PR #20, opened 2019-04-16
Add Laravel Echo Event Listeners

davidpiesse:
Allows use of Laravel Echo to fire an event in a component.
Is good for public, private and presence type channels.
Just ```event(new NewTodo($todo))``` or ```broadcast(new NewTodo($todo))``` using the normal Laravel means of broadcasting an event

Syntax for the listener key is currently:
```php
 'echo:todos,NewTodo' => 'updating', //public
 'echo-private:todos,NewTodo' => 'updating', //private
```

There are tests!

NOTE - I have not implemented some means to have dynamically named listeners - yet...


bbcreation:
Hi; When will you be able to dynamically add listeners?

davidpiesse:
@bbcreation have you got a use case for dynamic event listeners?
Maybe an example of how you would like to write one?

bbcreation:
I don't know how to work around the problem. I have to listen to many public channels at the same time in the same component. I want to read events from these feeds into html tags. I am building a time server with a friend and it would be good to listen to many at the same time. On one channel it transmits time to finish something and on the next channel it measures time for another thing. These two things relate to e.g. the same item. I will probably have to use the nest now.

PR #24, opened 2019-05-01
Allow custom attributes on validation.

cihantas:
The signature of `make` is `make(array $data, array $rules, array $messages = [], array $customAttributes = [])`. With the following changes users can specify custom attributes like they're used to without having to create a custom validator.


lancepioch:
@all-contributors please add @cihantas for code

PR #25, opened 2019-05-02
LivewireMakeCommand refactoring

hivokas:
This PR provides a refactoring for LivewireMakeCommand.

Regarding new features it adds:
- possibility to create component in a sub-directory;
- `--force` flag which can be used to force create a component.

PR #28, opened 2019-05-02
An ability to declare an alias inside a component

hivokas:
This PR adds an ability to declare an alias inside a component (instead of a declaring inside a routing file)  this way:

```php
class Hivokas extends LivewireComponent
{
    public static $alias = 'hivokas';

    public function render()
    {
        return view('livewire.hivokas');
    }
}
```

> should be accepted after #25


calebporzio (changes requested):
Thanks for your contribution. All that "finder"/reflection code is really helpful. Couple thoughts before merging though.

lancepioch:
@all-contributors please add @hivokas for code

Issue #43, opened 2019-07-12
Setup CI linting


fgilio:
Just as a reminder.

**Describe the solution you'd like**
Something like StyleCI or similar. So there's no fuss around code formatting once the repo is public.

calebporzio:
Sorry, by "Laravel Codebase" I meant: vendor/laravel/framework, not just a new laravel project. Make sense? I wonder if it would still pass from that folder?

wotta:
I will check out if the linter passes on de vendor/laravel folder. I will let you know what the results are.

calebporzio:
StyleCI and Travis are now implemented

fgilio:
Niice!

PR #50, opened 2019-07-23
Added phpcsfixer

wotta:
This PR is for #43.

So I included the phpcsfixer of friends of php and dropped in the laravel-shift configuration.
I ran this against Livewire and there were some errors which I will show in a screenshot.

To get full benefits from phpcs and maybe in the future even more CI tools, I would suggest using a build for the package that will test the code and run all linters.

This preset is copied from: https://gist.github.com/laravel-shift/cab527923ed2a109dda047b97d53c200.

The following screenshot is the result of a dry-run. This means that it didn't fix the errors that are shown.
It can however fix those errors pretty easily.
Hyper 2019-07-23 07-34-56

### DIFF
--- a/.php_cs.php
+++ b/.php_cs.php
@@ -3,6 +3,13 @@
 use PhpCsFixer\Config;
 use PhpCsFixer\Finder;
 
+/**
+ * For more information about this configuration you can go to
+ * the following website to see more about this config when
+ * you click on the "folder" icon on the website.
+ *
+ * https://mlocati.github.io/php-cs-fixer-configurator/#configurator
+ */
 $rules = [
     'phpdoc_indent' => true,
     'binary_operator_spaces' => [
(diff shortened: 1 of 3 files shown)


tabuna:
Hey.

Why not use the tools for all participants, such as StyleCI, this would be a long-playing solution. Instead of users taking care of it every time.

I see that there is such an offer in the issue, but I don’t understand why not to use it.

wotta:
@tabuna because Caleb needs to setup a service that runs the ci. And what tools are used is up to him.

I suggest this tool incase he will be using a more complete ci like travisci.

I do understand your question tho.
Hope I gave you a somewhat useful answer.

calebporzio:
yeah I definitely want this handled in CI, as well as running tests, I just haven't taken the time to do it. I will be out of town until next week and don't really see this as too urgent.

But I'd welcome a PR for the addition. 

I think we should go with what laravel/framework uses: StyleCI and Travis for testing

wotta:
I can adjust the pr accordingly but I can't setup those services. So maybe decline this pr for now?

And reopen when your back?

PR #53, opened 2019-07-23
Fix: "route:cache does not work"

joriskuijpers:
Fix for: https://github.com/calebporzio/livewire/issues/51

Haven't provided test or so but only copied the code into a HttpController.

### DIFF
--- a/src/Http/Controllers/AssetController.php
+++ b/src/Http/Controllers/AssetController.php
@@ -2,7 +2,7 @@
 
 namespace Livewire\Http\Controllers;
 
-class AssetsController
+class AssetController
 {
     public function __invoke()
     {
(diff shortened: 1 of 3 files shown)


wotta (approved):

francislavoie:
This should probably use `response()->file()` instead. See https://laravel.com/docs/5.8/responses#file-responses

The difference is that using `file_get_contents()` will load the file into memory before sending it, whereas using `file()` will stream the contents. It uses https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/BinaryFileResponse.php under the hood which uses `stream_copy_to_stream()`.

lancepioch:
@all-contributors please add @jjhamkuijpers for code

PR #59, opened 2019-07-25
Removes `composer.lock`

nunomaduro:
This pull request addresses the removal of the file `composer.lock` as it's important to run local tests against the latests versions of the dependencies. Also, since this a package, you don't need this.


ahinkle:
Wouldn’t be a bad idea adding it to`.gitignore` too. 👍

calebporzio:
Thanks fellas, will add to .gitignore

lancepioch:
@all-contributors please add @nunomaduro for code

PR #74, opened 2019-07-26
Remove wire data from assert see

### LINKED ISSUE #62: Testing Components can generate false positives from `wire:data`
**Describe the bug**
Calling `assertSee` on `Livewire::test(Component::class)` can produce false positives when looking for text on the page that exists within the `json` payload in `wire:data`.

**To Reproduce**
1. Create a component `Post` through the artisan command
2. Set up the route `Route::livewire('/posts', 'posts')` that lists a set of Posts with a `title` and `body` (just some basic data). Do not set up the template to show any Posts.
3. Write a test that checks that the first Post in the list is present on the page by checking the title appears using `assertSee`.
4. The test will pass.

**Expected behavior**
The test should fail because the first `Post`'s title shouldn't be listed on the page.

**Screenshots**
N/A

 - OS: MacOS
 - Chrome
 - Latest of both

**Additional Info**
I believe that because the models are being json encoded in `wire:data`, it ends up providing a false-positive when checking the DOM for text coming from the model.

pddevins:
Performing a basic `preg_replace` on the html to remove the `wire:data` attribute. Since many of the values within the data attribute are likely to be printed on the page, it's possible over the course of testing to receive a false positive that some text is visible on the page because it exists in this attribute in the dom. 

Should resolve [Issue 62](https://github.com/calebporzio/livewire/issues/62)

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -23,7 +23,7 @@ trait MakesAssertions
 
     public function assertDontSee($value)
     {
-        PHPUnit::assertStringNotContainsString((string) $value, $this->dom);
+        PHPUnit::assertStringNotContainsString((string) $value, preg_replace('(wire:data=\".+}")', '', $this->dom));
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @pddevins! I added some tests to this. Good PR!

lancepioch:
@all-contributors please add @pddevins for code

Issue #84, opened 2019-07-29
404 for livewire.js when not using php artisan serve


2kodes:
**Describe the bug**
I have the counter livewire example app  being served from the htdocs folder of my local xampp.
When I access the app I get a 404 for the livewire.js file. 
However when I use use `php artisan serve` to serve the app and access it the livewire.js is served correctly

**To Reproduce**
1. Create a new laravel app.
2. Install livewire in it
3. Serve or run it with a normal web server instead of `php artisan serve`
4. See error in the network tabs of the browser

**Expected behavior**
Expected the path to the livewire.js file to be resolved correctly no matter where itsa being served from

**Screenshots**

**Error when served from localhost with a webserver**

![liveiwire-2](https://user-images.githubusercontent.com/2084398/62039617-95c83b00-b1e7-11e9-8b44-5d819bfe65a9.png)

**Loaded correctly when served with `php artisan serve`**

![liveiwire-3](https://user-images.githubusercontent.com/2084398/62039682-abd5fb80-b1e7-11e9-9725-4c41a9bf3254.png)

**Desktop (please complete the following information):**
 - Browser - chrome

lancepioch:
I have this running with Apache just fine

calebporzio:
This has been addressed in the latest release: v0.1.1

2kodes:
@calebporzio ... sweet..will test and feedback..thanks

Krato:
Is giving me same error. Using artisan works fine.

Using a domain with vhost in my local, gives me a 404 NOt found. Route is registered but no working. :(

robsontenorio:
In my case i just run my localhost on 8080.
I can confirm. Error 404.

robsontenorio:
Just notice when publishing config files it works:

```
php artisan vendor:publish --tag=livewire:assets
```

According docs this **is not** mandatory.

faustbrian:
@robsontenorio is this still an issue?

lancepioch:
I tested this with Apache and this is not an issue. (same setup from [here](https://github.com/livewire/livewire/issues/243#issuecomment-565830288))

My guess is that OP is having issues because of using subdirectories in their url `/public/` instead of pointing the root path to the public directory itself.

I believe this can be closed.

calebporzio:
Closing this for now, thanks @faustbrian and @lancepioch for digging in and resolving.

RafaTorices:
This not work for me. I have error 404, 
how can i solve it?
[...]

PR #86, opened 2019-07-30
Allow livewire.js to be loaded from root or subfolder

twokodes:
Yes. #84 
No
No

`livewire.js` had 404 error when the livewire app was not served by `php artisan serve` or from the root of a webserver

This change allows the `livewire.js` file to be loaded irrespective of whether the app is being served from a subfolder or the root folder of the web server or `php artisan serve`

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -60,7 +60,7 @@ class LivewireManager
         return <<<EOT
 <!-- Livewire Assets-->
 <style>[wire\:loading] { display: none; }</style>
-<script src="/livewire{$versionedFileName}"></script>
+<script src="livewire{$versionedFileName}"></script>
 <script>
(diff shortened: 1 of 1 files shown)


davidpiesse:
Very cool!
Do you think the livewire() needs a bit more functionality around this? Such as the ability to force root?

D

2kodes:
@davidpiesse 
I think that will be great. also checked out your PR. I think thats better since on further testing my PR doesnt work. Even though livewire.js was loaded correctly, the http requests to /livewire were unable to go the right url since the root of the app was not given.

calebporzio:
This has been addressed in the latest release: v0.1.1

Thanks for the contribution though!

2kodes:
no problem..you are welcome..thanks too

PR #87, opened 2019-07-30
Allow requests to be made correctly

twokodes:
Yes. 
No. However this is linked to #86 
No

Currently requests made to `/livewire/` will 404 when the livewire app is not being served from the root folder of a webserver or being run with `php artisan serve`

This fix allows the request to be made to the correct root path of the app irrespective of the hosting path

### DIFF
--- a/src/js/connection/drivers/http.js
+++ b/src/js/connection/drivers/http.js
@@ -7,7 +7,7 @@ export default {
     },
 
     keepAlive() {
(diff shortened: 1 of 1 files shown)


calebporzio:
This has been addressed in the latest release: v0.1.1

calebporzio:
Thanks for the contribution though!

PR #89, opened 2019-07-31
Maintain url and asset path from Laravel config.

loganhenson:

Encompasses the requirements (while maintaining current behavior) for PRs:
#86  
#87 

No.

No. Manual tests only.

We should enable the user to be able to customize the `APP_URL` and `ASSET_URL` without losing the functionality of livewire.

For example (tested using the php built in webserver to simulate non-root paths):
` php -S localhost:3000` run from `/Users/myuser/Desktop` where livewire-pr is located at `/users/myuser/Desktop/livewire-pr`.

> .env
```
APP_URL=http://localhost:3000/livewire-pr/public/index.php
ASSET_URL=http://localhost:3000/livewire-pr/public/index.php
```

Or if they are serving their app from a subdirectory via apache.

This should allow users more flexibility in their use of Livewire.


calebporzio:
@loganhenson - I made a couple changes to the PR. Here were my reasons:

1. I moved the definition of `$assetsPath` from the directive method, to the `LivewireManager::assets()` method. This way, we won't break people's apps (they would need to run `artisan view:clear` because we modified the directive). Also, if we decide to allow an override via passed in `$options` in the future, this will be easy.
2. I removed the `$url` (config('app.url')). reference. Reason being: in most of my apps in development, I leave the `APP_URL` env variable to `localhost`. By hardcoding `localhost` in the `livewire/message/` endpoint, I would get errors in javascript because I serve my sites with Valet. Am I wrong in thinking it's best to keep this flexible? (just `/livewire/message` instead of `http::/localhost/livewire/message)? How often are people putting a non-root app_url? is that a problem I missed, or is it just people having a non-root `asset_url`?

calebporzio (approved):

calebporzio:
Ok, I now see that #87 needs `/livewire/message` to reference a non-root `APP_URL`.

Is there a way to both not break new `laravel new` apps being served by valet AND non-root URL apps?

We could allow an override to be passed in: `@livewireAssets(['app_url' => config('app.url))`.

OR we could detect if the .env file contains the default: "http://localhost" for APP_URL, and if so, just use the root `/livewire/message`.

That sounds nasty, but I actually kind of like it, hmm...

calebporzio:
Ok, I added the conditional to check for the Laravel default. I've curbed the dirtiness I feel by adding tests.

Thanks a bunch @loganhenson

lancepioch:
already added for code to contributors 🔢

PR #92, opened 2019-07-31
Container bindings for Component Actions

jrseliga:
**Needed :upside_down_face:**
**No, just wanted to put together a proof of concept**

**No**

**Yes**

**This will allow `Component` actions to type hint parameters that will be resolved out of the container while preserving the current affordance of passing values directly to the action.**

**Assume `Banana@peel()` is called like so:**
```php
peel(.75);
```

```php
use Animals\Monkey;
use Livewire\Component;

class Banana extends Component
{
    public function peel(Monkey $monkey, $percentage)
    {
        dd(get_class($monkey)); // \Animals\Monkey
        dd($percentage); // .75
    }
}
```
> **If this looks good and gets merged, I'll probably follow-up with some sort of `BindsMethodParameters` trait that both `Component` and `LivewireManager` can leverage so we can bind from the container in `mount` as well.**

> **This implementation could potentially also be leveraged to rehydrate models, i.e. when parameter of an `Action` is typehinted to a `Model` and the parameter name matches a `public` property on the `Component`.**

> **Getting a little ahead here but worth acknowledging upfront that if the original state of the `Component's` `public` property that is a `Model` contained loaded relationships, rehydrating those relationships could be tricky**

### DIFF
--- a/src/Concerns/HandlesActions.php
+++ b/src/Concerns/HandlesActions.php
@@ -60,6 +60,17 @@ trait HandlesActions
         }
     }
 
+    protected function resolveActionParameters($method, $params)
+    {
+        return collect((new \ReflectionMethod($this, $method))->getParameters())->map(function ($parameter) use ($params) {
+            return rescue(function () use ($parameter) {
+                return app($parameter->getClass()->name);
+            }, function () use ($params) {
+                return array_shift($params);
+            }, false);
+        });
+    }
+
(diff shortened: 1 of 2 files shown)


calebporzio:
Great PR, seriously - thanks

lancepioch:
@all-contributors please add @jrseliga for code

PR #93, opened 2019-07-31
Support Laravel redirects in Components

jrseliga:
**Needed :upside_down_face:**
**No, just wanted to put together a proof of concept**

**No**

**Yes**

**This will allow `Component` actions to use the core Laravel `Redirector` inside of a `Component` action, while preserving the current built-in redirect behavior in `Component`**

```php
use Livewire\Component;
use Illuminate\Support\Facades\Redirect;

class Banana extends Component
{
    public $peeled = false;

    public function peel()
    {
        $this->peel = true;

        return redirect()->route('monkeys.index');
        // or
        return redirect('/monkeys');
        // or
        return Redirect::route('monkeys.index');
        // or
        return Redirect::to('/monkeys');
        // or
        return $this->redirect('/monkeys');
    }
}
```
> **A little unsure of the safety of this implementation, especially if additional third party packages extend `Redirector`. I _think_ it should be safe in that case.**

> **Also feels a little wonky directly extending `Redirector`, this was done because core `Redirect@route and redirect()` offload behind the scenes to `Redirector@to` so we get both `to()` and `route()` for the price of one overridden method.**

> **Just putting this up for feedback. :tada:**

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -32,6 +32,10 @@ class LivewireServiceProvider extends ServiceProvider
 
     protected function interceptRedirect()
     {
+        if (!$this->isLivewireRequest()) {
+            return;
+        }
+
         $this->app->extend('redirect', function ($redirector) {
             $component = collect(debug_backtrace())->pluck('object')->first(function ($object) {
                 return $object instanceof Component;
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -15,16 +15,30 @@ use Livewire\Connection\HttpConnectionHandler;
 use Livewire\LivewireComponentsFinder;
 use Livewire\Macros\RouteMacros;
 use Livewire\Macros\RouterMacros;
+use Livewire\Routing\Redirector;
 
 class LivewireServiceProvider extends ServiceProvider
 {
     public function register()
     {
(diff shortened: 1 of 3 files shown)


calebporzio:
This is extremely interesting. I think I like it actually.

One modification I would make:

Instead of scouring the backtrace to detect if it's a livewire request, you can use the `LivewireServiceProvider::isLivewireRequest` method to detect if it's a Livewire request. THEN maybe get the component out of the stack trace or with some other method so you aren't running that logic on every request in the app.

jrseliga:
@calebporzio good call, definitely felt a little grim tracing every request of an application.

Does checking for `isLivewireRequest` make the ternary in the `extend()` unnecessary now? Seems like we would be guaranteed to have a `Component` in the trace

calebporzio:
Nah, I think I still dig it

calebporzio:
Really solid PR, great work here. Thanks

calebporzio:
@jrseliga - for what it's worth, I changed the strategy of this PR after merging it if you care to look.

I put the redirect logic inside "ConnectionHandler" so we don't have to do any conditional for "isLivewireRequest" AND we don't have to get the component out of the stack trace - it's already available.

lancepioch:
already added for code to contributors 🥇

PR #96, opened 2019-07-31
dump() support similar to error modal

allmanaj:
This pull request adds support for the `dump()` function in livewire components. It adds a dump watcher to the framework (similar to how Laravel Telescope works), stores the dumps made in the current request, and returns them from the request to the javascript where they're displayed in a modal similar to how the errors are displayed in Livewire.

This will enable the developer to follow data flow through the app easier and output values whenever they like! When multiple requests are made containing multiple `dump()` calls, they are each appended to the existing modal so as not to have lots of modals piled on top of each other.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
     ],
     "require": {
         "illuminate/support": "^5.7",
-        "illuminate/validation": "^5.7"
+        "illuminate/validation": "^5.7",
+        "symfony/var-dumper": "^4.1"
     },
     "extra": {
         "laravel": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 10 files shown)


calebporzio:
Thanks for the contribution @allmanaj!

So I was all set to pull this in, but started to add tests for the feature and realized this will break `dd()` for people.

Because you are changing the VarDumper handler to just store the dumps until a later point in execution, dd will add the dump to VarDumper, but will die() and never get a chance to do anything with the dumps.

We "could" just force people to use `dump()` and tell them in the docs, but I want Livewire to cater to a Laravel dev's intuition. And mostly that's dd().

Anyhow, I actually just built an implementation that is really simple and detects if the ajax request response is from dd(), and if it is, shows a modal with the dd contents. Super simple.

Thanks for the push on the feature! And thanks for being willing to contribute!

allmanaj:
Ah, that's fair enough! Saw your preview of it on twitter and it looks great! No problem, I'm just glad to have had even some little effect on the direction of the framework

PR #98, opened 2019-08-01
Stream the livewire.js file instead

francislavoie:
Nope, just something I noticed while browsing through old PRs while procrastinating instead of going to bed.

Nope

Nope, I haven't tried it either to confirm if it works, just made the edit in GitHub.

As mentioned in my comment here https://github.com/calebporzio/livewire/pull/53#issuecomment-517119194, this should stream the content of the file instead of loading it into memory before sending it off. Also, it handles Last-Modified itself, so that code can be removed.

### DIFF
--- a/src/LivewireJavaScriptAssets.php
+++ b/src/LivewireJavaScriptAssets.php
@@ -7,15 +7,12 @@ class LivewireJavaScriptAssets
     public function __invoke()
     {
         $file = __DIR__ . '/../dist/livewire.js';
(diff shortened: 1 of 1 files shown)


calebporzio:
That's awesome! Glad I don't have to set that lastModified manually anymore. I tested it, and the caching still works - thanks!

lancepioch:
@all-contributors please add @francislavoie for code

Issue #69, opened 2019-07-26
wire:loading delay


jaysson:
**Is your feature request related to a problem? Please describe.**
I want to show loading indicator only when request takes more than x milliseconds (usually 300 ms).

**Describe the solution you'd like**
Just like CSS transitions, `wire:loading.300ms` support would be great.

**Describe alternatives you've considered**
There is always the option of using JS, but I believe that's beside the point of this library.

**Additional context**
Turbolinks has same behaviour. The loading indicator is shown only if request takes more than 500ms.

calebporzio:
I agree I wanted this behavior too.

Do you think we should add a default timeout for all `wire:loading` or just let people add it themselves? (I could see arguments both ways)

I also don't think this will be too difficult to implement

jaysson:
I think it's better to let people specify the delay. Some people may want to show it after 300ms, others after 500ms and so on.

davidpiesse:
So would you expect something like this?
```html
<div wire:loading.delay.750ms >
    Processing Payment...
</div>
```
Where the loading state would only activate if it has not responded within the first 750ms?
I would also expect to have a default value (like wire:poll) of 500ms for the simpler ``` wire:loading.delay ```

Any initial thoughts on this style implementation? Would it interfere with other loading methods like wire:loading.class? I'm not sure if you can currently do a double like below either.
```html
<div wire:loading.delay.750ms wire:loading.class="some-class" >
    Processing Payment...
</div>
```

davidpiesse:
OK :)

Add two new modifiers to wire:loading directive:

- after [if after Xms you still don't have a response change the loading state]
- min [show the loading state for a min on Xms]

You can use the same 500ms / 2s syntax as the wire:poll (uses the same code)
after defaults to 500ms, and min defaults to 250ms

You won't be able to use both at the same time realistically as they oppose one another's raison d'etre (reason they exist)

The bonus to this is we will hopefully be able to test and prove that you can split the loading logic between two directives on the same element like this:
```html
<div wire:loading.class="foo" wire:loading.min.300ms ></div>
```

calebporzio:
Closing this due to inactivity

PR #111, opened 2019-08-02
Add loading delay and after

davidpiesse:
Yep #69 
Just a loading.min and a loading.after

Not yet..

**This is a first draft.**
It splits up the functionality in setLoading() and needs a look at with a fresh pair of eyes.
They both work as it stands using just two additional timers.

loading.min - Will set the minimum time a loading state can be active for (defaults to 500ms if no duration is set)

loading.after - Will delay the loading state for a duration (default 500ms) once a request has been sent to Livewire.

### DIFF
--- a/src/js/Component/index.js
+++ b/src/js/Component/index.js
@@ -5,6 +5,7 @@ import DOM from '../dom/dom'
 import DOMElement from '../dom/dom_element'
 import nodeInitializer from "../node_initializer";
 import store from '../store'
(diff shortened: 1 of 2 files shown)


davidpiesse:
This needs to wait for reset loading error to be pulled in first as I have changed the way loading states are set/unset.
I think I need to look at this again. It works but I don't know how to test this properly, and the code doesn't feel nice and squishy and fun to me yet.

PR #112, opened 2019-08-03
[WIP] Add protected properties to session

calebporzio:


tillkruss:
@calebporzio: The props are grouped into a single write/read call, right?

Issue #127, opened 2019-08-05
Component mount hook method doesn't receives method bindings when rendered


webong:
```
class Bar
{
    public $input = 'output-from-method-binding';

    public function output()
    {
        return $this->input;
    }
}

class Foo extends Component
{
    public $name;

    public function mount(Bar $bar)
    {
        $this->name = $bar->output();
    }

    public function render()
    {
        return view('livewire.view');
    }
}
```
Above is a LiveWire Component `Foo`, on mount it receives a `Bar` class

```
 // routes/web.php

  Livewire::component('foo', Foo::class);

  Route::livewire('/test1', 'foo');

```

If `/test1` route is visited, Laravel Route is able to resolve the dependency on the mount hook method. But if the livewire component is rendered as a blade directive, the mount method bindings are not received. 

```
 // resources/view/render.blade.php
<div>
    @livewire('foo')
</div>
```
```
 // routes/web.php
  Route::livewire('/test2', 'render');
```
Now if `/test2` is visited, we get an error *Too few arguments to function App\Http\Livewire\Foo::mount()*. This means the mount method dependecy is not resolved.

PR #128, opened 2019-08-05
Fix component mount hook method bindings

webong:
Yes #127 

It adds a DependecyResolver trait and updates the LiveWireManager mount method to resolve the class method bindings.

Yes a test is added

With this fix, livewire components are able to mount dependencies when rendered as blade directive, as this was only possible when components were rendered as a route view

### DIFF
--- a/src/Concerns/DependencyResolver.php
+++ b/src/Concerns/DependencyResolver.php
@@ -6,7 +6,7 @@ use ReflectionParameter;
 use Illuminate\Support\Arr;
 use ReflectionFunctionAbstract;
 
-trait DependencyResolverTrait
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks @webong for the contribution.

I'm guessing you copied RouteDependencyResolverTrait from Laravel instead of using it because of it's dependency on "$this->container". I messaged your code so that we don't have to maintain a copy of that trait (or maybe it'll end up biting us the other way around (depending on the framework)).

Anyhow, good PR - thanks.

lancepioch:
@all-contributors please add @webong for code

PR #131, opened 2019-08-06
avoid collision with component name property

faustbrian:
Yes https://github.com/calebporzio/livewire/issues/130

No

Yes

At the moment the base component has a very generic property called `$name` which can be used to modify the component name. This could easily be overwritten and hard to figure out why a wrong component name is now used to resolve the view name.

The newly introduced `getName()` method allows the use of a custom name that will be used by `name()` and apply all the same methods as perform like diffing and imploding. This should avoid issues with components that themselves use a `$name` property.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -19,7 +19,6 @@ abstract class Component
 
     public $id;
     public $redirectTo;
-    protected $name;
     protected $lifecycleHooks = [
         'mount', 'updating', 'updated',
     ];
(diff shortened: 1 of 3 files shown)


faustbrian:
@calebporzio thinking about it now looking at the code. Shouldn't it be enough if we completely remove the name property and add some docs for how to overwrite `name()` to use a custom component name?

That way the default assumption of a component would be that you follow the Livewire naming and location conventions and if you don't you will overwrite the `name()` method.

This could in the future maybe be expanded by allowing to use a global custom name resolver.

calebporzio:
@faustbrian - that makes a lot of sense.

Should we remove the `$name` property AND rename `->name()` to `->getName`? It would feel similar to a Model's `->getTable()`. What do you think?

faustbrian:
Sounds good to me, I will make the change in an hour when I am at a computer.

faustbrian:
@calebporzio renamed `name()` to `getName()` and added another test to make sure default and custom names work.

calebporzio:
Perfect - great PR thanks @faustbrian !

lancepioch:
@all-contributors please add @faustbrian for code

Issue #134, opened 2019-08-06
Pass data to updating/updated methods


marksalmon:
**Is your feature request related to a problem? Please describe.**
Less a problem than an inconvenience. With components that contain a lot of properties you can end up writing many identical/similar ```updatedFoo``` methods.

**Describe the solution you'd like**
I would like the generic ```updated``` / ```updating``` methods in a livewire component to receive data, so that logic can be performed conditionally depending on the name or value of the updated property

**Describe alternatives you've considered**
The ```updatingFoo``` and ```updatedFoo``` methods already receive ```$value``` so you can work round this just by writing a specific method for each property

**Additional context**
* Example (pseudo-code)

```php
public function updated($data)
{
    // ... do some validation.

    $this->model()->persistToDatabase($data['name'], $data['value']);
}
```

calebporzio:
This has been merged and added to the framework (will be in the next release):

```
public function updated($name, $value)
{
    // ... do some validation.

    $this->model()->persistToDatabase($name, $value);
}
```

PR #135, opened 2019-08-06
Pass data to updating updated methods

marksalmon:
#134 

Nope

Yep

This passes the $data array to the updating/updated methods on the livewire component so that logic can be performed in those methods dependant on the property or value that has been updated

### DIFF
--- a/src/Connection/ConnectionHandler.php
+++ b/src/Connection/ConnectionHandler.php
@@ -44,7 +44,7 @@ abstract class ConnectionHandler
 
     public function processMessage($type, $data, $instance)
     {
-        $instance->updating();
+        $instance->updating($data);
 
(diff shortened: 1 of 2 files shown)


calebporzio:
@marksalmon!!! Thanks for the PR.

I change the functionality a little bit as you can see and messaged the test a little.

instead of passing in `$data` to the updated property. I thought it would be cleaner to pass in `$name, $value` as separate params.

Also, I decided to scope `updated` and `updating` to only Livewire calls to update properties. Before they would run before and after actions, but I don't think that makes sense as you already have action methods to hook into, and I think people would assume this is how it should work.

Anyhow, thanks!

lancepioch:
@all-contributors please add @marksalmon for code

Issue #137, opened 2019-08-06
 ModelNotFoundException and NotFoundHttpException not handled in the component mount method


webong:
```
class Store extends Component
{
    public function mount()
    {
        abort(404);
     }

    public function render()
    {
        return view('livewire.store');
    }
}
```

```
Route::livewire('/test', 'store');
```
Visiting `/test` throws a whoops error page instead of a 404 page.

calebporzio:
`abort(404)` now works in components. As well as `abort(500)`.

This is in the latest release: v0.2.5

PR #138, opened 2019-08-07
Handle ModelNotFoundException and NotFoundHttpException in component mount hook

webong:
Yes #137 

No

Yes

Currently, ModelNotFoundException and NotFoundHttpException are not handled in the mount hook method. As a result of this, we cannot abort a livewire request or throw an exception when a model item is not found.
This PR wraps a try and catch block on when the component mount hook is called and updates the livewire blade directive too to verify the correct response type.


webong:
Hello, @calebporzio can you give this a second look ... The LivewireManager no longer extends the Handler class, but now gets an instance of the exception handler from the application instance.

calebporzio:
Hi @webong, thanks for the PR. It's an interesting problem you've described.

First, just  to make sure, the "route model binding" with livewire:
`Route::livewire('contacts/{contact}, ...)`, will throw and render a true 404 to the user like so:

image

As far as providing support for the same behavior in the `mount()` method, I think you're right that I would expect it to behave the same way, HOWEVER, there are some issues with your solution and I don't see a clear way around it.

The problem: You are basically bypassing the App's exception handling process (or I should say short-circuiting it). For example, if the app has bug tracking set up, the `->report()` method will never be called with this approach, the `->render` method is the only thing that is called.

This solution feels a bit off to me.

Is there any way you can think of to let the exception bubble all the way up to the exception handler?

It might just be a problem with Laravel itself and maybe a PR to the framework is warranted to allow running `abort(404)` inside the Blade compiler.

calebporzio:
I added some code to hijack the part of blade that wraps exceptions in `ErrorException` and made exceptions for NotFoundHttpException (and a bunch of other ones).

Fixes are available in v0.2.5

PR #141, opened 2019-08-07
forget flashed session content if we are not redirecting

faustbrian:
Yes https://github.com/calebporzio/livewire/issues/132

No

Yes

As described in https://github.com/calebporzio/livewire/issues/132 all flashed contents of a session is persisted for 1 in-browser page reload even though Livewire already rendered the contents.

This PR changes the connection handler to assume that if there is no redirect it means that the contents was already rendered by the component itself. This change results in 2 scenarios now.

1. We call `session()->flash()` inside our component but don't call `$this->redirect()` so the flashed content is removed after it has been added to the AJAX response.
2. We call `session()->flash()` inside our component and call `$this->redirect()` so we are redirected, the content is flashed and removed by Laravel and its session handler.

### DIFF
--- a/src/Connection/ConnectionHandler.php
+++ b/src/Connection/ConnectionHandler.php
@@ -30,7 +30,7 @@ abstract class ConnectionHandler
         $events = $instance->getEventsBeingListenedFor();
         $eventQueue = $instance->getEventQueue();
 
-        return new ResponsePayload([
+        $response = new ResponsePayload([
             'id' => $payload['id'],
             'dom' => $dom,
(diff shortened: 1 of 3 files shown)


faustbrian:
@calebporzio any chance of getting a review on this?

calebporzio:
@faustbrian, sorry for the lag on reviewing this. Like all your contributions, the problem is clearly stated and the solution is smart, clear, and well tested. Thanks a ton!

lancepioch:
faustbrian has already been added for code

PR #149, opened 2019-08-08
add hydrate to $lifecycleHooks of component

faustbrian:
Yes and No

No

No

https://github.com/calebporzio/livewire/commit/4c4a0ba589ad3aea29e1139e485a5d3545136abf introduced the `hydrate` lifecycle hook but didn't register it which results in `Method XXX::hydrate does not exist.` because the `__call` method doesn't catch the call.

The tests didn't catch this because the test file has an actual `hydrate` method defined.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -20,7 +20,7 @@ abstract class Component
 
     public $id;
     protected $lifecycleHooks = [
-        'mount', 'updating', 'updated',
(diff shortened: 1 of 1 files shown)


calebporzio:
Shoot, good call - thanks - patch released: v0.2.1

lancepioch:
already added for code, cheers

PR #152, opened 2019-08-09
JS Dirty Input detection

marksalmon:
I'm not sure. I did this to scratch my own itch and thought it might be useful for others.

No, 1 distinct feature

Yes. My first attempt at Jest tests

This feature adds a wire:model modifier called 'dirty'  eg. ```<input wire:model.dirty="foo">```

This modifier will prompt livewire js to compare the current input value with the value in the component data, on every input event, and add a 'dirty' class to the input if they are different.

This is useful for inputs or textareas, especially when using the 'lazy' modifier, to display that the input has been changed but not yet synced with the livewire component. The class will be removed on subsequent dom refresh, or if the input value is returned to its original value.

Optionally, the user can specify a different element to receive the dirty class, by applying the class ```dirty-target```

I know it is possible to do all of this using existing livewire features, but it does require a back-end round trip for every input event. I also know you already have the concept of dirty properties, so perhaps I need to come up with a different name for this feature.

I'd be happy to submit a PR to the docs if required.


calebporzio:
woah @marksalmon - I really dig this.

What about:

```
<input type="text" wire:dirty.class="bg-red">
```
and for a remote target
```
<input type="text" wire:ref="input">

<div wire:dirty.class="something" wire:target="input">
```

Basically, I would want it to behave the same way as `wire:loading`

If you add `wire:dirty` it will be hidden by default until it's dirty, and so on and so forth.

Thoughts?

calebporzio:
I like the idea of giving the user feedback without writing back to the backend. I actually love the idea.

marksalmon:
Excellent. I'm on the case.

marksalmon:
I think this is all working now. It goes like this:

_During component initialization_
1. Capture `wire:dirty` directives during walk of component in `initialize()`
2. If `wire:target` directive exists, store element in component's `dirtyElsByRef` object, keyed by target, or else store self in component's `dirtyEls` array
3. After all elements are initialized, add an event listener on the component element listening for input events.

_When an input event is triggered_
1. Check for target elements in the component's `dirtyElsByRef` object and `dirtyEls` array
2. Perform dirty logic and add/remove class/attribute/visibility changes to each of the targeted elements

calebporzio:
Hey Mark,

Couple points:
* The tests are passing, but the element with the dirty directive and no modifier will be hidden by default and shown when dirty" test isn't accurate and when I changed the test to actually test that behavior, it fails. (and I tested in real life)
* There are merge conflicts with master, you will need to resolve those before I can merge (Run `git merge master` locally, and work through the conflicts)
* The tests check that dirty classes are added and removed when a value is changed and then changed back. It would be good to see the scenario of: updated a field, dirty is shown, `wire:model` updates from server and dirty class is removed again. (as this is the more common scenario for `wire:dirty`)

This PR is looking good, and I'm excited for this feature. I just don't really have the time right now to spend too much time improving this PR myself. Thanks for pitching in Mark!

calebporzio:
It's in!

I tweaked a couple of things and refactored a bit. But the functionality and tests are there and great!

At some point, we may want to refactor the listener a bit. If I am seeing correctly, every single time anything is typed into an input with `wire:dirty` (or is a target), the entirety of the listener will run, including the "remove" class stuff which is probably not very good performance-wise, but we'll keep an eye on it.

Thanks for the good hard work on this Mark - great contribution

lancepioch:
@all-contributors please add @marksalmon for code

PR #153, opened 2019-08-09
Add loading.after and loading.min functionality

davidpiesse:
Yep #69

Just a loading.min and a loading.after

Yes...

This is a second draft.
It splits up the functionality in setLoading() and needs a look at with a fresh pair of eyes.
They both work as it stands using just two additional timers.

loading.min - Will set the minimum time a loading state can be active for (defaults to 500ms if no duration is set)

loading.after - Will delay the loading state for a duration (default 500ms) once a request has been sent to Livewire.


davidpiesse:
I think I am going to submit this working but with commented out tests... I can't get both to work at the same time; so basically my test suck!

davidpiesse:
No worries.

I will look into moving logic into the LoadingManager :)
I will remove the defaults from the documentation and reorder them.

The commented out test was the one that passes but sometimes fails. The issue is delaying an assert (await/async) in the middle of a request and checking the loading state is still alive (IIRC)...

I'll keep plodding along until its right 🎆

calebporzio:
Closing for now.

dubcanada:
@calebporzio - is there interest in having this in core?

I'd like this functionality so if there is interest I'll take over development unless @davidpiesse has it somewhere?

mblarsen:
Yeah, it is really nice functionality that I take care to implement when working with React or Vue. Like only showing loading if it takes more than some time and if showing loading show it at least for some time (to avoid flickering).

davidpiesse:
```wire:loading.delay``` is Coming in V2... 😄 
https://twitter.com/calebporzio/status/1296204700879065088

PR #154, opened 2019-08-10
Add @livewireassets blade directive

intellow:
Not sure if you are interested in merging this, but I spent way too much time trying to figure out why nothing was working until I realized I forgot to capitalize the A in assets. If you add this version of the blade directive it might help someone else not waste time with that mistake either.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -77,6 +77,10 @@ class LivewireServiceProvider extends ServiceProvider
         Blade::directive('livewireAssets', function ($expression) {
             return '{!! Livewire::assets('.$expression.') !!}';
         });
+        
+        Blade::directive('livewireassets', function ($expression) {
+            return '{!! Livewire::assets('.$expression.') !!}';
+        });
 
         Blade::directive('livewire', [LivewireBladeDirectives::class, 'livewire']);
     }


faustbrian:
Just my personal opinion but I don't feel like adding new methods or directives for anything because of possible typos is a good solution.

The documentation is already there and states how the directive is named so your first step would, or should, be to check if you did the setup of Livewire correct.

intellow:
> Just my personal opinion but I don't feel like adding new methods or directives for anything because of possible typos is a good solution.
> 
> The documentation is already there and states how the directive is named so your first step would, or should, be to check if you did the setup of Livewire correct.

I think it aligns well with the developer experience of Laravel. For example, the official Laravel docs state that an accessor should be called like `$model->your_accessor`, however if you type `$model->yourAccessor` it still works. Some of the beauty of Laravel is that in many cases you can just guess what you need to type without looking at the docs and it works.

There is no standard convention for Blade Directives that I'm aware of. I've seen some package camel case and others use all lowercase. Why not support both?

calebporzio:
Hey @intellow, you raise a good point.

I like how composer offers both `dump-autoload` AND `dumpautoload` because most poeple can never remember which one it is.

In this case, I feel like this is not a common hangup, so I'm closing this for now, but If anyone else runs across this issue that will add weight in favor of merging it.

Thanks for the care and contribution @intellow, hope you're enjoying the framework!

PR #157, opened 2019-08-12
allow page composition through Route::livewire

faustbrian:
Probably. No.

No

Yes

I find myself quite frequently in a situation where I end up composing pages with multiple components that are just a list of chained `@livewire(...)` calls in a blade template. This PR allows to pass in components as either a string or an array and will loop over them to include them.

**Old**
```blade
@extends('layouts.dashboard')

@section('content')
    @livewire('update-contact-information')

    @livewire('update-profile-information')
@endsection
```

**New**
```php
Route::livewire('profile', [
    'update-contact-information',
    'update-profile-information',
])->name('settings.profile');
```

This new syntax makes it very easy to compose pages that consist purely of Livewire components without the need of having an extra blade template for it.

### DIFF
--- a/src/Macros/RouterMacros.php
+++ b/src/Macros/RouterMacros.php
@@ -2,6 +2,8 @@
 
 namespace Livewire\Macros;
 
+use Illuminate\Support\Arr;
+
 class RouterMacros
 {
     public function layout()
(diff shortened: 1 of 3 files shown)


lancepioch:
This looks super cool, two side questions:

1. Does route caching still work?
2. Is there a method for the @extends?

faustbrian:
1. Seems to still work fine as only the composition of the view changed.
2. Not sure what you mean. You can use the `layout` method or route group property to change that property if that is what you mean. https://livewire-framework.com/docs/spa-mode/

calebporzio:
This is interesting, I'm not sure how I feel about it though.

Does anyone else find this useful? (in real life, not just in theory)

faustbrian:
I would imagine something like below to avoid an array syntax in the second argument of the `livewire` call.

```php
// Global
Route::livewire('profile', [
    'update-contact-information',
    'update-profile-information',
])->withParam('something')

// Local
Route::livewire('profile', [
    'update-contact-information',
    'update-profile-information',
])->withParamFor('something', 'update-contact-information')
```

Issue #155, opened 2019-08-10
Using a subdomain does not work


intellow:
If you try to use livewire from a subdomain (like if you have a multi-tenant app), you get this error in the browser console.

Access to fetch at 'https://domain.test/livewire/message' from origin 'https://sub.domain.test' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

ceejayoz:
Do you have `app.url` explicitly set to `domain.test`?

intellow:
> Do you have `config('app.url')` explicitly set to `domain.test`?

Yes, my .env has `APP_URL=https://domain.test`. I tried updating my .env to `APP_URL=https://*.domain.test` and when I did that Livewire did not work at all. Does Laravel support that type of setting for multi-tenant apps? If so, I'm not aware.

edgrosvenor:
I've used this package to simplify CORS management before and it's pretty straightforward. It might be the key to your happiness.

https://github.com/barryvdh/laravel-cors

intellow:
@edgrosvenor I tried to install that package and add HandleCors to the global middleware. It doesn't work but it does change the error:

```
Uncaught (in promise) SyntaxError: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
    at livewire.js?id=e21774621afc85fe1711:1
```

calebporzio:
@intellow, This is usually caused by something weird coming back from a Livewire ajax request. Something non-standard. Can you inspect the XHR requests and post the raw output from the one at fault?

intellow:
@calebporzio Thanks for the tip. It didn't solve the problem, but the response is the full HTML of my login page. I imagine somewhere along the line there was some return of a redirect to the login page. I thought it might be some custom middleware I've written but I disabled all my custom middleware and it's still happening. Again, if I remove the subdomain from the URL (in which case my app still works), Livewire works fine.

edgrosvenor:
@intellow That explains your problem then. The expected response is JSON, not HTML. If you fully trust the server you're requesting from you can use JSONP which would allow you to receive the HTML page back instead of JSON. https://www.w3schools.com/js/js_json_jsonp.asp

edgrosvenor:
Oh... but if it's a Livewire ajax request that's receiving it, you probably can't do that.

[...]

PR #170, opened 2019-08-16
Allow custom base_url to be passed into @livewireAssets()

intellow:

No, but there is an open issue #155 

Nope

I added one test that I *think* will work but I couldn't actually get the tests to run. There's also probably at least one more test to write (comments in the test file) 

If you use subdomains in your app, there is a CORS issue that won't allow livewire to work unless the domain matches the config value. With this change, a user can send in their subdomain URL through the blade template and everything works.

:)

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -75,7 +75,16 @@ class LivewireManager
 
     public function assets($options = null)
     {
-        $appUrl = $this->appUrlOrRoot();
+        if(!is_null($options) && array_key_exists('base_url', $options)) {
+            $appUrl = $options['base_url'];
+            // I'm not sure what you're doing with this options variable, so if base_url is passed in...
+            // I set $appUrl to the passed in value, then remove it from $options so it doesn't get passed through on line 101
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for the PR @intellow, I merged it but changed the behavior kinda significantly.

I am now using the relative root: `src="/livewire/livewire.js"` for everything unless you pass in a `@livewireAssets(['base_url' => '...'])`.

I think there will be fewer problems for most people this way. And the certain people who are serving from a non-root endpoint can just pass in the URL. Now subdomains won't get the CORS issue.

Let me know if you agree.

Thanks for the contribution!

intellow:
Sounds great to me. I think that makes it even easier for local environments and prod as well.

2kodes:
Curious.. Will there be a config file for livewire in the future?

calebporzio:
@intellow - glad to hear it, and you're right.

@2kodes - I'm sure there will be, but I want to see how far we can push this zero-config thing.

lancepioch:
@all-contributors please add @intellow for code

Issue #182, opened 2019-08-22
string to be encoded contains characters outside of Latin1 range


erikscheepers:
I get this error when polling a result (containing characters like ♫ and ♪) from my database:

> Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range

I could fix the issue for me by replacing "_btoa(string)_" by "_btoa(unescape(encodeURIComponent(string)))_" using [this google-result](https://stackoverflow.com/questions/23223718/failed-to-execute-btoa-on-window-the-string-to-be-encoded-contains-characte)

Unfortunately I couldn't reproduce this error in a simple environment:
```php
namespace App\Http\Livewire;
use Livewire\Component;
class Counter extends Component
{
    public $string;

    public function mount()
    {
        $this->string = '♪&nbsp;♫&nbsp;♫';
    }

    public function hydrate()
    {
        $this->string = '♪&nbsp;♫&nbsp;♫';
    }

    public function render()
    {
        return view('livewire.counter');
    }
}
```
```html
<div wire:poll="$refresh">
    {!! $string !!}
</div>
```
Maybe somebody else has more luck, or my fix helps others experiencing the same error.

calebporzio:
I'd love a PR for that fix if it works for you

erikscheepers:
Issue is fixed

PR #186, opened 2019-08-23
Fix for encoding string containing characters outside Latin1

erikscheepers:
Replaced "btoa(string)" by "btoa(unescape(encodeURIComponent(string)))" to fix issue where the string to be encoded contains characters outside of Latin1 range. See issue #182 . 
I didn't do a thorough investigation on the fix, just used the fix presented by googling.


erikscheepers:
I'm not so fluent with git(hub), so I'm reopening this pull request. Let me know if I have to make a new one. I think my new version of livewire.js is correct: but please check or regenerate before testing.

I succeeded in reproducing my issue: 

```php
class Latin1 extends Component
{
    public $string;

    public function mount()
    {
        $this->string = "♪ &nbsp; abc š ë";
    }

    public function hydrate()
    {
        $this->string = "♪ &nbsp; abc ë š Ð ̪̪  Ͳ Җ ج ቜ \" ";
    }

    public function render()
    {
        return view('livewire.latin1');
    }
}
```

```html
<div wire:poll="$refresh">
    {!! $string !!}
</div>
```

In an earlier version of this pull request ``unescape`` was used. I tested the issue without this and with only using ``encodeURIComponent`` everything still looks good in Chrome and Firefox. 

Making a test for my fix (replacing "``btoa(string)``" with "``btoa(encodeURIComponent(string))``") is a bridge too far.

erikscheepers:
Problem is fixed in pull request #211

PR #197, opened 2019-08-27
Chore/es6 updates

leemason:

This is a small syntax update in a few javascript file, no new features included.

The first commit introduces a new `.gitignore` item for the `.idea` folder created when using PHPStorm (and similarly WebStorm).

The rest of the pr contains only syntax changes in some js files, i toyed with the idea of breaking down the changes for 1 liner returns and destruction separately, but in most cases the changes are on the same lines so didn't seem correct to pollute the history with the 2 changes which both fall under formatting changes.

No new features, all existing js test continue to pass after the changes.

The change updates some syntax to make use of the new js formatting features, specifically object destruction, implied returns in arrow functions and a small tweak to the `getCSRFToken` method in the `http` driver.

The later converts the method signature to be declarative instead of imperative, reducing control flow.

The former simply reduce repetition. 


src/js/index.js
```diff
@@ -32,10 +32,10 @@ class Livewire {
     }
 
     activatePolyfills() {
-        ArrayFlat();
-        ArrayFrom();
-        ArrayIncludes();
-        ElementGetAttributeNames();
+        ArrayFlat()
```
leemason: Ive removed the semi colons here, it seemed inconsistent with the rest of the files, i couldn't figure out what the intention here was?

Maybe `eslint` could be put in place for the package (happy to look into this as a future pr if i have time).

dist/mix-manifest.json
```diff
@@ -1,3 +1,3 @@
 {
-    "/livewire.js": "/livewire.js?id=b4b27f4e53118a0cdaf3"
+    "/livewire.js": "/livewire.js?id=8317e62ea8d2c2ca1a87"
```
leemason: I couldnt find any docs for how/when the `dist` folder should be updated, ive assumed if changing js files you need to run `npm run prod` and commit the `dist` files as ive done here.

tillkruss:
Is this an optimization we could do as well? https://github.com/pusher/pusher-js/pull/370

stof:
The pusher-js optimization you linked does not apply here as you don't use typescript (or rather it is already applied, as you already feed ES modules to webpack).
The optimization that you could benefit from though is https://github.com/pusher/pusher-js/pull/361: making sure that the only difference between your output files are the minification, not prod vs dev mode of webpack, so that other webpack prod optimization (and especially the module concatenation, which helps a lot the minification) are applied in your output file (minification gains will be there if someone bundles an unminified file into their own minified bundle, but modules cannot be concatenated later).
Of course, this is not relevant at all if your package.json contains the `module` key pointing to your ESM sources, as then webpack will load the ES modules directly when consuming your package (and so will provide its own module concatenation and tree shaking).

tillkruss:
Thanks for the details @stof!

leemason:
@calebporzio not really sure where to go with this now. Comments all sound great but i think there probably better targeted at a separate mr (really thanks to all involved, ive learned some stuff via the links provided).

The js files are changing fast (as to be expected) so im hoping you could give me some insight into if this is a worthwhile addition. If so i can then get everything update to date for review.

That does again bring me back to the dist folder. I really think this should built via a CI job which maybe could improve the conflict flags happening in this folder.

Id be more than happy to investigate this in some free time should it be required.

dist/mix-manifest.json
calebporzio: yes, that's right for now

src/js/Component/PrefetchManager.js
```diff
@@ -15,16 +15,16 @@ class PrefetchManager {
         if (message) message.storeResponse(payload)
     }
 
-    actionHasPrefetch(action) {
-        return Object.keys(this.prefetchMessagesByActionId).includes(action.toId())
+    actionHasPrefetch({ toId }) {
+        return Object.keys(this.prefetchMessagesByActionId).includes(toId())
```
calebporzio: I honestly prefer the old style here.

src/js/Message.js
```diff
@@ -7,20 +7,19 @@ export default class {
 
     get refs() {
         return this.actionQueue
-            .map(action => {
-                return action.ref
-            })
+            .map(({ ref }) => ref)
```
calebporzio: I think I'd rather see `action => action.ref`

I like the single return thing, but not sure about the param destructuring sometimes

src/js/Message.js
```diff
@@ -7,20 +7,19 @@ export default class {
...
-            name: this.component.name,
-            children: this.component.children,
-            middleware: this.component.middleware,
-            checksum: this.component.checksum,
+            id,
+            data,
+            name,
+            children,
+            middleware,
+            checksum,
```
calebporzio: I dig, good call.

src/js/Message.js
```diff
@@ -31,16 +30,16 @@ export default class {
         }
     }
 
-    storeResponse(payload) {
+    storeResponse({ id, dom, children, dirtyInputs, eventQueue, events, data, redirectTo }) {
```
calebporzio: This is an example of a time I like the destructuring

src/js/Store.js
```diff
@@ -52,9 +50,7 @@ const store = {
     },
 
     componentsListeningForEvent(event) {
-        return this.components().filter(component => {
-            return component.events.includes(event)
-        })
+        return this.components().filter(({ events }) => events.includes(event))
```
calebporzio: yeah, I think for me, I like seeing "component" as the param (even if it's obvious because it's chained off a filter), it just reads better to me. But like I said, I like the single return.

src/js/connection/drivers/http.js
```diff
@@ -55,17 +55,15 @@ export default {
...
-        if (!tokenTag) {
-            if (!window.livewire_token) {
-                throw new Error('Whoops, looks like you haven\'t added a "csrf-token" meta tag')
-            }
+        const { content: token } = {
+            ...{
+                content: window.livewire_token || null
+            },
+            ...document.head.querySelector('meta[name="csrf-token"]'),
+        }
```
calebporzio: Going to be honest, I have a really hard time following that refactor. Even though the original is a bit ugly, I think I still prefer it for readability.

src/js/connection/index.js
```diff
@@ -25,20 +25,21 @@ export default class Connection {
...
-            componentStore.findComponent(payload.id).receiveMessage(payload)
+            componentStore.findComponent(id).receiveMessage(payload)
 
             dispatch('livewire:update')
         }
     }
 
-    onError(payloadThatFailedSending) {
-        componentStore.findComponent(payloadThatFailedSending.id).messageSendFailed()
+    onError({ id }) {
```
calebporzio: I don't want to lose the well-named param here - it provides helpful context

src/js/index.js
```diff
@@ -14,7 +14,7 @@ class Livewire {
             driver: 'http'
         }
 
-        options = Object.assign({}, defaults, options);
+        options = { ...defaults, ...options }
```
calebporzio: Good call

src/js/index.js
calebporzio: Yeah, possibly, but you're right - the semicolons should be gone. I think I actually removed them since this PR.

calebporzio (changes requested):
@leemason, thanks for the PR. I like housekeeping style PRs like this. Take a look at my notes. I like some of your improvements, but not others. If you're willing to make the modifications, I'll merge it. Thanks for pitching in!

faustbrian:
@leemason would you have the time to resolve the conflicts and finish this PR?

calebporzio:
I think this has drifted too far from the code base, and the contributer seems to not be responsive. So, I'm going to close this for now.

Thanks for being willing to contribute in the first place, and thanks for bringing this back up @faustbrian !

leemason:
yes sorry guys, got quite a few things to get done before the holidays, and @calebporzio is right, if i did have time i would probably start again with the number of changes.

leemason:
@calebporzio just some feedback incase anyone does start on a similar mission. It appears the destruction changes received a 50/50 response sometimes liking, others not.

If i'm honest i think it would be really difficult for anyone to get it quite right without some guidelines on a coding style, or some basic linting rules in place.

I would also argue for a consistent codebase it should be all in or out, just some clarity for future contributors.

Keep up the great work, ill try and contribute in the future!

calebporzio:
Great points @leemason!

PR #222, opened 2019-09-01
Minify asset code

tillkruss:
- switch to single quotes (js code)
- made styles/script more readable
- don’t print “Livewire Assets” comment when `app.debug` is disabled
- minify script when `app.debug` is disabled


tillkruss:
We could extract the heredocs into separate methods to make it shorter and easier to read.

calebporzio:
Thanks for the PR @tillkruss.

I refactored the Heredocs into their own methods.

Also, I added a "defer" attribute to the external Livewire script. I had to change the `DOMContentLoaded` listener, to listen for a new event I'm dispatching now: `livewire:available`. This should have been done before anyways.

Thanks again

tillkruss:
@calebporzio: Beautiful 😄

I'd _maybe_ define the listener before including the script, just in case some browsers are weird with `defer`.

calebporzio:
I "caniused" `defer`, and it is supported by most everything, even IEs.

But you're right, it makes sense and wouldn't hurt. will do, thanks.

lancepioch:
@all-contributors please add @tillkruss for maintenance

PR #229, opened 2019-09-02
#192 Config File with paths for #192

### LINKED ISSUE #192: Not Working if namespace is not App
**Describe the bug**
Unable to find component

**To Reproduce**
My laravel App is named "Portal" instead of "App" - the component discovering won't work

**Expected behavior**
Show Livewire Component

**Additional context**
namespace App is hardcoded in LivewireMakeCommandParser.php@classNamespace and Component.php@getName

Solution: change it to config('app.name') instead of 'App'

defenestrator:
Fixes #192 by adding config file for paths for Livewire views, javascript assets and class namespaces.

Currently, Livewire hardcodes the applications root namespace, and view path. This change allows for configuration of those values to be pulled from .env, to be set statically in the config file, or retain the existing defaults. Includes a third configuration value to set the js asset "base_url"


defenestrator:
Yikes, let's try that again

PR #230, opened 2019-09-02
Configurable paths and Namespace, fixes #192

### LINKED ISSUE #192: Not Working if namespace is not App

defenestrator:
Fixes #192 by adding config file for paths for Livewire views, javascript assets and class namespaces.

Currently, Livewire hardcodes the applications root namespace, and view path. This change allows for configuration of those values to be pulled from .env, to be set statically in the config file, or retain the existing defaults. Includes a third configuration value to set the js asset "base_url"


defenestrator:
I can squash these and rebase to a single commit and new PR, if this is too messy for the git history, haste makes waste.

calebporzio:
This is a very nice PR - I'm not that concerned about a cleaner history, but suit yourself!

defenestrator:
Hey, I want to re-do this today, I killed that fork and made a new. I was distracted, rushed, and on new meds, haha!

I'll resubmit.
<3 Livewire!

PR #239, opened 2019-09-04
Remove keep alive

### LINKED ISSUE #171: Session never expires
The user session never expires.

I'm guessing this is to avoid errors if the component has to refresh and the session expired. But what if we need the session to expire after the time set in config/session.php?

We could extend the StartSession middleware so that whenever the header X-Livewire-Keep-Alive is present in the request, we optionally don't extend the cookie lifetime, we just leave the expiration date as is.

calebporzio:
Remove the "keep-alive" ajax request that keeps the session open longer than it should be.

Closes #171


lancepioch:
already added to contributors :D

PR #240, opened 2019-09-04
Better cache headers

tillkruss:
- Changed `Cache-Control` from 1 hour to 1 year
- Added `Last-Modified` / `Expires` header for browsers to use
- Return `304` response when cache matches


tillkruss:
I removed the `ETag` logic, because it's too much complexity.

calebporzio:
Solid. Thanks @tillkruss!

lancepioch:
@all-contributors please add @tillkruss for code

PR #241, opened 2019-09-04
Support publishing assets

### LINKED ISSUE #223: Publish assets as files instead of dynamic routes
Serving the `livewire.js` file through PHP makes it harder to cache.

Can we have a `vendor:publish` command like other packages? I'm happy to submit a PR.

tillkruss:
This PR adds support for publishing static assets. It also removed the unused js sourcemap. Fixes #223.

To automate the asset publishing add `@php artisan vendor:publish --force --tag=livewire-assets` to the `post-autoload-dump` script in your `composer.json`.


tillkruss:
Part of the docs should be ignoring js requests in telescope if assets aren’t published.

calebporzio:
Good PR @tillkruss, what's the strategy for making sure people re-publish when we update.

Is there something we can do automatically? Or at least hook in and remind them with a prompt?

Livewire has knowledge of what the JS file version should be, so it could, in theory, check the users public/vendor directory for the existence of livewire.js, if it finds it, check if it is out of date, then prompt the user to update it.

tillkruss:
I was just thinking because it’s optional to move it to the bottom of the install docs and then say “if you choose to publish the assets, make sure you register the post-auto load-dump hook, otherwise you assets will be out of date”.

tillkruss:
I guess the assets directive could spit out a version warning if they don’t match, eh? Or is there a better place?

tillkruss:
@calebporzio: I think this is read to be merged. Can I PR to the docs somewhere?

calebporzio:
@tillkruss - you can PR to the docs here: https://github.com/calebporzio/livewire-docs

Maybe under "Installation"?

tillkruss:
@calebporzio: Done: https://github.com/calebporzio/livewire-docs/pull/41

calebporzio:
Thanks @tillkruss!

lancepioch:
Already added as a contributor, 🚀

PR #251, opened 2019-09-08
Route name

### LINKED ISSUE #248: Make debugging easier
Right now all polling requests to Livewire look the same. Could we maybe append the component name the the route?

If it this isn’t used and redundant it would help identifying requests much much faster.

tillkruss:
This PR adds the component name to the route for easier debugging.

Closes #248.

@calebporzio: I haven't tested this code, because of the asset compilation. Am I missing something?


calebporzio:
Good stuff, thanks @tillkruss

calebporzio:
Also, I built and tested locally, looks good.

lancepioch:
already added as a contributor, nice!

PR #266, opened 2019-09-13
Improve performance by removing getClass()

### LINKED ISSUE #256: Delete is Extremely Slow
I have built what is basically a simple ToDo except it is adding and removing emails. When I add an email, it takes about 100ms which is fast enough for me. However, the delete is taking close to 600ms. It's shocking that it takes this long considering all that function is doing is executing one line of code: `Model::find($id)->delete();`

I have other Livewire implementations where the same type of functionality is not nearly as slow. I'm just wondering where to start with debugging this. Is it possible there's some kind of cache that needs to be cleared or something? I'm running version 0.3.2 if that matters.

slowdelete

My component and blade:

```
<div class="form-row">
    <h4 style="width: 100%;">Email CC for Participants</h4>
    <h6 style="width: 100%">If a participant is not connected to any housing locator or case manager, cc the following emails.</h6>
    <!-- Email Form Input -->
    <div class="form-group label-floating bmd-form-group col-md-6" wire:ignore>
        <label for="email" class="bmd-label-floating">Add Email (Press Enter)</label>
        <input type="text" name="email" class="form-control" wire:model="email" wire:keydown.enter="addEmail"/>
        @if($errors->has('email'))<div class="input-error">{{$errors->first('email')}}</div>@endif
    </div>
    <div class="col-md-6" style="display: flex; justify-content: center; align-items: center; flex-wrap: wrap;">
        @foreach($emails as $email)
            <div class="badge badge-info" style="font-size: 14px; margin: 5px 5px 0;">{{$email->email}}<span><i class="material-icons" style="font-size: inherit; cursor: pointer; margin-left: 10px;" wire:click="deleteEmail('{{$email->id}}')" wire:ref="delete-button-{{$email->id}}">close</i></span></div>
            <div wire:loading wire:target="delete-button-{{$email->id}}" class="text-danger">
                Removing Email...
            </div>
        @endforeach
    </div>
</div>
```

```
class TenantCc extends Component
{
    public $email;

    public function render()
    {
        return view('livewire.tenant-cc', [
            'emails' => request()->user()->tenant->tenant_notification_emails,
        ]);
    }

    public function addEmail()
    {
        $this->validate([
           'email' => 'required|email',
        ]);

        request()->user()->tenant->tenant_notification_emails()->create([
            'email' => $this->email,
        ]);
    }

    public function deleteEmail($id)
    {
        TenantNotificationEmail::find($id)->delete();
    }
}
```

intellow:
Fixes #256 

Admittedly I don't really know what's going on here, but I do know that with `->getClass()` here, my delete function takes between 400ms and 800ms. When I remove it then it not only still works, but takes just about 50ms.

### DIFF
--- a/src/Concerns/HandlesActions.php
+++ b/src/Concerns/HandlesActions.php
@@ -73,6 +73,7 @@ trait HandlesActions
                 if ($class = $parameter->getClass()) {
                     return app($class->name);
                 }
+
                 return app($parameter->name);
             }, function () use (&$params) {
                 return array_shift($params);
--- a/src/Concerns/HandlesActions.php
+++ b/src/Concerns/HandlesActions.php
@@ -70,7 +70,11 @@ trait HandlesActions
     {
         return collect((new \ReflectionMethod($this, $method))->getParameters())->map(function ($parameter) use (&$params) {
             return rescue(function () use ($parameter) {
-                return app($parameter->name);
+                if($parameter->getClass()) {  
+                    return app($parameter->getClass()->name);
+                } else {
+                    return app($parameter->name);
+                }
             }, function () use (&$params) {
                 return array_shift($params);
             }, false);
--- a/src/Concerns/HandlesActions.php
+++ b/src/Concerns/HandlesActions.php
@@ -70,7 +70,7 @@ trait HandlesActions
     {
         return collect((new \ReflectionMethod($this, $method))->getParameters())->map(function ($parameter) use (&$params) {
             return rescue(function () use ($parameter) {
-                if($parameter->getClass()) {  
+                if ($parameter->getClass()) {
                     return app($parameter->getClass()->name);
                 } else {
                     return app($parameter->name);
--- a/src/Concerns/HandlesActions.php
+++ b/src/Concerns/HandlesActions.php
@@ -70,7 +70,7 @@ trait HandlesActions
     {
         return collect((new \ReflectionMethod($this, $method))->getParameters())->map(function ($parameter) use (&$params) {
             return rescue(function () use ($parameter) {
-                return app($parameter->getClass()->name);
+                return app($parameter->name);
             }, function () use (&$params) {
                 return array_shift($params);
             }, false);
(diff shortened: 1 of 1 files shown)


intellow:
A few other notes on this one. As an avid `dd()` developer, I narrowed down the slowness to this method. For my implementation, my `$parameter` was the following:

image

When I did `dd($parameter->getClass())` I got null, and then this line was trying to call `->name` on null, and this is where my performance issues came from. However, there is a `name` parameter on my variable, so it seems like we need to check for getClass() and if it's null then just do `$parameter->name`.

Like I said above, I'm not really sure what's going on in this code but it fixes my problem and passes CI. 

@jrseliga it looks like you committed this code originally. Does this look good or do you have any other input?

tillkruss:
Can you pull the output into a variable so it only called once?

src/Concerns/HandlesActions.php
```diff
@@ -70,7 +70,11 @@ protected function resolveActionParameters($method, $params)
     {
         return collect((new \ReflectionMethod($this, $method))->getParameters())->map(function ($parameter) use (&$params) {
             return rescue(function () use ($parameter) {
-                return app($parameter->getClass()->name);
+                if ($parameter->getClass()) {
+                    return app($parameter->getClass()->name);
+                } else {
+                    return app($parameter->name);
+                }
```
tillkruss: How about this?

```php
if ($class = $parameter->getClass()) {
    return app($class->name);
}

return app($parameter->name);
```

tillkruss (changes requested):

src/Concerns/HandlesActions.php
intellow: Looks good to me.

calebporzio:
Weird, I wonder if it's because the `rescue` helper calls `report($e)`, and maybe you have some error handling that takes a long time? Not sure.

I found a similar performance issue with the `report` method in the exception handler myself, but I can't replicate it to dig deeper.

Either way, I think this is a good code improvement anyways, although we probably don't need

lancepioch:
@all-contributors please add @intellow for maintenance

PR #287, opened 2019-09-22
Windows only fix for finding components

lancepioch:

Bug Fix for PHP on Windows

No

Components can't be automatically found at all before this change

Component File Paths:
Path before change: `C:\neard\www\library\app\`
*even though this looks correct ^, win/php actually ends it with a forward slash, so this causes Str::after to fail*
Path after change: `C:\neard\www\library\app/`

The problem is that PHP/Windows doesn't actually use the directory separator at the end.


calebporzio:
Thanks!

lancepioch:
This guy looks familiar 👀 - already added as a contributor

PR #300, opened 2019-09-25
Make base namespace removal less aggressive when getting the name of a Component

### LINKED ISSUE #267: Component or View not found if 'App' part of path other than the beginning
### Issue
If you make a component with `App` in the path, it causes the Component and View to not be found due to `Component->getName()` removing `App` too aggressively.

### Run
`php artisan make:livewire app.foo.livewire-is-awesome`

#### Creates the following in `livewire-components.php`
```php
<?php return array (
  'foo.livewire-is-awesome' => 'App\\Http\\Livewire\\App\\Foo\\LivewireIsAwesome',
);
```
*when it should create...*
```php
<?php return array (
  'app.foo.livewire-is-awesome' => 'App\\Http\\Livewire\\App\\Foo\\LivewireIsAwesome',
);
```

Also, if you remove the override `render()` function from the `LivewireIsAwesome` component so it uses the the base `render()` function in `Component`, Livewire tries to load the view `livewire.foo.livewire-is-awesome` instead of `livewire.app.foo.livewire-is-awesome`.

```php
<?php

namespace App\Http\Livewire\App\Foo;

use Livewire\Component;

class LivewireIsAwesome extends Component
{
    // public function render()
    // {
    //     return view('livewire.app.foo.livewire-is-awesome');
    // }
}
```

### Cause
The cause of this is the `getName()` method in `Component` being too aggressive. It is removing all instances of `App` in the path instead of just the first one.

```php
public function getName()
{
    return collect(explode('.', str_replace(['/', '\\'], '.', static::class)))
        ->diff(['App', 'Http', 'Livewire'])
        ->map([Str::class, 'kebab'])
        ->implode('.');
}
```

#### Option 1
```php
public function getName()
{
    return collect(explode('.', str_replace(['/', '\\'], '.', static::class)))
        ->slice(3)
        ->map([Str::class, 'kebab'])
        ->implode('.');
}
```
#### Option 2
```php
public function getName()
{
    $name = collect(explode('.', str_replace(['/', '\\'], '.', static::class)))
        ->map([Str::class, 'kebab'])
        ->implode('.');

    if (Str::startsWith($name, 'app.http.livewire.')) {
        return Str::replaceFirst('app.http.livewire.', '', $name);
    }

    return $name;
}
```

#### Option 3
And to accommodate https://github.com/calebporzio/livewire/pull/235 something like:

```php
public function getName()
{
    $name = collect(explode('.', str_replace(['/', '\\'], '.', static::class)))
        ->map([Str::class, 'kebab'])
        ->implode('.');

    $namespace = Str::kebab(config('livewire.namespace'));

    if (Str::startsWith($name, $namespace . '.http.livewire.')) {
        return Str::replaceFirst($namespace . '.http.livewire.', '', $name);
    }

    return $name;
}
```

wsamoht:
Yes, resolves #267 

Nope.

Most definitely.

This changes how the base namespace of Livewire Components is removed when getting the name for a Component. Before it was too aggressive removing all instances of a word from the entire path. 

This is needed to accommodate for example if `App` is in the namespace twice like `App\Http\Livewire\App\Test`. Before it would remove both instances of `App`. Now it just removes the first giving us `app.test` instead of just `test`.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -43,12 +43,15 @@ abstract class Component
 
     public function getName()
     {
-        $namespace = config('livewire.class_namespace', 'App\\Http\\Livewire');
-
-        return collect(explode('.', str_replace(['/', '\\'], '.', static::class)))
-            ->diff(explode('\\', $namespace))
+        $namespace = collect(explode('.', str_replace(['/', '\\'], '.', config('livewire.class_namespace', 'App\\Http\\Livewire'))))
             ->map([Str::class, 'kebab'])
             ->implode('.');
+
+        $name = collect(explode('.', str_replace(['/', '\\'], '.', static::class)))
+            ->map([Str::class, 'kebab'])
+            ->implode('.');
+
+        return Str::substr($name, strlen($namespace) + 1);
     }
 
     public function render()
--- a/src/Component.php
+++ b/src/Component.php
@@ -51,7 +51,11 @@ abstract class Component
             ->map([Str::class, 'kebab'])
             ->implode('.');
 
-        return Str::substr($name, strlen($namespace) + 1);
+        if (Str::startsWith($name, $namespace)) {
+            return Str::substr($name, strlen($namespace) + 1);
+        }
+
+        return $name;
     }
 
     public function render()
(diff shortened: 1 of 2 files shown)


calebporzio:
Great work! Thanks a bunch!

lancepioch:
@all-contributors please add @wsamoht for code

PR #317, opened 2019-10-04
Add support for Laravel version 5.7

kotfire:
Related to #315 

The only obstacle for using Liveware in a laravel 5.7 project was the way the package publishes it resources. Using multiple groups as an array was introduced in 5.8 (see https://github.com/laravel/framework/pull/26115)

For convenience I have created a method in the `LivewireServiceProvider` that provides the same behaviour.

Issue #315, opened 2019-10-04
Hacktoberfest: Support Laravel 5.7 (or maybe even earlier)


calebporzio:
Hey Hacktoberfesters! Here's is an isolated scope issue that shouldn't be too difficult to resolve.

Currently, Livewire doesn't work on any Laravel version below 5.8 (#296).

We could easily set the minimum version constraint to 5.8 (#314).

HOWEVER, I don't think there is any legitimate reason Livewire couldn't support earlier versions of Laravel. I believe the reason it currently doesn't is because of just a couple of PHP methods I use, that we could easily workaround.

So, if you're up for it. Maybe create a new (or if you have an existing) Laravel project with 5.7 (or maybe even lower), pull in Livewire, and find the exact spots it's broken. Outline what the hangups are, replace the methods with older-php-compatible ones, and submit a PR.

That would be GREAT!

Comment here for any questions or if you need any help with it, and I'll be sure to respond.

Thanks for Hacktoberfesting!

PR #318, opened 2019-10-04
Add support to Laravel version 5.6

kotfire:
Related to #315 

See #317 

This pull request goes one step further and downgrades the laravel version requirement to 5.6. I think we shouldn't encourage people to use version older than 5.6 because it is the oldest Laravel version with a PHP requirement including security fixes.

In addition to the changes in #317 in this case we need the `Filesystem::replace` method was introduced in 5.7 (See https://github.com/laravel/framework/pull/26010). I replaced it with the [5.6 code](https://github.com/laravel/framework/blob/d1be6e43207ed1bacd183f11b87c152dffd08a7c/src/Illuminate/Foundation/PackageManifest.php#L171)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,8 +9,8 @@
         }
     ],
     "require": {
-        "illuminate/support": "^5.7|^6.0",
-        "illuminate/validation": "^5.7|^6.0"
+        "illuminate/support": "^5.6|^6.0",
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks a bunch for doing both of these! Big help!

lancepioch:
already added as a contributor, thanks 🏆

Issue #321, opened 2019-10-07
Option to store protected properties in browser


slava-vishnyakov:
**Is your feature request related to a problem? Please describe.**
Currently the protected properties are stored in cache. If you have a reasonably popular site - your cache will explode with dozens of new cache items per each page visited by a robot or a human.

The bots can often be like 99% of visits to the site and they won't click any button. Therefore any state that I keep for them is useless. Even with users - most of them won't click anything, so again - most of the state I keep for them is useless waste of Redis memory.

**Describe the solution you'd like**
I'd like to be able to store everything client-side. Maybe something like

```
class MyComponent ... {
  use EncryptedClientStorage;
```

Basically, LiveWire can send something like `encrypt($protected)` to the browser and receive it back whenever a request goes back to LiveWire (just like it now sends `data`, `dirty`, it might send also `protected`, which will be decrypted server-side). 

Even though this is a trade-off (fatter network packets), but often having fatter packets is much more useful than having hundreds of gigabytes of RAM.

You can have hundreds of thousands of robots and humans on site and still not require any additional RAM. 

I don't expect majority (99.99% or something like that) of LiveWire widgets to receive any interactions in future, but still I have to keep data for each of them.

(P.S. That's actually how my current solution works, which is pretty similar to LiveWire. I'd like to use LiveWire on my current site, but my napkin-based calculations suggest that I would need hundreds of gigabytes of RAM to use LiveWire instead of my solution, which just sends all "protected" data encrypted to the user, for user to send it later back)

calebporzio:
Hey @slava-vishnyakov, yes, that makes sense.

I think most Livewire usage is in a non-public, app context. Therefore, I believe caching is a good default. Also, when Livewire is used on public facing stuff prone to bot usage, I imagine people using protected properties less, but I know I could be wrong AND there are exceptions.

Anyhow, I agree, it would be nice to offer the option to use the AJAX  request to transported encrypted protected properties.

I suppose the only question is what is the best way to configure that option.

If it is per component, we would need a trait or something like you proposed. It could also be site-wide and configured in the config file.

Anyhow, I like it. Would you be willing to PR this feature?

slava-vishnyakov:
Maybe we can have a config option, something like 
```
'protected_storage' => 'cache' / 'browser'
``` 

and also a base method 

```
protected function getProtectedStorage() { 
  return config('livewire.storage'); 
}
``` 

That way you could override it in your component like this:

```
protected function getProtectedStorage() { 
[...]

PR #323, opened 2019-10-08
Add protected storage in browser as discussed in #321

slava-vishnyakov:
Yes, #321 

No

Yes, but maybe needs more

See #321

👍

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -55,4 +55,17 @@ return [
 
     'middleware_group'  => 'web',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Livewire Protected Storage
+    |--------------------------------------------------------------------------
+    |
+    | This value defines where to store the protected properties of the
+    | component. Valid values are: 'cache' (store in server-side cache)
+    | and 'browser' (store encrypted in AJAX requests).
+    |
+    */
+
+    'protected_storage'  => 'cache',
+
 ];
--- a/src/Component.php
+++ b/src/Component.php
@@ -8,6 +8,8 @@ use Illuminate\Support\Str;
 use Illuminate\Support\MessageBag;
 use Illuminate\Support\ViewErrorBag;
 use Livewire\Exceptions\PublicPropertyTypeNotAllowedException;
+use Livewire\ProtectedStorage\ProtectedStorageInBrowser;
+use Livewire\ProtectedStorage\ProtectedStorageInCache;
 
 abstract class Component
 {
@@ -167,4 +169,24 @@ abstract class Component
             'Method %s::%s does not exist.', static::class, $method
         ));
     }
+
+    public function getProtectedStorage()
+    {
+        return config('livewire.protected_storage', 'cache');
+    }
+
+    public function getProtectedStorageEngine()
+    {
+        if ($this->getProtectedStorage() === 'cache') {
+            return new ProtectedStorageInCache();
+        }
+
+        if ($this->getProtectedStorage() === 'browser') {
+            return new ProtectedStorageInBrowser();
+        }
+
+        throw new \RuntimeException("Protected storage config option '" . $this->getProtectedStorage() . "' is invalid");
+
+    }
+
 }
--- a/src/Connection/ComponentHydrator.php
+++ b/src/Connection/ComponentHydrator.php
@@ -10,18 +10,12 @@ class ComponentHydrator
 {
     public static function dehydrate($instance)
     {
-        // Store the protected properties in the cache.
-        if ($protectedOrPrivateProperties = $instance->getProtectedOrPrivatePropertiesDefinedBySubClass()) {
-            (new ComponentCacheManager($instance))->put(
-                '__protected_properties',
-                $protectedOrPrivateProperties
-            );
-        }
+        $instance->getProtectedStorageEngine()->saveProtectedData($instance);
 
         return $instance->getPublicPropertiesDefinedBySubClass();
     }
 
-    public static function hydrate($component, $id, $publicProperties, $checksum)
+    public static function hydrate($component, $id, $publicProperties, $checksum, $protected = null)
     {
         // Make sure the data coming back to hydrate a component hasn't been tamered with.
         $checksumManager = new ComponentChecksumManager;
@@ -34,18 +28,12 @@ class ComponentHydrator
 
         $unHydratedInstance = new $class($id);
 
-        // Grab the protected properties out of the cache.
-        $protectedOrPrivateProperties = (new ComponentCacheManager($unHydratedInstance))
-            ->get('__protected_properties', []);
-
-        return tap($unHydratedInstance, function ($unHydratedInstance) use ($publicProperties, $protectedOrPrivateProperties) {
+        return tap($unHydratedInstance, function ($unHydratedInstance) use ($publicProperties, $protected) {
             foreach ($publicProperties as $property => $value) {
                 $unHydratedInstance->setPropertyValue($property, $value);
             }
 
-            foreach ($protectedOrPrivateProperties as $property => $value) {
-                $unHydratedInstance->setProtectedPropertyValue($property, $value);
-            }
+            $unHydratedInstance->getProtectedStorageEngine()->restoreProtectedData($unHydratedInstance, $protected);
         });
     }
 }
--- a/src/Connection/ConnectionHandler.php
+++ b/src/Connection/ConnectionHandler.php
@@ -12,7 +12,9 @@ abstract class ConnectionHandler
 {
     public function handle($payload)
     {
-        $instance = ComponentHydrator::hydrate($payload['name'], $payload['id'], $payload['data'], $payload['checksum']);
+        $instance = ComponentHydrator::hydrate(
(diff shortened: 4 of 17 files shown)


slava-vishnyakov:
@calebporzio 

I've added the config option and two methods to Component `getProtectedStorage()` which user can override on per-component basis, like this:
https://github.com/livewire/livewire/pull/323/files#diff-8a7a34e0fd91135184ee976810b4bfcbR34-R37  and `getProtectedStorageEngine()` which returns a specific implementation of storage.

I've extracted the protected storage behavior into two classes `ProtectedStorageInCache` (current behavior) and `ProtectedStorageInBrowser`. 

The only thing that I'm having problems with is how to test it end-to-end that it actually works. Like, for example this change: https://github.com/livewire/livewire/pull/323/files#diff-f22e3dd7206922d2012072b60be30644R21-R46 - it's not covered by any test. But I have no idea how to test it. Any help would be appreciated.

slava-vishnyakov:
Ok, I'm a bit stuck with this. I've run `npm install` and then `npm run unminified`, now I get this:

```
 ERROR  Failed to compile with 4 errors                                                                                                                                                                                                                        1:22:40 PM

These dependencies were not found:

* @/component/DirtyStates in ./src/js/index.js
* @/component/LoadingStates in ./src/js/index.js
* @/component/OfflineStates in ./src/js/index.js
* @/component/Polling in ./src/js/index.js

To install them, you can run: npm install --save @/component/DirtyStates @/component/LoadingStates @/component/OfflineStates @/component/Polling
       Asset     Size     Chunks             Chunk Names
/livewire.js  184 KiB  /livewire  [emitted]  /livewire

ERROR in ./src/js/index.js
Module not found: Error: Can't resolve '@/component/DirtyStates' in '.../livewire/src/js'
 @ ./src/js/index.js 19:0-50 151:0-11
 @ multi ./src/js/index.js

ERROR in ./src/js/index.js
Module not found: Error: Can't resolve '@/component/LoadingStates' in '.../livewire/src/js'
 @ ./src/js/index.js 18:0-54 150:0-13
 @ multi ./src/js/index.js

ERROR in ./src/js/index.js
Module not found: Error: Can't resolve '@/component/OfflineStates' in '.../livewire/src/js'
 @ ./src/js/index.js 20:0-54 152:0-13
 @ multi ./src/js/index.js

ERROR in ./src/js/index.js
Module not found: Error: Can't resolve '@/component/Polling' in '.../livewire/src/js'
 @ ./src/js/index.js 21:0-42 153:0-7
 @ multi ./src/js/index.js
```

slava-vishnyakov:
Ok, now (after https://github.com/livewire/livewire/pull/323/commits/ec5488d5d385624420e55e165e26173d3c21e5b7) it definitely works. Just need to figure out a few things:

1) Is there a way to add end-to-end tests? 
2) How to test JS part of all of this
3) Do I need to be concerned about `.children` (in JS/PHP) and their `.protected` fields somehow?

slava-vishnyakov:
@calebporzio Can you help a little with these questions?

calebporzio:
Ok, sorry I haven't gotten to this, I've been traveling.

Thanks a bunch for this PR, it's really well done.

I read through it and pulled it down locally, and I just wasn't happy with how complicated this feature needed to be because of the current architecture.

I refactored the Livewire codebase to support something like this better.

Now all your changes (except for your tests) can be isolated to one file: `src/HydrationMiddleware/HydrateProtectedProperties.php`

You should be able to add the switching from the config in there.

I know I just made more work for you, so let me know if you want me to work on this PR myself and I'm happy to, or you can continue finishing it.

Let me know.

About your questions:
1. Not at the moment (will definitely have a Dusk suite before 1.0)
2. Hmm, I'm not sure JS tests are necessary here. JavaScript is just passing through the extra encrypted property... yeah, E2E tests would be best for this. If there's a simple test to make sure that if "protected_properties" is present in the request, that it makes it's way back, maybe that'd be good, but not necessary
3. I don't think you have to be concerned with children.
4. Yes please
5. Hmm, I think you're right, it's already encrypted, let's leave it out of the checksum

Let me know, thanks again @slava-vishnyakov

slava-vishnyakov:
@calebporzio Cool! Thanks for the answers! 

I'll check out the new branch and see if I can do a PR.

slava-vishnyakov:
Closing this for now, will open a new PR.

slava-vishnyakov:
@calebporzio 
Just to be clear - did you mean something like this?

image

calebporzio:
Yeah, something along those lines, yes.

slava-vishnyakov:
Should we keep the ability to override the storage per component? I.e. I want for ex. some components to be stored in cache, but others in browser, like this?

image

calebporzio:
Yeah, I do think that'd be helpful

slava-vishnyakov:
Trying to wrap my head around the idea of "Now all your changes (except for your tests) can be isolated to one file"

1) When the component is first rendered into HTML - neither `hydrate`, nor `dehydrate` method is called, which seems like a bug, actually.

I've tried this and the component happily rendered.

image

Which leads me to this discovery. Here's a simple counter that doesn't start with 0, but rather takes its value from some secret secure storage:

image

Let's see what happens:

ScreenFlow

The initial value is always discarded and the counter doesn't work as expected. Seems like a bug to me, or is it intended?

2) Again, returning to the idea that "all your changes (except for your tests) can be isolated to one file". Here's the `InitialResponsePayload` being created:

image

I'm not sure where the `protected` properties would end up here if I can't touch this file. The only place is `data`, but it is initialized by the function `$properties = $instance->getPublicPropertiesDefinedBySubClass();` 

Here we have 2 problems: it returns only `public` properties and it doesn't call hydrator.. So I have no influence on this function (or any other in `new InitialResponsePayload` call).

So, now, do we hide the protected properties in the `data` somehow?

I'm kind of stuck here. Any ideas?

calebporzio:
Sorry, @slava-vishnyakov - you're right.

I refactored further. Check out the new tagged release for the code updates.

Basically, you would manipulate the `$response` object to add a "protected_properties" property. This property will now be automatically added the component's initial data payload AND  subsequent ones.

Hope this makes sense, if it doesn't I'm happy to hop  on a Zoom call or something.

slava-vishnyakov:
After updating and trying this:

```
@livewire('counter', 42)
```

I'm getting this:

```
Object of class Illuminate\Support\Fluent could not be converted to string (View: ../livewire/src/Macros/livewire-view.blade.php)
```

image

image

image

Either I did something obviously stupid or maybe we should add CircleCI + Dusk?..

slava-vishnyakov:
Ok, scratch that. It was a cached view.

slava-vishnyakov:
Ok, so I think the idea is like this:

```
    public static function hydrate($unHydratedInstance, $request)
    {
        // I should have $request->protected_properties here
    }

    public static function dehydrate($instance, $response)
    {
        $response->protected_properties = ['a' => 1];
    }

```

However it does not work:

```
    public static function hydrate($unHydratedInstance, $request)
    {
        # $request->protected_properties;
        # Trying to get property 'protected_properties' of non-object

        # $request['protected_properties'];
        # ErrorException: Undefined index: protected_properties
    }

    public static function dehydrate($instance, $response)
    {
        $response->protected_properties = ['a' => 1];
    }
```

I see this:

```
wire:initial-data="...],&quot;protected_properties&quot;:{&quot;a&quot;:1},&quot;children
```

Which is a good sign.

However the JS part still seems to have only a few hardcoded properties:

image

So my property never arrives back to the backend.

slava-vishnyakov:
The screenshot is `src/js/Message.js`

calebporzio:
@slava-vishnyakov, you're correct. We have to either update the JS to allow for "passing-through" data that isn't explicitly declared.

What do you think the best thing to do is here?

I'm tempted to just add the "protected_properties" as an optional param that will be added to the request if it exists from the previous response (same with the initial extraction of the data in component/index.js) - thoughts? Thanks for pushing on this!

slava-vishnyakov:
Well, off the top of my head I can't think of anything else that we might
need to pass and even if that happens - it's probably easier to just
hardcode one more property.

Another solution cold be to have a 'custom' prop and just json encode it
for props like '$response->custom->protected'.

Maybe the 'custom' way is more flexible, though maybe it opens some new
attack vectors... Can't think of any though

slava-vishnyakov:
Two things that bug me about `custom` approach. 

1. If we were decoding to a simple stdClass or array - we'd probably be much safer than if we use something like Fluent, which has a few methods of its own, maybe it's somehow possible to craft something like `{"custom": {"../data": "..."}}` to override the `.data`.. I'm not sure, doesn't seem likely, but still it's a class with a bunch of methods.

2. Having a `custom` I'd probably feel safer if it were checksummed too. But then again, what prevents somebody in future from setting `$response->custom->my_nice->nested_thing = [true, 0, (stdClass)['a' => null]]`, so now the checksumming becomes a small nightmare. :)

calebporzio:
True, I say we just hardcode an extra param on the front-end for now.

Yeah, I was thinking the same thing with attack vectors, but too couldn't think of any.

Thanks.

Issue #316, opened 2019-10-04
JS error with wire:model in select


azurinspire:
**Describe the bug**
I'm getting following error when using wire:model with select:
`TypeError: Element.setAttribute requires at least 2 arguments, but only 1 were passed`

Error comes from `specialElHandlers.js` file in where the second attribute of `setAttribute` was dropped in this commit: https://github.com/livewire/livewire/commit/38caadbfdb128252a1251354a4c0e15d0935bf55#diff-3e6ee14443ed0559b0e9e7fb4dfc26b2L7 

I think it shouldn't be dropped:
https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute

**Expected behavior**
No errors

**Desktop (please complete the following information):**
 - OSX  Firefox 69

azurinspire:
Actually the error comes if translation is used as option text and for some reason locales are not in sync on client and server. 

![image](https://user-images.githubusercontent.com/40248184/66219720-f5e9ca80-e6cb-11e9-8129-8695a0254450.png)

![image](https://user-images.githubusercontent.com/40248184/66219735-fbdfab80-e6cb-11e9-86e6-b99461115c74.png)

![image](https://user-images.githubusercontent.com/40248184/66219749-026e2300-e6cc-11e9-897f-99769edcdf5a.png)

azurinspire:
But actually this error happens always if option labels are dynamic for some reason. Don't know if there are really use cases when that would be needed...

Minimal example to reproduce:

```html
<div>
   <select wire:model="test">
        <option value="0">{{ $dynamic }}</option>
        <option value="1">option 1</option>
        <option value="2">option 2</option>
        <option value="3">option 3</option>
   </select>
</div>
```

```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class Test extends Component
{
    public $test = 0;
    public $dynamic = 0;

    public function render()
    {
        $this->dynamic++;
        return view('livewire.test');
    }
}
```

azurinspire:
This could be quite normal use case which is causing same error:
```html
<div>
    <select wire:model="test1">
        <option value="0">option 0</option>
        <option value="1">option 1</option>
        <option value="2">option 2</option>
        <option value="3">option 3</option>
    </select>

    <select wire:model="test2">
        @foreach ($options as $option)
            <option value="{{ $option['value'] }}">{{ $option['label'] }}</option>
        @endforeach
    </select>
</div>
```

```php
<?php

[...]

PR #326, opened 2019-10-09
add second parameter to setAttribute 

allmanaj:
This was raised in the issue #316 

no
no
The [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) for the method says that it needs 2 parameters and this removes the following error:
`TypeError: Element.setAttribute requires at least 2 arguments, but only 1 were passed`


calebporzio:
I believe it should be `setAttribute(name, '')` -> I think that's what it was before.

Thanks for the PR!

allmanaj:
I've amended that now!

calebporzio:
Thanks!

lancepioch:
@all-contributors please add @allmanaj for maintenance

Issue #175, opened 2019-08-19
Store View Templates


intellow:
Would you accept a PR that does the following:

`php artisan make:livewire counter --view=my_template`

And instead of scaffolding the standard view, it would look for an use a template of a view that stored in views/livewire/templates/my_template.blade.php as the template for the new component view.

This would allow people to store any customization they like for their component views.

calebporzio:
Ah, very interesting.

I love the concept. Will want to brainstorm on the specifics.

Something I've always wanted in Laravel is the ability to customize the stubs that `make:` commands use.

I would love the same for Livewire. I'm not sure of the best way to accomplish this though.

Here are some options that come to mind:

A) Provide some sort of configuration option (weather in a service provider or config file) for overriding the class and view stubs.
B) Provide some sort of config for different template options
```
return ['templates' => [
    ['admin' => ['class' => 'App\Livewire\Templates\AdminStub', 'view' => 'livewire.templates.admin-stub']
];
```
Then this would be used like: `artisan make:livewire --template=admin` or something
C) Do what you are proposing and provide `--view` and `--class` options to use files as a template. The questions I have about that approach are: 1) what about class templates? they will need to have the right class name and namespace 2) I'm not sure about the `livewire/templates/foo` naming convention (probably the best option, just not sure if relying on a convention here is best) 3) People like me have a hard time remembering command options and especially don't want to type all that extra text in the command (maybe there's a way to make it more memorable / intuitive)

Those are my thoughts, let's get the conversation going.

intellow:
I like option B a lot, but I know you were trying to avoid publishing a config file, which is why I proposed the options in the command.

I like having a first step where you run a command like `livewire:publish-stubs` to publish your current stubs for both class and view in `App\Livewire\Templates\StandardStub` and `livewire.templates.standard-stub`. If you haven't run this, `livewire:make` scaffolds from the livewire core. If you have run this, then `livewire:make` scaffolds out every new component based on these files instead of the files in the livewire core, and the user can customize the standard stubs. This allows some customization without needing a config file.

Step two is the config file where the user can add new templates and pass the template name through in the `make:livewire NewComponent --template=admin` command, just like you described.

edgrosvenor:
[...]

PR #355, opened 2019-11-08
Make Livewire Stubs

intellow:
Related to #175 

The stubs don't do anything right now. This is the first PR with more to come.


intellow:
I have been trying to write the test cases for this and have not been successful. I think I need to refactor to create a StubParser.php similar to the ComponentParser.php but have been struggling with that as well. @calebporzio any interest in a pairing session to help me get on the right track here?

intellow:
Okay, nevermind. I think I have the refactor working and some very basic tests working as well.

PR #357, opened 2019-11-10
Use stubs in livewire:make command

intellow:
Resolves #355 

### Use Custom Stub in Make Command
Run `php artisan livewire:stub modal`
Customize the views/livewire/stubs/modal.stub and Http/Livewire/Stubs/Modal.stub files
Run `php artisan make:livewire success-model --stub=modal`
See that the customized stubs are used to create this modal

### Customize Default Stubs in Make Command
Run `php artisan livewire:stub default`
Customize the views/livewire/stubs/default.stub and Http/Livewire/Stubs/Default.stub files
Run `php artisan make:livewire success-model`
See that the customized stubs are used to create this modal

### Regression Testing
Run `php artisan make:livewire success-modal`
See that no custom stubs are being used, and the default stubs from the vendor package are used.

I have added a basic test here as well.

### DIFF
--- a/src/Commands/ComponentParser.php
+++ b/src/Commands/ComponentParser.php
@@ -3,6 +3,7 @@
 namespace Livewire\Commands;
 
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\File;
 use Illuminate\Support\Str;
 use Illuminate\Console\DetectsApplicationNamespace;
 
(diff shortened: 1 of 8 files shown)


intellow:
Sorry, this is the second PR after #355. This resolves #175

calebporzio:
Thanks a bunch @intellow - sorry for taking so long to get to these. They're great!

PR #364, opened 2019-11-13
$this keyword in blade references the component class

shadoWalker89:
This PR allows to bind the `$this` php keyword inside of the livewire blade file to the related component class.

Before the PR your template will have somethings like this

```php
class MyComponent
{
    public $selected_ids;
}
```

```php
@if(in_array($id, $selected_ids))
    Do something
@endif
```

After the PR

```php
class MyComponent
{
    private $selected_ids;

    public function isSelected($id)
    {
        return in_array($id, $this->selected_ids);
    }
}
```

```php
@if($this->isSelected($id))
    Do something
@endif
```

This has 2 advantages :

1- Move the logic of checking if the item is selected from your template to the class. In the provided example, it looks simple and that doesn't hurt to have it within the template, but when you work on large components, things can get ugly very fast.

2- The `$selected_ids` is now private (was public before the PR), you don't need to pass every variable to the template because now the work is handled by the method within the class. This will lead to a more readable and clean code.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -92,6 +92,8 @@ abstract class Component
             $errorBag = $errors ?: ($view->errors ?: $this->getErrorBag())
         );
 
+        app('view.engine.resolver')->resolve('blade')->setLivewireComponent($this);
+
         $view
             ->with([
                 'errors' => (new ViewErrorBag)->put('default', $errorBag),
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -109,6 +109,7 @@ class LivewireServiceProvider extends ServiceProvider
         $this->registerRouterMacros();
         $this->registerBladeDirectives();
         $this->registerPublishables();
+        $this->registerViewCompilerEngine();
     }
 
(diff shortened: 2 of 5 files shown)


shadoWalker89:
About the CompilerEngine
I'm overriding it to allow for this feature (in boot()), and i see that you do as well (in register())
I think it's best if we move your override to the new class added by this PR

I can do that in another PR if this is merged.

roni-estein:
Sweet!

calebporzio:
I'm wondering if there's a way to only mess with Laravel's bindings on an "as-needed" basis.

For example, this solution hijacks the compiler for every request to a Laravel app that has Livewire installed (even pages with no Livewire components).

Any thoughts on moving to a strategy that only hijacks the compiler engine for the component renders (and then puts the old one back right after)?

shadoWalker89:
Hi,
I just saw your comment and also saw that you implemented that by calling the parent::evaluatePath() 👍

lancepioch:
Already added as a contributor, nice work ☕️

PR #365, opened 2019-11-13
Enable laravel Echo toOthers()

shadoWalker89:
When using laravel echo it also possible to broadcast an event to other clients except the source client.
This is done using the `toOthers()`
But in order for this to work, the socket id must be sent to the server.
This PR sends adds the socket id to the request headers


calebporzio:
Thanks for the PR @shadoWalker89, question: what system resolves the "X-Socket-ID" header? Is tha an Echo-specific thing or what? I don't see any code to handle that in the PR, so curious. 
Thanks!

shadoWalker89:
Yeah that is handeled by laravel echo, that's how he knows to exclude the client that sent the payload.
For more information check the `configuration` section under `Only To Others` in laravel docs https://laravel.com/docs/6.x/broadcasting#only-to-others

calebporzio:
Perfect, thanks!

lancepioch:
@all-contributors please add @shadoWalker89 for maintenance

Issue #372, opened 2019-11-19
Livewire null parameter bug


shadoWalker89:
**Describe the bug**
```php
class Foo
{
    public function mount($user = null, $options = [])
    {
    }
}
```

When i call it like this

```php
@livewire('foo', null, ['mode' => 'planner'])
```

I get `null` for both arguments `$user` and `$options`

When i call it like this

```php
@livewire('foo', '', ['mode' => 'planner'])
```

I get `empty string` for `$user` and `['mode' => 'planner']` for `$options`

If i remove the default null from `Foo::mount()` on the `$user`

```php
class Foo
{
    public function mount($user, $options = [])
    {
    }
}
```

And then call 

```php
@livewire('foo', null, ['mode' => 'planner'])
```

I'll get correct values `null` for `$user` and `['mode' => 'planner']` for `$options`

**Expected behavior**
Even when using a default value `null` in the `mount()` argument values should be mapped correctly and i shouldn't get `null` for the `$mode` argument

calebporzio:
Gotcha, looks like an issue that wouldn't be too hard to fix. Would love A PR with tests

shadoWalker89:
### Bug explanation
So this is what i found.
To resolve the dependencies you are using `resolveClassMethodDependencies()` of the following trait `Illuminate\Routing\RouteDependencyResolverTrait` from laravel framework. Now this method is then calling another method `resolveMethodDependencies()` on the same trait which contains the actual logic of resolving the parameters.

Let's assume i'm calling the component like this

```php
@livewire('foo', null, ['mode' => 'planner'])
```

If you check this line https://github.com/laravel/framework/blob/7ca29256b85a910585ab7492d22365c1c5a8e29b/src/Illuminate/Routing/RouteDependencyResolverTrait.php#L53
you will see that the laravel framework is using an `isset()` which of course when used with null it will return false. That line of code in laravel is checking if we have a null parameter and if a default value exists, if so it will add the default value in the parameters list.

So when the laravel method returns at this line (Now we are in livewire land)
https://github.com/livewire/livewire/blob/9f9dc2014cbabdc060795284a033128d2d1ca661/src/LivewireManager.php#L85

the `$parameters` variable will have the following value

```php
[null, null, ['mode' => 'planner']]
```
First null is the null from the original `$options` passed to the laravel `resolveClassMethodDependencies()`

Second null is added by laravel inside the `resolveClassMethodDependencies()` via `resolveMethodDependencies()`

[...]

PR #374, opened 2019-11-20
Fix mount parameter resolving

shadoWalker89:
Yes

No

Yes

This fixes the resolving of the parameters of the component mount method, when passing null values.
For more details please check this issue https://github.com/livewire/livewire/issues/372

So the change that i made on the trait copied from laravel is using `array_key_exists()` instead of `isset()` to avoid returning false for null values


calebporzio:
Great! Thanks!

Issue #382, opened 2019-11-27
Livewire does not work on Vapor environments


lucasromanojf:
**Describe the bug**

*First issue:*

Livewire throws an error when used on Vapor environments. The error details can be seen here:

https://flareapp.io/share/KPgnkJma#F63

This happens because Livewire's manifest path points to "<app_root>/bootstrap/cache/livewire-components.php" but Vapor uses "/tmp/storage/bootstrap/cache".

*Second issue:*

Vapor assets are uploaded to S3 and served using Cloudfront. If I set "asset_url" Livewire config, the assets are loaded successfuly, but they start to use the same Cloudfront URL to call the "message" Livewire endpoints. On Vapor, the assets URL and the app URL will not be the same.

**To Reproduce**

Just upload an app using Livewire features to Vapor and the issues will happen.

**Expected behavior**

Having a "manifest_path" config prop inside "livewire.php" config file (which has the default but we can change) would solve *first issue* above quickly, but I'm not sure if this is the best solution.

A possible solution for the *second issue* above is having, in addition of the "asset_url" config, a "api_url" or "app_url" or something, which would allow loading assets from the "asset_url" but calling Livewire endpoints from "api_url" or "app_url".

calebporzio:
Thanks for reporting this.

Do you feal capable of PRing fixes for this? It would be a big help.

lucasromanojf:
Hey Caleb, sure, if my suggestions are OK, I can do them and submit a PR.

lucasromanojf:
Issue solved through PR #385

PR #385, opened 2019-11-29
Make Livewire compatible with Laravel Vapor

lucasromanojf:

Yes

No

Not needed

This PR solves the issues reported on #382. Not sure if I did the best/expected solution, but I hope this helps!

Count on me!


calebporzio:
Hey @lucasromanojf, this is great!

Couple thoughts:
A) Are you able to add a test for the app_url vs asset_url config? Also one for the manifest path change
B) Is there anything we can do to make this more automatic? Like detect if the environment is Laravel Vapor and set the manifest path / asset path? Guessing no, just checking in case.

lucasromanojf:
Hey @calebporzio !

Sorry, have been busy lately. Here we go:

A) Sure, will do this as soon as I can.
B) Actually, at the beginning I was thinking about checking if some Vapor specific env var is present. But then I decided to make it configurable, to avoid issues with other environments other than Vapor. But if you think it's better to check if it is hosted on Vapor, no problem, I can do that.

georgeboot:
You can check if you're on Laravel Vapor by running:
```php
$onVapor = 'vapor' === $_ENV['SERVER_SOFTWARE'] ?? null;
```

georgeboot:
@calebporzio Can you merge please?

calebporzio:
I would love Livewire to just work on Vapor automatically, or at least group the configuration to a simple ENV variable: LIVEWIRE_SERVERLESS or LIVEWIRE_ENABLE_VAPOR

lucasromanojf:
Hey @calebporzio !

I think we can detect Vapor for the manifest file path, but I think we should keep the app_url config in addition to the asset_url config, since the assets url and the app url can be different, independently of hosting on Vapor or not. For example, I can have an app where the assets_url is "/assets" or "/js", but my app's base URL is "/admin". Anyway, we can make app_url default to assets_url, but allow app_url to be configurable. What is your opinion about this?

calebporzio:
Hey @lucasmichot, I'm sorry I didn't think this through more fully before.

Here's what this PR should contain:
A) A new config item for the manifest path

Another PR could contain the auto-detecting of a Vapor environment to set the default manifest path.

As far as the `app_url` config item, by default, Livewire does not specify a domain to load the assets and message endpoint, so they should both work (Livewire assets are "faked". They are actually just routes on the app, not static assets)

IF a vapor user decides to publish the JS assets, you are right, they would hit an error when they set the `assets_url` config. HOWEVER, there really isn't much of a reason to publish the assets. So I don't know that we need to address that.

Sorry for stringing you along, I just didn't have a clear mental picture on this issue before.

Thanks!

lucasromanojf:
Hey @calebporzio , no worries, it's important that we discuss the best solution!

The Livewire JS assets URL and the Livewire endpoint URL will always be different on Vapor. The Livewire JS assets, when deploying to Vapor, will always be on a Cloudfront URL generated by Vapor, and the Livewire endpoint URL will be the app's URL itself.

The way it is implemented currently, asset_url is both the path for the Livewire JS files and the base Livewire endpoint URL. This is why I've added the app_url config item. This way we can have different URLs for the JS assets and the Livewire endpoints.

Since this PR is for making Livewire compatible with Vapor, do you think it is necessary to do the following in another PR? Maybe I could add this to this same PR.
"Another PR could contain the auto-detecting of a Vapor environment to set the default manifest path."

The way I did this PR, Livewire will be compatible not only with Vapor, but with any situation where asset_url and app_url are different, and also any other environments where the "<project_root>/bootstrap/cache" folder is not writtable.

calebporzio:
Good point.

The part I'm getting hung up on is this:
>The Livewire JS assets, when deploying to Vapor, will always be on a Cloudfront URL generated by Vapor

Here's a code snippet from `src/LivewireServiceProvider.php`
```
public function registerRoutes()
    {
        RouteFacade::get('/livewire/livewire.js', [LivewireJavaScriptAssets::class, 'unminified']);
        RouteFacade::get('/livewire/livewire.min.js', [LivewireJavaScriptAssets::class, 'minified']);

        RouteFacade::post('/livewire/message/{name}', HttpConnectionHandler::class)
            ->middleware(config('livewire.middleware_group', 'web'));
    }
```

As you can see, both the assets and the message endpoint are served from the app. The assets aren't actually static. I would think Vapor would treat them the same as the message endpoing.

Does this make sense?

Also, yeah, I'm fine with both being in this PR. (the auto-detecting vapor thing) - thanks!

lucasromanojf:
@calebporzio I guess I understand now what you said. I will test here, but since not publishing the JS assets will work like a normal Laravel route, I guess it will work for Vapor. The additional app_url I added will be needed only for the cases where the JS assets are published.

calebporzio:
Right, and I'd rather not add the separate `app_url` until someone actually has a real problem

Let's stick with the manifest file config and auto-detection for now. (after you verify this works)

Thanks!

lucasromanojf:
Hey @calebporzio, just tested and it worked on Vapor in case of not publishing the JS assets. Then I reverted the app_url stuff and kept only:

1) manifest_path in livewire.php config file;
2) If manifest_path is null in livewire.php (default), detect if app is hosted on Vapor and then set the manifest path automatically to the Vapor one ("/tmp/storage/bootstrap/cache/livewire-components.php"), otherwise set manifest path to "<project_root>/bootstrap/cache/livewire-components.php" (the default).

calebporzio:
This is great! Thank you @lucasromanojf!

lancepioch:
@all-contributors please add @lucasmichot for code

lucasromanojf:
Hey guys I am @lucasromanojf not @lucasmichot haha!!!

lancepioch:
Thanks for your contributions! Sorry, I'm honestly not sure how that got mixed up

@all-contributors please add @lucasromanojf for code

tillkruss:
@calebporzio: We're seeing errors since this PR running on Heroku.

```
The /tmp/storage/bootstrap/cache directory must be present and writable. 
```

PR #394, opened 2019-12-07
add fill method for public properties

faustbrian:
Probably. No.

No.

Yes.

If you have a component that is a large form with a lot of fields that need to be pre-filled (when for example editing a model) it can be helpful to have a method to quickly fill the public properties of the component with values that are present on the model.

The fill method works with arrays and objects so it doesn't really matter what is passed in as long as the keys in those match the names of public properties on the component.

### DIFF
--- a/src/Concerns/InteractsWithProperties.php
+++ b/src/Concerns/InteractsWithProperties.php
@@ -120,4 +120,15 @@ trait InteractsWithProperties
             ->pluck('name')
             ->search($propertyName) !== false;
     }
+
+    public function fill($values): void
+    {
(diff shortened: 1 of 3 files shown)


calebporzio:
Perfect. (except for that silly `: void` ;) )

calebporzio:
Thanks @faustbrian !

faustbrian:
Oh yeah, forgot that Livewire doesn't use return types. Has become a habit since PHP 7 :sweat_smile:

lancepioch:
Already added as a contributor, you are a machine 🤖

PR #397, opened 2019-12-07
add method to reset public properties to initial values

faustbrian:
Probably. No.

No.

Yes.

Often times Livewire components hold temporary state like data that a user inputs into a contact form. When a user submits the form you most likely want to reset the components public properties to their initial state. If you have 10 fields that can get rather tedious as there is currently no way of reseting all public properties to their initial values.


calebporzio:
I'm confused that this test works?

I don't think a feature like this is actually possible without passing the "initial" values to the front-end back and forth. If you think about it, Livewire is stateless, and this would only work for "initial values" at the beginning of the request from a Livewire update. Make sense? Am I missing something?

faustbrian:
Yeah wasn't entirely sure why the test works because I would've expected `initialPublicProperties` to be empty when making a new request.

I'll close this for now as it isn't much work to implement a reset method on a per component basis if necessary or adding a trait of my own.

calebporzio:
Yep, thanks Brian, it's definitely a good use case though

PR #400, opened 2019-12-08
remove render from base class checks

### LINKED ISSUE #393: `render` via listeners results in NonPublicComponentMethodCall
Prior to 0.5.0 it was possible to omit the `render` method from a component and use it for a listener. After 0.5.0 the `render` method has to be defined, which is somewhat inconvenient if you're not doing anything with the view other than rendering it.

**Before**
```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;

{
    protected $listeners = [
        'refreshSomething' => 'render',
    ];
}

```

**After**
```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;

{
    protected $listeners = [
        'refreshSomething' => 'render',
    ];

    public function render()
    {
        return view('livewire.some-component');
    }
}

```

@calebporzio is this a regression and could be caused by components now being macroable or is this an intended breaking change that went unnoticed?

faustbrian:
Yes. No but fixes https://github.com/livewire/livewire/issues/393.

No.

No. Wasn't sure how to best test this without a mock to assert a method call.

Excludes the `render` method from the base class checks as it is an exception to how the rest of the methods work and are used. If `render` is omitted it will be called by the method on the abstract that is extended which means that if you don't define it in your own class it will result in an error if an even listener tries to call it because it only exists on the abstract but not on your component itself.

### DIFF
--- a/src/Concerns/HandlesActions.php
+++ b/src/Concerns/HandlesActions.php
@@ -85,6 +85,11 @@ trait HandlesActions
     {
         return collect((new \ReflectionClass($this))->getMethods(\ReflectionMethod::IS_PUBLIC))
             ->reject(function ($method) {
+                // The "render" method is a special case. This method might be called by event listeners or other ways.
+                if ($method === 'render') {
+                    return false;
+                }
+
                 return $method->class === self::class;
             })
             ->pluck('name')


calebporzio:
Perfect. Thanks @faustbrian!

lancepioch:
@all-contributors please add @faustbrian for maintenance

PR #401, opened 2019-12-08
avoid property name conflicts in tests

### LINKED ISSUE #381: Testing with prop $name conflict with Livewire\Testing\TestableLivewire name prop
**Describe the bug**
```
$form = Livewire::test(MyForm::class);
$form->set('name', 'testing name');

$this->assertEquals('testing name', $form->name);
```

**Unit test output**
```
1) Tests\Feature\Form::livewire_form_test
--- Expected
+++ Actual
@@ @@
-'testing name'
+'dS6VqTPZ2ngRNl4hFQUn'
```

```
Livewire\Testing\TestableLivewire^ {#326
  +name: "dS6VqTPZ2ngRNl4hFQUn" <--- This is the prop that will be tested.
  +id: "ByH0BEbqSa0muQ4uz1u5"
  +children: []
  +checksum: "11226a9b30ff8ec90f5558be00f878a8c634616c0ca84ddd5d0c7ad8bba22319"
  +prefix: "wire"
 
```

**Expected behavior**
It should collect the name prop from the data object of Livewire\Testing\TestableLivewire

```
Livewire\Testing\TestableLivewire^ {#326
+data: array:4 [
    "name" => "testing name"
  ]

```

faustbrian:
Yes. Fixes https://github.com/livewire/livewire/issues/381.

No.

Yes.

Prefixes the randomly generated `name` property with `component` to avoid naming conflicts and false results in tests.

### DIFF
--- a/src/Testing/Concerns/MakesCallsToComponent.php
+++ b/src/Testing/Concerns/MakesCallsToComponent.php
@@ -64,7 +64,7 @@ trait MakesCallsToComponent
 
         $result = $handler->handle([
             'id' => $this->id,
-            'name' => $this->name,
+            'name' => $this->componentName,
             'data' => $this->data,
             'children' => $this->children,
             'checksum' => $this->checksum,
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -7,7 +7,7 @@ use Livewire\Livewire;
 
 class TestableLivewire
 {
-    public $name;
+    public $componentName;
     public $id;
     public $children;
     public $checksum;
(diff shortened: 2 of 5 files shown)


faustbrian:
@calebporzio I think it would be a good idea to start to use a convention like `componentProperty` for internal things of Livewire to avoid naming conflicts with components since things like ID and name are not the only common names for properties.

Potential prefixes for internal properties could be `component`, `internal` or `_` to avoid naming conflicts. Even better maybe would be to use an array that holds all state that belongs to Livewire and not the part that is implemented by developers that extend the base component.

calebporzio:
Thanks @faustbrian!

And yeah, you're totally right. A dedicated array, or `_` prefixed props is probably the way to go. Thanks.

lancepioch:
@all-contributors please add @faustbrian for test

Issue #415, opened 2019-12-13
Make PassPublicPropertiesToView an Interface


faustbrian:
I just saw the new `PassPublicPropertiesToView` trait and saw that it is completely empty. What was the reason for making it a trait, are there any plans to add methods to it?

I think `class ShowUser extends Component implements PassPublicPropertiesToView` might be a better implementation in the sense that the interface can dictate how something should behave as opposed to a trait that should actually contain something that expands the functionality of whatever it is added to.

PR #416, opened 2019-12-13
turn PassPublicPropertiesToView trait into an interface

faustbrian:
Maybe. Yes https://github.com/livewire/livewire/issues/415.

No.

Yes.

Explained in https://github.com/livewire/livewire/issues/415. Only thing to note might be that the name `PassesPublicPropertiesToView` or `HasPublicViewProperties` might be better for an interface.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -92,8 +92,7 @@ abstract class Component
             $errorBag = $errors ?: ($view->errors ?: $this->getErrorBag())
         );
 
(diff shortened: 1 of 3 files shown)


calebporzio:
Hmm, you're right that it is a tad odd being empty. However, personally, I would just rather add traits for extra behavior (even if it doesn't contain the behavior it adds). I think an interface would be just as odd.

Might just be me, but I like it as a trait better for now. Thanks though @faustbrian.

PR #422, opened 2019-12-18
Reload the page if it hasn't already changed, resolves #289

### LINKED ISSUE #289: Redirecting to the current page with anchors
**Describe the bug**
Livewire redirects will not reload the current page if the route contains anchors, it will just add the anchor to the URL. For example if you are on the page `"/home"` and you return the following responses from your Livewire controller:

`return redirect('/home'); //Page properly reloads itself`

`return redirect('/home#anchor')` //Page will not reload, just appends the anchor to the url`

Notably, this only happens if you are trying to return back to the current page you are on. If you try to return a redirect to a different page with anchors, it works as intended. 

**Expected behavior**
I expect a redirect to reload the entire page, regardless if it's the same url or not (and with or without anchors). Oftentimes I have users taking actions with Livewire and when they are complete I want to reload the current page with the updates information.

Also, with the current behavior, Livewire will no longer work on a page if a controller returns a redirect and the page doesn't redirect.  

**Browser:**
I have tried in Safari, Chrome, and Firefox and get the same response.

lancepioch:

#289

No

No

Resolves the issue. If the redirect happens normally, then this extra code won't be fired. However, if the url is the same with the anchor changing, then this fires


calebporzio:
Hey @lancepioch  - thanks for the contribution.

I actually don't think this works though. Have you tried this locally?

I think this would just reload the page no matter what a user is redirecting to?

Issue #389, opened 2019-12-03
 Livewire\Exceptions\ComponentNotFoundException


bedus-creation:
I tried to change the default components directory, It couldn't resolve the components.

**To Reproduce**
1. Change the config
``` 
'class_namespace' => 'App\\Domain',
```
2. This is my components class
App\Domain\Settings\Components\PostIndex
3. 
```
Route::group(['namespace' => 'Settings/Components'], function () {
    Route::livewire('/home', 'PostIndex');
});
```
4. It says Components not found

bedus-creation:
Same problem with view_path components
**To Reproduce**
1. change the default view_path
```
'view_path' => resource_path('views'),
```
2. I have livewire components in this directory 
```
views/settings/samanKoKisim/components/index
```
3. And I couldn't access
```
@livewire('settings.samanKoKisim.components.index')
```

HenkMaller:
Did you run the `livewire:discover` command?

faustbrian:
@bedus-creation as @HenkMaller said you'll need to run `php artisan livewire:discover` if you ever change the name or location of your components or views.

faustbrian:
Should be safe to be closed after https://github.com/livewire/livewire/pull/429 @calebporzio @lancepioch

Issue #165, opened 2019-08-14
"Unable to find component" on sub-domain


NT-daviddefoe:
**Describe the bug**
I created a Livewire component on my local machine.  When I uploaded the code to my testing server it was showing the component @livewire as a text string in the browser (@livewire('inventory.inventory-pricing', $vehicle->prices, $vehicle->id, $vehicle->vin)).

I then made a very simple component and uploaded to the test server (test.domain.com) and getting "Unable to find component" error.

**To Reproduce**
1. run: php artisan make:livewire livewirejunk
2. In web.php add: Route::view('/livewire', 'livewire');
3. Source of livewire.blade.php:
```
@extends('layouts.app')

@section('headers')

<style>
.fixed-sn main {
    margin-left: 2% !important;
    margin-right: 2% !important;
}
</style>
@endsection

@section('content')
    <h1 class="text-center">Livewire Test!</h1>

    <div>
        @livewire('livewirejunk')
    </div>
@endsection

@section('scripts')
@livewireAssets
@endsection
```

4. Source of livewire\livewirejunk.blade.php:
```
<div>
    {{-- Care about people's approval and you will be their prisoner. --}}
    <div>
        Lorem ipsum dolor sit amet, consectetur adipisicing elit. Placeat unde repellat ea incidunt. Quos sequi aspernatur delectus vero iste, ut voluptates minima voluptatibus placeat, dolore doloribus iure expedita quia assumenda!
    </div>
</div>
```

5. Source of App\Http\Livewire\livewirejunk.php:
```
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class Livewirejunk extends Component
{
    public function render()
    {
        return view('livewire.livewirejunk');
    }
}
```

**Expected Result**
![image](https://user-images.githubusercontent.com/47991824/63052220-c88f5600-beac-11e9-8d62-f00832260de6.png)

**Error**
"Unable to find component: [livewirejunk] (View: /var/www/html/projectfolder/resources/views/livewire.blade.php)"

**Test Server**
 - Ubuntu 18.04.6 LTS
 - Server version: Apache/2.4.18 (Ubuntu)
 - Node v4.2.6
 - Laravel v5.8.32

lalov:
Have you tried to rediscover the components? It helped me: 

php artisan livewire:discover

intellow:
@lalov that command fixed my problem. Thanks!

cosmicvibes:
[...]

PR #429, opened 2019-12-19
Resolve ComponentNotFoundException when Manifest is Out of Date

intellow:
Yes. There are open issues that this will solve: #389 #165 

No.

There was an existing test that required a manual command to rebuild the manifest. I changed that one to remove the command to rebuild it, and verified that it is rebuilt automatically with no user input necessary.

There are many instances where the manifest can become out of date and throw an exception. Before any exception is thrown, I am rebuilding the manifest and double checking to see if the component can be found in the updated manifest.

I also saw a generic exception might be thrown for a component not found, and I changed it to the more specific exception class.

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -2,7 +2,6 @@
 
 namespace Livewire;
 
-use Exception;
 use Illuminate\Support\Str;
 use Illuminate\Support\Fluent;
 use Livewire\Testing\TestableLivewire;
@@ -59,6 +58,13 @@ class LivewireManager
             $this->componentAliases[$alias] ?? $finder->find($alias)
         );
 
+        if(!$class) {
+            $finder->build();
+            $class = $class ?: (
+                $this->componentAliases[$alias] ?? $finder->find($alias)
+            );
+        }
+
         throw_unless($class, new ComponentNotFoundException(
             "Unable to find component: [{$alias}]"
         ));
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @intellow!! Take a look at the refactors I made to your contributions in `master` if you get a chance. Want to make sure you agree with them:
https://github.com/livewire/livewire/commit/ba509a846bcf56ca2d9d2a00462ac986b0f07831

intellow:
Love the comments. Now I actually understand what's going on in there. Thanks!

I did notice you removed the check I added in the activate method, so let me explain more why I put that there. It's to cover the use case of someone changing their component class_namespace in the config. Take a look:

The following will find a match in the **outdated** manifest and return it to the activate method.
```
$class = $class ?: (
            // Let's first check if the user registered the component using:
            // Livewire::component('name', [Livewire component class]);
            // If not, we'll look in the auto-discovery manifest.
            $this->componentAliases[$alias] ?? $finder->find($alias)
        );
```
Now we end up here:
```
public function activate($component, $id)
    {
        $componentClass = $this->getComponentClass($component);
        // when the class_namespace in the config has changed and components have moved but
        // the manifest HAS NOT been regenerated, $componentClass points to a class that doesn't exist
        // if we don't check (class_exists($componentClass)) and try again if it fails, then
        // this use case throws the exception
        throw_unless(class_exists($componentClass), new ComponentNotFoundException(
            "Component [{$component}] class not found: [{$componentClass}]"
        ));

        return new $componentClass($id);
    }
```

The weird thing is when I tested this use case, I can actually have an existing component and decide I want my Livewire components in a new directory like /app/Http/Components. I can update the config, move my components (and update their namespace), and everything just works without needing to run `php artisan livewire:discover`.

However, if I change my mind and try to move everything back to the app/Http/Livewire directory, I do the same thing (update config, move files and update their namespace) and try to load a page with a Livewire component, it throws an Exception. Of course, the exception can be resolved by running `php artisan livewire:discover`, but I'm curious why this works changing your directory from the default to something else, but not moving from something else to the default.

fgilio:
Thanks for this!

Issue #483, opened 2019-12-21
Can't return file download from method


mokhosh:
My component represents a digital product. I'm using Spatie media library to attach files to my products. When I click on the *download* button, it should start downloading a file:

```php
// wire:click="download"

public function download()
{
    $files = Product::find($this->productId)->getMedia('download');
    return MediaStream::create($this->title)->addMedia($files);
}
```

These two lines work just fine if I put them in a normal controller, but not in livewire. Are there any workarounds?
The only quick fix that I can think of, and I don't like at all, is to redirect the user to another route that downloads the file, ugh.

mokhosh:
I tried adding a route for downloading the file and redirecting to it. It starts downloading but the component just freezes. Nothing works, the loading state never ends, other buttons don't work either.

So, if we're not going to fix this fundamentally by adding support for file download responses, we need to see what's going on with this workaround.

calebporzio:
Yeah, would you be able to investigate this further? It WOULD be good to support this case.

mokhosh:
I did.

I think the problem is the way `handleResponse` in `Component/index.js` behaves when `redirectTo` is set. It just changes the url and returns (line 160) and the rest of the function is never executed if you redirect.
This is fine if you actually move to another page, but since when `Content Type` or `Disposition` is set in the response headers, the browser just starts downloading and keeps the current url without refreshing or anything, it gets messy, because livewire has said its goodbyes but the door is locked, which is really awkward.

I thought we could fix it one way or the other, but now I think we have to fix both problems:
1) keep doing what you were doing if the redirect is not actually changing the url, or at least do a refresh to get out of the freez
2) add a method for sending a file response

mokhosh:
@lancepioch It's a bug report + a feature request, and an accepted call to collaboration. No questions.

lancepioch:
Agreed, it's definitely a bug!

intellow:
Jumping in here to say that I'm trying to use Livewire to download a file as well. I haven't come up with a good way to do this yet but will update if I find a solution.

calebporzio:
Awesome, I think the answer to this will end up being the same as this question:

"How do I trigger a file download from an ajax response"

If there is a good answer to that. This is possible.

I'm going to let you all figure this out though, because I have other priorities ATM, thanks for pitching in!

mokhosh:
@calebporzio looks like we have 3 options:

a) a more modern options that uses js to download the file but:
[...]

PR #491, opened 2019-12-24
fix file download redirect: issue 483

mokhosh:
Yes https://github.com/livewire/livewire/issues/483

No

I don't know if this can be tested and I didn't go through the tests, but I did check if not `return`ing after a redirect to a file download will let the rest of the code to be executed, and I think this will fix the problem. That line seems redundant anyway. Isn't it?

Currently livewire's `handleResponse` `return`s after redirecting if `redirectTo` is set, which is unnecessary if there's an actual url change AND breaks the page if there isn't (eg. in case the redirect url is just a file to be downloaded). By getting rid of the `return` we keep the current functionality and fix the problem.

### DIFF
--- a/src/js/Component/index.js
+++ b/src/js/Component/index.js
@@ -156,8 +156,6 @@ export default class Component {
         // This means "$this->redirect()" was called in the component. let's just bail and redirect.
         if (response.redirectTo) {
             this.redirect(response.redirectTo)
-
-            return
         }
 
         store.callHook('responseReceived', this, response)


mokhosh:
I think this pull request might fix [this](https://github.com/livewire/livewire/issues/289) issue too.
[this](https://github.com/livewire/livewire/pull/422) doesn't seem like a great solution.

mokhosh:
OK, further investigations. That's not the problem.

I published livewire assets, modified the js, and it didn't fix the problem. 😞 
I even tried a straight up `a` tag with my download url and even that makes livewire stop working.
I tried to read the code and understand what's going on, but it just got too complicated.

What do we do @calebporzio ?

calebporzio:
Closing this, for now, I'm sure there is a solution, but I don't think this early return is it. We can keep the discussion going in the issues.

MOHAMMED-ABD-RAZAQ:
I wish I knew the problem in the library before I used it in my project and I could almost complete it, but I stopped at this problem 👎

PR #498, opened 2019-12-28
Fix typo "Eloquen" to "Eloquent" in exception name

imliam:
Just a small typo fix in a class name to add the "t" to "Eloquent"


calebporzio:
Thanks @imliam!

PR #521, opened 2020-01-13
fix vapor condition

### LINKED ISSUE #520: The /tmp/storage/bootstrap/cache directory must be present and writable.
**Describe the bug**

#385 broke our app:

```
The /tmp/storage/bootstrap/cache directory must be present and writable. 
```

We reverted back to v0.5.2 and it's working just fine.

**Additional context**

It doesn't work with `livewire.manifest_path` set to `null` or not set at all.

The manifest path seems to evaluate to `/app/bootstrap/cache/livewire-components.php`.

tillkruss:
Fixes #520, which was introduced in v0.5.3 via #385.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -51,14 +51,15 @@ class LivewireServiceProvider extends ServiceProvider
         $this->app->singleton('livewire', LivewireManager::class);
 
         $this->app->singleton(LivewireComponentsFinder::class, function () {
-            $isHostedOnVapor = $_ENV['SERVER_SOFTWARE'] ?? null === 'vapor';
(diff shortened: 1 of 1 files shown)


src/LivewireServiceProvider.php
```diff
@@ -51,14 +51,15 @@ public function register()
         $this->app->singleton('livewire', LivewireManager::class);
 
         $this->app->singleton(LivewireComponentsFinder::class, function () {
-            $isHostedOnVapor = $_ENV['SERVER_SOFTWARE'] ?? null === 'vapor';
+            $isHostedOnVapor = ($_ENV['SERVER_SOFTWARE'] ?? null) === 'vapor';
+
```
tillkruss: @calebporzio: You should tweet this change to educate the people 😉

calebporzio:
Thanks @tillkruss!

PR #525, opened 2020-01-15
Make it compatible with PHP 7.1

lalov:
Removed some trailing commas to make it work with PHP 7.1

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -28,7 +28,7 @@ use Livewire\Commands\{
     MvCommand,
     RmCommand,
     StubCommand,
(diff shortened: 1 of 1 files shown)


cbaconnier:
Shouldn't we add PHP 7.1 to Github actions test workflow ?

https://github.com/livewire/livewire/blob/master/.github/workflows/test.yml#L17

lalov:
Should I add PHP 7.1 to Github actions test workflow before it can be pulled in?

calebporzio:
Thanks!

lancepioch:
@all-contributors please add @lalov for maintenance

PR #529, opened 2020-01-17
Refactor polyfills and update IE support

harrygulliford:

Unsure.

Hopefully a minor refactor of the current pollyfills and adding 2x new polyfills is ok.

No

A minor clean-up/refactor of the current polyfills implementation. Also added Object.assign and String.prototype.startsWith to support IE (currently broken on IE11).

It may be worth considering separating out the polyfills into a seperate file, so people can include it only if they want to support older browsers? Similar to @stimulus/polyfills (https://github.com/stimulusjs/stimulus/tree/master/packages/%40stimulus/polyfills).


calebporzio:
Thanks!

lancepioch:
@all-contributors please add @harrygulliford for maintenance

PR #531, opened 2020-01-17
Stop creation of livewire class and view if name is not allowed / reserved

intellow:
When you try to make a component with a reserved name, everything blows up. I think it's needed but there's no issue.

Nope, this is self contained.

Yes

When you make:livewire parent or make:livewire component the classes are created and everything breaks. I have added a check to make sure those classes cannot be created. If there is a definitive list of what class names are reserved, it would be great to use that. Right now I've just captured three I've seen (Parent, Component, Interface)

👍

### DIFF
--- a/src/Commands/MakeCommand.php
+++ b/src/Commands/MakeCommand.php
@@ -19,7 +19,9 @@ class MakeCommand extends FileManipulationCommand
             $this->option('stub')
         );
 
-        if(!$this->ensureClassNameIsNotReserved()) {
+        if($this->isReservedClassName($name = $this->parser->className())) {
+            $this->line("<options=bold,reverse;fg=red> WHOOPS! </> 😳 \n");
+            $this->line("<fg=red;options=bold>Class is reserved:</> {$name}");
             return;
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @intellow - good PR. Made a few tiny updates (will add GH review comments on the lines for you to look at)

src/Commands/MakeCommand.php
```diff
@@ -19,6 +19,12 @@ public function handle()
             $this->option('stub')
         );
 
+        if($this->isReservedClassName($name = $this->parser->className())) {
```
calebporzio: Changed from a double negative "!$this->ensureClassNameIsNotReserved()" to a single positive "$this->isReservedClassName"

Also, "ensureClassNameIsNotReserved" both performs an action (throwing the message) AND returns a boolean (weather or not the class is reserved). I generally like to have a method responsable for only one of those things.

intellow: Great to know for future PRs. Thanks!

PR #535, opened 2020-01-18
Laravel 7 support

driesvints:
This contains the bare minimum to get Livewire installed for Laravel 7. This was we can already start testing out Livewire for the upcoming Laravel 7 release. Tests seem to be passing. I don't know of any breaking changes by hand that could affect Livewire. I'll try it out on the Eventy codebase once this is merged and see if I stumble upon anything.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: true
       matrix:
         php: [7.2, 7.3, 7.4]
-        laravel: [5.6.*, 5.7.*, 5.8.*, 6.*]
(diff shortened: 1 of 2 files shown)


driesvints:
Hmm, it seems it's not installing dev-master from Laravel. Need to investigate why.

driesvints:
So it seems that the matrix is set on the build but it's never applied. So Livewire is only tested on 6.x atm for every matrix version: https://github.com/livewire/livewire/pull/535/checks#step:4:80

driesvints:
I'll try to fix the matrix build first before I send this in. So the PR stays clean.

PR #536, opened 2020-01-18
Fix matrix build and Laravel 7 support

driesvints:
I took me a few hours but I finally got the matrix build to run and Laravel 7 support in. I discovered that the matrix build wasn't really running and was always testing on 6.x. I took some inspiration from Spatie's builds to set things up. And also added Laravel 7 support in the meantime. I also cleaned up the composer.json file to be more similar to Laravel's one. I needed to set minimum stability for example. I also added a minimum PHP version to it which is the same one as the minimum Laravel one (5.6).

As you can see I needed to do quite some circumventing to get things to work. In the future it might be best to consider to drop support for older Laravel versions. 

Feel free to add comments on the pieces of code where you have questions and I'll answer them.


lancepioch:
This looks pretty awesome!

These builds failed because they couldn't run `composer require`:

* PHP:7.2 / Laravel:5.6.*(prefer-lowest)
* PHP:7.2 / Laravel:5.7.*(prefer-lowest)
* PHP:7.3 / Laravel:5.6.*(prefer-lowest)

How can these be fixed? I'm assuming because of the prefer-lowest there needs to be some minimum versions pinned possibly?

And the last build (PHP:7.3 / Laravel:5.7.*(prefer-lowest)) failed because it couldn't find the class Mockery. I feel like this test can just be changed to resolve this issue.

Finally, Laravel 5.6, 5.7, 5.8 all can run with PHP 7.1. Should that be added as well for just those versions? If not, then the composer file should lock that version too.

driesvints:
@lancepioch heya, thanks!

The builds that you listed were actually cancelled because of the other failing build (see the annotations in the bottom right). I suspect that composer will run just fine.

Screenshot 2020-01-19 at 09 59 24

Thanks on the note about PHP 7.1! I've added these builds now locally. I'll push them up in a bit.

I spotted the Mockery failure too late yesterday evening. I'll get that fixed as well 👍

driesvints:
Tests are passing 🥳

calebporzio:
This is great  @driesvints - thanks a ton!

As soon as all the checks report passing, I'll merge this in.

driesvints:
Guess we can merge this then 🥳

lancepioch:
@all-contributors please add @driesvints for code

Issue #541, opened 2020-01-21
Unable to build livewire after switch to rollup


nuernbergerA:
**Describe the bug**
After you replace laravel-mix with rollup #539 i'm unable to build livewire
 ##

**To Reproduce**
1. run  `./node_modules/rollup/dist/bin/rollup -c rollup.config.js`

**Expected behavior**
Successful build

**Additional context**
We should reflect changes also in [package.json](https://github.com/livewire/livewire/blob/065d2562c8c1d77ad7bc1d2d4ef47a93e55ba7db/package.json)

```js
{
  //...
  "scripts": {
    //...
    "unminified": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
    "minified": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
    "watch": "npm run unminified -- --watch",
    "build": "npm run unminified && npm run minified"
  },
  //...
}
```

harrygulliford:
@nuernbergerA Could you let me know if #543 fixes the issue?

nuernbergerA:
@harrygulliford unfortunately i encounter this error:
```console
$ npm run build
npm WARN npm npm does not support Node.js v10.15.2
npm WARN npm You should probably upgrade to a newer version of node as we
npm WARN npm can't make any promises that npm will work with this version.
npm WARN npm Supported releases of Node.js are the latest release of 4, 6, 7, 8, 9.
npm WARN npm You can find the latest version at https://nodejs.org/

> laravel-livewire@ build /home/adrian/code/livewire
> npx rollup -c


src/js/index.js → dist/livewire.js...
[!] (plugin Rollup Core) Error: Could not load /home/adrian/code/livewire/src/js/component/LoadingStates (imported by /home/adrian/code/livewire/src/js/index.js): ENOENT: no such file or directory, open '/home/adrian/code/livewire/src/js/component/LoadingStates'
Error: Could not load /home/adrian/code/livewire/src/js/component/LoadingStates (imported by /home/adrian/code/livewire/src/js/index.js): ENOENT: no such file or directory, open '/home/adrian/code/livewire/src/js/component/LoadingStates'

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! laravel-livewire@ build: `npx rollup -c`
[...]

Issue #534, opened 2020-01-18
Confusing $set behaviour


mokhosh:
This will throw `Unable to set component data. Public property [$1] not found on component`:
```php
<a href="#" wire:click.prevent="$set('showChat',true)">
```
After a lot of head scratching I found out that this will work as expected:
```php
<a href="#" wire:click.prevent="$set('showChat', true)">
```
Not putting a space after the comma and before the `value`, will make livewire want to evaluate the string for some reason. It evaluates `true` to `1` and tries to find a public property with that name.

**To Reproduce**
Just omit the space before tha `value`.

**Expected behavior**
The space should not have any effect.

cbaconnier:
I believe, this is related with https://github.com/livewire/livewire/issues/517

nuernbergerA:
I will open a PR to fix this after #541 is fixed

PR #542, opened 2020-01-21
make space around paramter delimiter optional

### LINKED ISSUE #517: Strange behavior with comma separated parameter
**Describe the bug**
Strange behavior with comma separated parameter

**To Reproduce**
1) Install latest Laravel and latest livewire (0.5.3)
2) Create livewire component:
`$ php artisan make:livewire test`
3) Update view and component class:
resources/views/livewire/test.blade.php:
```
<div>
    @php
    $text = 'test, test';
    @endphp
    <button wire:click="test('{{ $text }}')">Test</button>
</div>
```

app/Http/Livewire/Test.php
```
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class Test extends Component
{
    private $text;

    public function render()
    {
        return view('livewire.test');
    }

    public function test($text)
    {
        $this->text = $text;
    }
}
```
**Expected behavior**
In browser console:
```
SyntaxError: '' literal not terminated before end of script
```

But if you change the view to (delete one space in `$text`):
```
<div>
    @php
    $text = 'test,test';
    @endphp
    <button wire:click="test('{{ $text }}')">Test</button>
</div>
```
Everything works fine. 

 - Browser is any: FF, Chrome, Safari and etc.

nuernbergerA:
Fixes #517 and #534 

due to #541 there is currently no compiled versionand test contained


tests/js/component_api.spec.js
```diff
@@ -22,6 +22,22 @@ test('set component data item value', async () => {
         expect(payload.actionQueue[0].payload.name).toEqual('foo')
         expect(payload.actionQueue[0].payload.value).toEqual('bar')
     })
+
+    component.set('foo','bar')
```
calebporzio: I like the PR, but I don't think this test actually excercises the change.

You would need to test for the actual HTML string being used for an event handler or something. Something like:

`<button wire:click="callSomething('foo' , 'bar')">`

Does that make sense?

calebporzio (changes requested):

nuernbergerA:
replaced by #551

tests/js/component_api.spec.js
nuernbergerA: totally agree, opened a new PR due to rebase issus #551

PR #547, opened 2020-01-22
Add Vue Support back in with a single file for easier extraction later

calebporzio:


jgile:
Vue still does not display after first render in v0.7.2.

calebporzio:
v0.7.3 should have the fix - please verify when you get a chance

jgile:
looks good.  thanks.

PR #551, opened 2020-01-22
[JS] parameter parsing

### LINKED ISSUE #534: Confusing $set behaviour

nuernbergerA:
Fixes #534 and #517


calebporzio:
Thanks!

lancepioch:
@all-contributors please add @nuernbergerA for code

PR #556, opened 2020-01-23
fix error assertion check on undefined index

faustbrian:
Yes. No.

No.

No.

Take the following code where `token` will not be present in the errors. You will receive an error saying `ErrorException: Undefined index: token` instead of the expected `Component has no [required] errors for [token] attribute. Failed asserting that an array contains 'required'.` from PHPUnit.

```php
/** @test */
public function fails_to_store(): void
{
    Livewire::test(Component::class)
        ->set('name', 'John Doe')
        ->call('store')
        ->assertHasErrors([
            'name'  => 'required',
            'token' => 'required',
        ]);
}
```

The tests will fail but only because of the way an undefined index tried to be accessed instead of an actual code error in your component.

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Testing\Concerns;
 
+use Illuminate\Support\Arr;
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks Brian!

PR #564, opened 2020-01-30
[JS] Changed the way the query string is updated

tormjens:
I had an issue where I tried to pass an array/object to the querystring, which only resulted in the JS adding ?foo=[object Object] to my query string.

The `query-string` library did not support nested objects, and the alternatives either weren’t ES6 compatible or didn’t have the two-way functionality needed.

Changed this so that the JS is now using a custom implementation of parsing and stringify’ing the query string. This implementation supports nested objects and should be compatible with the previous implementation.

And, there's a JS test for it.


lancepioch:
@all-contributors please add @tormjens for code

PR #568, opened 2020-01-31
Change to use `component` rather than `yo`

tormjens:
This threw a error since `yo` is undefined. Looking at the blame the first argument in this was function was previously named `yo`.


calebporzio:
Oops, thanks so much!

Issue #201, opened 2019-08-28
Move js Echo logic to EchoManager


davidpiesse:
Along with other splitting of Component logic I think moving the Echo listener logic to its own manager looks nice.

Made the changes but want to see if its desired / wanted.

```js
        this.echoManager = new EchoManager(this)
        //...
        this.echoManager.registerListeners()
        // this.registerEchoListeners()
``` 

All tests pass as is :D

PR #570, opened 2020-01-31
Implement an Echo Manager

davidpiesse:
Updated for the new JS code styling.
Issue #201 - Bam done!

Sorry it took so long!


calebporzio:
Thanks @davidpiesse - sorry I took so long to get to this. You da man!

Issue #560, opened 2020-01-25
wire:click only sends the last parameter


snapey:
When wire:click is actioned, the network request only sends whatever is last in the parameter list.

blade;
```
<button wire:click="confirmDelete('a','b','c')"
```

Network request;
```
{"id":"RML8hMYmFd3eM3TtGuAQ","data":[],"name":"grid","checksum":"1a5d8f51f2272d0333c88ba8dad91de5f980e9af5f2caa3bb2484a65537d37b4","children":[],"actionQueue":[{"type":"callMethod","payload":{"method":"confirmDelete","params":["c"]}}]}
```
Blade;
```
<button wire:click="confirmDelete('a','b')"
```
Network;
```
{"id":"LvxbRJT9gsKGGUcg8Y6l","data":[],"name":"grid","checksum":"f95896a5cf82e4025c93b1a80758eca83d97fe969e1f101c90856f16c6cb9032","children":[],"actionQueue":[{"type":"callMethod","payload":{"method":"confirmDelete","params":["b"]}}]}
```
The complete parameter list should be sent, but only the last parameter is present in the request params

Version 0.7.0

nuernbergerA:
Should be fixed by #551

nuernbergerA:
@snapey can you confirm if it works with the latest release?

snapey:
Confirmed working in v0.7.2

I don't know if this is a problem or not, but there are no errors or warnings for different number of parameters on either side.

Eg, Method expects 2 parameters, client sent one / Method expects 1 parameter, client sent 2

snapey:
Is it reasonable for me to expect this to be sent as an array plus a string to the component?
```
wire:click="addConditionRow(['a', 'b'], 'c')"
```
Throws js error
```
Uncaught SyntaxError: Unexpected end of input
    at directive.js:54
    at Array.map (<anonymous>)
    at ElementDirective.parseOutMethodAndParams (directive.js:54)
    at ElementDirective.get method [as method] (directive.js:20)
    at node_initializer.js:135
    at Component.callAfterModelDebounce (index.js:413)
    at HTMLButtonElement.handler (node_initializer.js:127)
```

nuernbergerA:
@snapey should work since [v0.7.3](https://github.com/livewire/livewire/releases/tag/v0.7.3)

lancepioch:
@snapey Does the latest version resolve your issue?

snapey:
All good now. Sorry for the delay in replying.

PR #571, opened 2020-01-31
[JS] wire:click paramters

### LINKED ISSUE #517: Strange behavior with comma separated parameter

nuernbergerA:
fixes #517 

The current implementation will split the parameter string by comma without checking if the comma is part of a string inside a parameter

```js
let paramString = "'foo, bar', true , 'baz',null,'x,y'"
let params = paramString.split(/\s?,\s?/)

//result
["'foo", "bar'", "true", "'baz'", "null", "'x", "y'"]
```

With this PR my focus is not a delimiter, its the paramerter itself.
A paramter must follow some rules:
* be a constance sequence of word-charaters (true, null, someVar)
* start/end with ' ('some string')

so the above example code will look like this:
```js
let paramString = "'foo, bar', true , 'baz',null,'x,y'"
let params = paramString.match(/((?:'[^']*')|\w*)/g).filter(Boolean)
```

the only 'downside' now is that delimiters are optional so this would also be valid
```js
let paramString = "'foo, bar' true 'baz'null'x,y'"
let params = paramString.match(/((?:'[^']*')|\w*)/g).filter(Boolean)
```
or
```js
let paramString = "'foo, bar'|true|'baz'|null|'x,y'"
let params = paramString.match(/((?:'[^']*')|\w*)/g).filter(Boolean)
```

will all result in
```js
["'foo, bar'", "true", "'baz'", "null", "'x,y'"]
```

If you want to ensure a comma is used a a sperator we need a more complex regex


nuernbergerA:
Ok after some research i found a solution that might fit better.

- use `.split` again (detects only comma outside apostrophe, with optional space around)
- keep `.filter(Boolean)` to handle empty brackets `callSomething()`

nuernbergerA:
if array support is needed as mentioned in #560 we have to tweak the regex again
```regex
/ ?, ?(?=(?:[^'\[\]]|[\[\]'][^'\[\]]*['\[\]])*$)/
```

calebporzio:
Ugh, maintaining a regex for this kind of thing is really dark. Are there other pure-js solutions to this? like doing an actual function call on a made-up function to get the parameters passed in? I don't know, just seems like going further down the regex path is a bad idea. Thoughts?

nuernbergerA:
I would stick to the current state of this PR to fix the initial issue and open up a new issue to discuss futher steps.

calebporzio:
Ok, I can abide.

Can we expand the regex to include other forms of quotes? backticks and double quotes?

Sometimes people do things like this: `<input wire:click='doSomething("with-this")'`

I'd like to at least support double-quotes.

Thanks for this PR!

nuernbergerA:
[Visual Regex](https://embed.ihateregex.io/make/JUMyJUEwJTNGJTJDJTIwJTNGKCUzRiUzRCglM0YlM0ElNUIlNUUlMjInJTYwJTVDJTVDJTVCJTVDJTVDJTVEJTVEJTdDJyU1QiU1RSclNUQqJyU3QyUyMiU1QiU1RSUyMiU1RColMjIlN0MlNjAlNUIlNUUlNjAlNUQqJTYwJTdDJTVDJTVDJTVCJTVCJTVFJTVDJTVDJTVEJTVEKiU1QyU1QyU1RCkqJTI0KQ)
```regex
 ?, ?(?=(?:[^"'`\[\]]|'[^']*'|"[^"]*"|`[^`]*`|\[[^\]]*\])*$)
```

Ok with this boy we can support
- single-quotes
- double-quotes
- backticks
- arrays
- js-variables (incl. boolean/null)
- even all mixed together in a single line

nuernbergerA:
To ignore escaped single-quotes
```regex
'[^'\\]*(?:\\.[^'\\]*)*'
```

We end up with this monster
```regex
/ ?, ?                             #match comma optional surrounded by a space
(?=
    (?:
        [^"'`\[\]]|
        '[^'\\]*(?:\\.[^'\\]*)*'|  #single-quotes
        "[^"\\]*(?:\\.[^"\\]*)*"|  #double-quotes
        `[^`\\]*(?:\\.[^`\\]*)*`|  #backticks
        \[[^\]]*\]                 #array
    )*
$)/
```

calebporzio:
Yeah gross, I'm cool with it though. Regex is hard for me to read when it's short anyway, so it doesn't seem like a big deal to me (especially with a well-named variable or comment).

I'd rather support ""

calebporzio:
Love this. Thanks so much @nuernbergerA - great PR!

nuernbergerA:
closes #587 and #517

PR #596, opened 2020-02-06
added jest tests to github actions

### LINKED ISSUE #593: Add JS tests to github actions
On PRs we only check PHP test.
It would be nice to check if all jest test also pass.

We could use [stefanoeb/jest-action](https://github.com/stefanoeb/jest-action)

nuernbergerA:
closes #593


calebporzio:
Thanks @nuernbergerA !!!

PR #614, opened 2020-02-14
[JS] checkbox value to true/null, fix validation error

### LINKED ISSUE #611: Validation stops after toggling checkbox
**Describe the bug**
I have a checkbox that needs to be checked when a textarea is filled. I use validation to give a warning when that is not the case. The first time it works, but when I then uncheck and check my checkbox the "required" warning is not shown anymore.

**To Reproduce**
You can use the code snippets below to reproduce my issue:
```php
class Required extends Component
{
    public $checkbox, $textarea;

    public function render()
    {
        return view('livewire.required');
    }

    {
        $this->validateOnly($field, $this->validations());
    }

    private function validations()
    {
        return [
            'textarea' => 'digits_between:3,6',
            'checkbox' => 'required_with:textarea',
        ];
    }

    public function submit()
    {
        $this->validate($this->validations());
        //
    }
}
```
```html
<div>

        <textarea wire:model="textarea"></textarea>
        @error('textarea')Some wrong textarea input!!!@enderror

        <input type="checkbox"wire:model="checkbox" value="1">
        @error('checkbox')Checkbox is required!!!@enderror

        <button>submit</button>
    </form>
</div>
```
**Expected behavior**
I expect the warning to be shown always when the textarea is filled and the checkbox is not checked. Even if I toggle the checkbox.

**Additional context**
I haven't checked this issue with other types of validation types.

cloudstudio:
Fix #611 

When you check/uncheck a checkbox Laravel validation stop working, Its because for Larvel "false" is a value, so we need to change it to null again.

If you think is a bad solution, be free to change it :)

You welcome!

### DIFF
--- a/js/dom/dom_element.js
+++ b/js/dom/dom_element.js
@@ -229,7 +229,7 @@ ${this.el.outerHTML}
                 return modelValue
             }
 
-            return this.el.checked
+            return this.el.checked ? true : null;
         } else if (this.el.tagName === 'SELECT' && this.el.multiple) {
             return this.getSelectValues()
         }


calebporzio:
I feel like there are other implications here. I think it's better to set the initial state of the property to `false` yeah?:

```
public $terms = false;
```

This way, everything's predictable. I'd rather push people to set the intial state of the props to something meaningful.

If you disagree and have good reason, re-open.

Thanks for contributing!

cloudstudio:
Hey, @calebporzio  I think this doesn't fix the problem :(

You can try so easily yourself.

The problem is when you check and uncheck if you send the form validation pass, without checkbox pressed.

Livewire ignores the values, just send true/false.

If you can take a look :)

Livewire View
```
<div>
    <form class="form" wire:submit.prevent="submit">
        <input wire:model="terms" type="checkbox">
        <button type="submit" class="send">Send</button>
    </form>
</div>
```
Livewire Class
````
public $terms = false; // or null or whatever

    public function submit()
    {
        $this->validate([
            'terms' => 'required'
        ]);
        
        //
    }
```

CountZeroOne:
Try with 'accepted'
public $terms = null;
public function submit()
    {
        $this->validate([
            'terms' => 'accepted'
        ]);
        
        //
    }

Issue #371, opened 2019-11-16
Bug into select with values like 01


yashiroiori:
**Describe the bug**
When a select use this values:
"01":"01 - Efectivo",
"02":"02 - Cheque nominativo",
"03":"03 - Transferencia electrónica de fondos",
"04":"04 - Tarjeta de crédito",
"05":"05 - Monedero electrónico",
"06":"06 - Dinero electrónico",
"08":"08 - Vales de despensa",
"12":"12 - Dación en pago",

![image](https://user-images.githubusercontent.com/17031855/68999795-6f5c0780-088b-11ea-8eed-605d08f9cc03.png)

On choose value into select say:
![image](https://user-images.githubusercontent.com/17031855/68999760-f78ddd00-088a-11ea-94cb-df0ae7fea053.png)

But if use values without "0" into value:
"1":"01 - Efectivo",
"2":"02 - Cheque nominativo",
"3":"03 - Transferencia electrónica de fondos",
"4":"04 - Tarjeta de crédito",
"5":"05 - Monedero electrónico",
"6":"06 - Dinero electrónico",
"8":"08 - Vales de despensa",
"12":"12 - Dación en pago",

All work fine not error:
![image](https://user-images.githubusercontent.com/17031855/68999780-4b98c180-088b-11ea-87d7-1f11f82e9ac6.png)

lancepioch:
I think this is related to #283 and has to do with you using accented characters (non ascii).

calebporzio:
This issue needs more source code for reference and verification that this is still an issue. Closing for now.

emanuelErazo1:
the solution is check in backend if sleec firts is null or empty, in result is true so set value var $this->var = value in front select option

Issue #354, opened 2019-11-08
Livewire encountered corrupt data when trying to hydrate the component issue.


roni-estein:
**Describe the bug**
Short: I have an array of tasks, just a 2d array with text in keys, that livewire is loosing state on.

Longer: This is a volunteer school board project that I'm going VERY YANGI on, in the hopes it takes just a few days of spare time. We have a bunch of shared tasks, that I'm putting online for the right people to see.

I made a simple checklist component and added it to the site, I haven't even put items in a DB at this point, (perhaps that could help, but I decided to go all the way for the YAGNI)

Here is the component model, I"m just putting the whole thing in as is so it's easy to test:

```
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class Checklist extends Component
{
    public $flash = '';
    public $question = '';
    public $tasks = [
        'setup_the_system'=>[
            'title_note' => 'Most of this section is stuff that you won\'t readily be able to see,
            for example programming this checklist and the ability to ask questions here. However,
            it\'s all very helpful in moving this forward in the easiest way possible to solve problems
            in the easiest ways we can think of.',
            'create_visual_checklist_format' => 1,
            'create_version_repository'=>1,
            'create_version_repository_note'=> 'This basically will show all versions of the project, it\'s a <em>DR</em>
            (Disaster recovery) mechanism, in case I kick the bucket, another person can take over with a list of everything
            that\'s ever been done and things that need to get done in order to move forward. I will also endeavour to
            find someone to take over for me in case that should occur.',
            'kick_bucket_replacement_volunteer_found' => 0,
            'put_up_temporary_site'=>1,
            'put_up_temporary_site_note'=>'you are on it, the data on here will constantly be destroyed and re-written
            as we get ready to make something real! As real as this feels, don\'t save anything here that is important.
            Instead email me or keep a word doc, or basically anything else other than saving real data here.',
        ],
        'PAC USE CASES (A CHECK MEANS IT\'S AGREED UPON)' => [
            'Gatekeeper authorizes guardians who register'=>1,
            'Gatekeeper authorizes guardians who register_note'=> 'The <b>"Gatekeeper"</b> is a PAC member who has access
             to information and knows which families are attending the school and which ones are not.<br><br>
             Anyone who registers need to go through a 2 step process to have access to <b><em>any</em></b> information.
             Step 1 is simply verifying their email address which they can do without any intervention from the gatekeeper.
             Step 2 requires a person to actually allow the verified guardian access to see our internal information.
[...]

Issue #283, opened 2019-09-20
CorruptComponentPayloadException with carriage returns in data


aztekrw:
**Describe the bug**
Initial page loads fine. On any wire:model sending an update if any of the public fields has carriage returns in the data I get the error.

**To Reproduce**
eg:

data: 
address: "XXX XXX XX Avenue
↵Some End 
↵City

**Expected behavior**
Should handle these new lines in the data

**Desktop (please complete the following information):**
 - Browser  Google Chrome
Version 76.0.3809.132 (Official Build) (64-bit)

golam-sorwar:
<img width="1440" alt="Screenshot 2019-09-20 at 8 45 34 PM" src="https://user-images.githubusercontent.com/31072206/65336339-0425ea00-dbe8-11e9-95f2-9f0f632afbdf.png">
 I think same error

aztekrw:
I trimmed the data, but left the new lines, and it worked.

$this->address = trim($location->address);

golam-sorwar:
@aztekrw Thanks, it's working now

lancepioch:
Because of the carriage returns, it seems like this is a Windows specific issue. However, carriage returns alone shouldn't break like this.

I think we should have a notice/warning for Windows and then try to get this resolved!

calebporzio:
So what's the solution for this besides a warning? Is it a consequence of using json_encode? or the browser JSON.parse and re-stringifying thing? Or the fault of a middleware? Ideas?

lancepioch:
I tested this on the latest version both manually inserting carriage returns via alt+013 and %0d (urldecode) and also with <textarea> on Windows. There are no exceptions and Livewire handles it correctly.

Note: 0xd is carriage return and 0xa is regular new line

\* If you change the textarea after page load, there are still no exceptions, HOWEVER, the carriage return is converted to a new line character. This is rather odd behavior, but I believe all browsers ignore carriage returns from user input by default. So I think this behavior fits with all current browsers.

Result:
![image](https://user-images.githubusercontent.com/1296882/76268526-b8720600-6244-11ea-9e03-b9f5f7e7959a.png)

Component:
```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class Abc extends Component
{
    public $text;

    public function mount()
    {
        $this->text = urldecode("tes%0dting%20this+ou%0at");
    }
}
```

View:
```html
<div>
    <div style="margin-left: 50px;">
[...]

PR #630, opened 2020-02-22
Fixing unordered collection cause CorruptComponentPayloadException

Lastwisher:
Coused by JSON.parse auto sorting

This pr fix issues of #371  #354 #283 
No it does not.
Nope, this is one line fix that will not effect any behavior except fixing issue.

Ok I think this will be a long text to explain.

Issue was when you start your component with unordered collection at livewire, on any event you were getting CorruptComponentPayloadException .

Here is how you can reproduce bug ;
app/Http/Livewire/TestComponent.php
```
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class TestComponent extends Component
{
    public $data;

    public function mount($data) {
        $this->data = $data;
    }

    // just for test any event
    public function test() {
        
    }

    public function render()
    {
        return view('livewire.test-component');
    }
}

```
resources/views/livewire/test-component.blade.php

```
<div>
    <button wire:click="test">test</button>
</div>
```
resources/views/test.blade.php

```
<!doctype html>
<head>
    <meta charset="utf-8">
    @livewireStyles
    <meta name="viewport" content="width=device-width, initial-scale=1.0">

</head>
<body>

    @livewire('test-component',[collect([2 => "test",1=>"test2"])])
    @livewireScripts
</body>
</html>
```
When you click button you will get CorruptComponentPayloadException 

This is mostly can happen at elequent models that not returns with ordered id lets say you have a table users and have a column named external_id,
User 
id | external_id
1  | 3
2  | 2

when you sended 
```
$usersByExternalId = App\User::all()->keyBy('external_id');
@livewire('test-component',[$usersByExternalId])

```
you will face with this bug

Couse of this problem actually pretty nasty. 

Livewire at first request sets a initial-data as json.
After this whenever livewire js sends a request to livewire php, php checks this sum againist data comes from js to check is data corrapted. 

Unfortunately on ui side Chrome browser thinks that whenever a JSON.parse request comes i have to return an ORDERED list, you can test via :

```
JSON.parse(JSON.stringify({2:"test",1:"test"}))
```

its returns 

```
{1: "test", 2: "test"}
```

actually expected return is 
```
{2: "test", 1: "test"}
```

So what happens 
initial-data sended by livewirephp is :
```{2: "test", 1: "test"} ```
and sum is created by ComponentChecksumManager is lets say "X"

so Ui takes this data via json decode and Whoolaa array order is changed, at next event sended to server data is 
{1: "test", 2: "test"} and sum of this data ComponentChecksumManager  this time "y"

after generated 2 differnt data when ComponentChecksumManager  ->check compares this two data sees diffrence and thinks that, data modified at ui (actually works as expected data modified by chrome) and throws CorruptComponentPayloadException 

So as solution there is few ways can follow, maybe work around at JSON.parse to create not sorted array or adding ids as string (very dirty)

My solution is just sort values when create hash (dont worry its not manipulate original data just uses to generate sum). With this way even if chrome sorts data and returns back sum wont change.

I just realize i write 100 lines to explain 1 line of code :) 

Thank you guys for this awesome tool, its such a game changer.

### DIFF
--- a/src/ComponentChecksumManager.php
+++ b/src/ComponentChecksumManager.php
@@ -2,13 +2,15 @@
 
 namespace Livewire;
 
+use Illuminate\Support\Arr;
+
 class ComponentChecksumManager
 {
     public function generate($name, $id, $data)
     {
         $hashKey = app('encrypter')->getKey();
 
-        $stringForHashing = $name.$id.json_encode($data);
+        $stringForHashing = $name.$id.json_encode(Arr::sortRecursive($data));
 
         return hash_hmac('sha256', $stringForHashing, $hashKey);
     }


calebporzio:
Take a look at this test file: `tests/PublicPropertiesAreCastToJavaScriptUsableTypesTest.php`

I've already tried to solve this problem.

I could see that the current solution doesn't account for nested collection objects inside plain arrays (you can't actually set a property as a collection unless you're using a caster.)

If you want, you COULD add a test for a specific case involving Eloquent models and key-by's.

Also, I'd rather fix this upstream at around `src/Component.php@reindexArrayWithNumericKeysOtherwiseJavaScriptWillMessWithTheOrder` I believe.

I also kinda think this is a bit of an edge case, as we already have a fix for the basic issue you demonstrated.

Anyhow, I'm closing this, for now, re-open if you can add a test and more clearly define a common scenario that is affected by this.

Thanks!

Lastwisher:
Thanks for review this merge, 

Surely, i can write a test case and solution at 
`src/Component.php@reindexArrayWithNumericKeysOtherwiseJavaScriptWillMessWithTheOrder`
 but this approach will effect developer experience badly, end of the day livewire wont able to handle unordered arrays. its will convert them to ordered arrays.

Im currently working on a project netflix kind of streaming. I have episodes and seasons. At my database colums are 
id, season_no
1, 3
2, 1

If i took that approach i wont be able to get this by byKey ofc i can always order by season no. But if this was not an sql data that i can sort (eg: api response etc..) i have to write alot of sorting and re-keying just to make livewire response accordingly.

My suggestion if you like i can get around JSON.parse and add some prefixes to numeric keys so JSON will tread as string and when we get that data i can trim prefix.
This is little bit hacky and dirty but end result will be same as expected. Not just for elequent models its will work with all kind of unordered arrays and classes.

Which way you would preffer ? 
1) Order unordered arrays before send so user wont able to work with unordered arrays
2) Convert keys to string so JSON.parse wont mess with order, user able to work with unodered arrays.

Choice is all yours, personally i would go 2. option.

Cheers

PR #642, opened 2020-02-27
Fixes JS so it doesn't break when minified in production

kduma:
With this PR, minified code works 😉


kduma:
There ie even a notice above that block of code 😉
https://github.com/kduma/livewire/blob/e4d220d50a1722a15f49044aaf3c64d8c91bbfb5/src/LivewireManager.php#L233-L234

it-can:
Have the same problem...

calebporzio:
Thanks!

calebporzio:
Great catch. Thanks!

PR #661, opened 2020-03-01
asset_url from .env

Kamleshpaul:
added APP_URL from .env


besrabasant:
I do not think we need to set this to default to **APP_URL**. The assets path is already relative to the application URL. What do you say @calebporzio

Kamleshpaul:
for development we use `/public` route  that's why we need may be other 
linux use will also face same issue that's why

calebporzio:
Thanks for the contribution, however, I think I'm going to leave it alone for now.

If others feel the need for this, we can re-evaluate.

PR #674, opened 2020-03-03
Poll on minute

aryehraber:

I personally needed this in a project today and feel it has a good use case (more on this below). Also felt it was a good first issue to get my feet wet in helping with Livewire, so hopefully it gets merged (no hard feelings if not 😄).

---

I removed a duplicate test, can revert this back if needed.

---

Yes, however I feel the test is a little weak. I'd like to test `pollHappenedTimes` but am not sure how to make that happen, the `mountAsRoot` `requestInterceptor` seems to only be fired once.

Quick note: since this test was more sensitive to precise timing than the other polling tests, I added a test util to use Jest's Fake Timers feature. Calling `jest.useFakeTimers()` inside the test without calling `jest.useRealTimers()`afterwards broke _other_ tests, so this little wrapper keeps things contained. 

I've also updated the other polling tests locally using this util (instead of `timeout`) so let me know if you'd like me to push those changes.

---

I'm currently working on a project that shows a business's opening hours with an icon and green/red colours depending on whether they're open or closed. This was of course the perfect use case for Livewire and its `wire:poll` directive (from experience doing time-checking client side is a recipe for disaster as many users have their computer's clock/timezone misconfigured).

I started with a `15s` poll but the perfectionist in me was dying on the inside for 14 seconds while demoing the feature. Reducing to `10s` was better but still not great. Shortening this time even further seemed like the wrong way to go since the backend will get hit way too often while I know _exactly_ when it should refresh... _on the next minute_.

This PR adds a new modifier to the `wire:poll` directive which on init calculates how many seconds are left of the current minute and does a `setTimeout` for that duration, the action is then called after which the regular `fireActionOnInterval` gets set. By default the `defaultDuration` of 2 seconds gets increased to 60 seconds, ensuring the action is called once a minute, _on_ the minute. An additional modifier can override this duration, same as the current `wire:poll` API (eg: `wire:poll.on-minute.30s`)

---

🙌


aryehraber:
Hmm, just noticed the Jest Tests didn't pass, they're working locally so not sure why this happened...

Screenshot 2020-03-03 at 15 17 28

aryehraber:
Tweaks to tests didn't help... Let me know how you'd like to proceed from here!

nuernbergerA:
the jest action uses the compiled version so you have to commit them

aryehraber:
Oh! Let me try that, thanks! 🤞

aryehraber:
Sweet, that did the trick 👍  Thanks @nuernbergerA!

calebporzio:
Thanks for the contribution @aryehraber, however, I don't think this is a good idea for core.

You can accomplish this yourself from a component with something like this psuedo code:

```html
<div>
  ... your component's html
</div>

@push('scripts')
// Some custom interval code for your "on-the-minute" behavior"
setInterval(() => {
  @this.call('$refresh');
})
@endpush
```

aryehraber:
No problem, thanks for the alternative solution 👍

aryehraber:
@calebporzio Just implemented using `@this.call` and works great, is this documented anywhere?  Only found `window.livewire.emit('postAdded')` in [Event docs](https://laravel-livewire.com/docs/events/).

PR #676, opened 2020-03-04
Added support to start poll after x amount of time fixes #675

### LINKED ISSUE #675: Poll start after specified time
Yes. I've a requirement to start polling on multiple component & they are logically dependent on each other. My logic only full-fills when polling start on each component after specified time.

I want polling to start after x time on specific component. I've planned to use javascript `setTimeout` function with default time as 0. With this solution developers can set poll-start on every component the way they wants to like:
`<div wire:poll.3s wire:poll-start.10s>...</div>`

No I did not found any

**Additional context**
NA

mhetreramesh:
Yes. I've a requirement to start polling on multiple component & they are logically dependent on each other. My logic only full-fills when polling start on each component after specified time.

No

Yes

I want polling to start after x time on specific component. I've planned to use javascript `setTimeout` function with default time as 0. With this solution developers can set poll-start on every component the way they wants to like:
`<div wire:poll.3s wire:poll-start.10s>...</div>`


calebporzio:
Rather than bloating the API, for extra polling customization, it is recommended that you do so manually inside your component using `@this.call('$refresh')` inside a script tag:

```html
// Your component's view
<div>
  ...
  <script>
    // Do one poll after 10ms
    setTimeout(() => {
      @this.call('refresh')
    }, 10)
  </script>
</div>
```

mhetreramesh:
@calebporzio  Sure I'll try with this approach, thanks!

parmonov98:
Actually, you can conjugate ms via a variable  from component
image

PR #680, opened 2020-03-06
[Tests] Both slashes are valid directory separators for Windows

lancepioch:
Windows Only

The code for copying files is working fine. However the tests strictly compare slashes, but in Windows, both "\\" and  "/" are valid directory separators. Therefore, this adds a test helper that replaces all directory separators with a normal forward slash for these comparisons.

The following testing failures are resolved from this change:
image


calebporzio:
Thanks @lancepioch !

PR #681, opened 2020-03-06
Fix string representation display of directories

lancepioch:
Windows Only

There are two line changes here.

The second line change is a bug fix. Without it, on Windows, it returns the entire file path including C:\users\lance\app\... instead of just the regular relative path.

The first line changes the final path to use "/" as the directory separator. This is correct because the rest of the path already is using it. This is a visual change though because Windows treats "/" and "\" the same for directory separators. However, this lines it up with what the output is for Linux and makes the tests pass.

The following testing errors are resolved from this change:
image

### DIFF
--- a/src/Commands/ComponentParser.php
+++ b/src/Commands/ComponentParser.php
@@ -41,12 +41,12 @@ class ComponentParser
         return $this->baseClassPath.collect()
             ->concat($this->directories)
             ->push($this->classFile())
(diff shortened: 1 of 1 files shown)


lancepioch:
FYI #680 and #681 together resolve all current testing errors and failures for Windows.

calebporzio:
Thanks @lancepioch !

PR #683, opened 2020-03-06
[PHP] support fqcn for SPA routes

### LINKED ISSUE #682: [PHP] Allow fqcn for SPA routes
The current way to register SPA Routes isn't IDE friendly.

Lets use the example from the [docs](https://laravel-livewire.com/docs/spa-mode/#route-model-binding)

Wouldn't it be so much easier to write this
```php
Route::livewire('/contact/{user}', ShowContact::class);

//instead of
Route::livewire('/contact/{user}', 'show-contact');
```

nuernbergerA:
closes #682


nuernbergerA:
@calebporzio any planes to merge this in?

calebporzio:
What do you think of this strategy? Rather than modifying the finder itself, just grab the name in the `::livewire()` router macro and keep all the rest of the code the same.

If you like it, and are willing to change your implementation to it, I'll pull it in. Thanks!

image

nuernbergerA:
that was my first approach, but i thought the finder class should be responsible.

a question comes to my mind here is, why do i have to create an instance of my component with a fake id to call ```getName()``` this could be a static method.

so i remove all my changes and apply your code, my test fails now because the finder cant resolve the component.

this is due to how components are registered in this test.

this leads to another discussion on why do components have to have an alias and why isnt the compoenent responsible to gernerate a default if its empty.

or it's just me not knowing all LOC

how would you like to proceed?

nuernbergerA:
if we move the logic a step down, it could work.

test is green but not sure whats the impact to the variable `$component` passed to the blade

```php
public function livewire()
    {
        return function ($uri, $component = null) {
            $component = $component ?: $uri;

            return $this->get($uri, function () use ($component) {
                //added check here
                $componentClass = $component instanceof Component ? $component : app('livewire')->getComponentClass($component);
                $reflected = new \ReflectionClass($componentClass);

                return app('view')->file(__DIR__.'/livewire-view.blade.php', [
                    'layout' => $this->current()->getAction('layout') ?? 'layouts.app',
                    'section' => $this->current()->getAction('section') ?? 'content',
                    'component' => $component,
                    'componentParameters' => $reflected->hasMethod('mount')
                        ? (new PretendClassMethodIsControllerMethod($reflected->getMethod('mount'), $this))->retrieveBindings()
                        : [],
                ]);
            });
        };
    }
```

nuernbergerA:
@calebporzio your thoughts?

src/LivewireManager.php
```diff
@@ -42,6 +42,10 @@ public function componentResolver($callback)
 
     public function getComponentClass($alias)
     {
+        if (class_exists($alias)) {
```
nuernbergerA: add check for `$alias instanceof Component`

calebporzio:
Thanks a bunch @nuernbergerA!

Issue #684, opened 2020-03-06
Translation of computed property attributes to property methods should behave like Eloquent custom accessors


ninthspace:
Currently, the conversion from a computed property accessor such as `$this->foo` only performs a `ucfirst` on the property name. See:

https://github.com/livewire/livewire/blob/639ee48304760903b832db852592becf0fa805d6/src/Component.php#L169

This means `$this->my_foo` translates to `getMy_fooProperty`, rather than the more expected `getMyFooProperty`, which is similar to how Eloquent would translate it if it was a custom accessor (as `getMyFooAttribute`)

calebporzio:
Agreed, should be a simple enough PR. Feel capable? Thanks!

ManojKiranA:
@calebporzio  Just a Simple question are you going to make a breaking change on this by removing existing behaviour or we need to support for Both 

https://github.com/laravel/framework/blob/7.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L453

calebporzio:
This will be addressed in the next major version. It would be a breaking change to fix right now (which sounds odd I know, but it really would break a lot of people's apps and I don't want to do that in this version.)

PR #690, opened 2020-03-09
Change naming convention of computed properties to that of Eloquent f…

ninthspace:
As per #684 we replace `ucfirst` with a studly property calculation.

This is a breaking change for properties with more than one word.

Given that Livewire is currently compatible with many Laravel versions, and built-in string helpers have changed between these versions, we compute the studly property explicitly rather than defer to helpers.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -166,7 +166,9 @@ abstract class Component
 
     public function __get($property)
     {
-        if (method_exists($this, $computedMethodName = 'get'.ucfirst($property).'Property')) {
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks for the contribution @ninthspace. 

I agree this should be added.

Unfortunately, this is a breaking change, so we will need to wait until we're ready for v2 to add this.

For now, I'm going to close this PR but will add a reference to it in the Roadmap (which you can find in the README). We will re-open and merge when that time comes. (simpler than maintaining a separate v1 branch for now).

Thanks again!

calebporzio:
Thanks @ninthspace - finally got this merged!

PR #693, opened 2020-03-09
Add route layout optional params parameter

calebporzio:
This will allow you to pass parameters into the layout file with Livewire's shortcut:
```php
Route::livewire('/vittorio', 'vittorio')->layout('layouts.app', [
    'title' => 'foo',
]);
```


rodrigore:
Hi, I tried this method (passing argument to the layout) but it don't work for me.

It doesn't fail but the argument are not display on the blade render.

PR #695, opened 2020-03-09
ALSO disable selects and textareas on form submit.

lucasromanojf:

I imagine that @calebporzio is busy lately and then I did not open a feature-request first since this is something simple for me to help.

No.

Yes, tests were updated to cover this PR.

Currently, only inputs and submit buttons are disabled on form submission. This PR just includes selects and textareas too.

Count on me!


philipnewcomer:
@lucasromanojf I don't know if you want to include this in your PR or not, but it's a related change: checkboxes should be set to disabled on submit, as setting them to readonly (as is the case currently) does nothing.

I was going to create a PR myself for selects and checkboxes but you beat me to it with selects.

lucasromanojf:
> @lucasromanojf I don't know if you want to include this in your PR or not, but it's a related change: checkboxes should be set to disabled on submit, as setting them to readonly (as is the case currently) does nothing.
> 
> I was going to create a PR myself for selects and checkboxes but you beat me to it with selects.

Hey @philipnewcomer, sure I can add this, but I will only be able to do this tomorrow, is that OK?

philipnewcomer:
@lucasromanojf certainly, there’s no rush on my end.

js/component/DisableForms.js
```diff
@@ -13,10 +13,10 @@ export default function () {
 
                 if (! el.el.contains(node)) return
 
-                if (node.tagName.toLowerCase() === 'button' && node.type === 'submit') {
+                if ((node.tagName.toLowerCase() === 'button' && node.type === 'submit') || node.tagName.toLowerCase() === 'select') {
```
mikemand: Does this prevent the value from being sent with the request? Or is the data already on the way before the handler is called? I know if a `select` is disabled when a form is submitted, the value doesn't get included.

lucasromanojf: I thought about the same thing before doing this PR, but in case of Livewire, the default behavior is the fields with wire:model and on wire:submit, all fields values will be already in their respective public properties.

mikemand: Good to know, thank you!

lucasromanojf:
@philipnewcomer just added checkboxes and radio buttons to disable list on form submit!

lucasromanojf:
@calebporzio I was thinking, instead on disabling some kinds of fields and setting others to readonly, why can't we just disable all of them? Is there a reason for setting inputs as readonly instead of disabled?

calebporzio:
@lucasromanojf, thanks for the PR, and thanks for including tests.

Yeah, I think I took that pattern from Phoenix LiveView's behavior?

I like what you have here for now though. Merging it. Thanks again.

lucasromanojf:
@calebporzio thanks too

PR #696, opened 2020-03-09
Fix #289 by reloading the page if it's a hash change

### LINKED ISSUE #289: Redirecting to the current page with anchors

lancepioch:
This change listens for hash changed events and then automatically refreshes the page if those occur. I wasn't quite sure of the best place for the window.onhashchange listener.

`window.livewireInitiatedRedirect` makes sure that only hash changed events from Livewire will reload the page, other js libraries won't cause a full page refresh because of this new change.


calebporzio:
Hey @lancepioch, thanks for the contribution.

I like the idea of changing the current behavior, it seems a little odd to call `return redirect()` and not actually redirect.

As far as the implementation, It feels a little wonky to me setting a global variable in one place in the codebase and referencing it in another.

I also feel like it's not ideal because if the order gets messed up for some reason, the `livewireInitiatedRedirect` thing will be stuck on true and might cause weird hash change bugs.

I'm open to a fix, but I think it will need to look a little different and ideally have tests with it.

Thoughts?

lancepioch:
> As far as the implementation, It feels a little wonky to me setting a global variable in one place in the codebase and referencing it in another.
> 
> I also feel like it's not ideal because if the order gets messed up for some reason, the livewireInitiatedRedirect thing will be stuck on true and might cause weird hash change bugs.

I thought about it and if we remove it, then every hashchange would completely refresh the page, which might still be okay. We could just make this a toggleable option too (whether to refresh or not) and remove the global variable.

I'm not sure if scoping the variable to the Component would make a difference.

Also I'm for the tests, but I would need help with them probably because I'm not used to JS testing, especially with redirects.

calebporzio:
Ok, let's hold off on this for now.

If more people encounter this behavior as a "problem" we can re-examine.

(For some people, they might actually WANT the hashchange behavior with Livewire redirects)

Thanks for being willing to contribute, closing for now.

PR #702, opened 2020-03-11
Updated default asset URL in configuration file

stevebauman:
When you have a Laravel application running in a sub-folder, an relative path will be generated to the root of the URL, and not the absolute URL for the Livewire script.

For example, I have the following URL (Laravel application inside of the `erm` folder):

http://reports.company.com/erm/

The `<livewire:scripts>` annotiation will generate:

```html
<script src="/vendor/livewire/livewire.js?id=c6f7d3682bc8d847a034" data-turbolinks-eval="false"></script>
```

Which 404's. With the proposed change using the `url()` helper, it is set to the proper sub-folder:

```html
<script src="http://reports.company.com/erm/vendor/livewire/livewire.js?id=c6f7d3682bc8d847a034" data-turbolinks-eval="false"></script>
```

This is simply a change to the default asset URL. This won't break and users current asset URLs set.

I believe this is needed, as it will help anyone running Livewire on a Laravel application in a sub-folder / sub-directory.

No, just one single change.

No tests are required for this.

This sets the default asset URL properly when you have a Laravel application using Livewire in a sub-directory / sub-folder. This default will simply allow a greater initial flexibility when publishing the config file - as many will do when using their application in a sub-directory.


calebporzio:
I dig, thanks @stevebauman !

lancepioch:
@all-contributors please add @stevebauman for maintenance

PR #703, opened 2020-03-11
Enhance IDE support

devmsh:
After watching your Laracon Online video, I really appreciate the testing workflow Livewire provide, this is a simple PR to make it easier for IDE like PHPStorm to autocomplete the testing related function just like official Laravel Facades do.

Sample from Laravel Queue Facades:
https://github.com/illuminate/support/blob/master/Facades/Queue.php

### DIFF
--- a/src/Livewire.php
+++ b/src/Livewire.php
@@ -4,6 +4,11 @@ namespace Livewire;
 
 use Illuminate\Support\Facades\Facade;
 
+/**
+ * @method static \Livewire\Testing\TestableLivewire test($name, $params = [])
+ *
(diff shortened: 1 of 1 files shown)


calebporzio:
Generally, I resist docblocks, but using them on Facades is really extremely helpful. Thanks a lot for this!

devmsh:
@calebporzio :D once I create a new file using PHP artisan command, the first thing I do is to delete the docblocks 😆 but I really like using them to overcome the PHP magic, so I supposed that using them in the Facade will make the developer life easier.

Thanks for your greate work 👍

Issue #185, opened 2019-08-23
Snake case naming convention


oliverbj:
**Describe the bug**
Consider below Livewire component and view:

```
php artisan make:livewire text_replace
```
This will create two files:

```
-> [TextReplace.php]
-> [text_replace.blade.php]
```

When trying to render the view file:
```
@livewire('text_replace')
```

It will result in an error:

> Unable to find component: [text_replace]

However, if I change the view name to follow the kebab naming convention like: `text-replace.blade.php`

Then the view can load just fine, without error.

**Expected behavior**
I am following the snake convention when naming my blade views and would like to follow the same convention for my Livewire views as well.

calebporzio:
Would love a PR for this - it shouldn't be too hard.

oliverbj:
Will take a look at this!

PR #705, opened 2020-03-12
automatically convert Snake case naming convention

ManojKiranA:
1️⃣ Is this something that is wanted/needed? #185

2️⃣ Does it contain multiple, unrelated changes? 
NO

YES


ManojKiranA:
Will add More tests and Make New PR

Issue #708, opened 2020-03-12
Component doesn't work when using tag-syntax + passing an empty string


code-distortion:
**Description**
Hi there.  I've found a situation which stops the Livewire component from initialising.

If I have a component that has a value passed to it, when the string is empty and I'm using the new tag-syntax, the component isn't applied and the original tag is left in the output.

**To Reproduce**
``` PHP
<?php
// routes/web.php
Route::get('/', function () {
    return view('index');
});
```

``` Blade
// resources/views/index.blade.php
<html>
<head>
    <livewire:styles/>
</head>
<body>

{{-- works (using tag-syntax - a value is passed) --}}
<div style="background-color: #ddd; padding: 5px; margin-bottom: 10px;">
    <livewire:do-something value="abc"/>
</div>

{{-- works (using the old syntax - an empty string is passed) --}}
<div style="background-color: #ddd; padding: 5px; margin-bottom: 10px;">
    @livewire('do-something', ['value' => ''])
</div>

{{-- works (using tag-syntax - an empty string is passed via a variable) --}}
@php( $emptyValue = '' )
<div style="background-color: #ddd; padding: 5px; margin-bottom: 10px;">
    <livewire:do-something :value="$emptyValue"/>
</div>

{{-- doesn't work (using tag-syntax - an empty string is passed) --}}
<div style="background-color: #ddd; padding: 5px; margin-bottom: 10px;">
    <livewire:do-something value=""/>
</div>

<livewire:scripts/>
</body>
</html>
```

``` PHP
<?php
// app/Http/Livewire/DoSomething.php
namespace App\Http\Livewire;

use Livewire\Component;

class DoSomething extends Component
{
    public $value = '';

    public function mount($value = '')
    {
        $this->value = $value;
    }

    public function render()
    {
        return '<div>The value is "{{ $this->value }}"</div>';
    }
}
```

**Current Output:**
![image](https://user-images.githubusercontent.com/56794290/76530816-64775380-64c8-11ea-9355-b1d511825d0f.png)
**Expected behavior**

> <livewire:do-something value=""/>

Should render as:

> <div>The value is ""</div>

**Expected Output:**
![image](https://user-images.githubusercontent.com/56794290/76530668-29752000-64c8-11ea-8769-1f9cb398fcc5.png)

PR #709, opened 2020-03-12
Allow an empty literal string value when using tag-syntax

code-distortion:
Yes, I have just created bug-issue #708.  It seems like a small and straight-forward change,  I hope that's ok.

No. The change is simply one character.

Yes. I have added a new test.

If I have a Livewire component that has a value passed to it, when the string is empty and I'm using the new tag-syntax, the component isn't applied and the original tag is left in the output.

<livewire:do-something value=""/>

This is tag is currently skipped by Livewire as the regex in LivewireTagCompiler doesn't pick it up when no characters exist within the attribute's quotes.

The syntax above looks valid so I think it should work accordingly.

### DIFF
--- a/src/LivewireTagCompiler.php
+++ b/src/LivewireTagCompiler.php
@@ -26,7 +26,7 @@ class LivewireTagCompiler extends ComponentTagCompiler
                         (
                             =
                             (?:
-                                \\\"[^\\\"]+\\\"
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks @code-distortion !

lancepioch:
@all-contributors please add @code-distortion for maintenance

PR #704, opened 2020-03-12
Fix undefined $errors issue in nested components

merthurturk:
1️⃣ Is this something that is wanted/needed? Did you create a feature-request issue first?

Yes, this is a fix related to #621, #646 and https://github.com/livewire/livewire/issues/621#issuecomment-597593824.

2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No

3️⃣ Does it include tests if possible? (Not a deal-breaker, just a nice-to-have)

Yes

4️⃣ Please include a thorough description of the feature/fix and reasons why it's useful.

PR #646 fixed flashing validation error messages to the session but `$errors` variable was only available to the root view only, it was undefined in nested components. This was causing issues with `@error` directive usage in nested components and throwing `Undefined variable: $errors` exceptions. In order to fix the issue, I've implemented exactly what  `Illuminate\View\Middleware\ShareErrorsFromSession` middleware does to share errors with all the views.

5️⃣ Thanks for contributing! 🙌

👍

PR #710, opened 2020-03-12
Fix Error Sharing

taylorotwell:
This PR fixes error sharing with Livewire components as well as nested Blade components within those Livewire components. It also addresses a few other related problems.

Validation errors should not be flashed to the session as previous PR attempted to do. This doesn't solve the root problem and also creates new problems of making validation errors eternally persistent on the screen even after refresh. IMO errors should not be accessed in this way - the Laravel documentation never demonstrates accessing errors in this fashion either. You should use the global `$errors` variable that available to you in your views (and which is also utilized by the `@error` Blade directive).

After sharing the Livewire component's errors globally during that render cycle we reset the shared error bag to its previous state to clean up after ourselves.

Added new test for this based on https://github.com/livewire/livewire/pull/704 and also an additional check to make sure the error bag is restored to its previous state.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -107,10 +107,18 @@ abstract class Component
             new \Exception('"render" method on ['.get_class($this).'] must return instance of ['.View::class.']'));
 
         $this->setErrorBag(
-            $errorBag = $errors ?: ($view->errors ?: $this->getErrorBag())
+            $errorBag = $errors ?: ($view->getData()['errors'] ?? $this->getErrorBag())
         );
 
+        $previouslySharedErrors = app('view')->getShared()['errors'] ?? new ViewErrorBag;
+
         $errors = (new ViewErrorBag)->put('default', $errorBag);
+
+        $errors->getBag('default')->merge(
+            $previouslySharedErrors->getBag('default')
+        );
+
+        app('view')->share('errors', $errors);
 
         $view->with([
             'errors' => $errors,
@@ -120,6 +128,8 @@ abstract class Component
         session()->flash('errors', $errors);
 
         $output = $view->render();
+
+        app('view')->share('errors', $previouslySharedErrors);
 
         Livewire::dispatch('view:render', $view);
 
--- a/src/Component.php
+++ b/src/Component.php
@@ -125,8 +125,6 @@ abstract class Component
             '_instance' => $this,
         ] + $this->getPublicPropertiesDefinedBySubClass());
 
-        session()->flash('errors', $errors);
-
         $output = $view->render();
 
         app('view')->share('errors', $previouslySharedErrors);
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks Taylor!

lucasromanojf:
@taylorotwell what I did was just a best effort that fixed the empty @error issue (at least partially). I think it worked because of the Laravel's ShareErrorsFromSession middleware, that gets errors that may be in session and shares with view (not sure). This is why I did this way, I mean, even if this is not "the right way", the framework predicts this kind of situation in this middleware, right? Anyway, thanks for fixing this issue the right way, obviously you know the right way of addressing this better than me.

taylorotwell:
@lucasromanojf  that middleware runs before Livewire.

lucasromanojf:
@taylorotwell I see... So how did flashing errors to session could make @error inside blade components work? Do you have any idea? Just trying to learn.

lucasromanojf:
After this PR (and removing flashing errors to session), @error directives stopped working again inside Blade components. @taylorotwell please can you help with this? I'm not sure what to do now.

lancepioch:
@all-contributors please add @taylorotwell for code

petericebear:
I came across an issue with error sharing with named ErrorBags. Then the errors are blank within the Component. Any advice how to implement that?

boaterdan:
As has been mentioned above, this fix appears to be problematic.  As cited above, the documentation clearly states validation errors are automatically flashed to the session, even in the current 9.x version.  So why @taylorotwell do you state that it's wrong to do so, and why did you remove functionality that made things work according to the documentation?

The remaining problem for me is that when you have a standard component nested inside a livewire component, you can't get errors out of the session in the component class.  Yes, the $errors variable is available and contains the expected errors by the time you get to the blade, but I want to see if there are errors in the (non-livewire) class  _before_ rendering the component blade. I'm using this technique on another project using standard Laravel without livewire.

Please advise on how to accomplish this now that livewire makes Laravel behave differently (and not in compliance with documentation).

lucasromanojf:
> As has been mentioned above, this fix appears to be problematic.  As cited above, the documentation clearly states validation errors are automatically flashed to the session, even in the current 9.x version.  So why @taylorotwell do you state that it's wrong to do so, and why did you remove functionality that made things work according to the documentation?
> 
> The remaining problem for me is that when you have a standard component nested inside a livewire component, you can't get errors out of the session in the component class.  Yes, the $errors variable is available and contains the expected errors by the time you get to the blade, but I want to see if there are errors in the (non-livewire) class  _before_ rendering the component blade. I'm using this technique on another project using standard Laravel without livewire.
> 
> Please advise on how to accomplish this now that livewire makes Laravel behave differently (and not in compliance with documentation).

As you could see, he is more interested on pointing someone else's mistake than helping providing a solution.

PR #718, opened 2020-03-15
Allow empty value (using single quotes) to be passed

code-distortion:

This change is almost the same as one of my previous PRs.  I wasn't thorough enough and only made the fix for double quotes (sorry about that).  This one is for single quotes.

No,  only one fix.

I noticed the test I wrote for my previous PR (for almost the same thing) was removed. So I haven't added any for this one.

This change allows Livewire components to be used when passing an empty string value with single quotes.

<livewire:do-something value=''>

### DIFF
--- a/src/LivewireTagCompiler.php
+++ b/src/LivewireTagCompiler.php
@@ -28,7 +28,7 @@ class LivewireTagCompiler extends ComponentTagCompiler
                             (?:
                                 \\\"[^\\\"]*\\\"
                                 |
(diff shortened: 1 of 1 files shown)


code-distortion:
Does anybody have an idea why the test is failing?  The same suite of tests completes successfully [here](https://github.com/code-distortion/livewire/actions/runs/55868713)

lancepioch:
@all-contributors please add @code-distortion for maintenance

PR #720, opened 2020-03-15
remove vue support

jgile:
I did not create a feature request.  

No

No

this removes vue support.  current docs say that vue is removed, but it isn’t.  this causes problems for people following the docs as they will install the vue plugin and (possibly) have unexpected results.


calebporzio:
Wow, I thought I already did that, good catch. Thansk!

Issue #599, opened 2020-02-08
Default query string 'page' applies to all paginators on page.


xxdalexx:
The changes in [this commit](https://github.com/livewire/livewire/commit/6beffd1d6708f35e64f90d835bdd7e08a9b0baa2) do not work well if there is more than one pageinator in a page.

I don't know if doing something like the regular paginator with `$results->setPageName($name)`, or the ability to disable it per component would be a better fix.

lancepioch:
I believe that this has been answered in #738

PR #738, opened 2020-03-17
Disable or Rename Pagination Query String.

xxdalexx:
#599 
No
No
### API

**Disable query string for pagination:**
_Add to component_
`protected $paginationQueryStringEnabled = false;`

**Rename query string:**
_Declare property to use_
`public $differentPage;`
_Tell Livewire to use it_
`protected $paginationQueryString = 'differentPage';`

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -8,19 +8,40 @@ trait WithPagination
 {
     public $page = 1;
 
+    public function isQueryStringEnabled()
+    {
+        if (property_exists($this, 'paginationQueryStringEnabled')) {
+            return $this->paginationQueryStringEnabled;
+        }
+        return true;
+    }
+
     public function getUpdatesQueryString()
     {
-        return array_merge(['page' => ['except' => 1]], $this->updatesQueryString);
+        if ($this->isQueryStringEnabled()) {
+            return array_merge([$this->paginationQueryString], $this->updatesQueryString);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


ManojKiranA:
how to handle multiple pagination in same component

xxdalexx:
Having two paginators within the same component, or having two versions of the same component on a page with different query strings?

ManojKiranA:
> Having two paginators within the same component, or having two versions of the same component on a page with different query strings?

Having Multiple Paginator Instance on Same Component  or 

Handling  Multiple Paginator Instance on Multiple Component rendered in Same Page

xxdalexx:
I don't think the first way is possible currently, I'd have to play with around with it to see what would be involved to make it work. But you are probably better off extracting into a different component if you need two paginators.

Two components with different query strings: Rename the query string in one of them.

Multiple versions of the same component: Declare all possible query string tied variables in your component, set one as the default, and then pass in your mount() method which one to use if it's not the default. If you override the default, also call `$this->initializeWithPagination();` to reset the pagination with the new query string.

calebporzio:
The functionality within `WithPagination` is purposely extracted to an optional trait so that people can extend the functionality to meet their own needs.

I would recommend maintaining your own version of the trait locally where you modify the property name to avoid conflicts yourself.

Hopefully, that makes sense, thanks for being willing to contribute!

PR #740, opened 2020-03-17
fix minify bug in LivewireManager.php + add option for minifying styl…

it-can:
- Fixed a bug in production when minifying.
- Added options in config to set minifying instead of using app.debug


calebporzio:
This has been fixed, thanks for the PR though.

PR #646, opened 2020-02-27
Flash Livewire validation errors to session like Laravel validations do

lucasromanojf:
1️⃣ Is this something that is wanted/needed? Did you create a feature-request issue first?

This is needed, for example, for using @error blade directives inside new Laravel 7 blade components. Please see issue #621 .

2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No

3️⃣ Does it include tests if possible? (Not a deal-breaker, just a nice-to-have)

Yes

4️⃣ Please include a thorough description of the feature/fix and reasons why it's useful.

Please see issue #621

5️⃣ Thanks for contributing! 🙌

Count on me!

PR #747, opened 2020-03-19
Flash errors to session again (@error and $errors are not working in nested blade components)

### LINKED ISSUE #621: Validation errors are not flashed to session
If a validation error happens in Livewire, the ViewErrorBag (or MessageBag?) is not flashed to session as Laravel validation normally does (check https://laravel.com/docs/master/validation#quick-displaying-the-validation-errors). This means that when validations occur inside Livewire components, we can't access "**session()->get('errors')**". 

This is a problem for new Laravel 7 components syntax. For example, when creating a custom input, I want to use the **@error** blade directive inside the component. For this to work, I need to set the error bag in a public $errors property, so it gets injected inside the component. Something like this:

```
class Input extends Component
{
    public string $name;
    public string $label;
    public string $type;
    public $errors;

    /**
     * Create a new component instance.
     *
     * @return void
     */
    public function __construct(string $name, string $label, string $type = 'text')
    {
        $this->name = $name;
        $this->label = $label;
        $this->type = $type;
        $this->errors = session()->get('errors') :? new ViewErrorBag(); //This does not work, since session()->get('errors') is null
    }

    /**
     * Get the view / contents that represent the component.
     *
     * @return \Illuminate\View\View|string
     */
    public function render()
    {
        return view('components.input');
    }
}
```

For now, we are doing this:

**Input.php (component class):**

```
class Input extends Component
{
    public string $name;
    public string $label;
    public string $type;
    public $errors;

    /**
     * Create a new component instance.
     *
     * @return void
     */
    public function __construct(string $name, string $label, $errors, string $type = 'text')
    {
        $this->name = $name;
        $this->label = $label;
        $this->type = $type;
        $this->errors = $errors;
    }

    /**
     * Get the view / contents that represent the component.
     *
     * @return \Illuminate\View\View|string
     */
    public function render()
    {
        return view('components.input');
    }
}
```

input.blade.php (view):

```
...
<x-input name="email" type="email" label="E-mail" :errors="$errors"/> // Passing $errors on every component isn't a good solution
...
```

I tested and I can confirm that just changing the following code inside **Livewire\Component@output** method solves this issue. I submitted the PR #622 which does this:

```
public function output($errors = null)
{
        ...
        ...
        $errors = (new ViewErrorBag)->put('default', $errorBag); //<- add this

        $view->with([
            'errors' => $errors, //<- change this
            '_instance' => $this,
        ] + $this->getPublicPropertiesDefinedBySubClass());

        session()->flash('errors', $errors); //<- add this
        ...
        ...
}
```

Thanks!!

lucasromanojf:

Flashing errors to session was removed in another PR and @error and $errors stopped working again inside nested blade components. Should fix #621.

No.

The tests made on the other PR should cover this too.

Flashing errors to session was removed in another PR and @error and $errors stopped working again inside nested blade components. This PR should fix #621. I tested before and after the change and I can confirm It was not working and started working after change.

Please, if this is not the best solution, please let's discuss another one that fixes this issue and is better then. I'm asking this because another PR reverted my PR #646 and the issues that were fixed started happening again.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -119,6 +119,8 @@ abstract class Component
             $previouslySharedErrors->getBag('default')
         );
 
+        session()->flash('errors', $errors);
+
         app('view')->share('errors', $errors);
         app('view')->share('_instance', $this);
 


lucasromanojf:
Trying to check what's wrong in my project since another developer was able to make it work. Closing for now.

PR #768, opened 2020-03-25
Update Redirector.php

### LINKED ISSUE #762: redirect()->away() does not work
**Describe the bug**

Documentation suggests that any Laravel redirection notation may be used, but the away() function does not work.

**To Reproduce**

1. Create a Livewire function that uses redirect()->away() instead of redirect()->to()

**Expected behavior**

A redirect mimicking Laravel's own redirect()->away() . (Essentially a redirect to the exact URL specified rather than what Laravel thinks the URL should be.)

Currently the call does nothing at all. No redirect, no error.

nielspeen:
Fixes #762

### DIFF
--- a/src/Redirector.php
+++ b/src/Redirector.php
@@ -14,6 +14,13 @@ class Redirector extends BaseRedirector
         return $this;
     }
 
+    public function away($path, $status = 302, $headers = [])
+    {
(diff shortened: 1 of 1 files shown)


lancepioch:
@all-contributors please add @nielspeen for code

Issue #707, opened 2020-03-12
Accessing properties before initialization (PHP 7.4)


gjm:
**Describe the bug**

Having uninitialized typed properties in components triggers `ErrorException` saying **typed property must not be accessed before initialization**.

**To Reproduce**

* Create a component with an uninitialized typed property:

```php
class Foo extends Livewire\Component
{
    public string $bar;
    
    public function mount(string $bar)
    {
        $this->bar = $bar;
    }

    public function render()
    {
        return view('components.foo');
    }
}
```

* Use it in a view
```html
<x-foo bar="baz" />
```

**Expected behavior**
Component should be rendered successfully but an `ErrorException` is thrown:

`Typed property Foo::$bar must not be accessed before initialization`

**Aditional information**

When the component is constructed it calls the `ensureIdPropertyIsntOverridden` method which, in turn, calls the `getPublicPropertiesDefinedBySubClass` method in the `InteractsWithProperties` trait. This method tries to access the property and that triggers the error.

basepack:
Can confirm this using the old `@livewire('foo', ['bar'=> 'baz']` syntax too.

tanthammar:
confirmed, this gives an error
```php
public string $title;
public function mount()
{
        $this->title = trans('crud.account.title');
}
```
setting a default value removes the error
```php
public string $title = '';
public function mount()
{
        $this->title = trans('crud.account.title');
}
```

yob-yob:
I don't know if e89b6a1 fixed this issue for **Binding Directly To Model Properties**
since I'm still getting an error on  `ComponentConcerns/HandlesActions.php:24`

this is how I implemented it

```
public Course $course;

    protected $rules = [
        'course.category' => 'required',
        'course.expiration' => 'required',
        'course.instructor' => 'required',
        'course.title' => 'required',
        'course.excerpt' => 'required',
        'course.description' => 'required',
        'course.price' => 'required',
    ];
```
How I used it
```
<x-slot name="form">
    <x-inputs.select-menu wire:model.defer="course.category"  />
    <x-inputs.select-menu wire:model.defer="course.expiration"/>
    <x-inputs.select-menu wire:model.defer="course.instructor"/>
    <x-inputs.textbox wire:model.defer="course.title" />
    <x-inputs.textarea wire:model.defer="course.excerpt" />
    <x-inputs.textarea wire:model.defer="course.description" />
    <x-inputs.textbox wire:model.defer="course.price" />
</x-slot>
```

[...]

Issue #600, opened 2020-02-09
[PHP 7.4] v0.7.3 broke typed public model properties in components


riesjart:
v0.7.3 includes a new [HydrateEloquentModelsAsPublicProperties.php](https://github.com/livewire/livewire/blob/26b36828c8e994a7748f5afe06a85f4d5278fec4/src/HydrationMiddleware/HydrateEloquentModelsAsPublicProperties.php).

It contains `$instance->$property = (array) $serializedModel;` on [L33](https://github.com/livewire/livewire/blob/26b36828c8e994a7748f5afe06a85f4d5278fec4/src/HydrationMiddleware/HydrateEloquentModelsAsPublicProperties.php#L33).

Trying to set the public model property as an array will throw an exception when it is typed (available as of PHP 7.4). For example:

```
class Form extends \Livewire\Component
{
    public \App\Models\User $user;
}
```

PR #785, opened 2020-03-30
Fixes 'Typed property must not be accessed before initialization'

iksaku:
- This is indeed something reported previously:
  - #600
  - #707 

- Nope, straight-forward fix.

- It is not possible (yet) to add tests, since typed properties are supported since PHP >=7.4, and this package only has official support for PHP ^7.1.3

- Commit message is pretty detailed, but will copy-paste anyways...
<details>
<summary>Click to expand commit message</summary>

```
Typed properties are a feature of PHP 7.4, which allows
properties to reference specific types, helping both during
development and runtime.

This new implementation has the need to provide an "initialized" state for
all typed properties, which breaks Livewire's implementation when accesing
properties in early component construction.

The 'ReflectionProperty' class has a new 'isInitialized' method, which
helps us to determine if the value of the property is accessible,
but is only available for PHP >= 7.4.

In order to prevent Livewire's property implementation from breaking
when developers want to use typed properties, we check for the existance
of the 'isInitialized' method in the 'ReflectionProperty' class.
If the method doesn't exist, it must mean that the running PHP version
does not support typed properties, so we just access the property value directly,
expecting a default null value if not initialized, or an actual value if already initialized.

If the method exists, then we return a value depending on the property state.
If the property has not been initialized, we just return a null value (like earlier PHP
versions), otherwise, we access the property value and return it.
```

</details>

- Thank you for this awesome package!


nuernbergerA:
He @iksaku nice PR!

>  Does it include tests if possible? (Not a deal-breaker, just a nice-to-have)
> 
> * It is not possible (yet) to add tests, since typed properties are supported since PHP >=7.4, and this package only has official support for PHP ^7.1.3

We test against [PHP 7.1, 7.2, 7.3 and 7.4](https://github.com/livewire/livewire/blob/master/.github/workflows/test.yml#L22)

To run tests only for the specific version you can compare version, which you don't need in this case.

```php
/** @test */
public function public_properties_are_initialized_properly()
{
    if (version_compare(PHP_VERSION, '7.4', '<')) {
        $this->markTestSkipped('not supported...');
    }

   //....
}
```

If you need any help writing these tests, I would be happy to help you out.

iksaku:
Hello @nuernbergerA!

I will look into adding the needed classes using version comparison for tests. Hope I can push them later today.

iksaku:
@nuernbergerA Pushed some tests, which look for initialization states for both non-typed and typed properties, now the thing is that PHP Compiler will complain about the typing in earlier PHP Versions.

The only way to prevent this is to eval class strings into code when needed, but I don't think its good to force this use for the tests.

Any other way around?

iksaku:
Bump on this...

\cc @calebporzio @nuernbergerA

nuernbergerA:
Hey @iksaku sorry for my long absence, I still have a lot todo.
The only option comes to my mind is something with eval or anonymous classes.

My intention wasn't to block this PR for so long.

I think @calebporzio is also fine as it is now.

Thanks for your patience.

iksaku:
Thanks for your response!

Actually, I tried using anonymous classes offline, but the parser still complains while trying to compile the test file on lower PHP versions... Maybe the only option left is eval... 

But _should_ we use that kind of hacky implementation?

nuernbergerA:
For me, eval in a test for this special edge case would be fine.

nuernbergerA (approved):

nuernbergerA:
@calebporzio bump

calebporzio:
Thanks for this @iksaku  and @nuernbergerA  for tending to this PR 👏🏻

PR #790, opened 2020-04-01
Remove typehints from assertSee methods

driesvints:


calebporzio:
Thanks @driesvints !!

Issue #672, opened 2020-03-03
Allow class_namespace to sit outside of \App


code-distortion:
Hi there. I originally wrote a solution to the problem outlined below, but before posting I found an alternative solution here https://github.com/livewire/livewire/issues/306 which lets me do what I was trying to do.  I thought I'd still post this in case the solution I proposed is useful to someone else.  I'm happy to discuss it further, create a PR, paste the code, and if it's not useful feel free to close.

**ORIGINAL POST...**

Currently the root namespace for Livewire component classes must sit within the \App namespace. It will be useful in some circumstances to allow them to sit elsewhere.

**Example:**

_Laravel Modules_ is a package ( https://nwidart.com/laravel-modules/v6/introduction ) that lets you segregate your Laravel code with different concerns into their own directory structures (Console, Http, Routes, Tests etc. effectively their own App directory). It creates a /Modules directory and puts each module inside that,  however this is outside of the \App namespace.

Changing the _livewire.class_namespace_ config setting to 'Modules' for this situation will break Livewire's conversion between path names and component class namespaces.

**Desired outcome:**

I would like for Livewire to allow its component classes to sit outside of the App directory.

**Thoughts:**

A drawback of this would be if a user happened to set _livewire.class_namespace_ to '' (an empty string),  the LivewireComponentsFinder->getClassNames() method would loop through the files in /node_modles and /vendor when looking for Livewire component classes (slow and unnecessary).

**Alternatives:**

Leaving Livewire's components classes grouped together within App, as can be done now would work but would not take advantage of what Laravel Modules is trying to achieve.

It might be possible to update Laravel Modules so it puts its modules within /App/Modules.  I haven't explored this option.

**Proposed changes:**

I have made updates to the following files to get this to work. The changes don't break any tests:

- **LivewireComponentsFinder->getClassNames()** and
- **ComponentParser::generatePathFromNamespace($namespace)**

I updated them to take composer.json's _autoload.psr-4_ values in to account (similar to what **Illuminate\Foundation\Application->getNamespace()** does).

ComponentParser doesn't take advantage of the manifest file (/bootstrap/cache/livewire-components.php) like LivewireComponentsFinder does which would save this step,  however they both previously called Illuminate\Foundation\Application->getNamespace() which would parse composer.json anyway. So this is no worse than before.

austenc:
Also just ran into this -- to tack on a little bit more, it might be advantageous to allow the user to specify multiple locations and namespaces -- so it would sort of cascade?

[...]

PR #791, opened 2020-04-02
Allow Livewire components to sit outside of \App namespace

code-distortion:

This is potentially wanted, it adds extra flexibility without affecting existing usage - I created #672 to propose the change.

No.

No, sorry.

#672 explains it in more detail however the synopsis is:

Livewire currently expects its components to exist within the \App namespace, the _livewire.class_namespace_ config must start with "App\\".  This PR relaxes this restriction and allows components to exist in other namespaces,  such as "Modules\Xyz".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,7 @@
     ],
     "require": {
         "php": "^7.1.3",
+        "ext-json": "*",
         "symfony/http-kernel": "^4.0|^5.0",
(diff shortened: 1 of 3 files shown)


ryangjchandler:
This can already be achieved using `Livewire::component()` inside of a service provider.

calebporzio:
Thanks for the PR, for now if your classes are outside the `App` directory, I recommend registering them with `Livewire::component()` like @ryangjchandler said.

Thanks for being willing to contribute.

PR #803, opened 2020-04-11
Fix bug for keydown debounced event

### LINKED ISSUE #729: Keydown debounce not firing
**Describe the bug**

As stated in title 
`<input wire:keydown.debounce.500ms="method">` not fire any event, while works correctly without debounce

**To Reproduce**
```
<div>
  {{ $counter }}

   <input wire:keydown.debounce.500ms="increment"  name="ssss" id="ssss">
</div>
```

```
class Foo extends Component
{
    public $counter = 0;
    public function render()
    {
        return view('livewire.foo');
    }

    public function increment(){
        $this->counter++;
    }
}
```

davideperfetto:
Fixes #729 

Just the fix for the issue reported

It test correctly for keydown, unfortunately did not know how to properly test for elapsed time after debounced request fired .

in the documentation at [Livewire, keydown modifiers](https://laravel-livewire.com/docs/actions#keydown-modifiers) 

> Livewire directives sometimes offer "modifiers" to add extra functionality to an event. Below are the available modifiers that can be used with any event. .... 
debounce.150ms | Adds an Xms debounce the handling of the action.

Unfortunately the debounce worked correctly in model binding but not for the keydown event

Thank you for this awesome package!

### DIFF
--- a/tests/js/event_actions.spec.js
+++ b/tests/js/event_actions.spec.js
@@ -297,6 +297,7 @@ test('action parameter can use double-quotes', async () => {
         expect(payload.actionQueue[0].payload.params).toEqual(['double-quotes are ugly', true])
     })
 })
+/* How can I correctly test for this?
 test('debounce keydown event', async () => {
     var payload
     mount('<input wire:keydown.debounce.1000ms="someMethod"></button>', i => payload = i)
@@ -310,6 +311,7 @@ test('debounce keydown event', async () => {
         expect(payload.actionQueue[0].payload.params).toEqual([])
     })
 })
+*/
 
 test('keydown event', async () => {
     var payload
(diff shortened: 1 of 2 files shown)


tests/js/event_actions.spec.js
```diff
@@ -297,3 +297,32 @@ test('action parameter can use double-quotes', async () => {
         expect(payload.actionQueue[0].payload.params).toEqual(['double-quotes are ugly', true])
     })
 })
+/* How can I correctly test for this?
```
nuernbergerA, line 300: maybe this is a good [starting point](https://stackoverflow.com/questions/52224447/jest-unit-test-for-a-debounce-function#answer-52226973)

calebporzio:
Thanks @DavideMoney! The implementation was a little broken so take a look at my changes after the merge and the test I added if you're interested. Thanks again!

lancepioch:
@all-contributors please add @DavideMoney for code

PR #809, opened 2020-04-12
refactor: remove lonely ? in query-string

mblarsen:

Feature was wanted by myself. So created the PR.

No, only a small change.

I updated the relevant test.

When using the feature `updatesQueryString` it would leave append `?` a single (lonely) sign if every parameter for whatever reason has been filtered away (by except or otherwise). 

So your URL would go from a clean `/` to `/?`.

This PR prevents adding an empty query set.


mblarsen:
That failing test is a bit of a mystery. What values after being stringified would be truthy and then show up as an empty string? (in the context of the unit test)

nuernbergerA:
@mblarsen please make sure you run `npm run build` and commit the `dist` files

mblarsen:
> @mblarsen please make sure you run `npm run build` and commit the `dist` files

Thanks for pointing that out @nuernbergerA - The tests are passing now.

calebporzio:
Love it, thanks!

lancepioch:
@all-contributors please add @mblarsen for maintenance

Issue #802, opened 2020-04-10
Optional route prefixing


alfiehd:
**Is your feature request related to a problem? Please describe.**
This request is in relation to a current problem I have. I want to be able to use Livewire to have realtime charts without the need for having to write *much* javascript, and in my test application this works great! The problem is that when I'm using Livewire in our live project which uses multi-tenancy where the domain prefix is the tenant identifier (and a bound parameter), the Livewire routes break because I cannot prefix them. An example route would be something like ```{tenant}.foo.bar/test```.

**Describe the solution you'd like**
The sort of solution I'd like to see is for there to be an option to include the Livewire routes in any of the ```routes/*.php``` files, wherever necessary. Possibly something like ```Livewire::routes();```.

PR #812, opened 2020-04-13
Support for domain prefixed multi-tenancy domains

alfiehopkin:
This is something that is wanted for my current scenario (needed for domain prefixed multi-tenancy to work). I have raised an issue (feature-request) but has been left uncommented for 3 days, the reason my proposed solution. :)

All of the changes here are relating to the same issue.

There are no tests included, if accepted as a PR tests can be made for this specific use-case?

As described in #802, the issue is around not being able to use this framework in areas of an application that use a domain structure that has any route model binding provided in the domain prefix.

Example: ```http(s)://{tenant}.myurl.com/livewire``` or any route model binding that is provided before the livewire routes.

This fix keeps the current functionality, whilst allowing for the user to optionally _use custom routes_, placing them where they feel needed within their routes.php files.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -72,5 +72,18 @@ return [
     */
 
     'manifest_path' => null,
+    
+    /*
+    |--------------------------------------------------------------------------
+    | Custom Routes
+    |--------------------------------------------------------------------------
+    |
+    | Setting this value to true allows for you to decide where to put the routes.
+    |
+    | Example: Wanting to use a prefix that uses Route Model Binding "http(s)://{tenant}.foo.com/{bar}/livewire..."
+    |
+    */
+
+    'use_custom_routes' => false,
 
 ];
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -80,7 +80,7 @@ return [
     |
     | Setting this value to true allows for you to decide where to put the routes.
     |
-    | Example: Wanting to use a prefix that uses Route Model Binding "http(s)://{tenant}.foo.com/{bar}/livewire..."
+    | Example: Wanting to use a prefix that uses Route Model Binding "http(s)://{tenant}.myurl.com/{bar}/livewire..."
     |
     */
 
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -5,6 +5,7 @@ namespace Livewire;
 use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Foundation\Application;
 use Illuminate\Support\Fluent;
+use Illuminate\Support\Facades\Route as RouteFacade;
 use Illuminate\Support\Str;
 use Livewire\Exceptions\ComponentNotFoundException;
 use Livewire\HydrationMiddleware\AddAttributesToRootTagOfHtml;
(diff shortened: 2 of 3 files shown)


calebporzio:
Going to close this for now as I don't think it's a common enough need.

I encourage you to just register your own routes that point to Livewire's internal controllers with whatever bindings you need.

(Shouldn't be a problem to leave the existing ones there.)

alfiehd:
@calebporzio I did originally try that, but couldn't get it to work. Found another workaround using middleware. There was an issue where obviously the Livewire routes don't have the route parameters bound, so I modified my ```tenant:set``` middleware to check for the subdomain via regex and resolve using that if the parameter could not be found in the request.

Love the work on Livewire and I'm glad I finally get to use it for this project! 😄

volkeransmann:
@calebporzio I am having problems declaring the livewire-routes manualy inside a route:group with subdomain parameter and several middlewares. Unless i comment the registerRoutes() function out in the LivewireServiceProvider, my custom routes never get called because the generic livewire ones superceed them. Would it be possible to make the registering of routes optional via the config? i think the tenant scenario is not so rare after all - and currently the default livewire routes do not fit in very well...
The use_custom_routes option in config as suggested bz @alfiehopkin would be perfect.
thanks for considering...

Hesesses:
I would like to see this implemented 👍 

@alfiehopkin could you show how your routes + middleware looks?

alfiehopkin:
@Hesesses sure. Here you go. There are much cleaner ways to do this, but this is how it would work.

Middleware handle method:
```php
...
// get company slug from request
$company_slug = $request->company;

if (! $company_slug) {
    // get slug from subdomain
    preg_match('/:\/{2}([\w|-]*)\.[\w|-]*\./', $request->url(), $url_params);
    $company_slug = Arr::get($url_params, 1);
}

// get company from slug
$tenant = Company::whereSlug($company_slug)->first() ?? abort(404);
...
```

RouteServiceProvider set up with custom domain method:
```php
private function getDomainName()
{
    $base_domain = config('app.base_domain');
        
    return "{company}.{$base_domain}";
}
```
used like this:
```php
protected function mapWebRoutes()
{
    Route::domain($this->getDomainName())
        ->middleware('web')
        ->namespace($this->namespace)
        ->group(base_path('routes/web.php'));
}
```

alfiehopkin:
@Hesesses one other thing I forgot to mention is that using route model binding for the tenant turned out to be a massive pain in this project. You have to use this in every controller method and route call, which gets rather messy. A better approach is to set the tenant in the container using the regex approach and ignore the RMB method all together. This gives you better access to the tenant model as well as not messing up your controllers/routes.

Hesesses:
Thank you @alfiehopkin for comments!

I have currently:
```
web.php:
Route::group(['domain' => '{subdomain}.{domain}.{tld}', 'middleware' => ['subdomain']], function () { ...

subdomain middleware handle:
...
URL::defaults(['subdomain' => $request->route('subdomain')]);
URL::defaults(['domain' => $request->route('domain')]);
URL::defaults(['tld' => $request->route('tld')]);
...


class CompaniesCreate extends Component
{
    public $routing = [];

    public function mount()
    {
        $this->routing = ['subdomain' => request()->route('subdomain'), 'domain' => request()->route('domain'), 'tld' => request()->route('tld')];
    }
    public function createCompany()
    {
        URL::defaults(['subdomain' => $this->routing['subdomain'],'domain' => $this->routing['domain'],'tld' => $this->routing['tld']]);

     // ...
    }
}
```

This seems to work, but needs to be done on every livewire component =/

@calebporzio would be awesome to have support for subdomains 🙏

PR #816, opened 2020-04-14
Make views publishable

dododedodonl:
Registers views for publication

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -195,7 +195,7 @@ class LivewireServiceProvider extends ServiceProvider
         ], ['livewire', 'livewire:config']);
         
         $this->publishesToGroups([
-            __DIR__..'/views' => resource_path('views/vendor/livewire'),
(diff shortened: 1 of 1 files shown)


calebporzio:
I'm going to hold off on this. If people want to override the pagination view they can do so by modifying the trait, or by passing there own view to `->links('their-view')`

Thanks for being willing to contribute though!

mblarsen:
@dododedodonl I was looking for this option too. I manually copied it from the package. The default view worked fine for me except for the classes; using Tailwind.

@calebporzio how do you propose modifying the trait - something like this?

```php
<?php

namespace App\Http\Livewire\Concerns;

use Livewire\WithPagination as LivewireWithPagination;

trait WithPagination
{
    use LivewireWithPagination {
        paginationView as _paginationView;
    }

    public function paginationView()
    {
        return 'partials.pagination';
    }
}
```

And use this "modified" trait instead? Or what did you have in mind?

PR #819, opened 2020-04-15
Fix error assertions on rules that contain underscores

### LINKED ISSUE #818: Unable to assert errors on rules that contain underscores
**Describe the bug**
Livewire is unable to assert that a component contains validation errors for rules with underscores in their names.

**To Reproduce**
This example uses a simple component that checks if `$value` passes the validation rule `alpha_dash`.
This rule checks that the field under validation may have alpha-numeric characters, as well as dashes and underscores.
To test this validation, I pass a Faker sentence to `$value` and validate it. I then assert that the component contains an `alpha_dash` validation error. Instead of the expected pass, I receive -
```
Component has no [alpha_dash] errors for [value] attribute. Failed asserting that an array contains 'alpha_dash'.
```
The same bug is present with `assertHasNoErrors()`.
I have managed to reproduce this unexpected behaviour with other validation rules that contain underscores such as `active_url`, `alpha_num`, `ends_with` and `starts_with`, but none so far with rules that don't contain underscores.

```
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class AlphaDash extends Component
{
    public $complete = false;

    public $value;

    public function submit()
    {
        $this->validate([
            'value' => ['alpha_dash'],
        ]);

        $this->complete = true;
    }
}
```

```
<?php

namespace Tests\Feature;

use App\Http\Livewire\AlphaDash;
use Illuminate\Foundation\Testing\WithFaker;
use Livewire\Livewire;
use Tests\TestCase;

class AlphaDashTest extends TestCase
{
    use WithFaker;

    /** @test */
    public function value_is_alpha_dash()
    {
        Livewire::test(AlphaDash::class)
            ->set('value', $this->faker->sentence)
            ->call('submit')
            ->assertHasErrors(['value' => 'alpha_dash']);
    }
}
```

**Expected behavior**
I expect Livewire to assert that an `alpha_dash` validation error is thrown.

danharrin:
Fixes #818.

The issue is with `$this->lastValidator->failed()` returning rule names in StudlyCase, which are then formatted in lower case before being asserted.

This results in a scenario whereby to test for an `alpha_dash` error, you'd need to check for `alphadash`.

This fix involves converting the StudlyCase rule names into snake_case, which also formats them in lower case as well, killing two birds with one stone.

You can check that this works using the example described in #818.

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -3,6 +3,7 @@
 namespace Livewire\Testing\Concerns;
 
 use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
 use Illuminate\Support\MessageBag;
 use Illuminate\Database\Eloquent\Model;
 use PHPUnit\Framework\Assert as PHPUnit;
@@ -26,7 +27,7 @@ trait MakesAssertions
     public function assertSee($value)
     {
         PHPUnit::assertStringContainsString(
-            e($value),
+            e((string) $value),
             $this->stripOutInitialData($this->payload['dom'])
         );
 
@@ -36,7 +37,7 @@ trait MakesAssertions
     public function assertDontSee($value)
     {
         PHPUnit::assertStringNotContainsString(
-            e($value),
+            e((string) $value),
             $this->stripOutInitialData($this->payload['dom'])
         );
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @DanHarrin!!

lancepioch:
@all-contributors please add @DanHarrin for code

PR #820, opened 2020-04-15
Remove unspecified config

### LINKED ISSUE #817: Use of unspecified config value
**Describe the bug**
When registering the views, the name is set to `config('livewire.view-path', 'livewire')` [on line 117](https://github.com/livewire/livewire/blob/f9f2cb4d2a587a8b5414858fb07e5d1fea589d0b/src/LivewireServiceProvider.php#L117) of the LivewireServiceProvider.

This is probably not intended, as the same config value is not used in the WithPagination trait [on line 31](https://github.com/livewire/livewire/blob/f9f2cb4d2a587a8b5414858fb07e5d1fea589d0b/src/WithPagination.php#L31), and not defined in the default config file. Using this config value would break the loading of this view.

**Solution**
Remove the config value when registering the views, and set it to the now used `livewire`.

dododedodonl:
Resolves  #817

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -114,7 +114,7 @@ class LivewireServiceProvider extends ServiceProvider
         // when a user applies the WithPagination trait to a component.
         $this->loadViewsFrom(
             __DIR__.DIRECTORY_SEPARATOR.'views',
-            config('livewire.view-path', 'livewire')
+            'livewire'
         );
     }
 


lancepioch:
The paginator should instead be changed to include the config value.

dododedodonl:
Why? The config entry is not specified in the config file, and the value is a namespace used in the blade engine so package views don’t name collision with user views.

iksaku:
I support this pr, after all, it removes a mistype that references an unexisting config entry. Only [`mount-component` and `pagination-links` views](https://github.com/livewire/livewire/tree/758a30ee08f9c2f4b8caf9ed11086c49e5b13036/src/views) are affected by this.

`mount-component` view is used for Tests, and is only [referred using with a relative path](https://github.com/livewire/livewire/blob/758a30ee08f9c2f4b8caf9ed11086c49e5b13036/src/Testing/TestableLivewire.php#L76-L81).

`pagination-links` view is [referred using Blade's view namespacing](https://github.com/livewire/livewire/blob/758a30ee08f9c2f4b8caf9ed11086c49e5b13036/src/WithPagination.php#L29-L32), which as @dododedodonl says, it will prevent collisions with other user-defined views. Official docs do not state that you're able to override the pagination view by placing a file in your livewire view folder. So I believe that is not an intended feature. But instead, by using Blade's view namespacing you could override it at `views/vendor/livewire`. I think we should stick to that, no modifications to `WithPagination` trait are needed at all.

calebporzio:
Agreed, good PR. Thanks!

lancepioch:
@all-contributors please add @dododedodonl for code

PR #829, opened 2020-04-17
Issue 649 - Testing - passing data for mount()

### LINKED ISSUE #649: Testing - passing data for mount() as of v1.0?
**Describe the bug**
When testing a component that expects a model in it's mount method until 0.7.4 we would do something like :
```php
Livewire::test(MyComponent::class, $user)
```
But now that gives `ErrorException: Only arrays and Traversables can be unpacked`.  So I tried making the syntax more like the v1 @livewire with :
```php
Livewire::test(MyComponent::class, ['user' => $user])
```
Which in turn gives `ErrorException: Cannot unpack array with string keys`.  I've tried variations of the syntax, calling `->toArray()` on the model etc, but can't seem to make it happy.

**To Reproduce**
^ As above

**Expected behavior**
To be able to pass a model (or array version of it) to the component.

i-bajrai:
Fixes https://github.com/livewire/livewire/issues/649

### DIFF
--- a/src/Exceptions/InvalidComponentMountStateException.php
+++ b/src/Exceptions/InvalidComponentMountStateException.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Livewire\Exceptions;
+
+class InvalidComponentMountStateException extends \Exception
+{
+    use BypassViewHandler;
+
+    public function __construct($component)
+    {
+        parent::__construct(
+            "Livewire encountered a missing mount method when trying to initialise the [{$component}] " .
+            "component. \n When passing non-numeric keyed arrays to components ensure you have a mount method " .
+            "on the component."
+        );
+    }
+}
--- a/src/Exceptions/MountMethodMissingException.php
+++ b/src/Exceptions/MountMethodMissingException.php
@@ -2,7 +2,7 @@
 
 namespace Livewire\Exceptions;
 
-class InvalidComponentMountStateException extends \Exception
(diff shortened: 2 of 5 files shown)


calebporzio:
Hey @i-bajrai, I like this, I just wonder if there is a simpler implementation.

I don't think we really have to use reflection, just a simple:

* If params were passed into the component
* If ! method_exists "mount"
* Throw an exception

(I'm not really worried about checking that mount is not declared in parent classes)

But actually, now that I'm thinking about it, I could see a case where you should be able to pass parameters into a Livewire component that the component doesn't use.

For example: if you had a dynamic component like this:

```
@livewire($dynamicName, ['foo' => 'bar'])
``` 

you may want `$dynamicName` to be set to SOME components that DO accept those params and others that DONT.

I'm curious how PHP behaves if you new up a class with no constructor but pass initial params to the constructor like: `new ClassWithNoConstructor('som-param')`.

Make sense?

Thoughts?

i-bajrai:
@calebporzio Thanks for the feedback! This was more of a test the waters with an idea kind of PR so would like to evolve into a better solution.

The reason I used reflection (shudder) was to check that the component being instantiated had the mount method and not the parent. Double-checked with method exists and still works so that's an improvement.

I will investigate the `$dynamicName` scenario, thanks!

calebporzio:
Perfect, thanks!

lancepioch:
@all-contributors please add @i-bajrai for code

Issue #836, opened 2020-04-20
Payload value isn't sent in Internet Explorer 11


chrisrhymes:
**Describe the bug**
When a model is updated the actionQueue[0].payload.value isn't sent when using IE11. The name of the model is set correctly but the value is missing from the payload object. 

For reference, I had to add a polyfill for `fetch` (using whatwg-fetch) and `prepend` from [MDN Webdocs](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/prepend#Polyfill) for Livewire to work with IE11. 

**To Reproduce**
1. Create a text input with a wire:model, such as 'title', in a livewire blade view. 
2. Type a value into the input
3. A 500 error is returned

`[2020-04-20 10:32:29] local.ERROR: Undefined index: value {"userId":xx,"exception":"[object] (ErrorException(code: 0): Undefined index: value at project-path/vendor/livewire/livewire/src/HydrationMiddleware/PerformPublicPropertyFromDataBindingUpdates.php:14)`
4. The payload sent does not contain `value`

```
"actionQueue": [
    {
        "type": "syncInput",
        "payload": {
            "name": "title"
        }
    }
]
```

**Expected behavior**
The value for title model should be sent in the request

**Screenshots**

**Desktop (please complete the following information):**
 - Browser Internet Explorer 11

**Additional context**
Looking through the JavaScript I can see it gets the value for each model in `dom_element.js` but I can't see any reason why it wouldn't get the value.

chrisrhymes:
It seems that this is caused by IE11 always returning true for `e instanceof CustomEvent` in 'js/node_initializer.js' line 62 so it tries to return `e.detail` which is undefined, instead of the value from the input.

```js
const value = e instanceof CustomEvent
                ? e.detail
                : el.valueFromInput(component)
```

lancepioch:
Thanks for the contribution!

PR #838, opened 2020-04-20
Check that e.detail is not undefined

chrisrhymes:
I have created an issue for this #836 as this is causing a bug in IE11

Single change for a single bug 

No

It seems IE11 always returns true when checking `e instanceof CustomEvent` leading to the e.detail to be returned when it is undefined. This checks that the e.detail is not undefined and falls back to using the valueFromInput instead. 


calebporzio:
Thanks @chrisrhymes !

lancepioch:
@all-contributors please add @chrisrhymes for code

PR #843, opened 2020-04-22
Don't wrap authorization exceptions in 500

danharrin:
This issue is present when authorization exceptions are thrown from a component class. Instead of a 403 being thrown and cleanly rendered, Livewire wraps it in a 500, which causes this eyesore...

ew

You can see this behaviour in action simply by using `$this->authorize()` on an unauthorized model.

I've written a test that demonstrates this bug and asserts that the change fixes it.


calebporzio:
This is perfect, thank you!

lancepioch:
@all-contributors please add @DanHarrin for code

PR #862, opened 2020-04-23
Stop polling when poll directive is removed from element

dennisvandalen:
**1️⃣ Is this something that is wanted/needed? Did you create a feature-request issue first?**
I wanted it, I expected the polling to stop when I deleted the `wire:poll` from my element. I also saw it [requested in the forums](https://forum.laravel-livewire.com/t/any-way-to-halt-polling/65/4).

**2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.**
Nope just one. 

**3️⃣ Does it include tests if possible? (Not a deal-breaker, just a nice-to-have)**
Yes. I added a test to check if the polling has stopped after the 

**4️⃣ Please include a thorough description of the feature/fix and reasons why it's useful.**
My use case is: I want to poll until a certain value is retrieved which is stored in the database from a queue, but this could take a while, thus we need to poll multiple times.

What I'd like to do is this:
```
<div @if(empty($posts))wire:poll @endif> ... </div>
```
By default `$posts` is an empty array. And in the `hydrate()` method I check if the data is already available. The component stops polling when `$posts` stops being an empty array. 

**5️⃣ Thanks for contributing! 🙌**
Thanks for making it possible 🍻


calebporzio:
Love it, thank you!

lancepioch:
@all-contributors please add @dennisvandalen for code

nkeena:
This doesn't seem to be working? Am I missing something?

I've pulled in the latest version of Livewire `v1.0.14` and added a blade conditional to remove the directive but polling is still happening.

```html
<section class="flex py-6 -mx-4" @unless($editMode)wire:poll.5000ms @endunless>
```

dennisvandalen:
> This doesn't seem to be working? Am I missing something?
> 
> I've pulled in the latest version of Livewire `v1.0.14` and added a blade conditional to remove the directive but polling is still happening.
> 
> ```
> <section class="flex py-6 -mx-4" @unless($editMode)wire:poll.5000ms @endunless>
> ```

Works for me after updating to `1.0.14`. Can you double check you're on the latest version?

Also if `wire:poll` is removed it should start throwing `TypeError`s but you mention it keeps polling. Maybe `$editMode` is still set to true?

nkeena:
I was getting a `TypeError` before I upgraded to `v1.0.14`. I can confirm I'm on the latest version. Here's what my `composer.lock` file has:

```json
"name": "livewire/livewire",
            "version": "v1.0.14",
            "source": {
                "type": "git",
                "url": "https://github.com/livewire/livewire.git",
                "reference": "a6b43843c89304fe96f9f98fcb3f1e2660b0deae"
            },
```

Also, I can confirm that `$editMode` is returning `true` from the livewire request in chrome dev tools:

```json
data: {selectedStudentId: null, class: {class: "App\ClassRoom", id: 6,…}, editMode: true}
```

Also, when inspect the element, the `wire:poll` directive is being removed, however the polling still continues to happen.

I'm also running the latest Laravel `v7.8.1`

intrepidws:
I'm also experiencing problems here. My use case might be a little different in that I'm initially polling, then removing, then adding polling back again, and so on and so on. The initial poll seems to work OK, as does the first removal - but then it won't add polling back again. The code does reflect that `wire:poll.2s` is present, but the polling doesn't actually occur.

I can confirm that I am on `v1.0.14`.

MrWeb:
Same here, very simple:

```
<div @if($chat_open == true) wire:poll="loadFeed" @endif>...</div>
```

When $chat_open is false the polling continues

dennisvandalen:
I can't reproduce this. 

```php
    public $polling = true;

    public function togglePolling()
    {
        $this->polling = !$this->polling;
    }
```
```php
<div @if($polling) wire:poll.1000ms="updatePosts" @endif >
    <button wire:click="togglePolling">toggle polling</button>
</div>
```

Seems to work on my end.

Can you add some more details or maybe supply a demo project where this doesn't work?

lancepioch:
@all-contributors please add @dennisvandalen for code

amitavroy:
Yes, it doesn't work as mentioned. I am on 1.0.14. And my code is:
@if($show) wire:poll.1000ms="showMessage" @endif

kevyworks:
I don't know if its obvious but, `wire:...` wont work if its not rendered with-in `<livewire:... />` directive. Just saying... The poll example works and so with the code example if you put it inside `hello-world.blade.php` from the hello-world tutorials.

PR #870, opened 2020-04-24
Adding dedicated controller for Route::livewire() to allow route caching

### LINKED ISSUE #746: Route Caching on SPA examples aren't cachable
**Describe the bug**
When utilizing a livewire route like:

    Route::livewire('/home', 'counter')
        ->layout('layouts.app', [
            'title' => 'foo'
        ]);

The route becomes uncachable from laravel's eyes and running route:cache result in an exception that the route isn't cacheable.

**To Reproduce**
In a laravel +livewire configuration, use a route as above, then run `php artisan route:cache`. 

**Expected behavior**
Routes will be cached correctly without error.

**Screenshots**

georgeboot:
Fixes #746

This adds a separate `LivewireSPAController` class in favor of the old closure route that wasn't route cachable.

### DIFF
--- a/src/Macros/LivewireSPAController.php
+++ b/src/Macros/LivewireSPAController.php
@@ -41,14 +41,11 @@ class LivewireSPAController
             ? (new PretendClassMethodIsControllerMethod($reflected->getMethod('mount'), $this->router))->retrieveBindings()
             : [];
 
-        return view()->file(
-            __DIR__ . '/livewire-view.blade.php',
-            [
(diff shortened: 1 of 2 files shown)


georgeboot:
Oh no :( It seems this trick only works on PHP >= 7.4.

I've now added a version check, so the up2date people can at least benefit from it.

Too hacky or fine like this?

calebporzio:
Yeah, thanks for being willing to contribute, but I'd rather not add this kind of conditional for a PHP version unless absolutely necessary. Thanks again.

Issue #853, opened 2020-04-22
Fast follow - dynamic constructor arguments


i-bajrai:
Original idea:

```
... now that I'm thinking about it, I could see a case where you should be able to pass parameters into a Livewire component that the component doesn't use.

For example: if you had a dynamic component like this:

@livewire($dynamicName, ['foo' => 'bar'])
you may want $dynamicName to be set to SOME components that DO accept those params and others that DONT.

I'm curious how PHP behaves if you new up a class with no constructor but pass initial params to the constructor like: new ClassWithNoConstructor('some-param').
```

Proposal to understand behavior here and potentially open a PR to support this idea.

i-bajrai:
@calebporzio I'm assuming `@livewire($dynamicName, ['foo' => 'bar'])` ultimately always calls the mounting method and not actually instantiating the component?

If overloaded params should be set to null the functionality is already correct :)

calebporzio:
Thanks for the PR. Yes, it calls "mount()"

PR #876, opened 2020-04-26
Testing behaviour of components with constructor overload

i-bajrai:
#853

This test determines what happens when you overload a component. I believe it aligns with what would be expected.

Unless and exception needs to be triggered to warn users?


lancepioch:
@all-contributors please add @i-bajrai for code

Issue #402, opened 2019-12-09
Loading state not working when using ref & target


mokhosh:
**Describe the bug**
Loading state works for all buttons when I don't use ref and target, and doesn't work for any of them when I do.

**To Reproduce**
```php
@foreach($products as $product)
<button
    class="btn btn-success"
    wire:click="addToCart({{ $product->id }})"
    wire:ref="product{{ $product->id }}"
>
    <i class="fas fa-spinner fa-spin" wire:loading wire:target="product{{ $product->id }}"></i>
    @lang('shop.add_to_cart')
</button>
@endforeach
```

**Expected behavior**
See spinner for the button you clicked on.

**Desktop (please complete the following information):**
 - Windows 10
 - Chrome latest version

adrianschubek:
`wire:ref` has been removed in [v0.5](https://github.com/livewire/livewire/releases/tag/v0.5.0)

mokhosh:
Oh thanks, I forgot I red that in the release notes.

So, is there any way I can set loading on a specific button in the loop?
If I use the action (`addToCart`) all buttons will show the loading spinner while any item is being added to the cart.

calebporzio:
Hey @mokhosh - that's a good call. Wondering what the best solution here is.

Thinking it'd be something like this:

A) `wire:target="addToCart"` would scope to all "addToCart" actions
B) `wire:target="addToCart({{ $product->id }})"` would scope to "addToCart" with the productId parameter.

What do you think?

mokhosh:
Although it seems weird to include parameters as target, and something tells me we're breaking SOLID or DRY or some other set of rules somehow, I think we need to do this.

Firstly, because for some reason it seems like the obvious solution, and it was the only solution I thought I could try after I remembered there are no `ref`s in 0.5.
And more importantly, this doesn't work now:
```php
<button
    wire:click="addToCart({{ $product->id }})"
    wire:loading.attr="disabled"
>
```
I guess because it's trying to target `addToCart` with parameters which isn't supported.

mokhosh:
What I did for now is I extracted items to a `ProductItem` component which holds its `productId` and doesn't need an attribute for its `addToCart` method.

But my biggest concern is efficiency.
I'm loading products in my main layout. And passing each product's Id to a component in a foreach loop.
Then in each component I store the Id and `find` the product from the database individually and every time it's needed.

mokhosh:
[...]

PR #881, opened 2020-04-26
Loading target with parameters

dsazup:
Hello, this PR is related to issue #402 and also a point in the roadmap:

`Scope loading targets to actions WITH parameters: <div wire:loading wire:target="updateTodo({{ $todo->id }})">`

I have run into this issue today. Imagine I have a list of payment plans. Each row has 'upgrade' button that you can click. Once you click it, I need to show a spinner on the button. This is currently not really possible, as the spinner would be displayed on all buttons. I read through the issues and also noticed the roadmap and decided to give it a go.

I was wondering if you would consider a slightly different, though easy to achieve syntax. 

Instead of parsing the` updateTodo('title', 1)`, I propose to have syntax such as `updateTodo:title:1`. This way we do not need to worry about combining the parameters and having exact same signature, as there could be a lot of issues, for example double or single quotes, spacing etc. 

In my PR you will notice if there are params it returns 2 actions, for example with the todo case it would return `updateTodo` and `updateTodo:title:1` so that both

`<div wire:target="updateTodo">` and `<div wire:target="updateTodo:title:1">` could work.

Do you see any issues with it?


basepack:
This is a great idea imo. Your PR has my +1 👍

jjjrmy:
need this! thanks 🙏

calebporzio:
Thanks for submitting this PR.

I agree, it would be great if we could add support for targeting specific method/param combinations.

I think in my ideal world, the syntax would be as follows:
```
<button wire:click="foo('bar', 'baz')">
<span wire:loading wire:target="foo('bar', 'baz')">
```

Rather than the proposed `foo:bar:baz`

To me, this would be a more intuitive syntax. I imagine this is the syntax a developer would "guess" would work. I'd rather not introduce a new syntax here if we can help it.

As far as retrieving the method and parameters form something like `wire:target="foo('bar', 'baz')"`, that part should actually be easy. There is a "directive.method" and "directive.params" accessors on the element's directive object to use the same regex we use for firing methods and params.

That being said, currently `wire:target` supports targeting multiple actions separated by a comma: `wire:target="foo, bar"`.

I imagine this would trip up the existing method/params regex.

So, my assessment of this feature is this:
* I agree Livewire should offer this capability
* I'd much rather implement the more intuitive `foo('bar', 'baz')` syntax over `foo:bar:baz`
* I think it's possible, it will just be even more difficult because the existing regexes will have to be updated, or new ones will need to be created.

I'm interested in others thoughts on my notes.

Thanks!

dsazup:
Thanks for the feedback. I tried to achieve that syntax at first, but just did not find an easy way to do it. Will need to familiarise with the codebase more. I might look into this again later, but if anyone wants to do it in the meantime, feel free.

jjjrmy:
could this also work with dirty state?

mokhosh:
@calebporzio I think supporting multiple actions with `wire:target="foo|bar"` is even better. Fixes the conflict too.

kentin84000:
This feature still works with the latest version of Livewire?

Issue #890, opened 2020-04-28
wire:offline cannot Add & Remove CSS class


jjjrmy:
This bug is related to #885 

By default `wire:offline` sets an inline style of `display: inline-block` to the element if you used by itself, as mentioned in the related issue. This is problematic when using Tailwind classes.

The solution would be to use `wire:offline.class` instead, but this will only add the included classes. So not possible to have the element hidden then shown when offline.

The best solution would be to combine `wire:offline.class` and `wire:offline.class.remove`, currently this does not work as only one can be used at a time.

```html
<div wire:offline.class="absolute block" wire:offline.class.remove="hidden" class="hidden">
    You Are Offline!
</div>
```

calebporzio:
This has been fixed. The following snippet works perfectly:

```html
    <div wire:loading.class="absolute block" wire:loading.class.remove="hidden" class="hidden">
        I am loading!
    </div>
```

PR #892, opened 2020-04-28
wire:offline cannot Add & Remove CSS class

jjjrmy:
Fixes issue #890

Previously you were unable to use `wire:offline.class` and `wire:offline.class.remove` together. This fixes that issue!

```html
<div wire:offline.class="absolute block" wire:offline.class.remove="hidden" class="hidden">
    You Are Offline!
</div>
```

### DIFF
--- a/js/component/OfflineStates.js
+++ b/js/component/OfflineStates.js
@@ -31,22 +31,28 @@ export default function () {
 }
 
 function toggleOffline(el, isOffline) {
-    const directive = el.directives.get('offline')
+    const directives = el.directives.directives.filter(obj => { return obj.type === 'offline' })
 
-    if (directive.modifiers.includes('class')) {
-        const classes = directive.value.split(' ')
-        if (directive.modifiers.includes('remove') !== isOffline) {
-            el.rawNode().classList.add(...classes)
-        } else {
-            el.rawNode().classList.remove(...classes)
+    directives.forEach(function(directive) {
+        if (directive.modifiers.includes('class') || directive.modifiers.includes('attr')) {
+            if (directive.modifiers.includes('class')) {
+                const classes = directive.value.split(' ')
+                if (directive.modifiers.includes('remove') !== isOffline) {
+                    el.rawNode().classList.add(...classes)
+                } else {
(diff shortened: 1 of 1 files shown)


jjjrmy:
This same solution can probably be applied to Loading and Dirty states

tomirons:
I haven't used `offline` so I'm not answering on that part. But the `loading` state directives work as expected.

This: `wire:loading.class.remove="bg-gray-900" wire:loading.class="bg-gray-500 cursor-not-allowed" wire:loading.attr="disabled"` works well for me.

tomirons:
@jjjrmy why would you want to toggle the position of the element? Shouldn't it already be in position? 🤔

jjjrmy:
> @jjjrmy why would you want to toggle the position of the element? Shouldn't it already be in position? 🤔
> 
> Example:
> 
> ```
> <span wire:offline.class.remove="hidden" class="absolute block hidden">
>     You Are Offline!
> </span>
> ```

Yes that example does work, but seem bad to have to apply all the classes you want to the element and remove them when offline/loading. This change makes it more predictable.

tomirons:
What is the case that you want to use both of those directives at the same time? I'm curious.

devcircus:
@tomirons your example above was the way it was designed to be used. I can't think of a situation in which you would need to use both.

oceanapplications:
This removes one more pitfall. Better for things to work in unforeseen use cases. Don't let failure of imagination be a reason to maintain status quo.

calebporzio:
Thanks for this PR @jjjrmy, I'm willing to add this to the API, but we'll need tests. Are you able to add tests for this?

Here's the file to add them to: `tests/js/offline_states.spec.js`

jjjrmy:
> Thanks for this PR @jjjrmy, I'm willing to add this to the API, but we'll need tests. Are you able to add tests for this?
> 
> Here's the file to add them to: `tests/js/offline_states.spec.js`

I will work on it, thanks!

calebporzio:
Closing this due to inactivity and lack of interest.

PR #893, opened 2020-04-29
Fix #814 Allow injecting dependencies at render() method

### LINKED ISSUE #814: Allow injecting dependencies at render time
Hi there. While adding Livewire to a project, we missed a way to have dependencies injected at render time.

I was wondering if it would be ok to make it possible to inject dependencies in the render method.

```php

{
  public string $search = '';

  public function render(MyRepository $repo)
  {
    return view('livewire.my-component', [
       'items' => $repo->getItems($this->search),
    ]);
  }
}
```

Right now we're relying on resolving the dependencies using the IoC container directly.

amaelftah:
this **PR** closes #814 and allows laravel container to **resolve render method** and inject any dependencies needed.

```php

class PostsComponent extends Component
{
       public function render(App\Post $post)
       {
             //some logic
       }
}

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -80,11 +80,6 @@ abstract class Component
         return $this->casts;
     }
 
-    public function render()
-    {
-        return view("livewire.{$this->getName()}");
-    }
-
     public function output($errors = null)
     {
         // In the service provider, we hijack Laravel's Blade engine
@@ -95,7 +90,7 @@ abstract class Component
         $engine = app('view.engine.resolver')->resolve('blade');
         $engine->startLivewireRendering($this);
 
-        $view = $this->render();
+        $view = method_exists($this, 'render') ? app()->call([$this,'render']) : view("livewire.{$this->getName()}");
 
         if (is_string($view) && Livewire::isLaravel7()) {
             $view = app('view')->make((new CreateBladeViewFromString)($view));
--- a/tests/ComponentCanResolveRenderDependencies.php
+++ b/tests/ComponentCanResolveRenderDependencies.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Tests;
+
+use Livewire\Component;
+use Livewire\LivewireManager;
+
+class ComponentCanResolveRenderDependencies extends TestCase
+{
+    /** @test */
+    public function public_name_property_is_set()
+    {
+        $component = app(LivewireManager::class)->test(DummyComponent::class);
+
+        $component->assertSee('bar');
+    }
+
+}
+
+class Dummy
+{
+    public $foo = 'bar';
+}
+
+class DummyComponent extends Component
+{
+    public function render(Dummy $dummy)
+    {
+        return view('dummy-view', ['dummy' => $dummy]);
+    }
+}
--- a/tests/views/dummy-view.blade.php
+++ b/tests/views/dummy-view.blade.php
@@ -1,0 +1,3 @@
+<div>
+    {{ $dummy->foo }}
+</div>

### CODE REVIEW

**Comment by mblarsen** [comment]
Cool. Just out of curiosity can you point me to the part that actually handles the injection?

**Comment by Te7a-Houdini** [comment]
@mblarsen this [line](https://github.com/livewire/livewire/pull/893/files#diff-c694a29ec41709ca1a5f5e8a29f6467eR93)

**Comment by mblarsen** [comment]
@Te7a-Houdini ah, so the `app()->call(...)` will enable injection?

**Comment by Te7a-Houdini** [comment]
@mblarsen yes it will . the laravel service container will resolve those dependencies from render method . cause we are using **app()->call()**

also check t[his gist it explains the power of laravel container ](https://gist.github.com/davejamesmiller/bd857d9b0ac895df7604dd2e63b23afe)

**Comment by tomirons** [comment]
Very nice addition, I like the fallback to the component's view. 👍

File: src/Component.php [hunk 1]
```diff
@@ -80,11 +80,6 @@ public function getCasts()
         return $this->casts;
     }
 
-    public function render()
-    {
-        return view("livewire.{$this->getName()}");
-    }
-
```
nuernbergerA, line 87: Just keep this as it is

src/Component.php
```diff
@@ -95,7 +90,7 @@ public function output($errors = null)
         $engine = app('view.engine.resolver')->resolve('blade');
         $engine->startLivewireRendering($this);
 
-        $view = $this->render();
+        $view = method_exists($this, 'render') ? app()->call([$this,'render']) : view("livewire.{$this->getName()}");
```
nuernbergerA, line 93: Keep it readable like so
```php
$view = app()->call([$this, 'render']);
```

nuernbergerA (commented):
Nice addition, maybe keep changes minimal and readable.

src/Component.php
Te7a-Houdini: @nuernbergerA if we kept render() as it is . the generated components must implement it and can't inject dependencies in it

src/Component.php
Te7a-Houdini: i wanted to do that . but something in the tests blowed up :thinking:

src/Component.php
nuernbergerA: I just mean to leave it here as default
I've pulled your pr and change the code like I suggested and had no errors

Te7a-Houdini: if we left the render() method in the parent this exception will happen if someone tried to inject dependency

```php
 class ExecuteQuery extends Component
{
    public function render(User $user)
    {
        dd($user);
    }
}
```


Screenshot from 2020-04-30 11-25-14

nuernbergerA: Yea you're right due to PHP limitations you would have to make any additional parameter optional
```php
class ExecuteQuery extends Component
{
    public function render(User $user = null)
    {
        dd($user);
    }
}
```

calebporzio:
I like the idea of being able to dependency inject the `render` method. HOWEVER, removing the base `->render()` would be a pretty big breaking change.

People rely on that method for rendering their components.

I'm going to put this on a new "Consider for V2" section of the docs so we can revisit when we're ready to work towards a V2 (which I am currently not right now).

Thanks for this. Closing for now.

Te7a-Houdini:
@calebporzio not sure if it's going to be a breaking change . for two reasons 

- if render() doesn't exist in child component we will fallback to the default behavior of guessing the view name

- second render() is resolved by the di container . so if someone is using render() at child component without any dependencies it will work 

so i think this implementation is backward compatible unless you have something else in your mind

@calebporzio

mblarsen:
> so if someone is using render() at child component without any dependencies it will work 

Yes in fact they wouldn’t have been able to use it with params before.

> so i think this implementation is backward compatible unless you have something else in your mind

I agree. Would like to hear how this could be backward incompatible. 

The only backwards compatibility is a meta issue. Your IDE won’t be able to see the render method. Not a big issue in my book.

devcircus:
Just a thought, but could we just give the constructor the ability to auto-resolve dependencies?

instead of ```return new $componentName``` in LivewireManager, we could resolve from the container, set the id, then run the two methods that are currently in the current constructor. Those two methods would need to be public to account for grandchildren+ components, but then we could just use the constructor and not worry about passing down the id to the parent. 

🤷🏽‍♂️

mblarsen:
IMHO the render method works much like an action on a controller. For many smaller components that still has some logic this is where your code lives until the component starts to grow.

Close proximity of the code makes it quick to read and injection gives you types and IDE goodness. 

To me adding injection to the constructor doesn’t help that much in that aspect over `resolve()` + doc-block `@var`. I would prefer the doc-block as using the constructor would require at min 4 extra lines to access the injected object somewhere else in your code. 

That said I think constructor, mount, and render should all be DI’ed alike. Or at least mount (as it is) and one of the other.

PR #894, opened 2020-04-29
Add deeply nested array support to validateOnly method

yateric:
Yes.

No.

Yes.

```
<input wire:model="products.0.quantity">
<input wire:model="products.1.quantity">
<input wire:model="products.2.quantity">

class Demo extends Component
{
    {
        $this->validateOnly($field, [
            // This rule won't be triggered
            'products.*.quantity' => 'required|integer',
        ]);
    }
}
```

Currently value update of these inputs won't trigger the `products.*.quantity` rules, because `validateOnly()` use `Arr::only($rules, $field)` to get the necessary rules.  In this case, `$field` is `products.0.quantity` so no rules will be returned.

In this commit, I used `Str::is()` for rules matching to fix this issue.

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -6,6 +6,7 @@ use Livewire\ObjectPrybar;
 use Illuminate\Support\Arr;
 use Illuminate\Support\MessageBag;
 use Illuminate\Support\Facades\Validator;
+use Illuminate\Support\Str;
 use Illuminate\Validation\ValidationException;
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Great fix, thank you!

calebporzio:
If there is still an issue open, can you link to it here so we can close it?

yateric:
> If there is still an issue open, can you link to it here so we can close it?

Thanks for the merge!

After a quick check, no related issue is open.

lancepioch:
@all-contributors please add @yateric for code

PR #902, opened 2020-05-01
Prevent wrapping of exceptions when in debug mode

mikepsinn:
Errors thrown while a view is rendering are caught by the Blade compiler and obscured in an "ErrorException".  This change enables you to see where these exceptions originated while in debug mode.

### DIFF
--- a/src/CompilerEngine.php
+++ b/src/CompilerEngine.php
@@ -29,6 +29,8 @@ if (Application::VERSION === '7.x-dev' || version_compare(Application::VERSION, 
                 || $e instanceof NotFoundHttpException
                 // Don't wrap "abort(500)".
                 || $e instanceof HttpException
+                // Don't wrap in debug mode
+                || config('app.debug')
                 // Don't wrap most Livewire exceptions.
                 || isset($uses[BypassViewHandler::class])
             ) {


calebporzio:
Will need more info on this PR. Can you clearly show the current behavior, why it's unintended and why this is an improvement?

Also, I'm hesitant to merge anything that has a conditional for something like app.debug - I want Livewire to behave the same way in production as it does locally (when possible) to avoid odd problems popping up in production that can't be reproduced locally.

I'm going to close this for now because of that, but am open to hearing more about the problem to explore other solutions.

Thanks.

mikepsinn:
Hi, @calebporzio !  Thanks for your great work here!  :D  

So here's what my Ignition page looks like with this commit.  I can easily see the call stack within my project triggering the exception and fix it quickly.
image

So here's what my Ignition page looks like without this commit.  It only shows the location of the exception and skips straight to the Livewire compiler engine without any context so it's more of an ordeal to debug. 

image

I know exactly what you mean about wanting local behavior to be identical to production to avoid problems popping up in production that can't be reproduced locally.

However, it seems like you may consider making an exception in this case for 2 reasons:

Number A.
This will only change the behaviour when the APP_DEBUG env is true and the app is already going to behave very differently from production.  If people need to simulate production, they can just set APP_DEBUG=0 and the behaviour will be identical to before. 

Number B.
This is only going to occur at the time of a fatal exception anyway and makes the cause of the exception much more transparent in development.  Thus, it saves dev time which can be used for fixing more bugs, which will result in fewer bugs getting to production. 

So it seems like the overall costs of the change are exceeded by the overall benefits IMHO. 

Thanks for your time! :D

PR #903, opened 2020-05-01
Add wire:append directive

amaelftah:
this **PR** adds **wire:append** directive

```html
<div>
   <h1> {{$foo}} <h2>
    
    <div wire:append> <!-- any changes to $bar will be happened to this div -->
       {{$bar}}
   </div>
</div>
```
two notes regarding testing

- first one if we tried to remove [this part of code](https://github.com/livewire/livewire/blob/master/js/component/index.js#L299-L306) that is responsible for **wire:ignore** and runner js test we will find it will pass

- for the above reason i felt that the **ignore.spec.js** isn't testing the actual behavior so i tried to that for testing **wire:append** . but it didn't work . hope someone could shade some light on that

```js
test("wire:append appends to the declared element", async () => {
    mount("<div wire:append>foo </div>");
    let root = window.livewire.find("123");

    root.handleMorph('<div wire:id="123"><div wire:append>bar </div></div>');
    await timeout(20);
    console.log(root.el.rawNode().innerHTML); //this outputed <div wire:append>bar </div> which is not the expected result

//expected result <div wire:append>bar foo </div>
});
```

### DIFF
--- a/js/component/index.js
+++ b/js/component/index.js
@@ -305,6 +305,10 @@ export default class Component {
                     }
                 }
 
+                if(fromEl.directives.has('append')){
+                    to.insertAdjacentHTML('afterbegin', from.innerHTML)
+                }
+
                 // Children will update themselves.
                 if (fromEl.isComponentRootEl() && fromEl.getAttribute('id') !== this.id) return false
 


calebporzio:
Yeah, that's a pretty interesting implementation. Unfortunately, I don't think it would handle well for components that had other areas of interactivity.

There will likely need to be an "keyed" system to track which things to append and which things to keep. Does that make sense?

Te7a-Houdini:
@calebporzio am not quite sure we will need a keyed system . if you have an example code in your mind that would be better so i can tackle it.

 here's one i was trying and the reactivity worked fine .

when i click on Click me or write any thing inside the input . the view will be rendered and it will only append a new <button> & <input> inside the first <article> tag . while the second one will remain the same since it's child element isn't changed

and the rest of template will not be affected by the append .

```html
<div >
    <div>
        @foreach ($users as $user)
        {{$user->name}}<br>
        @endforeach
    </div>
    {{$users->links()}}

    <article wire:append >
        @php
            $str = Str::random();
        @endphp
        <button wire:click="clicked">Click me</button>
        <input type="text" wire:model="search" value="{{$str}}">
        <div>
            {{$search}}
            {{$str}}
        </div>
    </article>

    <article wire:append >
      Ahmed
    </article>

</div>

```

calebporzio:
Closing this for now. Thanks for the work, but I'll be taking a different direction with the features. Thanks!

jayenne:
I appreciate this thread is closed but an append and even prepend feature would really solve the infinite-scroll efficiency/history problem and wondered is there was either another way to achieve it on a "solution" was on the cards?

*At present the only articles on infinite-scroll just increase the pagination number and thus require larger and large requests for and transfer of data. I concede that this approach is a 'load more' but falls far short of being and 'infinity scroll'.

Issue #518, opened 2020-01-11
Listener parameter is casted because of its name


cbaconnier:
**Describe the bug**
A listener with the parameter named `$date` will contain an instance of `Illuminate\Support\DateFactory` despite we are not sending any date object. 

Changing the parameter name to `$test` fix the problem.

**To Reproduce**
```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class Pagination extends Component
{
    public int $month;
    public int $year;

    public function mount()
    {
        $this->month = now()->month;
        $this->year = now()->year;
    }

    public function updated($name, $value)
    {
        $this->emit('paginate', "{$this->year}-{$this->month}-01");
    }

    public function render()
    {
        return view('livewire.pagination');
    }
}
```
```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class Account extends Component
{
    protected $listeners = [
        'paginate' => 'loadTransactions',
    ];

    public function loadTransactions($date)
    {
        // Expect "2020-1-01" 
        // but receive an instance of `Illuminate\Support\DateFactory` instead
        dd($date);
    }

    public function render()
    {
        return view('livewire.account');
    }

}
```

**Expected behavior**
The parameter `$date` should contain the emitted value.

**Desktop (please complete the following information):**
 - Tested on Firefox

**Additional context**
 - Tested with PHP 7.4
 - Tested with livewire `0.5.3` and `dev-master`
 - Using `$collection` returns the expected string "2020-1-01"

calebporzio:
Yeah, this is an odd one. I imagine this is something going wrong with the auto-dependancy-injecting system for methods. Would you be willing to dig and PR a fix? Thanks!

cbaconnier:
@calebporzio Sorry, I haven't had too much time to look into this theses days.
The line responsible of the bug is this one.
```php 
return app($parameter->name);
```
https://github.com/livewire/livewire/blob/v0.7.4/src/ComponentConcerns/HandlesActions.php#L103


It means any parameter with an accessor name will cause troubles; 

- date
- cache
- session
- view...

But to be honest, I don't really know what could be done.

calebporzio:
oof, yeah that's rough.

So the solution would be to prioritize provided params over injected ones yeah?

robinvdvleuten:
I have the same issue; when I have a `$date` parameter, an instance of `Illuminate\Support\DateFactory` gets injected.

robinvdvleuten:
[...]

PR #910, opened 2020-05-02
Fix #904 resolving laravel container aliases variables

### LINKED ISSUE #904: Some parameter names cause param value to be switched with container-resolved value.
**Describe the bug**
The parameter names that you choose for your actions can result in the value of the param being switched out with something from Laravel's container. 
This causes a LOT of param names to be unavailable in livewire actions.

**To Reproduce**
in template:
```html
<h1 wire:click="test('/path/to/some/file')>Click Me</h1>
```

in component class:
```php
public function test($path)
{
    dump($path);
}
```

**Expected behavior**
The passed path, ```/path/to/some/file``` is dumped.

**Actual result**
Behind the scenes, livewire returns ```resolve('path')``` as the value of the $path param.
I'm assuming any parameter name that can be resolved from Laravel's container will result in the same thing happening. I've only noticed it when using ```$path``` and ```$url```.

I haven't spent enough time in the Livewire codebase to understand exactly why the ```resolveActionParameters``` functions this way, so at this point I don't know how to change this in a way that doesn't break Livewire.

**Guilty code**
code from HandlesActions.php
```php
    protected function resolveActionParameters($method, $params)
    {
        return collect((new \ReflectionMethod($this, $method))->getParameters())->map(function ($parameter) use (&$params) {
            return rescue(function () use ($parameter) {
                if ($class = $parameter->getClass()) {
                    return app($class->name);
                }

                return app($parameter->name);
            }, function () use (&$params) {
                return array_shift($params);
            }, false);
        });
    }
```

The ```return app($parameter->name);``` returns ```path``` from the Laravel container instead of the value that was passed to your action.

amaelftah:
this PR fix #904 . for example when using [**any laravel container aliases variables as a dependency**](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Foundation/Application.php#L1200-L1240) it will get resolved by laravel container which is not the desired behavior .

```php
<button wire:click="customMethod('someRandomValue')">Click</button>

class FooComponent extends Component
{
   public function customMethod($path) //or $session or any aliased variable
   {
       dd($path); //print out laravel app path not 'someRandomValue'
   }

}

```

so this PR fixes the above issue

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -84,9 +84,7 @@ trait HandlesActions
                 throw_unless(method_exists($this, $method), new MethodNotFoundException($method, $this->getName()));
                 throw_unless($this->methodIsPublicAndNotDefinedOnBaseClass($method), new NonPublicComponentMethodCall($method));
 
-                $this->{$method}(
-                    ...$this->resolveActionParameters($method, $params)
-                );
+                app()->call([$this, $method], $this->resolveActionParameters($method, $params));
 
                 break;
         }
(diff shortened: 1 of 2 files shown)


devcircus:
Looks good.

I was genuinely curious why Livewire would allow auto-resolving of container aliases when using the alias as a parameter name. The behavior seemed strange, but since I'm not very familiar with the inner-workings of Livewire, I though maybe some important functionality depended on it.

👍

cbaconnier:
Would also close https://github.com/livewire/livewire/issues/518

calebporzio:
Thanks for the contribution, this has been fixed already and will be available in the next release.

PR #912, opened 2020-05-02
add assertSeeHtml

vardyb:
No

No

Yes

I've been using Livewire for a work project and find myself wanting to test that the correct SVG and other HTML tags are displayed to the user based on decisions made in my code.  However the current `assertSee()` method escapes HTML.  This PR adds `assertSeeHtml()` and `assertDontSeeHtml()` that allows tests that check that the correct HTML is presented or not.

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -54,6 +54,16 @@ trait MakesAssertions
         return $this;
     }
 
+    public function assertDontSeeHtml($value)
+    {
+        PHPUnit::assertStringNotContainsString(
+            $value,
+            $this->stripOutInitialData($this->payload['dom'])
+        );
+
+        return $this;
+    }
+
     protected function stripOutInitialData($subject)
     {
         return preg_replace('(wire:initial-data=\".+}")', '', $subject);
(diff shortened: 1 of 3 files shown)


nuernbergerA (approved):
Nice addition

roni-estein:
Yeah, I'd love this! I would be curious why someone wouldn't want it!

vardyb:
@nuernbergerA @roni-estein Thanks for your comments.  Much appreciated.  I was thinking maybe `assertHasHtml()` and `assertDontHaveHtml()` may be better names?  Technically you don't see the HTML.  What do folks think, I'm happy to rename?

nuernbergerA:
Maybe something with DOM or raw?

vardyb:
@nuernbergerA something like `assertDomHasHtml()` and `assertDomDontHaveHtml()`?

calebporzio:
I love it as-is. Thanks @vardyb !

vardyb:
@calebporzio No worries, and thanks for the merge.

lancepioch:
@all-contributors please add @vardyb for test

lalov:
Another way would bee to do as is Illuminate\Testing\TestResponse, where the second parameter can be used to disable escaping. That way it would be exactly the same api as the TestResponse.

`public function assertSee($value, $escape = true)`

PR #915, opened 2020-05-02
Adds forward slash nested component names to livewire:make command

### LINKED ISSUE #913: Artisan livewire:make livewire/with/slashes does not work
Running this should work:
```bash
php artisan livewire:make livewire/with/slashes
```
But it gives an error because the generated class name is based on periods.

This works:
```bash
php artisan livewire:make livewire.without.slashes
```
But both ways should work, because that's for everything else in Laravel (config, routes, views, etc.). 

I prefer to use slashes. When using slashes, I would also expect the render template to use slashes:

```php
public function render()
{
    return view('livewire/livewire/with/slashes');
}
```

MuzafferDede:
> 1️⃣ Is this something that is wanted/needed? Did you create a feature-request issue first?

Yes. Fixes #913

> 2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No.

> 3️⃣ Does it include tests if possible? (Not a deal-breaker, just a nice-to-have)

Yes.

> 4️⃣ Please include a thorough description of the feature/fix and reasons why it's useful.

`livewire:make` command will support `.` or `/` nested names when creating, which is a common thing in laravel ecosystem. 

Now we can use as : 

`php artisan livewire:make foo/bar/baz`
`@livewire('livewire/with/slashes')`
or 
`php artisan livewire:make foo.bar/baz`
`@livewire('livewire.with/slashes/and.dots')`

results will be the same `Foo/Bar/Baz`

### DIFF
--- a/src/Commands/ComponentParser.php
+++ b/src/Commands/ComponentParser.php
@@ -24,7 +24,7 @@ class ComponentParser
         $this->baseClassPath = rtrim($classPath, DIRECTORY_SEPARATOR).'/';
         $this->baseViewPath = rtrim($viewPath, DIRECTORY_SEPARATOR).'/';
 
-        $directories = preg_split('/[.]+/', $rawCommand);
+        $directories = preg_split('/[.\/]+/', $rawCommand);
 
 
         $camelCase = Str::camel(array_pop($directories));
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks!

lancepioch:
@all-contributors please add @muzafferdede for code

PR #924, opened 2020-05-05
[PHP] Dedicated test to cover current DI implementation

nuernbergerA:
Added a dedicated test to cover the current dependency injection implementation.


calebporzio:
FANTASTIC!! Thanks @nuernbergerA!

Issue #701, opened 2020-03-11
Method model binding


mokhosh:
This has been suggested before https://github.com/livewire/livewire/issues/295, but "held off to focus on stabilizing the core code and API".
Do you think now is a good time to bring this up?

```php
<button wire:click="remove(78)">

function remove(User $user){
     $user->delete();
}
```

calebporzio:
This is on the roadmap for the next major version. Thank you.

PR #934, opened 2020-05-06
Add route model binding to actions

lancepioch:
Allows you to pass the model's route key (value) into a component's public function and have it be auto resolved to your type hinted model's instance.

This is based on #910

I'm working on tests still.

Todo:

* [ ] Implement Tests
* [ ] Whether to base off current code or pull request above
* [ ] Consider security implications

Example Component Function:
```php
public function setUsername(User $user, $username)
{
    $user->username = $username;
    $user->save();
}
```

Example View Call:
```html
<a wire:click.prevent="setUsername(1, 'test')"></a>
```

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -93,10 +93,17 @@ trait HandlesActions
     protected function resolveActionParameters($method, $params)
     {
         return collect((new \ReflectionMethod($this, $method))->getParameters())->flatMap(function ($parameter) use (&$params) {
-            if(!$parameter->getClass()){
+            if (!$parameter->getClass()) {
                 return [$parameter->getName() => array_shift($params)];
             }
-        })->filter()->toArray();
+
+            $instance = rescue(function () use ($class) { return app($class->name); });
+            if ($instance instanceof UrlRoutable) {
+                if ($model = $instance->resolveRouteBinding(array_shift($params))) {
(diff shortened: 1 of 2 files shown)


DanHarrin:
Closes #701.

AbdelElrafa:
Is there any update on this?

mattharcourt:
@lancepioch - this PR could be expanded a bit to also solve #1067. 

If `ComponentConcerns\HandlesActions@resolveActionParameters()` is generalized to accept a `ReflectionMethod` instead of the method (to remove the dependency on `this`), it could be utilized in `Macros\RouterMacros@livewire()` in place of the call to `Macros\PretendClassMethodIsControllerMethod@retrieveBindings()`. Would have to make it a static method, or perhaps move it to a `ComponentMethodBinder` class so it could be used in both places.

I should've looked closer before submitting #1068 as this would be a better solution. I'm willing to close that PR and add to this if interested.

mattharcourt:
Regarding my previous comment: after digging deeper, route binding is fundamentally different from action binding. PR #1068 has been updated to pursue route binding. 

This PR should continue on it's own for action binding. 

There may be additional logic to pull from Laravel's [`ImplicitRouteBinding`](https://github.com/laravel/framework/blob/192356b8d519dc307bd453786551fc5c765c7cab/src/Illuminate/Routing/ImplicitRouteBinding.php#L21) if there's interest in implementing scoped binding of related models for actions.

PR #939, opened 2020-05-08
Add ->only() method to components

imliam:
This change adds a new `->only()` method to the base `Component` class that mimics the method of the same name in Laravel's own `Request` class.

This method takes an array of strings as an argument, the strings should each match the name of a property on the component. The method then returns an associative array the property names and values.

This is just a small convenience method that can be used for many things and I often find myself using in regular controllers, particularly when creating models, for example:

```php
User::create(request()->only(['name', 'email', 'password']));
```

Here's another example where such a method would help to clean up the code a little bit: https://github.com/laravel-frontend-presets/tall/blob/master/stubs/auth/app/Http/Livewire/Auth/Passwords/Reset.php#L41-L45

### DIFF
--- a/src/ComponentConcerns/InteractsWithProperties.php
+++ b/src/ComponentConcerns/InteractsWithProperties.php
@@ -126,4 +126,15 @@ trait InteractsWithProperties
             $this->{$property} = $freshInstance->{$property};
         }
     }
+
+    public function only($properties)
+    {
+        $results = [];
+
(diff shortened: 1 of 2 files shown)


danharrin (approved):

calebporzio:
Love it! Thanks @imliam!

calebporzio:
@imliam - would you mind PRing this to the docs at some point?

PR #946, opened 2020-05-09
Missing root tag exception

danharrin:
Before, a component view without a root HTML tag would cause this exception, and it's kinda unclear what's wrong...

Before

This mistake is now caught and clearly presented:

After

A simple component to demonstrate...

Test.php
```php
class Test extends Component
{
    public function render()
    {
        return render('test');
    }
}
```

test.blade.php
```
Test
```


calebporzio:
I dig, thanks @DanHarrin !

PR #947, opened 2020-05-09
Reset properties from array

danharrin:
Currently, `$this->reset()` accepts a [variable-length argument list](https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list) of properties.

This change also allows it to accept an array of properties to reset, for a more dynamic usage possibilities.

A simple component to demonstrate...

```php
class Test extends Component
{
    public $foo = 'bar';
    public $bob = 'lob';
    public $mwa = 'hah';

    public function mount()
    {
        $this->foo = 'baz';
        $this->bob = 'law';
        $this->mwa = 'aha';

        $this->reset(['foo', 'bob']);

        // $this->foo is bar
        // $this->bob is lob
        // $this->mwa is aha
    }
}
```

### DIFF
--- a/src/ComponentConcerns/InteractsWithProperties.php
+++ b/src/ComponentConcerns/InteractsWithProperties.php
@@ -116,6 +116,12 @@ trait InteractsWithProperties
     {
         $propertyKeys = array_keys($this->getPublicPropertiesDefinedBySubClass());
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks Dan!

PR #950, opened 2020-05-10
WIP: Move livewire router macro logic to controller to allow proper route caching

brenjt:
1️⃣ **Is this something that is wanted/needed? Did you create a feature-request issue first?**
Yes

2️⃣ **Does it contain multiple, unrelated changes? Please separate the PRs out.**
No

3️⃣ **Does it include tests if possible? (Not a deal-breaker, just a nice-to-have)**
No, sorry. I tested it locally as best I could and it seems to work as expected. I did have a really hard time getting the unit tests to run correctly. It seems there needs to be a specific setup in order to properly test this and I haven't figured it out.

4️⃣ **Please include a thorough description of the feature/fix and reasons why it's useful.**
This will allow us to use `php artisan route:cache`. We use the component name as magic method call on a new `LivewireController`. The original livewire macro logic is moved to the new controller.

### DIFF
--- a/src/LivewireController.php
+++ b/src/LivewireController.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Livewire;
+
+use Illuminate\Http\Request;
+use Livewire\Livewire;
+use Livewire\Macros\PretendClassMethodIsControllerMethod;
+
+class LivewireController
+{
+    public $request;
+
+    public function __construct(Request $request)
+    {
+        $this->request = $request;
+    }
+
+    public function __call($component, $parameters)
+    {
+        $componentClass = app('livewire')->getComponentClass($component);
+        $reflected = new \ReflectionClass($componentClass);
+
+        $paramters = $reflected->hasMethod('mount')
+                ? (new PretendClassMethodIsControllerMethod($reflected->getMethod('mount'), $this))->retrieveBindings()
+                : [];
+
+        $lastRenderedLivewireView = null;
+        Livewire::listen('view:render', function ($view) use (&$lastRenderedLivewireView) {
+            $lastRenderedLivewireView = $view;
+        });
+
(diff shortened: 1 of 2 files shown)


georgeboot:
See #870

calebporzio:
Thanks for the PR, couple things:

1. There is a new "Controllers" folder this should go under. Something like: `Controllers\LivewireRouteHandler`

2. Very interesting approach with the magic __call, I think I like it

3. The `$this` context of the closure before was probably the Router.php, not Request.php. So although they both have ->current() methods, I imagine this implementation would break trying to set global ->layout()s from route groups.

4. Also, the implementation of the closure has changed on master since you PR'd this if you could update.

Thanks!

georgeboot:
@calebporzio @brenjt Please see PR #870

I've PR'd this before, but `__call()` only works with non-function names since php 7.4.

brenjt:
@georgeboot Yes. I tested this out and can confirm that that is the case. I am working to find a solution that'll work in old php versions.

@calebporzio Any resources you can direct me to in order to set up a project properly in order to develop on this package as well as run its unit tests?  It seems there needs to be a specific setup in order to properly test this and I haven't figured it out yet.

mblarsen:
> It seems there needs to be a specific setup in order to properly test this and I haven't figured it out yet.

@brenjt If you are referring to unit testing maybe all you need is to make sure you run the included phpunit instead of why you have installed globally if any. 

```bash
composer install # you probably did this :)
./vendor/bin/phpunit optional/path/to/test
```

calebporzio:
Planning on this new syntax for V2. Closing this. Thanks!

georgeboot:
@calebporzio Master, you have done well. Really dig the new v2 syntax!

PR #954, opened 2020-05-11
Add latent script evaluation

calebporzio:
The following Livewire view snippet didn't work before, but now it will:
```html
<div>
  @if ($someCondition)
    <script>
      alert('Hello World!')
    </script>
  @endif
</div>
```

This will now work because Livewire will check if an added element is a `<script>` tag, and it will evaluate its contents after being added.


jjjrmy:
Thank you thank you!

jjjrmy:
@calebporzio I don't think this is working if the script is in a template, not in the Livewire component directly.

For example when we discussed doing the inputs as a loop using @include, the scripts in those inputs do not run.
```
@foreach($this->inputs as $input)
    @include('components.form.inputs.'.($input->type))
@endforeach
```

mgussekloo:
Upon dynamically adding a Livewire component that includes javascript, e.g. 

@push('scripts)
alert('Hello');
@endpush

The script doesn't get evaluated and there is no alert.

cosmospham:
It does not work for me. Laravel 7.x, PHP 7.3, Livewire 1.1.0 with Koolreport 4.5, Koolreport Laravel 2.0. The script tag inside component is not executed.
Screen Shot 2020-06-02 at 16 17 10

nuernbergerA:
@mgussekloo and @cosmospham please feel free to open a [new issue](https://github.com/livewire/livewire/issues/new?labels=Type%3A+Bug&template=bug_report.md)

avadis:
Is it solved?

livewire 2.6.7
Laravel 8.x

joshhanley:
@avadis yeah, I just tested it and it works fine for me.

If you are having an issue, please open a help discussion and share your component class and blade view code.

Thanks!

PR #957, opened 2020-05-11
Swap replaceState with pushState

calebporzio:
This PR replaces the current history.replaceState implementation for properties that update the query string using `$updatesQueryString`.

Now, we will use `.pushState` so that users can hit the back button while on the same page that made the query string update.

The back button, instead of triggering an entire page refresh, it will only update the component's state that originally updated the query string.


acurrieclark:
I can't seem to see this in the 1.2.0 release? Is there anything further which needs to be done for this feature to be added to a release?

devcircus:
It was reverted here.  https://github.com/livewire/livewire/commit/94a15745745e4d40e2c90bba2ec151aefd13bfb8

acurrieclark:
> It was reverted here. [94a1574](https://github.com/livewire/livewire/commit/94a15745745e4d40e2c90bba2ec151aefd13bfb8)

Thanks @devcircus! Just stumbled upon that myself too. Will look at the new #1079 pull request for details.

PR #958, opened 2020-05-12
Support file uploads

calebporzio:
This PR will add support for the following:

```php
```

Then, in your Livewire component `$photo` would be a normal property, but would be set to "UploadedFiled" just like a controller would do.

```php
class PhotoUpload extends Component
{
  public $photo;

  public function save()
  {
    $this->validate(['photo' => 'image|mimes:png,jpeg']);
  }

  ...
}
```

### Strategy
* When Livewire detects a file upload on a `wire:model`, it will "sideload" the file via ajax to an internal endpoint that stores it in a temporary Livewire files directory. (with basic validation)
* The endpoint will return the filename (not including the path)
* Livewire will set the value of the `$photo` property to that path on the frontend
* When a request to the backend is made, Livewire will detect that the `$photo` property is a file and convert it to an "UploadedFile" instance.
* Now everything works as normal. Laravel's validator will treat the file as if it was uploaded like a controller
* The user can then store the file any way they like: `$this->photo->store()` and such will work perfectly.

### Security Concerns
Because this feature deals with uploading files in a generic way abstracted from the developer, I'm just worried I'm missing a security concern. Here are some off the top of my head:
* What if a user just hammers the internal Livewire file upload endpoint? Filling up the tmp file directory?
* I have to override Laravel's `UploadedFile` with a `LivewireUploadedFile` to bypass the `is_uploaded_file` call in the `->isValid()` method because the file isn't actually uploaded with the request, we are just faking it. `is_uplaoded_file` is a PHP security feature we are now bypassing.
* Because we're storing files, I'm just generally concerned that a user will be able to fool Livewire into thinking a file is ok, or worse, fool Livewire into storing a file somewhere else in the directory.

I don't want security suspicions to hold this feature back, so I'd love more input and knowledge on these concerns so that we can clearly identify which concerns are actually valid before moving forward with this.

Thanks for your input!


devcircus:
Would it be feasible to offload any of the upload functionality as options in a local service provider where the developer could make some of those decisions. Other packages do a similar thing. I admit I haven't thought through what that would look like, just a thought.

calebporzio:
Yeah, good call. It could be a simple callback to override the default behavior:

```php
Livewire::handleFileUpload(function ($request) {
    $request->validate()...

    $file = $request->file('file');

    $request->file('file')->store(...);

    return $file;
})
```

This would work great I think, the developer could customize validation, storage, everything. All they have to do is return the file at the end of the callback and everything will work.

I like it.

I still would want to provide default behavior out of the box though, so the questions still stand.

georgeboot:
How would you handle scenarios with multiple servers? In that case, the file might get uploaded to a different server than the one that handles the Livewire request?

Vapor also does something similar to get around this problem, by using pre-signed S3 url’s to upload to, and then informing the app soon as the upload finished.

assertchris:
Weird suggestion but what about passing the file data through the normal emit process? You wouldn't be storing it then. Validation would need to change though...

iAmKevinMcKee:
If there are multiple servers, wouldn’t your app already have an S3 bucket or some other shared drive set as the default storage driver if you are uploading images? If you didn’t set this up then you can’t do image uploads. As long as we can define the storage driver for the side loading then everything should work unless I’m missing something.

calebporzio:
Re: multiple servers: Yes, presumably the default storage driver would be something like S3 anyways. There will be a config with this feature to set the config driver likely as well.

@assertchris, good suggestion. I don't want to do that because I don't want to hold up the rest of the component requests AND increase the payload size.

georgeboot:
> Re: multiple servers: Yes, presumably the default storage driver would be something like S3 anyways. There will be a config with this feature to set the config driver likely as well.

Would you consider a pre-signed approach as well, in addition to direct uploads? This would be amazing for the Vapor user out here.

calebporzio:
Hey @georgeboot, please explain how Vapor differs from this strategy as if I was a child 🙏

georgeboot:
Sure haha.

In your approach, we upload the file to the server and have the server store it on a ‘secret’ location. After the upload has finished, we do the LiveWire Ajax request and in that request, we basically fake a file upload so that you can easily validate and stuff. We can now move the “uploaded” file to it’s final destination.

Pros:
- Users can use the familiar file validation methods

Cons:
- Security. Uploading files (and more specifically, storing them) is always a bit of a security risk, even more if it isn't rate limited etc
- On platforms like Vapor, you might be bound to very specific uploads constraints resulting in this approach not working in all cases

In Vapor, we could first request an s3 pre-signed url from the server. Secondly; we have the user upload the file to s3 using that pre-signed url. After the upload has completed, we attach s3’s file ID to the request and tell the server where to find the file.

Pros:
- The file upload isn't actually handled by the server, faster (and cheaper, but let's not get there)
- The file is stored offsite: safer and easier in multi-server setup
- You could still "fake" the uploaded file to give the user standard good old Laravel request validation

Cons:
- you need s3 (or similar)


Given the fact that more and more people are using s3 (or similar) these days, it might be worth it to support this out of the box.

My proposal would be to support both filesystem (for local and/or remote filesystem) and also support direct uploads to s3 for the larger apps

gwleuverink:
This is really great! Exactly how I’d expect it to work.

Just one additional thing that comes to mind concerning security.
From a conference talk about securing php I remember serializing classes that make use of php file functions can potentially cause your app to be vulnerable to [object injection](https://owasp.org/www-community/vulnerabilities/PHP_Object_Injection) attacks.

I  remember hearing on a podcast somewhere that Livewire serializes and deserializes component state. (Not sure though) 

File functions are used in HydrationMiddleware/HydrateFileUploadsAsPublicProperties and also in Laravel’s own UploadedFile class. Just wondering if this can be an issue.

Maybe someone with knowledge on the subject can verify if this is safe? Mine is severely lacking. 

For reference, [this is the conference talk I mentioned](https://youtu.be/kKGGVGiq2y8)
And a [blogpost](https://www.tarlogic.com/en/blog/how-php-object-injection-works-php-object-injection/) explaining how the vulnerability can be exploited

calebporzio:
Great reference, one of the best laracon talks ever.

Livewire doesn't unseriialize any objects. The hydration is all manual. Good call though.

mblarsen:
Is multiple file uploads considered?

gwleuverink:
> Great reference, one of the best laracon talks ever.
> 
> Livewire doesn't unseriialize any objects. The hydration is all manual. Good call though.

Thanks! That's all I needed to hear :) 

I'm going to pull this in and play with it a little when I have some spare time today.

gwleuverink:
> What if a user just hammers the internal Livewire file upload endpoint? Filling up the tmp file directory?

Wild idea. Is using a signed route to upload the files an option? 

This way you can limit the time the upload endpoint stays alive for each component and prevent people from hammering it with files, at least to some extent. 

This isn't fool proof, but maybe something worthwhile. At least you don't have an endpoint that is always open.

---
Also I've been having some trouble getting the file to validate. 
Storing the file if I skip validation does work like expected.

Probably some things still have to be ironed out, or am I missing something?

roni-estein:
> > What if a user just hammers the internal Livewire file upload endpoint? Filling up the tmp file directory?
> 
> Wild idea. Is using a signed route to upload the files an option?
> 
> This way you can limit the time the upload endpoint stays alive for each component and prevent people from hammering it with files, at least to some extent.
> 
> This isn't fool proof, but maybe something worthwhile. At least you don't have an endpoint that is always open.
> 
> Also I've been having some trouble getting the file to validate.
> Storing the file if I skip validation does work like expected.
> 
> Probably some things still have to be ironed out, or am I missing something?

In the past, I've created a hashed route for people to upload files, to a matched hashed directory. Basically it limits them to only pounding only while on that page. Submitting the form processes the files and removes the exposed route and directory.A refresh of the session a timeout or hitting the session too many times (6) in our case did the same thing. In a worst case scenario something similar could probably be adapted here.

gwleuverink:
@roni-estein Wholeheartedly agree. Signing it with the component name limits the url only to work from it's intended component in addition to a maximum amount of time is a nice little extra layer. Not a bad idea to keep in the back of the mind.

In addition to that the route should be throttled I think. In case someone does get through you can at least limit the amount of requests actually being handled.

roni-estein:
> In addition to that the route should be throttled I think. In case someone does get through you can at least limit the amount of requests actually being handled

In our case it was public facing and heavily scrutinized, but it a route like that is heavily vulnerable to some odd hacks. I'm not a network admin, but throttling it to a number of attempts per minute typically prevents some bot attack vectors. Anyhow I hope the ideas help out :)

gwleuverink:
@calebporzio would you consider supporting opt-in chunked uploads? Maybe for a next iteration?

I would gladly have a go at that if this is merged

JayBizzle:
@gwleuverink now THAT would be awesome!

mokhosh:
A couple of questions
Is there an api for getting the uploaded percentage?
Is the service provider hook enough to make this work nicely with libraries like Spatie Medialibrary?

mokhosh:
btw, since we're discussing uploads, do you think we can tackle downloads as well?
https://github.com/livewire/livewire/issues/483

gwleuverink:
@mokhosh  Good question. I don't feel I should go too deep into this, since we don't know if Caleb even considers adding this feature. So 'll keep it brief.

> Is there an api for getting the uploaded percentage?

I suppose the js responsible for uploading the chunks can keep track of the progress and make the input fire an event. That way you have the freedom to handle it any way you like.
something to the effect of this (using alpine):
``` html
<input type="file" wire:model="file" x-on:progress="handleWithJs" />
```

> Is the service provider hook enough to make this work nicely with libraries like Spatie Medialibrary?

Good point. You mean the handleFileUpload callback right? I think you should do any definitive storage logic (using medialibrary for example) inside the component itself. That hook you speak of is for customizing the temporary storage logic for sideloading.

That being said. They do have to play nice. Let's wait and see what caleb thinks.

dev-techguy:
> This PR will add support for the following:
> 
> ```
> <input type="file" wire:model="photo">
> ```
> 
> Then, in your Livewire component `$photo` would be a normal property, but would be set to "UploadedFiled" just like a controller would do.
> 
> ```
> class PhotoUpload extends Component
> {
>   public $photo;
> 
>   public function save()
>   {
>     $this->validate(['photo' => 'image|mimes:png,jpeg']);
>   }
> 
>   ...
> }
> ```
> 
> ### Strategy
> * When Livewire detects a file upload on a `wire:model`, it will "sideload" the file via ajax to an internal endpoint that stores it in a temporary Livewire files directory. (with basic validation)
> * The endpoint will return the filename (not including the path)
> * Livewire will set the value of the `$photo` property to that path on the frontend
> * When a request to the backend is made, Livewire will detect that the `$photo` property is a file and convert it to an "UploadedFile" instance.
> * Now everything works as normal. Laravel's validator will treat the file as if it was uploaded like a controller
> * The user can then store the file any way they like: `$this->photo->store()` and such will work perfectly.
> 
> ### Security Concerns
> Because this feature deals with uploading files in a generic way abstracted from the developer, I'm just worried I'm missing a security concern. Here are some off the top of my head:
> 
> * What if a user just hammers the internal Livewire file upload endpoint? Filling up the tmp file directory?
> * I have to override Laravel's `UploadedFile` with a `LivewireUploadedFile` to bypass the `is_uploaded_file` call in the `->isValid()` method because the file isn't actually uploaded with the request, we are just faking it. `is_uplaoded_file` is a PHP security feature we are now bypassing.
> * Because we're storing files, I'm just generally concerned that a user will be able to fool Livewire into thinking a file is ok, or worse, fool Livewire into storing a file somewhere else in the directory.
> 
> I don't want security suspicions to hold this feature back, so I'd love more input and knowledge on these concerns so that we can clearly identify which concerns are actually valid before moving forward with this.
> 
> Thanks for your input!

Hello...when will you relaese this feature i persoanally i need it.

neeravp:
A thought:
Can we have a `FileBag` just like `ErrorBag` available globally for all Livewire components.
When a file is uploaded, save the file to `sys_temp_dir` and create an instance of `UploadedFile` and push it to the `FileBag`. 
When the user wants to save the form with other fields, the file can be pulled from the `FileBag` and since it is an instance of 'UploadedFile` normal Laravel validation checks can be applied.
My current implementation (rough) for the file upload (image) is as under:

Laravel anonymous component for Image(File) upload
```blade
@props(['fileTypes' => "jpg,jpeg,png,webp", 'maxUploadSize' => '1MB', 'name' => '', 'src' => ''])
<div 
    class="mb-4" 
    x-data="{ 
        label: 'Choose File', 
        fileTypes: '{{$fileTypes}}', 
        maxUploadSize: '{{$maxUploadSize}}', 
        onFileLoad: onFileLoad,
        dataUrl: '{{$src}}'
    }"
>
    <img x-show="dataUrl" :src="dataUrl" :alt="'{{$name}} image'">
    <label
        {{$attributes->merge([
            "class" => "flex items-center w-full border border-dashed border-gray-200 rounded outline-none bg-transparent p-2 rounded focus:border-blue-300"
        ])}}
    >
        <svg class="w-8 h-8" fill="currentColor" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
            <path
                d="M16.88 9.1A4 4 0 0 1 16 17H5a5 5 0 0 1-1-9.9V7a3 3 0 0 1 4.52-2.59A4.98 4.98 0 0 1 17 8c0 .38-.04.74-.12 1.1zM11 11h3l-4-4-4 4h3v3h2v-3z"
            />
        </svg>
        <span class="ml-2 text-base leading-normal" x-text="label"></span>
        <input type="file" class="invisible" name="{{$name}}" @change="onFileLoad" />
    </label>
   
        @foreach($errors as $error)
            <p class="text-xs text-red-500">{{$error}}</p>
        @endforeach
   
</div>

<script>
    function onFileLoad($event) {
        let errors = [];
        const maxUploadSizeInBytes = function(value) {
            value = value.trim().toUpperCase();
            if(value.endsWith('MB')) {
                return parseInt(value.replace(/[KkBbMm]/g, '')) * 1024 * 1024;
            }
            if(value.endsWith('KB')) {
                return parseInt(value.replace(/[KkBbMm]/g, '')) * 1024;
            }

            return parseInt(value);
        }
        const isValidFile = function(file, fileTypes, maxUploadSize) {
            maxUploadSize = maxUploadSizeInBytes(maxUploadSize);
            // console.log(`isValidFile fileTypes: ${fileTypes}, maxUploadSize: ${maxUploadSize}`);
            let acceptedMimes = fileTypes.trim().split(',').map(type => `image/${type}`);
            let valid = true;
            if(! acceptedMimes.includes(file.type)) {
                valid = false;
                let phrase = fileTypes.replace(/,/g, " or ");
                errors.push(`The file type is not allowed. Please upload ${phrase} file.`)
            }
            if(file.size > maxUploadSize) {
                valid = false;
                let sizeInKb = maxUploadSize / 1024;
                let sizeForHumans = sizeInKb < 1024 ? `${sizeInKb} KB` : `${sizeInKb/1024} MB`;
                errors.push(`The file exceeds the maximum allowed size of ${sizeForHumans}.`)
            }
            window.livewire.emit('uploadErrors', errors);
            return valid;
        }

        /** If there are no files return without doing anything */
        if (!$event.target.files.length) {
            return;
        }
        let file = event.target.files[0];

        /** Check if the file is valid as per the accepted file type & max allowed size */
        if (isValidFile(file, this.fileTypes, this.maxUploadSize)) {
            // console.log(file);
            /** use FileReader to get the dataUrl for preview */
            let reader = new FileReader();
            reader.readAsDataURL(file);
            reader.onload = e => {
                let dataUrl = e.target.result;
                this.dataUrl = dataUrl;
                this.label = file.name;
                // window.livewire.emit(`${$event.target.name}Uploaded`, dataUrl, file.name, file.type.split('/')[1], file.size);
                window.livewire.emit('fileUploaded', $event.target.name, dataUrl, file.name, file.type.split('/')[1], file.size);
            };
        }
        // console.log("changed", $event.target.name, $event.target.files, this.fileTypes, this.maxUploadSize, errors);
    }
</script>
```
Then I have a trait to be used with Livewire components

```php
<?php

namespace App\Http\Livewire\Concerns;

use finfo;
use Exception;
use Illuminate\Support\Str;
use Illuminate\Http\UploadedFile;
use Illuminate\Support\Facades\File;
use Illuminate\Support\Facades\Storage;

trait InteractsWithFiles
{
    protected $files = [];

  //Trying to simulate a FileBag with protected $files property
    public function fileUploaded($field, $file, $filename, $extension, $size)
    {
        $file = $this->createTempFile($file, $filename);
        if ($file->isValid()) {
            $this->files[$field] = $file;
        } else {
            $this->feedback('File upload error. Please try again', 'error');
        }
    }

    protected function createTempFile($content, $filename)
    {
        $path = tempnam(sys_get_temp_dir(), uniqid('livewire', true));

        $finfo = new finfo(FILEINFO_MIME_TYPE);
        File::put($path, base64_decode(Str::after($content, 'base64,')));

        return new UploadedFile(
            $path,
            $filename,
            $finfo->file($path),
            0,
            true
        );
    }

}
```

In any Livewire component view

```blade
<x-image-field name="avatar" :src="$user->avatar" file-types="jpg, png, webp" />
```
And in component class

```php
<?php

namespace App\Http\Livewire\Users;

use Livewire\Component;

class UserProfile extends Component
{
    public $user;
    public $name;
    public $email;
    public $avatar;

    public function mount($user)
    {
        $this->user = $user;
    }

    public function save()
    {
        $validated = $this->validate([
            'name' => ['required', 'string', 'min:6'],
            'email' => ['sometimes', 'required', 'email', 'unique:users'],
            'avatar' => ['image']
        ]);

        $this->user->fill($validated)->save();
    }
        
}

**Comment by marbuser** [comment]
Would this be strictly for use with input type file or would you also be open to exposing the internals so that if a developer chooses they could implement an custom upload component like Dropzone or Filepond as well?

**Comment by mokhosh** [comment]
Who is this Caleb guy? He’s really active.

**Comment by JohnRoux** [comment]
Big fan of the activity here. Got a project which I've had to pause until the upload feature (ideally via S3 as well) is in livewire.

So really appreciate the work being done here 🙏👍

**Comment by GertjanRoke** [comment]
The only thing that is missing is #noPlansToMerge

File: src/TemporarilyUploadedFile.php [hunk 1]
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public static function createFromLivewire($filePath)
+    {
+        return new static($filePath, FileUploadConfiguration::disk());
+    }
+
+    public static function canUnserialize($subject)
+    {
+        return Str::startsWith($subject, 'livewire-file:')
```
gregoriohc, line 117: This fails if you are using public component properties other than anything that could be converted to a string (an array or a model for example).
In the documentation, it says: "Properties can ONLY be either a JavaScript-friendly data types (string, int, array, boolean), OR an eloquent model (or collection of models)."
So, maybe it could be necessary to check if the subject is a string?

georgeboot: @gregoriohc please open an issue for this.

gregoriohc: Done: #1073 👍

marioguima:
Congratulations on the great job
However I am getting an error when using wire: model with input file
**Cannot set property ‘value’ of undefined (FileUploads.js)** 
_I detailed the problem and the steps to reproduce in the forum_
https://forum.laravel-livewire.com/t/cannot-set-property-value-of-undefined-fileuploads-js/2474

I believe that I may have missed something to have caused this error, however, it's been 5 days and I can't find a clean solution
I appreciate your time and your help

PR #961, opened 2020-05-13
Fix #953 Emit function on the tests

### LINKED ISSUE #953: Events emitted on tests don't send the correct argument
**Describe the bug**
Testing a component that expects an event to be emitted with parameters, the parameters are received on the method as only one on array format.
This happen just in the test not from the browser.

**To Reproduce**
Create a component, 
Create a listener for an event that expects a parameter
Create a test for the component in which emits the event with a parameter
See the error that the parameter receive is an array that contains the parameter sent

**Expected behavior**
The parameter should be the same that was send on the test.

**Additional context**
I found a fix for this. I am creating first the bug as the readme indicates before crating the PR.
The fix is in the trait of MakesCallsToComponent on the function emit is sending the $params variables to the "fireEvent"
        $this->fireEvent($event, $params);
To fix it just add the spread operation on $params
        $this->fireEvent($event, ...$params);

adalessa:
Creating tests for my project I found that the emit method was not sending the values as expected. Instead of the value I was receiving an array, after digging in the code I found that the emit was getting the parameters as spread but instead of sending in the same way to to fire event was just sending the array.

Note: I didn't find any test using the emit method, all the tests were using the "fireEvent" directly that's why I create a new test file for this.

### DIFF
--- a/src/Testing/Concerns/MakesCallsToComponent.php
+++ b/src/Testing/Concerns/MakesCallsToComponent.php
@@ -6,7 +6,7 @@ trait MakesCallsToComponent
 {
     public function emit($event, ...$parameters)
     {
-        return $this->fireEvent($event, $parameters);
+        return $this->fireEvent($event, ...$parameters);
     }
 
     public function fireEvent($event, ...$parameters)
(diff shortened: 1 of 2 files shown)


calebporzio:
Ooof, good catch, thanks!

Issue #965, opened 2020-05-14
Checkbox's value is ignored when provided, boolean used instead


ragulka:
**Describe the bug**
When a [value is provided for a checkbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value), it's ignored when bound using `wire:model`. Instead fo sendign the actual value, a boolean `true/false` is sent.

**To Reproduce**
1. Create a checkbox `<input type="checkbox" wire:model="wantsCoffee" value="yes" />`
2. Click on the checkbox to trigger the AJAX request to backend
3. Inspect the request sent to the server :

```
"actionQueue": [{
    "type": "syncInput",
    "payload": {
        "name": "wantsCoffee",
        "value": true
    }
}]
```

**Expected behavior**
I would expect the `value` in the `payload` to be `"yes"`. This would follow the default browser behavior.

**Desktop:**
 - Browser: Firefox

ahmedash95:
Hey @ragulka, I was able to reproduce it and purposed a fix. hope it gets merged soon :)

calebporzio:
This has been fixed and will be in the next release.

PR #969, opened 2020-05-15
Fix checkbox custom value always boolean

ahmedash95:
This PR fixes an issue #965 with checkbox values. usually, the browser would send the text in the `value` attribute. the current behavior is boolean which is not what we expect.

We have to check if the attribute exists because `return this.el.value || this.el.checked` would return the value always. as default value on chrome is `on`


calebporzio:
Thanks!

PR #973, opened 2020-05-16
chore: add prettier config that disables prettier

mblarsen:
For users of IDEs with a global prettier/formatting config it can be
difficult to work on projects that has no configuration.

This config adds `requirePragma` (which you can also set globally),
which require you to at a pragma to the top of the file before prettier
starts formatting.  So unless you add `// @format` or `// @prettier` to
the top of the file nothing is formatted.

This allows Livewire to start adapting formatted code little by little.

I've also added a few configs that makes prettier behave very close to
the way the code has been formatted already.

Ideally someone should just run prettier on all the code and get it done
with :rocket:

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,5 +39,11 @@
         "@rollup/plugin-node-resolve": "^7.0.0",
         "rollup-plugin-output-manifest": "^1.0.2"
     },
-    "dependencies": {}
+    "dependencies": {},
+    "prettier": {
+        "arrowParens": "avoid",
+        "requirePragma": true,
+        "semi": false,
+        "singleQuote": true
+    }
 }


lancepioch:
We currently use StyleCI. Would this collide? If not, what extras would it have?

mblarsen:
> We currently use StyleCI. Would this collide? If not, what extras would it have?

Prettier formats code in your editor or pre-commit however you set it up. No linting or anything.

Many developers use prettier (because it is awesome) by default in their editor. That means when saving the Livewire JS there will be a bunch of unnecessary changes to the code that makes it harder to commit

This PR adds a config that prevents Prettier form messing with the code unless told to.

You could say my intension is to encourage Livewire to use automatic formatting to avoid style guide discussions. 

It doesn't affect StyleCI AFAIK.

Perhaps this issue should be converted to a discussion. Is that possible?

mblarsen:
Anyway, the PR doesn't change much for most users, but it allows for adaption of auto-formatted code.

calebporzio:
Thanks for the contribution, however, I'd prefer to keep prettier out of the repo for now.

Issue #845, opened 2020-04-22
Component Key ID for $emitTo, like: $emitTo('order-item', {{ $order->id }}, 'eventName', params...)


basepack:
**Is your feature request related to a problem? Please describe.**
- I got a long list of Livewire components **(list-item)** think a long table. 
- Each individual list-item component has a good bunch of markup, so its HTML heavy
- When I do a `EmitTo('list-item')` kind of event, _**they all reload :astonished:**_, so in my case 300 XHR calls are made in the background.

**Describe the solution you'd like**
Well... consider this beauty: 
```
$emitTo('component', 'componentID', 'eventName', params...)
```

And with `componentID` I mean the given key. For example`:key="$order->id"`, and **not** the random id's Livewire assigns, like `wire:id="ZnmQ6dpTXFtk5a1C2v2H"`.

So the end result could be:
```
$emitTo('order-item', {{ $order->id }}, 'refresh')
```

mblarsen:
I was think of a similar use case. A have a survey form builder with lots of questions. Questions has an input put type, think widgets. When you select the type question component emits to a “singleton” type selector catalog component to present itself. Question ID included in event. Once the user selects the question type a second event is emitted, say “type selected”. Like you I have narrowed it down with emitTo question component. But there are just too many and all of them have to check if they are the question that initiated the type selector. 

So the pattern is:

1. Emit to start something 
2. Emit to get a reply back

I’m not sure how similar this is to your case. Anyway, I imagined an api like:

Initial event passes a specific Livewire token data type which is both stored in session and included in payload to browser. 

```php 
emit(“typeRequested”, Livewire::token(“question-type”))
```

Anyone listening to the event can store the token and later use it to “call back” or “reply to” the token owner. 

```
emitTo(“#question-type”, “typeSelected”, $selected);
```
(The hash-sign denotes a token rather than a component name)

So only one component needs update and on the plus side you don’t need to know the component ID.

Issues: 

- multiple use of same token (solution: last issuer wins or let Livewire pack and unpack a counter)
- multiple calls to same token (solution: ignore, fail?)

Was going to try to make a PR for this, but love to get your input the use case.

basepack:
Well, in my use-case I want Livewire to be able to callable on their keys. Because the key is something that is known. The generated Livewire token is not actually.

So you create a bunch of components like this:

```
@foreach($this->orders as $order)
      <livewire:order-item :order="$order" :key="$order->id"/>
@endforeach
```

[...]

PR #975, opened 2020-05-17
emitTo directly to a component instance whithout knowing wire:id

mblarsen:
**Introduction**

The is a work-in-progress PR that lets the user `emitTo` directly to a single component instance without knowing its ID using a *token* concept.

I've posted this as a _draft_ PR in hope that people who has a bit deeper understanding of Livewire can tell me if there are a better approach to solve this problem, fully or in part. Also there are some logic issues that I would like input on, so think of this PR as a starting point we can talk about and not a final solution.

**Motivation**

> I found that when building forms or list-like components with lots of child-components of the same kind that in order to emitTo them you'll either have to suffer lots of requests because Livewire emits event to all components (of the same kind when using emitTo). You can resort to JavaScript that passes along the wire:id, but that seems hackish. 

_Example use case_: Survey form builder. A survey has lots of questions of various types: text, multiple choice, file uploads, etc. When creating a new question a modal pops up that lets the user chose a question type after which they are token to the appropriate question editor.

1. Question notifies that it is new and needs a type
2. Type selector opens and presents options for types
3. A type is clicked and an event is emitted to question component

The last point is where the problem lies since I can at best `emitTo('question'`)` which will limit the number of requests to the number of questions which could still be a lot (see more #845).

The alternative approach would be to emit the wire:id along with the first event that causes the type selector to open to the effect that it knows who to "call back to" when the type is selected. This however would require some custom wiring with JavaScript (see example in #845).

This PR attempts to solve the problem a more Livewire-esque way which require less manual wiring in the front-end. 

**Implementation**

I've thought to solve the problem using a concept of **tokens**. A token is a named value a user can pass along as parameter when emitting events. It is a way by which you refer to a component instance by a discrete name. ~Much the same way you name events. In fact you can think of them as "future events"~ (updated see 4303c42).

```php
// Question.php
$this->emit("questionCreated", $question->id, Token::for("question", $this));
```

The question question editor above emits an event that a new question was created (saying that it needs a type from the type selector). By including a token it lets another component **refer to it by name** instead of by wire:id.

For example when we emit from the question type selector that a type was selected we can not emit directly to the token:

```php
// TypeSelector.php
public function selectType($id)
{
    $this->emitTo(Token::for('question'), 'typeSelected', $id);
    // easy to add this API
    // $this->emitToToken("question", "typeSelected", $id);
}
```

The type selector doesn't know anything about what exact question asked for the type as it shouldn't. But since a token is targeted **only the issuer of that token will be notified**.

On the receiving end of an emit to token it works ~much~ like normal events ~except that you don't have to manually register events. You can but you don't have to~.

**Details**

- `Token` resolves a `TokenStorage` from the app container which is currently set to `LivewireManager` that through a new `StoresToken` trait provides a session storage for setting and retrieving tokens.
- Hooked into `RegisterEmittedEvents` to extract the token parameters and register the tokens on the payload.
- Hooked into `emitTo` on the client-side to emit an event to the specific component that issued the token
- ~Modified `firesEvent` to make it work seamless with the existing event system.~

**Outstanding issues, notes, and questions**

- API can be improved for convenience to get rid of `Token::for`, e.g. `$this->emitToToken()`
- Currently tokens can only be added through the event system. By allowing the user to add a token to the payload one could target any component without an "initiating" an event in the first place. That scenario just fit my case.
- What if several tokens are created for the same value. Currently last one wins. Should exception be thrown for better DX?
- Currently the token isn't removed after "being used", it probably should. Ideally it should follow the `mount` cycle so that the storage is fresh on each first page load. Not sure where to hook into that?
- I opted for creating `tokens` on the response payload instead of using `events` directly which would have been another approach. However, the `RegisterEmittedEvents` is applied in the wrong order for that to work nicely.
- Currently it is assumed that the token is also the event name, but would it be better to introduce back a specific event name? Personally I think yes, as this makes for clearer tokens and only one token per instance per event is needed. **UPDATE** implemented in 4303c42
- Client side `$emitTo` not supported yet
- TEEEEEEEEEEEEEEEEEEEEESTS


calebporzio:
Thanks for the contribution @mblarsen, however, I'm hesitant to go with something like this, I think it adds too much complexity to the codebase.

I actually am a little fuzzy on the problem you're trying to solve.

Please explain the component hierarchy and the communication needs a bit better, I may be able to help with an existing solution. Thanks.

mblarsen:
Hi @calebporzio I agree with you on the complexity. That is should kept to a minimum.

Some of the my implementation could be simplified by extracting to its own HydrationMiddleware, like you did recently with file upload.

Anyway, this PR is presents a way to solve a problem in a less hack-ish way than currently seems to be the way to do it.

In my case I have a hierarchy of components like the following (simplified):

```
Survey
  TypeCatalog
  Question
  Question
  Question
  Question
  Question
  Question
  Question
  .. and many many more
```
The component `TypeCatalog` is quite big, but essentially a modal that lets you select what type a question is. I've opted to include it at a higher level to avoid duplication.

When the user clicks 'select type' the `TypeCatalog` opens as a response to listening to "selectType". When the catalog is open the user selects a type which `emitTo(question)` an event "selectedType". 

However, `emitTo` isn't very targeted so you'll see one post for every question. Here comes the hack-ish part. 

To reduce the posts I need to know the wire id of the exact question that "requested" the catalog. To do so I need to include that in the initial event, I need to create a public property on `TypeCatalog` (to have it "persisted") and then I have to write custom JavaScript (`window.find...`) mimicking the way Livewire JS part works. This is what I do now. I doesn't feel good. I'm writing code to work around my frameworks intended use.

I know that I can just include the `TypeCatalog` in each and every question. But I'm left with the same feeling that I'm doing this because of Livewire. Am I whining too much? :)

So the PR adds a token or a key to the payload identical to the way events works. The complexity you refer to comes from the fact that I didn't want to change the API and keep it as usual. E.g. the token is included as a regular param (one or more), but is removed as a param and added to `tokens` in payload. 

(maybe you Token interfaces and what not is also what you think of as added complexity. Can remove that, just added for flexibility)

The PR provides a way for components to communicate (vertical and horizontal alike) without the need to know of implementation details of Livewire.

Another user had a different use case with a list of things where each thing had some relation to other elements in the same list. The requested feature was emitTo that allowed using the key. There is some overlap in these cases. 

If you don't mind I'll return this PR to draft state and try remove some of the complexity (e.g. using a middleware). I've learned a bit along the way. This is my first time looking at the inards of Livewire :)

basepack:
@calebporzio could you elaborate on the answer @mblarsen provides? 

I think this PR would be a huge benefit for the flexibility of Livewire. It lets one target a single (child) component from everywhere with `emitTo` with a know id. Also see my issue for an explanation why one would use this: https://github.com/livewire/livewire/issues/845

Thanks!

mblarsen:
I’ll try to make a cleaner implementation that doesn’t hook into everything. Just a bit short of time.

mattharcourt:
Hi @mblarsen, I think there may be a simpler way to have emitTo target a specific instance using the key values. I can work on a PR if interested.

mblarsen:
Yeah, that’s the path I was going to take. You are welcome to go ahead. I’m a bit short of time.

mattharcourt:
See #1181

mblarsen:
Hi Matt, I think this fits well with the use case described in #845; I think @basepack will be happy about that. 

Unless I'm missing something you still have to know the key (which is often many times easier than knowing the component id). Is that right?

This PR introduces a token, a discrete value used in place of an ID. The difference is that "something else" would have to define that beforehand. E.g. as exemplified with the survey question asking the question type catalog for help picking a new type. The question takes the place of that token and when it comes to selecting the type the catalog emits to the token unknowing of who it is. Practically it means not having to pass values around with and maintain props of the component being called.

When I mentioned keys as a solution, it was to reduce some of the complexity I had introduced with adding another key type.

Anyway, your solution is definitely an improvement upon the current functionality. Thanks for the PR. I hope it gets merged.

PS: you reference #945 (issue and PR template) in your PR was that a mistake?

mattharcourt:
@mblarsen, yes 945 was mistake, meant to reference this one - I fixed it. Thanks!

Yes, you still have to know the key. But presumably the dev wanting this type of specific communication between components will know the key (either will have assigned it, or used some known algorithm for assigning). In other words, if I intend to emit event to component X, I should assign X a key and then use that in the emitTo.

PR #977, opened 2020-05-18
Typo: window.wire and not window.livewire

JunaidQadirB:
You must mean `window.wire` and not `window.livewire` under **V2 Roadmap** heading.

Since it is related to documentation so, not Applicable. 

No

Not Applicable

As of the Roadmap for v2 the references for `livewire` will be replaced with `wire`, so this should be `window.wire` and not `window.livewire`


devcircus:
I believe it's correct. The window reference will be capitalized. 🤷🏻‍♂️

PR #980, opened 2020-05-18
Fix Testing Livewire Route Conflict

wsamoht:
# Problem
There is a conflict with Livewire's testing route in an edge case.

If a Laravel project has a route like the following
```php
Route::get('/{type}/{number}', [OrganizationController::class, 'show']);
```
it causes a conflict with Livewire's testing route making the route above get called instead of Livewire's. This makes all Livewire tests fail.
```php
$randomRoutePath = '/testing-livewire/'.Str::random(20);

Route::get($randomRoutePath, function () use ($name, $params) {
    return View::file(__DIR__.'/../views/mount-component.blade.php', [
        'name' => $name,
        'params' => $params,
    ]);
});
```

# Solution
This PR changes the code to put Livewire's testing route as the first route instead of the last.

### DIFF
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -4,8 +4,9 @@ namespace Livewire\Testing;
 
 use Livewire\Livewire;
 use Illuminate\Support\Str;
+use Illuminate\Routing\Route;
 use Illuminate\Support\Facades\View;
-use Illuminate\Support\Facades\Route;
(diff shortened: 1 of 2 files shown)


wsamoht:
@calebporzio I've added a test for this. Let me know your thoughts!

vitalii-mv:
Have same problem

calebporzio:
Ah, this makes sense. Thanks!

PR #981, opened 2020-05-19
Rename the component in views when using the move command

Elhebert:
First of all, thanks for this awesome project, it's 🔥 !

**Description**

The `livewire:move` command is great, it saves me a lot of time every time I need to rename or move a component. However I tend to forget to update the views and routes where that component is used. This PR solves this issues.

If this feature is something that is welcome in the codebase, I'd like to also update the route files to rename component in `Route::livewire` calls. 

I'll also update the test suite to support the feature.

**Caveats**

Something to keep in mind is that it'll slow the command if the project contains a lot of view files and/or big views.

### DIFF
--- a/src/Commands/MoveCommand.php
+++ b/src/Commands/MoveCommand.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Commands;
 
+use Illuminate\Support\Str;
 use Illuminate\Support\Facades\File;
 
 class MoveCommand extends FileManipulationCommand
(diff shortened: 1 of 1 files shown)


calebporzio:
Nice PR! However, this feels too dangerous an operation to perform from Livewire. Thanks for being willing to contribute!

mblarsen:
Maybe a notice/warning telling the user to do so. Even though it is obvious it is easy to forget.

Elhebert:
Thanks @calebporzio. It could indeed have unwanted side effects, so I totally understand.

PR #997, opened 2020-05-21
[PHP] Resolve default value if given value is null

### LINKED ISSUE #992: Actions ignore parameter default values
### Description
If an action parameter has a default value, the value is ignored when the action is run.

### Steps to reproduce
```php
// in component

public function handleClick($first, $second = 'hello'): void
{
    dump($first);
    dump($second);
}
```

```html
<!-- in html -->
<div wire:click="handleClick('imasay')" class="cursor-pointer">Click Me</div>
```
### Result
```sh
"imasay"
null
```

**Context**
- Livewire version: 1.1.0
- Laravel version: 7.9.2
- Browser: N/A

nuernbergerA:
Currently, action calls with optional parameters are ignored and set to null.

This PR fixes this, without changing the current DI implementation.

```php
public $foo = '';

public function doSomething($foo = 'default') {
    $this->foo = $foo;
}
```

```blade
<button wire:click="doSomething">
<!-- $foo should be 'default' // currently its null -->
```

```blade
<button wire:click="doSomething('bar')">
<!-- $foo should be 'bar' -->
```

---
Closes #992


devcircus:
Hey, thanks for jumping on this. Check out PR #910 though. I reported another bug(#904)  a couple of weeks ago, and the resulting PR made changes to the same method. If that one is merged, there will be merge conflicts. Looks like we'll need a solution that takes both issues into consideration.

nuernbergerA:
I know there is this PR but I don't use `app()->call()` and just extend the current implementation.

calebporzio:
I believe this will break when intentionally passing `null` as a parameter explicitly.

nuernbergerA:
How can we differentiate between `not given` and `null`
Thinking ahead for DI rework.

calebporzio:
Thanks Adrian! @devcircus - thanks for that, but app()->call() won't work for us because it expects a key->value pair of params and we don't have one.

PR #999, opened 2020-05-22
Allow expressions in action params

mattharcourt:
The current implementation does not allow an expression to be used as an action parameter. For example:
```javascript
<button wire:click="submit(new Array('An', 'expression').join(' '))>Submit</button>
```
Will fail with an error:
```javascript
Uncaught SyntaxError: missing ) after argument list
    at directive.js:54
```

This PR alters the way action parameters are parsed and evaluated to utilize `eval`, and an inline function that returns it's arguments, to do the parsing and evaluating of the action parameters.

A test is included.


js/dom/directive.js
```diff
@@ -51,7 +51,8 @@ export default class {
             // This "$event" is for use inside the livewire event handler.
             const $event = this.eventContext
             method = methodAndParamString[1]
-            params = methodAndParamString[2].split(/ ?, ?(?=(?:[^"'`\[\]]|'[^'\\]*(?:\\.[^'\\]*)*'|"[^"\\]*(?:\\.[^"\\]*)*"|`[^`\\]*(?:\\.[^`\\]*)*`|\[[^\]]*\])*$)/).filter(Boolean).map(param => eval(param))
+            // use a function that returns it's arguments to parse and eval all params
+            params = eval('((...params) => [...params])(' + methodAndParamString[2] + ')')
```
nuernbergerA, line 55: Bonus: you remove my nasty and unreadable regex 👍

calebporzio: Seriously. This is dope.

calebporzio:
Fantastic contribution. Thanks!

PR #1009, opened 2020-05-24
feat: call 'updating' and 'updated' for $set and $toggle

mblarsen:

Asked on discord:

> Is there any reason why updating and updated isn't run when using $set and $toggle? Only the property specific methods are run e.g. updatingMyProp and updatedMyProp.
PerformPublicPropertyFromDataBindingUpdates calles updating and updated but it skips anything that is not syncing (as the name suggests).
I would expect set and toggle would call them as well.

No, just invokes updating and updated.

Works of existing tests, which passes. Doesn't add extra tests.

Currently `updatingMyProperty` and `updatedMyProperty` is called as expected, however, surprisingly `updating` and `updated` isn't.

This PR adds those calls.

This is useful in cases where `$set` and `$toggle` is called with dynamic input and you want to trigger something based on the affected property value change. Just as it would be when model binding is used.

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -48,11 +48,19 @@ trait HandlesActions
         $beforeMethod = 'updating'.$propertyName;
         $afterMethod = 'updated'.$propertyName;
 
+        if (method_exists($this, 'updating')) {
+            $this->updating($name, $value);
+        }
+
         if (method_exists($this, $beforeMethod)) {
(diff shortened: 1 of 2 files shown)


calebporzio:
Ah, thanks for this. Good catch!

Would you mind adding tests that would currently fail, but will now pass with your PR? Thanks!

mblarsen:
Yes, will add tests.

mblarsen:
@calebporzio added tests and then → https://github.com/livewire/livewire/pull/1045

calebporzio:
Thanks!

PR #1025, opened 2020-05-29
set-runs-hooks

### LINKED ISSUE #1024: $set does not run updated hook
### Description
When we call wire:click.prevent="$set('amount', 1)" in the browser it do not run updated($key) hook on server side.

### Steps to reproduce
1 - Set the listener: wire:click.prevent="$set('amount', 1)"
2 - Set the hook: public function updated($key) { info('updated', compact('key')); }
3 - Try to click and check laravel logs.

**Context**
- Livewire version: 1.1.0
- Laravel version: 7.13.0
- Browser: Chrome

indigoram89old:
When we call wire:click.prevent="$set('amount', 1)" in the browser it do not run updated($key) hook on server side.

Fixes #1024

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -48,6 +48,10 @@ trait HandlesActions
         $beforeMethod = 'updating'.$propertyName;
         $afterMethod = 'updated'.$propertyName;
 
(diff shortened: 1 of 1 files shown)


mblarsen:
Duplicate of #1009

PR #1028, opened 2020-05-29
[Feature] Add new resetExcept()

rasim:

Yes it is needed. eg: `request()->except()` exists for laravel.

No it doesn't

Yes it is attached in the test.

`$this->reset()` exists. But there are no exclusions to be excluded. 

The usage for this is as follows;

`$this->resetExcept([ 'foo', 'bar']);`

In this way, external reset of the 'foo' and 'bar' is performed.

### DIFF
--- a/src/ComponentConcerns/InteractsWithProperties.php
+++ b/src/ComponentConcerns/InteractsWithProperties.php
@@ -143,4 +143,15 @@ trait InteractsWithProperties
 
         return $results;
     }
+    
+    public function resetExcept(array $exceptProperties = [])
+    {
+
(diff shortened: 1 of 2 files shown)


calebporzio:
Thank you for your contribution. However, I'm going to hold off on this for now.

You can always use a "whitelist" with `->reset()` rather than the blacklist you are trying to use (I understand that it's more verbose, but I figured I'd mention it. Thanks!)

rasim:
Hey @calebporzio,
When I define a new variable to the component, I have to specify it every time. Creating a blacklist can be more useful with this method.

PR #1029, opened 2020-05-29
Added Layout Config on component Layer

vittoriodev:
Yes, it's needed. But regarding the need for others: I'm not sure but I think this should be available. There is no other easy way to do things like Metadata or strings in layouts when they're computed.

Nope, just one.

Yes 😎 

If you want to give computed layouts or layout params on a component layer to your layout file while using `Route::livewire()` you can now use two new functions directly in the livewire component instead of using the `->layout()` function: `layout()` and `layoutParams()` like this:

```

class ComponentWithLayout extends Component
{

    public function layout()
    {
        return 'layouts.app-with-bar';
    }

    public function layoutParams()
    {
        return [
            'bar' => 'baz'
        ];
    }

    public function render()
    {
        return app('view')->make('null-view');
    }
}

```

You're welcome 😎 Thanks back, for building livewire. 😉

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -341,4 +341,15 @@ HTML;
             new MountMethodMissingException($instance->getName())
         );
     }
+
+    public function getLayoutParams($component)
+    {
+        $element = $this->activate($component, Str::random(20));
+        return method_exists($element, 'layoutParams') ? $element->layoutParams() : null;
+    }
+
+    public function getLayout($component) {
+        $element = $this->activate($component, Str::random(20));
+        return method_exists($element, 'layout') ? $element->layout() : null;
+    }
 }
--- a/src/Macros/RouterMacros.php
+++ b/src/Macros/RouterMacros.php
@@ -31,9 +31,14 @@ class RouterMacros
                 $componentClass = app('livewire')->getComponentClass($component);
                 $reflected = new \ReflectionClass($componentClass);
 
+                // the default is catched by the route action "layout" or the fallback.
+                // This can be overwritten from the layout function of a component.
+                $layout = $this->current()->getAction('layout') ?? 'layouts.app';
+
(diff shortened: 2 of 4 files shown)


calebporzio:
Hey @vittorioe, thanks for the contribution!

Some thoughts:
1. With this approach people wouldn't be able to use data set from their mounted() hooks to provide to the layout.
For example:
```php
public function mount($user) {
  $this->user = $user;
}

public function layoutParams() {
  return ['title' => $this->user->name];
}
```
(Because the component is being created as a fresh instance for the sole purpose of running this method)

2. I think the word "->layout" may be a little confusing because people might assume that it applies that layout to EVERY request, but it really only applies it to the first request IF the component is used as a Route::livewire() component.

3. I would rather have one method that does both things than 2 methods: layout & layoutParams.

4. I'm not sure a dedicated method is the best API here. I know it's the easiest, but there may be a better API for this behavior.

Something that was talked about (and actually implemented before backing out), was forwarding all view data for a component to the layout file. For example:

```php
public function mount($user) {
  $this->user = $user;
}

public function render() {
  return view('livewire.foo', ['title' => $this->user->name]);
}
```
(Now all views (including the layout) would receive this data)

However, there are implementation problems with that.

I WILL say, a current way to achieve this behavior (I believe) is using View::share() from within the `->render()` method:

```php
public function render() {
  View::share('title', $this->user->name);

  ...
}
```

This is my currently recommended solution to the problem.

I understand the need and desire, but there is no easy solution to this. Thoughts?

vittorioe:
You’re totally right! Let me cry a bit, close the PR and rethink the solution! 😜🤔

booni3:
I have been using the `View::share()` to get the title show on the pages. A word of warning with this method - when using this, it will share that variable to ALL on the page. i.e. if you are using laravel components such as;

```blade
<x-panel :title="$title">
Some panel conent
</x-panel>
```

You page title will be shared to all components.

PR #1030, opened 2020-05-29
Add assertCount() to testing options

JayBizzle:

We need it in a lot of tests in our project.

Nope

Not required

We have a lot of public properties on our components that are arrays. It is useful to be able to assert the count of these.

One problem we have come across though is if you have a property that contains an `EloquentCollection`.

If for example you had the following code...

```php
class ExampleComponent extends Component
{
    public $users;

    public function mount()
    {
        $this->users = Users:limit(5)->get();
	}
}
```

Then this test will not pass...

```php
Livewire::test(ExampleComponent::class)
    ->assertCount('users', 5);
```

To get it to pass you have to change the test slightly to the following (notice `users.id`)

```php
Livewire::test(ExampleComponent::class)
    ->assertCount('users.id', 5);
```

As you know, this is because internally, that `EloquentCollection` is stored like this...

```
"users" => array:4 [
    "class" => "App\Users"
    "id" => [1, 2, 3, 4, 5]
    "relations" => []
    "connection" => "mysql"
  ]
```

Not sure if there is a way around this?

There are a few other assertions that we could do with. Would you be open to adding more?

Thanks 👍

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -23,6 +23,13 @@ trait MakesAssertions
 
         return $this;
     }
+    
+    public function assertCount($name, $value)
+    {
+        PHPUnit::assertCount($value, $this->get($name));
+
+        return $this;
+    }
 
     public function assertSee($value)
     {


calebporzio:
Going to hold off on this for now.

Here are better ways to make the assertions you are looking for:

image

PR #1043, opened 2020-06-01
strip wire:key from params

### LINKED ISSUE #1042: Using wire:key without mount method defined
### Description
When livewire component doesn't have mount function it can't accept wire:key assignment
### Steps to reproduce
`<livewire:app.product.import wire:key="import-products-{{Str::random(10)}}"/>`
```
<?php

namespace App\Http\Livewire\App\Product;

use Livewire\Component;

class Import extends Component
{
    public $modalOpen = 'false';

    protected $listeners = ['importProductsModalOpen'];

    public function importProductsModalOpen()
    {
        $this->modalOpen = 'true';
    }

    public function render()
    {
        return view('livewire.app.product.import');
    }
}
```
**getting the following error:**
```
Livewire\Exceptions\MountMethodMissingException
Livewire encountered a missing mount method when trying to initialise the [app.product.import] component. When passing component parameters, make sure you have a mount method.
```

**Context**
- Livewire version: 1.1.0
- Laravel version: 7.13.0

nuernbergerA:
Given a component without any parameters.

```blade
<livewire:app.product.import wire:key="import-products-{{Str::random(10)}}"/>
```
Would force the component to have a `mount` method defined.

As of my knowledge, we can safely remove `wire:key` before trying to `mount` the component, since its used only to help the dom finding the element again.

---
Closes #1042

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -83,8 +83,9 @@ class LivewireManager
 
     public function mount($name, $params = [])
     {
-        // This is if a user doesn't pass params, BUT passes key() as the second argument.
-        if (is_string($params)) $params = [];
+        $params = is_array($params) ?
+            $this->stripParams($params):
+            []; // This is if a user doesn't pass params, BUT passes key() as the second argument.
 
         $id = Str::random(20);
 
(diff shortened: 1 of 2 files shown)


mblarsen:
I'm not sure if the docs are wrong or out of date, but when used on `<livewire:comp/>` you should use `key` not `wire:key`.

https://laravel-livewire.com/docs/troubleshooting

nuernbergerA:
@mblarsen you are totally right.

mblarsen:
So when you have only `<livewire key=foo"/>` it works without this PR? @nuernbergerA

nuernbergerA:
Yes
```blade
<livewire key="foo" />
```
or

```blade
<livewire :key="$foo->id" />
```

mblarsen:
Yeah, I ran into this issue when trying to use `livewire/sortable`. The example uses `wire:key` on regular tags, but I was using nested components thus having to use `key`. A bit confusing and terrible error messages.

Will make a PR to "clean up" the examples there.

PR #1045, opened 2020-06-02
test: avoid test contamination in lifecycle hooks

mblarsen:
This PR splits out `mount_hook` test into multiple tests so that each concern is tested in isolation.

Ideally the hook calls should be tested with mocks, but after playing around with it a bit I realize that may be hard in the case of Livewire where reflection is used heavily.

It is really hard to extend and even understand the logic of the tests with all those `||` in the component hooks. 

**UPDATE**: refactored the tests so that each test defines they own assertions, effectively getting rid of the `||` issue. (in e1c4576)

> For reviewer. Check each of the two commits individually as the 'Files changes' makes the changes look more extensive than they are.

### DIFF
--- a/tests/LifecycleHooksTest.php
+++ b/tests/LifecycleHooksTest.php
@@ -6,6 +6,12 @@ use Livewire\Component;
 use Livewire\LivewireManager;
 use PHPUnit\Framework\Assert as PHPUnit;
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

PR #1049, opened 2020-06-03
fix: onNodeAdded script not executed

phamquocbuu:
execute all script tag, include all children tags


calebporzio:
Will there be any problems with double script execution? I forget how onNodeAdded works, but I thought that it would run on the parent node and then on all the children nodes, but I could be wrong - mind looking into this?

cosmospham:
I tracked by `console.log(node)` before they are `eval`ed. There is no child script tag. If the script tag is the node which will be eval, then the code return there.

```
if (node.tagName.toLowerCase() === 'script') {
                    eval(node.innerHTML)
                    return false
                }

                let scripts = node.querySelectorAll('script');
                if (scripts && scripts.length) {
                    scripts.forEach(function(el, i) {
                        eval(el.innerHTML);
                    });
                }
```

cosmospham:
For testing, I have a small project here. With this [commit](https://github.com/cosmospham/laravel-test/commit/23ba651d9620fa53bc380291df62dfd88724039e), script cannot be executed if script stay inside a component. Koolreport works like this.

In this [commit](https://github.com/cosmospham/laravel-test/commit/95839d578d7e17a803b8984af51df210dbde9188), script tag stay alone in a component, so it can be executed.

Sometime we cannot manage where to put script tags.

calebporzio:
Thanks for the PR, I believe this has been fixed in this recently merged PR: https://github.com/livewire/livewire/pull/1179

PR #1051, opened 2020-06-03
Correctly detect vapor

### LINKED ISSUE #1050: Laravel Vapor no longer correctly detected
Livewire tries to detect Vapor runtimes using `($_ENV['SERVER_SOFTWARE'] ?? null) === 'vapor'`. This doesn't work anymore.

It is now (only) available as `$_SERVER['SERVER_SOFTWARE']`.

This effectively means that users will see a 500 page when they try to load something that contains a Livewire component.

Sounds to me Vapor changed something in the way they inject these variables. I've contacted them to verify this is the case.

georgeboot:
Fixes #1050 

It checks both `$_ENV` and `$_SERVER` when detecting for Vapor.

Theoretically we could also remove the `$_ENV` check and only use `$_SERVER` but if the Vapor team ever decides to revert their (presumably) change, it will break again.

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -219,9 +219,7 @@ HTML;
             $publishedManifest = json_decode(file_get_contents(public_path('vendor/livewire/manifest.json')), true);
             $versionedFileName = $publishedManifest['/livewire.js'];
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Great, thanks!

PR #1053, opened 2020-06-03
Rename /livewire-tmp back to /tmp for Vapor

georgeboot:
This PR renames the manifest temp directory for Vapor users, back to `/tmp`, since that is the only writable directory.

This was changed in the recent file upload PR, specifically in commit 83a5a21.

This PR also reverts the changes made in #1051 since that was not the underlying issue. 
Appologies for that, jumpted too quickly to conclusions (and the `dd()` search thing bugs me every time).)

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -320,7 +320,7 @@ HTML;
 
     public function isOnVapor()
     {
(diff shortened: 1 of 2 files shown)


calebporzio:
Ooooh good catch, that was a regression from a find/replace all. Thanks!

Issue #1000, opened 2020-05-22
Route parameter casing does not match Laravel behaviour


adevade:
### Description

Laravel automagically converts route parameters between different casings _(snake, camel)_, but Livewire does not.

### Steps to reproduce

Create some routes: 

```php
// Both of these work
Route::get('/laravel-snake/{big_fish}', function ($bigFish) { return $bigFish; });
Route::get('/laravel-camel/{bigFish}', function ($big_fish) { return $big_fish; });

// Only the first one works
Route::livewire('/livewire-snake/{big_fish}', 'big-fish');
Route::livewire('/livewire-camel/{bigFish}', 'big-fish');
```
_Notice the casing on the parameters_

And catch the parameter in the Livewire component:

```php
namespace App\Http\Livewire;

use Livewire\Component;

class BigFish extends Component
{
    public function mount($big_fish)
    {
        return $big_fish;
    }
}
```

**Context**
- Livewire version: [e.g. 1.1.0]
- Laravel version: [e.g. 7.11.0]

calebporzio:
Valid point.

However, we're going to stick with the current behavior for now.

I recommend making both parameters match. I believe that's reasonable given both the component and the route declaration are under the devs control.

Thanks!

PR #1058, opened 2020-06-05
Add route parameter support for camel case

armancodes:
This PR solves #1000 issue.

This is a livewire component class called `Foo`:
```
use Livewire\Component;

class Foo extends Component
{
   public $name;

   public function mount($first_name)
   {
        $this->name = $first_name;
   }

   public function render()
   {
        return view('foo');
   }
```

Before:
```
// Only the first one works
Route::livewire('/foo/{first_name}', 'foo');
Route::livewire('/foo/{firstName}', 'foo');
Route::livewire('/foo/{FirstName}', 'foo');
Route::livewire('/foo/{first-name}', 'foo');
```

After:
```
// All of them work
Route::livewire('/foo/{first_name}', 'foo');
Route::livewire('/foo/{firstName}', 'foo');
Route::livewire('/foo/{FirstName}', 'foo');
Route::livewire('/foo/{first-name}', 'foo');
```

### DIFF
--- a/src/DependencyResolverTrait.php
+++ b/src/DependencyResolverTrait.php
@@ -2,6 +2,7 @@
 
 namespace Livewire;
 
+use Illuminate\Support\Str;
 use ReflectionMethod;
 use ReflectionParameter;
 use ReflectionFunctionAbstract;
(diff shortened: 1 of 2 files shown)


calebporzio:
Thank you for submitting this. However, we're not going to merge this at this time.

I would recommend changing the casings in both to match each other.

PR #1066, opened 2020-06-06
Validation not working with cast properties

morrislaptop:

Needed. This is an issue in the form of a PR with tests to help solve the issue. 

No

Yes, but only tests! I'm not sure on the solution

Laravel's validation only works with primitive types with the exception of DateTimeInterface. However, the components properties are cast into collections / dates / custom objects BEFORE we have the chance to do validation, so Laravel's validation will not work properly (tests attached)

Validation should be performed on the uncast values rather than the cast values. 

As far as I can tell, the cast process completely overwrites the uncast values so I'm not sure on the best approach for this. 

### DIFF
--- a/tests/ValidationTest.php
+++ b/tests/ValidationTest.php
@@ -55,7 +55,7 @@ class ValidationTest extends TestCase
         $component
             ->set('date', 'tomorrow')
             ->runAction('runValidationWithDateCastAttribute')
-            ->assertHasErrors('date', 'before');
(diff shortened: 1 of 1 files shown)


calebporzio:
Yes, in that case, I recommend storing a raw date string or collection array as a public property for validation, and then using a computed property to convert the primitive to a DateTime or Collection. Closing for now, thanks!

PR #1068, opened 2020-06-06
Fully implement route model binding for components

### LINKED ISSUE #1067: Livewire::test() causes an exception under certain conditions
### Description
When testing, calling `Livewire::test()` for a component that has a `mount()` method after calling a request testing method (e.g. `$this->get('/foo')`) generates an exception
```php
ErrorException: trim() expects parameter 1 to be string, object given
```

### Steps to reproduce
Run the following tests:
```
use Illuminate\Foundation\Testing\RefreshDatabase;
use Livewire\Component;
use Livewire\Livewire;
use Tests\TestCase;

class LivewireRouteRegistrationTest extends TestCase
{
    protected function setUp() : void
    {
        parent::setUp();

        Livewire::component('foo', ComponentWithoutMount::class);
        Livewire::component('moo', ComponentWithMount::class);

        Route::livewire('/foo', 'foo');
        Route::livewire('/moo', 'moo');
    }

    public function testTestsWorkBeforeRoutes()
    {
        Livewire::test('foo')->assertSet('foo', 'bar');
        $this->get('/foo')->assertSuccessful();

        Livewire::test('moo')->assertSet('foo', 'bar');
        $this->get('/moo')->assertSuccessful();
    }

    public function testTestWorksAfterRouteForComponentWithoutMount()
    {
        $this->get('/foo')->assertSuccessful();
        Livewire::test('foo')->assertSet('foo', 'bar');
    }

    public function testTestWorksAfterRouteForComponentWithMount()
    {
        $this->get('/moo')->assertSuccessful();

         /* this test causes an exception */
        Livewire::test('moo')->assertSet('foo', 'bar');
    }
}

class ComponentWithoutMount extends Component
{
    public $foo = 'bar';

    public function render()
    {
        return '<div>Foo is {{ $foo }}</div>';
    }
}

class ComponentWithMount extends Component
{
    public $foo = '';

    public function mount()
    {
        $this->foo = 'bar';
    }

    public function render()
    {
        return '<div>Foo is {{ $foo }}</div>';
    }
}
```

**Context**
- Livewire version: 1.1
- Laravel version: 7.0 
- Browser: Chrome

### Analysis
The current implementation of the `Route::livewire()` macro causes side effects. **PR to follow**.

mattharcourt:
This fixes #1067 

The `Route::livewire()` route registration macro causes side effects when it retrieves the route model bindings for the component's `mount()` method. 

This occurs in the [`retrieveBindings()`](https://github.com/livewire/livewire/compare/master...mattharcourt:route-livewire-issue?expand=1#diff-55630020df0f5f2f2e4a31f156876a9a) method of the `Macro\PretendClassMethodIsControllerMethod` class when the `'uses'` property of the route's action is temporarily replaced to enable binding to the `mount()` method.

However, when the router's `substitueImplicitBindings()` is called, the macro's `Closure` ends up being assigned as the `'controller'` property of the route's action, and this property doesn't get restored when restoring the `'uses'` property of the route's action.

This PR makes a change so that the entire route action is cached/restored.

The [`tests/RouteRegistrationTest.php`](https://github.com/livewire/livewire/compare/master...mattharcourt:route-livewire-issue?expand=1#diff-56f30375ee57f1e2c362514058f7ae96) test was extended to test for this issue.

The reason this issue only appears when calling `Livewire::test()` after calling a request/response testing method is because the mount bindings occur in the callback registered by the `Route::livewire()` macro, and the callback is only called when the route is requested. Then the `Livewire::test()` function attempts to insert a test route which adds all routes to a new route collection and that's where the inconsistent  `'controller`' property of the action causes an error (see [`Testing/TestableLivewire.php`](https://github.com/livewire/livewire/blob/07fa6212a0a456fd419866a31678ff464daffebb/src/Testing/TestableLivewire.php#L104)).


mattharcourt:
After digging deeper into route model binding, the PR has been updated to provide components the same route model binding that is done for controllers. Including
- explicit route binding (as before) - e.g. `Route::bind()`
- implicit route model binding - e.g. `Route::livewire('/post/{id}', 'post')`
- custom keys & scoping of related models appearing on a route (Laravel 7.0)

Tests for route binding were split into a separate file from the method binding tests.

nuernbergerA:
@mattharcourt you merged the wrong branch into yours.
image

Make sure to revert to be70653 and merge `1.x`

mattharcourt:
@nuernbergerA - I reverted to be70653, but not sure I did it correctly - any guidance would be appreciated.

mattharcourt:
Still fixes original issue, but with change in directions and mistaken merge this has gotten a bit messy.

Will resubmit a cleaner PR.

PR #1070, opened 2020-06-06
Make upload directory formatting consistent

danharrin:
Format is now consistently `directory` instead of `directory/`, as described [here](https://github.com/livewire/livewire/blob/master/config/livewire.php#L75).


calebporzio:
Closing this for now, thanks though @DanHarrin

PR #1072, opened 2020-06-07
Format the file upload rules as an array

jayhealey:
# Description
This PR changes Livewire so that file upload validation is formatted using the array format of rules, instead of a concatenated string. 

## Reasoning
Laravel offers two ways to forma validation rules.

 As pipe separated string:
- `'rules' => 'file|mimes:png,jpg', `

Or as an array:
- `'rules' => ['file' , 'mimes:png,jpg']`

Livewire's default configuration relies on it being a string as it prepends it with a `required` before using the validation. 

If you customise is to be an array, `FileUploadHandler::validateAndStore` throws an exception as it's attempting to merge the string of `required` with the array from the configuration.

In my use case, I need to add a custom rule which can only be added via the array format. This offers more flexibility when customising file validation, for example:

- `'rules' =>  ['required', 'file', new \App\Rules\ValidProductFile()]`, 

This PR changes Livewire to given the option of both - but defaults to array to imply more flexibility is available.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -70,10 +70,10 @@ return [
     */
 
     'temporary_file_upload' => [
-        'disk' => null,        // Example: 'local', 's3'         Default: 'default'
-        'rules' => null,       // Example: 'file|mimes:png,jpg'  Default: 'file|max:12288' (12MB)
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks!

Issue #1074, opened 2020-06-07
TypeError: Echo.join(...)[event_name] is not a function


rainaldoc:
### Description
When setting up a listener for a presence channel in livewire, it throws the error below
![image](https://user-images.githubusercontent.com/66362893/83973238-9b33ee80-a8aa-11ea-9e37-a3effbc8fdfb.png)

### Steps to reproduce

1) Create a [laravel](https://laravel.com/docs/7.x/installation#installing-laravel) app
2) Install laravel [echo](https://laravel.com/docs/7.x/broadcasting#installing-laravel-echo)
3) Install [livewire](https://laravel-livewire.com/docs/installation)
4) Create a livewire [component](https://laravel-livewire.com/docs/making-components)
5) Set up the listener array to listen on presence channel
`protected array $listeners = [
        'echo-presence:chatRoom,PlayerJoined' => 'notifyPlayerJoined',
    ];
    public function notifyPlayerJoined()
    {
        //do something interesting
    }
`
6) Add component to [routes](https://laravel-livewire.com/docs/rendering-components#route-registration) 
4) Browse to the component
5) View the console.

**Context**
- Livewire version: 1.1.0
- Laravel version: 7.0.0
- Browser: Chrome

calebporzio:
Closing due to inactivity

redbastie:
Did you figure it out? I'm getting the same error from this code:

    protected $listeners = [
        'echo-presence:lobby,MatchMade' => 'matchMade',
    ];

king724:
I am also getting this error, fresh install of Laravel 8.x and Livewire 2.x

redbastie:
Yeah I just ended up using the JavaScript code and having it emit events to the components.

guoxiangke:
presence report this error while private is works.

PR #1075, opened 2020-06-07
Update the event registration according to laravel-echo

rainaldoc:
Needed for applications using Laravel Echo. I created an [issue](https://github.com/livewire/livewire/issues/1074) outlining the problem

No

No tests

It is a change to the event registration in the EchoManager.js for presence channels. See the image below on how to register events on a presence channel taken from the [laravel](https://laravel.com/docs/master/broadcasting#broadcasting-to-presence-channels) doc

image


jonfackrell:
I ran into this same issue today, but I don't think this is the appropriate fix. I think the other methods (here, joining, and leaving) need to be considered as well.

larasocket:
Also running into this issue. Agreed with @jonfackrell, we should try and find a solution that can account for join, here, leaving, as well as custom events.

Could we add a simple switch to check if `event_name` is "here", "joining", "leaving" to call the methods as it currently does. If not, listen to the event like you suggest in your pull request?

I would love to see this fixed :) Let me know if I can help.

Does @calebporzio have any suggestions on the best fix?

calebporzio:
As it seems this isn't the appropriate fix, I'm going to close this.

Feel free to re-submit a PR or issue with a more thorough description of the problem that's easier for me and others to understand, ideally with sample, copy-pastable code.

Thanks!

redbastie:
Did you figure it out? I'm getting the same error from this code:

    protected $listeners = [
        'echo-presence:lobby,MatchMade' => 'matchMade',
    ];

PR #1079, opened 2020-06-08
[WIP] Support .pushState instead of .replaceState for `$updatesQueryString`

calebporzio:
### Ideal Magic Behavior
This component is shown on the `/component` route:
```php
class Foo extends Component
{
  public $foo = 'bar';

  protected $queryStringBindings = ['foo'];

  public function render()
  {
    return view('livewire.foo');
  }
}
```
* When the page loads, the URL goes from `/component` to `/component?foo=bar`
* When the `foo` property is changed to "baz", the URL is also changed to `/component?foo=baz`
* When the user hits the back button, the `$foo` property is reverted back to "bar" and the component will update with a server roundtrip, the forward button would set it back to "baz" -> all without actually changing pages.
* When the page loads from scratch, with the URL `/component?foo=bob`, then the `$foo` property will be set to `"bob"` initially.

In addition, you could opt-out of the magic "sync" behavior, and utilize manual methods:

```php
class Foo extends Component
{
  public $foo = 'bar';

  public function mount()
  {
    $this->foo = request()->query('foo', $this->foo);
  }

  public function updatedFoo()
  {
    $this->pushState('foo', $this->foo);
    // or...
    $this->replaceState('foo', $this->foo);
  }
```

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -69,12 +69,12 @@ return [
     |
     */
 
-    'temporary_file_upload' => [
-        'disk' => null,        // Example: 'local', 's3'              Default: 'default'
-        'rules' => null,       // Example: '['file', 'mimes:png,jpg'] Default: ['required', 'file', 'max:12288'] (12MB)
-        'directory' => null,   // Example: 'tmp'                      Default  'livewire-tmp'
-        'middleware' => null,  // Example: 'throttle:5,1'             Default: 'throttle:60,1'
-    ],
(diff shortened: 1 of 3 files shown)


acurrieclark:
Hi @calebporzio, would love to see this feature added. Anything I can do to assist?

calebporzio:
Thanks! But I think I'm good for now. Hopefully will be in the next release

PR #1099, opened 2020-06-13
created a fix for sorted/associative eloquent collections which cause…

### LINKED ISSUE #987: some collections evoke errors when passed as parameter to a component
### Description

In my `home.blade.php` I use a Livewire directive and need to pass two parameters like so:

`@livewire('price-table', ['table' => $table, 'local_country' => $local_country]) `

Both parameters are collections. 

When I use a drop down in my Livewire component on the front end I get the following error: 

```
Livewire\Exceptions\CorruptComponentPayloadException
Livewire encountered corrupt data when trying to hydrate the [price-table] component. Ensure that the [name, id, data] of the Livewire component wasn't tampered with between requests.
```

If I only pass `'local_country' => $local_country` as a parameter all works fine. 

If I only pass `'table' => $table` I get the error. 

When I create the table collection within the Livewire component itself, then all works fine too. 
But, since I need the `$table` in several parts of my page, I need to generate it in a controller and then pass it down to Livewire and other blade components. 

The only thing that is structurally different between $table and $local_country is that $table also contains relations. 

The problem only starts when I assign the table to a public variable with the mount method.

This evokes the error: 

```
public 	$table_y;
public 	$local_country_x;

public function mount($table, $local_country)
{
	$this->table_y = $table;
	$this->local_country_x = $local_country;
}
```

This works fine:

```
public 	$table_y;
public 	$local_country_x;

public function mount($table, $local_country)
{
	// $this->table_y = $table;
	$this->local_country_x = $local_country;
}
```

**Context**
- Livewire version: 1.1.0
- Laravel version: 7.11.0
- Browser: Chrome, Safari

jonneroelofs:
fixes #987

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -146,8 +146,9 @@ abstract class Component
             if (is_array($value)) {
                 $this->$key = $this->reindexArrayWithNumericKeysOtherwiseJavaScriptWillMessWithTheOrder($value);
             }
-            if(is_object($value) && is_a($value,Collection::class)){
-                $this->$key = $value -> values();
+            
+            if (is_object($value) && is_a($value, Collection::class)) {
+                $this->$key = $value->values();
             }
         }
     }
--- a/src/Component.php
+++ b/src/Component.php
@@ -8,6 +8,7 @@ use BadMethodCallException;
 use Illuminate\Support\Str;
 use Illuminate\Support\ViewErrorBag;
 use Illuminate\Support\Traits\Macroable;
+use Illuminate\Database\Eloquent\Collection;
 use Livewire\Exceptions\CannotUseReservedLivewireComponentProperties;
 
 abstract class Component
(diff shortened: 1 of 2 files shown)


src/Component.php
```diff
@@ -145,6 +146,9 @@ public function normalizePublicPropertiesForJavaScript()
             if (is_array($value)) {
                 $this->$key = $this->reindexArrayWithNumericKeysOtherwiseJavaScriptWillMessWithTheOrder($value);
             }
+            if(is_object($value) && is_a($value,Collection::class)){
+                $this->$key = $value -> values();
+            }
```
lancepioch, line 151:
```suggestion
            }
            
            if (is_object($value) && is_a($value, Collection::class)) {
                $this->$key = $value->values();
            }
```

jonneroelofs:
@lancepioch thanks for your help
Out of curiosity, do you have any clue why I had a failing test before? All tests were passing in my local enviroment with php 7.2 , 7.3 , laravel 6  and laravel 7.
The failing test 'NestingComponentsTest' seems unrelated to both my PR and your changes, but seems to be working after you change.

PR #1100, opened 2020-06-14
Fixed a typo which was causing filesystem default driver recognition …

merthurturk:
…issues

This is a must-have fix for file system driver recognition to work properly. Without this fix, S3 file previews are not working at all. I haven't created an issue because it's a very straightforward fix.

No.

No, but really surprised the existing tests didn't catch this.

The uploads with S3 driver is working but preview temporary url generation isn't working because of this issue. Due to this problem, Livewire can't detect that the app is using an S3 driver for its configured filesystem disk.

Thank you for building this beautiful thing 👍

### DIFF
--- a/src/FileUploadConfiguration.php
+++ b/src/FileUploadConfiguration.php
@@ -37,7 +37,7 @@ class FileUploadConfiguration
 
     public static function isUsingS3()
     {
-        $diskBeforeTestFake = config('livewire.temporary_file_upload.disk') ?: config('filsystems.default');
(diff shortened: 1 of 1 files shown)


merthurturk:
Watching the screencast (https://laravel-livewire.com/screencasts/s5-upload-previews) and seeing how it works very smooth but the same thing failing in the local instance was a crazy moment... The good news is that the fix was pretty easy :)

nuernbergerA (approved):

nuernbergerA:
Good catch @merthurturk :+1:

merthurturk:
Accidentally closed the pull request, but reopened it immediately. Sorry 🤭

Issue #1112, opened 2020-06-16
Internet Explorer 11 support 


ChrisSantiago82:
The Internet Explorer support is pretty good. But I' having a problem when I'm working with listeners.
I'm getting a Syntax error in the directive.js file.

I'm not a Javascript expert but I think the Arrow Syntax is not supported by the IE. Could there be another way to write that method that would also work for the ie?

![image](https://user-images.githubusercontent.com/43000085/84784395-3d915780-afb8-11ea-9331-13a48cb252ba.png)

**Context**
- Livewire version: [e.g. 1.2.0]
- Laravel version: [e.g. 6.2.0]
- Browser: [IE11]

nuernbergerA:
Relates to #999

PR #1124, opened 2020-06-18
Replace ES6 code with traditional js for parsing action expressions

mattharcourt:
This addresses #1112 by replacing the ES6 code with traditional javascript.

Will admit to not doing extensive browser testing, but this should resolve the issue.

### DIFF
--- a/js/dom/directive.js
+++ b/js/dom/directive.js
@@ -52,7 +52,13 @@ export default class {
             const $event = this.eventContext
             method = methodAndParamString[1]
             // use a function that returns it's arguments to parse and eval all params
-            params = eval('((...params) => [...params])(' + methodAndParamString[2] + ')')
+            getParams = `function (${methodAndParamString[2]}) {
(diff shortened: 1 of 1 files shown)


calebporzio:
At first glance it doesn't look like this code will work because the function is never being called.

You'll notice the original snippet in the `eval()` is a function that calls itself. You are replacing it with a function definition.

Did you test that this works?

I'm going to close this for now. Feel free to re-submit after you've tested it in both a current browser, and IE. Thanks!

mattharcourt:
I know it didn't look like it should work, but it did - passed tests and worked in Chrome (don't have IE so didn't test there).

I updated so that the dynamically defined function is explicitly called with the method params. Still passed tests and works in Chrome.

nuernbergerA:
@mattharcourt make sure to commit also your `dist`
`jest` tests against `dist` so you have to run `npm run build` and `npm run test`

mattharcourt:
@nuernbergerA thanks for the help! - I updated as suggested.

PR #1132, opened 2020-06-19
Fixed typo in isUsingS3 function

xitox97:
Before this PR, 
if `livewire.temporary_file_upload.disk` is null it will set `$diskBeforeTestFake` to null also due to the typo of `config('filsystems.default')` instead of `config('filesystems.default')`


xitox97:
my bad this PR already merged #1100

PR #1137, opened 2020-06-20
Laravel 5.6 and 5.7 eloquent collection hydration compatibility

jonneroelofs:
The original fix from #1099 was not working correctly for Laravel 5.6 and 5.7 because Laravel was restoring collections differently in those version.

This fix copies the behaviour (and some of the code) that was introduced in Laravel 5.8 in SerializesAndRestoresModelIdentifiers.php for the restoreCollection method.


calebporzio:
Wow, thanks for digging into this @jonneroelofs, this couldn't have been easy.

Anyhow, instead of just patching in the fix and running it twice on Laravel > 5.7 systems (which are the majority).

I think it's clearer to just include the newer method as a way of overriding the faulty trait method.

Merging this now. Thanks again!

PR #1146, opened 2020-06-22
Improve handling of 419 errors (session expired)

calebporzio:
Currently, when a users' session is expired due to inactivity, a Livewire component will show the following when interacted with:
image

This modal is unhelpful and cryptic to users.

This PR offers 2 things:
1. A better default behavior
2. A JS API for extending and customizing this behavior

**1. Better Default:**
image
A native `confirm()` message will be shown to the user allowing them to refresh or cancel (continue using the expired page)

**2. JS API**
Alternatively, this PR introduces the following method:
```js
window.livewire.onError(statusCode => {
    //
})
```

Currently, this callback will get executed for all error status codes (500s, 400s) and BEFORE the `confirm()` message is shown.

If you wish to prevent the default behavior for an error, you can return `false` explicitly from this callback.

Here's an example of overriding the current 419 behavior using this method:
```js
window.livewire.onError(statusCode => {
    if (statusCode === 419) {
        alert('Your own message')

        return false
    }
})
```


lucasromanojf:
Hey @calebporzio thanks for this PR! I was checking it out and I think that it is better to pass the entire response to the onError callback instead of the statusCode only. The reason is that sometimes a different message is returned from backend for a same status code (e.g.: a JSON response with an error message that varies). So we can't show the correct message since we just have the statusCode. I can PR this if it is ok.

jjjrmy:
> Hey @calebporzio thanks for this PR! I was checking it out and I think that it is better to pass the entire response to the onError callback instead of the statusCode only. The reason is that sometimes a different message is returned from backend for a same status code (e.g.: a JSON response with an error message that varies). So we can't show the correct message since we just have the statusCode. I can PR this if it is ok.

Please please please 🙏

jhoff:
Third. It would be nice to have full interceptor functionality, similar to [Axios](https://github.com/axios/axios#interceptors).

andcl:
> Hey @calebporzio thanks for this PR! I was checking it out and I think that it is better to pass the entire response to the onError callback instead of the statusCode only. The reason is that sometimes a different message is returned from backend for a same status code (e.g.: a JSON response with an error message that varies). So we can't show the correct message since we just have the statusCode. I can PR this if it is ok.

I also think this proposed functionality would be a smart replacement for the default error handling with modals in Livewire (for any kind of HTTP errors). What is the current situation about this topic? Was the PR added at the end? I haven't found anything after this and I consider worthy asking.
Thanks!

denfield-matt:
Am I right in saying if you did something like:

```
window.livewire.onError(statusCode => {
    return false
})
```

It would stop the error popups all together? We are having loads of issues for about 1-2% of users that any of our components are throwing hydration errors. As a lot of these are utility components we'd rather just stop the popups whilst we work it out.

TonyPartridgeR:
> Am I right in saying if you did something like:
> 
> ```
> window.livewire.onError(statusCode => {
>     return false
> })
> ```
> 
> It would stop the error popups all together? We are having loads of issues for about 1-2% of users that any of our components are throwing hydration errors. As a lot of these are utility components we'd rather just stop the popups whilst we work it out.

yes

Baspa:
Nowhere in the docs I can find how to redirect the user to a page instead of showing the error in a modal? How can I do that? @calebporzio

PR #1149, opened 2020-06-22
remove default label

nuernbergerA:
remove default labeling for new issues


calebporzio:
Thanks @nuernbergerA!

PR #1150, opened 2020-06-22
Eliminate route registration side effects

### LINKED ISSUE #1067: Livewire::test() causes an exception under certain conditions

mattharcourt:
This fixes #1067

This PR provides components the same route model binding that is done for controllers.  Including
- explicit route binding (as before) - e.g. `Route::bind()`
- implicit route model binding - e.g. `Route::livewire('/post/{id}', 'post')`
- custom keys & scoping of related models appearing on a route (Laravel 7.0) - e.g. `Route::livewire('/post/{post:id}/comment/{comment:id}')`

The issue occurs in the `Route::livewire()` route registration macro which causes side effects while retrieving the route model bindings for the component's `mount()` method  in `Macro\PretendClassMethodIsControllerMethod::retrieveBindings()` when the 'uses' property of the route's action is temporarily replaced to enable the binding. 

This PR replaces (renames)  `PretendClassMethodIsControllerMethod` and clones the `Route` to avoid side effects and relies more fully on Laravel's route-binding logic.

Tests are included - tests for route binding were split into a separate file from the method binding tests.


mattharcourt:
Discovered an issue with camel case route parameters. 

Laravel tries to match camel case route parameters with snake case method names, but some of this occurs in the way the request is dispatched to the controller method which wasn't replicated here since we're not dispatching to a controller.

Changes were added so that camel case and snake case differences between route parameters and component mount parameters are considered equivalent. Tests were also added to verify behavior matches Laravel.

As a result this should also address #1000.

calebporzio:
Much of this has been added to Livewire V2. Thanks, but closing.

PR #1152, opened 2020-06-22
Follow S3 disk root configuration

### LINKED ISSUE #1125: File uploader doesn't respect disk root
### Description

The file uploader tells the browser to upload to `/` + the `livewire-tmp/` dir, regardless of what the root disk is.

### Steps to reproduce

Tell Livewire to use the `s3` disk, set `root` in the disk to `/foo`, open dev tools when uploading a file and see that it uses just `/livewire-tmp/` rather than `/foo/livewire-tmp/`.

**Context**
- Laravel version: [e.g. 7.0.0]
- Browser: Chrome

danharrin:
Closes #1125.

- [x] Permanent S3 file uploads respect the disk root.
- [x] Temporary S3 file uploads respect the disk root.
- [x] Temporary S3 files are automatically cleaned up after 24 hours.

### DIFF
--- a/src/Commands/S3CleanupCommand.php
+++ b/src/Commands/S3CleanupCommand.php
@@ -21,7 +21,7 @@ class S3CleanupCommand extends Command
 
         $diskConfig = FileUploadConfiguration::diskConfig();
         $bucket = $diskConfig['bucket'];
-        $prefix = FileUploadConfiguration::directory();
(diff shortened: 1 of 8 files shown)


calebporzio:
Thanks @DanHarrin! Is it possible to add a test? (for local roots at least)

PR #1153, opened 2020-06-23
Use response()->make to work on Laravel 5.6

mpociot:
The `noContent` method was only added in Laravel 5.7 (see https://github.com/laravel/framework/pull/23744/files).
But since Livewire supports Laravel 5.6 as well, this was causing a MethodNotFoundException.

This PR simply uses the `response()->make` method with an empty content, as this method always exists.


nuernbergerA (approved):
Good catch @mpociot :+1:

lancepioch (approved):

calebporzio:
Thanks @mpociot!

PR #1154, opened 2020-06-23
Fix for wire:loading with a redirectTo

dubcanada:
Hello,

If you have a wire:loading with a button that does the recommended download (redirectTo a route with a laravel file download controller) it downloads the file successfully but the loading does not go away.

This is due to the fact most redirectTo send you to a new URL and thus the loading is removed. But a download will keep you on the same page while your browser downloads the file.

To fix this we need to call the `responseReceived` store hook before we handle the redirectTo.

Let me know if there is a better way of handling this.


nuernbergerA:
Please make sure to also compile assets and comit the dist files.

dubcanada:
@nuernbergerA - my apologies first time contributing.

I have added the compiled assets!

calebporzio:
Thanks @dubcanada !

Issue #420, opened 2019-12-17
Pagination with tag a sometime not working ???


doanhung1996:
**Describe the bug**
I'm using livewire with custom pagination tag a some page not working after click .

**To Reproduce**
1. Create custom pagination view with tag a to change page .
2. paginationView() set return section 1 view. 
3. Open browser click change page 
4. Some page not change after click .

**Expected behavior**
The page change correct.

**Desktop (please complete the following information):**
 - Browser chorme 
   Os: Ubuntu 18.04

calebporzio:
Hey @doanhung1996 - please include code samples so we can help you better:

Specifically the following:
* Your custom pagination view
* The code you use to swap it in instead of the default

doanhung1996:
Here is my pagination

```
@if ($paginator->hasPages())
    <ul class="options-middle tt-paginate">
        @if ($paginator->onFirstPage())
            <li class="disabled">
                <a href="javascript:void(0)">&lsaquo;</a>
            </li>
        @else
            <li class=""><a href="javascript:void(0)" wire:click="previousPage" aria-label="@lang('pagination.previous')">&lsaquo;</a></li>
        @endif

        @foreach ($elements as $element)
            {{-- "Three Dots" Separator --}}
            @if (is_string($element))
                <li class="">
                    <a href="javascript:void(0)">{{ $element }}</a>
                </li>
            @endif

            {{-- Array Of Links --}}
            @if (is_array($element))
                @foreach ($element as $page => $url)
                    @if ($page == $paginator->currentPage())
                        <li class="active"><a href="javascript:void(0)">{{ $page }}</a></li>
                    @else
                        <li class=""><a href="javascript:void(0)" wire:click="gotoPage({{ $page }})">{{ $page }}</a></li>
                    @endif
                @endforeach
            @endif
        @endforeach

        @if ($paginator->hasMorePages())
            <li class=""><a href="javascript:void(0)" wire:click="nextPage" aria-label="@lang('pagination.next')">&rsaquo;</a></li>
        @else
            <li class="disabled"><a href="javascript:void(0)">&rsaquo;</a></li>
        @endif
    </ul>
@endif
```

vanthao03596:
i have same issue @doanhung1996

JellyBool:
Same issue here.

linuxstreet:
The Livewire pagination is not working when using the default Laravel pagination views.

For example:
{{ $data->links('pagination::default') }}

[...]

PR #1155, opened 2020-06-23
Wrap pagination view's root conditional in an element to prevent DOM diffing issues

calebporzio:
Issue #420 started this.

Because the pagination view for Livewire contains a root-level blade conditional, depending on where it's included in a component's Blade view, it MAY create DOM-Diffing issues for morphdom.

The fix is to ALWAYS have a wrapping <div> for this view's inclusion.


IsraH:
can you post an example of this wrapping?

PR #1160, opened 2020-06-25
adds support for s3 prefixes

shanerbaner82:
I believe this is needed, [Issue](https://github.com/livewire/livewire/discussions/1161)

No.

No.

This PR allows for your env to have an S3 prefix. Laravel vapor has reserved keys that Livewire is also looking for, by adding a prefix to .env we can use whatever name we want for S3 credentials.


shanerbaner82:
Adding [storage](https://docs.vapor.build/1.0/resources/storage.html#introduction) to vapor.yml fixed the issue

PR #1162, opened 2020-06-26
Replace action param parsing ES code with equivalent babel polyfill

### LINKED ISSUE #1112: Internet Explorer 11 support

mattharcourt:
This is a re-submission of #1124 to fixes #1112 

The ES code was replaced with what Babel produces for a polyfill. I also verified it works in latest versions of Chrome and Firefox, but I don't have IE so need some help to verify there.


dubcanada:
I tested it in IE 11 and it fixed the issue.

mgussekloo:
This would be perfect! Any chance of a timely merge?

jwohlfert23:
Tested this in IE11 and it does the trick! Any change to get this merged soon?

calebporzio:
Thanks for the PR!

jwohlfert23:
Any timeline for releasing v1.3.3?

Issue #716, opened 2020-03-14
Custom input readonly after validation error


azurinspire:
**Describe the bug**
Custom input component is left to readonly state after form submission if there is validation errors.

**To Reproduce**
1. Go to https://github.com/azurinspire/livewire-tel-input
2. clone, `composer install && php artisan serve`
3. Open in browser, click the green button and phone input is readonly

**Expected behavior**
Custom input is not readonly

**Desktop (please complete the following information):**
 - OSX + Firefox 73

**Additional context**
What I think happens: when custom component uses `wire:ignore` attribute, it is still set as `readonly` when form is submitting by this code: https://github.com/livewire/livewire/blob/73bd75b830005d0f412c7347cca6ff1b95229746/js/component/DisableForms.js#L31
But cleaning readonly attribute doesn't happen because the dom inside wire:ignore is not updated.

codeduction:
I also ran into this issue today, What is the best approach to resolve this?

tell the js to ignore any inputs that are wire:ignore'd ?

rodrigore:
Im also facing this issue using mapbox library. I try to use the geocoder plugin which contains an input which generated some custom dom.

In order to fix this problem I have to remove manually the readonly and other undesirable attributes with the afterDomUpdate hook.

I put this inside an script tag:

``` html
<div wire:ignore id="geocoder" class="geocoder"></div>
```


``` javascript
document.addEventListener("livewire:load", function(event) {
    window.livewire.hook('afterDomUpdate',  () => {
        document.getElementsByClassName('mapboxgl-ctrl-geocoder--input')[0].removeAttribute('readonly');
        document.getElementsByClassName('mapboxgl-ctrl-geocoder--button')[0].removeAttribute('disabled');
    });
});


```

I hope that this trick can help people like  @azurinspire @codeduction  and others.

calebporzio:
Thanks! This issue will be fixed in the next release!

PR #1166, opened 2020-06-26
Fix form disabling after errors

### LINKED ISSUE #624: Form does not get reenabled in case of submit error
**Describe the bug**
When we submit a form, the input fields and button gets disabled, but in case of exception on submit, the fields and button keep disabled.

**To Reproduce**
1. Create a form with inputs and submit button and use wire:submit on form
2. Throw an exception from within wire:submit method
3. Submit the form. Fields and button will be disabled
4. Fields won't be reenabled after submit error

**Expected behavior**
Fields and button should be reenabled in case of exception on form submit.

**Suggestion of fix** 
Inside **DisableForms.js**, add (I can open a PR if it's ok):

```
store.registerHook('messageFailed', component => {
    component.call('$refresh')
})
```
I did not test this change but tested the following directly in my project and it solved the issue:
```
window.livewire.hook('messageFailed', component => {
    component.call('$refresh')
})
```

calebporzio:
Fixes #624 & #716


js/component/DisableForms.js
```diff
@@ -21,16 +27,28 @@ export default function () {
...
                 }
             })
         })
     })
+
+    store.registerHook('messageFailed', (component, message) => {
+        if (! cleanupStackByComponentId[component.id]) return
+
+        while (cleanupStackByComponentId[component.id].length > 0) {
+            cleanupStackByComponentId[component.id].shift()()
```
asantibanez, line 51: Very elegant 🧐

PR #1169, opened 2020-06-27
Shows the error keys when a bare call to assertHasNoErrors() fails

ohnotnow:
This makes it a lot easier to debug when you are getting unexpected errors in a call to a bare `assertHasNoErrors()`

It certainly saved me some time. Mentioned it on discord but thought I'd PR it anyway as discord is quiet today.

No.

No - sorry.  I looked through the existing tests and couldn't see anywhere that it fitted.  It's also a fairly minor change - but I can add a test pretty easily I think - just didn't want to have a jarring out-of-place test splatted in the middle of existing test cases.

At the moment a failing call to a bare (no keys) `assertHasNoErrors()` simply says `Component has errors.` which isn't very helpful when you're not sure what they are.  This PR just makes it display `Component has errors: "field1", "some_other_field"`

And thank you for livewire!

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -172,7 +172,7 @@ trait MakesAssertions
         $errors = new MessageBag($this->payload['errorBag'] ?: []);
 
         if (empty($keys)) {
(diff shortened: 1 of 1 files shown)


calebporzio:
I dig. Thanks

Issue #1034, opened 2020-05-31
Microsoft Edge Autofill Error AutoFill_FillEvent


jjjrmy:
### Description
When doing autofill on Microsoft Edge, the inputs are updating with the proper value then flashing to 'AutoFill_FillEvent'. It only seems to be happening on Microsoft Edge
![Broken](https://i.imgur.com/y7vzNN8.gif "Broken")

### Steps to reproduce
```php
<input type="text" autocomplete="family-name lname" name="lname" wire:model.lazy="a.lname" class="@error('a.lname') error @enderror" />
```

**Context**
- Livewire version: Current
- Laravel version: Current
- Browser: Microsoft Edge

nuernbergerA:
Hey @jjjrmy

since autofill is still a pain to work with, could you try if this works for you?
```blade
<input x-data x-on:AutoFill_FillEvent="$dispatch('change', @this.value)" type="text" autocomplete="family-name lname" name="lname" wire:model.lazy="a.lname" class="@error('a.lname') error @enderror" />
```

jjjrmy:
I must do this on all inputs?

nuernbergerA:
Just for now to test this out

jjjrmy:
@nuernbergerA 
I don't think the `x-on:AutoFill_FillEvent` is firing at all.

nuernbergerA:
Can you share your HTML so I can test it on my windows machine?

nkeena:
I'm experiencing the same issue. My registration form is a Livewire component and the form inputs are blade components. Here's what the input for collecting the user's first name looks like:

```
<x-inputs.text 
    wire:model.lazy="firstName"
    name="firstName" 
    label="First name" 
    placeholder="John"
    required
/>
```

And the markup for the actual input field in the blade component:

```html
<input 
class="..."
value="{{ isset($value) ? old($name, $value) : old($name) }}"
autocomplete="{{ $name }}"
{{ $attributes->merge(['id' => $name]) }}
/>
```

Wondering if the problem is here: `autocomplete="{{ $name}}"` ?

calebporzio:
I'm new to this, but I just put a $200 USD bounty on fixing this issue on BountySource.

Here's the link: https://www.bountysource.com/issues/91607963-microsoft-edge-autofill-error-autofill_fillevent

BountySource already collected the $200 from me, so they are holding it until someone fixes this.

nuernbergerA:
[...]

PR #1180, opened 2020-06-29
Resolve an issue with IE Edge and Autofill

codeduction:

Yes, https://github.com/livewire/livewire/issues/1034

No

No

It seems with a line that resolves IE 11 problem, it was still catching the if case on IE Edge which handles values differently. So I have put in place a documentMode check to make sure we are on <IE11


ATholin:
Nice! I was just about to PR the same thing, so this should fix the issue.

Tested it on Edge and IE11.

codeduction:
Ah thanks @ATholin for testing it also.

calebporzio:
Thanks @codeduction! I'm going to merge this, as soon as we verify it is the fix, you should get your loot! Thanks so much!

PR #1181, opened 2020-06-29
Allow emitTo to target an instance of a component identified by it's key

### LINKED ISSUE #845: Component Key ID for $emitTo, like: $emitTo('order-item', {{ $order->id }}, 'eventName', params...)

mattharcourt:
This PR resolves #845 

It enables a component rendered by a directive such as `<livewire:counter :key="foo" />` to receive an event via `$emitTo('foo', 'increment')` without also sending the event to all other instances of the component. See #975 for further discussion of use cases.

Tests are included.

**Notes**

- This PR uses the component children that are maintained in the payload that gets passed in Livewire requests and responses - relying on the fact that they are indexed by the `:key` values assigned to components and contain the id of the corresponding component.

- When performing an `$emitTo('key', 'event')`, *all* Livewire components identified by `:key="key"` (regardless of type) will receive the event. So when targeting a specific component, it should have a unique key among all components on the page (and unique from component names). For example, you may see unexpected behavior when using the Blade `@foreach` `$loop->index` as a key in multiple places on a page. Though I suppose someone could want to utilize this behavior to send an event to the n<sup>th</sup> component instance (of any type of component) in several lists.

- Here is a [Gist](https://gist.github.com/mattharcourt/d4b84f7903797afdf2f83556c6f82446) with components that can be used to demonstrate the resulting behavior.


basepack:
This is a good solution to #845, but maybe we need to stick more to the Livewire docs/current implementation when add this to issue the event via the `emitTo` function.

Because `emitTo` normally needs a target, to quote the docs about `emitTo`:

```
Sometimes you may only want to emit an event to other components of the same type.

In these cases, you can use emitTo:

$this->emitTo('counter', 'postAdded');
```

So, may I suggest the following options, to stick better to the current implementation:
- change it to the `emit` event (because this is a broad one), like `$emit('key', 'event')`
- or change your implementation to support it like `$emitTo('component', 'key', 'event')`
- or maybe Caleb likes this implementation that much, that we can create a new function like `$emitToKey('key', 'event')` 

I will add a comment to your PR code too, to explain my point.

js/Store.js
```diff
@@ -75,6 +81,8 @@ const store = {
     emitTo(componentName, event, ...params) {
         let components = this.getComponentsByName(componentName)
 
+        components = components.concat(this.getComponentsByKey(componentName))
```
basepack, line 84: I can understand your code, but this lead possible to confusing later. I think a dedicated `$emitToKey` function would be better to seperate responsibilities?

basepack (changes requested):

mattharcourt:
@basepack - I see your point. However, I view the emit functionality slightly differently...

- As you stated, `emit` is broad. `emit` is used to broadcast an event to the entire Livewire landscape.
- The other emit methods apply scope and limit the event broadcast to a subset of the Livewire landscape - `emitUp` to parents of the current component, `emitSelf` to self, and `emitTo` to a specified component type.
- I see emitting to an instance of a component as a further scoping (specificity) of `emitTo`. In addition to specifying (scoping to) a subset of Livewire components (by type) this PR allows further specificity - narrowing the scope to a components with a given key. It's simply being more specific in the target.
- The only complicating factor is if someone tries to use a key that's the same as a component name. But I don't think this would be very common (except in trivial cases where we're trying to use 'foo' and 'bar' values everywhere :grinning:), and could fairly easily be worked around.

To me, this seems like a natural extension to `emitTo` - it's just a refinement of what the *to* is. But if that's just me, an `emitToKey` method could certainly be added.

(Not to further complicate things, but one could argue that if a separate `emitX` is needed, the existing scoping/naming suggests that the current `emitTo` functionality should be named `emitToSiblings` or `emitToType` and then `emitTo` would target a specific component instance. I'm not suggesting we go there :grinning:)

basepack:
All valid points @mattharcourt, I am super happy with your current implementation. Which would help a lot in my current projects if accepted. 

I can understand why you choose this naming, seems fine. And maybe @calebporzio also wants to say his thing about naming.

calebporzio:
Thanks for this PR. I do understand the need for something like this, but the appropriate API still isn't clear to me.

I'm going to close this for now and maybe revisit later.

Thanks!

booni3:
I would love something like this. 

Does anyone know of a suitable workaround for something like this?

In the below example, we have an accordion-type layout. When you click the "more details" button a large amount of database work is done. This might be a couple of hundred queries in total.

For this reason (as an example only) I only want to run the queries when the button is clicked. The problem is when you click the button we run all product-detail components inside the loop.

```blade
<div>
    @foreach($products as $product)
        {{ $product->name }}
        <button wire:click="openDetail">Details</button>
        <livewire:product-detail :product="$product" :key="'product_'.$product->id"/>
    @endforeach
</div>
```

basepack:
At the moment you will have to to the following to get similar functionality.

On your Product component:
```php
public function listeners() 
{
    return [
        'openDetail' . $product->id => 'openDetail'
    ];
}

```
In your blade view:
```blade
<button wire:click="$emit('openDetail{{$product->id}}')">
```
Written from my phone, so check carefully for errors. But you get the idea.

prathamesh-gharat:
I am trying to use [SortableJS](https://github.com/SortableJS/Sortable).
To explain, I got an instance of a Livewire "List" with nested "ListItem" components.
I need to emit event from JavaScript to specific instances of the "List" component but when I do that it currently triggers 3 ajax requests (1 for each instance of the list).

`emitToKey` as explained in this PR might help with that.

PR #1182, opened 2020-06-29
Refactor LivewireBladeDirectives class

mo7amed-3bdalla7:
It is just a small refactoring to reduce execute `explode` twice if condition is true.


calebporzio:
Thanks for the contribution, however, I don't like that $args is being assigned earlier and mutated within the conditional. Going to keep the current code. Thanks again!

PR #1183, opened 2020-06-29
Updated modifier debounce and time filter regex.

### LINKED ISSUE #1177: Keydown event always triggers for some modifiers
### Description
When using certain modifier (I've tried backspace and escape) together with wire:keydown e.g. wire:keydown.backspace, it triggers for all keydown events.

### Steps to reproduce

**Context**
- Livewire version: 1.2.0
- Laravel version: 7.17.2
- Browser: Chrome 83 on MacOS

Given this Livewire component.

```php
class Test extends Component
{
    public $data;

    public function render()
    {
        return view('livewire.test');
    }

    public function removeLast()
    {
        info('removeLast');
    }
}
```

```
<div>
    <input x-ref="input" wire:model="data" wire:keydown.backspace="removeLast" />
</div>
```

For every keydown event in the input removeLast() is triggered and logs to the log file. This is not the case for all modifiers. Using arrow-down seems to be fine.

ATholin:

Yes, fixes #1177.

No.

As mentioned in #1177, some events, such as the `keydown` event, is triggered even when a modifier is explicitly specified. Two examples are `escape` and `backspace`. An example where this does not happen is `enter`, which is the key used in the tests.

The error causing this is the regex for filtering 'debounce' and time modifiers. If the array is empty after filtering, we assume that there is no explicit modifier set, and it should always trigger. The specified regex looked like this: `/[0-9ms]/`, which finds a string that contains either 0-9, 'm' or 's' and removes them. `escape` and `backspace` both contain 's', and they will be removed. The array will be empty, and all modifiers are triggering the event.

This pull request resolves this issue by updating the specified regex to: `/^[0-9]+m?s$/`.

Thank you for all of your hard work!


stromgren:
I just want to say that this fix works perfectly for me. Is there any chance of getting this merged in the near future?

calebporzio:
Great PR, thanks!

PR #1186, opened 2020-06-30
Added support for validation in mount()

### LINKED ISSUE #1178: $this->validateOnly from Mount: Call to a member function toArray() on null: $this->errorBag->toArray()
I am getting an error from the ErrorBag when calling another function with validateOnly.

```php
Call to a member function toArray() on null
vendor/livewire/livewire/src/ComponentConcerns/ValidatesInput.php
---
$this->setErrorBag(
    Arr::except($this->errorBag->toArray(), $field)
 );
```

Here is a simplified version of my component:
```php
public $a;

public function mount() {
    if(true) {
        $this->a = "##########";
        $this->validateOnly('a', ['a' => 'required'], ['a.required' => 'Fill this field please.']);
        // ^^^ This is what is causing the error!
        // Or if I were to do this instead:
        // $this->phone();
        // ^^^ This has the same error!
    }
}

public function phone() {
    $this->validateOnly('a', ['a' => 'required'], ['a.required' => 'Fill this field please.']);
    // ^^^ This errors only if triggered by mount()
    // blah blah blah
    // do more stuff
}
```

If you comment out the `$this->validateOnly` in mount it works fine, so that's definitely the issue.
If the conditional in mount is false and you trigger the `phone()` function from the View with `wire:click` it validates the input fine and runs the rest of the function fine.

The problem is only caused when running $this->validateOnly from mount.

**Context**
- Laravel version: [e.g. 7.0.0]

ATholin:

Yes, fixes #1178.

No

Not sure how to test. It should assert that `ValidationException` is thrown in `mount()`, but tests fail saying that the exception thrown is `ErrorException`. This does not seem right.

Earlier, validating on `mount()`  was not possible, throwing an error.

It seems like the `errorBagExcept($field)` method call `$this->errorBag` directly instead of using the `$this->getErrorBag()` helper. When updating the method from:
```php
public function errorBagExcept($field)
{
    return new MessageBag(Arr::except($this->errorBag->toArray(), $field));
}
```
to
```php
public function errorBagExcept($field)
{
    return new MessageBag(Arr::except($this->getErrorBag()->toArray(), $field));
}
```

It seems to work when the validation passes. When the validation fails, however, the `ValidationException` is thrown and sent as a 500 response. This is resolved by wrapping the `$instance->mount(...$resolvedParameters)` in a ValidationException try-catch block, same as all other method calls. Though I'm not sure if this is the right approach.

Thank you for all of your hard work!


jjjrmy:
Thank you thank you!

jjjrmy:
Working, good job!

jjjrmy (approved):

ATholin:
Thanks!

calebporzio:
Thanks!

PR #1192, opened 2020-06-30
Fix conditional select options problems

calebporzio:
Because of the way browser handle default select values on <select> tags, when you have multiple select tags who's options depend on each others values, sometimes the value gets "reset"

This PR fixes this problem.


colinmac17:
@calebporzio I noticed some weird behavior when upgrading to 1.3 from 1.2. What I noticed is all my select elements were not showing the text of the selected option. When changing options the data is updated appropriately - it seems to only be a UI issue.

PR #1194, opened 2020-07-01
Register macros early

hxnk:
This moves the registration of (route) macros to the register method in the Livewire Service Provider.

Registering it here makes sure the macros are available in the boot method of other (Package) Service Providers wanting to use "Route::livewire" for route registration for example.


calebporzio:
Thanks!

PR #1197, opened 2020-07-01
Persist locale

### LINKED ISSUE #765: Re-rendering points to default locale (Localisation)
**Describe the bug**
Trying to validate in the updated() method. For English, it is working fine. But of other locales, the page re-renders to default locale.

**To Reproduce**
1. Create locale routes as `/en` and `/ar`.
2. Create an input field and add validation on `updated()`. 
3. When you get a validation error, despite having the `/ar` locale in the URI, the page turns English.

**Expected behavior**
It should stick to the locale while validation or `$refresh` from which the page was initially loaded.

**Screenshots**
Have a look at the URI. It is `/ar` but the validation is coming for`/en` instead of `/ar`
image

 - Browser: Chrome

danharrin:
Closes #765 by persisting locale between AJAX requests.


tests/ComponentLocaleIsPersistedTest.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Illuminate\Support\Facades\App;
+use Livewire\Livewire;
+use Livewire\Component;
+
+class ComponentLocaleIsPersisted extends TestCase
+{
+    /** @test */
+    public function a_livewire_component_can_persist_its_locale()
+    {
```
calebporzio, line 13: This test is a tad confusing to me.

Here's how I would re-structure:

1. Set the app locale
2. Boot up the component and check that the locale is set properly from mount to match the previously set locale
3. Change the local
4. Run a component method called "refreshLocale" (instead of "check")
5. Check that the right locale is set

calebporzio:
Couple merge conflicts, and then I can hit merge! Thanks Dan!

calebporzio:
Thanks Dan!

nuernbergerA:
Nice work @DanHarrin :+1:

PR #1199, opened 2020-07-01
Allow customized exception handling

### LINKED ISSUE #625: Allow customizing exception handling behavior
Currently, when an exception happens in a Livewire request, a modal is displayed with the response HTML. There is no (documented) way of overriding/customizing this.

The goal of this ticket is to request a (official/documented) way of customizing the exception handling behavior in frontend.

For now, we are overriding the showHtmlModal function, like this:
```
window.livewire.connection.driver.showHtmlModal = function(response) {
    const json = JSON.parse(response);
    // Handle JSON response...
    // E.g.: alert(response.error);
};
```
And in the Laravel's **Handler.php** class, inside the **render** method, we are doing something like this:

```
$response = parent::render($request, $exception);
if ($response->getStatusCode() == 500) {
    if ($request->header('X-Livewire')) {
        return response()->json(['success' => false, 'status' => 500, 'error' => 'Internal server error'], 500);
    }
}
if ($response->getStatusCode() == 404) {
    if ($request->header('X-Livewire')) {
        return response()->json(['success' => false, 'status' => 404, 'error' => 'Page or resource not found'], 404);
    }
}
return $response;
```
For now this works, but maybe there is a better and more elegant solution.

**Additional context**
I am VERY involved with Livewire project and I want it to be great (more than it already is)! I am available for helping with anything you need.

mattharcourt:
This PR resolves #625 

The logic in the connection http driver is changed so that a custom error handler defined as
```js
    window.livewire.onError(status => {
        return false
    }
```
will prevent errors or Laravel dumps from going to Livewire's default error handling. 

A return of exactly `false` from the custom error handler will prevent the default behavior (including the default page inactivity prompt). Handler logic can block default behavior for selected errors. For example, `if (status !== 419) return false` would prevent default behavior (i.e. show error in modal) for all but the page inactivity prompt.

The status information from the response (`ok`, `status` & `statusText`) are passed to the custom error handler after Livewire processes the failed payload. The entire response could be sent if there is a need for the custom error handler to access the response text (the dump, error text and/or Livewire data).

Tests are included which use Jest to mock `fetch` so that the driver code which responds to fetch is exercised. 
image


calebporzio:
Thanks for the contribution, however, this is already taken care of in this PR: https://github.com/livewire/livewire/pull/1146

PR #1202, opened 2020-07-01
Add macro support to TestableLivewire

jhoff:

No discussion or issue, though It seems like something people would find useful?

Nope

All current tests pass, doesn't really require additional testing.

Adds the ability to add custom assertions to the `TestableLivewire` class. For example:

```php
        TestableLivewire::macro('assertComputed', function ($property, $value) {
            PHPUnit::assertEquals($value, $this->viewData('_instance')->$property ?? null);

            return $this;
        });

            ->assertComputed('foobar', 42);
```

P.S. Fair warning that accepting this PR means this will no longer be true ☹️


calebporzio:
Thanks!

Issue #1007, opened 2020-05-24
Input values flicker when AJAX requests in flight 


joshhornby:
### Description

First can I just say this library is brilliant! 🎉

If I update an `input` element, quickly tab to the next input and update this as well. Due to the first AJAX request being inflight the 2nd input will update to the original data, then update once it's AJAX request has returned.

This is only noticeable with slow AJAX requests, in my example I'm using my local Docker which is quite slow.

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/5455767/82748864-17074480-9d9d-11ea-9765-df7521ca0c88.gif)

```
<div>
    <form wire:submit.prevent="submit">
        <input placeholder="Name" type="text" wire:model.lazy="name">
        @error('name') <span class="error">{{ $message }}</span> @enderror

        <input placeholder="Email" type="text" wire:model.lazy="email">
        @error('email') <span class="error">{{ $message }}</span> @enderror

        <button type="submit">Save Contact</button>
    </form>
</div>
```

### Steps to reproduce

- Create a component with two inputs
- Add values to both inputs
- Update the value of input 1
- Before the AJAX request has returned for input 1, update the value of input 2

**Context**
- Livewire version: 1.1
- Laravel version: 7.12
- Browser: Chrome

keepanitreel:
I do not believe this is a bug. 
Use Lazy Loading.
https://laravel-livewire.com/docs/properties#lazy-updating
If you can also add a debouncing value that will produce the result you are looking for.

joshhornby:
I am using lazy loading (see code above) and debouncing doesn't solve the issue unfortunately.

ryangjchandler:
Does adding a `wire:ignore` attribute to your input make any difference?

joshhornby:
Would adding `wire:ignore` mean I wouldn't be able to get the input values though?

I think the issue is more of a conceptual one, if a request is in flight & I edit another field on the form should the in fight request not be cancelled? Is there a reason you'd want the in flight request to update the DOM when you're already updating the DOM again?

ryangjchandler:
`wire:ignore` just ensures that Livewire doesn't swap out the DOM node, on seconds thoughts though `wire:model` is obviously mutating the input values via the JavaScript.

khanalpride:
@joshhornby I think the wire:model.passive which is in the roadmap will fix this issue.
https://github.com/livewire/livewire#roadmap

[...]

Issue #712, opened 2020-03-13
Avoid unnecessary update / re-render


tiago-sb:
**Is your feature request related to a problem? Please describe.**

Everytime a component renders, all its DOM needs to be generated and, in cases where this generation is heavy, we could be spending time (cpu/user) unnecessarily.

Concrete example:
I have a search form that, depending on the inputs, can query the DB for 0.5 million records. (these are then paginated of course). I want this search to happen only when the user presses the "submit" button on the form.
With the current implementation, there is no way (that I know of!) to avoid a re-render when any of the wired inputs change (wire:model.lazy is still "too much").

Similar issues have been raised that maybe could maybe benefit from a common solution:
- [#575](https://github.com/livewire/livewire/issues/575)
- [#313](https://github.com/livewire/livewire/issues/313)
- [#494](https://github.com/livewire/livewire/issues/494)

**Describe the solution you'd like**

Have a modifier like 'wire:model.passive' (i'm not good at naming things :)) that prevents a syncInput action from being sent to the server, but instead remains in the actionQueue until another action (that does not have '.passive') is queued. At this time both actions are sent to the server (maybe in 1 single network request?).

**Describe alternatives you've considered**

1. Issue #313 solution would work here, although we'd have to manually manage the form data when the "livewire-way" would be to have the inputs wired and the component public variables automatically hydrated.

**Additional context**

Please note I've started using livewire 1 week ago, so please forgive me if this can be solved in a simple way!

austenc:
Just ran into this as well. For this component I'm searching a bunch of text files, and when using `wire:model` the component re-renders too often for it to be performant. 

I ended up using the "hacky" solution Caleb mentioned in #313 for now.
```
<form wire:submit.prevent="submit(Object.fromEntries(new FormData($event.target)))">
```
Although this _will work_, a built-in solution would make me feel a lot better about the stability of this code... The question is, can this be implemented in a simplistic way that doesn't add much complexity to the framework? Could see either of these two approaches working:

### Option A
Add a modifier for `wire:model`. Similar to the proposed `.passive` option above, a bit different. What if (instead of queuing the updates) the ajax request to update the `wire:model` data could update the public property, but it wouldn't trigger a full render? So in this case you'd still have the ajax request, but it would just be "silent". This _could_ work, but in some cases may lead to race conditions... for example, what if the submit handler runs before the public property is updated?

### Option B
[...]

PR #1211, opened 2020-07-03
Implement wire:model "defer" modifier

ATholin:

Specified in [roadmap](https://github.com/livewire/livewire#roadmap), discussed in #1007, #712 and #1205 

No

Tests defined in `data_binding.spec.js`

By default, Livewire inputs send data on the `input` or `change` event. Using the `lazy` modifier, this can be changed to update when the user clicks away. In some cases, where a re-render might not be necessary for each input change, constantly updating these might be redundant. As discussed in #712, a solution to this could be a `passive` or `defer` modifier, that stores all of the changes in Javascript, and sends them all in a single request.

This pull request adds the `defer` modifier, which adds the action to the `actionQueue` but does not explicitly fire the message.

Let me know if there's anything I've missed!


ATholin:
I saw that the plan was to add this feature in the next major version, which could change the action queue strategy. I understand this PR might be a little early to consider 😄

calebporzio:
Closing, thanks though!

Issue #872, opened 2020-04-24
When using a custom pagination view, unable to go to the previous page.


alexjustesen:
**Describe the bug**
While using a custom pagination view, using wire:click"previousPage" doesn't trigger an action.

**To Reproduce**
```php
// Livewire/User/History.php

namespace App\Http\Livewire\User;

use Auth;
use Livewire\Component;
use Livewire\WithPagination;

class History extends Component
{
    use WithPagination;

    public function render()
    {
        $actions = Auth::user()->actions()
                            ->orderBy('created_at', 'desc')
                            ->paginate(2);
        
        return view('livewire.user.history', [
            'actions' => $actions,
        ]);
    }
}
```

```php
// livewire/user/history.blade.php

<div class="flex flex-col">
    <div class="-my-2 py-2 overflow-x-auto sm:-mx-6 sm:px-6 lg:-mx-8 lg:px-8">
        <div class="align-middle inline-block min-w-full shadow overflow-hidden sm:rounded-lg border-b border-gray-200">
            <table class="min-w-full">
                <thead>
                    <tr>
                        <th class="px-6 py-3 border-b border-gray-200 bg-gray-50 text-left text-xs leading-4 font-medium text-gray-500 uppercase tracking-wider">
                            Action
                        </th>
                        <th class="px-6 py-3 border-b border-gray-200 bg-gray-50 text-left text-xs leading-4 font-medium text-gray-500 uppercase tracking-wider">
                            IP address
                        </th>
                        <th class="px-6 py-3 border-b border-gray-200 bg-gray-50 text-left text-xs leading-4 font-medium text-gray-500 uppercase tracking-wider">
                            Time
                        </th>
                    </tr>
                </thead>

                <tbody>
                    @if($actions)
                        @foreach($actions as $action)
                        <tr class="bg-white hover:bg-gray-50">
                            <td class="px-6 py-4 whitespace-no-wrap text-sm leading-5 font-medium text-gray-900">
                                {{ $action->log_name }}.{{ $action->description }}
                            </td>
                            <td class="px-6 py-4 whitespace-no-wrap text-sm leading-5 text-gray-500">
                                {{ $action->ip_address }}
                            </td>
                            <td class="px-6 py-4 whitespace-no-wrap text-sm leading-5 text-gray-500">
                                {{ \Carbon\Carbon::parse( $action->created_at )->DiffForHumans() }}
                            </td>
                        </tr>
                        @endforeach
                    @endif
                </tbody>
            </table>

[...]

PR #1213, opened 2020-07-03
Handle adding/removing directives to/from an element

mattharcourt:
This PR adds tests to demonstrate issue #872 - see my [comment](https://github.com/livewire/livewire/issues/872#issuecomment-653601968) describing the use case and work-around.

It appears that when a directive is added to an element in place (by `morphAttrs`), that an event handler is not created for the newly added directive (it reads null as the existing value and replaces it with the new value). A similar issue can occur when a directive is removed - in that case the directive value is null and you get the "Cannot read property 'match' of null" error in `parseOutMethodAndParams`.

@calebporzio I'm not sure where/how you'd want to handle this use case - in the `onBeforeElUpdated` or `onElUpdated` hook or somehow in `morphDom` itself to recognize a difference in the directives between the from and to elements and do a wholesale replacement. For the case of a directive being removed, without a wholesale replacement, it might require tracking of the event handler so that it can be removed before tear down. In the least I thought these tests would be helpful to illustrate the issue.


mattharcourt:
A potential solution for the case where a directive is removed is for the handler to check if the directive is still valid. This doesn't remove the listener for the directive that was removed, but it eliminates the `"Cannot read property 'match' of null"` error in `parseOutMethodAndParams`.

mattharcourt:
For the case where a directive is added, the `beforeElUpdated` and `onElUpdated` callbacks passed to `morphdom` can record and the before update derivatives and compare them to the after update derivatives and initialize any that were added. 

Could try to utilize this to handle listeners for removed derivatives, but listeners aren't being tracked per directive so the best we can do without additional cataloging of listeners is let them run but ensure there are no catastrophic side effects (see 5c191b3). This has the potential to build up obsolete listeners if there are may re-renders of a component which add and remove directives.

calebporzio:
Thanks for the contribution, but closing as it has become stale and no one else has requested this in a while. Thanks!

Issue #885, opened 2020-04-28
wire.loading inline style display:inline-block


jjjrmy:
I am trying to put a loading icon in my input, it requires display flex.

The wire:loading automatically applies the inline style display:inline-block
Is there any way to override this other than adding !important to a CSS class?

https://imgur.com/F7JoXwY

jjjrmy:
Also happens on wire:offline.

This is a problem if you are using tailwinds as the inline style will always override it

indigoram89:
+1

capjjs:
try wrapping it in another div/span and apply wire:loading on that.

calebporzio:
I'm not clear exactly on the scenario you are describing.

I would think that Livewire would only add display: none when the icon should be hidden and that would be the behavior you would want.

Anyhow, if wrapping the icon in a parent element with `wire:loading` is not an option, then please re-submit with clear code examples I can easily copy and paste.

Thanks!

cdbessig:
I believe we are hitting the same issue.  The use case we have is that we want to display the loading icon centered in a table cell that spans 3 columns.  The display:inline is being attached to the tr element and throwing off all the formatting. 

`<tbody wire:init="loadSites">

                        <tr wire:loading>

                            <td colspan="3" class=" align-middle text-center items-center">

                                <div class="la-ball-pulse la-dark">

                                    <div></div>
                                    <div></div>
                                    <div></div>
                                </div>
                            </td>
                        </tr>
                        
                        @foreach($sites as $site)
                            @if($loop->odd)
                                <tr class="bg-white">
                            @else
                                <tr class="bg-gray-50">
                            @endif
                                    <td class="px-2 py-1 whitespace-nowrap text-sm font-medium text-gray-900">
                                        {{ $site['something'] }}
                                    </td>
                                    <td class="px-2 py-1 whitespace-nowrap text-sm text-gray-500">
                                        {{ $site['else'] }}
                                    </td>
                                    <td class="px-2 py-1 whitespace-nowrap text-sm text-gray-500">
                                        {{ $site['here'] }}
                                    </td>
                                </tr>
                        @endforeach
                        </tbody>`

joshhanley:
@cdbessig try to make use of toggling classes instead of using `wire:loading` on its own. https://laravel-livewire.com/docs/2.x/loading-states#toggling-classes

That way you could have a `hidden` class on the loading TR by default and remove it when loading.

Hope this helps!

cdbessig:
[...]

PR #1224, opened 2020-07-06
feat: add loading.display modifier to allow other display types

mblarsen:
This PR adds a modifier to `wire:loading` that lets you override the default display property being set when loading (`inline-block`). 

```
wire:loading.display="flex"
```

You can probably get the same functionality with clever use of `wire:loading.attr` and `style`, however that doesn't feel very Livewire.

Anyway this is useful when you have just a little more than a `Loading...` text. In my particular case my loading is an overlay on an avatar that is using the padding-ratio hack for sizing. I use flex to center the 'Uploading' text.

Test included.


mblarsen:
Hmm, would I need to add this?

```
[wire\:loading.display] {
  display: none;
}

mblarsen:
Hey @calebporzio, any change you'd consider this PR or a variation of it?

calebporzio:
Hmmm... I'm not sure about this API.

Are you using Tailwind? Wondering if it would be crazy to just check for the existence of a "flex" class (or "block" or "inline") and set the display type accordingly. Am I crazy?

mblarsen:
I know what you mean about the API. But the more I look at it it grows on me. It is like a modifier for loading. (which is a display directive). Like enter for `keydown.enter`.

The default is fine for most cases, so I don't think you'll start seeing code littered with `wire:loading.display="flex"`.

I happen to use Tailwind, but it seems like a worse idea to code for a specific CSS framework, no matter how awesome it is.

underdpt:
I just hit this issue, but not using tailwind but bootstrap4. I think it would be an error to tail livewire to an specific css framework.

Also, I like the solution the PR provides, as it allows the developer to use different properties in different contexts.

devinfd:
I'm not sure that this PR is what is needed. The bottom line is that livewire should not add `display:inline-block` to `wire:loading` elements. When the element is loading the display property should just be removed so that either the element will behave it's natural way or use any display css given to it via classes.

ref: https://github.com/livewire/livewire/issues/885

mblarsen:
Yes I guess it can be solved by removing the display style and then make use us .class and .class.remove

Perhaps keep it if class modifier isn't used. In other words when using the .class modifier no display is set.

loading use style
loading.class no style change
loading.class.remove no style change 

This is a breaking change though. 

The wire:loading css rule that hides can be kept as well.

Francismori7:
> I'm not sure that this PR is what is needed. The bottom line is that livewire should not add `display:inline-block` to `wire:loading` elements. When the element is loading the display property should just be removed so that either the element will behave it's natural way or use any display css given to it via classes.
> 
> ref: #885

CC @calebporzio 

Hey there,

You requested a demo, here's a quick screen capture of the problem. I disabled the "display: none" on wire:loading for testing purposes. As you can see, Livewire adds a style property with `display: inline-block` which in turns breaks the Tailwind "flex icon" positioning on the input, since it needs to be flex (as in my classes).

IMHO: Livewire should not add `display: inline-block` but rather just add `display: hidden` when it should be hidden, the display property's responsibility should be to the styling of the item being hidden/shown, rather than to Livewire itself.

https://streamable.com/tpf4pz

calebporzio:
Closing this.

I've implemented a feature that solves this problem here: https://github.com/livewire/livewire/pull/1935

PR #1226, opened 2020-07-06
fix: correct typeof check for IE11 documentMode

mblarsen:
Currently it does a string comparison which would always be false.

I.e. `window.document.documentMode == 'undefined'`.

It should be:

`typeof window.document.documentMode == 'undefined'` in order to use `event.detail`

---

I ran into this issue as my payload was sent back for `syncInput` with `name` set but `value` was always undefined. This is when using Apline's `$dispatch('input', value')`.

I think it is a regression introduced from the autofill issue:

 Commit:  15ac0ba7a39591ac0698a03d777764c28254b1a7
 Author:  Jason Mountcastle <git@aspn.dev>
 Date:    Tuesday, 30 June, 2020 01:15:34 AM PST


mblarsen:
Come to think about it. This bug must impact all non-native events emitted from non-input type elements.

calebporzio:
Ooof, thanks so much!

eduarguz:
Hi @mblarsen given you already addressed a related issue, could it be possible that other browsers have some similar issues?

I am debugging a similar problem in safari and someone reported the issue while using chrome

somewhere in js/node_initializer.js where you added the IE11 support
```
                // We have to check for typeof e.detail here for IE 11.
                const value =
                    e instanceof CustomEvent && typeof e.detail != 'undefined' && typeof window.document.documentMode == 'undefined'
                        ? e.detail
                        : el.valueFromInput(component)
```

I get that the e var is not a CustomEvent

image

Could you think what the issue is or guide me into how to fix it?

----- 

I am using this approach in my livewire/alpine component https://dev.to/eduarguz/adding-quilljs-editor-to-livewire-alpinejs-app-389l

PR #1232, opened 2020-07-07
Fix session expiry reload confirmation dialogue

### LINKED ISSUE #1220: "Session expire" loops for every page component not refreshing the page.
### Description

I've updated Livewire from 1.2 to 1.3.
I opened my website which has many livewire components in the same page and I left the office, I came back and there was an HTML alert saying something like "Your session expired would you like to to refresh the page?"  I clicked "ok" and a new modal showed up, I clicked ok and another one showed up and so on. I think that fired for every component in the page, until I pressed "cancel" and it finally refreshed the page.

### Steps to reproduce

Just place some components on the page an let the session expire.

**Context**
- Livewire version: 1.3.0
- Laravel version: 7.18.0
- Browser: Chrome

danharrin:
Closes #1220.


calebporzio:
Thanks Dan!

PR #1231, opened 2020-07-07
Fix file upload dehydration

iantasker:
1️⃣ Is this something that is wanted/needed? Did you create an issue / discussion about it first?

This fixes https://github.com/livewire/livewire/issues/1222

2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No

3️⃣ Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have)

No due to https://github.com/livewire/livewire/issues/1238 

4️⃣ Please include a thorough description of the improvement and reasons why it's useful.

N/A

5️⃣ Thanks for contributing! 🙌

Issue #1222, opened 2020-07-05
File Uploads within Array Public Property


jjjrmy:
When using file-upload and wire model as an array, there are some issues on hydration where the data is lost.
![Help](https://i.imgur.com/j39w5C1.gif)

```php
<?php

namespace App\Http\Components;

use Livewire\Component;
use Livewire\WithFileUploads;

class Form extends Component
{
    use WithFileUploads;
    public $a = [];

    public function mount() {
        $this->a['file_upload'] = []; // Get Input Ready For MultiFile
    }
}
```

```html
<input type="text" name="first_name" wire:model.lazy="a.first_name" />
<input type="file" multiple name="files" wire:model="a.file_upload" />

<?php dump($this->a); ?>
```

jkque:
Got this working by tweaking the `WithFileUploads.php` .

```php
    protected function dehydratePropertyFromWithFileUploads($name, $value)
    {
        if ($value instanceof TemporaryUploadedFile) {
            return $value->serializeForLivewireResponse();
        } elseif (is_array($value) && isset($value[0]) && $value[0] instanceof TemporaryUploadedFile) {
            return $value[0]::serializeMultipleForLivewireResponse($value);
        } elseif (is_array($value) && $this->isAssoc($value)) {
            $key = collect(array_keys($value))->map(function ($key) use ($value) {
                if ($value[$key] instanceof TemporaryUploadedFile) {
                    return [$key, 'singleUpload'];
                } elseif (is_array($value[$key]) && isset($value[$key][0]) && $value[$key][0] instanceof TemporaryUploadedFile) {
                    return [$key, 'multipleUpload'];
                }
            })->first();

            if($key) {
                if ($key[1] === 'singleUpload') {
                    $value[$key[0]] = $value[$key[0]]->serializeForLivewireResponse();
                }

                if ($key[1] === 'multipleUpload') {
                    $value[$key[0]] = $value[$key[0]][0]::serializeMultipleForLivewireResponse($value[$key[0]]);
                }
            }
        }

        return $value;
    }

    protected function isAssoc(array $arr) 
    {
        if (array() === $arr) return false;
        return array_keys($arr) !== range(0, count($arr) - 1);
    }
```

added 2nd elseif checking if the value is an associative array

mblarsen:
Can you use the diff tag for the code block above to better illustrate your changes:

```diff
- removed
  unchanged
+ added
```

Like so:

    ```diff
    - removed
      unchanged
    + added
    ```

[...]

PR #1234, opened 2020-07-08
fixed fileupload unserialize process

jkque:
This fixes [commit](https://github.com/livewire/livewire/pull/1231/commits/60bf5233d0c4a8cd740ee8e9556693c6df366179) for the issue #1222 
No
No, but tested on local
n/a

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -145,20 +145,11 @@ class TemporaryUploadedFile extends UploadedFile
         if (is_string($subject)) {
             return Str::startsWith($subject, 'livewire-file:')
                 || Str::startsWith($subject, 'livewire-files:');
-        } elseif (is_array($subject)) {
-            return collect($subject)->filter(function ($key) {
-                return ! empty($key);
-            })->contains(function ($value) {
-                if (is_string($value)) {
-                    return Str::startsWith($value, 'livewire-file:')
-                    || Str::startsWith($value, 'livewire-files:');
+        }
 
-                } elseif (is_array($value)) {
-                    return collect($value)->contains(function ($value) {
-                        return Str::startsWith($value, 'livewire-file:')
-                        || Str::startsWith($value, 'livewire-files:');
-                    });
-                }
+        if (is_array($subject)) {
+            return collect($subject)->contains(function ($value) {
+                return static::canUnserialize($value);
             });
         }
 
@@ -174,17 +165,15 @@ class TemporaryUploadedFile extends UploadedFile
                 $paths = json_decode(Str::after($subject, 'livewire-files:'), true);
 
                 return collect($paths)->map(function ($path) { return static::createFromLivewire($path); })->toArray();
+            } else {
+                return $subject;
             }
-        } elseif (is_array($subject)) {
-            collect(array_keys($subject))->map(function ($key) use (&$subject) {
-                if (Str::startsWith($subject[$key], 'livewire-file:')) {
-                    $subject[$key] = static::createFromLivewire(Str::after($subject[$key], 'livewire-file:'));
-                } elseif (Str::startsWith($subject[$key], 'livewire-files:')) {
-                    $paths = json_decode(Str::after($subject[$key], 'livewire-files:'), true);
-
-                    $subject[$key] = collect($paths)->map(function ($path) { return static::createFromLivewire($path); })->toArray();
-                }
-            });
+        }
+        
+        if (is_array($subject)) {
+            foreach (array_keys($subject) as $key) {
+                $subject[$key] =  static::unserializeFromLivewireRequest($subject[$key]);
+            }
 
             return $subject;
         }
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -172,8 +172,8 @@ class TemporaryUploadedFile extends UploadedFile
         }
         
         if (is_array($subject)) {
-            foreach (array_keys($subject) as $key) {
-                $subject[$key] =  static::unserializeFromLivewireRequest($subject[$key]);
+            foreach ($subject as $key => $value) {
+                $subject[$key] =  static::unserializeFromLivewireRequest($value);
             }
 
             return $subject;
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -142,20 +142,51 @@ class TemporaryUploadedFile extends UploadedFile
 
     public static function canUnserialize($subject)
     {
-        if (! is_string($subject)) return false;
+        if (is_string($subject)) {
+            return Str::startsWith($subject, 'livewire-file:')
+                || Str::startsWith($subject, 'livewire-files:');
+        } elseif (is_array($subject)) {
+            return collect($subject)->filter(function ($key) {
+                return ! empty($key);
+            })->contains(function ($value) {
+                if (is_string($value)) {
+                    return Str::startsWith($value, 'livewire-file:')
+                    || Str::startsWith($value, 'livewire-files:');
 
-        return Str::startsWith($subject, 'livewire-file:')
-            || Str::startsWith($subject, 'livewire-files:');
+                } elseif (is_array($value)) {
+                    return collect($value)->contains(function ($value) {
+                        return Str::startsWith($value, 'livewire-file:')
+                        || Str::startsWith($value, 'livewire-files:');
+                    });
+                }
+            });
+        }
+
+        return false;
     }
 
     public static function unserializeFromLivewireRequest($subject)
     {
-        if (Str::startsWith($subject, 'livewire-file:')) {
-            return static::createFromLivewire(Str::after($subject, 'livewire-file:'));
-        } elseif (Str::startsWith($subject, 'livewire-files:')) {
-            $paths = json_decode(Str::after($subject, 'livewire-files:'), true);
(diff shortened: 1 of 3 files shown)


src/TemporaryUploadedFile.php
```diff
@@ -142,20 +142,51 @@ public static function createFromLivewire($filePath)
 
     public static function canUnserialize($subject)
     {
-        if (! is_string($subject)) return false;
+        if (is_string($subject)) {
+            return Str::startsWith($subject, 'livewire-file:')
+                || Str::startsWith($subject, 'livewire-files:');
+        } elseif (is_array($subject)) {
```
calebporzio, line 148: I would break this elseif out into a standalone "if". It will make it look nicer and less like a big nested chunk of code.
The first if returns so this is possible

src/TemporaryUploadedFile.php
```diff
@@ -142,20 +142,51 @@ public static function createFromLivewire($filePath)
...
+            return collect($subject)->filter(function ($key) {
```
calebporzio, line 149: First, a ->filter() without any args should work fine for what you want.

Second, why do you even need a filter here at all? I don't think you do.

src/TemporaryUploadedFile.php
```diff
@@ -142,20 +142,51 @@ public static function createFromLivewire($filePath)
...
+                return ! empty($key);
+            })->contains(function ($value) {
+                if (is_string($value)) {
```
calebporzio, line 152: This should be recursive so you aren't repeating yourself, AND so it covers more than 2 layers deep.

src/TemporaryUploadedFile.php
```diff
@@ -142,20 +142,51 @@ public static function createFromLivewire($filePath)
...
-            return collect($paths)->map(function ($path) { return static::createFromLivewire($path); })->toArray();
+        if (is_string($subject)) {
+            if (Str::startsWith($subject, 'livewire-file:')) {
+                return static::createFromLivewire(Str::after($subject, 'livewire-file:'));
+            } elseif (Str::startsWith($subject, 'livewire-files:')) {
+                $paths = json_decode(Str::after($subject, 'livewire-files:'), true);
+
+                return collect($paths)->map(function ($path) { return static::createFromLivewire($path); })->toArray();
+            }
+        } elseif (is_array($subject)) {
```
calebporzio, line 178: same as before, break this out into a new line and standalone "if"

src/TemporaryUploadedFile.php
jkque: updated my pr, thanks @calebporzio

src/TemporaryUploadedFile.php
```diff
@@ -142,20 +142,40 @@ public static function createFromLivewire($filePath)
 
     public static function canUnserialize($subject)
     {
-        if (! is_string($subject)) return false;
+        if (is_string($subject)) {
+            return Str::startsWith($subject, 'livewire-file:')
```
nuernbergerA, line 146: 2nd param takes also array
`return Str::startsWith($subject, ['livewire-file:', 'livewire-files:']);`

src/TemporaryUploadedFile.php
```diff
@@ -142,20 +142,40 @@ public static function createFromLivewire($filePath)
...
+        return false;
     }
 
     public static function unserializeFromLivewireRequest($subject)
     {
-        if (Str::startsWith($subject, 'livewire-file:')) {
-            return static::createFromLivewire(Str::after($subject, 'livewire-file:'));
-        } elseif (Str::startsWith($subject, 'livewire-files:')) {
-            $paths = json_decode(Str::after($subject, 'livewire-files:'), true);
+        if (is_string($subject)) {
```
nuernbergerA, line 161: early return for the win:
```php
if (is_string($subject)) {
    if (Str::startsWith($subject, 'livewire-file:')) {
        return static::createFromLivewire(Str::after($subject, 'livewire-file:'));
    }
    
    if (Str::startsWith($subject, 'livewire-files:')) {
        $paths = json_decode(Str::after($subject, 'livewire-files:'), true);

        return collect($paths)->map(function ($path) { return static::createFromLivewire($path); })->toArray();
    }
    
    return $subject;
}
```

src/WithFileUploads.php
```diff
@@ -102,8 +102,12 @@ protected function dehydratePropertyFromWithFileUploads($name, $value)
     {
         if ($value instanceof TemporaryUploadedFile) {
```
nuernbergerA, line 103: Again early returns would be more readable
```php
if ($value instanceof TemporaryUploadedFile) {
    return $value->serializeForLivewireResponse();
}

if (is_array($value) && isset($value[0]) && $value[0] instanceof TemporaryUploadedFile) {
    return $value[0]::serializeMultipleForLivewireResponse($value);
}

if (is_array($value)) {
    foreach (array_keys($value) as $key) {
        $value[$key] = $this->dehydratePropertyFromWithFileUploads(null, $value[$key]);
    }
}

return $value;
```

nuernbergerA (changes requested):

nuernbergerA:
`removing_first_item_from_array_of_temporary_uploaded_files_serializes_correctly` test need to be fixed

src/WithFileUploads.php
jkque: @nuernbergerA - push some updates and thanks for the tip :)

src/WithFileUploads.php
```diff
@@ -102,9 +102,17 @@ protected function dehydratePropertyFromWithFileUploads($name, $value)
...
             return $value->serializeForLivewireResponse();
-        } elseif (is_array($value) && isset(array_values($value)[0]) && array_values($value)[0] instanceof TemporaryUploadedFile) {
+        }
+
+        if (is_array($value) && isset(array_values($value)[0]) && array_values($value)[0] instanceof TemporaryUploadedFile && is_numeric(key($value))) {
             return array_values($value)[0]::serializeMultipleForLivewireResponse($value);
         }
+        
+        if (is_array($value)) {
+            foreach (array_keys($value) as $key) {
```
calebporzio, line 112: why not use a regular `foreach ($value as $something)` instead of the key thing?

src/TemporaryUploadedFile.php
```diff
@@ -142,20 +142,41 @@ public static function createFromLivewire($filePath)
...
+                $paths = json_decode(Str::after($subject, 'livewire-files:'), true);
+
+                return collect($paths)->map(function ($path) { return static::createFromLivewire($path); })->toArray();
+            }
+
+            return $subject;
+        }
+        
+        if (is_array($subject)) {
+            foreach (array_keys($subject) as $key) {
```
calebporzio, line 175: again, use a normal foreach here

calebporzio:
Will need to have tests included in this PR to merge it. Thanks!

iantasker:
@calebporzio Unless the scope of this PR is widened a test can't be included due to this issue #1238 that I found when I was testing PR #1231. I believe you have fixed it in version 2.0.

src/TemporaryUploadedFile.php
iantasker: @calebporzio  A filter is required to skip all null string, if not `Str::startsWith` will throw an exception.

src/TemporaryUploadedFile.php
iantasker: Here is the recursive code from my PR.

```suggestion
            return collect($subject)->filter(function ($key) {
                return ! empty($key);
            })->contains(function ($value) {
                return static::canUnserialize($value);
            });
```

src/WithFileUploads.php
jkque: we have to assign it back the dehydrated property of the value to its key and return the whole array. Maybe $key variable
`foreach ($value as $key => $item) {` will do

jkque:
@calebporzio - updated my pr and added tests, thanks :)

tests/FileUploadsTest.php
```diff
@@ -472,6 +472,73 @@ public function removing_first_item_from_array_of_temporary_uploaded_files_seria
...
+    /** @test */
+    public function removing_first_item_from_array_of_temporary_uploaded_files_serializes_correctly_with_in_array_public_property()
+    {
+        $file1 = UploadedFile::fake()->image('avatar1.jpg');
+        $file2 = UploadedFile::fake()->image('avatar2.jpg');
+        $file3 = UploadedFile::fake()->image('avatar3.jpg');
+        $file4 = UploadedFile::fake()->image('avatar4.jpg');
+
+        $component = Livewire::test(FileUploadInArrayComponent::class)
+                             ->set('obj.file_uploads', [$file1, $file2, $file3, $file4])
```
iantasker, line 485: AFAIK `set` is not setting `obj.file_uploads` as a nested array it's setting `obj.file_uploads` as a string, which would mean the test isn't complete even though it passes.

jkque: i think it does setting `obj.file_uploads` as nested array and not as a string.
try dumping `$component->get('obj')` and `$component->get('obj.file_uploads')`

but i got Undefined index: obj.file_uploads in `$component->viewData('obj.file_uploads')` and a tweak was to get the temp files like this `$component->viewData('obj')['file_uploads']`

GertjanRoke:
@calebporzio do you have any idea when this is going to fixed/merged? Or is this going to be in v2? I could also test this code if needed, I have some code waiting for this fix so would love to help.

Great job guys! 👍

PR #1239, opened 2020-07-08
Fixing the class name to comply with psr-4 autoloading standard

arcanedev-maroc:
Fixing The Deprecation Notice: 
`Class Tests\ComponentLocaleIsPersisted does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0.`


calebporzio:
Thanks!

Issue #1263, opened 2020-07-12
Fileupload not respecting config values


ClausMunch:
### Description
When using a S3 compatible disk, Livewire Fileupload builds .amazonaws.com strings to upload files and thus not respecting the config in filesystems.php

### Steps to reproduce
Basic example of a clean Laravel install w. livewire and minio or digitalocean space configured.
Basic tutorial from the documentation will fail and try to upload file to an {bucket}.S3.{region}.amazonaws.com address.

**Context**
- Livewire version: dev-master ad2f1fd
- Laravel version: v7.19.1
- Browser: Chrome

iAmKevinMcKee:
Are you sure you have the correct region? Can you double check that your .env has the right code from here: https://docs.aws.amazon.com/general/latest/gr/rande.html

ClausMunch:
Yes, I'm sure. I'm not using an Amazon S3, but have tried with both digitalocean spaces and Minio S3 storage.
Works brilliantly, except for with Livewire.

PR #1294, opened 2020-07-18
Fix file upload to s3 compatible storage

gdebrauwer:
**What is the problem?**

The file upload to s3 storage always uses the env-variables with `AWS_` prefix to create the signed url.
This will not be the case when you use another s3-compatible storage service, for example, "DigitalOcean Spaces". Your config variables will be called differently.

```php
// in 'disks' of 'config/filesystems.php'

'spaces' => [
    'driver' => 's3',
    'key' => env('DO_SPACES_KEY'),
    'secret' => env('DO_SPACES_SECRET'),
    'endpoint' => env('DO_SPACES_ENDPOINT'),
    'region' => env('DO_SPACES_REGION'),
    'bucket' => env('DO_SPACES_BUCKET'),
    'url' => env('DO_SPACES_URL'),
],
```

The signed url will currently not be created correctly by Livewire, because it assumes env-variables that are prefixed with `AWS_`.

This issue has also been reported by someone else (https://github.com/livewire/livewire/issues/1263).

**How did I fix the problem?**

This PR fixes the issue by using the s3 client/driver/adapter that has been created by the Laravel Storage Facade.

_I tested this implementation using an AWS S3 bucket and a DigitalOcean Spaces bucket._


xitox97:
@gdebrauwer did you know why Im getting this errors? when try uploading the files
`has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

gdebrauwer:
@xitox97 you have to set the CORS configuration of you bucket, because you are uploading a file directly to an other domain.

ClausMunch:
I have played around with this PR. Nice work @gdebrauwer !!
One thing that is weird is, I have spun up a DO Spaces. Having livewire-tmp there works like a charm.
Doing the same on a Minio instance spun up by ploi.io, fails. But only for the temp files? If I have the temp files on eg. local or DO Spaces and store the image onto Minio (S3 compatible), it saves it just right.

So I'm thinking, what is done differently on temp and norm. storage?

joaokamun:
Just to know, what's missing to merge this?

vikas5914:
Just tested with Linode Object Store as well. Working fine. 

Please merge this.

glaesser:
This works great and is really an essential feature, please merge?

Rusya13:
Without extracting hardcoded envs to config you can't use laravel-livewire file uploads deploying via Vercel 
 for example: https://vercel.com/docs/platform/limits#reserved-variables

jimmypuckett:
I found one more small issue where the path to the temp URL is not taking into consideration the configuration.  I made a small PR to correct.

Thanks @gdebrauwer & @calebporzio for making this work!

glaesser:
Also noting here that ->storePublicly() seems to have no effect (tested on DigitalOcean using the new livewire compatibility), i.e. uploads still default to private permission.

jimmypuckett:
I have found an issue with using this on GCS where the file name is not safe.  I opened issue #1456

calebporzio:
@glaesser  - ->storePublicly() has been fixed for cloud storage providers already I believe.

PR #1297, opened 2020-07-20
Fix Bug with Success Message on Make Component Command

iAmKevinMcKee:

I didn't create an issue, but found the bug and fixed it.

Just a single PR for this bugfix.

I didn't add tests because I didn't see any other tests related to the console messaging so wasn't sure if you wanted them. I can add if you'd prefer.

People will not get a confusing message when they try to create a component and it fails.

You're welcome! 👍


iAmKevinMcKee:
For more context on what this is actually fixing, here's a screenshot. If you try to make a component and nothing is actually created, you still get a success message "COMPONENT CREATED" in the console.

This PR removes that message when no new files were created.

image

danharrin (approved):

PR #1318, opened 2020-07-25
Add assertSeeInOrder() and assertSeeHtmlInOrder() assertions

intrepidws:
I think so, as it creates further parity with available Laravel testing assertions. No discussion was created first.

No.

Sorry, no. Testing the testing is beyond my brain power.

Assertions that the given string (or HTML) are contained in order within the response. Similar to the Laravel equivalents.


Patabugen:
I came here looking for something like `assertSeeInHtmlIgnoringWhitespace` but this would also suit my use case.

calebporzio:
Thanks Craig!

PR #1325, opened 2020-07-26
added version to package.json

cbl:
The package cannot be installed via **npm** unless a version is given in the `package.json` file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
     "main": "dist/livewire.js",
     "name": "laravel-livewire",
+    "version": "1.0",
     "scripts": {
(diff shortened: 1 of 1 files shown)


DanHarrin:
Hey, Livewire isn't meant to be installed through NPM, it's a Laravel package that should be installed through Composer. Many thanks.

cbl:
Allowing users to install just the javascript components makes it possible to create and test extensions for the the Livewire javascript functionality. For example I am currently working on a debug tool for Livewire and I would like to use the javascript classes without having to install the composer package since I am not writing any php.

PR #1329, opened 2020-07-27
Added a prepareForValidation() method to modify the attributes before validation

rennokki:
I stumbled upon making the users able to add tags using commas, but on validation, there was no way of preparing my data for validation to swap the words separated by commas to an array.

I have added a `prepareForValidation()` method just like Laravel has: https://laravel.com/docs/7.x/validation#prepare-input-for-validation

It can be overwritten in the component class:

```php
class MyLivewireComponent
{
    protected function prepareForValidation($attributes)
    {
        $attributes['tags'] = explode(',', $attributes['tags'] ?? null);

        return $attributes;
    }
}
```

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -74,6 +74,8 @@ trait ValidatesInput
                 = $this->getPropertyValue($propertyNameFromValidationField);
         }
 
+        $result = $this->prepareForValidation($result);
+
         $result = Validator::make($result, Arr::only($rules, $fields), $messages, $attributes)
             ->validate();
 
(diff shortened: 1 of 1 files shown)


danharrin (approved):

nuernbergerA:
Hey, @rennokki thanks for the PR.

The implementation looks good.
Would be awesome if you could add some tests.

Cheers Adrian

rennokki:
@nuernbergerA I thought of adding some, but as long as the validation tests pass, it means it gets through the implemented method. 🤷

ryangjchandler:
@rennokki - that's true, but it's always good to test that the data being returned from the `prepareForValidation()` method is working correctly too, for sanity's sake :)

ceejayoz:
Should this also fire in `validateOnly` calls?

PR #1345, opened 2020-07-31
Fix file upload error message formatting

### LINKED ISSUE #1303: File upload validation message displays "files.0"
### Description
When using translation translation strings with uppercase attributes the file upload error messages displays "files.0" instead of the attribute name.

### Steps to reproduce
1. Change the validation.max.file message from `'The :attribute may not be greater than :max kilobytes.'` to for example `':Attribute may not be greater than :max kilobytes.'` (uppercasing the first attribute letter).
2. Upload a file larger than max. The error message will now be displayed as: "files.0 may not be greater than x kilobytes."

andreasnij:
Fixes #1303 

When using translation translation strings with uppercase attributes
the file upload error messages displays "files.0" instead of the
attribute name.

### DIFF
--- a/src/WithFileUploads.php
+++ b/src/WithFileUploads.php
@@ -57,8 +57,8 @@ trait WithFileUploads
         }
 
         $errorsInJson = $isMultiple
-            ? str_replace('files', $name, $errorsInJson)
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

Issue #1107, opened 2020-06-15
.csv files uploads with .txt extension


HalfBottleOfMind:
### Description

Files with extension .csv uploads as .txt files. I think problem in laravels guessExtension() method.

### Steps to reproduce

Try to upload .csv file. In livewire-tmp folder it will be .txt.

**Context**
- Livewire version: 1.2
- Laravel version: 7.15
- Browser: Chrome v84.0.4147.45

calebporzio:
It looks like this is a problem with Laravel itself. There are simple workarounds described here: https://stackoverflow.com/questions/51569475/how-to-store-csv-file-laravel-5-php

Thanks!

nowendwell:
Using the suggested workaround doesn't work as the extension is already gone by the time it gets to the Livewire component.

If you change the following, you can use getClientOriginalExtension() to get the correct extension.

[TemporaryUploadedFile.php #L128](https://github.com/livewire/livewire/blob/master/src/TemporaryUploadedFile.php#L128)
```php
$extension = '.'.$file->guessExtension();
```
to
```php
$extension = '.'.$file->getClientOriginalExtension();
```

bernhardh:
This ist still an issue. Uploading an csv file leads to a txt file. :(

saagoor:
This issue still remains, not just for .csv file but for .srt, .vtt etc files also.

majid-al-saidi:
I have this issue,

NeoLarson:
Have the same issue in Livewire 3.

PR #1359, opened 2020-08-05
Change the mimetype for mp3 files from 'mp3' to 'mpga'

MichaelBrauner:
I think for this no tests are needed. 

It just makes it possible to preview mp3 files because the mimetype for mp3 files is 'mpga'.

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -69,7 +69,7 @@ class TemporaryUploadedFile extends UploadedFile
         $supportedPreviewTypes = [
             'jpeg', 'png', 'gif', 'bmp', 'svg', 'webp',
             'mp4', 'mov', 'avi', 'wmv',
-            'mp3', 'wav', 'm4a', 'wma',
+            'mpga', 'wav', 'm4a', 'wma',
         ];
 
         if (! in_array($this->guessExtension(), $supportedPreviewTypes)) {


nuernbergerA:
Thanks for your PR @Slowwie.

I think it's not a good idea to replace `mp3` with `mpga`, because we currently use Laravel's `guessExtension()` which uses the file extension to guess the type.

I think we have 2 options here:
A) Move extensions to config so people can overwrite it easily
B) Refactor to another solution to tackle also issues like #1107

ryangjchandler:
Laravel uses the Symfony file stuff under the hood, which in turn uses `finfo()` iirc, so a config option might the most sense, I'm not sure how other libraries would do it without some sort of huge lookup table.

Slowwie:
@nuernbergerA maybe you can just add mpga to the list. So its supported and dont crash. Also for me - I hacked it like this and it workes without issues with file extensions or something else.

Slowwie:
@ryangjchandler As I know - Caleb is pretty proud of the "non-config" out of the box functionality of Livewire.  
From this perspective - I don't think its user friendly when u have to config something to be able to upload a simple mp3 file.

ryangjchandler:
> 
> 
> @ryangjchandler As I know - Caleb is pretty proud of the "non-config" out of the box functionality of Livewire.
> From this perspective - I don't think its user friendly when u have to config something to be able to upload a simple mp3 file.

That is true, but this a limitation of the underlying components. I'll do some research into alternatives.

ryangjchandler:
Had a look at this briefly & lots of people have been confused about the same thing with various different frameworks. 

AFAIK, the `guessExtension()` does in fact return `mpga`, not `mp3` - therefore this PR is actually a bug fix. 

The test coverage seems to be quite low for this functionality, since the specifics of previewing the specified file yours aren't being tested. 

@Slowwie would you like to have a go at writing a test for your changes? That's generally the best thing to do when you're fixing a bug, to ensure that there isn't a regression.

Slowwie:
@ryangjchandler I can try it. I am a Symfony Dev and wrote once my own tests for a Symfony CMS project wich I did not finished :D. 
But how to write tests for this as a contributer - I don't really know. 
Are there any guidelines to follow. I mean I don't even know what this tests should cover... .

ryangjchandler:
> 
> 
> @ryangjchandler I can try it. I am a Symfony Dev and wrote once my own tests for a Symfony CMS project wich I did not finished :D.
> But how to write tests for this as a contributer - I don't really know.
> Are there any guidelines to follow. I mean I don't even know what this tests should cover... .

No worries, if you don't feel like you can do it, I can always write a test and send it in to your branch :)

Slowwie:
That would be nice - so I see how u do this. I always worked alone and I wouldn't even know how to send a test to a branch. I guess I have to learn github stuff.

calebporzio:
This is a bit of a tricky one to test, I understand.

Thanks for the contribution!

PR #1364, opened 2020-08-07
Render CSP Nonce On Style Tag If Available As An Option

faradele:
Fixes #1364 

This adds support for passing a **nonce** option to the `@livewireStyles` directive and have the generated ```<style>``` tag contain a **nonce** attribute.

E.g. 
`@livewireStyles(['nonce' => 'somenoncevalue'])`
Will output:
`<style nonce="somenoncevalue">...</style>`

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -154,7 +154,7 @@ class LivewireManager
     {
         $debug = config('app.debug');
 
-        $styles = $this->cssAssets();
+        $styles = $this->cssAssets($options);
 
         // HTML Label.
         $html = $debug ? ['<!-- Livewire Styles -->'] : [];
@@ -180,10 +180,11 @@ class LivewireManager
         return implode("\n", $html);
     }
 
-    protected function cssAssets()
+    protected function cssAssets($options)
     {
+        $nonce = isset($options['nonce']) ? " nonce=\"{$options['nonce']}\"" : '';
         return <<<HTML
-<style>
+<style{$nonce}>
     [wire\:loading] {
         display: none;
     }
--- a/tests/LivewireAssetsDirectiveTest.php
+++ b/tests/LivewireAssetsDirectiveTest.php
@@ -95,4 +95,14 @@ class LivewireAssetsDirectiveTest extends TestCase
             $output
         );
     }
+
+    /** @test */
+    public function nonce_passed_into_directive_gets_added_as_style_tag_attribute()
+    {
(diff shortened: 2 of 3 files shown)


calebporzio:
Curious about his. I understand wanting this for your site's CSP, but is this a real issue you've faced? Or just a theoretical improvement? I believe Livewire's JS uses "eval" which would violate CSP anyways.

I'd love more info on your situation. I don't want to just pull this in because it's a good idea, you know?

faradele:
@calebporzio you're totally right 🤦‍♂️

Turns out I have unsafe-eval allowed for script-src on the page where I'm using livewire that was why I wasn't getting the related csp errors for livewire's 'eval' usage. 
I didn't have the unsafe-inline set for style tags on that page hence the csp errors I got with the generated <style> tag and why I thought nonce support would be needed for the generated style tag as well.

Seems full csp support is already toast on pages with livewire anyways so this is a big non-issue.
I'm sure we can close this out now since it seems to be a solution in search of a problem 😔.

calebporzio:
Great, thanks for following up!

davidjr82:
@calebporzio I think this shouldn't be closed without merging. CSP gives real security value to software. Eval is not that evil if is used with "nonce" attribute, because nonce allows the browser that the evaluated code is a trusted source.

In the same way nonce is merged for the script tag, it would be useful to merge it too for styles tag. 

If it is not merged, and someone still wants to use CSP headers, the solution is to copy the content of the style tags and add it to an app.css file, so no CSP issue will arise.

P.S.: This is a real issue I faced right now, solving it moving the styles to a app.css file, but I will have to be aware of possible future changes in the css assets.

alexandre67fr:
@faradele is right. Whats's the point of nonce support for script tag, but not for style tag?
Plus, without this PR https://github.com/spatie/laravel-csp is simply not compatible with livewire. 

Please reconsider the pull request.

RomainMazB:
@calebporzio I was about to create a duplicate PR for this!

I can confirm that spatie/laravel-csp isn't compatible out of the box with livewire, this is actually what made me come here.
We need this to make it compatible without allowing all unsafe inline which would result in a regression in our CSP.

> Note: Disallowing inline styles and inline scripts is one of the biggest security wins CSP provides. However, if you absolutely have to use it, there are a few mechanisms that will allow them.
> Source: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_inline_styles)

PR #1398, opened 2020-08-21
Restrict error bag persistence to component property errors

### LINKED ISSUE #962: Have to call resetErrorBag and resetValidation when using a custom validator
**Describe the bug**
I have an `update()` lifecycle hook which has some regular validation, and the a custom validator.  When the custom validator fails, the error is never cleared even if you resolve the date causing the validation problem.  Putting in a manual call to `$this->resetErrorBag()` and `$this->resetValidation()` before the call to `Validator::make()` seems to clear it up though.

**To Reproduce**

View :
```html
       @error('date1')
        <span class="has-text-danger">{{ $message }}</span>
        @enderror
<!-- ... various plain wire:models on inputs, eg -->
          <input class="input" type="text" ref="date2" name="formatted_date2" id="date2" wire:model.lazy="booking.formatted_date2">

```
Component :
```php
    {
        $this->validateOnly($field, [
            'booking.formatted_date1' => 'required|date_format:d/m/Y',
            'booking.formatted_date2' => 'required|date_format:d/m/Y',
        ]);

        $date1 = $this->booking['formatted_date1'] . $this->booking['preferred_tod1'];
        $date2 = $this->booking['formatted_date2'] . $this->booking['preferred_tod2'];

        // removing these lines leaves the custom validator 'date1' error always in the session once it's failed
        $this->resetErrorBag();
        $this->resetValidation();

        Validator::make(
            [
                'date1' => $date1,
                'date2' => $date2,
            ],
            ['date1' => 'different:date2'],
            ['different' => 'Preferred dates must be different']
        )->validate();
    }
```
**Expected behavior**
When the input date is corrected, the `date1` validation error should be gone.

Firefox, macos, but also in phpunit tests

mattharcourt:
This PR implements the alternate approach discussed in #1264 (see the discussion there for details), and fixes #962.

### DIFF
--- a/src/HydrationMiddleware/PersistErrorBag.php
+++ b/src/HydrationMiddleware/PersistErrorBag.php
@@ -14,7 +14,10 @@ class PersistErrorBag implements HydrationMiddleware
     public static function dehydrate($instance, $response)
     {
         if ($errors = $instance->getErrorBag()->toArray()) {
-            $response->errorBag = $errors;
+            $response->errorBag = collect($errors)
+                ->filter(function ($value, $key) use ($instance) {
+                    return $instance->hasProperty($key);
+                })->toArray();
         }
     }
 }
--- a/tests/ValidationTest.php
+++ b/tests/ValidationTest.php
@@ -3,6 +3,7 @@
 namespace Tests;
 
 use Illuminate\Contracts\Validation\Rule;
+use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\ViewErrorBag;
 use Livewire\Component;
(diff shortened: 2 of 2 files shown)


calebporzio:
Thanks @mattharcourt!

redbastie:
OK so how do we retain the errors for custom validators?

For example, I prefer using a `state` array so I don't have to declare 100 class properties in my components.

To validate, I use:

	Validator::make($this->state, $this->rules())->validate();

The problem is when a non-deferred model changes, the validation errors disappear.

PR #1407, opened 2020-08-22
Fix temporary file casting while validation errors are present on other properties

### LINKED ISSUE #1122: File upload hydration (and preview) fails when another field has a validation error in an updated-hook
File upload hydration (and preview) fails when another field has a validation error in an updated-hook

I have a component with two fields. A text field an a input type = file field. The text field has validation in a updated() - hook. After uploading a file, and then changing the text field, the updated file is not hydrated anymore. 

After debugging, I found that the HydratePropertiesWithCustomRuntimeHydrators::class is skipped for the incoming request, because the validation throws an error. 

It could maybe be fixed by switching the order of the middlewares. 

#### step 1
Create a component called Example

<pre>

namespace App\Http\Livewire;
use Livewire\Component;
use Livewire\WithFileUploads;

{
    use WithFileUploads;

    public $name;

    public $photo;

    public function updated(){
        $this->validate(['name' => 'min:20']);
    }

    public function render()
    {
        return view('livewire.example');
    }
}

</pre>

And a view with input type="file" wire:model="photo, and a preview (using temporaryUrl() ), and the input for the $name property: <input class="border" type="text" wire:model="name">

<pre>
< div>
< input class="border" type="text" wire:model="name">
        < input type="file" wire:model="photo">
        @if($photo)
            < img src="{{ $photo->temporaryUrl() }}" alt="Photo">
        @endif
 < /div>
</pre>

#### step 2
Upload a file

#### step 3
Change the text field

#### step 4
Watch the exception...

**Context**
- Laravel version: [e.g. 7.11.0]
- Browser: Any.

danharrin:
Closes #1122.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -261,12 +261,12 @@ class LivewireServiceProvider extends ServiceProvider
         /* v */ RegisterEventsBeingListenedFor::class,              /* ^ */
         /* v */ RegisterEmittedEvents::class,                       /* ^ */
         /* v */ PersistLocale::class,                               /* ^ */
-        /* v */ PersistErrorBag::class,                             /* ^ */
         /* v */ HydratePublicProperties::class,                     /* ^ */
         /* v */ HashPropertiesForDirtyDetection::class,             /* ^ */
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks Dan, closing this though - I commented on #1122

PR #1420, opened 2020-08-27
add $reset magic action

mokhosh:
Please include a thorough description of the improvement and reasons why it's useful.

I understand the reason we have magic actions is to avoid having common one line methods.
There is this one line method that I have in many projects:
```php
public function cancelEdit()
{
    $this->reset();
}
```

This little magic action helps me do this instead:
```html
<button wire:click="$reset">Cancel</button>
```

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -91,6 +91,10 @@ trait HandlesActions
                 return;
                 break;
 
+            case '$reset':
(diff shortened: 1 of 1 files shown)


gdebrauwer:
I could be useful to allow parameters as well, if you want to reset specific fields.

```php
<button wire:click="$reset('lastname')"></button>
````

calebporzio:
Going to hold off on this for now. Thanks!

PR #1422, opened 2020-08-27
[V2] Fix file upload in new $wire handler

maxeckel:
The new $wire magic has broken the upload functionality using `@this.upload` as the public API methods on the component (in JS) have to be whitelisted. Every call to a method not whitelisted will be redirected to the class (PHP). 

Because of this, using the `@this.upload` results in an error, because the "upload" method can not be found on the component class (PHP).

With this PR I whitelisted the upload methods `upload`, `uploadMultiple` and `removeUpload`, so that these methods get correctly called on the JS component.

Probably it's too early to send in a PR for V2 and this might already be fixed, if so please ignore/close this PR.


calebporzio:
Thanks @maxeckel!

PR #1427, opened 2020-08-28
Allow components to update full history path (not just query string)

inxilpro:
This is WIP PR — I was running into issues getting the jest tests to run (tried Node 10, 12, and 14 and ran into regeneratorRuntime errors), so the hope is that the tests will work in CI.

```

```

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -75,6 +75,11 @@ abstract class Component
     {
         return $this->updatesQueryString;
     }
+    
+    public function mapStateToUrl()
+    {
+    	return null;
+    }
 
     public function getCasts()
     {
--- a/src/HydrationMiddleware/UpdateHistoryPath.php
+++ b/src/HydrationMiddleware/UpdateHistoryPath.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Livewire\HydrationMiddleware;
+
+use Illuminate\Support\Str;
+use Livewire\Exceptions\CannotCombineHistoryPathAndQueryStringException;
+use Livewire\Exceptions\InvalidHistoryPathException;
+
+class UpdateHistoryPath implements HydrationMiddleware
+{
+    public static function hydrate($instance, $request)
+    {
+        //
+    }
+
+    public static function dehydrate($instance, $response)
+    {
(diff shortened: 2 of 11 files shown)


src/HydrationMiddleware/UpdateHistoryPath.php
```diff
@@ -0,0 +1,27 @@
...
+        //
+    }
+
+    public static function dehydrate($instance, $response)
+    {
+        if (! empty($url = $instance->mapStateToUrl())) {
+        	$root = url()->to('/');
+	        
+        	throw_if(isset($response->updatesQueryString), new CannotCombineHistoryPathAndQueryStringException($instance->getName()));
+        	throw_unless(Str::startsWith($url, $root), new InvalidHistoryPathException($instance->getName()));
```
inxilpro, line 22: This needs to handle relative paths.

src/HydrationMiddleware/UpdateHistoryPath.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+        //
+    }
+
+    public static function dehydrate($instance, $response)
+    {
+        if (! empty($url = $instance->mapStateToUrl())) {
+        	$root = url()->to('/');
+	        
+        	throw_if(isset($response->updatesQueryString), new CannotCombineHistoryPathAndQueryStringException($instance->getName()));
```
inxilpro, line 21: Theoretically, we could merge both the query string and the path, but it's a lot more legwork for a use-case that is unlikely (you'll probably use one or the other).

src/Component.php
```diff
@@ -75,6 +75,11 @@ public function getUpdatesQueryString()
     {
         return $this->updatesQueryString;
     }
+    
+    public function mapStateToUrl()
+    {
+    	return null;
+    }
```
inxilpro, line 82: We could get a performance gain if we also include the props that affect the URL in this response, but I feel like the API that I would want is just return a string and have livewire work.

One solution would be to add an optional property to the component, something like `$urlProps`, that could be used to filter down the props that are associated with the url state (this is the pattern that React uses for memoizing hooks).

PR #1428, opened 2020-08-28
[2.x] Check if Alpine is present

mpociot:
This code currently breaks, when using the Livewire JS component API on a page without Alpine.

This PR fixes that.


calebporzio:
Thanks Marcel!

Issue #1429, opened 2020-08-28
7.4 Typed Properties & Route Param Binding


calebporzio:
This is an outline for two Livewire features:

## #1: Directly assign properties from route parameters without the `mount()` method:

```php
Route::get('/post/{post}, ShowPost::class)
```
```php
class ShowPost extends Component
{
    public Post $post;

    // Notice no mount method needed.
}
```

## #2: Mutate the browser URL at runtime if a route parameter's property changes.

* Given:
```php
Route::get('/post/{post}, ShowPost::class)
```
```php
class ShowPost extends Component
{
    public Post $post;

    public function changePost(Post $newPost)
    {
        $this->post = $newPost;
    }
}
```

A user visits `/post/1`, and then clicks a Livewire button that calls "changePost(2)", the URL now says `/post/2` without ever changing pages. (using pushState).

This should work with ANY route parameters no matter if they are route model bound or not.

DanielCoulbourne:
We just had a long call with @calebporzio, @inxilpro and myself about these features and we're going to be implementing them on our end.

tanthammar:
Just tested it AWESOME!!!

PR #1441, opened 2020-08-30
[2.0] Implicit prop binding

inxilpro:
This is the first part of #1429 — now you can essentially do:

```php
class UserProfileComponent extends Component
{
    public User $user;
    public function render(Router $router)
    {
        return "<div>Hello {$this->user->name}!</div>";
    }
}

Route::get('/users/{user}', UserProfileComponent::class);

// Or…

@livewire(UserProfileComponent::class, ['user' => $user])
```

Leaving this as a draft as I investigate cleaning it up a bit more.


src/HydrationMiddleware/HydratePublicProperties.php
```diff
@@ -2,13 +2,17 @@
...
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Collection;
 use Livewire\Exceptions\PublicPropertyTypeNotAllowedException;
 
 class HydratePublicProperties implements HydrationMiddleware
 {
     public static function hydrate($unHydratedInstance, $request)
     {
-        $publicProperties = $request->memo['data'];
+        $publicProperties = static::removeTypedProperties(
```
calebporzio, line 13: Everything is looking good. My only thought is, is it possible to not have the "model property hydration" concerns leaked to other hydration middleware.

For example, we have to remove them here because we already dehydrated them earlier. Does it make sense to colocate the removal of the properties with the other middleware?

inxilpro: I think we'd have to mess with the hydration middleware order if we wanted to do that. It's just a trade-off re: which is preferable.

inxilpro: I guess we could call `HydrateEloquentModelsAsPublicProperties` before `HydratePublicProperties` and then remove the data associated with Eloquent models from the data array…

PR #1448, opened 2020-09-01
Allow use of custom rule classes in validation

rookwood:
This is a relatively minor change that uses the `rules` _method_ as the sole accessor for the validation rules array.  As the existing method uses a class property, one cannot instantiate a new rule inside it.

```php
class SomeComponent extends Component {
    public $rules = [
        'phone' => ['required', new ValidPhone], // Not valid php
    ];
}
```

The intent here is to allow the use of the `$rules` property if the user wants, but to also allow the `rules` method as a way to make the above possible.


calebporzio:
Thanks!

PR #1449, opened 2020-09-01
Allow psr client to build URL

jimmypuckett:

This is part of #1294 in that it allows the PSR client to set the URL.

No

No

In the current code, when the temp URL is created, it is being manually made, but not taking into consideration `schema` & `port`, so it will not work when running s3 compatable storage on non https.  This just cast `GuzzleHttp\Psr7\Uri` to a string, so it keeps up with the correct path

### DIFF
--- a/src/GenerateSignedUploadUrl.php
+++ b/src/GenerateSignedUploadUrl.php
@@ -35,11 +35,9 @@ class GenerateSignedUploadUrl
             '+5 minutes'
         );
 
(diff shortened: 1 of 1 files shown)


jimmypuckett:
@calebporzio I know you are busy with v2, but any chance you can review this?  We trying to use a minio container locally to develop & test, but cannot with the way that URL's are made.

calebporzio:
Thanks @jimmypuckett !

PR #1464, opened 2020-09-04
Bugfix: re-add window.location.hash to updated url

gdebrauwer:
**What is the problem?**

When the current url has a hash/fragment (e.g. `http://localhost/hello-world#hash`), that hash/fragment disappears when the querystring is being updated.

**How did I fix the problem?**

I updated `UpdateQueryString` js-file to re-add `window.location.hash` to the newly constructed url.

_I did not include a new build of `livewire.js` because building the javascript did not work on my local machine 😔_

### DIFF
--- a/js/component/UpdateQueryString.js
+++ b/js/component/UpdateQueryString.js
@@ -43,6 +43,7 @@ export default function () {
 
         var stringifiedQueryString = queryString.stringify(queryData)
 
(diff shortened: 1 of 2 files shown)


inxilpro:
This logic has been moved to the server in `SupportBrowserHistory` for version 2.0. Would you be open to submitting a PR to persist the fragment in the `master` branch?

gdebrauwer:
Ok 👍 I will try to create a PR on monday

calebporzio:
Thanks! Would still love a PR for the feature on V2 of course, but this is great 👍

Issue #799, opened 2020-04-09
Allow polling to execute as usual in the background


joelhinz:
**Is your feature request related to a problem? Please describe.**
I use and love Livewire for a few things where I want parts of the page to update at set intervals, e.g. personal dashboards at work. Since I rely on some of these checks alerting me when something happens, I need them to be constantly updated even when the tab is open in the background, but the polling script prevents most such calls from being executed. This is currently the only thing stopping me from removing Vue entirely from two of my apps.

**Describe the solution you'd like**
Going from e.g.
`wire:poll.750ms="lookForTickets()"`
to e.g.
`wire:poll.750ms.alwaysOn="lookForTickets()"`
 would solve the problem completely.

**Additional context**
From what I can tell, a simple change to https://github.com/livewire/livewire/blob/master/js/component/Polling.js would be enough, something like this:

```js
// before
if (store.livewireIsInBackground && Math.random() < .95) return

// after
const alwaysOn = directive.modifiers.includes('alwaysOn')
if (!alwaysOn && store.livewireIsInBackground && Math.random() < .95) return
```

I'd be happy to submit a PR myself, but I'm new to that and I don't know how to run and properly test for it locally, so I'd be doing it blind unless there's a guide for that as well. :)

calebporzio:
I could see wanting this.

I'm going to hold off on adding this unless more people submit issues requesting it.

Here's something you can do in the meantime from within your Livewire component to achieve what you're looking for:

```html
<script>
setInterval(() => {
    @this.livewire.call('lookForTickets')
}, 750)
</script>
```

joelhinz:
Oh, that's a great workaround. Thanks! And totally fair. :)

PR #1466, opened 2020-09-04
Added "persist" modifier so it won't throttle polling on focus lost

EmilioBravo:
Added a "persist" modifier to the polling component so it won't throttle the polling on browers window loss of focus. 

Maybe helps with #799


calebporzio:
I believe this logic is inverted? Won't this make it so that adding `.persist` NEVER updates in the background?

Anyhow, I'm open to this, but can you change the modifier to `.keep-alive` instead of `persist`? Thanks!

Issue #1387, opened 2020-08-18
WithPagination removes element from custom paginator template


MaxGiting:
When using a custom pagination template I have seen strange behaviour. Below are three images in chronological order. When I go from page 1 -> 3 -> 1 the word `to` is removed from the custom paginator template. I have linked an example repo below that is a fresh Laravel install with livewire a custom pagination template and nothing else.

**Start on page 1**
<img width="200" alt="page1-correct" src="https://user-images.githubusercontent.com/9828591/90559774-039a2880-e196-11ea-82e1-44605c56f94f.png">

**Click to page 3**
<img width="200" alt="page3-correct" src="https://user-images.githubusercontent.com/9828591/90559781-05fc8280-e196-11ea-91cf-4b6db39685fb.png">

**Click back to page 1 and the word `to` has been removed**
<img width="200" alt="page1-wrong" src="https://user-images.githubusercontent.com/9828591/90559790-09900980-e196-11ea-91d2-0f728f2421ec.png">

I have narrowed it down to the [`custom-paginator.blade.php`](https://github.com/MaxGiting/Livewire-Pagination-Issue/blob/d189055691dc4e89e5934a490bd0639a3ddd4d83/resources/views/custom-paginator.blade.php#L3-L9) file I have used for the pagination.
```
    Showing
    <span>{{ $paginator->firstItem() }}</span>
    to
    <span>{{ $paginator->lastItem() }}</span>
    of
    <span>{{ $paginator->total() }}</span>
    results
```

If I remove the `<span>` tags then the word `of` does not disappear. It does disappear if the word `of` is between any element tags.

### Steps to reproduce
 - Clone [my example repo](https://github.com/MaxGiting/Livewire-Pagination-Issue)
 - `composer install`
 - Connect to a database in `.env`
 - `artisan migrate`
 - `php artisan db:seed --class=ItemSeeder` to populate the database
 - `artisan serve` and open browser
 - Click on the final page number link
 - Now click on any other page link and the word `to` will be removed from the string.

**Context**
- Livewire version: [e.g. 1.3.2]
- Laravel version: [e.g. 7.25.0]
- Browser: [Chrome]

JayBizzle:
Also, when you go from `1` to `3`, if you notice, the space between the `1 and the `2` button disappears

mpbarlow:
[...]

PR #1473, opened 2020-09-06
[2.x] Improve pagination morphdom

timrspratt:
This PR corrects pagination so that `simplePaginate` uses the `paginationSimpleView`.

I've also added `wire:key` to the dom elements, since there is an issue when clicking the pagination links, whereby words disappear. I believe the issue is related to https://github.com/livewire/livewire/issues/1387, however @MaxGiting is using a custom pagination view so would need to add. Keying the buttons also stops the focus remaining on the wrong button when the dom is updated (since the `button` changes to a `span` and so the browser shifts its focus to the previous button, if that makes any sense).

The particular issue I had that this aims to resolve is the word "of" vanishes when going back a page.

1) Initial state
Screenshot 2020-09-06 at 14 01 50

2) Click on page 2 (note button focus shifts back one despite clicking page 2 and not the "previous" button)
Screenshot 2020-09-06 at 14 01 58

3) Click on page 3 (note button focus shifts back one despite clicking page 3 and not page 2)
Screenshot 2020-09-06 at 14 02 06

4) Click back to page 2 (note wording has lost "of")
Screenshot 2020-09-06 at 14 02 14

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -22,7 +22,7 @@ trait WithPagination
         });
 
         Paginator::defaultView($this->paginationView());
-        Paginator::defaultSimpleView($this->paginationView());
+        Paginator::defaultSimpleView($this->paginationSimpleView());
     }
 
     public function paginationView()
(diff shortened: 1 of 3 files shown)


calebporzio:
This is great! Thanks!

PR #1476, opened 2020-09-06
[2.0] Browser history feature clean up

inxilpro:
I took one pass over the browser history implementation and cleaned things up a bit:

- Minor tweak to JS
- Renamed `getFromQueryString` to `getQueryString` to keep naming consistent with the rest of the component API
- Removed `$casts` property and `getCasts()` method because that functionality was removed
- Removed leftover call to an `AllCapsCaster` class
- Refactored `SupportBrowserHistory` for consistent naming (there was still a mix of `$queryStringParams` vs `$queryStringParams` vs `$queryParams` throughout)
- Other general refactors to `SupportBrowserHistory` to remove unnecessary duplication/etc

### DIFF
--- a/js/component/SyncBrowserHistory.js
+++ b/js/component/SyncBrowserHistory.js
@@ -19,7 +19,7 @@ export default function () {
         let { response } = message
         let effects = response.effects || {}
 
-        if ('path' in response.effects && effects.path !== window.location.href) {
+        if ('path' in effects && effects.path !== window.location.href) {
             let state = generateNewState(component, response)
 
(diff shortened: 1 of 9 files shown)


timrspratt:
👍 This should hopefully resolve an issue I'm currently having where Livewire is adding `?` onto the url despite no query params being present when using:

```php
protected $queryString = [
    'search' => ['except' => ''],
];
```

inxilpro:
Cool. I just added a test for that specific case just to be safe :)

calebporzio:
Thanks @inxilpro, this is great!

PR #1480, opened 2020-09-07
Add getEventPayload() for consuming event payloads in tests

jameslkingsley:
This PR adds a method to the `TestableLivewire` class that lets you grab the event payload to use in your tests. Apologies if this is already available in some way - I couldn't find a method to achieve this so I've been using this macro, and thought I'd PR it incase it's beneficial to others.

It allows you to make assertions about the data you've emitted in an event, for scenarios where you're handling it on the client and don't necessarily have that state to call on from the Livewire component directly.

Here's a snippet from a real-world test I use this on:
```php
/** @test */
public function upgrade_modal_completes_successfully()
{
    $user = factory(User::class)->create();

    $livewire = Livewire::actingAs($user)
        ->test(Upgrade\PremiumPlan::class)
        ->call('billingSetupAuthorised', ['payment_method' => 'pm_card_visa'])
        ->assertEmitted('billingSetupComplete');

    /** @var \Illuminate\Support\Carbon $renewal */
    $renewal = $livewire->getEventPayload('billingSetupComplete');

    $this->assertEquals(now()->addMonth()->toDateString(), $renewal->toDateString());
}
```

### DIFF
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -6,6 +6,7 @@ use Mockery;
 use Livewire\Livewire;
 use Illuminate\Support\Str;
 use Illuminate\Routing\Route;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\View;
 use Livewire\GenerateSignedUploadUrl;
 use Illuminate\Routing\RouteCollection;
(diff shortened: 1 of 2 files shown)


calebporzio:
This is already possible:

```php
/** @test */
public function upgrade_modal_completes_successfully()
{
    $user = factory(User::class)->create();

    $livewire = Livewire::actingAs($user)
        ->test(Upgrade\PremiumPlan::class)
        ->call('billingSetupAuthorised', ['payment_method' => 'pm_card_visa'])
        ->assertEmitted('billingSetupComplete', function ($event, $params) {
            return now()->addMonth()->toDateString() === $params[0]->toDateString();
        });
}
```

PR #1484, opened 2020-09-07
Fixes #1483 : error with query string array

### LINKED ISSUE #1483: SupportBrowserHistory error with query string array
### Description

> json_decode() expects parameter 1 to be string, array given
>vendor/livewire/livewire/src/RenameMe/SupportBrowserHistory.php on line 32
> http://myproject.local/nouveautes?f%5B1%5D%5B0%5D=1

I'm encountering an issue when query string parameter is an array.

```
    public $f = [];

    protected $queryString = [
        'f'                => ['except' => false],
    ];
```

In mount()
```
        $this->fill($request->only(array_keys($this->queryString)));
```

And the checkbox : 
```
 wire:model="f.{{ $option_id  }}.{{ $loop->index }}"
```

### Steps to reproduce
- Make a Full-Page Component with a checkbox data binded and querystring updated to a public property 
- Check the box
- Reload the page (with the new url)

**Context**
- Livewire version: 2.0.0
- Laravel version: 7.27.0
- Browser: Chrome

SRWieZ:
Fixes #1483 

Maybe we should create a test for this case.

### DIFF
--- a/src/RenameMe/SupportBrowserHistory.php
+++ b/src/RenameMe/SupportBrowserHistory.php
@@ -29,7 +29,7 @@ class SupportBrowserHistory
                 $fromQueryString = Arr::get($queryParams, $property);
 
                 if ($fromQueryString !== null) {
-                    $component->$property = ($decoded = json_decode($fromQueryString)) === null ? $fromQueryString : $decoded;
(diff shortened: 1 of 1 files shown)


calebporzio:
This has already been fixed in v2.0.2. Thanks for submitting!

PR #1492, opened 2020-09-07
Allow expressions inside the @entangle directive

imliam:
When trying to use the `@entangle()` directive, you currently can't pass through actual PHP expressions because of the operator precedence in the `(object) {$expression}` cast, particularly in this bit of code:

```php
<?php if ((object) {$expression} instanceof \Livewire\WireDirective) : ?>
```

For example, using:

```
@entangle('foo' . 'bar')

// Or something more realistic
@entangle('inputs.' . $input->id)
```

This if statement will essentially result in:

```
(object) 'foo' . 'bar'

// aka with the precedence:
((object) 'foo') . 'bar'
```

This triggers an error when rendering the view and it tries to concatenate these two values together:

```
Object of class stdClass could not be converted to string
```

In order to support actual expressions like this, this pull request simply wraps the expression in parenthesis to change the precedence - so the result of the expression is then cast to an object instead.

```
<?php if ((object) ({$expression}) instanceof \Livewire\WireDirective) : ?>
```

### DIFF
--- a/src/LivewireBladeDirectives.php
+++ b/src/LivewireBladeDirectives.php
@@ -14,7 +14,7 @@ class LivewireBladeDirectives
     public static function entangle($expression)
     {
         return <<<EOT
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @imliam! Good contribution!

PR #1495, opened 2020-09-08
Fix CI test

nuernbergerA:
Alpine L77 was failing only on CI.
We need to simulate a real network speed with some ms delay - maybe we should do this for all internal tests.


calebporzio:
Thanks Adrian!

Issue #1493, opened 2020-09-07
No error when multiple root tags are used


Roboroads:
### Description
I have been guessing for hours why my app wasn't working - until I remembered that Vue.JS wasn't happy with more than one root tag - wrapping everything in a div resolved the issue. Unsure if this is intended or not since I haven't found this yet in the documentation (or read over it 15 times - in which case excuse me). If this is intended, there should be some kind of error/warning.

### Steps to reproduce
Create a new livewire component with multiple root tags like so:

```php
// Reproduce.php
class Reproduce extends Component
{
    /** @var string */
    public $someText = '';
}
```

This works:
```html
<!--reproduce.blade.php-->
<div>
    <div>
        <input wire:model="someText">
    </div>
    <div>
        <input wire:model="someText">
    </div>
</div>
```

This does not:
```html
<!--reproduce.blade.php-->
<div>
    <input wire:model="someText">
</div>
<div>
    <input wire:model="someText">
</div>
```

**Context**
- Livewire version: 2.0.3
- Laravel version: 7.27.0
- Browser: Chrome

Note: I do have some time to look into creating a warning for this and PR-ing it - but would like to know what is intended before I put time into it :smile:

mblarsen:
Yes, this is a known limitation. I'm not sure that anything has changed for version 2; but I think that it is being addressed. You need to have a single root for now.

nuernbergerA:
It's still unsupported

damms005:
As long as this remains unsupported, it will be helpful if livewire throws an exception message when there are more than one root tags.

Also, it will be more helpful if the exception message says "...read more at https://laravel-livewire.com/docs/1.x/rendering-components#returning-blade"

This will help debug issues faster, as I slept on a bug and just waking up to see this multi-root tag issue was the thing

Roboroads:
Ah!! There it is, i was sure it should have been somewhere in the docs and i read over it. I will see if i have some time today to create a PR to add a warning.

calebporzio:
Closing (See my comment on the PR) - Thanks!

PR #1499, opened 2020-09-08
Adds error when multiple root tags are present

Roboroads:
#1493  

Nop.

Yep!

Someone who did not thoroughly read the documentation (like me - sorry) an error like this can save hours of debugging time. Since it just silently fails right now although it's intended behaviour. 

You're welcome :smile:


Roboroads:
Busy fixing the tests ;)

Roboroads:
NOTE: This test fails on 2.0.4/master as well - this will be fixed with #1495

Roboroads:
Since #1495 got merged - tests are now resolved :smile:

calebporzio:
Thanks for this PR.

I totally understand the need for this, but I really don't want to add extra processing to every single Livewire request for this case.

Going to hold off for now and just try to make the docs better (would love a PR for that)

Thanks!

damms005:
@calebporzio Since such extra processing is a concern, something like `dev mode` can be made available (like Vue JS does)

calebporzio:
Ah, interesting. Like only check when app debug is true or something. I could get with that

Roboroads:
I can change the PR to include this!

Issue #1498, opened 2020-09-08
Updating to v2 throws error with laravel/framework < 7.18


erikgaal:
### Description
Commit 1adc05a added a change that should have increased the version requirement of `laravel/framework` to ^7.18` as this code was only added after that version (https://github.com/laravel/framework/pull/33354).

As a developer upgrading to Livewire v2, you have to figure out yourself that this is due to your Laravel version being behind. Unfortunataly, v2 has already been tagged, so changing the requirement now would still result in the following error. Instead we should document this here so that others will find this and know how to resolve.

### Steps to reproduce
```$ composer require livewire/livewire:^2.0
Using version ^2.0 for livewire/livewire
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
...
Writing lock file
Generating optimized autoload files
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover --ansi

   Error 

  Call to undefined method Illuminate\View\ComponentAttributeBag::macro()

  at vendor/livewire/livewire/src/LivewireServiceProvider.php:196
    192|     }
    193| 
    194|     protected function registerViewMacros()
    195|     {
  > 196|         ComponentAttributeBag::macro('wire', function ($name) {
    197|             $entries = head($this->whereStartsWith('wire:'.$name));
    198| 
    199|             $directive = head(array_keys($entries));
    200|             $value = head(array_values($entries));

      +9 vendor frames 
  10  [internal]:0
      Illuminate\Foundation\Application::Illuminate\Foundation\{closure}(Object(Livewire\LivewireServiceProvider))

      +5 vendor frames 
  16  artisan:37
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
Script @php artisan package:discover --ansi handling the post-autoload-dump event returned with error code 1

Installation failed, reverting ./composer.json to its original content.
```

### How to resolve
Upgrade to laravel/framework ^7.18 before upgrading to Livewire v2.

```
$ composer require laravel/framework:^7.18
```

**Context**
- Livewire version: 2.0.0
- Laravel version: <7.18

nuernbergerA:
Feel free to PR

calebporzio:
I fixed updated to only macro if macro is available. v2.1.1 should fix this. Thanks!

PR #1506, opened 2020-09-08
Install dependencies with prefer-lowest

erikgaal:
Adding `prefer-lowest` to the build matrix will discover issues like #1498 in CI.

Can be merged after #1507.


calebporzio:
Good call. As this doubles the amount of builds in the matrix (slowing it down), I'm going to hold off for now. Thanks though, will likely add this back in the future.

PR #1513, opened 2020-09-09
fix: decode query string to associative array instead of stdClass

cappuc:
this fix a problem with query string array properties during component initial rendering: it encode the value and decode it but it's decoded to sdtClass instead of array and it throw an exception


calebporzio:
Thanks!

PR #1517, opened 2020-09-09
Add special condition for spacebar to wire:click

### LINKED ISSUE #1516: Spacebar doesn't work with wire:keydown
### Description

wire:keydown.space doesn't work.

It's because it looks at the KeyboardEvent.key which for spacebar is ` `.

### Steps to reproduce

**Context**
- Livewire version: 2.1.2
- Laravel version: 8
- Browser: Chrome

dubcanada:

Fixes #1516 

Nope

Yes

Spacebar is weird and it's handled completely different then the other keyboard events, rather then returning a name, the key is literally an empty space. Which is not possible within the current html attributes. So we add a little exception to fix this.


dubcanada:
Not sure why the test is failing, it seems to be related to the Alpine test, not the new spacebar test.

calebporzio:
Thanks!

PR #1518, opened 2020-09-09
Adds the ability to define Validation Messages via  property

iksaku:
* Nope, just went directly to implementation

* Single feature

* Yup, and they're working (at least on my end)

As PR title says, this feature is intended to mirror the new `$rules` property for validation introduced in v2, but instead of working with rules, now we work with validation messages.

This will help people that abstract component logic (like myself) in traits, so they can share, in this case, most of the validation logic between similar components, without needing to implement custom validation functions nor global validation messages.

Just, define your `$rules` and your `$messages` in a trait, then call `$this->validate()` on your actions and you're ready to go!

PS: I added this feature because most of my work is done in Spanish, and I love to give context on why the validations fail, so I kept creating my own `validateData()` functions in order to provide custom messages without overriding the core `validate()` function, nor creating generic app-wide messages. So, that's the true background for this PR.

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -66,6 +66,14 @@ trait ValidatesInput
         return [];
     }
 
+    protected function getMessages()
+    {
+        if (method_exists($this, 'messages')) return $this->messages();
+        if (property_exists($this, 'messages')) return $this->messages;
+
+        return [];
+    }
+
     public function rulesForModel($name)
     {
         if (empty($this->getRules())) return collect();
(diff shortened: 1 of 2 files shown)


calebporzio:
I like this. Perfect PR. Thanks ❤️

jimpeters:
Thank you!  Now I can delete my crazy method override to accomplish the same thing.  Have a great weekend!

ninthspace:
Should we do the same with `validateOnly()` since this has the same pattern of behaviour?

iksaku:
> Should we do the same with `validateOnly()` since this has the same pattern of behaviour?

Oh god, surely we need to, I missed that part. Let me open another PR.

Thanks for the heads up!

PR #1522, opened 2020-09-10
Fixed bootstrap pagination button type

davidrushton:
I ran into an issue using a Livewire component with pagination inside an HTML form.

Since the pagination `<button>` tags don't have a type attribute, they default to `submit` and therefore try to submit the form / hit HTML5 validation in Chrome.

This PR adds `type="button"` to the pagination buttons to stop them from submitting an outer form.

Thanks for Livewire 🙌


calebporzio:
Perfect, thanks!

PR #1534, opened 2020-09-10
Add option to select a node to rescan

imliam:
This PR simply adds the ability to pass a DOM node to the `window.Livewire.rescan()` function, which will be used as the root element to scan - falling back to `document` as the root node if none is supplied (the current behaviour).

This means that when rescanning components, we can be a bit more selective about what components are scanned and initialised - particularly when we need to manually do so in some use cases.

The use case in my example was that I had some Livewire components inside an Alpine `<template x-if="">` - and due to the contents of the template tag not being in the DOM at initial page load, the Livewire components never actually get initialised. This change would allow the following block of code to work, without affecting other components outside the template:

```
<template x-if="true">
    <div x-data x-init="window.Livewire.rescan($el)">
        <livewire:component />
    </div>
</template>
```


calebporzio:
Perfect, thanks Liam!

Issue #1482, opened 2020-09-07
@v2 Component Pagination causes 404 error


JustSteveKing:
### Description

Using Laravel 7.24 and Livewire 2.0 after following the upgrade guide, Pagination links now open a modal window with a 404 error when hitting the route: `/livewire/message/`

### Steps to reproduce

Upgrade following the documented approach on the Livewire website, do not upgrade anything else.  Click on a pagination link to go to the next page. You should see a modal window pop up that contains the standard 404 error page for Laravel. You will also notice that the component endpoint is no longer called, instead livewire/message is called which causes the 404 itself.

**Context**
- Livewire version: 2.0
- Laravel version: 7.24
- Browser: Chrome

JustSteveKing:
This is my own fault 🤦 kinda

A step that needs adding to the upgrade guide is to also ensure you republish livewire assets:

```bash
$ php artisan livewire:publish
```

This has solved my problem. I will leave this for @calebporzio to close when ready

JustSteveKing:
I have craeted a PR in the docs to add this step now: https://github.com/livewire/docs/pull/219

calebporzio:
Thanks for the solution. I added it to the upgrade guide.

coolsam726:
> ### Description
> 
> Using Laravel 7.24 and Livewire 2.0 after following the upgrade guide, Pagination links now open a modal window with a 404 error when hitting the route: `/livewire/message/`
> ### Steps to reproduce
> 
> Upgrade following the documented approach on the Livewire website, do not upgrade anything else. Click on a pagination link to go to the next page. You should see a modal window pop up that contains the standard 404 error page for Laravel. You will also notice that the component endpoint is no longer called, instead livewire/message is called which causes the 404 itself.
> 
> **Context**
> 
>     * Livewire version: 2.0
> 
>     * Laravel version: 7.24
> 
>     * Browser: Chrome

The issue is still persistent on my end, but ONLY when I am serving my project from a sub-path of the root domain.

coolsam726:
This issue is linked to   #1535

calebporzio:
Closing this as duplicate, thanks!

PR #1544, opened 2020-09-11
Fix 404 Error when making ajax Requests in projects hosted on a path in the domain

### LINKED ISSUE #1535: 404 Error when hosting the project on a subpath
### Description
I have tried both livewire ^2 on Laravel 7 and a fresh installation of Laravel 8. When I serve my project from the domain root everything works fine. However when I host my project on a sub-path of the domain i.e myurl.com/my-app, EVERY POST operation fails and raises a modal with the 404 error page. However the data that is fetched from the component is fine. The problem is posting to `/myproject/livewire/message/[component]`

### Steps to reproduce
1. Install livewire
2. Publish assets: `php artisan livewire:publish`
3. Set the asset_url in livewire config: 
```
...
'asset_url'  => env('APP_URL'),
...
```
4. Create a new livewire component: `php artisan make:livewire users.index`
5. Create an alias for the project in your web server.
6. Serve the project from a sub-path e.g `http://myurl.com/myapp`

**Context**
- Livewire version: 2.1.3
- Laravel version: 7.28.1 and 8.x
- Browser: Firefox

coolsam726:
This PR fixes the issue of having 404 errors when the asset_url setting is set to a sub-path, especially for projects NOT hosted on the root domain.

Fixes #1535 , #1482

### DIFF
--- a/src/RenameMe/SupportBrowserHistory.php
+++ b/src/RenameMe/SupportBrowserHistory.php
@@ -51,8 +51,16 @@ class SupportBrowserHistory
 
             if (! $referer = request()->header('Referer')) return;
 
+            $assetUrl = config("livewire.asset_url");
+            // Get the path for sure;
+            $path = parse_url($referer,PHP_URL_PATH);
+            // Remove the asset_url segment
+            if ($assetUrl) {
+                $path = Str::replaceFirst($assetUrl,"",$path);
+            }
+
             $route = app('router')->getRoutes()->match(
-                Request::create($referer, 'GET')
+                Request::create($path,"GET")
             );
 
             $queryParams = $this->mergeComponentPropertiesWithExistingQueryParamsFromOtherComponentsAndTheRequest($component);


calebporzio:
Thanks @coolsam726 for looking into this so deeply.

Here are my thoughts.

I merged this, but changed the code significantly on master: https://github.com/livewire/livewire/commit/21c9c32d664cd83199119c7b100a4bcdccf417df

Take a look.

I don't think it's a good idea to look at the "asset_url" for reformatting the "referer".

My reasoning is that it'd be kind of a hidden behavior: That livewire helps Laravel understand a route match by using the "asset_url" because MOST of the time the asset_url is the same as the sub-path URL.

For example, for you, you are setting `env('APP_URL')` to it in the config, which means it's a full URL. However, the examples show just paths or even urls without schemes (like `hey.com/something`).

In these cases, your parsing would break down and cause odities.

I think for now it's better to just attempt to get the route from the referer, if we can't, just use the current route.

All that means is that you won't get a very specific feature that most people don't use anyways.

I'm going to tag this tomorrow, and I imagine all your problems will go away.

Thanks!

coolsam726:
> Thanks @coolsam726 for looking into this so deeply.
> 
> Here are my thoughts.
> 
> I merged this, but changed the code significantly on master: [21c9c32](https://github.com/livewire/livewire/commit/21c9c32d664cd83199119c7b100a4bcdccf417df)
> 
> Take a look.
> 
> I don't think it's a good idea to look at the "asset_url" for reformatting the "referer".
> 
> My reasoning is that it'd be kind of a hidden behavior: That livewire helps Laravel understand a route match by using the "asset_url" because MOST of the time the asset_url is the same as the sub-path URL.
> 
> For example, for you, you are setting `env('APP_URL')` to it in the config, which means it's a full URL. However, the examples show just paths or even urls without schemes (like `hey.com/something`).
> 
> In these cases, your parsing would break down and cause odities.
> 
> I think for now it's better to just attempt to get the route from the referer, if we can't, just use the current route.
> 
> All that means is that you won't get a very specific feature that most people don't use anyways.
> 
> I'm going to tag this tomorrow, and I imagine all your problems will go away.
> 
> Thanks!
I have looked at your commit. This is just brilliant. Thank you.

jamesbhatta:
Thank you @coolsam726 for your pull request. It saved my project.

PR #1545, opened 2020-09-11
Feature/persist polling

EmilioBravo:
Added "keep-alive" directive to polling so it will keep polling when the browser window is in the background


calebporzio:
Thanks for this!

EmilioBravo:
> Thanks for this!

Thank you for this great package!

Issue #1549, opened 2020-09-11
Livewire 2.0 query string state management alphabetizes query strings even when no query string manipulation is happening; causes SEO errors


erikwestlund:
### Description

In Livewire 2.0, if you have a livewire component on a page, even if you are not manipulating query strings in any way, Livewire will alphabetize query strings on page load. 

Unless Livewire is working with query strings, I do not think it should be manipulating them. I also do not think it should be alphabetizing for SEO reasons. I work on a legacy message board which we've incorporated Livewire into where query strings are used for both pagination and for differentiating threads. Therefore, query string order is important for SEO, as we have concerns over duplicate content if the order changes.

Our links look like this. They are generated using Laravel's standard pagination (Not Livewire's)

domain.com/forum?thread=1&page=2

When you load that page, even though we do nothing with the query string, Livewire changes it to:

domain.com/forum?page=2&thread=1

The change is also visible.

**I can also confirm that Google is able to learn about these different query string-ordered URLs, as we have already received duplicate content errors in our dashboard. As this could affect SEO of users, I think this should be high priority to fix.** 

Regardless, the changing order of query strings without clicking is distracting and makes it feel like something is broken. I do not think it should be re-ordering query strings by default, especially when we're not using the $queryString property in any of the loaded Livewire components.

Video attached shows this in action.

### Potential fix

In SyncBrowserHistory.js, these two methods:

```
    store.registerHook('component.initialized', component => {
        let state = generateNewState(component, generateInitialFauxResponse(component))
        let url = initializedPath ? undefined : component.effects.path

        history.replaceState(state, '', url)
        initializedPath = true
    })

    store.registerHook('message.processed', (message, component) => {
        if (message.replaying) return

        let { response } = message
        let effects = response.effects || {}

        if ('path' in effects && effects.path !== window.location.href) {
            let state = generateNewState(component, response)

            history.pushState(state, '', effects.path)
        }
    })
```

get run once a component is initialized and/or a message is processed.

If I remove the `history.replaceState`  and `history.pushState` code from the published vendor assets, this problem goes away. Obviously, this also breaks page tracking. 

Of note, I need to remove *both* of those lines to make it stop doing this. I believe this is because any time a livewire message gets received, the query state gets pushed.

I think the only solution as currently architected is to make the `component.effects.path` property not alphabetize query strings.

I believe the code tries not to replace query strings when the URL hasn't changed:

`if ('path' in effects && effects.path !== window.location.href) `

[...]

PR #1553, opened 2020-09-11
Update SupportBrowserHistory.php

erikwestlund:

There is an issue here:

https://github.com/livewire/livewire/issues/1549

No. 

No tests.

The `sortKeys()` method call in the collection pipeline within the `mergeComponentPropertiesWithExistingQueryParamsFromOtherComponentsAndTheRequest()` method causes query strings to be re-written in alphabetical order on any page with multiple query strings in the URL.

This should not be default behavior. Sites that rely upon query strings for locating separate resources can receive duplicate content SEO warnings from this behavior. We have confirmed this side effect in production on a high traffic message board with Google Search.

### DIFF
--- a/src/RenameMe/SupportBrowserHistory.php
+++ b/src/RenameMe/SupportBrowserHistory.php
@@ -134,8 +134,7 @@ class SupportBrowserHistory
             })
             ->map(function ($property) {
                 return is_bool($property) ? json_encode($property) : $property;
-            })
-            ->sortKeys();
+            });
 
         return $this->mergedQueryParamsFromDehydratedComponents;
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks @erikwestlund!

calebporzio:
This will be out in the next version

erikwestlund:
Awesome, thanks! Glad to contribute! Looking forward to removing this gem from my deploy script 

`sed -i "s/->sortKeys()//g" vendor/livewire/livewire/src/RenameMe/SupportBrowserHistory.php` 😂

PR #1562, opened 2020-09-12
Wrap pagination text in spans

### LINKED ISSUE #1558: Morphdom issue with tailwind pagination
### Description

When navigating pages, the wording is being lost.

### Steps to reproduce

Note the word "of" vanishes when returning from the final page. It seems to only be certain numbers it triggers on though.

Screen Recording 2020-09-12 at 09 37 31

(Apologies for the gif being super slow 🙈)

**Context**
- Livewire version: 2.2.1
- Laravel version: 8.1.0
- Browser: Chrome 85

imliam:
Fixes #1558

With freeform text mixed with tags inside an element, Morphdom occasionally removes some of the text as it fails to diff it properly. Wrapping the free text in elements like spans helps prevent this.

This is the same fix we had in the TALL preset a while back too: laravel-frontend-presets/tall#45

### DIFF
--- a/src/views/tailwind.blade.php
+++ b/src/views/tailwind.blade.php
@@ -30,13 +30,13 @@
             <div class="hidden sm:flex-1 sm:flex sm:items-center sm:justify-between">
                 <div>
                     <p class="text-sm text-gray-700 leading-5">
-                        {!! __('Showing') !!}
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks Liam!

PR #1568, opened 2020-09-12
Allow 'validationOnly' to use validation messages property

iksaku:
- Was [commented](https://github.com/livewire/livewire/pull/1518#issuecomment-691505392) on a previous PR.

- Straight implementation

- Simple working test :star2: 

- The #1518 pull added the ability to define a `protected $messages` property that allowed validation function to use those messages, but it only worked for the `validate()` function. I missed the `validateOnly()` one, but it was brought to my attention after the PR was merged. So, this is to fix that :sweat_smile: 


calebporzio:
Perfect! Thanks!

PR #1574, opened 2020-09-13
Cleanup CI

nuernbergerA:


calebporzio:
Thanks Adrian!

I'm curious, why were you able to remove the orchestral testbench version in the matrix?

composer.json
```diff
@@ -20,8 +20,8 @@
         "mockery/mockery": "^1.3.1",
         "phpunit/phpunit": "^8.4|^9.0",
         "laravel/framework": "^7.0|^8.0",
-        "orchestra/testbench": "^5.0",
-        "orchestra/testbench-dusk": "^5.2",
+        "orchestra/testbench": "^5.0|^6.0",
```
nuernbergerA, line 23: Because we added it here

PR #1577, opened 2020-09-13
remove spans from bootstrap pagination view

deanzod:
didn't create an issue as it was just as quick to fix it
no
no
Bootstrap pagination was being output with spans wrapping the li tags which caused the layout of the numbers to break. Simply removed the spans leaving a clean list ul>li


calebporzio:
Thanks!

PR #1580, opened 2020-09-14
add property check on reset

redbastie:
its useful

no

no

this will ensure a property exists when included in the reset function

youre welcome

### DIFF
--- a/src/ComponentConcerns/InteractsWithProperties.php
+++ b/src/ComponentConcerns/InteractsWithProperties.php
@@ -163,9 +163,11 @@ trait InteractsWithProperties
         }
 
         foreach ($properties as $property) {
-            $freshInstance = new static($this->id);
+            if ($this->hasProperty($property) {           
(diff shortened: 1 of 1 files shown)


src/ComponentConcerns/InteractsWithProperties.php
```diff
@@ -163,9 +163,11 @@ public function reset(...$properties)
         }
 
         foreach ($properties as $property) {
-            $freshInstance = new static($this->id);
+            if ($this->hasProperty($property) {
```
nuernbergerA, line 166: This is invalid PHP code you are missing a `)`
```php
if ($this->hasProperty($property)) {
```

nuernbergerA (changes requested):
This is invalid PHP code you are missing a `)`

calebporzio:
I'm closing this. I think it's ok to error out. Also the PR doesn't have sufficient reasoning behind it, tests, or valid PHP.

Issue #1591, opened 2020-09-14
Issue when testing component that has temporaryUrl except specifc extensions


jimmypuckett:
### Description

Geting an error on temporaryUrl for sideloaded pdfs

```
Tests\Feature\Forms\FileUploadTest::it_allows_specific_file_types with data set "pdf" ('pdf', 'create', true)
ErrorException: This driver does not support creating temporary URLs. (View: /some/path/resources/views/components/forms/upload.blade.php) (View: /some/path/resources/views/components/forms/upload.blade.php)
```

### Steps to reproduce

We have a blade that has a preview of the sideloaded file...

```php
@if($upload and ! $errors->has('upload'))
                <embed
                    src="{{ $upload->temporaryUrl() }}"
                    type="{{ $upload->getMimeType() }}"
                    width="100%"
                    height="500px"
                    class="my-4"
                />
@endif
```

Since the `env` is testing, then is goes into the check to see if supported type, which it is not, so it has passes the request to storage...

https://github.com/livewire/livewire/blob/8247b1fd7221624ce2d58a08135a68666baea9d9/src/TemporaryUploadedFile.php#L79

I know that I can mock all of this out, but that just seems fragile as I have to know the exact methods inside of livewire that is calling to storage to mock out.

If `pdf` was added to the built in supported files, this would fix my issue, but that is not really a good fix for other extensions.

What is the preferred solution?

**Context**
- Livewire version: 1.3.3
- Laravel version: 7.28.1
- Browser: CLI

jimmypuckett:
What if the `supportedPreviewType` was set in the config?  Then a user could override?

PR #1593, opened 2020-09-14
Move supported preview types to config

jimmypuckett:

Yes, it is here...

#1591 

Nope

Updated existing

Allow people to controll what they want to allow

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -56,23 +56,6 @@ return [
     */
 
     'middleware_group'  => 'web',
-    
-    /*
-    |--------------------------------------------------------------------------
-    | Supported Preview Types
-    |--------------------------------------------------------------------------
-    |
-    | Only allow the the following filetypes when generating 
-    | pre-signed temporary urls for files on local storage.
-    |
-    */
-
-    'supported_preview_types' => [
-        'jpeg', 'png', 'gif', 'bmp', 'svg', 'webp',
-        'mp4', 'mov', 'avi', 'wmv',
-        'mp3', 'mpga', 'wav', 'm4a', 'wma',
-    ],
-
 
     /*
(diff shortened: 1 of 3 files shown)


calebporzio:
Hey @jimmypuckett, couple notes:

This makes sense, however:
* If we push this, it will fail because people already have published config files and updating this repo's config file won't automatically update theirs - you will need to provide a fallback as a "default" parameter in `config()->get()`
* I think this belongs in the upload subsection of the config as the last parameter as it strictly deals with that system and doesn't make sense to stand on it's own
* In your test, I would hardcode the expectation rather than loading it directly from the config file - less dynamicism in tests the better

Just some notes for you, thanks!

jimmypuckett:
@calebporzio A few clarifications...

1) Good catch, I will address that.  (Have a question about the how, so please see below)
2) I toyed with that location, so I will move it
3) I almost put the array in here to not duplicate the array, but I wanted a single source of truth on the expected array.  I was concerend that someone would remember to change the config but not the test.

I can dupilcate the array in 1 & 3 or I can make 1 use the require from the configs like I do in the test.

Whcih do you prefer & I will make it so?

calebporzio:
Yeah, so actually you can just remove the code you added to the tests, and ideally add a new test that tests that you can't preview a certain extension until you add it to the config.

jimmypuckett:
OK, @calebporzio, I finally had some time to make those changes that you suggested.  Please let me know if there is anything else that you would like to see changed.  If/when this gets merged, then I will port it to the 2.x branch.

Thanks!

calebporzio:
Thanks for this! This is perfect!

calebporzio:
Also, no need to re-PR to master - already merged up.

Issue #1602, opened 2020-09-15
Downloads with special characters in filename not downloading with correct filename


jimpeters:
### Description

When using response()->streamDownload(function() { }, 'filename.txt') the browser downloads the file with the correct filename of "filename.txt".

When using any special characters like ampersand Chrome will prepend and append underscores to the filename rendering it unopenable.  I know this is the download attribute of a hidden A element and not a Content-Disposition header.  

I have tried to htmlspecialchars the filename from the backend and still can't make it download with the correct filename.

Any help would be appreciated.  Thanks!

### Steps to reproduce

https://laravelplayground.com/#/snippets/d29b28f0-fefb-4567-90d1-2fbeba8ab8a3

After clicking on the 'Download Test File' button you should see a download in Chrome with a filename called "\_This & That.txt\_" instead of "This & That.txt"

**Context**
- Livewire version: [2.2.3]
- Laravel version: [7.27.0]
- Browser: [Chrome]

jimpeters:
Found the issue.  Working on a pull request now.

PR #1604, opened 2020-09-15
Support quoted Content-Disposition filenames for downloads

jimpeters:

Needed.  Yes. #1602 

No.

Yes

You are unable to download a file in Livewire using a filename like "This & That.txt".  The contents will download but the filename will be surrounded by quotes in the JSON response to Livewire causing weird filename behavior on the saved computer.  Quotes will be replaced with underscores in the filename, etc.

The Content-Disposition header should contain quotes around the filename when special characters or spaces are involved but it is not required when dealing with simple filenames.  It seems Laravel responds with and without quotes depending on the situation.

Content-Disposition: attachment; filename=filename.jpg
Content-Disposition: attachment; filename="filename.jpg"

Are both valid.

### DIFF
--- a/src/RenameMe/SupportFileDownloads.php
+++ b/src/RenameMe/SupportFileDownloads.php
@@ -20,7 +20,13 @@ class SupportFileDownloads
 
             $response = $returned;
 
-            $name = Str::after($response->headers->get('Content-Disposition'), 'filename=');
+            preg_match(
+                '/^.*?filename=(?:(?:(["\'])([^"\']+)\1)|([^"\';]+))/m',
+                $response->headers->get('Content-Disposition'),
+                $matches
+            );
+
+            $name = (count($matches) == 3) ? $matches[2] : $matches[3];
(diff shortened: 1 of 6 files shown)


src/RenameMe/SupportFileDownloads.php
```diff
@@ -20,7 +20,13 @@ function __construct()
 
             $response = $returned;
 
-            $name = Str::after($response->headers->get('Content-Disposition'), 'filename=');
+            preg_match(
```
calebporzio, line 23: Please extract this to a separate function, and maybe add a comment to accompany the regex expression - it's not obvious at first glance

jimpeters: Here is my first attempt.  I also simplified the regexes for more readability since we are in a new method now.

calebporzio: This is much better, thanks!

Issue #1605, opened 2020-09-15
iOS Safari cannot download files


jimpeters:
### Description

iOS Safari cannot download files.  Inspecting a simulator of iPad/iPhone in Safari I see the following when Livewire Javascript attempts to create the hidden link containing the createObjectURL.

Failed to load resource: The operation couldn’t be completed. (WebKitBlobResource error 1.)

I have attempted to find solutions to this and am unable to come up with a competent answer.

### Steps to reproduce

https://laravelplayground.com/#/snippets/d29b28f0-fefb-4567-90d1-2fbeba8ab8a3

Try to download the file on an iOS device.

**Context**
- Livewire version: [e.g. 2.2.3]
- Laravel version: [e.g. 7.27.0]
- Browser: [iOS Safari]

calebporzio:
Fixed in merged PR

PR #1606, opened 2020-09-15
iOS 13 Safari download issue

jimpeters:

Wanted downloads in iOS devices.  Yes.  #1605 

No

No new tests other than regular file download testing that already exists.

iOS was unable to download files using Livewire.  It was suggested that revoking the object URL may be happening too fast and to set a setTimeout of zero for that call.  

Tested with using window.URL it did not work. 

Only window.webkitURL to create the objectURL worked in iOS and Desktop Chrome/Safari/Firefox


calebporzio:
Thanks!

Issue #1515, opened 2020-09-09
It cannot find component by id


VahanDrnoyan:
Hello,

I have doubts if this is an issue or misconfiguration by me, however, I have implemented all according to the documentation.

I am trying to implement a modal opening and closing with @entangle.

An example from the documentation is this:

<img width="844" alt="Screenshot 2020-09-09 at 19 47 22" src="https://user-images.githubusercontent.com/23120503/92630122-60de5100-f2e0-11ea-9f1b-0bd8f8704d32.png">

Here is how I use the blade component:

<img width="523" alt="Screenshot 2020-09-09 at 19 35 53" src="https://user-images.githubusercontent.com/23120503/92630151-718ec700-f2e0-11ea-980b-cdd2d9c0a7c7.png">

Here is how the x-data is used:

<img width="822" alt="Screenshot 2020-09-09 at 19 36 32" src="https://user-images.githubusercontent.com/23120503/92630228-92efb300-f2e0-11ea-8e0d-759920a3b1a0.png">

Here is the error I get:

<img width="822" alt="Screenshot 2020-09-09 at 19 37 02" src="https://user-images.githubusercontent.com/23120503/92630278-a9960a00-f2e0-11ea-8461-ed7ea3e4d2d0.png">

Here is the code from which the error originates, within Alpin js code.

<img width="822" alt="Screenshot 2020-09-09 at 19 37 34" src="https://user-images.githubusercontent.com/23120503/92630380-cdf1e680-f2e0-11ea-8de7-8a170a9321c6.png">

Here is how it looks in the elements panel:

<img width="844" alt="Screenshot 2020-09-09 at 19 44 51" src="https://user-images.githubusercontent.com/23120503/92630391-d64a2180-f2e0-11ea-81f2-5cbdc20e2354.png">

**Context**
- Livewire version: [e.g. 2.0.0]
- Laravel version: [e.g. 8.0.0]
- Browser: [Chrome]

Thank you!

ryatkins:
What does your Livewire component look like and does it have showNoTriggerModal as a public property?

Perhaps you want to break it down to a simpler example without the Blade components and attributes passing.

VahanDrnoyan:
Hi, @ryatkins !

 Yes here is my public property:

<img width="566" alt="Screenshot 2020-09-09 at 21 59 58" src="https://user-images.githubusercontent.com/23120503/92636101-e1548000-f2e7-11ea-8f90-c0dd881f497e.png">


[...]

PR #1609, opened 2020-09-15
Warn alpine not deferred

jimpeters:

Not sure.  I discussed it on #1515 

No

No

I started using the @entangle blade directive in my Alpine components and kept running into exceptions from window.Livewire.find() returning undefined even though the ID passed by entangle was correct according to the DOM.

An hour or two (I'm slow) of banging my head on the desk I found out I had originally included alpine.js BEFORE Livewire and the deferLoadingAlpine callback was not being used as the DOMContentLoaded events are run in the order they are defined.  This caused my Alpine component with @entangle to initialize and run before Livewire was completed initializing the component.  Switched the script order around and no issues since.  

This little warn would of kept me sane that day.  I'm not sure this is the right place for it but this is the only method that makes sense to me.  I don't know how to do a backtrace in JS and verify the find call came from Alpine to only warn then.


calebporzio:
I agree there should be some guidance for this case, however, I'm not sure this is the place.

Livewire.find() is a popular function used often and there are many other reasons it might fail. Suggesting that it's likely the cause of Alpine is misleading I think.

I think documentation can help this problem a bit, and maybe there is a good place to throw a warning (which I'm open to), but I don't think that's here, so I'm going to close this for now.

Feel free to submit a new one.

Thanks!

PR #1627, opened 2020-09-17
Add assertListeningFor

ahinkle:
This PR adds `assertListeningFor($string)` to components; Allowing developers to ensure the event is listening on the component.

We recently ran into an issue where an event was being emitted from a different component but the listening components property `$listeners` was inadvertently removed. The supplied value to `assertListeningFor` will check component payload to ensure the value is within the array of listeners.

Example:
```php
public function test_listens_for_refresh_shipping_methods()
{
    Livewire::test(ShippingMethods::class)
        ->assertListeningFor('refreshShippingMethods');
}
```

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -267,4 +267,13 @@ trait MakesAssertions
 
         return $this;
     }
+
+    public function assertListener($listener)
+    {
+        $test = collect($this->payload['effects']['listeners'])->contains($listener);
+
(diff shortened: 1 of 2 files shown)


calebporzio:
What do you think about `->assertListeningFor('event-name')` instead?

ahinkle:
Updated 👍

calebporzio:
Thanks @ahinkle, however, given the current implementation, will the test still pass if you have a `->call('$refresh')` in there before the assertion?

I believe the listeners ARENT send with the payload effects on every response - which would mean this assertion would provide a false negative in those cases

calebporzio:
Closing, feel free to re-submit after my feedback @ahinkle - thanks!

PR #1633, opened 2020-09-17
Adds error in debug mode when multiple root tags are present

Roboroads:
#1493  - this was an earlier PR as well, made an edit to onlyt do the check in debug mode. See https://github.com/livewire/livewire/pull/1499#issuecomment-694330337

Nop.

Yep!

Someone who did not thoroughly read the documentation (like me - sorry) an error like this can save hours of debugging time. Since it just silently fails right now although it's intended behaviour. 

You're welcome :smile:


calebporzio:
Is there any other good way of testing this besides using Regex? Like maybe this should be tested in the JS front-end?

I just feel like this is going to be a never ending pit of issues with more complex component HTML you know?

calebporzio:
Closing due to inactivity - thanks though!

Issue #1546, opened 2020-09-11
livewire is not using Custom Validation Attributes language from validation.php


falconeri:
### Description
I have an array component ```<input type="file" name="directors[{{ $index }}][passport_image_front]"/>``` and want to change error message attribute to custom name, but livewire not using validation.php language.

### Steps to reproduce

1. i have set a rules in temporary_file_upload livewire config. 
2. If I select a file that does not pass validation, the error is displayed.
2. Because it is an array, it will shows array attribute error message ```directors.0.passport_image_front```
3. I have set a Custom Validation Attributes inside validation.php ```'directors.*.passport_image_front' => 'Company Name'```
4. Livewire not respected / not using custom validation attributes

### Screenshots
![Screenshot 1](https://user-images.githubusercontent.com/9047407/92921127-b3914780-f45d-11ea-8546-cc2f57ef2783.png)
![Screenshot 2](https://user-images.githubusercontent.com/9047407/92921136-b9872880-f45d-11ea-9a70-f7fe9b5a4980.png)

**Context**
- Livewire version: 2.1.3
- Laravel version: [7.28.1
- Browser: Chrome, Firefox

zepfietje:
Could be related to #1524.

zepfietje:
@falconeri, could you show the parts of your code related to validation?

falconeri:
> @falconeri, could you show the parts of your code related to validation?

I am using temporary rules inside config/livewire.php, so not using my own custom validation but validation rules from livewire

Here https://laravel-livewire.com/docs/2.x/file-uploads#configuration

zepfietje:
Okay, not sure if this is the exact same issue as #1524, but maybe it is related to https://github.com/livewire/livewire/issues/1524#issuecomment-691506776.

calebporzio:
This will be fixed in the next release. See comment here: https://github.com/livewire/livewire/issues/1524#issuecomment-692904182

PR #1643, opened 2020-09-18
Add validation `$attributes` property

lvxianchao:
Using Custom Validation Attributes.

See #1546.

Yep.
Nop. I found a issue maybe about this: #1546 

Nop.

Yep.

The `$attributes` property not work when my code like this:

```php
    public $form = [
        'email', 'password', 'remember', 'password_confirmation',
    ];

    protected $rules = [
        'form.email' => 'required|email|unique:users',
        'form.password' => 'required|min:6|confirmed',
        'form.password_confirmation' => 'required',
    ];

    protected $attributes = [
        'form.email' => '邮箱',
    ];
```

Sorry, my English is not so good 😢, but this PR really fixes the problem.

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -74,6 +74,14 @@ trait ValidatesInput
         return [];
     }
 
+    protected function getAttributes()
+    {
+        if (method_exists($this, 'attributes')) return $this->attributes();
+        if (property_exists($this, 'attributes')) return $this->attributes;
+
+        return [];
+    }
+
     public function rulesForModel($name)
     {
         if (empty($this->getRules())) return collect();
@@ -96,7 +104,7 @@ trait ValidatesInput
 
     public function validate($rules = null, $messages = [], $attributes = [])
     {
-        [$rules, $messages] = $this->providedOrGlobalRulesAndMessages($rules, $messages);
+        [$rules, $messages, $attributes] = $this->providedOrGlobalRulesAndMessages($rules, $messages, $attributes);
 
         $data = $this->prepareForValidation(
             $this->getDataForValidation($rules)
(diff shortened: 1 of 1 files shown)


iksaku:
Please add corresponding tests for this feature.

It looks great to me :+1:

redbastie:
You can already do this with the validator facade.

	Validator::make($this->form, $this-rules)->validate();

iksaku:
@redbastie Yeah, however, the aim of this PR is to be able to define an `$attributes` property in your Component's class scope, and not have to define it in every function that you want to perform validations.

This way, you only call `$this->validate()` with no arguments and all validation rules, messages and attributes will be automatically injected.

redbastie:
> @redbastie Yeah, however, the aim of this PR is to be able to define an `$attributes` property in your Component's class scope, and not have to define it in every function that you want to perform validations.
> 
> This way, you only call `$this->validate()` with no arguments and all validation rules, messages and attributes will be automatically injected.

Ohhhh I see, you want an easier way than having to go into the validation language file and specifying your keys/messages that way?

calebporzio:
Yeah, I could see adding this, however, a part of me thinks that `$attributes` could be misleading. People might not know it corresponds to the validation system. Like how a model has `$attributes`.

`$validationAttributes` might be better, but then you would want `$messages` to be `$validationMessages` to be symmetrical.

Though one.

Also, I haven't heard too many people ask for this, so I'm going to close for now. Re-submit if there's more demand. Thanks!

ChrisSantiago82:
I have the same problem.
Like in the old version, you couldn't bind models to the input and I think because of this there was no demand for it.
Now, you can bind models, your error messages will look like team.name or something like this. I think for a normal user who doesn't need to know anything about the database and models it just doesn't look so good.
Now I'm torn between not using model binding or those "weird" error messages ;)

Of course I understand the discussion if attributes is the correct name for this or not....

michabbb:
i would love to see `$validationAttributes` - because using `$this->validate()` is so easy, but at the same time useless if i am not able to use all laravel features 😕

iksaku:
@ChrisSantiago82 @michabbb It is already possible to customize both the Validation Messages and provide custom Attributes.

To learn more, you can checkout [Livewire's Validation Docs](https://laravel-livewire.com/docs/2.x/input-validation).

michabbb:
yeah, pardon, i have found it in the docs... will test it, thanks a lot for the notice!

michabbb:
just a notice for others: i was missing that feature because i had an older version `v2.2.9` running. 🙈 so take care, you are up 2 date 😏

PR #1644, opened 2020-09-18
Fixed component create with custom path (outside app folder and custom views).

maab16:
I'm trying to implement livewire in my module based project. My modules outside the app folder. This package forces to generate component class inside the app folder. So I needed to load the component classpath dynamically. I updated a couple of files and it working for me.

I hope it helps when someone tries to generate a component class in the custom path (Outside or inside the app folder).

Thanks

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -1,7 +1,7 @@
 <?php
 
 return [
-    
+
     /*
     |--------------------------------------------------------------------------
     | Root Path
@@ -10,8 +10,6 @@ return [
     | This value sets the root path for Livewire component classes in
     | your application. This value affects component auto-discovery and
     | any Livewire file helper commands, like `artisan make:livewire`.
-    |
-    | After changing this item, run: `php artisan livewire:discover`.
     |
     */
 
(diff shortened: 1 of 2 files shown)


calebporzio:
This would work for generating classes and views, but it would still not make auto-discovery work with the new directories right? You'd still have to register them using `Livewire::component()` right?

maab16:
Thanks for your reply. 

Yes, I used Livewire::component() to register components manually. It was one of the problems. But the main problem that I faced to generate a Component and view to the custom path. My custom path is base_path('project/modules/{module}/components').

BTW, I make a generator for myself by this pull request and I think it may help other guys who want to generate livewire components to the custom directory (inside and outside app directory).

calebporzio:
Ok thanks for clarifying.

I think having a "Root" and "class namespace" config is confusing.

It's not clear by their name what the different purposes they serve are.

I think ideally, we would just use the configured "class namespace" to handle generating livewire components.

A better solution would be something that doesn't rely on app()->namespace() to generate the target directory, but that might be difficult to impossible.

I'm going to close this for now.

We will keep an eye on this request though. Thank you for being willing to contribute!

PR #1645, opened 2020-09-18
Allow pushState on subsequent requests when only the path changes

inxilpro:
When a route is bound to a Livewire component, the expected behavior is that changing a route-bound parameter will update the path via `pushState`. This is true when both a path segment **and** a query string parameter are bound, but not when just the path segment is bound.

For example, given:

```php
// Route:
Route::get('/users/{user}/awards', UserAwardComponent::class);

// Component:
class UserAwardComponent extends Component
{
  public User $user;
  public $page = 1;

  protected $queryString = ['page'];
}
```

Going to `/users/1/awards?page=2` will set `$user` to the `User` with ID of `1` and the `$page` to `2`. And if I set `$user` to the `User` with ID of `2`, the URL will update to `/users/2/awards?page=2`.

This is exactly as is expected.

The bug is when there is no `$queryString` present:

```php
// Route:
Route::get('/users/{user}', UsersComponent::class);

// Component:
class UsersComponent extends Component
{
  public User $user;
}
```

Going to `/users/1` will properly set `$user` to the `User` with ID of `1`. But if I set `$user` to the `User` with ID of `2`, the URL will not update at all.

This is because `SupportBrowserHistory::shouldSendPath()` uses `app('router')->current()` for initial and subsequent requests, but on subsequent requests, `app('router')->current()` will return the `HttpConnectionHandler` controller. This is not correct in the context of `SupportBrowserHistory` on subsequent requests—we need to rely on `SupportBrowserHistory::getRouteFromReferer()` to determine the appropriate route.

This PR implements this change.


calebporzio:
Thanks @inxilpro, can you get the tests passing? And ideally provide a bit more info in the description pointing to the issue. Thanks!

inxilpro:
@calebporzio sorry—got caught up with other things and wasn't able to revisit this until just now. I updated the description. I'm honestly not sure why that test was failing. Seems completely separate from the changes I made, and it failed in different ways on each commit. I think it was just some Dusk wonkyness.

nuernbergerA:
> I think it was just some Dusk wonkyness.

It was and it was fixed by merging master ;)

calebporzio:
Thanks @inxilpro !

PR #1648, opened 2020-09-19
Last attempt to fix failing CI tests

nuernbergerA:
Changed the assertion to reflect what we want to actually test - that only 1 request fires with `@entangle`

also removed some `pause()` calls and wrapped assertions within `waitUsing()` to eliminate all timing issues in CI

after some test re-runs, we are still 🟢


calebporzio:
Thanks Adrian! Hopefully this does it!

PR #1655, opened 2020-09-20
ensure the element has a parent element

timrspratt:
This PR aims to solve errors such as:

```
TypeError: Cannot read property 'closest' of null
```

which occur when a nested component has a running requests (ajax) which returns an event to fire, but the parent re-renders and no longer includes that nested component.


calebporzio:
Are you able to write a test for this fix?

calebporzio:
Thank you for the bugfix, however, I will need a test added to merge this. Please add a test and re-submit. Thanks!

timrspratt:
Will do, I'm on holiday at the moment, so will sort when I return

PR #1659, opened 2020-09-21
Only return validated data

### LINKED ISSUE #1649: break codes with livewire update [2.2.5]
your new commit [2.2.5] break users codes.
for example in this code i dont want use $ruels property
and i just use this code:
```
  $data = $this->validate([
            'phone_number' => ['required', 'digits:11', 'regex:/^09([0-9]){9}$/'],
            'province' => ['required' , Rule::in($this->provinces) ],
            'marital_status' => ['required' , Rule::in(['married' , 'single']) ],
        ]);
```
i suppose to this code return just rule fields in my validation to $data but it returns all other properties in my component merge with $data !!!
it's not right !

**Context**
- Livewire version: 2.2.5
- Laravel version: 8.0

SiebeVE:
Currently all public properties are returned on validation. This is rather counterintuitive when working with Laravel's validation. https://github.com/livewire/livewire/issues/1649

Nope, one change.

Yes!

This PR add backs the default behaviour from Laravel that when you validate data only the validated data is returned. 

Thanks for maintining!

Resolves #1649

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -106,11 +106,11 @@ trait ValidatesInput
 
         $this->shortenModelAttributes($data, $rules, $validator);
 
-        $validator->validate();
+        $validatedData = $validator->validate();
 
         $this->resetErrorBag();
 
-        return $data;
+        return $validatedData;
     }
 
     public function validateOnly($field, $rules = null, $messages = [], $attributes = [])
--- a/tests/Unit/ValidationTest.php
+++ b/tests/Unit/ValidationTest.php
@@ -294,6 +294,18 @@ class ValidationTest extends TestCase
             ->call('runSameValidation')
             ->assertDontSee('The password and password confirmation must match');
     }
+
+    /** @test */
+    public function only_data_in_validation_rules_is_returned()
+    {
+        $component = new ForValidation();
+        $component->bar = 'is required';
+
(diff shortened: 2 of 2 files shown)


src/ComponentConcerns/ValidatesInput.php
```diff
@@ -106,11 +106,11 @@ public function validate($rules = null, $messages = [], $attributes = [])
 
         $this->shortenModelAttributes($data, $rules, $validator);
 
-        $validator->validate();
+        $validatedData = $validator->validate();
```
iksaku, line 109: Why not just override the `$data` variable? It is not used afterwards.

iksaku (changes requested):

src/ComponentConcerns/ValidatesInput.php
SiebeVE: I find this more clear.
It is no longer the _data_, but the _validated data_. The variable name reflects what is in it.
Off course, this can be changed to `$data`

imliam:
Is it worth considering this as a security patch? The regression has caused a handful of odd issues, where passing the result of `$this->validate()` into an unguarded model to mass-assign properties is now also assigning extra values that were not previously intended to be set.

It's not been too bad for stuff I'm working on, but I'm sure other people could potentially run into worse problems with this.

calebporzio:
Indeed this is a regression and introduces a security risk.

Thank you for such a good, concise PR with tests!

Just merged and tagged 2.2.6 with the fix.

Thanks!

PR #1664, opened 2020-09-22
Added keep-alive directive to polling on v1

EmilioBravo:
Added "keep-alive" directive to polling so it will keep polling when the browser window is in the background

Is the same as #1545 but for v1


calebporzio:
Going to leave this out of V1 actually. Thank you though!

PR #1665, opened 2020-09-23
Allow Models to cast its attributes when serialized

iksaku:
Asked about this in [Livewire's Discord](https://discordapp.com/channels/698229985755791471/698231501501890600/758124903714127904) and started a little research on the topic.

Single implementation :)

Surely it does, however, I would love to receive some feedback from them. I'm not entirely sure if there are edge-cases that should be tested against, however, feature does work as expected.

In v1 of Livewire, one could call casts for public properties directly on the component class, however, in v2 this feature [was removed in order to use better methods](https://github.com/livewire/docs/blob/master/resources/views/docs/2/upgrading.blade.php#L185-L258).

I think it is worth using Laravel's own Model casting methods, since this would help reduce component noise.

**EDIT 1**: This affects Livewire's JS-serialization process, so only browser-based actions (Like `wire:model`) will be affected by this. If you're using your properties directly in Blade (with the sintax `{{ $yourProperty }}`), you'll still be using the goold ol' Laravel way of working with your PHP object.


nuernbergerA:
Make sure to merge master again to fix failing tests 😉

iksaku:
I think I discovered some more possible use cases, and since I can't mark as draft, I'll just put it in the title for now.

I'll look into getting this PR ready tomorrow with more tests in place :)

iksaku:
Since I got a little more time and tests are not blowing up, I'll flag this as **Ready** again.

xfudox:
i have the same need as explained [here](https://stackoverflow.com/questions/64064239/laravel-livewire-bind-model-carbon-attribute), so this would be very helpful

calebporzio:
I like this as well. Would this be a breaking change though? I'm open to considering this a bug and therefore it'd be warranted, but I would want to know ALL the cases where merging this in would cause breakages for people's applications.

Thanks for this! And thanks for the good testing.

iksaku:
Oh yeah, I tried to make sure to test for some cases, but let me expand the test suite and try to break it to ensure there are no problems with most built-in casters.

iksaku:
Hmmm... Looks like browser tests are getting funny again 😅

iksaku:
Up to this point... Most casters work as one would expect, keeping PHP types for PHP execution, and JSON-like type preparation for Payload only.

Only need to finish find a test case for Object casting, and maybe create a better custom caster example... Hope to get this ready tomorrow :smile_cat:

iksaku:
Finally finished all test cases. Hope this is good enough @calebporzio.

On the _breaking change_ note, we should warn developers somewhere that using Model casting will definitely affect their component's data. Before, the model would be natively casted to JSON-type by PHP. This new feature will allow JSON-type casting via Laravel's model casters. Which will work the same way as `$model->toArray()` or `$model->toJson()`.

egknight:
Will this PR allow you to set `$dates` and `$dateFormat` on the model and those be honoured in the blade view?

https://laravel.com/docs/8.x/eloquent-mutators#date-mutators

iksaku:
@egknight This PR was entirely born with that purpose in mind.

Take a look at the provided tests to get an idea of what else is possible 😉

egknight:
Thanks for the reply, it looks like exactly what I need according to the tests, can't wait!

I did merge the changes from `src/HydrationMiddleware/HydratePublicProperties.php` to my vendor folder but couldn't get it working. I'll just wait for the release :)

iksaku:
@egknight Try pulling directly from my GitHub fork (using composer’s git imports)  and tell me if it works, if it doesn’t, send me a message to try figure out why it doesn’t work for you and be able to either make a note for the upcoming docs, or add more tests if needed.

PS: I’ve tried it with one project of mine and works great, maybe there are some changes in master that you are not accounting for?

egknight:
@iksaku still no luck. So I have either totally misunderstood this PR, or configured something wrong.

Let me show you my steps...

Install a new Laravel with Jetstream
`code laravel new casting --jet`

Update composer.json

`"livewire/livewire": "dev-model_attribute_casting as 2.0"`

Add your repo
```
"repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/iksaku/livewire"
        }
    ]
```
Update composer
```
➜  casting composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Removing livewire/livewire (v2.2.9)
  - Installing livewire/livewire (dev-model_attribute_casting 41b43c9): Loading from cache
Writing lock file
Generating optimized autoload files
```
Create SQLite
`casting touch storage/database.sqlite`
Update .env
```
DB_CONNECTION=sqlite
DB_DATABASE="/Users/mbp/code/casting/storage/database.sqlite"
```
Migrate database
`php artisan migrate`

Register a new user at http://casting.test/register

Add casting to `User` model
```
    protected $dates = [
        'created_at',
    ];
    
    protected $dateFormat = 'Y'; 
```
Create a Livewire Component
```
➜  casting php artisan make:livewire ShowUsers
 COMPONENT CREATED  🤙

CLASS: app/Http/Livewire/ShowUsers.php
VIEW:  resources/views/livewire/show-users.blade.php
```
Create the route
`Route::get('/users', \App\Http\Livewire\ShowUsers::class);`

Edit ShowUsers.php
```
<?php

namespace App\Http\Livewire;

use Livewire\Component;

{
    public function render()
    {
        	'users' => \App\Models\User::all(),
        ]);
    }
}
```

Edit show-users.blade.php
```
<div>
    <x-slot name="header">
        <h2 class="font-semibold text-xl text-gray-800 leading-tight">
            Users
        </h2>
    </x-slot>
    <div class="py-12">
        <div class="max-w-7xl mx-auto sm:px-6 lg:px-8">
            <div class="bg-white overflow-hidden shadow-xl sm:rounded-lg">
                @foreach ($users as $user)
                    {{ $user->name }} was created {{ $user->created_at }}
                    <br>
                @endforeach
            </div>
        </div>
    </div>
</div>
```
Visit http://casting.test/users

Result: John Smith was created 2020-10-06 01:45:05

I was expecting to see: John Smith was created 2020

By the way, I also tried with the following on the User model, and same result: John Smith was created 2020-10-06 01:45:05
```
protected $casts = [
        'created_at' => 'datetime:Y-m-d',
    ];
```

iksaku:
@egknight Oh, I see why it didn't work with your first approach.

Laravel's global date caster is not considered for Array/JSON serialization. You should manually add your property according to [Laravel's Model Casting property](https://laravel.com/docs/8.x/eloquent-mutators#date-casting).

This is, however, the way that you provided your second example, however, if you take a closer look to the provided tests, a PHP object is used for PHP works, and Blade rendering is still executed server side, so it'll work as Laravel intends it to work, while the casted value is used for Livewire's Request lifecycle, since those values are needed to be serialized as JS-compatible types.

Could you try appending your date property to an input via `wire:model`? Like:
```blade
<input type="date" wire:model="user.created_at" required>
```

Maybe I should add a note about it.

egknight:
@iksaku I have done some more work here..

Here is the updated component

```
<?php

namespace App\Http\Livewire;

use \App\Models\User;
use Livewire\Component;

{
	public $user;

	protected $rules = [
		'user.name' => 'required|min:6',
		'user.created_at' => 'required|date',
    ];

	public function mount() 
	{
		$this->user = \App\Models\User::find(1);
	}

    public function render()
    {
    	return view('livewire.show-users');
    }
}

```
Added the cast on the `User` model

```
protected $casts = [
        'created_at' => 'datetime:Y-m-d',
    ];
```

Then in the blade view

```
                {{ $user->name }} <br>
                {{ $user->created_at }} <br>
                <input type="text" wire:model="user.name" required> <br>
                <input type="date" wire:model="user.created_at" required> <br>
```

This outputs

John Smith
2020-10-06 01:45:05
John Smith
2020-10-06

So that's better, but forgive me, how can we use `{{ $user->created_at }}` and get 2020-10-06?

iksaku:
Since it’s raw Blade echoing PHP values, you’ll need to use Carbon to format your dates when not using `wire:model` 😉

egknight:
Got it, just like regular vanilla Laravel. I think just make it clear in the docs and all good 😀

calebporzio:
Thanks for this @iksaku, especially for such thorough testing 👌

Issue #1667, opened 2020-09-23
Can't validate values from an array attribute in Models


iksaku:
### Description
If you have a Model, which has an 'array' casted attribute, Livewire is unable to validate its nested values.

### Exact steps to reproduce
Take for example the provided Model and Component below.

You surely can change the Name of your actual Super Hero, and you should be able to bind (`wire:model`) into the values of each Super Hero contact and modify its name and rank values, however, once you start modifying contact values, validation messages will start popping from livewire, telling you:

`Cannot bind property [superHero.contacts.0.name without a validation rule present in the [$rules] array on Livewire component: [super-hero.list].`

But, we do have a validation rule set... Oh, but our validation rule uses an `*` for array, instead of a numerical index as the error output said.

After doing some research, it looks like a [regression was implemented](https://github.com/livewire/livewire/commit/0313887e8c355bac985a87a7580b0bee5533edc5#diff-13d0c883643367d8306b4ff4bdbe48f3L81-R89).

### Stripped-down, copy-pastable code snippets
```php
// Model
class SuperHero extends Model {
    protected $casts = [
        'contacts' => 'array'
    ];
}
```

```php
// Component
class SuperHeroList extends Livewire\Component {
    public SuperHero $superHero;

    protected $rules = [
        'superHero.name' => ['required'],
        'superHero.contacts.*.name' => ['required', 'distinct'],
        'superHero.contact.*.rank' => ['required', 'integer']
    ];

    public function save() {
        $this->validate();
        $this->superHero->save();
    }

    public function render() {
        return view('super-hero.list');
    }
}
```

**Context**
- Livewire version: 2.2.6
- Laravel version: 8.6.0
- Browser: Firefox

joowdx:
Hi. I just wanna ask if this fix was implemented along with the most recent version of livewire coz I'm still facing the same issue😫

```
composer show -i | awk '/laravel\/framework|livewire\/livewire/'
laravel/framework                  v8.7.1    The Laravel Framework.
livewire/livewire                  v2.2.9    A front-end framework for Laravel.
```

iksaku:
It has been fixed on master, no release yet for it. Maybe Caleb is waiting for another feature implementation to tag a 2.3 minor release.

joowdx:
@iksaku so far are you having any problems regarding the validation rules set? It doesn't seem to respect what's set on the rules like date, date_format, numeric, etc.. The validation just passes despite the rules. Or is it just me?

[...]

PR #1668, opened 2020-09-23
Fix not being able to set values on specific numerical indexes of array attributes of models

iksaku:
Previously created an issue: #1667 

Straight to the neck... Ermm... Implementation...

Yup, tests where added to ensure it is not broken in the future.

Mostly, what #1667 says.

If your model contains an array of arrays on which you would like to `wire:model` a specific index, validation would fail, even if your rules are properly set as arrayable (`model.attribute.*.nested`).

This is an attempt to fix that.

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -86,12 +86,22 @@ trait ValidatesInput
 
     public function missingRuleFor($dotNotatedProperty)
     {
-        return ! collect($this->getRules())
+        // Prepare numerical-indexed properties for array attributes
+        $preparedProperty = (string) Str::of($dotNotatedProperty)
+            ->replaceMatches('/(?<=(\.))\d+\./', '.*.')
+            ->replace('..', '.');
+
+        $isANumericallyIndexedProperty = $dotNotatedProperty !== $preparedProperty;
+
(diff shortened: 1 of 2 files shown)


erikgaal:
I think we need another test for the case where the key of property that's going to be set is _ending_ with an index, e.g. `foo.zap.0`. This would be so when the items of the array aren't `objects`, but `strings` or other primitives.

```php
Livewire::test(ComponentForEloquentModelHydrationMiddleware::class, [
    'foo' => $foo = Foo::first(),
])  ->set('foo.zap.0', 'glob')
    ->call('save')
    ->assertHasErrors('foo.zap.*')
    ->set('foo.zap.0', 'globbbbb')
    ->call('save')
    ->assertHasNoErrors();

    $this->assertEquals(['globbbbb'], $foo->fresh()->zap);
}

/* ... */

protected $rules = [
    ...,
    'foo.zap.*' => 'required|min:5',
]
```

iksaku:
@erikgaal thanks for the heads up! Let me push another commit for this 😄

calebporzio:
Thanks, I'll wait for that next commit before merging.

iksaku:
All ready to go @calebporzio :wink:

calebporzio:
Thanks @iksaku & @erikgaal !

PR #1681, opened 2020-09-23
Extract Dusk macros

imliam:
**This pull request just pulls the macros for the `Laravel\Dusk\Browser` class out into its own mixin class.**

The reason for this is that these macro functions would be extremely handy to use in our own applications that use Livewire and need Dusk testing. With this change, we could easily register the mixins (or individual macros from it) into our own tests as needed.

The `Livewire::test()` utility is nice for testing backend logic, but when we have increasingly complex frontends with a mix of Alpine and other stuff happening on the frontend, Dusk still serves a good purpose in testing our own apps - and having access to some of these more complex test methods like `waitForLivewire()` would help make testing using Dusk a lot easier.

### DIFF
--- a/src/Macros/DuskBrowserMacros.php
+++ b/src/Macros/DuskBrowserMacros.php
@@ -1,0 +1,156 @@
+<?php
+
+namespace Livewire\Macros;
+
+use PHPUnit\Framework\Assert as PHPUnit;
+use Illuminate\Support\Str;
+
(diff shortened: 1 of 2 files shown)


nuernbergerA (approved):
That's a good idea! We would have to make a own section in the docs for it.

calebporzio:
I dig. Good call Liam!

Issue #1678, opened 2020-09-23
Undefined index: value


giorgiopogliani:
### Description
This is the code where the error comes from. Probably, I am missing something but with an if containing `array_key_exists` it works as it should. Maybe there is some double call? I have no idea.

```
...
$data = $update['payload'];
$unHydratedInstance->syncInput($data['name'], $data['value'])
...
```

### Exact steps to reproduce

The following blade templates plus a normal livewire component with a `$foo` property

```
@push('styles')
    <link href="https://cdn.quilljs.com/1.3.6/quill.snow.css" rel="stylesheet">
@endpush

@push('scripts')
    <script src="https://cdn.quilljs.com/1.3.6/quill.js"></script>
@endpush

<div class="mt-2 bg-white" wire:ignore>
    <div
        x-data="{}"
        x-init="
            quill = new Quill($el, {theme: 'snow'})
            quill.on('text-change', function () {
                $dispatch('input', quill.root.innerHTML);
            });
        "
        wire:model="foo"
    >
        {!! $foo !!}
    </div>
</div>
```

**Context**
- Livewire version: 2.2.6
- Laravel version: 8.6.0
- Browser: Chrome, Safari

dennisvandalen:
I'm running into the same thing.

```json
   "updates":[
      {
         "type":"syncInput",
         "payload":{
            "name":"short_description",
            "value":"<p>D</p>"
         }
      },
      {
         "type":"syncInput",
         "payload":{
            "name":"short_description"
         }
      }
   ]
```

giorgiopogliani:
It looks like the `keydown` event (I think) setuped by quilljs triggers the wire:model of livewire.   

As workaround I removed the wire:model and I'm setting the property using `@this.set('foo', text)` and it's working for now.

It would be cool to stop that empty update though.

glaesser:
v2.2.8 now throwing this error constantly when dynamically showing/hiding (updating) page elements. Worked perfectly on v2.2.7 and below for me.

calebporzio:
@glaesser  - thank you for catching that regression. I just fixed it and tagged it in v2.2.9

@giorgiopogliani - this issue is caused because the package you are using is firing an input event on the div with the value in the .data property. That is specific to that package. Even if Livewire didn't through the undefined index error you would still hit bad behavior.

You need to prevent that event from hitting wire:model by either listening for something else, or preventing default on it before it reaches wire:model. THEN, use your `$dispatch('input')` like you are.

[...]

PR #1685, opened 2020-09-24
fix(using property.match when property is not always string)

ryangjchandler:

This is actually a bug I encountered whilst trying to loop over `window.Livewire.all()`.

No.

No.

The `$wire` Proxy assumes that `property` is a string and will try to call `property.match`. In some cases though, `property` might not be a string and could actually be a `Symbol` or similar.

I experience this in Firefox whilst trying to `console.log(window.Livewire.all())`, but have also reproduced in Chrome.

### DIFF
--- a/js/component/index.js
+++ b/js/component/index.js
@@ -645,7 +645,7 @@ export default class Component {
                 if (property === '__instance') return component
 
                 // Forward "emits" to base Livewire object.
-                if (property.match(/^emit.*/)) return function (...args) {
+                if (typeof property === 'string' && property.match(/^emit.*/)) return function (...args) {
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks @ryangjchandler !

glaesser:
@calebporzio While this latest release (v2.2.8( has indeed fixed safari (including safari mobile) autofill (thank you!), it has also broken basic functionality. Now receiving error:

`Undefined index: value `
` $unHydratedInstance->syncInput($data['name'], $data['value']);`

when toggling showing/hiding components (e.g. a button that hides/shows text). This works perfectly in 2.2.7 and below, and affects all browsers (safari, chrome etc.).

May be related to https://github.com/livewire/livewire/issues/1678?

PR #1687, opened 2020-09-24
Added flags for type checking set/not-set on TestableLivewire

iksaku:
No previous discussion nor issue, just something I've been looking for for quite some time

Single feature

I don't think there are tests for the "test" definitions, but correct me if I'm wrong.

The `TestableLivewire` is the founding block for component testing, and it has two basic assertions: `assertSet()` and `assertNotSet`.

These two assertions use [PHPUnit's `assertEquals()`](https://phpunit.readthedocs.io/en/9.3/assertions.html#assertequals), which helps to determine if a certain value is properly assigned to a given property. However, `assertEquals()` doesn't test for type matching, it just test for value equality.

[As mentioned in PHP's Documentation](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison), some fun things may happen if you only rely on value equality, when you should be looking for both, value and type equality. This is however not always the case, and some developers may not always need type-checking.

Instead creating new `assertSameType()` and `assertNotSameType()` functions, we could just introduce an optional flag in the existing functions to switch from `assertEquals()` to [`assertSame()`](https://phpunit.readthedocs.io/en/9.3/assertions.html#assertsame).

This would be a backwards-compatible change, since it'll default to use `assertEquals()`, and still allow type testing when needed.

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -11,12 +11,16 @@ use PHPUnit\Framework\Assert as PHPUnit;
 
 trait MakesAssertions
 {
-    public function assertSet($name, $value)
+    public function assertSet($name, $value, $same = false)
     {
         if (is_callable($value)) {
             PHPUnit::assertTrue($value($this->get($name)));
         } else {
-            PHPUnit::assertEquals($value, $this->get($name));
+            if ($same) {
+                PHPUnit::assertSame($value, $this->get($name));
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the PR.

Closing as I think we will need a better real-life example of where this is necessary. Also I want to know others would need this.

Also also, there ARE tests for TestableLivewire assertions.

Thanks for being willing to contribute! It means the world.

PR #1688, opened 2020-09-25
Fix autofill bug for Safari

redbastie:
yes

no

no

this fixes broken autofill in Safari which was caused by the way the compiler handled the undefined comparison and converted it into code that Safari doesn't understand

### DIFF
--- a/js/node_initializer.js
+++ b/js/node_initializer.js
@@ -78,8 +78,8 @@ export default {
                 // We have to check for typeof e.detail here for IE 11.
                 let value =
                     e instanceof CustomEvent &&
-                        typeof e.detail != 'undefined' &&
-                        typeof window.document.documentMode == 'undefined'
+                        typeof e.detail !== undefined &&
+                        typeof window.document.documentMode === undefined
                         ? e.detail
                         : DOM.valueFromInput(el, component)
 


amcsi:
What was wrong before with Safari out of curiosity by the way?

Also, are you sure this is going to work? This seems like it goes against my knowledge of JS.

```js
typeof e.detail !== undefined
```

`typeof` should always return a string AFAIK, so that condition will always pass.
Same with `typeof window.document.documentMode === undefined` which will never pass.

So I'm worried that this commit would break IE11.

redbastie:
It gets compiled to the code mentioned in my comment:

```
void 0 !== _typeof(e.detail) &&
void 0 === _typeof(window.document.documentMode)
```

The problem with Safari is that currently Livewire doesn't pick up autofilled values.

glaesser:
@calebporzio please include this in the next release - autofill in safari (especially safari iOS) has been broken for months, and prevents login forms from being used effectively (i.e. with password managers or keychain).

amcsi:
@redbastie and can you tell me what `_typeof()` does?

redbastie:
> @redbastie `_typeof()` is the same as the `typeof` operator, except it supports babel transpiled Symbols.
> 
> Could the problem with Safari rather maybe be that the operator presedence is simply wrong, and that's what was causing the problem?
> 
> If you rush this PR, you're definitely breaking IE11.

The current `livewire.js` build already uses `_typeof` in a few different places:

https://github.com/livewire/livewire/blob/master/dist/livewire.js

Your concerns about IE 11 having an issue with this function are irrelevant.

amcsi:
> 
> The current `livewire.js` build already uses `_typeof` in a few different places:
> 
> https://github.com/livewire/livewire/blob/master/dist/livewire.js

Yes, if you search for the usages for `_typeof()` in what you linked, you'll see that they compare the result with a string: `value&&"object"===_typeof(value)&&value.livewireEntangle`. In this case, with "object".

> Your concerns about IE 11 having an issue with this function are irrelevant.

I didn't say that IE11 had a concern with that function. What I said was that the this entire expression will always resolve to `false`, because comparing the `undefined` type to the result of `typeof` (or even `_typeof()` will always resolve to false, because `undefined !== "undefined"`.

So if you look at the entire ternary...

```js
                // We have to check for typeof e.detail here for IE 11.
                let value =
                    e instanceof CustomEvent &&
                        typeof e.detail !== undefined &&
                        typeof window.document.documentMode === undefined
                        ? e.detail
                        : DOM.valueFromInput(el, component)
```

...since the condition expression will always resolve to `false`, the ternary condition will never be able to return `e.detail` for IE11.

_That's_ why this breaks IE11.

redbastie:
The browser doesn't actually pass `undefined` as a string, it is a primitive type as I told you in the github issue.

redbastie:
Heres some proof for you: https://jsfiddle.net/uhcdvyka/

amcsi:
> The browser doesn't actually pass undefined as a string, it is a primitive type as I told you in the github issue.

Exactly, that's what I'm saying too. And if you agree with me, then why are you comparing the `undefined` _type_ with a string?

```js
typeof e.detail !== undefined // this will always be FALSE
void 0 !== _typeof(e.detail) // this will also always be FALSE, because _typeof() also returns a string
```

redbastie:
You're out of your element.

amcsi:
Fine, attack me and don't listen to my arguments. Merge this in and screw this up for IE11. Suites this really buggy software well anyway.

redbastie:
> Fine, attack me and don't listen to my arguments. Merge this in and screw this up for IE11. Suites this really buggy software well anyway.

Did you not bother checking the jsfiddle link I posted proving you wrong? Why aren't those 3 lines of code all logging as false?

🤔

amcsi:
@redbastie that's not what i got. Did you check in chrome?
The first two as true and the third was false.

redbastie:
You mean to tell me they aren't *always returning false*?

amcsi:
Ah crap, i was mixed up in boolean logic. What I meant to say there was that it would always evaluate to TRUE (not false).
And that's the case with the first two logs of yours.
In your third log you correctly compare the _string_ (not type) "undefined" with the `typeof` result, and that is how you're able to get FALSE.

At least according to Chrome 🤔

So I have a feeling the source of the Safari issue will have something to do with the different console results we are getting between the browsers.

redbastie:
Post your solution if mine isn't good enough then.

redbastie:
Internet explorer: still impeding progress in 2020.

redbastie:
Also IE 11 < 1% market share

Safari = 17% market share

mblarsen:
> Heres some proof for you: https://jsfiddle.net/uhcdvyka/

As @amcsi is pointing out, I'm not sure what you are trying to prove here. 

```js
let e = {};

console.log(void 0 !== typeof e.detail); // can be replaced with console.log(true) as it doesn't matter what value detail has or if it is defined or not.
console.log(undefined !== typeof e.detail); // same for this one.
console.log("undefined" !== typeof e.detail); // this works as expected.
```
As you say yourself you can exchange `_typeof()` with `typeof`.

@redbastie perhaps as the PR author you can do the service and test in IE11 or at least encourage the test instead of rudely disregarding @amcsi 's concern. Not really helpful. You may be right, but raised concerns do seem valid.

Your argument about market share is completely irrelevant.

redbastie:
enjoy fucking around with IE11 and other garbage browsers that have and will continue to fuck modern web development for the forseeable future despite being less than 1% market share.

i see people arguing with me about this bullshit for hours instead of offering a solution. if you're such a stone genius then suggest a change for these huge 2 lines of code that babel will compile to js that doesn't fuck up browsers that people actually use in 2020.

redbastie:
@amcsi you were right and i apologize.

amcsi:
@redbastie no worries

Issue #1662, opened 2020-09-22
Better Serverless support (LivewireManager::javaScriptAssets() should be more configurable, care taken for component auto-discovery)


ralphschindler:
### Description

In attempting to make Livewire work in a Serverless / AWS Lambda environment (specifically Bref, which I assume is similar in scope to Vapor), it is hard to achieve a deployment where assets are located in S3 and the application is served from Lambda.

The problem is that LivewireManager::javaScriptAssets() is trying to compute and detect too many values. Typically I would want assets delivered from S3 which is generally configurable (typically at the env `ASSET_URL` / config value `app.asset_url` and the app's base URL for routes is at the configured location (typically at env `APP_URL` or config value `app.url`).

Currently, `javaScriptAssets()` is trying to deduce the *`appUrl`* from the *assets url*:

https://github.com/livewire/livewire/blob/master/src/LivewireManager.php#L168

This value is later used on the rendered javascript:

https://github.com/livewire/livewire/blob/master/src/LivewireManager.php#L209

Generally, people will actually [publish their js vendor file, so then the Vapor detection](https://github.com/livewire/livewire/blob/master/src/LivewireManager.php#L179-L184) would be expected to run to change some of these values.

I've made this work by hacking `isOnVapor()`, and then configuring `livewire.asset_url` to use the `APP_URL` value (in `config/livewire.php`: `'asset_url'  => env('APP_URL', null),`):

```php
    public function isOnVapor()
    {
        return ($_ENV['SERVER_SOFTWARE'] ?? null) === 'vapor'
            || ($_ENV['SERVER_SOFTWARE'] ?? null) === 'bref';
    }
```

But this then further causes issue b/c in a serverless environment you need to be able to write a file to `/tmp` at a deep location, so this is also required in bootstrap.php (to facilities component auto-discovery caching):

```php
if (isset($_ENV['SERVER_SOFTWARE'])
    && $_ENV['SERVER_SOFTWARE'] === 'bref'
    && !file_exists('/tmp/storage/bootstrap/cache'))
{
    mkdir('/tmp/storage/bootstrap/cache', 0777, true); // recursive create directories
}
```

#### Proposed Solutions?

I hesitate to make a PR because there is a range of solutions, but I don't mind taking a stab at it if you had a preferred direction.

My best advice for a PR would be more specific (optional) config keys that when set will also disable some of the code-paths that do some of the more auto-magic discovery (eg: `isOnVapor()`).

### Exact steps to reproduce

[...]

PR #1693, opened 2020-09-25
Fix app/asset URL usage

ragulka:

Yes, see https://github.com/livewire/livewire/issues/1662. Note that https://github.com/livewire/livewire/discussions/1674 would also benefit from this, although this PR does not directly related to the proposal there.

No.

It does :)

Please see https://github.com/livewire/livewire/issues/1662. The way Livewire currently uses the `asset_url` config value breaks Livewire under certain circumstances (see issue for more details). Given that `asset_url` is _not_ used for serving (static, published) assets at all, but rather for determining the path to LW routes on the frontend, the key naming is also misleading.

Note that this PR only handles the asset/app URL handling reported in #1662, it does not handle tmp files/bootstrapping mentioned in the original issue.

Consider this PR a proposed fix - I'm happy to change the implementation, so long it addresses the underlying issue.

### Changes
* Rename `livewire.asset_url` to `livewire.app_url`. 
* Always use `app.asset_url` when serving published assets, regardless if on Vapor or not
* Add tests to ensure the above scenarios work

### Other considerations

**`app.url` vs `livewire.app_url`**

I initially wanted to do away with the Livewire-specific app URL configuration value entirely, because we already have `config('app.url')` (set by `APP_URL` ENV variable) available. However:
* if the APP_URL is not correctly set, it might break existing installations (although [you should always set the APP_URL](https://quickadminpanel.com/blog/why-its-important-to-change-app_url-in-laravel-env-file/))
* even if `APP_URL` is correctly set, it may not work in a multi-tenant app with different subdomains, where the `APP_URL` points to the main domain, but livewire is only available on a specific subdomain

Then again, Laravel expects the `APP_URL` to be correctly set when you're using CLI, emails, etc anyway, so I'd be happy to remove the LW config value and simply use `APP_URL`.

**Deprecating `asset_url`**

This PR might be a breaking change for some sites (depending on their setup), so we might want to soft-deprecate the `asset_url` option instead of outright removing/renaming it and issue console warnings if it's still used?

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -165,7 +165,8 @@ HTML;
     {
         $jsonEncodedOptions = $options ? json_encode($options) : '';
 
-        $appUrl = config('livewire.asset_url', rtrim($options['asset_url'] ?? '', '/'));
+        $appUrl = rtrim($options['app_url'] ?? config('livewire.app_url'), '/');
+        $assetUrl = rtrim(config('app.asset_url') ?? '', '/');
 
         $csrf = csrf_token();
 
@@ -181,7 +182,7 @@ HTML;
             $publishedManifest = json_decode(file_get_contents(public_path('vendor/livewire/manifest.json')), true);
             $versionedFileName = $publishedManifest['/livewire.js'];
 
-            $fullAssetPath = ($this->isOnVapor() ? config('app.asset_url') : $appUrl).'/vendor/livewire'.$versionedFileName;
+            $fullAssetPath = $assetUrl.'/vendor/livewire'.$versionedFileName;
 
             if ($manifest !== $publishedManifest) {
                 $assetWarning = <<<'HTML'
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -31,18 +31,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Livewire Assets URL
+    | Livewire App URL
     |--------------------------------------------------------------------------
     |
-    | This value sets the path to Livewire JavaScript assets, for cases where
-    | your app's domain root is not the correct path. By default, Livewire
-    | will load its JavaScript assets from the app's "relative root".
+    | This value sets the path to your Laravel app, for cases where
+    | your app's domain root is not the correct path. It's used by Livewire's
+    | JavaScript portion when communicating with the server. By default, Livewire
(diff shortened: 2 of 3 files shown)


src/LivewireManager.php
```diff
@@ -165,7 +165,8 @@ protected function javaScriptAssets($options)
     {
         $jsonEncodedOptions = $options ? json_encode($options) : '';
 
-        $appUrl = config('livewire.asset_url', rtrim($options['asset_url'] ?? '', '/'));
+        $appUrl = rtrim($options['app_url'] ?? config('livewire.app_url'), '/');
```
ralphschindler, line 168: On it's own, that didn't work for me, I changed that line to this:

```php
        $appUrl = $options['app_url'] ?? config('livewire.app_url');
        $appUrl = rtrim($appUrl ?? config('app.url', '/'), '/');
```

ragulka: @ryangjchandler could you also set `'app_url' => config('app.url')` in livewire config?

ralphschindler: That may work(?) but I don't believe that is a good practice to suggest.

If it does work, it only works because config files *happen* to be processed alphabetically. Also, I don't know how well that works when config caching is taking into consideration.

I feel like sensible fallbacks are the path forward for a good out of the box experience... or let the default value be `env('LIVEWIRE_APP_URL', null)`

ralphschindler: *sidenote: if you google `Laravel config within config` you get a general sense of why that's not a generally acceptable practice, even though it is not explicitly ever discussed.

ragulka: I see - yeah, relying on load order is probably not a good idea. In the end I'd like to hear @calebporzio's thoughts on the PR before moving forward.

calebporzio:
Good PR. However, this would be a breaking change, AND I have other reasons that I outlined in a comment at the end of this issue: #1662 

Going to close this for now.

I've added a link to this in the README roadmap so we can re-open for the next major version when we're ready.

Thanks!

erensanlier (commented):
Totally solves the problem! Thanks a lot

onderic:
Its my first time working with livewire ,am getting this error when i upload my project to cpanel in a subdomain "UnexpectedValueException
Invalid route action: [App\Http\livewire\HomeComponent]." , but the same project is perfectly working in local host ,how can i solve it?

PR #1694, opened 2020-09-25
Allow customizing livewire route prefix

ragulka:

Yes. See https://github.com/livewire/livewire/discussions/1674

No

It sure does

This makes possible to identify a tenant on livewire routes in a multi-tenant app using path identification. Consider `example.com/{tenant}/livewire/message` - if the middleware group includes tenant identification, there's no need to identify the tenant in each individual livewire component. The app would need to provide a default value for the `{tenant}` param, but that's expected in such a scenario anyway.

It also makes it possible to simply adjust the livewire route namespace/prefix. Ie use `foo/message` instead of `livewire/message`. Given that this prefix is somewhat internal, I'd be happy remove this possibility and simply allow adding an additional prefix to `/livewire`, so you'd have `/foo/livewire/message`

Note: if https://github.com/livewire/livewire/pull/1693gets merged, this PR might need to be refactored a bit.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -31,6 +31,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Livewire Route Prefix
+    |--------------------------------------------------------------------------
+    |
+    | This value sets the prefix for Livewire routes, for cases where
+    | you need to customize where livewire mounts it's routes. By default, Livewire
+    | will prefix its routes with "/livewire".
+    |
+    | Examples: "/livewire", "/{tenant}/livewire".
+    |
+    */
+
+    'route_prefix'  => '/livewire',
+
+    /*
+    |--------------------------------------------------------------------------
     | Livewire Assets URL
     |--------------------------------------------------------------------------
     |
--- a/js/connection/index.js
+++ b/js/connection/index.js
@@ -22,7 +22,7 @@ export default class Connection {
 
         // Forward the query string for the ajax requests.
         fetch(
-            `${window.livewire_app_url}/livewire/message/${payload.fingerprint.name}`,
+            `${window.livewire_app_url}${window.livewire_prefix}/message/${payload.fingerprint.name}`,
             {
                 method: 'POST',
                 body: JSON.stringify(payload),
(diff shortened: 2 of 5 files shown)


calebporzio:
Thanks for this contribution, however, I think I'd prefer to go with a customization that can be managed in user-land.

I like your recommendation about being able to attach custom headers to Livewire requests.

I'm going to close this, but feel free to PR something that adds the ability to add headers to Livewire requests.

Thanks!

ragulka:
> I think I'd prefer to go with a customization that can be managed in user-land

@calebporzio can you be more specific? In which way is the proposed custom prefix not managed in user-land if it takes a configuration value?

With regards to custom headers, how do you imagine these could be provided?

ragulka:
@calebporzio here's my initial, possibly naive take for custom headers:

```js
getHeaders() {
    return Object.assign(this.customHeaders(), {
        'Content-Type': 'application/json',
        Accept: 'text/html, application/xhtml+xml',
        'X-CSRF-TOKEN': getCsrfToken(),
        'X-Socket-ID': this.getSocketId(),
        'X-Livewire': true,

        // We'll set this explicitly to mitigate potential interference from ad-blockers/etc.
        Referer: window.location.href,
    })
}

customHeaders() {
    return window.livewire_custom_headers ?? {}
}
```

Users could simply set any custom headers on `window.livewire_custom_headers`. This would work for use-cases like in my original PR description, but would not work if the headers would need to be customized per-component. 

To allow per-component headers, maybe headers could be defined on the Livewire PHP component:

```php
public function customHeaders() : array
{
  return ['X-Custom' => $this->id];
}
```

These would be rendered in the `wire:initial-data` tag of the component.

Then the `Message` class in JS would have a `headers()` method that could be used like `payload()` in `/connection/index.js`.

yet another approach would be to introduce an event here, before the message is sent: https://github.com/livewire/livewire/blob/master/js/component/index.js#L212

Thoughts, preferences?

kevyworks:
I am also having issues somehow on this on NGINX using [Nginxconfig.io's](https://nginxconfig.io) configurations, without CORS.. Causing `fetch` preflight due to chromes self signed certificates or maybe something is wrong with my certificate but or configurations. 

But id like to have the ability to change this as-well for customizability and perhaps added custom security which can be manage through NGINX, lastly also for tenant or modular approaches.

~ Best Regards.

PR #1696, opened 2020-09-25
Fix autofill in non-IE11 browsers (Safari, etc.)

redbastie:
yes

no

no

fixes autofill for non-IE11 browsers which are currently broken (Safari, etc.)


iksaku (approved):
As far as I can see, it doesn't interfere with any other browsers. Will properly scope as an IE11-only check :older_man:

calebporzio:
This is not a valid fix for a number of reasons. Thanks so much for tackling this though!

This problem should be fixed in v2.2.8

PR #1704, opened 2020-09-26
Allow passing a message to the assertions

jimmypuckett:

I hope so, as I know that we want it.  I could not find any discussion on it.

No

The existing test covers the use-case & makes sure that variables are passed as expected.

When chaining together multiple assertions, it is nice to set a message, so this PR allows passing messages to the tests.


calebporzio:
Thanks for the contribution.

However, this would stray from the API of the Laravel's testing methods.

Not that that's a real problem, I would just want to hear more people request this feature before introducing it.

Thank you, but closing for now.

jimmypuckett:
ok, thanks.

PR #1708, opened 2020-09-27
Allow DI for computed properties

morrislaptop:

Wanted. No issue / discussion first. 

No

Yes

This change will allow dependencies to be injected into computed property methods. This is useful if the computed property requires another class to return a result. A common example would be a Repository class..

```php
public function getAccountsProperty(AccountsRepo $repo) {
    return $repo->all();
}
```


calebporzio:
This is perfect, thanks!

PR #1709, opened 2020-09-27
Add ability to setup default Component property values with the use of a "created()" method

iksaku:
There haven't been any discussions for this type of feature, but there has been some cases in which I stumbled upon needing this feature.

Single feature implementation

Yup, all :green_circle: as it should be :) Also, added related tests on the `LifecycleHookTest` suite.

So, the reason for this PR is to introduce the ability to have a `created` method for Livewire\Component clases, which allows developers to configure their components every time they're called, independently from the context (First render or Property Updating).

This feature is based on [Vuejs' `created` lifecycle hook](https://vuejs.org/v2/api/#created).

In order to assign default property values in a component, a developer may need to define values "inline" with the property definition, or use the `mount` method to do so. But this is not a bulletproof way to do so, since you can't call functions nor create classes "inline" with property definitions, and `mount` will only be called the first time the component is rendered, not after subsequent requests.

So, with this in mind, the `created` method would come in handy. It'll be called every time a component is instantiated, without needing to override the `__construct()` method, and will help to not have to override the `reset()` method either whenever a component is able to be reset. Also, developers will be able to assign class-based values to their components, and would not need to have to do so twice: once in a `mount()` method and once in the `reset()` action. It'll all be down to a single `created()` method.

I'll try to illustrate in code a simple case, which may not be as error-prone in PHP 7.3, but starting with PHP 7.4, things start to break immediately.

Consider the following code:
```php
class MyComponent extends Livewire\Component
{
    public User $user;

    public function mount()
    {
        if (!isset($this->user)) {
            $this->user = User::make();
        }
    }

    public function resetComponent()
    {
        $this->reset();
    }
}
```
Our default class for `$user` is an empty model, and we can call for a _component reset_, so we would get a new instance of our `$user` class. Under the hood, this will try to access "$freshInstance->user" from Livewire's default `reset()` function, which is the most logical way and it is _OK_ in **PHP 7.3**, but it'll be assigned to null instead of our intended default class, so this is not _good_ at all. However, in **PHP 7.4**, this will directly result in a Fatal error, because we're trying to access an "uninitialized" typed property, which is even worse.

So, a way to mitigate this issue is to either duplicate the `mount()` method into the `resetComponent()` method, or reference one to the other:
```php
class MyComponent extends Livewire\Component
{
    public User $user;

    public function mount()
    {
        $this->resetComponent();
    }

    public function resetComponent()
    {
        if (!isset($this->user)) {
            $this->user = User::make();
        }
    }
}
```
This would work "great", however, there's still noise in the component, since we need our `mount()` function to call the `resetComponent()` every time, and that type of spaghetti-boilerplate may not be needed at all.

We can solve this specific case using a `created()` method, which will keep our code simple to the eye, and not having to reference functions one to another, since it'll not be needed:

```php
class MyComponent extends Livewire\Component
{
    public User $user;

    public function created()
    {
        if (!isset($this->user)) {
            $this->user = User::make();
        }
    }

    public function resetComponent()
    {
        $this->reset();
    }
}
```
Now, we're back to our first implementation, but using `create()` instead of `mount()`. Now, when calling `resetComponent`, in both **PHP 7.3** and in **PHP 7.4** the new component will have a fresh User model in the respective property, keeping out the `null` value in 7.3, and the _Uninitialized_ error in 7.4.

I hope the intended use case is clear enough. If not, I'll be happy to keep thinking of useful use cases :smile:.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -37,8 +37,8 @@ abstract class Component
         $this->ensureIdPropertyIsntOverridden();
 
         if (method_exists($this, 'created')) {
-            // At this point, the component's traits have been initialized,
-            // but neither Request Data nor Property Bindings have been executed.
+            // At this point, the component has been created,
+            // but neither Request Data, Property Bindings, nor Traits have been done.
             // This is intended for assigning default values to mission-critical properties,
             // or class-based properties, like Collections or Models, since in PHP you can't
             // create classes nor call functions inline when defining class properties)
--- a/src/Component.php
+++ b/src/Component.php
@@ -36,6 +36,15 @@ abstract class Component
 
         $this->ensureIdPropertyIsntOverridden();
 
+        if (method_exists($this, 'created')) {
+            // At this point, the component's traits have been initialized,
+            // but neither Request Data nor Property Bindings have been executed.
+            // This is intended for assigning default values to mission-critical properties,
+            // or class-based properties, like Collections or Models, since in PHP you can't
+            // create classes nor call functions inline when defining class properties)
+            $this->created();
+        }
+
         $this->initializeTraits();
     }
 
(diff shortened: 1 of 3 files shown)


calebporzio:
1. Thanks for submitting this. Unfortunately, I want more people to express the need for something like this before merging it. A new lifecycle hook introduces more overhead.

2. I don't think I would use `created`, as it doesn't map perfectly to Vue's created hook and may cause confusion.

3.  The "hydrate" lifecycle hook may be helpful here

Closing for now. Thanks!

Issue #1585, opened 2020-09-14
[v2.2.2] value not shown using wire:model with eloquent collection


masterix21:
![example](https://user-images.githubusercontent.com/6555012/93079397-59c29480-f68c-11ea-8eb7-a5512361cc6a.gif)

### Description

*Preamble:* the bug isn't present using the version 2.2.1

*Component:*
```php
class Setup extends Component
{
    public Company $company;
    public Collection $sections;

    protected $listeners = [
        'refresh-setup' => 'refresh',
    ];

    public function mount()
    {
        $this->refresh();
    }

    public function refresh()
    {
        $this->sections = CompanyAccountingDaySection::query()
            ->whereCompanyId($this->company->id)
            ->orderBy('name')
            ->get();
    }
}
```

```blade
@foreach ($sections as $key => $section)
    <input type="text" wire:key="cads-id-{{ $section->id }}" wire:model.lazy="sections.{{$key}}.name" value="{{$section->name}}" />
@endforeach
```

###  Differences with the previous 2.x versions
The previous version only has the bug of `value="{{$section->name}}"` attribute: without, doesn't show the initial data.

### Steps to reproduce

**Context**
- Livewire version: 2.2.2
- Laravel version: 8.1.0
- Browser: Chrome

calebporzio:
You need to have a `$rules` property to specify keys in the nested collection that should be made available to the front-end.

If THAT doesn't work, please re-submit an issue for that bug.

Thanks!

masterix21:
@calebporzio, I have tried already to use `$rules = ['sections.*.name' => 'required']`, but doesn't work.

austenc:
This should be fixed with `2.3.0`

masterix21:
Yes, seems to be fixed.

Thanks

calebporzio:
Wooot!

PR #1710, opened 2020-09-28
Fix dehydrate using Collections

masterix21:
Issue related: #1585 

From Livewire 2.2.2 there is a problem to dehydrate Collection. For example, if you follow the code like so:
```php
namespace App\Http\Livewire\Users\Index;

use App\Models\User;
use Livewire\Component;

class Index extends Component
{
    public Collection $users;

    public function mount()
    {
        $this->users = User::orderBy('name')->get();
    }
}
```

```blade
@foreach ($users as $key => $user)
    <input type="text" 
           wire:key="users-{{ $user->id }}"
           wire:model.lazy="users.{{$key}}.name" />
@endforeach
```

Using the code above, you won't see the users' names.

The PR fixes the problem.

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -106,9 +106,12 @@ class HydratePublicProperties implements HydrationMiddleware
 
         // Deserialize the models into the "meta" bag.
         data_set($response, 'memo.dataMeta.models.'.$property, $serializedModel);
-
         $modelData = [];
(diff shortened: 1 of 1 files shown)


calebporzio:
This is actually intended behavior.

Livewire shouldn't expose model data to the frontend that hasn't been added to the `$rules` array.

Ideally, you would have validation `$rules` property to specify nested values that are able to be wire:model bound.

masterix21:
@calebporzio, `$rules` aren't enough with Eloquent collections: doesn't work without the changes of this PR.

PR #1721, opened 2020-09-29
Fix Safari autofill

calebporzio:


glaesser:
While this has indeed fixed safari (including safari mobile) autofill (thank you!), it has also broken basic functionality. Now receiving error:

`Undefined index: value" `
` $unHydratedInstance->syncInput($data['name'], $data['value']);`

when toggling showing/hiding components (e.g. a button that hides/shows text). This works perfectly in 2.2.7 and below, and affects all browsers (safari, chrome etc.).

May be related to https://github.com/livewire/livewire/issues/1678?

PR #1728, opened 2020-09-30
Consider view path configuration when guessing render() view path

danharrin:
Consulted Caleb to confirm that this is a bug.

When guessing a view path for `render()`, Livewire now considers the `view_path` configuration.


calebporzio:
Thanks Dan!

calebporzio:
Hey @danharrin, I had to revert these commits after merging, heads up.

This PR actually breaks all my Livewire apps because it forces something like `view('the.full.path.to.resources.views.livewire.etc')`

Just letting you know.

danharrin:
Oh wow, really? I was not aware of that at all, especially since all my tests passed fine.

Is there any way we could implement this behaviour in a way that doesn't break your apps like that?

PR #1735, opened 2020-10-01
Add assertViewIs testing helper

jasonlbeggs:
This PR adds an `assertViewIs` testing helper method. This will allow users to ensure that a certain view is returned from a component. It's similar to [Laravel's `assertViewIs` method](https://laravel.com/docs/8.x/http-tests#assert-view-is).

Usage:

```php
    ->assertViewIs('livewire.some-component');
```


calebporzio:
Thanks!

PR #1738, opened 2020-10-01
Allow a component to skip its initial render

danharrin:
Components can now skip their initial render using `$this->skipRender()`, which has been part of the component API since v2.

This is especially useful when redirecting in a `mount()` method, as it skips the requirement for an obsolete view:

```php
class ComponentWithSkipRender extends Component
{
    public function mount()
    {
        $this->redirect('foo');

        $this->skipRender();
    }
}
```

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -107,6 +107,8 @@ abstract class Component
 
     public function renderToView()
     {
+        if ($this->shouldSkipRender) return null;
+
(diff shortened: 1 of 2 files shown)


barryvdh:
Why did you close this? Without this PR, the view is not rendered but the `render()` method is still called on the component, so executing queries/gathering data.

PR #1742, opened 2020-10-02
add safari autofill support for .lazy fields

jimpeters:

Wanted by safari users, hated by anyone writing code to detect input changes.

No

If safari would let you automate autofill it sure would!

When using Safari autofill with fields set with .lazy modifier the changes are not picked up.  This is all I could see to do and it works 100% for me after the change and 0% of the time before.  All other field modifiers work without this change so thank you for your work on this Caleb!  Am I barking up the wrong tree in terms of where to handle this?  I ran all the dusk tests and they passed still.


calebporzio:
I think this is the PERFECT solution actually. Thanks!

PR #1745, opened 2020-10-03
Failing test - showcase DOM update from no events/with directive for binding events issue

fabriciojs:
This is a failing test for the case of DOM updating issue which misses updating the directive event binding `wire:click`.

Test `test_button_dom_update_click_event` will fail, while test `test_button_dom_update_click_event_with_id` will succeed.

There will be an issue for this soon.

### DIFF
--- a/tests/Browser/Actions/Test.php
+++ b/tests/Browser/Actions/Test.php
@@ -120,7 +120,7 @@ class Test extends TestCase
         });
     }
 
-    public function test_sibling_buttons()
+    public function test_sibling_buttons_no_ids()
     {
         $this->browse(function ($browser) {
             Livewire::visit($browser, Component::class)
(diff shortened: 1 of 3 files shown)


tests/Browser/Actions/Test.php
```diff
@@ -119,4 +119,34 @@ public function test()
...
+    {
+        $this->browse(function ($browser) {
+            Livewire::visit($browser, Component::class)
+                /**
+                 * sibling buttons
+                 */
+                ->press('@show.button.actions')
+                ->waitForLivewire()->pause(100)->assertPresent('@button.with-actions')
+                ->press('@button.with-click')
+                ->waitForLivewire()->pause(100)->assertSeeIn('@output', 'button with wire:clicked got triggered')
```
nuernbergerA, line 133: thats not how `waitForLivewire()` is designed
For multiple assertions and actions before livewire renders AGAIN
```php
waitForLivewire(function ($b) {
    $b->click('something')
})
// livewire has done its magic and you proceed here after all is rendered
```

sometimes its just a oneliner so you can just pass the single action like HigherOrderProxy stuff
```php
waitForLivewire()->click('something')
// livewire has done its magic and you proceed here after all is rendered
```

```php
            Livewire::visit($browser, Component::class)
                /**
                 * sibling buttons
                 */
                ->waitForLivewire()->press('@show.button.actions')
                ->assertPresent('@button.with-actions')
                ->waitForLivewire()->press('@button.with-click')
                ->assertSeeIn('@output', 'button with wire:clicked got triggered')
            ;
```

just to streamline, we use `click()` over `press()` in 99%

tests/Browser/Actions/Test.php
```diff
@@ -119,4 +119,34 @@ public function test()
...
+            ;
+        });
+    }
+
+    public function test_button_dom_update_click_event_with_id()
```
nuernbergerA, line 138: Due to **performance** reasons, we structured all browser **tests** to have **only one** test **method** per file and just chain all "real" tests within it.

To separate them we use doc-blocks and if needed a `->refresh()` call.

nuernbergerA (changes requested):

nuernbergerA:
Hey @fabriciojs,

thanks for your PR!
That's the way we like it, you deliver a failing test - we fix it :+1: 

Cheers Adrian

fabriciojs:
@nuernbergerA thanks for the review! I updated the test accordingly to your remarks.

Now the last assertion is the only failing part - `->assertSeeIn('@output', 'button with wire:clicked got triggered')` at line 136.

We will try explore on the solution as well - we can keep discussions here on any progress to keep everyone posted.

Thanks!

carlalexander:
Just wanted to tag in with a workaround if someone runs into this issue. I couldn't figure out why my modal wouldn't close after the content was swapped. I just added a random ID with `Str::random()` and the `wire:click` event was picked up.

```blade
<div class="mt-5 sm:grid sm:grid-cols-2 sm:gap-3 sm:grid-flow-row-dense">
    @if (empty($token))
        <button>{{ __('button.create') }}</x-button-primary>
        <button wire:click="$set('opened', false)">{{ __('button.cancel') }}</x-button-secondary>
    @else
        {{-- This needs an ID --}}
        <button id="{{ Str::random() }}" wire:click="$set('opened', false)">{{ __('button.close') }}</x-button-secondary>
    @endif
</div>
```

calebporzio:
Thanks for posting the solution @carlalexander - there is no good way for morphdom to know that those two conditionals are different, so it just tries to update one to the other.

Adding keys is the way to go here.

Closing this

PR #1746, opened 2020-10-03
Check if manifest.json exists

princejohnsantillan:
The issue is when vendor/livewire folder exists but is empty it will still go into the if block and look for the manifest.json file therefore throwing an error. In my case this happened when I ran "php artisan vendor:publish" and selected livewire, once I saw the files created I discarded the scrips and only kept the config file because that's all I needed. VS code only deleted the files but not the folder. 

Googling the issue, I found that I'm not the only one experiencing this. Most advise was to publish the assets but I don't want to keep a copy since I'm not customizing it and I want it kept in the vendor folder.

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -191,7 +191,7 @@ HTML;
         $assetWarning = null;
 
         // Use static assets if they have been published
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

PR #1773, opened 2020-10-07
Improve handling for missing properties in testing

jasonlbeggs:
This PR improves how missing properties or computed properties are handled by the `TestableLivewire` class. Previously, the `TestableLivewire` class would catch all exceptions that occured in a computed property making it harder to debug errors when writing tests. With these changes, only the `PropertyNotFound` exception will be caught and all other exceptions will be allowed through.


jasonlbeggs:
I'm not sure where those two errors are coming from. If someone who is more familiar with the Livewire internals has time to take a look, that'd be great! I'm not sure if it's an issue with my changes or if my changes might've uncovered an issue.

calebporzio:
Perfect! Thank you Jason!

PR #1777, opened 2020-10-08
Support stringables as public properties

michaeldyrynda:

I did not, however, it makes sense to support Laravel's `Stringable` class, in the same way Livewire supports `Collection` and `Model` objects.

No

Yes; best I can tell. Let me know if I've missed some coverage and I'll add it!

We use `Stringable` - or `Str::of('string')` - in a couple of places in our Livewire components, but in order to render them as public properties, we need to explicitly call `__toString()`, which is extraneous and contrary to Livewire's penchant for killing the noise 🥳


calebporzio:
Woot! This is perfect, THANKS @michaeldyrynda !

PR #1778, opened 2020-10-08
Unique rule failes during hydration with ignore()

nuernbergerA:
When validating a **user** with a **unique** rule I can't **ignore** the given user because it's **not hydrated** and the fields a still prefixed during hydration.

I added a failing test with 3 variations to illustrate both issues.

Happy fixing @calebporzio :smile:

### DIFF
--- a/tests/Unit/EloquentModelValidationTest.php
+++ b/tests/Unit/EloquentModelValidationTest.php
@@ -3,8 +3,10 @@
 namespace Tests\Unit;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Validation\Rule;
 use Livewire\Component;
 use Livewire\Livewire;
+use Livewire\ObjectPrybar;
 use Sushi\Sushi;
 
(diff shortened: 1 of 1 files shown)


iksaku:
To me, this seems related to #1652 for the underlying bug specification. Maybe we could open a discussion for an upcoming fix to this test?

calebporzio:
@nuernbergerA, I think @iksaku is right. We previously closed an issue related to this issue. here was my response: https://github.com/livewire/livewire/pull/1652#issuecomment-700857927

However, apparently it has cropped up again. Thoughts?

iksaku:
Maybe give a way for devs to specify sensible read-only fields to populate beforehand?

Complexity increases in the code base, but may effectively help with this type of cases, when you need to use a non-modifiable attribute before validating or working with the upcoming modifiable attributes.

calebporzio:
Because this is such a sensitive are of the code flow, I'm going to close this in favor of more people expressing this need.

Thanks!

PR #1797, opened 2020-10-10
[CI] fixed indentation

nuernbergerA:
CI currently failing because PHP7.2 fails due to a wrong indentation.


calebporzio:
Thanks Adrian!

PR #1800, opened 2020-10-11
Autoload components by namespace

iamabdeldayem:
This PR will allow developers to autoload components in a similar way to Laravel autoload class components feature https://github.com/laravel/framework/pull/34144

```
Livewire::componentNamespace('Package\\Http\\Livewire', 'package');
```

Then use it as

```
@livewire('package::livewire-component')
```

or

```
<livewire:package::livewire-component />
```

Also, It will look for the view (if the render method isn't provided for Livewire component) using

```
view('package::livewire.livewire-component');
```

which is registered using

```
$this->loadViewsFrom('path/to/package/views', 'package');
```

-----
:heavy_check_mark: Unit test

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -5,8 +5,10 @@ namespace Livewire;
 use Illuminate\View\View;
 use BadMethodCallException;
 use Illuminate\Support\Str;
+use Illuminate\Support\Arr;
 use Illuminate\Routing\Route;
 use Illuminate\Support\ViewErrorBag;
+use Illuminate\View\ViewFinderInterface;
 use Illuminate\Support\Traits\Macroable;
 use Illuminate\Contracts\Container\Container;
 use Livewire\Exceptions\CannotUseReservedLivewireComponentProperties;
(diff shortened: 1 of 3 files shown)


src/Component.php
```diff
@@ -97,6 +99,22 @@ public static function getName()
         return $fullName;
     }
 
+    public static function getPrefix()
+    {
+        return Str::startsWith(static::class, app('livewire')->getComponentNamespaces())
+            ? array_search(Arr::first(app('livewire')->getComponentNamespaces(), function ($namesapce) {
+                return str_starts_with(static::class, $namesapce);
+            }), app('livewire')->getComponentNamespaces())
+            : null;
```
iksaku, line 108: Looks like you're iterating the same array 3 times in a row. I'll send a PR to your branch with a simplified version of this algorithm :smiley:

iksaku (changes requested):

src/Component.php
iamabdeldayem: Thanks, merged 👌

PR #1802, opened 2020-10-12
fix assertSet() when testing string that matches a function name

vardyb:

This relates to the following forum discussion:
https://forum.laravel-livewire.com/t/issue-with-basic-assertset-following-upgrade-to-v2/1977

No, the PR is for related changes

Yes, I have updated the existing tests.

After upgrading to Livewire v2.2.9 some of my assertSet() failed and not related to using assertPayloadSet().

In v2 if you try to test a class property with a string value that is the same name as a valid function it runs the function rather than testing the string for a match.  e.g.

```
Livewire::test(SomeComponet::class)
                ->assertSet('name', 'info');
```

Will fail with the error

```
Failed asserting that null is true.
```

In this case the `assertSet()` function runs `info('info')` which returns a null and actually writes

```
[2020-10-12 07:06:35] testing.INFO: info
```

In the Laravel log.

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -13,7 +13,7 @@ trait MakesAssertions
 {
     public function assertSet($name, $value)
     {
-        if (!is_string($value) && is_callable($value)) {
+        if (! is_string($value) && is_callable($value)) {
             PHPUnit::assertTrue($value($this->get($name)));
         } else {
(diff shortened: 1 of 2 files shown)


vardyb:
Not sure if `assertPayloadSet()` should also be updated in a similar way?  Please advise.

calebporzio:
Ooooh that's a nasty little bug. Thanks!

Issue #1748, opened 2020-10-04
Undefined variable $attributes with default layout


mokhosh:
### Description
In my `layouts.app` I have `$attributes` on the `body` tag, to be able to pass `Alpine.js` or `class` attributes when using the layout.
If I use the Livewire shorthand for my `route` I get the error:
```
Undefined variable $attributes
```
And I have to do this to fix it:
```php
public function render()
{
    return view('livewire.component-page')->layout('layouts.app', [ 'attributes' => '' ]);
}
```

### Exact steps to reproduce
1. Put `{{ $attributes }}` in your `layouts.app`
2. Use shorthand routing with Livewire

**Context**
- Livewire version: latest
- Laravel version: latest
- Browser: Chrome

RomainMazB:
@mokhosh 
Maybe I'm facing the same problem, and I'm currently trying to fix it.

Are you using an AppLayout class and blade "x-tag" to render your "basic" views?

Like so:
```html
<x-app-layout>
    
</x-app-layout>
```

mokhosh:
Yes I am

RomainMazB:
So the problem is described and solved in the PR I've made https://github.com/livewire/livewire/pull/1807

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

PR #1807, opened 2020-10-12
Allow the use of layout as a Component

RomainMazB:
I did not create an issue, but it's probably related with this one: #1748 

No.

No.

When using layout as a Component, like it's done in Jetstream (as an example) for the [AppLayout](https://github.com/laravel/jetstream/blob/master/stubs/livewire/app/View/Components/AppLayout.php) and [GuestLayout](https://github.com/laravel/jetstream/blob/master/stubs/livewire/app/View/Components/AppLayout.php), we are able to define some "complex data" we don't want to put in our blade file:
```php
class AppLayout extends Component
{
    public bool $canCreateCharacter;

    public function __construct()
    {
        $user = Auth::user();

        $userCharacters = $user->characters->map(
            static fn ($character) => $character->id
        );

        // Retrieve the world not full where the user doesn't have a character yet
        $this->canCreateCharacter = World::whereDoesntHave('characters',
                static fn (Builder $query) =>
                $query->whereIn('id', $userCharacters)
            )
                ->subscriptionAvailable()
                ->count() > 0;
    }

    public function render()
    {
        return view('layouts.app');
    }
}
```

But since our blade file is strongly attached to our Component class, some errors appear when we are trying to render our Livewire components using its class directly:
```php
Route::get('/', 'Livewire\\Messenger\\Messenger')->name('messenger.index');
```
> ErrorException
Undefined variable: canCreateCharacter (View: /home/maz/websites/terrajdr/resources/views/layouts/app.blade.php) (View: /home/maz/websites/terrajdr/resources/views/layouts/app.blade.php)

As mentioned in a thread on the [Livewire forum](https://forum.laravel-livewire.com/t/layout-class-constructor-is-not-triggered-while-rendering/1985), I've succeed into resolving it using base Livewire methods by manually instantiate the AppLayout class and passing its data as parameters:
```php
// Somewhere in my Livewire Component class
    public function render()
    {
        $view = view('livewire.messenger.index')->layout('layouts.app', (new AppLayout())->data());
    }
```
But manually add this code, referencing the default layout and passing its data manually is totally boring and we don't want that.

So I've dive into the Livewire code and searched why I needed to do that.
I finally understood that Livewire do not use x-blade component syntax, which is (I suppose) the source of the problem.

First I tried to search for a method to programmatically know if x-component was used by the layout file, but didn't find one (open to suggestion).
Then I figured out that I just needed to create a view file to render the layouts that are extending Component class
I've made some tests, using a x-blade specific layout to render my components and it worked perfectly well.

Since x-blade syntax is not natively supported by Laravel 7 which is supported by Livewire and assuming the fact that most of us don't use Blade Component to render layouts, I propose to just set a config boolean to change the default behavior to use this feature.

I've created the viewMacro method to change the default layout and pass data to it the same way we use `->layout()` or `->extend()`. You can just use `->xLayout('custom-layout', ['parameter' => 'custom data'])`

Doing so, we can use basic feature for some layout, and xLayout to render specific Component extending layout.

I'm totally open to suggestion to upgrade my code, I think this feature in my implementation-style or another should be merged in near future.

5️⃣ Thanks @calebporzio for providing this good package! 🙌

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -97,4 +97,15 @@ return [
 
     'manifest_path' => null,
 
+    /*
+    |--------------------------------------------------------------------------
+    | X-Layout
+    |--------------------------------------------------------------------------
+    |
+    | This value sets the default layout render behavior to use x-blade components styles.
+    |
+    */
+
+    'uses_x-layout' => false
+
 ];
--- a/src/Component.php
+++ b/src/Component.php
@@ -49,14 +49,16 @@ abstract class Component
             ->mount($componentParams)
             ->renderToView();
 
-        $layoutType = $this->initialLayoutConfiguration['type'] ?? 'component';
+        $layoutType = $this->initialLayoutConfiguration['type'] ?? config('livewire.uses_x-layout') ? 'x-layout' : 'component';
+
+        $view = $this->initialLayoutConfiguration['view'] ?? config('livewire.uses_x-layout') ? 'app-layout' : 'layouts.app';
 
         return app('view')->file(__DIR__."/Macros/livewire-view-{$layoutType}.blade.php", [
-            'view' => $this->initialLayoutConfiguration['view'] ?? 'layouts.app',
+            'view' => $view,
             'params' => $this->initialLayoutConfiguration['params'] ?? [],
             'slotOrSection' => $this->initialLayoutConfiguration['slotOrSection'] ?? [
-                'extends' => 'content', 'component' => 'default',
-            ][$layoutType],
+                    'extends' => 'content', 'component' => 'default', 'x-layout' => 'default'
+                ][$layoutType],
             'manager' => $manager,
         ]);
     }
--- a/src/Component.php
+++ b/src/Component.php
@@ -49,9 +49,9 @@ abstract class Component
             ->mount($componentParams)
             ->renderToView();
 
-        $layoutType = $this->initialLayoutConfiguration['type'] ?? config('livewire.uses_x-layout') ? 'x-layout' : 'component';
+        $layoutType = $this->initialLayoutConfiguration['type'] ?? (config('livewire.uses_x-layout', false) ? 'x-layout' : 'component');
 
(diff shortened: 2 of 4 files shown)


calebporzio:
Thank you for the PR, however I don't think this is the right way forward.

I could see wanting to pass a Blade component class into `->layout()` and I would be open to that, but none of the extra config-related or "x-layout" nomenclature stuff.

Thanks!

RomainMazB:
> Thank you for the PR, however I don't think this is the right way forward.
> 
> I could see wanting to pass a Blade component class into `->layout()` and I would be open to that, but none of the extra config-related or "x-layout" nomenclature stuff.
> 
> Thanks!

I will review my code to allow `->layout()` to get a Blade component class in parameter and parse it with the special `livewire-view-x-layout.blade.php` (this part is OK I suppose, because I don't see any way to handle without this ? @calebporzio ).

This is I agree a proper way.

RomainMazB:
On the other hand, after thinking about it. Using the native ->layout() will force the programmer to pass his default layout into each component.
```php
// First component
{
    ...
    public function render()
    {
        return view('livewire.show-posts')
            ->layout(AppLayout::class);
    }
}


// Second component
class ShowComments extends Component
{
    ...
    public function render()
    {
        return view('livewire.show-posts')
            ->layout(AppLayout::class);
    }
}


// Third component
class ShowRates extends Component
{
    ...
    public function render()
    {
        return view('livewire.show-posts')
            ->layout(AppLayout::class);
    }
}

// And so on ...
```

Maybe a config key to set the default layout may be useful in this case and could provide both "x-layout" or basic layout (if the file of the default layout is not layouts/app.blade.php)

rjdjohnston:
would be great if this was merged

RomainMazB:
@rjdjohnston A new PR is in progress here https://github.com/livewire/livewire/pull/2499 , stay tuned ;)

PR #1809, opened 2020-10-12
Ensure base64 encoded strings dont lead to unwanted sub-directories

### LINKED ISSUE #1798: Error when uploading files with Arabic names
# Description
When uploading a file with an Arabic name (Not all names) and try to store the file, an exception is thrown: `League\Flysystem\FileNotFoundException`

The initial upload to the temp directory is successful, and the file exists in the `livewire-tmp` directory. However, it's stored in a subdirectory, which makes it unavailable because the file name is stored without the subdirectory. 

My guess is that when the original file name is encoded using `base64_encode` [here](https://github.com/livewire/livewire/blob/0cde489ba3881fb03802f9c9e0b820640f4e7a42/src/TemporaryUploadedFile.php#L127), the generated name has a forward slash, thus creating a subdirectory. 

### Exact steps to reproduce
- Create a basic file upload input (as explained in docs)
- Upload a file with an Arabic name. Try this: `دليل.pdf` or this file as an example: [https://gofile.io/d/xQM66H](https://gofile.io/d/xQM66H)
- Check the `livewire-tmp` for the file, it will be in a subdirectory.
- Try to store the file using: `$this->photo->store('photos')` or even validate the file on submit.

**Context**
- Livewire version: 2.27
- Laravel version: 8.6.0
- Browser: Chrome

nuernbergerA:
The real issue behind #1798 is that a base64 encoded string can contain a `/`, so the uploaded file will be stored in a sub-folder.
@alhoqbani Please could you check if this fixes your issue?

---
Closes #1798 
Co-Authored-By: @mathiasonea

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -124,7 +124,7 @@ class TemporaryUploadedFile extends UploadedFile
     public static function generateHashNameWithOriginalNameEmbedded($file)
     {
         $hash = Str::random(30);
-        $meta = '-meta'.base64_encode($file->getClientOriginalName()).'-';
+        $meta = Str::of('-meta'.base64_encode($file->getClientOriginalName()).'-')->replace('/', '_');
         $extension = '.'.$file->guessExtension();
 
         return $hash.$meta.$extension;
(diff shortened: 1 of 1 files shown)


alhoqbani:
Yes, it does fix the issue. Thanks @nuernbergerA

calebporzio:
Thanks @nuernbergerA !

PR #1836, opened 2020-10-17
don't publish config with --assets flag - closes #1831

### LINKED ISSUE #1831: php artisan livewire:publish --assets should not copy the config over
### Description

Running: `php artisan livewire:publish --assets` then generated:

```
Copied File [/vendor/livewire/livewire/config/livewire.php] To [/config/livewire.php]
Publishing complete.
Copied Directory [/vendor/livewire/livewire/dist] To [/public/vendor/livewire]
Publishing complete.
```

You just overwrote my config forcing me to set it up again - or in the future - make backups before publishing new assets after updates.

### Exact steps to reproduce

- Run: `php artisan livewire:publish --assets`
- See: 

```
Copied File [/vendor/livewire/livewire/config/livewire.php] To [/config/livewire.php]
Publishing complete.
Copied Directory [/vendor/livewire/livewire/dist] To [/public/vendor/livewire]
Publishing complete.
```

- Should see:

```
Copied Directory [/vendor/livewire/livewire/dist] To [/public/vendor/livewire]
Publishing complete.
```

**Context**
- Livewire version: [2.3.0]
- Laravel version: [8.10.0]
- Browser: N/A - **Command issue**.

austenc:
A potential fix for #1831, which brings up a good point -- the `publish` command probably shouldn't overwrite previously-published config files.

This feature could be supported in several ways, so the intent of this PR is to open up that discussion. Here are a couple options:

**Option 1 -- This PR**
`livewire:publish` does only config (same as now)
`livewire:publish --assets` does only assets and no config (currently it does both, this PR changes that)

**Option 2 -- More flexible?**
`livewire:publish` would do everything (change from current behavior, docs would need updating)
`livewire:publish --config` is only config
`livewire:publish --assets` is only assets

Thinking Option 2 might be nicer for end users, if you agree I'll update this PR to support that. 👍

### DIFF
--- a/src/Commands/PublishCommand.php
+++ b/src/Commands/PublishCommand.php
@@ -6,16 +6,31 @@ use Illuminate\Console\Command;
 
 class PublishCommand extends Command
 {
-    protected $signature = 'livewire:publish {--assets : Indicates if Livewire\'s front-end assets should be published}';
+    protected $signature = 'livewire:publish 
+        { --assets : Indicates if Livewire\'s front-end assets should be published }
+        { --config : Indicates if Livewire\'s config file should be published }';
 
     protected $description = 'Publish Livewire configuration';
 
     public function handle()
     {
         if ($this->option('assets')) {
-            $this->call('vendor:publish', ['--tag' => 'livewire:assets', '--force' => true]);
+            $this->publishAssets();
+        } elseif ($this->option('config')) {
+            $this->publishConfig();
         } else {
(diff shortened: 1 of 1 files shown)


calebporzio:
Option 2 is great.

Not a breaking change so I'm good with it and yeah, we'll just update the docs.

Thanks Austen!

calebporzio:
Thanks Austen!

Issue #1783, opened 2020-10-08
The shortenModelAttributes does handle properties with underscores


aromer:
### Description
When using model binding, with keys like `model.property` there is a method that removes the model prefix. When properties have underscores like `model.snake_case_property` these are skipped, resulting in error messages like "The model.snake case property is required."

### Exact steps to reproduce
Create a component with a validation rule like:
```php
    protected array $rules = [
        'post.snake_case_property' => 'required',
    ];
```
Submit the form and fail the validation, you should see an improperly formatted message.

### Stripped-down, copy-pastable code snippets
https://github.com/livewire/livewire/blob/90f12fee988f5084758c514b77eb87c9306fedb7/src/ComponentConcerns/ValidatesInput.php#L191-L205

The comparison ` if ($key === $validator->getDisplayableAttribute($key))` converts the key to remove underscores in the getDisplayableAttribute() method causing the comparison to fail.

**Context**
- Livewire version: [e.g. 2.2.3]
- Laravel version: [e.g. 8.3.1]
- Browser: [e.g. Chrome, Safari]

zepfietje:
I think I ran into this issue with nested array properties too.

zepfietje:
@calebporzio, I suppose this issue can be closed now?

danharrin:
Thanks @zepfietje!

aromer:
Thank you everyone. I can confirm this works in the original reported application.

PR #1838, opened 2020-10-17
Fix snake cased model attribute short name in validation error - #1783

austenc:
This is a fix related to #1783.

The included test represents the error from the issue. However, the code to make it pass is kinda gross. 

Perhaps we can figure out a better way to handle that comparison and conversion to a "displayable" attribute?


calebporzio:
Thanks @austenc!

PR #1845, opened 2020-10-19
Fix Laravel Echo presence channels listening

developeritsme:
Fix Laravel Echo presence channels registering listeners

`Uncaught TypeError: Echo.join(...)[event_name] is not a function`


redbastie:
Did you figure it out? I'm getting the same error from this code:

    protected $listeners = [
        'echo-presence:lobby,MatchMade' => 'matchMade',
    ];

Issue #1607, opened 2020-09-15
Array reindexing with nested arrays not always working right


austenc:
### Description

Sometimes when an array public property is reindexed on the PHP side things go haywire. I was working with an array of cities and noticed that they weren't in perfect alphabetical order in the dropdown (select) field I was building.  

From what I can tell it's caused by the `reindexArrayWithNumericKeysOtherwiseJavaScriptWillMessWithTheOrder` method. It has something to do with the `uksort` being called recursively with `array_map`. I would've preferred to submit a PR but haven't figured it out yet!

Here's the line: 
https://github.com/livewire/livewire/blob/master/src/HydrationMiddleware/NormalizeDataForJavaScript.php#L25

In `1.x` it's here:
https://github.com/livewire/livewire/blob/1.x/src/Component.php#L175

### Steps to reproduce
In the example, you'll notice the array is alphabetized in the component definition, but when it gets to the view it is all jumbled.

https://laravelplayground.com/#/snippets/ccd2908b-0f16-422f-998f-18f2c82118e8

**Context**
- Livewire version: 1.3.4 (likely 2.x also)
- Laravel version: 7.x
- Browser: Brave

austenc:
So this actually comes down to how javascript doesn't care about the ordering of keys within objects.

Even though it's reordered on the PHP side before hitting JS, it gets reordered again after JS gets ahold of it anyway. 

The _solution_ is to use numerically keyed arrays when you need order to matter. Not associative ones. 

Every solution (on the livewire side) that I can think of is **really gross.** So given that, I'm closing this issue.

calebporzio:
Thanks for following up @austenc !

PR #1859, opened 2020-10-21
Fix uksort() issue with arrays of greater than 16 items

### LINKED ISSUE #1858: select tags with more than 16 elements and strings as keys render shuffled
### Description
So I have an array that serves as a list of choices for a `<select>`. The keys are strings and the values are strings. If the number of choices is 16 and below, no issues at all. But as soon as I use more than 16 choices, the <option> values are shuffled on initial render and in every update.

### Exact steps to reproduce
Create a component with this code below and try to view it in the browser. You'll see all the options are shuffled and when you select a different option, the options are shuffled again.

```blade
<div>
    <select name="provinces" id="provinces" wire:model="choice">
        @foreach ($choices as $key => $value)
            <option value="{{ $key }}">{{ $key }}: {{ $value }}</option>
        @endforeach
    </select>

    <div>Choice: {{ $choice }}</div>
</div>

```

```php
class Provinces extends Component
{
    public $choices;

    public $choice;

    public function mount()
    {
        $this->choices = [
            'a' => 'A',
            'b' => 'B',
            'c' => 'C',
            'd' => 'D',
            'e' => 'E',
            'f' => 'F',
            'g' => 'G',
            'h' => 'H',
            'i' => 'I',
            'j' => 'J', // 10
            'k' => 'K',
            'l' => 'L',
            'm' => 'M',
            'n' => 'N',
            'o' => 'O', // 15
            'p' => 'P',
            'q' => 'Q',
        ];

        $this->choice = 'c';
    }

    public function render()
    {
        return view('livewire.provinces');
    }
}
```

**Context**
- Livewire version: 2.3.0
- Laravel version: 8.11.2
- Browser: Chrome

danharrin:
Closes #1858.

`uksort()` does not retain the original array order when it contains over 16 items, as demonstrated [here](https://laravelplayground.com/#/snippets/898cda72-bebd-4bf8-8594-80f9351ab1a6).


austenc:
Wondering if this was the cause of my issue here also: https://github.com/livewire/livewire/issues/1607 

Will test this out, nice one @danharrin!

nuernbergerA:
For me `uksort` works fine, the example with echo inside the sort function makes no sense.

https://web.tinkerwell.app/#/snippets/798811d6-65ca-4ab5-99ea-c7ca28a683f2 would work fine

danharrin:
For ref: just had a chat with @nuernbergerA, although his example works, the original order of string keys is not retained after the `uksort()`.

calebporzio:
Wow, that's a weird one, thanks @danharrin  - mind adding a test for this?

nuernbergerA:
Just for clarification how `uksort` works under the hood:
for smaller arrays (16 or fewer items) it compares entry 1/2, 2/3, 3/4, and so on
for a larger array (17 or greater) it compares entry 1/9, 9/17, and so on to gain performance.

In our use-case we dont want to sort the entire array so a custom solution has to be used as @danharrin provided :+1:

danharrin:
> Wow, that's a weird one, thanks @danharrin  - mind adding a test for this?

@calebporzio I _would_, not sure if a test for 17 array items would be useful or not though, since we've completely removed the use of `uksort()` so that limitation is completely gone. Any other ideas?

barryvdh:
`uksort()` like it's implemented now does have an issue on PHP8 though:

> ErrorException
> uksort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero

nuernbergerA:
@danharrin you can merge master to fix your tests

calebporzio:
Thanks @danharrin !

ahinkle:
Hey there, unfortunately this PR caused a breaking change. Please see #2049.

iAmKevinMcKee:
Confirming this broke my app as well. Had to revert to 2.3.1 to fix.

src/HydrationMiddleware/NormalizeDataForJavaScript.php
```diff
@@ -10,17 +10,19 @@ protected static function reindexArrayWithNumericKeysOtherwiseJavaScriptWillMess
...
+        }, ARRAY_FILTER_USE_KEY);
+        ksort($itemsWithNumericKeys);
+
+        $itemsWithStringKeys = array_filter($value, function ($key) {
+            return ! is_numeric($key);
+        }, ARRAY_FILTER_USE_KEY);
 
-            if (! is_numeric($a)) return 1;
-        });
+        $normalizedData = array_merge($itemsWithNumericKeys, $itemsWithStringKeys);
```
nuernbergerA, line 25: Instead of `array_merge` we should use `+` to keep keys

nuernbergerA (commented):
WIll PR soon

nuernbergerA:
#2058 should fix it

PR #1864, opened 2020-10-21
Add devtools support when debug mode is enabled

mpociot:
This PR adds a new property to the Livewire object, depending on whether or not debug mode is enabled.
The Livewire Devtools can then inspect this property and show or hide the Livewire devtools.


ryangjchandler:
Since the dev tools aren't official, doesn't it make more sense to have a generalised name such as `developmentMode` or something similar?

calebporzio:
Thanks Marcel!

Issue #1776, opened 2020-10-07
fill() method doesn't work with model-properties binding


zofe:
### Description
I'm trying to set property that is binded to a model:

```
...
class PartnerRegistration extends Component
{
    public $company;

    protected $listeners = [
        'set:address' => 'setAddress'
    ];

    protected $rules = [
        'company.address' => 'required',
    ];

    public function mount($rid = null)
    {
        $this->company = new Company();
    }

    public function setAddress($address)
    {
        $this->fill(['company.address'=>$address]);
    }
...

```

set:address is called via `window.livewire.emit('set:address', 'my new address')` 
from a vue component (i've also livewire-vue 0.3.1).
the problem is inside livewire, because setAddress is called, but "company.address" doesn't change.

- Livewire  2.2.9
- Laravel 8.8.0
- Chrome 85.0.4183.121

moonofmylife:
Use $this->syncInput

austenc:
@zofe couldn't you just use $this->company->address = $address?

Or, even better, use `wire:model`?

zofe:
@austenc wire:model is absolutely the best choice, but in my case the "component" where address  handled is a vue component (with google places address api interactions, so a complex stuff to use direct wire:model binding)

and yes $this->company->address it's a smarter way to do it, thanks.
Anyway `$this->fill(['company.address'=>$address]);` doesn't seem to work, so  I decided to create this issue report.

moonofmylife:
@zofe have you tried using sync Input? 
`$this->syncInput('company.address', $address);`

zofe:
@moonofmylife really thanks, 
tested now  this->syncInput('company.address', $address);  and it works, and seems smarter than `$this->company->address `  because i need to fill  multiple fields (address, streetnumber, etc..)

There is a docs about  **syncInput** ?  I did not find on livewire site

moonofmylife:
@zofe binding (wire:model) uses this function. I found it when I was looking at the source code. It was also very necessary for me in Vue

calebporzio:
Hmm I think `$this->fill()` should work for this case.

I consider this a bug.

The current implementation of `$this->fill()` doesn't support dot notation.

I think this would be a fairly simple PR

zofe:
ok
https://github.com/livewire/livewire/pull/1872

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

[...]

PR #1866, opened 2020-10-21
Fixed broken test since Laravel v8.11.0

nuernbergerA:
fixes #1865

PR #1872, opened 2020-10-21
fix fill property with dot notation

zofe:
fix for https://github.com/livewire/livewire/issues/1776

### DIFF
--- a/src/ComponentConcerns/InteractsWithProperties.php
+++ b/src/ComponentConcerns/InteractsWithProperties.php
@@ -136,11 +136,11 @@ trait InteractsWithProperties
     public function fill($values)
     {
         $publicProperties = array_keys($this->getPublicPropertiesDefinedBySubClass());
-        
+
         if ($values instanceof Model) {
             $values = $values->toArray();
         }
-        
+
         foreach ($values as $key => $value) {
             if ($this->containsDots($key)) {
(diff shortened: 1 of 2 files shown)


src/ComponentConcerns/InteractsWithProperties.php
```diff
@@ -136,14 +136,22 @@ public function propertyIsPublicAndNotDefinedOnBaseClass($propertyName)
...
-            if (in_array($key, $publicProperties)) {
-                $this->{$key} = $value;
+            if ($this->containsDots($key)) {
+                $propertyName = $this->beforeFirstDot($key);
+
+                $keyName = $this->afterFirstDot($key);
+                $targetKey = $this->beforeFirstDot($keyName);
+                data_set($this->{$propertyName}, $targetKey, $value);
+            } else {
+                if (in_array($key, $publicProperties)) {
```
iksaku, line 152: You could remove the nesting level by using an `else if`

iksaku:
Could you please revert your changes for the #1866 PR? They should be handled separately.

danharrin:
Hey, thanks for your work here! Would you mind adding a couple of tests for this if you have time please? I'd suggest two:
- Filling a model property.
- Filling an associative array property.

calebporzio:
Thanks for the PR. Just echoing what @danharrin said - mind adding tests for this new case?

PR #1873, opened 2020-10-21
support lifecycle hooks from multiple traits - closes #1860

### LINKED ISSUE #1860: Issue with Trait Lifecycle Hooks
When applying multiple traits to a Livewire Component the mount lifecycle hooks will only run on the last trait with a lifecycle hook in the 'use' list.

Create a component and multiple traits and add to the use statement of the component.  Within the traits add the MountTraitName function.

When the component is mounted only the last declared trait will run the mountTratName lifecycle hook.

**Case 1**

```
class Resource extends Component

    use WithSort, WithSearch;

trait WithSort

    public function mountWithSort(){
          // Do Something
    }

trait WithSearch

    public function mountWithSearch(){
          // Do Something
    }
```

In the above only mountWithSearch will get called.  mountWithSort is not called.

**Case 2**

```
class Resource extends Component

    use WithSearch, WithSort;

trait WithSort

    public function mountWithSort(){
          // Do Something
    }

trait WithSearch

    public function mountWithSearch(){
          // Do Something
    }
```

In the above only mountWithSort will get called.  mountWithSearch is not called.

Livewire Version 2.3.0
Laravel Version 8.11.2

austenc:
This PR adds support for using multiple traits with lifecycle methods. 

Original issue here: 
https://github.com/livewire/livewire/issues/1860

### DIFF
--- a/src/RenameMe/SupportComponentTraits.php
+++ b/src/RenameMe/SupportComponentTraits.php
@@ -87,7 +87,7 @@ class SupportComponentTraits
             $methods = $this->componentIdMethodMap[$component->id]['dehydrate'] ?? [];
 
             foreach ($methods as $method) {
-                ImplicitlyBoundMethod::call(app(), $method,);
+                ImplicitlyBoundMethod::call(app(), $method);
             }
         });
     }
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @austenc !

gwleuverink:
Any idea when it'll be tagged for release? No pressure :) I'm trying to plan my work around it. Thanks!

Issue #1592, opened 2020-09-14
Initial data "effects" don't run the emit events


timrspratt:
### Description
I have a use case where a nested component triggers `$this->emitUp()` within its `mount` method, and this data is included in the `wire:initial-data` (as `effects.emits`), yet it appears not to be triggered in the `Component` constructor `js/component/index.js` when the component is processed by the browser.

**Context**
- Livewire version: 2.2.2
- Laravel version: 8.1.0
- Browser: Chrome

timrspratt:
I've managed to use `wire:init` for my use case, but leaving this open for your consideration incase it would still be counted as a bug. Thanks

calebporzio:
Thanks for reporting this.

Yeah, `$this->emit` calls in `mount` SHOULD be received by other components on the page on-load.

isaackearl:
This bug has affected my team recently. I made another ticket that was closed because I didn't realize this already existed.  Here is the simple example I created there


```
class ComponentOne extends Component
{

    // emitting an event to be picked up by the other component
    public function mount(){
      $this->emit('hello');
    }

    public function render()
    {
        return view('component-one');
    }
}
```
second component
```
class ComponentTwo extends Component
{

    protected $listeners = ['hello' => 'doSomething'];    

    public function doSomething(){
        dump('this never happens');
    }

    public function render()
    {
        return view('component-two');
    }
}
```

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

isaackearl:
REOPEN

isaackearl:
I went ahead and said re-open because I believe this is a legitimate issue even though there are some workarounds.  I plan on trying to create a dusk failing test eventually, but unfortunately haven't had the time.

MACscr:
I am having this issue as well. Would love to see it working. I am finding a lot of use cases where emits aren't working as someone would expect nor documented that they won't as well and that leads to a lot of frustration.

whoisthisstud:
[...]

PR #1875, opened 2020-10-22
Providing a failing test for #1592

isaackearl:

I am trying to provide a failing test case for #1592 as requested.

no

it's only a test

I'm trying to show that when you emit an event from the mount method.  Listeners don't "hear" it.  I created a new listener on the test class so it wouldn't break other tests that also rely on the 'bar' listener within the class (or the foo property).

### DIFF
--- a/tests/Unit/ComponentEventsTest.php
+++ b/tests/Unit/ComponentEventsTest.php
@@ -19,6 +19,15 @@ class ComponentEventsTest extends TestCase
     }
 
     /** @test */
+    public function receive_event_from_mount()
+    {
+        $component = Livewire::test(ReceivesEvents::class);
+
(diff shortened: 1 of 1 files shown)


isaackearl:
Please let me know if I did this wrong.  Or if I should have submitted it some other way.  Someone suggested providing a failing test and this is my attempt to do so.

calebporzio:
Thanks for submitting @isaackearl! I love when people submit failing tests.

Unfortunately, I don't think a Unit test will properly test this because it's likely a JS issue I believe and Unit tests won't cover JS.

Do you feel capable of creating a browser test w/ Dusk?

isaackearl:
@calebporzio that makes alot of sense! I'll see if I can get to it this weekend.

Issue #1893, opened 2020-10-24
@entangle not syncing correctly.


joshhanley:
### Description

When you have an array that is @entangled with livewire and an input field that is also @entangled. 

Changes in the input are synced back to livewire, which filters down the array, which should sync it back to alpine but it doesn't.

### Exact steps to reproduce

I have created a demo repo here: https://github.com/joshhanley/EntangleArrayBug

1. Pull down the repo
2. Composer install
3. Open browser and go to site, its in the base route

I have just replaced the contents of welcome.blade.php and added a Livewire component called ListTest

**Screenshot**

![EntangleArrayBug](https://user-images.githubusercontent.com/882837/97065547-e2244580-15f9-11eb-9eae-241289af4888.gif)

### Stripped-down, copy-pastable code snippets

**Context**
- Livewire version: 2.3.0
- Laravel version: 8.11.2
- Alpine version: 2.x.x (from CDN, so I think 2.7.1)
- Browser: Chrome

vladrusu:
+1
I also run into this bug. I don't think it is array-specific, but I think that @entagle is somewhere broken. Consider this code:
`<div x-data="{ open: @entangle('showDropdown') }">
	<p>Dropdown: {{ (int)$showDropdown }}</p>
	<input class="selection form-control" type="text"
	    	x-on:keydown.arrow-down.prevent="open = true"
	    	x-on:keydown.arrow-up.prevent="open = false"
	/>
</div>
<div class="ac-dropdown" x-show.transition="open" x-on:click.away="open = false">
... dropdown contents ...
</div>
`
No matter if you press arrow up or down, $showDropdown will stay false. I expected "Dropdown" to show 0/1 when pressing arrow up/down.

joshhanley:
I have tested this with 2.3.1 and the issue is still happening. I have rebased my PR #1894  to pull in the new changes

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

joshhanley:
REOPEN

[...]

PR #1894, opened 2020-10-24
Add test for @entangle array sync issues and potential fix

joshhanley:
I have added a failing dusk test for the @entangle syncing issue I raised in #1893 

I have also added a potential fix by allowing livewire watchers to run on the way back, but I'm not sure if it is going to cause circular dependency issue. 
Maybe watchers should still be fired except for the property that just got changed by alpine.

Not sure how to go about implementing that.
I'm happy to add some tests if you can give me an idea of what sort of input/change would cause circular dependency issues.

Hope this helps! Let me know if I need to change anything.


joshhanley:
@calebporzio I have rebased this into master after you merged #1911 and just pushed the test alone, to demonstrate the failing test still happening on Livewire 2.3.1

Then I have pushed up the original fix I suggested. But as stated above, I don't know if it is going to cause circular dependency issues.

Hope this helps!

calebporzio:
Thanks for submitting a failing test. That's very helpful.

Turns out the issue has nothing to do with arrays.

Livewire blocks response watchers to prevent circular dependancies (which we've established)

The reason why this test was failing was because the "search" data triggered the request, but the "list" data changed inside the response, so the watcher for "list" was blocked because "search" was blocked.

I made the watcher-blocker (lol) more specific to only block watchers related to changes that fired the request itself.

Thanks!

joshhanley:
No worries! Glad I could help!

Yep, that makes sense. Heh watcher-blocker, I like it.

Thanks for everything you do!

PR #1896, opened 2020-10-24
Fixed pagination overflow of available page range - closes #1889

### LINKED ISSUE #1889: nextPage and previousPage method exceed page range
### Description
I'm using Livewire Pagination, and in my pagination component i use `wire:click="nextPage"` and `wire:click="previousPage"` to navigate the pages. 
If i rapidly click the pagination links before the component refresh, the method exceed upper and bottom limit, for example:

- I click rapidly previous page link and in my url i see "**page=-1**", loading no items, and the same happen for the upper bound

In the blade component i always check for `$paginator->onFirstPage()` and `$paginator->hasMorePages()`, but the click happen so quickly and the view not refresh in time.

### Exact steps to reproduce
Rapidly call nextPage and previousPage in a pagination component

```
<div class="ui pagination menu">
                @if(!$paginator->onFirstPage())
                    <div wire:click="gotoPage(1)" class="link item"> << </div>
                    <div wire:click="previousPage" class="link item"> < </div>
                @endif

                {{-- Pagination Elements --}}
                @foreach ($elements as $element)
                    {{-- "Three Dots" Separator --}}
                    @if (is_string($element))
                        <div class="disabled item"> ... </div>
                    @endif

                    {{-- Array Of Links --}}
                    @if (is_array($element))
                        @foreach ($element as $page => $url)
                            @if ($page == $paginator->currentPage())
                                <div class="active link item"><span style="font-weight:bold"> {{ $page }} </span></div>
                            @else
                                <div wire:click="gotoPage({{ $page }})" class="link item"> {{ $page }} </div>
                            @endif
                        @endforeach
                    @endif
                @endforeach

                @if($paginator->hasMorePages())
                    <div wire:click="nextPage" class="link item"> > </div>
                    <div wire:click="gotoPage({{ $paginator->lastPage() }})" class="link item"> >> </div>
                @endif
            </div>
```

**Context**
- Livewire version: 2.2.9
- Laravel version: 8.10.0
- Alpine version: 2.3.0
- Browser: Firefox

austenc:
If you have a slow connection (or are really fast at clicking), you can exceed the range of available pages.

This is described in #1889.

I added some Unit tests for this, but ran into a blocker when it came to implementing the `nextPage` portion. 
Since there's not a specific paginator instance in the `WithPagination` trait, we can't easily determine the total number of pages to restrict the `$page` variable (as I did with `previousPage()`).

Instead, I added `wire:loading.attr="disabled"` to the default pagination views to prevent this from happening. It's a bummer we couldn't test _everything_ in the unit test, but it seemed better than trying to hack a paginator instance into everything 🤮 .

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -37,7 +37,7 @@ trait WithPagination
 
     public function previousPage()
     {
-        $this->setPage($this->page - 1);
+        $this->setPage(max($this->page - 1, 1));
     }
 
     public function nextPage()
--- a/src/views/bootstrap.blade.php
+++ b/src/views/bootstrap.blade.php
@@ -9,7 +9,7 @@
                     </li>
                 @else
                     <li class="page-item">
-                        <button type="button" dusk="previousPage" class="page-link" wire:click="previousPage" rel="prev" aria-label="@lang('pagination.previous')">&lsaquo;</button>
+                        <button type="button" dusk="previousPage" class="page-link" wire:click="previousPage" wire:loading.attr="disabled" rel="prev" aria-label="@lang('pagination.previous')">&lsaquo;</button>
                     </li>
                 @endif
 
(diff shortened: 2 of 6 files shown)


calebporzio:
Beautiful! Thanks Austen!

PR #1897, opened 2020-10-24
[Maintenance] use str all over the place

nuernbergerA:
- replaced `Str` with `str()`
- to make sure a string is returned some methods now have a return type `string` (could be removed and return needs to cast stuff to `string`)
- kept `Str` only for `Stringable` tests to be explicit
- in some tests we now use `assertStringContainsString` and `assertStringNotContainsString`
- [separated commit](https://github.com/livewire/livewire/pull/1897/commits/2e94090ad360e281b480d62fc08a6113bcb6810e#diff-bc1a1c611931be05f4fd946bdc7dff27c6a67685f272b6e1369017a259d2b950) where I replaced a strange between hack with the actual `between()`

### DIFF
--- a/src/Commands/ComponentParser.php
+++ b/src/Commands/ComponentParser.php
@@ -5,6 +5,7 @@ namespace Livewire\Commands;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Illuminate\Support\Facades\File;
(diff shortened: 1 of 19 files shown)


calebporzio:
Haha, thanks for this @nuernbergerA - I dig!

Issue #1868, opened 2020-10-21
Cannot bind to hidden model attributes


jasrys:
### Description

**wire:model does not bind to hidden model attributes**

My use case is that I need to allow users to view/update hidden attributes via a livewire powered form. e.g. an internal note attribute for a laravel model should be a hidden model attribute so that it doesn't leak externally (e.g. via being cast to JSON in an external API call) but may need to be viewed or updated internally by users from time to time.

Is this possible to accomplish with livewire?

### Exact steps to reproduce

1. Install new laravel 8 project and add livewire 2.3
2. Create laravel model with hidden attributes
3. Create livewire component which has a wire:model reference to the model's hidden attribute. e.g. `<input wire:model="model.hidden_attribute">`
4. Observe that hidden attributes are not bound and that updating hidden attributes causes the input to be removed (perhaps because the data object in the response to the `syncInput` request contains null for the hidden attributes)

### Stripped-down, copy-pastable code snippets

Model with hidden attribute:
```
<?php

namespace App\Models;

use Illuminate\Database\Eloquent\Model;

class Post extends Model
{
    protected $fillable = [
        'title',
        'body',
        'note,'
    ];

    protected $hidden = [
        'note',
    ];
}
```
Example component:
```
<?php

namespace App\Http\Livewire;

use App\Models\Post;
use Livewire\Component;

class PostComponent extends Component
{
    public Post $post;

    protected $rules = [
        'post.title' => 'required',
        'post.body' => 'required',
        'post.note' => 'nullable',
    ];

    public function mount()
    {
        $this->post = Post::first();
    }


    public function render()
    {
        return view('livewire.post-component');
    }
}
```
 
Example component view:
```
<div>
    <label for="post.title">Title</label>
    <input
        wire:model="post.title"
        name="post.title"
        type="text"
    >

    <label for="post.body">Body</label>
    <input
        wire:model="post.body"
        name="post.body"
        type="text"
    >

    <label for="post.note">Note</label>
    <input
        wire:model="post.note"
        name="post.note"
        type="text"
    >
</div>

```

The hidden attribute is not bound on initial render:

<img width="609" alt="image" src="https://user-images.githubusercontent.com/11322354/96772442-611e4180-13a0-11eb-8b6e-ca75773bfed3.png">

The hidden attribute cannot be set/sync'd by the input:
[...]

PR #1899, opened 2020-10-24
Support binding to hidden model attributes when explicitly present in Component $rules.

iksaku:
It was rised on #1868.

Straight to implementation :nerd_face: 

Yup, all looking good :green_circle: 

As discussed in #1868, sometimes you may want to hide certain attribute values from a Model (for implementation reasons), but you may stumble across feature implementations that require that you bind into those attributes using Livewire (or maybe entangled to Alpine).

Well, this PR tries to solve that.

In order to prevent all hidden attributes to be exposed via Livewire, only the attributes specified in the `$rules` property (or `rules()` method) will be exposed to Livewire. This way, we try to force developers to think about their implementation needs, unless they're just copy-pasting without understanding anything.

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -164,9 +164,9 @@ class HydratePublicProperties implements HydrationMiddleware
         if ($rules = $instance->rulesForModel($property)) {
             $keys = $rules->keys()->map(function ($key) use ($instance) {
                 return $instance->beforeFirstDot($instance->afterFirstDot($key));
-            });
+            })->toArray();
 
-            $fullModelData = $instance->$property->toArray();
+            $fullModelData = $instance->$property->makeVisible($keys)->toArray();
 
             foreach ($keys as $key) {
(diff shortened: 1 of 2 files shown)


jessarcher:
Hey @calebporzio, did you happen to see my related scenario at https://github.com/livewire/livewire/issues/1868#issuecomment-714438553?

It seems we can currently write to bound hidden attributes, but their values will always be returned as null on read. The same is true for a custom accessor/mutator pair, which aren't explicitly hidden, but just haven't been appended. The mutator will work, but not the accessor.

This can cause bugs that are hard to track down because the failure is silent and things can appear to be working. This can also leave the data in a bad state because of the mismatch between what is stored vs. what can be seen.

I think the reason this is so confusing is that binding is a single action to a developer, rather than separate read and write action like a typical server-rendered app or JSON API. Therefore, I think either both should work or neither should work. And if neither should work, then an exception would be safer than failing silently. It also lets the developer know that it's not a bug or missing feature, and that there is something they can do to get their current solution to work.

Thanks for reading!

danharrin:
Hey @jessarcher! This is still an issue, especially in Livewire's testing environment. Did you ever find a good solution?

PR #1900, opened 2020-10-24
support numbers larger than javascript maximum safe integer - #1857

austenc:
As reported in #1857, using integers larger than Javascript's `Number.MAX_SAFE_INTEGER` causes problems. 

This PR should solve that problem by casting large integers to a string when dehydrating (nice suggestion @nuernbergerA !) 

For more info on Javascript's limitation see here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

### DIFF
--- a/src/HydrationMiddleware/NormalizeComponentPropertiesForJavaScript.php
+++ b/src/HydrationMiddleware/NormalizeComponentPropertiesForJavaScript.php
@@ -3,11 +3,10 @@
 namespace Livewire\HydrationMiddleware;
 
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
+use Illuminate\Support\Collection;
 
 class NormalizeComponentPropertiesForJavaScript extends NormalizeDataForJavaScript implements HydrationMiddleware
 {
-    protected const JAVASCRIPT_MAX_SAFE_INTEGER = 9007199254740991;
-
     public static function hydrate($instance, $request)
     {
         //
@@ -16,10 +15,7 @@ class NormalizeComponentPropertiesForJavaScript extends NormalizeDataForJavaScri
     public static function dehydrate($instance, $response)
     {
         foreach ($instance->getPublicPropertiesDefinedBySubClass() as $key => $value) {
-
-            // The javascript maximum integer is 9007199254740991. 
-            // If the value is larger than that, it should be converted to a string
-            // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
+            // If the value is larger than the javascript integer maximum, it should be converted to a string
             if (is_numeric($value) && $value > self::JAVASCRIPT_MAX_SAFE_INTEGER) {
                 $instance->$key = (string) $value;
             }
--- a/src/HydrationMiddleware/NormalizeComponentPropertiesForJavaScript.php
+++ b/src/HydrationMiddleware/NormalizeComponentPropertiesForJavaScript.php
@@ -6,6 +6,8 @@ use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 
 class NormalizeComponentPropertiesForJavaScript extends NormalizeDataForJavaScript implements HydrationMiddleware
 {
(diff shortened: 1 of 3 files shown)


adamthehutt:
@austenc I added a comment on #1857 as well, but I'm still getting a CorruptComponentPayloadException every time there's a data round-trip. Looking at the wire:initial-data in the rendered HTML, the encoded JSON includes the big integer as a number, not a string.

nuernbergerA:
@austenc can you add a dusk test to make sure a whole roundtrip would work?

austenc:
Thanks for testing @adamthehutt -- any chance you have a simple repo already set up to save me some time when reproducing this? 

I think we just need to ensure it's cast to a string in the initial data too. Will check.

adamthehutt:
@austenc Sure, I created a simple repo demonstrating the issue here: https://github.com/adamthehutt/livewire-bigints

If you visit the home page of that laravel app it loads a Livewire component that demonstrates the issue with an integer property, a model, and a collection of models.

FWIW, another way to approach this would simply be to ensure that jsonSerialize() rather than toArray() is called on models and collections before json encoding them. (Actually, you could accomplish the same thing by just calling json_encode() on the model or collection directly, without first calling toArray().) That way you'd leave it up to the user to handle this and any similar issues that might arise.

adamthehutt:
@austenc I'm sure this isn't the cleanest way to accomplish this, but I don't understand the Livewire internals well enough to know what would be optimal. That said, I *think* I was able to get it working with the following two changes:

```php
class Response
{
...
    public function embedThyselfInHtml()
    {
        if (! $html = $this->effects['html'] ?? null) return;

        $initialData = $this->toArrayWithoutHtml();
        array_walk_recursive($initialData, function (&$value) {
            if (is_int($value) && $value > 2**53) {
                settype($value, 'string');
            }
        });

        $this->effects['html'] = (new AddAttributesToRootTagOfHtml)($html, [
            'initial-data' => $initialData,
        ]);
    }
}
```

and

```php
class SecureHydrationWithChecksum
{
...

    public static function dehydrate($instance, $response)
    {
        $memo = $response->memo;
        array_walk_recursive($memo, function (&$value) {
            if (is_int($value) && $value > 2**53) {
                settype($value, 'string');
            }
        });
        
        $response->memo['checksum'] = (new ComponentChecksumManager)->generate($response->fingerprint, $memo);
    }
}
```

These two changes together ensure that the initialization data _and_ the checksum hash are both based on recursive type-casting of too-large integers to strings. 

I can't say for sure that this handles every scenario, but in my very simple demo repo it appears to prevent errors.

adamthehutt:
@austenc Update to my last comment. This doesn't, in fact, fully fix the problem (although it seems to be getting close). If models (including in the collection) have big integer IDs, then the checksum still fails validation on subsequent (i.e. not-initial) hydration.

austenc:
@adamthehutt I have been working on this and unfortunately don't have a solution yet. 

I've been able to fix it for properties of the following data types: `integers, arrays, model properties`
I have NOT been able to fix it for collections of models yet. This is where it becomes tricky. For instance, we don't want to have to parse through every attribute for every model inside a collection just to see if it's greater than the javascript maximum. 

It does seem like this is something Livewire should handle automatically, I'm just not sure how to do it (yet). I'm leaving a note here for other folks to weigh in (and to document for myself).

The only way I've been able to get it working is using the (updated) code in my branch combined with the following on the model:
```
protected $casts = [
    'big_integer' => 'string'
];
```

Thank you for your demo repo by the way, that was a huge help!

adamthehutt:
Thanks @austenc. I appreciate your time and effort on this. You're right that collections are the toughest aspect.

I'm afraid I didn't see your last comment until just now, and I put together a pull request of my own (#1938). It does seem to be working for me, including with collections. It may be "too recursive" for your liking, however. That said, if your concern is performance related, perhaps we could run some benchmarks? I'd be really surprised if doing these checks and typecasting takes more than a few microseconds per request.

calebporzio:
Closing for same reasons as #1938

PR #1935, opened 2020-10-30
Add wire:loading display modifiers

calebporzio:
Currently, `wire:loading` will force the element to use `inline-block` when showing.

This PR adds an API to tell Livewire which display property to use when showing elements:

```html
<div wire:loading.flex>...</div>
<div wire:loading.inline>...</div>
<div wire:loading.block>...</div>
<div wire:loading.grid>...</div>
<div wire:loading.table>...</div>
```


Francismori7:
Thanks a lot.

PR #1936, opened 2020-10-30
fixed escaping in heredoc string

austenc:
Fixed escaping in the heredoc string so it doesn't cause a javascript syntax error.


calebporzio:
Thanks Austen!

PR #1938, opened 2020-10-31
String encode large integers for javascript safety

adamthehutt:
This is in reference to #1900 and #1857. The goal is to safely encode large integers (that exceed Javascript's maximum) as strings whenever Livewire generates JSON. 

These changes appear to solve the problem for me. However, I don't understand Livewire's internals well enough to feel confident, so this warrants a careful review from someone who has a better understanding than I. Unfortunately, I'm not able to write any useful tests since I can't seem to get Livewire's Dusk tests to run locally, and that's the only good way to test this. (Sorry, I tried, really.) Unit tests are passing, at least.

### DIFF
--- a/src/ComponentChecksumManager.php
+++ b/src/ComponentChecksumManager.php
@@ -2,10 +2,16 @@
 
 namespace Livewire;
 
+use Livewire\Concerns\EncodesJsonSafely;
+
 class ComponentChecksumManager
 {
+    use EncodesJsonSafely;
+
     public function generate($fingerprint, $memo)
     {
+        self::stringEncodeTooLargeIntegers($memo);
+
         $hashKey = app('encrypter')->getKey();
 
         // It's actually Ok if the "children" tracking is tampered with.
--- a/src/Concerns/EncodesJsonSafely.php
+++ b/src/Concerns/EncodesJsonSafely.php
@@ -1,0 +1,20 @@
+<?php
+declare(strict_types=1);
+
+namespace Livewire\Concerns;
+
+trait EncodesJsonSafely
+{
+    protected static $javascriptMaxInteger = 2**53-1;
+
+    protected static function stringEncodeTooLargeIntegers(&$mixed)
+    {
+        if (is_numeric($mixed) && $mixed > self::$javascriptMaxInteger) {
+            $mixed = (string) $mixed;
+        } elseif (is_array($mixed)) {
+            array_walk_recursive($mixed, function (&$value, $key) {
+                self::stringEncodeTooLargeIntegers($value);
+            });
+        }
+    }
+}
--- a/src/HydrationMiddleware/AddAttributesToRootTagOfHtml.php
+++ b/src/HydrationMiddleware/AddAttributesToRootTagOfHtml.php
@@ -2,10 +2,13 @@
 
 namespace Livewire\HydrationMiddleware;
 
+use Livewire\Concerns\EncodesJsonSafely;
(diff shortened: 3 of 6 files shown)


calebporzio:
Hi @adamthehutt  - thanks for your willingness to contribute.

3 thoughts:
- This logic should probably live in: `src/HydrationMiddleware/HydratePublicProperties.php`
- How expensive is this check? It will be run on EVERY peice of data for Livewire which makes me curious what kind of overhead this will add.
- I wonder if we should hold off on this in favor of more bespoke approaches outside of Livewire's core. (Also I think you're the only person who has brought this up)

I'm going to close this one for now. Thanks though!

adamthehutt:
@calebporzio 

HydratePublicProperties middleware was the first place I looked. Spent a long time trying to get it to work from there, but I couldn't figure it out. The encoding seems to happen in lots of places scattered throughout the codebase, which makes things really difficult. (You know the code base 1000x better than I do, so you might disagree and realize something I don't, obviously...)

As far as bespoke approaches outside of Livewire's core... How would that work? If you can give me even a tiny hint about how to go about that I'll give it a shot. Are you suggesting to add additional middleware that rewrites the request and response data, for example? Casting big ints to strings in the Laravel models doesn't work, btw, since Livewire undoes that and encodes them as integers. 

I'm also happy to do some benchmarking to address your performance concerns.

Otherwise, I have to say this is understandable but pretty disappointing. I may be the only one who has encountered this to date, but I promise I won't be the last. Anyone dealing with really big data sets (even if that's as simple as consuming the Facebook API) will run into this, and Livewire becomes borderline unusable. I love the Livewire concept enough that I guess I could maintain my own fork, but I just don't know how sustainable that would be.

seabasss:
@calebporzio @adamthehutt 

I think this topic should be opened up again. A lot of us are using snowflake ID's with long integers and we all get this message when using livewire:

```
Livewire encountered corrupt data when trying to hydrate the [laravel.jetstream.http.livewire.update-profile-information-form] component. Ensure that the [name, id, data] of the Livewire component wasn't tampered with between requests.
```

If you want a quick example, try to set `current_team_id` on the user model to a high integer like `23535702465712128` in Jetstream. If you try to update your user profile you get the error above and by using Josh's brilliant checksum logger you'll see that the returned integer now ends with 30 instead of 28 because of the javascript integer limit.

In a large app this is really a pain and it would be great with a built-in solution in Livewire to handle it. Would it even be harmful to just convert large integers to strings as default since javascript can't handle them anyway?

Right now I'm casting all id fields to strings on the models as a work-around. I should probably cast it in the controllers instead, but that's even more work.

Ideas are highly appreciated. Thank you! :)

jtomek:
I agree. I have just encountered the same issue. A good start would be to have at least a better error handling for people who will encounter this problem in the future as it really hard to understand what's happening (over 1 day of debugging in my case) 

@seabasss how did you make the casting work? I was unsuccessful setting it both for 'id' and 'foreign_id's. I used the typical $casts array on a model. For IDs, I have also set $increamenting to false. To no avail.

seabasss:
> I agree. I have just encountered the same issue. A good start would be to have at least a better error handling for people who will encounter this problem in the future as it really hard to understand what's happening (over 1 day of debugging in my case)
> 
> @seabasss how did you make the casting work? I was unsuccessful setting it both for 'id' and 'foreign_id's. I used the typical $casts array on a model. For IDs, I have also set $increamenting to false. To no avail.

Just cast array on the model. All ids and foreign ids as strings on all models with big int ids.

Issue #1904, opened 2020-10-25
Image/file upload not updating temporary image / model binding


spaantje:
### Description

When i have a file upload field with a temporary image viewing and a delete button to clear out the field and upload the file, delete it again and upload the same file again it does not update. When i choose a different file it's working. I don't know if this is a file input field limitation, something i'm doing wrong or a real bug.

### Exact steps to reproduce

I have created a clean L8 application with only Livewire to test it out: 
https://gitlab.com/spaanproductions/livewire-upload-bug-test

Or a gist with only the Livewire Parts:
https://gist.github.com/spaantje/1c729ac972e33646e78e3beff3ddc80e

Steps: 
1. Go to the page. 
2. Upload an image. 
3. Click "Delete Photo"
4. Upload the same image. (You see it does not fill the field/updates the temp image).
5. Upload a different image (it works).

### Stripped-down, copy-pastable code snippets

**Context**
- Livewire version: v2.3.0
- Laravel version: v8.11.2
- Browser: Chrome

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

spaantje:
REOPEN

alibori:
Hi!

I've a very similar problem. I can upload an image the first time, delete it, but when I try to upload a new one the file input seems to keep a temporary file and, of course, the validation detects that is not an image.

Any idea?

Thanks!

calebporzio:
Thanks for the PR @joshhanley, this has been merged and fixed. Will be available in the next release.

Thanks!

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

[...]

PR #1939, opened 2020-10-31
Add on click listener to clear current file

joshhanley:
See issue #1904 

No

No, tried to setup a dusk test, but couldn't get it working, it was triggering the upload fine each time.

Basically, if a user uploads a photo, deletes that photo, and tries to use the same file input field to upload the same photo, nothing happens.

This happens because the onChange event handler isn't fired again if the same file is selected.
To combat this, you need to clear the file input when the file input dialog is opened.
See this https://stackoverflow.com/questions/12030686/html-input-file-selection-event-not-firing-upon-selecting-the-same-file


calebporzio:
Thanks so much @joshhanley for this.

spaantje:
Thanks!

joshhanley:
No worries!

Issue #1618, opened 2020-09-16
Mocking request query parameters for Livewire


faustbrian:
### Description

Since Livewire 2.0 now has the ability to automatically fill properties from the query string, what would be the recommended way to test that Livewire picks up properties from the query string? Is there any easy way to set the query string before the Livewire component is mounted in a test so that it could pick up the query parameters?

**Context**
- Livewire version: 2.2.3
- Laravel version: 8.0
- Browser: n/a

faustbrian:
@calebporzio any chance to get some help on this?

austenc:
@faustbrian If you want to test query string stuff, there are examples of how to do it in Livewire's tests. 

See here:
https://github.com/austenc/livewire/blob/b92bef82525268ffbc6f761f4bab285fa3dd7105/tests/Browser/QueryString/Test.php#L11

Since that's tested by Livewire, testing it in your own code may be redundant, but that's your call. 👍

faustbrian:
Yeah I know how to do it with Dusk from the test suite but doing it in unit tests would be my use-case.

nuernbergerA:
Could you provide some sudo code?

So we know how you would imagine an API for this should look like.

I don't think that's possible without dusk but you never know 🙂

calebporzio:
Good call @faustbrian, I think it would be handy to have a way to do this in Livewire's testing utilities.

In the past, I've used more outside-in approaches to assert that Livewire handled a query string setter.

Here's some not-tested-off-the-top-of-my-head code:
```php
$this->get('/the-url/with/the/livewire/component?foo=bar')
    ->assertSee('bar'); // Assuming that `$foo` is echoed out on the page
```

Other than this I'm not sure what the API would be for the testing utilities.

Maybe something like this?
```php
Livewire::withQueryParams(['foo' => 'bar'])->test('livewire-component');
```

Because this isn't a bug necessarily, I'm going to close this issue. Open to a PR thought.

faustbrian:
Bit busy at the moment but I'll try to take a stab at a PR when I find time.

PR #1940, opened 2020-10-31
Allow query parameters to be passed into TestableLivewire

mahansky:
#1618 

No

Yes

You can use `withQueryParams` method to specify query parameters to be used when testing your Livewire components, for example to access `request('name')` in your `mount` method
```php
Livewire::withQueryParams(['name' => 'Livewire'])->test('livewire-component');
```


calebporzio:
Thanks!

PR #1941, opened 2020-10-31
Ability to fill component arrays or model properties

iksaku:
Yeah, this was discussed in #1776 and a pending PR #1872, but the PR failed to add tests.

Nope, straight implementation.

Now, it does.

As discussed in #1776, when using binded models, you can expect the `$this->fill()` method in a component to 
assign values to the model property using dot notation (`'user.name' => $user->name`).

This PR will enable that, as well as to provide support for filling array properties.


calebporzio:
This is perfect, thanks!

PR #1943, opened 2020-11-01
Respect existing history state

robertvansteen:
There was no issue or discussion about this yet.

Nope.

Yes, I have added a browser test that tests this added behaviour.

Livewire adds an object to `history.state` when calling `pushState` or `replaceState`. Since `history.state` is a global object other libraries or user code can write to that as well. Right now Livewire will overwrite any other values in `history.state` when writing to it. This PR changes that so Livewire will respect any values in `history.state` if we can copy them over.

To give some context on why a user might need to use `history.state`: Let's say you have a long list in an overflowing div with links to detail pages, if the user opens a detail page and clicks the back button in the browser, they expect to be taken back to the same scroll position in the list. This works by default for the `body` element, but not for  any other element where you can scroll in. To implement something like that with JS we need to keep track off the scroll position for that element for every entry in the browser history and `history.state` is the only place that allows us to do so.

Thank you for creating Livewire!


calebporzio:
Thanks for this @rovansteen!

Heads up for next time: check out a new branch on your fork for your feature (not `master`) - this way I can push to this branch before merging. No big deal though.

Thanks again!

Issue #1774, opened 2020-10-07
BUG: Livewire + Alpine entangle.defer


deanzod:
If you use entangle and defer in Alpine it causes an issue where every second submit is ignored by livewire. When doing this with a more complex form using arrays it causes the "Livewire encountered corrupt data when trying to hydrate the ['x'] component. Ensure that the [name, id, data] of the Livewire component wasn't tampered with between requests." error

See video for example with requests in dev tools:
https://www.youtube.com/watch?v=uJhY9Wv5KDw

Steps to reproduce:

Create component:

```php
class Test extends Component
{
    public $testing;
    public function submit()
    {
    }

    public function render()
    {
        return view('livewire.test');
    }
}
```

View:

```html
<div>
    <div x-data="{ testing:  @entangle('testing').defer }">
        <input type="text" x-model="testing">
        <p>Alpine: <span x-text="testing"></span></p>
        <p>Livewire: {{$testing}}</p>
        <button wire:click.prevent="submit">Submit</button>
    </div>
</div>
```

Then type a letter, click submit, type another click submit etc...

Removing defer works perfectly but sends a request with every keystroke

**Context**
- Livewire version: 2.2.9
- Laravel version: 8.8.0
- Browser: Chrome

deanzod:
To add to this, after a bit more testing it seems the bug is only present if I change a single character.

**So if I:**

type: 'a' -> submit -> livewire result =  'a' (correct)
append another 'a' -> submit -> livewire result = 'a' (incorrect)

**But if I:**

type: 'a' -> submit -> livewire result = 'a' (correct)
append another 'ab' -> submit -> livewire result = 'aab' (correct)

All of this behaviour is very consistent. It has become a major issue for me now as it happens every time after failing validation and it just doesn't seem feasible to exclude `defer` and allow a request with every key stroke on a large form

zepfietje:
Hi @deanzod, could you include the validation part of your code? I believe I've been running into this issue too.

deanzod:
@zepfietje  Ok, here is a video of the other issue which I believe is related to the core issue:
https://youtu.be/xvIN6kc8cYw

Component:

```php
class Test extends Component
{

    public $testing = ['', ''];
    public function submit()
    {
        $this->validate(['testing.1' => 'required']);
    }

    public function render()
    {
        return view('livewire.test');
    }
}
```

View

```html
<div>
    <div x-data="{ testing:  @entangle('testing').defer }">

[...]

PR #1947, opened 2020-11-01
Add failing test for entangle defer

joshhanley:

Yes, see latest comments on #1774 

No

Yes, a failing one 😬

Entangle defer seems to still be having some issues with subsequent Livewire requests.

With a defered input field, and a submit button.
The first input change is batched, and then sent with submit button click.
But a follow up input change is not batched, so when submit button is clicked again, the changes are not sent with the request.

Hope this helps!


joshhanley:
So I've debugged what could be causing this, and it seems to be this call on line 78 in AlpineSupport.js

```
if (blockAlpineWatcher === true) {
    blockAlpineWatcher = false
    return
}
```

which is getting triggered by this on line 105

```
livewireComponent.watch(
    livewireProperty,
    value => {
        blockAlpineWatcher = true
        component.$data[key] = value
    }
)
```

Hope this helps!

calebporzio:
Ok I got to the bottom if it. Thanks SO much for such a helpful PR @joshhanley - SO NICE 🤩

joshhanley:
Great work! No worries!

PR #1959, opened 2020-11-03
use interface Arrayable instead of Model

whizzrd:
InteractsWithPoperties::fill() explicitly checks that the object is a Model before calling toArray. Thus using fill with objects that do not extend Model but do implement the toArray method will not be converted to array and will not fill.

Allow using fill with objects that are not models but do implement \Illuminate\Contracts\Support\Arrayable

resolves #1958


calebporzio:
Thanks!

Please review this PR: #1941 

If that doesn't resolve your issue, then please re-open this PR with a better statement of what problem this solves and tests to support the new feature.

Thank you!

PR #1973, opened 2020-11-05
Support layout specified by route

jszobody:
In Livewire 1 I was specifying a custom layout file in my route group, something like this:

```php
Route::group(['prefix' => 'admin', 'layout' => 'layouts.admin']) {
    Route::livewire('/', 'admin.dashboard');

    // Dozens more!
});
```

With Livewire 2, we lost the ability to specify a custom layout file in the routes file, now it has to be specified inside each individual component. For a situation where there are dozens of components, all that need the same custom layout file, this can be tedious.

This PR re-introduces support for the route-specified layout. It still will use the component specified layout first if it exists, but then fall back to the route layout before falling all the way back to the hardcoded default.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -51,7 +51,7 @@ abstract class Component
         $layoutType = $this->initialLayoutConfiguration['type'] ?? 'component';
 
         return app('view')->file(__DIR__."/Macros/livewire-view-{$layoutType}.blade.php", [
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the PR!

I'd rather not return to the route-level configuration, but I agree there should be a story for this problem.

I've written up a feature spec that we'll leave open.

Feel free to PR if you like. Thanks!

https://github.com/livewire/livewire/issues/2000

PR #1976, opened 2020-11-05
Use Arrayable Interface for casting array types in the component fill method.

iksaku:
Suggested in #1958

Nope

Yup, perfectly balanced... :green_circle: 

When using the `fill` method in a component, only instances of `Model` are converted to Array representation
to fill public properties in the component, however, there may cases when you need to use other type of
objects that are indeed castable as Arrays, but since we're looking only for `Model` instances, properties are
not being properly segregated.

This PR fixes that.

### DIFF
--- a/src/ComponentConcerns/InteractsWithProperties.php
+++ b/src/ComponentConcerns/InteractsWithProperties.php
@@ -2,7 +2,7 @@
 
 namespace Livewire\ComponentConcerns;
 
-use Illuminate\Database\Eloquent\Model;
+use Illuminate\Contracts\Support\Arrayable;
 use function Livewire\str;
 
 trait InteractsWithProperties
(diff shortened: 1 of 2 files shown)


iksaku:
Looks like there was a previous PR similar to this (#1959), but I think these provided tests better illustrate the possible use case when not dealing with Models nor Collections

calebporzio:
Thanks for submitting.

Because `fill` is already a convenience helper, I'm hesitant to continue expanding it's capabilities until either:
- More people express this as a pain point
- You present a concrete identifiable and common use case that this addresses

Closing this for now. Thanks for contributing.

whizzrd:
Hey @calebporzio , since you also closed #1959 which fixed the same where the use-case is described in #1958 .. how many people does it take?

PR #1987, opened 2020-11-07
call hydrate once

whizzrd:
See #1986 
No
Yes, LifecycleHooksTest now tests that the methods are called the expected amount of times 
Hydrating an already hydrated object makes it too wet.
You're welcome


whizzrd:
Fixes #1986

calebporzio:
Thanks for submitting.

Unfortunately because the problem scenario isn't clear to me and there are no tests present, I'm going to close this. Thanks!

Issue #763, opened 2020-03-24
x-on:click called twice after livewire call under certain circumstances


jdssem:
**Describe the bug**
`x-on:click` event called twice after livewire communication under certain circumstances.

**To Reproduce**
- laravel v7.7.2
- livewire v1.0.9
- alpinejs v2.2.1

### Test.php
```php
class Test extends Component
{
    public $name;

    public $items;

    public function mount()
    {
        $this->items = [1, 2];
    }

    public function render()
    {
        return view('livewire.test');
    }
}
```

### test.blade.php
```html
<div>
    <input type="text" wire:model="name"/>

    @foreach($items as $item)
        <div x-data="{ open:false }" wire:key="{{ $item }}" x-bind:key="{{ $item }}">
            <div> <!-- If I remove this div, it works fine. -->
                <button x-on:click="console.log('clicked');">
                    click
                </button>
            </div>
        </div>
    @endforeach
</div>
```

1. click the button and see the browser console. it works nomally.
2. put some characters in the input box. It will make a livewire call.
3. click the button again and see the browser console. it will be called twice.

**Expected behavior**
It should be called once.

**Screenshots**
![livewire](https://user-images.githubusercontent.com/978944/77432415-607cf700-6e21-11ea-83af-508ae27b7248.gif)

**Desktop (please complete the following information):**
 - chrome, safari, firefox

**Additional context**
I have originally made a replies component. It has several buttons like edit and delete in every replies. When I click delete button, it shows a delete modal. If I click 'delete' button, it calls livewire action using javascript. But If I make some livewire actions such as deleting or editing, delete action runs twice. The code above is the simplified version of that.

vrajroham:
@jdssem Encountered same issue. On clicking the submit button, only one xhr request is sent to a server to the specified function `E.g. wire:submit.prevent="submit"` but the code in submit function runs twice. 

In my case, 2 records are created with same input and 2 different otps are being sent.

vrajroham:
@jdssem @calebporzio @lancepioch 

**Update** : After removing id from the form attribute all works perfectly no more running the twice. @jdssem Check if this solves yours too.

jordyvanderhaegen:
Ran into the same issue, it looks like 2 click events are present on the DOM element after the component is updated.

nuernbergerA:
Hey @jdssem I follow your steps and all works fine, even with your alpine and livewire versions.

Have I done something wrong?

Cheers Adrian

[...]

PR #1995, opened 2020-11-08
Add failing test Alpine clicks firing twice after Livewire change

joshhanley:

Yes - see issue #763 

No

Yes

This is a failing tests to demonstrate that Alpine click handlers are firing twice after Livewire changes the content on the page.

It seems like any x-on:click that is attached to an element inside blade conditional or loop is getting registered twice after a Livewire request is sent that may change/modify that element (e.g. show).


austenc:
Continued in #2434, thanks @joshhanley!

PR #1996, opened 2020-11-09
Only initialize child components when necessary - closes #1726

### LINKED ISSUE #1726: Cannot read property 'fingerprint' of null
@amaelftah  I think I've figured it out. 

I noticed when the result is returning properly, this happened

Screenshot 2020-09-25 at 2 03 36 AM

and when the error appear, this is what I see
Screenshot 2020-09-25 at 2 03 46 AM

So I'm guessing the nested component for some reason is being generated twice.
It might be because it has the same id for key.

So I tried this hacky method to always generate a different key 

```html
@php
    $randomKey = time().$watchedFile->id;
@endphp
<livewire:files-table-actions :watched-file="$watchedFile" :key="$randomKey"/>
```

and I'm no longer facing the issue.

I'm not sure why is the previous nested component not clearing itself. 

Perhaps @calebporzio could point us to what are we doing wrongly :)

_Originally posted by @jasontxf in https://github.com/livewire/livewire/issues/1686#issuecomment-698503788_

Hello @calebporzio I'm sorry to reopen this, but the original issue wasn't resolved.
Is the intended behavior to always generate a different key (by appending time()) when using nested component?

Thank you for your time!

austenc:
This fixes a case where `initial-data` could be undefined on nested components as reported in #1726.

After an initial filter with an input, if you then clear the input, Livewire was _always_ calling the JS constructor when adding elements back to the DOM, even if the element was a component that had already been initialized. Since the `initial-data` is removed when the constructor is called, it was causing the fingerprint error reported in the issue.

The error message isn't as detailed as I'd prefer from the failing test, but it does fail without the fix in place. 👍 

**Here's a GIF of the original issue**
livewire-bug

### DIFF
--- a/js/component/index.js
+++ b/js/component/index.js
@@ -475,11 +475,14 @@ export default class Component {
                         return false
                     }
                 } else if (DOM.isComponentRootEl(node)) {
-                    store.addComponent(new Component(node, this.connection))
+                    // Only initialize "new" components if they don't exist on the page already
+                    if (node.hasAttribute('wire:initial-data')) {
+                        store.addComponent(new Component(node, this.connection))
 
-                    // We don't need to initialize children, the
-                    // new Component constructor will do that for us.
-                    node.skipAddingChildren = true
+                        // We don't need to initialize children, the
+                        // new Component constructor will do that for us.
+                        node.skipAddingChildren = true
+                    }
                 }
 
                 this.morphChanges.added.push(node)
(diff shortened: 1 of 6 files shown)


austenc:
The tests should pass when assets are built. I didn't include new assets in `dist/` intentionally. Let me know if you'd like me to do that! Thought that they'd be rebuilt during a release cycle anyhow.

evanbarter:
Happily confirming this fixes my reproduction for #1726 

/sparkle @austenc For taking care of this!

calebporzio:
This problem arrises because of the lack of a `wire:key` within the @foreach loop: `@foreach <div wire:key="???"`

PR #2027, opened 2020-11-13
Add uniqid to the paginator links wire:key

### LINKED ISSUE #1582: Pagination display error when call links() 2 times
image

Related #1536

likeadeckofcards:
Fixes #1582 pagination issue when rendering 2 or more pagination links for the same livewire component

### DIFF
--- a/src/views/pagination/tailwind.blade.php
+++ b/src/views/pagination/tailwind.blade.php
@@ -1,4 +1,4 @@
-<div>
+<div @if($id ?? false) id="{{ $id }}" @endif>
     @if ($paginator->hasPages())
         <nav role="navigation" aria-label="Pagination Navigation" class="flex items-center justify-between">
             <div class="flex justify-between flex-1 sm:hidden">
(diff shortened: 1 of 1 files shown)


calebporzio:
Hey there, thanks for the PR, does this require anything on the usage part? In other words, where does that `$id` come from? Is that something that would be documented?

Please provide a more thorough explanation of the problem and solution and ideally a test. Thanks!

likeadeckofcards:
`$id` does not have to be set but if a user does set it in `$models->links(null, ['id' => 'my-unique-id'])` it will be set otherwise it generates a unique id using `uniqid('paginator')`. Typically this would result in `paginator-1`, `paginator-2`, etc.

calebporzio:
Ah, I see.

I like this PR. Would you mind adding a test for the case that this fixes? Also, should we add this to the other non-tailwind pagination views?

Am open to merging if the PR is more fleshed out.

Thanks!

calebporzio:
Closing. Please re-submit if you address my comment above. Thanks!

jdlien:
This solved the problem I was having. I think simply changing wire:key to
<span wire:key="{{ uniqid('paginator') }}-page{{ $page }}">
is sufficient to allow the inclusion of the pageselector multiple times. This seems like it'd be a common use case as long pages might want a page selector at the top and bottom of a page.

If I get less busy/lazy I may submit a proper PR with tests and modifications for the other templates, but at least this works for me using my own custom blade template.

Issue #2040, opened 2020-11-17
wire:modeling separate single file uploads to an array property causes error


amirzpr:
### Description
I have 4 different single file upload inputs in a page and connected them to integer keys of an _array_ property in the livewire component. Because I want to use the keys of this array to determine order of images in `sort_order` column of database table, the keys are from 1 to 4.
```php
class ProductComponent extends Component
{
    public $images;

    public function mount($product)
    {
        $this->images = array_fill_keys(range(1, 4), null);
    }
}
```

I'm using `null` as default array values to identify absence or deletion of an image.

```blade
@foreach( $images as $key => $image )
    <input wire:model="{!!"images.$key"!!}" type="file">
@endforeach
```

Now the problem is that when I upload the _first_ image ( `images.1` ), it causes this error:
```
ERROR: Call to a member function getRealPath() on null
```

Uploading other images except the first one won't raise this error.

### Why this occurs?
I made a dive into the framework's code and realized that line 48 of `Livewire\RenameMe\SupportFileUploads` is causing this issue:
```php
if (is_array($value) && isset(array_values($value)[0]) && array_values($value)[0] instanceof TemporaryUploadedFile && is_numeric(key($value))) {
    return array_values($value)[0]::serializeMultipleForLivewireResponse($value);
}
```
Actually, this condition passes and the `serializeMultipleForLivewireResponse` is called, although **it's not a real multiple file upload array** and this misinterpretation causes the error.

### Exact steps to reproduce
Just do whatever I explained above.

#### Context
- Livewire version: 2.3.1
- Laravel version: 8.11.2
- Browser: Chrome

I will make a PR to resolve this.

PR #2041, opened 2020-11-17
wire:model separate file uploads to array property bug

amirzpr:
Fixed wire:model separate file uploads to array property bug explained in this issue: #2040

Unwrapping `array_values($value)` fixed the issue in my case. I didn't see any reasons why they're wrapped.

### DIFF
--- a/src/RenameMe/SupportFileUploads.php
+++ b/src/RenameMe/SupportFileUploads.php
@@ -10,7 +10,7 @@ class SupportFileUploads
 {
     static function init() { return new static; }
 
-    function __construct()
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for pitching in on this.

Unfortunately, because there isn't more understanding of the importance of array_values AND this PR is missing tests to provide assurance, AND the tests for this PR are failing. I'm closing this.

joshhanley:
@Mutembeijoe can you please open a discussion with your issue and provide example code so we can replicate it? Thanks!

PR #2051, opened 2020-11-18
Create tests from Make Component command

scottzirkel:
When creating new Livewire components, I find myself immediately creating tests for them as well. This PR adds a flag in the livewire:make (and livewire:touch) command to also generate a test file.

Running `php artisan livewire:make Foo --test` would create the Component, View, and Test files.

I have also included tests to ensure this works.

Here's a photo of a puppy chewing on a wire (probably not live) ~as a bribe~.

Cute, right?

### DIFF
--- a/src/Commands/ComponentParser.php
+++ b/src/Commands/ComponentParser.php
@@ -11,6 +11,7 @@ class ComponentParser
 {
     protected $appPath;
     protected $viewPath;
+    protected $testPath;
     protected $component;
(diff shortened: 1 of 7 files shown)


calebporzio:
Thanks for this Scott, and lol/thanks for the puppy pic.

Sorry to say though that in the interest of keeping the codebase slim, I'm going to pass on this. Thanks anyway though!

PR #2054, opened 2020-11-19
Feature/emit livewire event after validation failure

chrisrhymes:
I created a discussion, but there wasn't any responses, so apologies if this is too much of a niche request, but I thought it would be useful to others to easily display a toast message each time validation fails. 

Single change

A simple test has been added, but could probably be improved

You can listen for the failed-validation [event in JavaScript](https://laravel-livewire.com/docs/2.x/events#in-js) to display a toast message. This may not be needed on short forms, but on larger forms that require scrolling it will allow a message to be displayed to users each time they press submit on the form to inform them of validation errors. 

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -153,6 +153,10 @@ trait ValidatesInput
 
         $this->shortenModelAttributes($data, $rules, $validator);
 
+        if($validator->fails()) {
+            $this->emit('failed-validation');
+        }
+
         $validatedData = $validator->validate();
 
         $this->resetErrorBag();
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -153,11 +153,12 @@ trait ValidatesInput
 
         $this->shortenModelAttributes($data, $rules, $validator);
 
-        if($validator->fails()) {
+        try {
+            $validatedData = $validator->validate();
+        } catch (ValidationException $e) {
             $this->emit('failed-validation');
+            throw ValidationException::withMessages($e->validator->errors()->toArray());
         }
-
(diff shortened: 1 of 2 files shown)


iksaku:
Wouldn't it be better to surround the event emit in a `try... catch`, emit then re-throw the exception?

Just tinkering to prevent double validation execution, since you evaluate for a `fail` in the if statement, and, independent of the result, you then re-validate.

Couldn't this be a considerable hit for complex validation rules or bigger ones?

What do you think?

chrisrhymes:
Thanks for your feedback @iksaku I have done as you have suggested and wrapped in a try catch.

src/ComponentConcerns/ValidatesInput.php
```diff
@@ -153,7 +153,12 @@ public function validate($rules = null, $messages = [], $attributes = [])
 
         $this->shortenModelAttributes($data, $rules, $validator);
 
-        $validatedData = $validator->validate();
+        try {
+            $validatedData = $validator->validate();
+        } catch (ValidationException $e) {
+            $this->emit('failed-validation');
+            throw new ValidationException($validator);
```
iksaku, line 160: To reduce memory usage, you can simply rethrow the catched exception: `throw $e`.

iksaku (changes requested):

src/ComponentConcerns/ValidatesInput.php
chrisrhymes: Thanks for your help and patience. Every day is a school day.

crnkovic:
I'd avoid this in core of Livewire, couples too much to your application code. I'd rather leave you do it yourself in your components.

chrisrhymes:
Now I understand I can do this in my component without putting it in the Livewire core. Thanks for the help @iksaku and @crnkovic

Issue #2059, opened 2020-11-20
TrimString middleware causing checksum issues


julienarcin:
### Description

I traced back a checksum problem when my public properties are string containing spaces at the beginning/end.

### Exact steps to reproduce

```
class Component {

    public $test;

    public function mount() {
        $this->test = '    my string     ';
    }
}
```

**Context**
- Livewire version: 2.3.2
- Laravel version: 8.15.0
- Alpine version: ?
- Browser: Tested on Chrome + Safari

julienarcin:
Seems like the TrimString middleware is the source of the problem.

Reference: https://github.com/livewire/livewire/issues/354#issuecomment-552514576

julienarcin:
I found the origin of the problem.

My Laravel Middlewares (including TrimStrings) are in App\Http\Middleware\Template

And in LivewireServiceProvider, line 78 :

```
// Bypass specific middlewares during Livewire requests.
// These are usually helpful during a typical request, but
// during Livewire requests, they can damage data properties.
$this->bypassTheseMiddlewaresDuringLivewireRequests([
    TrimStrings::class,
    ConvertEmptyStringsToNull::class,
    // If the app overrode "TrimStrings".
    \App\Http\Middleware\TrimStrings::class,
]);
```

So in my specific case, the Middleware is not bypassed by Livewire.

julienarcin:
I created a PR here: https://github.com/livewire/livewire/pull/2060

austenc:
Closing this issue for now (and for reasons mentioned in #2060), but please reopen if you'd like to continue the discussion or propose another approach for making Livewire respect middleware in atypical locations. 

Thanks!

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

PR #2060, opened 2020-11-20
Add excluded middlewares in configuration file

julienarcin:
Related to this issue: https://github.com/livewire/livewire/issues/2059

My middlewares are not namespaced \App\Http\Middleware (they are in a subfolder), so the TrimString middleware isn't bypassed and it causes checksum errors.

This is an edge-case, but I think a package like Livewire shouldn't rely on a specific application directory structure (or at least, we should be able to modify it).

Second reason: The problem can be silent for a long time, and is difficult to identify and debug. It took me a few hours to understand where the checksum error was coming from. If other developers changed their middleware directory structure (like me), it can take them a long time to understand the origin of the problem.

Third reason: Not related to my problem, but it's nice to have the option to bypass other middlewares in the config file. Particularly if the user has some middlewares that can mess-up livewire requests (I'm thinking about some Pagespeeds/Cache/etc. middlewares).

I didn't delete the "overridden" \App\Http\Middleware\TrimStrings::class from the LivewireServiceProvider to avoid any backward compatibility issue. But I think it should be only in the config file by default.

Let me know what you think 😀

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -59,6 +59,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Livewire Excluded Middlewares
+    |--------------------------------------------------------------------------
+    |
+    | This value sets the middlewares that will be excluded from the main Livewire 
+    | "message" endpoint. It is set to null by default, but should be set to 
+    | exclude the TrimStrings Middleware if the location is different.
+    |
+    */
+    'middlewares_to_exclude' => [
+        //\App\Http\Middleware\TrimStrings::class,
+    ],
+    
+
+    /*
+    |--------------------------------------------------------------------------
     | Livewire Temporary File Uploads Endpoint Configuration
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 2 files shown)


joshhanley (commented):
My only concern with this, is that those middleware are disabled so the security features of livewire work correctly. There are people asking how to re-enable these middlewares but they shouldn't be re-enabled. The wording of this in the config suggests that TrimStrings is enabled and can be excluded.

I would consider changing it so it's clearer that you are just providing an updated location for these middleware, not that you can include/exclude. Maybe use a map or something, with a comment that explains the potential checksum error and to uncomment and point at the new TrimStrings location if someone is getting that error.

I believe being able to exclude additional middleware should be a seperate PR with is own section of the config. Ie: 'specify any additional middleware you want excluded from running on your Livewire components'.

Let me know what you think.
Hope this helps!

calebporzio:
Yep, I think @joshhanley explained it perfectly. Thanks for contributing! Closing this for now.

Issue #2070, opened 2020-11-21
Typed property must not be accessed before initialization on rules method during wire:model update


kingpabel:
##### Here is the component

```
<?php

namespace App\Http\Livewire\Patient;

use App\Models\Patient;
use Livewire\Component;

class Create extends Component
{
    public Patient $patient;

    public function rules()
    {
        return [
            'patient.number' => "required|unique:App\Models\Patient,number,{$this->patient->id}",
        ];
    }

    public function render()
    {
        return view('livewire.patient.create');
    }
}
```

##### Simple blade page with wire:model
```
<input type="text" wire:model="patient.number" />
```

##### During patient number update, exactly the error I am getting
```
Typed property App\Http\Livewire\Patient\Create::$patient must not be accessed before initialization
```

**Context**
- Livewire version: [2.3.2]
- Laravel version: [8.15.0]
- Browser: [Chrome]

abrardev99:
I think this is not a bug, this is a behavior. 
This should be the fix. 

```
public mount(){
         $patient = new Patient()
}
```

kingpabel:
@abrardev99 you should be right to define a new instance on the mount method, if your component purpose is creating a patient. 

I am editing my patient where patient model is binding automatically by livewire without mount method. [Reference](https://laravel-livewire.com/docs/2.x/properties#binding-models)

For editing purpose on component, sometimes you need to access model property on rules like unique validation except id.

Personally, I think for many other reason you need to access model property on rules.

3alampro:
> public mount(){
>          $patient = new Patient()
> }

I'm having the same issue on both when trying to create new model or editing existing one 

even when initiating the user model on the mount method produce the same error

3alampro:
the same issue happen when trying to access any typed properties 

```php
public string $password;
```

and try to validate the field under the rules method 

```php
'password' => [
    'required',
    'min:8',
],
```

when the form submit and trying to validate the password input you will get the same error

```
Error
Typed property App\Http\Livewire\User\Form::$password must not be accessed before initialization
```

austenc:
@kingpabel could you please provide us some more context? Just tested this in one of my own projects and it worked correctly when using route model binding on a full page livewire component. 

Is this a full page component, or are you including it in another page with `@livewire` or `<livewire:`?
What is your route definition? 

Closing, but feel free to reopen if you'd like to present more info and an example repository we may use to reproduce this. 

Thank you!

squishythejellyfish:
[...]

PR #2077, opened 2020-11-22
Model Should Assigned before call rulesForModel

imtiazPabel:
Suppose I am using livewire component for route model binding, where I have to validate ```unique``` a field with except current model id. In that scenario, during hydrate a field I can't use the model from public property to set except id on rules.

Right now, the ```rulesForModel()``` method call before model assigned on property and generate the error ```Typed property must not be accessed before initialization```

If we assigned the model on property first then call "rulesForModel()", it will work.

I have created an issue also #2070

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -102,6 +102,8 @@ class HydratePublicProperties implements HydrationMiddleware
         } else {
             $model = new $serialized['class'];
         }
+        
+        $instance->$property = $model;
 
         $dirtyModelData = $request->memo['data'][$property];
 
@@ -114,8 +116,6 @@ class HydratePublicProperties implements HydrationMiddleware
                 data_set($model, $key, data_get($dirtyModelData, $key));
             }
         }
-
-        $instance->$property = $model;
     }
 
     protected static function hydrateModels($serialized, $property, $request, $instance)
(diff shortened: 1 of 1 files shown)


iksaku:
This has been discussed, and @calebporzio left it open for more people to raise interest in this kind of case.

I think its time to consider this already :sweat_smile:.

For the record, these are previously closed PRs:
  - #1778 (Add failing tests).
  - #1652 (Previous attempt to fix this).

kingpabel:
@iksaku thanks for those references. It seems many people expect to access model property on rules. I also read your [#1652 comment](https://github.com/livewire/livewire/pull/1652#issuecomment-695813933) to serve the purpose a way. That should definitely work but not seems as clean as much.

I have done a very small rearrange to make accessible model property on rules and I am confident that should not break anything existing.

iksaku:
Yup, I don't see why this rearrange should break anything, but just wanted to provide historical reference to keep pushing the need of this feature 👀

joshhanley (commented):
I've also ran into this problem a few times, so hopefully this is a fix for it!

@kingpabel have you confirmed that the dirty data still gets applied correctly to the instance property, with model now being assigned to the instance before the dirty data is replayed onto it?

kingpabel:
@joshhanley 

> I have done a very small rearrange to make accessible model property on rules and I am confident that should not break anything existing.

joshhanley (approved):
@kingpabel I have tested your changes in a local version of Livewire and it all looks good to me!

@calebporzio I don't see any reason why this can't be merged.

All dirty data gets reapplied to the model correctly, even after the rearrange. 

I have also tested and accessing the model inside the rules method now works happily, so rules such as the below works using this change!

```php
protected function rules()
{
    return [
        'user.name' => 'unique:users,name,'.$this->user->id
    ];
}
```

iksaku (approved):

3alampro:
I'm having the same issue when trying to validate unique rules or any Dynamic or conditional  rules that depends on the status of model property inside the rules method

Hopefully this PR fixe it and got merged soon

calebporzio:
Thank you for the PR. However, it has already been fixed and will be available in the next release. #2248

PR #2123, opened 2020-11-30
use rfc 3986 in query string

mix5003:
No 

No 

Yes

This will make url consistency between `url()->full()` and real current url. 
in my case i use  `<link rel="next" href="{{ $paginate->withQueryString()->nextPageUrl() }}" />`. and google report that duplicate other url. because livewire encode whitespace to `+` but normal paginate encode it as `%20` .


calebporzio:
Thanks!

Issue #2140, opened 2020-12-01
Nested Keys passed to Updated and Updating hooks are instances of Stringable


nedwors:
### Description
Since [#1897](https://github.com/livewire/livewire/pull/1897), nested keys passed to the `Updated` and `Updating` hooks are instances of `Stringable`.

In our app, this caused a break as the equality of the key was being checked.

### Exact steps to reproduce
This [Playground](https://laravelplayground.com/#/snippets/f396211f-d58b-4711-8b74-7d759d5c0ae1) demonstrates an example of the setup that is now failing in our app. It seems that the Playground is not beyond 2.3.1 at the moment of writing, hence why it works as expected in this playground.

**Context**
- Livewire version: [^2.3.2]

I am about to submit a PR with a fix for this.

This is my first time contributing so please feel free to let me know if I have not submitted properly etc.

Thanks!

(Ps. Thanks for the amazing work!!)

nedwors:
Here is the [PR](https://github.com/livewire/livewire/pull/2141)

ryangjchandler:
@nedwors You can get scalar strings by type-hinting the property:

```php
public function updatedFoo($value, string $key)
{

}
```

nedwors:
Thanks @ryangjchandler, indeed that is a workaround 👍 We actually fixed the issue by casting to a string inside the method.

The reason for the issue and the [PR](https://github.com/livewire/livewire/pull/2141) is not so much how to solve this issue in implementation. Rather, it is that prior to [PR 1987](https://github.com/livewire/livewire/pull/1897) this was not needed. `$key` was a scalar string by default and not a `Stringable`, without need for casting or type hinting. 

If it is desired to have the `$key` now passed as a `Stringable` then that is great 👍 But it seems this behaviour wasn't an intentional addition to Livewire?

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

PR #2141, opened 2020-12-01
Adds toString for Stringables

nedwors:
This fixes a break we picked up on in an app we were using Livewire in. Issue [#2140](https://github.com/livewire/livewire/issues/2140) is open and describes further.

No

I have added PHPUnit assertions in `LifecycleHooksTest`.

Since 2.3.2, nested keys passed to the Updated and Updating hooks are instances of `Stringable`. This PR casts the key to a string before passing to the hooks so they can be interacted with as such - as it was prior to 2.3.2.

This is my first contribution to this project so if I have misunderstood/not done anything as preferred, please let me know. Thanks!

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -59,8 +59,8 @@ trait HandlesActions
         $name = str($name);
 
         $propertyName = $name->studly()->before('.');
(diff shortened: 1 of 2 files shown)


calebporzio:
Ooops, perfect, thanks!

PR #2156, opened 2020-12-03
Make extractOriginalNameFromFilePath() static

stancl:

I did not.

No

No

The method doesn't use any object state, and is basically a counterpart to the already-static `generateHashNameWithOriginalNameEmbedded()`.

Making this method static will make it possible to convert the hash back to the human-readable filename, without having to awkwardly instantiate the object.

Example from my package:
```php
->resolveDisplayValueUsing(fn (File $field, string $path) => (new TemporaryUploadedFile($path, 'public'))->extractOriginalNameFromFilePath($path))
->storeFileUsing(fn (File $field, TemporaryUploadedFile $file) => $file->storePubliclyAs('files', $file->generateHashNameWithOriginalNameEmbedded($file), ['disk' => 'public']))
```

This would make it possible to do just:
```php
->resolveDisplayValueUsing(fn (File $field, string $path) => TemporaryUploadedFile::extractOriginalNameFromFilePath($path))
```

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -54,7 +54,7 @@ class TemporaryUploadedFile extends UploadedFile
 
     public function getClientOriginalName()
     {
-        return $this->extractOriginalNameFromFilePath($this->path);
+        return static::extractOriginalNameFromFilePath($this->path);
     }
 
     public function temporaryUrl()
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for being willing to contribute, however, going to hold off on this - doesn't add enough value to most, while changing a sorta-public api

stancl:
Understandable. If I remember, I'll re-submit the PR when you have a branch for the next major release.

Issue #2025, opened 2020-11-13
After installing livewire, Stack trace showing errors in cached php files, not in original blade.php views


almazdanilov:
### Description
If we have error inside View, Laravel shows blade.php file in Stack trace.
But after installing livewire, Laravel shows cached php file in Stack trace.

### Exact steps to reproduce
Install livewire according to documentation:
```
composer require livewire/livewire
```
After that make little mistake in welcome.blade.php, f.e. put {{$unknownVariable}} at the beginning of the file. Open website and in Stack trace you will see **Illuminate\View\Engines\CompilerEngine::handleViewException** in cached php file located in storage/framework/views/. But before installing Livewire, Laravel was showing **Illuminate\Foundation\Bootstrap\HandleExceptions::handleError** in welcome.blade.php. It's impossible to quickly trace errors when Stack trace shows them in cached files. Cached view files absolutely different than their source blade.php files.

**Context**
- Livewire version: 2.3.1
- Laravel version: 8.10.0
- PHP Version: 7.4.12
- Browser: Chrome

austenc:
Thanks for reporting this @almazdanilov. It's related to https://github.com/livewire/livewire/issues/711. 

The reason you're seeing this happen is that Ignition's (the Laravel error page) main compiler engine is not running.

Livewire registers a compiler engine override, and so does Ignition. Currently, Laravel only allows one of these, so Livewire is taking precedence and the extra nice stuff from Ignition isn't working. Unfortunately, it's not a simple fix. We'll leave this open and investigate it more as we explore other options.

freekmurze:
We'll continue this in #2160 

Thanks for addressing this @austenc

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

PR #2160, opened 2020-12-04
fix ignition compatibility - #711 #2025

### LINKED ISSUE #711: Call to undefined method CompilerEngine::startLivewireRendering()
I got an error like this when displaying a livewire component in Laravel 7.0.8.

```
Facade\Ignition\Exceptions\ViewException
Call to undefined method Facade\Ignition\Views\Engines\CompilerEngine::startLivewireRendering() (View: /path/to/my/laravel7x/resources/views/livewire.blade.php)
```

 I guess the order of loading classes is the problem because `IgnitionServiceProvider` is loaded after `LivewireServiceProvider` in my environment.

Is there a way to fix it?
Thank you!

austenc:
This will fix #711 and #2025 with an additional PR submitted to facade/ignition (which I will link soon) 

No, but there are a few things we might consider (listed below)

No, but it shouldn't break any. 🤫 

Laravel expects only a single view compiler engine to be registered, and the last provider in the `bootstrap/cache/packages.php` file wins out.

There are **two scenarios** that come into play:

1. Livewire's service provider is loaded first, so ignition's compiler engine is the one the app uses (causes #711)
2. Ignition's provider is loaded first, so livewire's compiler engine is used (#2025 - missing ignition features)

This needs to be fixed from both angles. The changes should fix things on the Livewire side, and provide an additional Livewire-friendly compiler engine class that Ignition can load for apps utilizing Livewire. 

I tried to put as much of the code on the Livewire side of things as possible so that it would be the source of truth and Ignition would require minimal changes. This way, if Livewire changes, Ignition shouldn't require additional PRs. 

Another improvement (in a future PR) could be to move the `views` folder and make a `src/Views/Engines` folder to mimic the structure of both Ignition and Laravel. Putting both engine files in there as well as the `RendersLivewireComponents` trait seems like it would compartmentalized this stuff somewhat. 

Marking this as draft so we can be extra sure of how everything is named (so we don't need to PR ignition again). 

_Perhaps we could provide a static constant in Livewire which is the classname Ignition should use?_ 🤔 

🍻 🤜


austenc:
PHP 7.3 why you do this to me? 😢

joshhanley:
@austenc I had issues with the tests yesterday, I knew they should all pass, but they failed almost instantly. I waited a bit then pushed a comment change up just to re-trigger the tests and all passed successfully.

nuernbergerA (approved):

calebporzio:
This was a deep problem with a deep solution. Nice work Austen!

Issue #1886, opened 2020-10-23
Binding deep nested data broken.


InWave:
### Description
First of all, I love this package! It takes a way the hurdle of JS in dynamic interfaces. 

My problem:
I have build a very dynamic form. This is the context: A candidate can have multiple employers and at a employer he can have (had) multiple positions. Therefore we have to bind the form item of a position on the following way: 

`
protected $rules = [
        'employers.*.employer_name' => 'required|string|min:3',
        'employers.*.positions.*.position_title' => 'required',
];
`
The binding of the employer name works. That is also synced between the frontend and the Livewire component. 
However when loading the page the **position title** is not displayed. It stays blank. 

I can conclude that the bind is successful because of two reasons:
- as soon as I enter data in the form field after a few seconds the input disappears.
- when I change the name of the field, an error pops up that I have to bind the input with $rules. 

In my opinion this means that somewhere in the package the data is incorrectly handled. I have tried to figure out the behavior by looking into the hydrate / dehydrate method for dotNotedData. 
However by checking the data via dd() debugging all seems fine. I even see an update object with the original value (which is not displayed in the text input) and the new input.

This supports my hypothesis that somewhere in the source of Livewire the data is not getting correctly placed. However I ran into the limits of my debugging skills and PHP skills.

I hope someone can help me figure this out and find a solution. I am more then willing to submit a PR for this problem however I am unable to find the core issue. 

### Exact steps to reproduce

This is a example of the data structure I use:
```

  0 => array:6 [▼
    "id" => 91
    "employer_name" => "Employer 1"
    "candidate_id" => 33
    "created_at" => "2020-08-24T09:06:39.000000Z"
    "updated_at" => "2020-08-24T09:06:39.000000Z"
    "positions" => []
  ]
  1 => array:6 [▼
    "id" => 92
    "employer_name" => "Employer 2"
    "candidate_id" => 33
    "created_at" => "2020-08-24T09:06:58.000000Z"
    "updated_at" => "2020-08-24T09:06:58.000000Z"
    "positions" => array:2 [▼
      0 => array:11 [▼
        "id" => 70
        "position_title" => "Example position"
        "description" => """
          Test description
          """
        "start_date" => "2016-03-24"
        "end_date" => null
        "date_notation" => "m-Y"
        "is_current_position" => "true"
        "employer_id" => 92
        "candidate_id" => 33
        "created_at" => "2020-08-24T09:06:58.000000Z"
[...]

PR #2162, opened 2020-12-04
Fix magic toggle nested property

joshhanley:
Currently using the magic action `$toggle` doesn't work with dot notation.

Example: `wire:click="$toggle('users.0.active')"` throws the below error due to the `$this->{$prop}` call.

>Livewire\Exceptions\PropertyNotFoundException
>Property [$users.0.active] not found on component

This PR fixes that and adds tests for `$toggle`.

Hope this helps!

**Side Note:** I think a helper function should be made for getting and setting properties using dot notation. That would extract the logic out of this code and clean it up, and it could be useful in other areas.

**Related Issues:** There as a comment on #1886 noting this issue, but I don't think it's the original issue.
https://github.com/livewire/livewire/issues/1886#issuecomment-715872847

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -119,9 +119,16 @@ trait HandlesActions
 
             case '$toggle':
                 $prop = array_shift($params);
-                $propertyName = $this->beforeFirstDot($prop);
-                $targetKey = $this->afterFirstDot($prop);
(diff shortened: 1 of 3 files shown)


Melcus:
Got the issue with nested toggle today myself. Glad someone put in the effort for it.

calebporzio:
This is great, thanks @joshhanley!

PR #2176, opened 2020-12-05
Support constructor DI

phanan:
This PR attempts to add support for constructor's dependency injection, allowing usage of dependent services throughout the component's actions and events:

```php
class FooComponent extends Component {
    private Dep $dep;

    public function __construct(Dep $dep) 
    {
        $this->dep = $dep;
    }

    public function hydrate(Request $request): void
    {
        $this->dep->handleRequest($request);
    }

    public function updatedFoo(string $value): void
    {
        $this->dep->doSomethingWithFoo($value);
    }
}
```

Also this way, one wouldn't need to resort to method DI. For components that heavily depend on external services, this should keep the codebase as DRY as possible.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -239,4 +239,9 @@ abstract class Component
             'Method %s::%s does not exist.', static::class, $method
         ));
     }
+
+    public function setId($id): void
+    {
+        $this->id = $id;
+    }
 }
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -65,7 +65,9 @@ class LivewireManager
             "Component [{$component}] class not found: [{$componentClass}]"
         ));
 
-        return new $componentClass($id);
+        return tap(app($componentClass), static function (Component $component) use ($id): void {
+            $component->setId($id);
+        });
     }
 
     public function mount($name, $params = [])
--- a/tests/Unit/ComponentDependencyInjectionTest.php
+++ b/tests/Unit/ComponentDependencyInjectionTest.php
@@ -9,6 +9,14 @@ use Illuminate\Routing\UrlGenerator;
 class ComponentDependencyInjectionTest extends TestCase
 {
     /** @test */
+    public function component_constructor_with_dependency()
+    {
+        $component = app(ComponentWithDependencyInjection::class);
+        
+        $this->assertInstanceOf(Dependency::class, $component->getDep());
+    }
+
+    /** @test */
     public function component_mount_action_with_dependency()
     {
         $component = Livewire::test(ComponentWithDependencyInjection::class, ['id' => 123]);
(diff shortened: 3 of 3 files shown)


ryangjchandler:
From a project maintenance point of a view, the constructor is one of the few methods Livewire has internal control of so a change like this wouldn't make much sense.

Currently, the constructor isn't really available as part of Livewire's public API so any backwards-incompatible changes don't need to made as part of a major version change (the beauty of internal use only, even if PHP has no nice way of making this strict).

ryangjchandler:
Also, as an extension, the example given in the PR description isn't a very good one since the resolution of `$this->dep` could be moved into a getter function on the component, or even into a computed property.

phanan:
I'm not sure I follow. The fact that Livewire has internal control of the constructor (?) and constructor DI shouldn't be mutually exclusive. 

One thing I believe, having no support for constructor DI is a bummer, and IMHO one of the biggest limitations of Livewire at the moment. For instance, correct me if I'm wrong, but method DI for hooks (`updated`, `updating` etc.) doesn't work. Of course, one can resort to other solutions, but IMO they tend to be clumsy workarounds and at times can be violations of software design principles.

Granted, this PR might not have the best shape (I don't even know how test it with the current Livewire test flow), but having no constructor DI is, again, a major drawback.

ryangjchandler:
The PR's logic is fine and is how I'd tackle the problem too. The only reason I put my thoughts on the PR is because if the change is so simple, you have to ask why it isn't supported in the first place.

I'm pretty sure this functionality has been asked for in the past too but turned down. Of course, it's down to Caleb!

phanan:
> The only reason I put my thoughts on the PR is because if the change is so simple, you have to ask why it isn't supported in the first place.

I actually did a search before the attempt, but to my surprise, no actual mention of constructor DI showed up. Unless I missed something obvious, my best guess is that most developers in Laravel/Livewire community tend to use facades (`Dependency::foo()`) and dynamic container resolving (`app(Dependency::class)`) instead of constructor DI.

ryangjchandler:
> > The only reason I put my thoughts on the PR is because if the change is so simple, you have to ask why it isn't supported in the first place.
> 
> I actually did a search before the attempt, but to my surprise, no actual mention of constructor DI showed up. Unless I missed something obvious, my best guess is that most developers in Laravel/Livewire community tend to use facades (`Dependency::foo()`) and dynamic container resolving (`app(Dependency::class)`) instead of constructor DI.

It has come up in the Discord quite a bit tbf, but most people will use the methods that Livewire provides instead of the constructor for ease I imagine.

joshhanley:
@phanan any chance you can provide a specific use case for this? 

The reason I ask is because you can use DI through the mount method, hydrate methods, and any of the action methods. Is there a reason you can't use these?

If you can provide a specific use case, that will help support your PR!

phanan:
@joshhanley As mentioned in my previous comment, AFAIK method DI hooks doesn't work. 

Even though, that it's not just about "specific use cases." Constructor DI is a popular, standard, pattern. I'm sure we can find a bunch of articles about it on Google, but a couple of (buzzwords) in short: testing, maintainability, and reasonability. 

Also, different DI approaches (setter/method/constructor) are not mutually exclusive, though IMO constructor DI most of the time eliminates the need of others.

nathanjansen:
DI works in the mount method as wel if I am not mistaken. So the mount method can be the starting point for your component (as it currently is for Livewire) in steady of the construct.

calebporzio:
Yeah, thanks for weighing in everyone. That is my thinking. For Livewire component's __construct is internal, use `mount()` (it supports the same DI and what-not)

PR #2180, opened 2020-12-06
Replace partial HTML by Selector

barryvdh:
This is based on the concept of SelectorMorphs from https://docs.stimulusreflex.com/morph-modes#selector-morphs

This allows to replace a partial of the component instead of the full component. This avoids re-rendering the entire view if you just want to show a result box / update a count/name/value. 

- [x] Update `handleMorph` to allow custom root element
- [x] Method to add HTML for an selector
- [x] Add html/selector to output
- [x] Add tests
- [x] Replace the inside of the selector, not the selector itself (or add root element?)
- [ ] Think about new events, lifecycle etc?

What do you think?

Usage inside component:

```php
function someAction(){
    // Do something
    $this->morphSelector('#result', 'Success!');
    $this->morphSelector('.counter', $this->counter);
}
```


js/component/index.js
```diff
@@ -362,11 +369,11 @@ export default class Component {
...
     }
 
-    handleMorph(dom) {
+    handleMorph(dom, el) {
         this.morphChanges = { changed: [], added: [], removed: [] }
 
-        morphdom(this.el, dom, {
-            childrenOnly: false,
+        morphdom(el || this.el, dom, {
+            childrenOnly: !!el,
```
barryvdh, line 376: Replace the children only, when using a custom selector element.

src/Component.php
```diff
@@ -104,6 +106,17 @@ public function skipRender()
         $this->shouldSkipRender = true;
     }
 
+    public function renderSelector($selector, $html)
+    {
+        $this->outputSelector = $selector;
+        $this->preRenderedOutput = '<div>' . $html .'</div>';
```
barryvdh, line 112: Wrap it inside a div for the morphdom.

src/RenameMe/OptimizeRenderedDom.php
```diff
@@ -25,7 +25,7 @@ function __construct()
 
             $response->memo['htmlHash'] = $newHash = hash('crc32b', $response->effects['html']);
 
-            if ($oldHash === $newHash) {
+            if ($oldHash === $newHash && !isset($response->effects['selector'])) {
```
barryvdh, line 28: Not sure if we should invalidate it or store selector HTML differently

barryvdh:
Thoughts: 
- perhaps the selector should be scoped inside the root element of the component?
- should it be possible to update multiple selectors?

barryvdh:
Tweak the naming a bit and scoped the selectors inside the base element, to provide conflicts.
Also made it an array to be able to morph multiple selectors in 1 action.

https://github.com/livewire/livewire/pull/2181 should be merged to make it a bit more efficient.

ryangjchandler:
@barryvdh This is some impressive work so far and would be a game-changer for Livewire in my opinion! Great stuff.

barryvdh:
@ryangjchandler Thank you.

It's perhaps worth noting that in actual cases, the output should still match a full re-render for best experience. So the updates that were made with the query selectors, should also be visible when just re-rendering the page.

calebporzio:
This is super interesting!

calebporzio:
Ok, while I really like the idea of providing an affordance for people to target specific morphs to not waste resources (response payload size, morphdom diffing time and complexity), I'm hesitant to introduce a solution for a problem I haven't heard much about.

I think this might be a good feature, I just want to make sure the community really expresses a need for it with practical reasons.

Also, I would be interested to see what this would look like as a userland implementation. I think this would be fairly straightforward to introduce as a composer package or something, with the exception of the skipping morphdom on the front-end part.

But we can discuss if there is a change to Livewire's internals that would allow you to skip the morphdom step from a hook? 

I don't know, just thinking out loud.

barryvdh:
Well I guess it isn't so much as a solution to a problem, more of an optimisation. It works already, only can be more efficient in more cases.

barryvdh:
I think an example would be a result message (success/fail or an output), similar to what you would doe with ajax on a button. For tables/list it could be updating a single row/item (eg a nested blade template) after op pushing an action button. I think removing a row would mean just updating it with an empty string.

But I could also imagine that a different approach might be better, if you want to take appends/prepends/deletes into account, eg. add a chat message without re-rendering all. Not sure how Liveview or perhaps Hotwire/Turbo? do this.

joshhanley:
I imagine this could also be used in an infinite scroll/ load more situation for appending new results to the end of the page. That way only the new results html would be transmitted rather than all html with results being transmitted each time.

One area that I would use it in is with autocomplete search dropdown results. 

Currently I have a form with 4 autocomplete dropdown inputs where everything is in the one form component. I don't want to split it into components as trying to handle communication between the form component and each autocomplete component, plus interaction between inputs (as they cascade) using events and keeping state in sync to me seems too problematic.

So being able to just target the dropdown of a specific input to update its results and html would be great.

I also have dynamic inputs in rows on that form as well, which get generated when the users clicks add another, which I think could also benefit from this.

Hope that gives you an insight to the benefits for me 🙂  I'm also open to other options/ ideas if anyone has any!

barryvdh:
But good point on userland implementations, because now that I think of it, sometimes you might want to replace the contents of a div (result output etc), but sometimes you would just want to replace the entire div (eg update row or a blade partial). So maybe a mode option would be better, but more complicated.

calebporzio:
Good input.

I definitely agree there should be an API for more granular HTML updates. Especially in the case of infinite-scroll.

I'm still not sure the right abstraction though. Ideally we arrive on an approach that doesn't involve referencing CSS selectors, but we'll see.

I'm going to close this for now as I don't think we'll go with this API, but we should certainly keep the conversation going. Maybe in a GH discussion?

Thanks so much for the original PR @barryvdh ! And thank you for all your contributions to the community in general over the years!

Issue #940, opened 2020-05-08
skip render when redirect is set


snapey:
**Is your feature request related to a problem? Please describe.**
I have a component that renders a model.  In this component is a delete button. When the delete is hit, the function removes the model and sets the redirect.  Control then passes to the render() function which tries to render the now deleted model, resulting in an error.

**Describe the solution you'd like**
Issue a redirect without needing to render the full component.

**Describe alternatives you've considered**
I have a workaround that checks `$this->redirectTo` in the render method, ane immediately returns.  Unfortunately it also needs to return a view, and this must be populated with valid data even though it will not be seen.
```
    public function render()
    {
        if($this->redirectTo){
            return view('livewire.foodbanks.foodbank-card')->withFoodbank(new Foodbank());
        }

        // etc with the required render functionality
```

jhoff:
I came across this as well. A slightly more efficient workaround is to override the `output()` method and return a simple string if `$this->redirectTo` is set:

```

    /**
     * Prevent render when redirecting
     *
     * @param Illuminate\Support\MessageBag $errors
     *
     * @return string
     */
    public function output($errors = null)
    {
        if ($this->redirectTo) {
            return '<div></div>';
        }

        return parent::output($errors);
    }
```

calebporzio:
Ah interesting.

I think you're right. If an action before ->render() redirects, we should skip the rendering.

This would be a breaking change, so I will make a note to do this for V2.

Thanks!

zepfietje:
Has this been implemented in V2? As far as I can see, I don't think so.

SRWieZ:
Similarly we can't do redirections in mount() function in a full page component.

yoeriboven:
Curious as well. Would be very useful to redirect before a user sees a page they shouldn't see.

redbastie:
You can use `$this->skipRender()`

zepfietje:
> You can use `$this->skipRender()`

This doesn't work for me, as it breaks rendering the page being redirected to.

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

[...]

PR #2181, opened 2020-12-06
Fix skipRender() to actually skip the render method

barryvdh:
Skip the actual render on skipRender(). Same as https://github.com/livewire/livewire/pull/1738 but with a test to confirm it.

Not sure why the render() method is still called otherwise, which can load/query data still, but it's not used.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -106,6 +106,8 @@ abstract class Component
 
     public function renderToView()
     {
+        if ($this->shouldSkipRender) return null;
+
         Livewire::dispatch('component.rendering', $this);
 
         $view = method_exists($this, 'render')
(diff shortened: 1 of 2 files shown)


barryvdh:
Not really sure why builds fail, seems like a network issue?

joshhanley:
@barryvdh yeah I had the same issue the other day, I force pushed a commit message change just to re-run them to make sure they're all ok.

PR looks good though!

nuernbergerA:
@barryvdh Is there another use-case for this aside from `redirect()` in `mount()`?

austenc:
Looks like this will fix #940 also.

nuernbergerA (approved):

calebporzio:
Thanks @barryvdh for such a good clean tested PR!

Issue #1965, opened 2020-11-03
Loading target for action with argument


mokhosh:
### Description
I have a loop within which there are buttons that trigger an action and have loading states. So:
```php
@foreach($accounts as $account)
    <button wire:click="edit({{ $account->id }})">
        <x-icon.pencil wire:loading.remove wire:target="edit"/>
        <x-icon.spinner wire:loading wire:target="edit"/>
    </button>
@endforeach
```
But when I click on one, all of them get the loading state. I used to do this in version one:
```php
@foreach($accounts as $account)
    <button wire:click="edit({{ $account->id }})">
        <x-icon.pencil wire:loading.remove wire:target="edit({{ $account->id }})"/>
        <x-icon.spinner wire:loading wire:target="edit({{ $account->id }})"/>
    </button>
@endforeach
```
But that doesn't work in v2.

austenc:
Hey @mokhosh this is on the roadmap, and is a duplicate of https://github.com/livewire/livewire/issues/1780 👍

mokhosh:
As the great @calebporzio once said on that issue:
> Yep! Good to know there's more demand for it thought.

😄
Thanks @austenc

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

calebporzio:
On the roadmap.

ejulfaey:
Does is support with multiple target? Below code is not working.


                                                                                                                                                                                       
        @foreach($cars as $car)                                                                                                                                                                                  
        <div wire:loading.flex wire:target="updateStatus('add', '1'), updateStatus('remove', '1')" class="justify-content-center p-2">
            <div class="spinner spinner-success"></div>
        </div>
        <div wire:loading.remove wire:target="updateStatus('add', '1'), updateStatus('remove', '1')">
            <button wire:click="updateStatus('add', 1)" class="btn btn-icon btn-sm btn-outline-danger">
                <i class="uil uil-multiply"></i>
            </button>
[...]

Issue #1780, opened 2020-10-08
wire:target for action + parameter


redbastie:
I'm trying to having a loading state for a specific action with a specific parameter.

I've tried this:

`wire:target="myAction(22)"`

Which does not work. However, just using `myAction` works, but effects all buttons with that action.

How do I target an action AND parameter for loading states?

iksaku:
Not possible at the momento, but you could try and PR that feature, I'm pretty sure it'll come in handy for more people

austenc:
@redbastie another option may be to make a new child component?

danharrin:
This is on the [roadmap](https://github.com/livewire/livewire#roadmap) 👍

calebporzio:
Yep! Good to know there's more demand for it thought. Thanks @danharrin !

PR #2189, opened 2020-12-09
Support loading targets with params

austenc:
This PR adds parameter support to `wire:target` directives as requested in #881 #1965 #1780. 

For example:
```
wire:target="foo('bar')"
```

Let me know if you'd like me to fix some of the weird formatting that prettier did, or you can think of other scenarios we should add tests for.


joshhanley (approved):
@austenc looks good!

I was originally thinking to use some form of wire:key or id or something to make the target specific, but I think this is a simpler solution.

The only disadvantage I can see is having to duplicate the whole method call (including all params).

Thanks!

calebporzio:
Sick, thanks Austen!

xdimension:
Thanks for the fix. But I have an issue when I'm using pagination and go to another page, somehow the loading element doesn't show up (I notice that livewire doesn't add `inline-block` style).

PR #2202, opened 2020-12-11
Remove unused uploadDangerous function on FileUploadsTest

manelgavalda:
No

No

Yes

The `uploadDangerous` function from `FileUploadsTest` is not used.

This function was introduced (maybe for convenience while testing) in [this commit](https://github.com/livewire/livewire/commit/7d4af7df6160792ef5eb148e0a21b2ab9b19d576#diff-92ed53a475f7ff939fc2af939bbe806ff880b10dd936a0867565a544b40b331f), but is not removed in the next cleanup commit


calebporzio:
Perfect, thanks!

Issue #2209, opened 2020-12-11
TemporaryUploadedFile serialization issue


gaetan-hexadog:
### Description
When trying to pass a TemporaryUploadedFile to an event (to pass it to a parent component), the class is not well serialized:

Following object passed into an `emit` call
```php
Livewire\TemporaryUploadedFile {#1659 ▼
    +"disk": "local"
    #storage: Illuminate\Filesystem\FilesystemAdapter {#1893 ▶}
    #path: "livewire-tmp/eypmQmwEA2I438gXJCvNnRUwY0yXOB-metadGltZS1tYW5hZ21lbnQucG5n-.png"
    -test: false
    -originalName: "eypmQmwEA2I438gXJCvNnRUwY0yXOB-metadGltZS1tYW5hZ21lbnQucG5n-.png"
    -mimeType: "application/octet-stream"
    -error: 0
    #hashName: null
    path: "/private/var/folders/vv/6gx44q_s3_q2pws4hwhxdcs00000gn/T"
    filename: "eypmQmwEA2I438gXJCvNnRUwY0yXOB-metadGltZS1tYW5hZ21lbnQucG5n-.png"
    basename: "phpqHUhwY"
    pathname: "/private/var/folders/vv/6gx44q_s3_q2pws4hwhxdcs00000gn/T/phpqHUhwY"
    extension: ""
    realPath: "/Users/gaetan/Desktop/Projets/laravel/storage/app/livewire-tmp/eypmQmwEA2I438gXJCvNnRUwY0yXOB-metadGltZS1tYW5hZ21lbnQucG5n-.png"
    aTime: 2020-12-11 20:41:11
    mTime: 2020-12-11 20:41:11
    cTime: 2020-12-11 20:41:11
    inode: 3616610
    size: 125358
    writable: false
    readable: false
    executable: false
    file: false
    dir: false
    link: false
  }
```

is retreived like that at the end:

```php
array:1 [▼
    "disk" => "local"
  ]
```

only `+"disk": "local"` property is serialized. Not the all object. So I can't access to the TemporaryUploadedFile into listeners.

**Context**
- Livewire version: 2.32
- Laravel version: 8.13

PR #2212, opened 2020-12-12
Get the correct data when json TemporaryUploadedFile

mtawil:
In somehow, I keep getting `[{"disk": "local"}, {"disk": "local"}]` as a result when I do something like `@this.set(field, ['livewire-file:***', 'livewire-file:***'])` which means there is a `json_encode` somewhere on response lifecycle and it leads to an incorrect `TemporaryUploadedFile` serialized data, so this fixed it.

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -7,7 +7,7 @@ use Illuminate\Http\UploadedFile;
 use Illuminate\Support\Facades\URL;
 use Illuminate\Support\Facades\Storage;
 
-class TemporaryUploadedFile extends UploadedFile
+class TemporaryUploadedFile extends UploadedFile implements JsonSerializable
 {
     protected $storage;
     protected $path;
@@ -183,5 +183,10 @@ class TemporaryUploadedFile extends UploadedFile
     public static function serializeMultipleForLivewireResponse($files)
     {
         return 'livewire-files:'.json_encode(collect($files)->map->getFilename());
+    }
+
+    public function jsonSerialize()
+    {
+        return $this->serializeForLivewireResponse();
     }
 }
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -6,6 +6,7 @@ use Illuminate\Support\Arr;
 use Illuminate\Http\UploadedFile;
 use Illuminate\Support\Facades\URL;
 use Illuminate\Support\Facades\Storage;
+use JsonSerializable;
 
 class TemporaryUploadedFile extends UploadedFile implements JsonSerializable
 {
--- a/tests/Browser/FileUploads/Component.php
+++ b/tests/Browser/FileUploads/Component.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Tests\Browser\FileUploads;
+
+use Illuminate\Support\Facades\Storage;
+use Illuminate\Support\Facades\View;
+use Illuminate\Support\Stringable;
+use Livewire\Component as BaseComponent;
+use Livewire\WithFileUploads;
+
+class Component extends BaseComponent
+{
+    use WithFileUploads;
+
+    public $foo;
+
+    public function render()
+    {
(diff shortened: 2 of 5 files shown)


mtawil:
I just noticed that there is an opened issue for it: https://github.com/livewire/livewire/issues/2209

calebporzio:
The serialized version of a temporary file is currently meant to be an internal concern ("livewire-file:...").

File uploads have serious security implications and I would only feel comfortable merging something like this if there were ample tests and explanations why this won't now allow people to manipulate the temporary file and gain access to other files they otherwise couldn't.

I'm open to it, especially the use case described in the linked issue, just need someone to do some legwork on justification here.

Thanks for contributing!

mtawil:
Hi @calebporzio 

The problem here, is that I am using FilePond for handling file uploads

To reproduce it:
1. edit the record (model) to preview the uploaded files. 
2. upload any file side by side with the old uploaded files.
3. delete one of the them previous uploaded files.

So, to do it, we will use the FilePond event `onremovefile` then use `@this.get('uploaded_files')` to get the uploaded files first (so we can get the new files uploaded) as an array, then reject the deleted one, then upload the accepted `@this.set('uploaded_files', uploaded_files);`

Anyway, in a simplest case, you can reproduce it by just doing this `@this.set('uploaded_files', @this.get('uploaded_files'))` you will see that the property `uploaded_files` value is interrupted and turns into something like `[{"disk": "local"}, {"disk": "local"}]`

That is because the response we've got when using `@this.get` is json, which means the property has been json encoded somewhere; and that's why I implement `JsonSerializable` to get the correct data.

mtawil:
@calebporzio I've added a test for this, please have a check on it.

Thanks!

calebporzio:
Hi @mtawil, thanks for adding a test.

After digging further into this, I don't think either of us understand the full scope of a change like this.

If someone is willing to put in the work to fully understand the upload system and list out all of the exact reasons this doesn't currently work, why it should, and why it's safe - I'd be happy to look at a PR. But unfortunately, there isn't enough demand here to justify my own time personally.

Thanks so much for being willing to contribute!

mtawil:
Hello @calebporzio, there is still a bug that caused when trying to `get` then `set` the file-upload field property on the client-side. You can still try to test and reproduce it using the test case I provided in this PR, and definitely, it will be failed on the last release.

Honestly, I am using `get` and `set` methods for Fileponds to get previous uploaded files for preview; that's why I need this issue to get fixed. Otherwise, the only functionality that Filepond can use with livewire and alpine is for uploading only.

If you won't proceed with this, then is there any way to get this fixed? Some workaround, maybe?

Thank you.

elbakkali:
REOPEN

elbakkali:
I am having multiple instances of filepond on the same page, filepond is a livewire component, I am using it to add an array of files to a value that's in an array of tweets

So i have a an array of tweets, each tweet has an array of attachements

The issue is, some of the file objects get assigned as `array:1 ["disk" => "local"]` instead of `Livewire\TemporaryUploadedFile` instance, please check the screenshot https://imgur.com/a/o5MBVYz for a better visualization, I tried everything i could for the past few days, and my project depends on it, I am desesperate for any help I could get...

This is the code inside my filepond component:

```
<div
    x-data="{ id : 'input' + index }"
    wire:ignore
    wire:key="input"
    x-data="{pond: null}"
    x-init="
        FilePond.registerPlugin(FilePondPluginImagePreview);
        pond = FilePond.create($refs.filepond_input);
        pond.setOptions({
            allowMultiple: true,
            server: {
                process: (fieldName, file, metadata, load, error, progress, abort, transfer, options) => {
                    @this.upload(`tweets.${tweet.id}.attachements`, file, load, error, progress)
                },
                revert: (filename, load) => {
                    @this.removeUpload(`tweets.${tweet.id}.attachements`, fileName, load)
                },
            },
        });
    "
>
    <input type="file" class="testfile" name="filepond" x-ref="filepond_input">
</div>

```

Issue #1110, opened 2020-06-15
Redirect not working in mount.


sohail-panezai:
### Description
Cannot redirect from mount lifecycle hook.

1. redirect('/') doesn't work.
2. $this->redirect('/') doesn't work.

### Steps to reproduce
1. Make livewire component
2. Redirect from mount method.

**Context**
- Livewire version: 1.2.0
- Laravel version: 5.8.28
- Browser: Chrome

nuernbergerA:
What would be a use case to redirect on mount?

acurrieclark:
I have a case where there is a 3 part form, with the `part` being added as a query string. If the user attempts to go straight to part 3, then I need to redirect them back to 2. 

I can do it in the view, but the URL doesn't update without a redirect.

sohail-panezai:
The component depended upon some data in the session. If that is not present, I redirect. 
I was using
`Route::livewire('route', 'component')`
and changed that to good old controller method in which I checked for that session data and redirected from there.
But, I still hoped it worked.

jrleonr:
Same here! I am using Route::livewire, and I check some values and redirect some times, it kind of stops the execution for mount() but still renders the view.

I had to create a middleware for that but it is annoying...

sohail-panezai:
Caleb had this issue fixed I guess in a previous update. However, its not working on the master branch now.
Check issue [#378 ](url)

georgeboot:
This is actually something we do a lot. Almost all pages with forms (create X page) are livewire components.

If the mount, we usually check some conditions and redirect away if not allowed.

However, livewire will still load the component, but push a redirect instruction in the js payload. In practice, the component loads and some 300ms later, your browser will redirect.

It would be amazing if Livewire would implement "real" redirects here - so that the render method will not even be called.

tanthammar:
I also have the need to redirect in the mount() method.
My use case is a create form that should not be displayed if the user has not yet created relational data.

amcsi:
I'm getting this issue as well.

I simply want to redirect the user to a different page (based on a condition) on landing on my Livewire component/page, but it has no effect in the `mount()` method.

Qirel:
Can confirm that `redirect()` **works** within the `mount()` method in the latest release. If you're not running lower versions for a specific reason and can work around that, I suggest you update your Livewire version. 

Regarding authorization - you can throw exceptions with `abort(403)` / `abort(401)` that will fire before the component is rendered.

amcsi:
[...]

PR #2223, opened 2020-12-15
Skip rendering on redirects

### LINKED ISSUE #940: skip render when redirect is set

nuernbergerA:
Fixes #940

### DIFF
--- a/src/ComponentConcerns/PerformsRedirects.php
+++ b/src/ComponentConcerns/PerformsRedirects.php
@@ -9,15 +9,18 @@ trait PerformsRedirects
     public function redirect($url)
     {
         $this->redirectTo = $url;
+        $this->shouldSkipRender = true;
     }
 
     public function redirectRoute($name, $parameters = [], $absolute = true)
     {
         $this->redirectTo = route($name, $parameters, $absolute);
+        $this->shouldSkipRender = true;
     }
 
     public function redirectAction($name, $parameters = [], $absolute = true)
     {
         $this->redirectTo = action($name, $parameters, $absolute);
+        $this->shouldSkipRender = true;
     }
 }
--- a/tests/Unit/ComponentSkipRenderTest.php
+++ b/tests/Unit/ComponentSkipRenderTest.php
@@ -27,6 +27,15 @@ class ComponentSkipRenderTest extends TestCase
 
         $this->assertNull($component->payload['effects']['html']);
     }
+
+    /** @test */
+    public function on_redirect_in_mount_render_is_not_called()
+    {
+        $component = Livewire::test(ComponentSkipRenderOnRedirectInMountStub::class);
+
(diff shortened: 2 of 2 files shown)


zepfietje:
This probably also fixes https://github.com/livewire/livewire/issues/1110.

calebporzio:
Thanks @nuernbergerA !!!

AdvancedStyle:
How does this actually work?

I've added a redirect to mount and it just shows my empty layout and doesn't redirect?

```
public function mount(){
return redirect()->to(route('home'));
}
```

nuernbergerA:
@AdvancedStyleits `$this->redirect()`

AdvancedStyle:
I tried this too, same thing happens, just shows empty layout, no redirect:

```
public function mount(){
$this->redirect(route('home'));
}
```

```
public function mount(){
return $this->redirect(route('home'));
}
```
etc... (I tried many different ways to use redirect())

zepfietje:
@AdvancedStyle @nuernbergerA, this hasn't been released yet. So if your project is still in development, you can use `dev-master` for the Livewire dependency. Else, you'll have to wait until a new version gets tagged (no idea when).

nuernbergerA:
@AdvancedStyle can you please open an issue with exact steps to reproduce. i assume it's related to turbolinks.

cheers
adrian

AdvancedStyle:
I don't think it is turbolink related seeing as I've tried removing turbolinks from the project and the problem is still the same.

Unfortunately I can't make a laravel playground example seeing as this version is not yet tagged :( I'll make an example and open issue once this is possible.

Just to be sure that I'm not doing something dumb; it should work like this right?

If you add a web route directly to a Livewire Component (such as):

```
Route::get('mypage', \App\Http\Livewire\MyPage\Index::class)->name('mypage');
```

Then put a redirect in the mount, such as below, then it should never render a 'mypage' and just redirect straight to the homepage?

```
public function mount(){
$this->redirect(route('home'));
}
```

nuernbergerA:
@AdvancedStyle can you test again with the latest dev-master?

AdvancedStyle:
@nuernbergerA Yes seems to work just like a normal Laravel redirect now 👍

PR #2224, opened 2020-12-16
Fix entangle nested arrays

joshhanley:
If you use entangle on a nested array, and an element gets removed from the array, you get a type error
> Uncaught (in promise) TypeError: Cannot read property 'id' of undefined

It is caused by the reduce function trying to access a property on undefined, throwing this error.

This PR adds a check to the reduce function to fix this and includes a test (which was failing before this fix).

Hope this helps!

---
Fixes #2191


calebporzio:
Thank you so much, this is such a good PR and makes `get()` so much more robust for all of it's uses

joshhanley:
No worries!

uxm:
I am experiencing an issue with removing the first item in an array of arrays. Am using the latest version of livewire. My error looks like this one here. (Uncaught (in promise) DOMException: String contains an invalid character livewire.js:13)

How would I resolve the issue? Thanks

joshhanley:
@uxm can you see if there are any issues that match your problem and add some details there, or open a new issue if there isn't any that are similar.

PR #2225, opened 2020-12-16
Fix hydration hook, property hydration called twice

SanderSander:
Currently property hydration functions like `hydrateFoo` are called twice, this PR simply fixes that.
I assumed that calling the hydrate function should always be done because of the following documentation in the code.

```
// Call magic hydrateProperty methods on the component.
// If the method doesn't exist, the __call with eat it.
```

I didn't create an issue because the fix seems to be trival.


calebporzio:
Thanks for contribution! Do you feel able to add tests to this PR. This changes a critical part of Livewire and I won't feel comfortable merging without tests.

SanderSander:
@calebporzio Done, would it be nice to count all life cycle invocations? Currently I only did the (de)hydration hooks.

calebporzio:
Perfect, thanks!

I'm fine with only the count assertions for this case.

PR #2226, opened 2020-12-16
Add failing test for server-rendered selected option

ragulka:

This PR demonstrates an unexpected behavior/bug in Livewire when using `select` elements with options where the `selected` attribute is set on the server.

No

Only a test is included!

I stumbled upon this issue when I created a simple blade component for form selects, where I can pass in the options as an array. The component loops over the options, rendering each one and setting the `selected` attribute on the currently selected value. I was using this component on regular server-rendered pages and it works fine. However, when dropping it into a Livewire component, and adding a `wire:model` property to it - it would start deselecting the selected value.

Here's a demo on Laravel Playground: https://laravelplayground.com/#/snippets/70b3675e-286f-4509-906d-3be0508a403d

I think the behaviour at the moment can be considered a bug. I would expect I can just re-use my blade select component for both regular blade views and inside Livewire components.

I was able to fix the issue by commenting out https://github.com/livewire/livewire/blob/master/js/component/index.js#L424-L429 - all tests were green. However, I'm not sure of the full implications of that change, so I didn't want to PR it as a fix myself.

Sidenote: it seems that MorphDOM itself might have some issues with selecting the correct element when any option attributes aside from `selected` have changed (such as some `data-` attribute).

### DIFF
--- a/tests/Browser/DataBinding/InputSelect/Component.php
+++ b/tests/Browser/DataBinding/InputSelect/Component.php
@@ -11,6 +11,8 @@ class Component extends BaseComponent
     public $singleValue;
     public $singleNumber = 3;
     public $placeholder = '';
+    public $selected = '';
+    public $selectOptions = ['', 'foo', 'bar', 'baz'];
     public $multiple = [];
 
     public function render()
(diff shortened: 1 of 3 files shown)


calebporzio:
Ah I gotcha here.

Yeah, I can see why this seems confusing and frustrating.

For now, I'll say you should add a blade conditional that checks if the user of the component is trying to `wire:model` and selectively removes the selected markings on the `<option>` tags.

Like you mentioned I think there would be other (morphdom related?) implications by supporting this use case.

Also I aim to make `wire:model`, `x-model` (Alpine), and `v-model` (Vue) all behave the same so that you don't have to remember the in's and outs of each.

Closing this. Thank you for the great PR with tests!

Issue #2061, opened 2020-11-20
Previewing uploaded SVG image does not work


AntonioDiPassio-AppSys:
Trying to preview an uploaded svgfile with $img->temporaryUrl() doesn't work.
The response generated by the FilePreviewHandler always contains the header Content-Type: application/javascript 
For a png or jpg, this is no problem because Chrome or Edge will render it regardless.
But when used with an svg the image is not rendered.

Tested on Laravel 8.14 and livewire 2.3.2

austenc:
This should be fixed in the next version, thanks for reporting it!

See #2230

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

PR #2230, opened 2020-12-16
fixed svg preview mime type detection

austenc:
**Is this something that is wanted/needed? Did you create an issue / discussion about it first?**

An issue was opened here: https://github.com/livewire/livewire/issues/2061

This PR improves the `Content-Type` sent for preview image responses by first trying to guess the file's mime type. Additionally, it accounts for a PHP bug with `.svg` mime type detection so that preview images for uploaded SVGs also work.

While investigating this, I noticed Livewire is always sending `application/json` as the mime type for file previews. Instead, it should probably try to guess the mime type and respond with that in the `Content-Type` header. 

There's a problem with SVG mime type detection in PHP though, see [here](https://github.com/symfony/symfony/issues/37148) and [here](https://bugs.php.net/bug.php?id=79045).

The PR includes a workaround for this PHP bug. Without it, all SVG's are detected as `image/svg` (instead of `image/svg+xml`) and the previews do not show up correctly. 

**Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have)**
It doesn't, but if you have an idea of how to test this, let me know!

### DIFF
--- a/src/Controllers/CanPretendToBeAFile.php
+++ b/src/Controllers/CanPretendToBeAFile.php
@@ -4,7 +4,7 @@ namespace Livewire\Controllers;
 
 trait CanPretendToBeAFile
 {
-    public function pretendResponseIsFile($file)
+    public function pretendResponseIsFile($file, $mimeType = 'application/javascript')
     {
         $expires = strtotime('+1 year');
         $lastModified = filemtime($file);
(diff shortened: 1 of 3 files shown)


joshhanley:
@austenc I looked at that and couldn't work out best how to solve it. That's a nice fix!

calebporzio:
Looks great, thanks Austen!!

PR #2234, opened 2020-12-17
Changed messages to validationMessages for better namespacing

colynb:

Yes, see my PR in the docs over at 
https://github.com/livewire/docs/pull/307

No.

Yes; I modified the existing ValidateTest to include this change.

This started out with me creating a component for storing chat messages. Logically, this would include a property called "messages" for storing a collection of chat messages. There were no issues until I executed `$this->validate()` when I started seeing a `Validation::factory` error that the 3rd argument was expected to be an array. It was too generic for me to understand what the problem was so I started going into the Livewire vendor code to trace the issue. After "dd"ing  various arguments in the validate method, I noticed that the messages being passed in which were meant for the Validation factory, were MY chat messages. The error occurred because they were a collection and not an array. Had they been a normal array this would have gone on silently and I would be none the wiser.

So, anyway, this is obvious in retrospect, but I at least updated the docs to include this caveat in case it trips anyone else up like me. However, in this particular instance, "messages" _is_ fairly generic and it seems unfair that I can't use that for my own property. I also get that this PR would be a major breaking change for those using the messages property as intended. But I thought at the very least this could be useful as a discussion around how you name internal properties moving forward. 

Thanks!

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -79,10 +79,10 @@ trait ValidatesInput
         return [];
     }
 
-    protected function getMessages()
+    protected function getValidationMessages()
     {
-        if (method_exists($this, 'messages')) return $this->messages();
-        if (property_exists($this, 'messages')) return $this->messages;
+        if (method_exists($this, 'validationMessages')) return $this->messages();
+        if (property_exists($this, 'validationMessages')) return $this->validationMessages;
 
         return [];
     }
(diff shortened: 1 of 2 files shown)


calebporzio:
Yes, I agree that "messages" is a common noun in apps and by restricting it's use, I'm causing confusion and pain in those cases.

Fortunately, I don't think those cases are overwhelmingly common enough to break the API by changing "messages" to "validationMessages"

Something that if we could go back in time, maybe should have done, but now it's too breaking.

PR #2239, opened 2020-12-18
Fix error when new variable is present

striebwj:
This fixes an error that occurs when a new variable is added to an existing component.

When a user is on an old session when re-rendering the component happens it throws a 500 since the array offset does not exist in the `requestMemo['data']` object.


calebporzio:
Thanks! I'm open to this addition. Will just need a test covering this case (a test that fails without this PR but passes with it) to feel comfortable merging though.

calebporzio:
Closing this for now because of lack of response.

Please re-submit if you add those tests. Thanks for contributing!

PR #2245, opened 2020-12-20
[Update] Fixes wire:model.lazy syncInput action being called on subsequent requests

### LINKED ISSUE #2148: lazy inputs fire extra `syncInput` next time form is used if initial form is submitted while edited lazy input still has focus
### Description
When (re-)using a form (in a modal) to edit multiple models (e.g. transactions), when saving the form it fires `syncInput` events for `.lazy` inputs that were edited during the form's previous use(s) - *only* when the Save button is clicked when focus is still in the edited input (e.g. lazy hasn't fired yet on loss of the input's focus)

### Exact steps to reproduce
- using Surge as an example
- login to Surge
- open browser's devtools to Network
- in Surge's dashboard, click `Edit` for one of the transactions
- edit the `Title` field, *without* moving to another input or using the tab key etc.
- click `Save`, examine the payload in devtools for Save event (e.g. see screenshot below):
image
- now, click `Edit` for a different Transaction
- edit the `Amount` field *again, without leaving the field*
- click `Save`, examine payload in devtools (see below):
image

- what seems to be happening is, in the 2nd edit, Livewire sees both the `Title` *and* `Amount` fields as dirty when only the `Amount` field was edited.  Looks like it is remembering the `Title` field's edit from the first `Transaction's` edit.

- easiest way is to use the Surge repo and follow above steps

**Context**
- Laravel v8.16.1
- Livewire v2.3.4
- Alpine.js v2.7.3
- browsers tested = Brave, Chrome, Safari

I'd also like to make a suggestion - a method/function to enable the manual clearing of the 'dirty' state (and, therefore, I assume the related pending syncInput events) related to a Livewire component's attribute(s) used in a form (e.g. `public $editing` in Surge) might be useful.  It would enable a more thorough resetting between a form's usages (vs. `$this->reset($editing)` + `$this->resetValidation()`).

Thanks!

joshhanley:
UPDATE: I have added a dusk test, merged in master and built the assets.

Adding a listener seemed too complex.

So instead I am getting the updates straight from the request and outputting them to the screen and running checks against that. Hope that's ok, seemed the simplest solution!

Ran the test without the fix and it failed, reapplied the fix and it passed again, so should be all good to go!

Hope this helps!

---
Original:

This fixes issue #2148 where, if you have an input with wire:model.lazy on it, update the input and submit to the server, any subsequent requests to the server also contain the a syncInput action for that same input even if it hadn't been changed.

The reason was that the callback inside the callback register wasn't being cleared out correctly, as such it still contained the callback from the previous network request, which was causing it to fire when iterating over the register on the next request.

This fixes it so the callback is returned back to an empty callback after it has run.

Don't know how to test for this, sorry! Is there a way to test what actions are being sent to the backend so we can make sure only the correct actions are being sent? If so I'd be happy to put a dusk test and component together if there is a way.

Hope this helps!

Closes #2148


jeffchown:
Hey @joshhanley - just made your edit to my local Livewire fork and took it for a test-drive on Surge - lookin' good! 👍    
Then I tested the update on one of my app's modals and all looked good there too :)

Then, I decided to go for the gold and tested the update with all of my app's modals wrapped in `@if($isShow) ... @endif` blade directives (yes, each is within a <div></div> to be 'morph-safe' ;)... this is what originally brought the issue to my attention - I have multiple large modals for editing different models' data in my app (which is effectively an SPA) and I don't want the modals markup polluting my HTML (for a variety of reasons including security) except when a modal is in use. ...and... it worked!  No more 'old' queued `syncInput`s from previous uses of the modal firing and overwriting the wrong record's data!

This issue caused multiple issues and became a major frustration and limitation - no more!

Thank You! for your help!

(I'll be running extensive tests tomorrow to double-check/confirm and will let you know how they go)

joshhanley:
@jeffchown awesome! That's great to hear! Thanks for testing it 🙂

nuernbergerA:
@joshhanley would it be possible to add a dusk test?

joshhanley:
@nuernbergerA can you test the payload using dusk? If so, I'd be happy to add it, just wasn't sure how to do that with dusk 🙂

nuernbergerA:
you could register a listener `Livewire.hook('message.received', (message, component) => {})` like we did here https://github.com/livewire/livewire/blob/ba70e2ddb00cc45a7063b2fdf106c0f9aa83478a/src/Macros/DuskBrowserMacros.php#L128

then you can grab the payload from `message.response`

joshhanley:
@nuernbergerA thanks! I’ll check it out in the morning and see if I can get something running

calebporzio:
Thanks Josh! Great PR!

joshhanley:
No worries!

Issue #2000, opened 2020-11-09
Add per-component layout configuration


calebporzio:
This is a feature spec:

```php
class Foo extends Component
{
    protected $layout = ['layouts.dashboard', 'body'];
    // or
    protected $layout = ['view' => 'layouts.dashboard', 'slot' => 'body'];

    protected $extends = ['layouts.dashboard', 'body'];
    // or
    protected $extends = ['view' => 'layouts.dashboard', 'section' => 'body'];

    // Also add support for methods:
    protected function layout()
    {
        return [...];
    }

    protected function extends()
    {
        return [...];
    }
}
```

This will allow apps with multiple different layouts to be able to extract the configuration from individual component's `render()` method into whatever abstraction they choose.

For example, you could extract a trait called `use WithDashboardLayout`, or even a parent component like `extends DashboardComponent` that has the layout settings.

calebporzio:
I don't want to support setting the layout at the route level like Livewire V1, so this is an alternative to that solution that should provide mostly the same benefits.

jszobody:
What if this isn't about making it like Livewire v1, and more about being consistent with the vanilla Laravel routing features?

If this was something Livewire v1 had invented, I get it. Let's do it better. But isn't there a lot of value to leveraging as much of the built-in Laravel routing features as possible, and making the transition from normal blade views to LW components that much simpler?

calebporzio:
Please clarify @jszobody 

Laravel doesn't allow you to set any default layout for a blade view from a route.

Interested to know what you mean. Sample code would be great.

iksaku:
The implementation of this spec will help for sure! It would help me reduce "boilerplate" defining the layout and section (or slot) for every "dashboard" component page (for example).

Totally into this :+1:

joshhanley:
The other thing I think would be handy to support, would be not specifying a layout in the component class at all.

Since blade components, all my blade files have `<x-layouts.app></x-layouts.app>` tags wrapping the content in the .blade.php file and it would be nice to do the same with Livewire.

Not sure if that's even possible or will work though, but would be nice!

iksaku:
@joshhanley yeah, that would be the nicest way, but due to the way Livewire works for component reactivity, it is not possible ATM, unless we use some kind of custom blade diffing to identify the root view part to replace 🤔

joshhanley:
@iksaku yeah exactly, just thought I'd share the hopes and dreams 😁

adamthehutt:
[...]

PR #2246, opened 2020-12-21
Configurable default layout

adamthehutt:
Per #2000 and #2221, this PR adds support for an optional config setting to specify a default layout for full-page components (i.e. an alternative to "layouts.app").


calebporzio:
Thanks for this!

I'm moving this to #2247 because you PR'd your fork's master branch which locks it from me contributing. (in the future please take out a feature branch on your fork)

See you over there!

PR #2247, opened 2020-12-21
Make layout configurable

calebporzio:
This is a fork of #2246 because that PR was made from the `master` branch for a fork which means I can't push to it.

This will be the place to discuss PR #2246


calebporzio:
Hey @adamthehutt 

I made the following changes to your PR:
* renamed the config item to: 'livewire.layout' (cleaner IMO)
* refactored your tests a tiny bit

The reason I'm not just merging this is that config changes are fairly permanent.

My main question is: If we merge this right now, will we feel stuck in the future if we want to introduce config options for the default "slot" or "section"?

I lean towards just merging this because the default "$slot" is good enough, but just want to make sure we won't wish we would have done something like:

```
'layout' => [
    'view' => 'layouts.app',
    'slot' => 'default',
];
```

Or something like that. But that case is a little misleading, because "default" will come through as "$slot" and when a user changes it to something like "body", it will be "$body".

Anyhow. Again, I'm leaning towards what you've proposed just want to get other opinions on it.

Thanks!

adamthehutt:
Hi @calebporzio. I tend to prefer more verbose and explicit config options, so your suggestion makes sense to me. IMO it works fine with verbose comments, like:

```php
/**
 *  These options set defaults for rendering full-page Livewire components.
 */
'layout' => [

   /**
    *  Which view should Livewire use by default when rendering a full-page component? 
    */
    'view' => 'layouts.app',   

    /**
     *  Which slot in the layout view should Livewire use by default when rendering a full-page component?
     */
    'slot' => 'default',
];
```

The reason I keep specifying "full-page" (in the comments if not in the config keys) is that I'm concerned some users may be confused and assume this applies to "normal" components as well. The documentation uses the term "Full-Page Components", so I'm just going for consistency. Your call, obviously.

If "layout.slot" is available as a config option, it needs to be supported, right? How hard is that to add to the PR?

adamthehutt:
...aaaand I just saw the comments you added to the config file, which are even clearer and more explicit than mine, so you can ignore that part of my comment above. :)

joshhanley:
Copying my comment from the issue thread here too, I believe also worth exploring at the same time:

I also had a thought for those that might be using the old layout extends instead of components:

- Add a config option for 'layoutType' which can be 'extends' or 'component'.
- Or add an option 'useExtendsLayout` and is false by default, can be set to true.

Have a look at Component.php:51 where you can set it on the component or has a default of 'component', that default could be a config option as well.

PR #2251, opened 2020-12-22
Fix entangle consecutive actions

### LINKED ISSUE #2250: Checksum Throwing Error With Entangled Data
### Description

I am getting a checksum error (corrupt data) when I try to change entangled values.

### Exact steps to reproduce

I need dependent selectpickers. The contrived example is where you have states and cities. Select a state, and the `cities` update with just the cities associated to that state. Then you can choose multiple cities.

1. Use the code below
2. Click on 'Missouri'
3. Click on both 'St. Louis' and 'Kansas City'
4. Click on 'Texas' and see the error

https://user-images.githubusercontent.com/4503765/102845164-8b0cd800-43d2-11eb-8dc4-062b801e0657.mp4

```
class Debug extends Component
{
    public $states = [];
    public $cities = [];
    public $state;
    public $city;
    public $selectedCities = [];

    public function mount()
    {
        $this->states = collect([
            1 => 'Missouri',
            2 => 'Texas',
        ]);

        $this->cities = collect([
            1 => 'St. Louis',
            2 => 'Kansas City',
            3 => 'Dallas',
            4 => 'Houston',
        ]);
    }

    public function updatedState($value)
    {
        if($value == 1) {
            $this->cities = collect([
                1 => 'St. Louis',
                2 => 'Kansas City',
            ]);
        }
        if($value == 2) {
            $this->cities = collect([
                3 => 'Dallas',
                4 => 'Houston',
            ]);
        }
    }
}
```

```
<div x-data="{
        city: @entangle('city'),
        state: @entangle('state'),
        cities: @entangle('cities'),
        states: @entangle('states'),
        selectedCities: @entangle('selectedCities'),
    }">
    <h3>State (click on one)</h3>
    <template x-for="item in Object.entries(states)" :key="item[0]">
        <div x-text="item[1]" x-on:click="state = item[0]"></div>
    </template>
    <hr/>
    <h3>Cities (click to select multiple)</h3>
    <template x-for="item in Object.entries(cities)" :key="item[0]">
        <div x-text="item[1]" x-on:click="selectedCities.push(item[0])"></div>
    </template>
    <hr/>
    Selected Cities
    <template x-for="item in Object.entries(selectedCities)" :key="item[0]">
        <div x-text="item[1]"></div>
    </template>
</div>
```

**Context**
- Livewire version: 2.3.5
- Laravel version: 8.19
- Alpine version: 2.8
- Browser: Chrome

joshhanley:
This PR fixes #2250 where Alpine watchers are not firing when consecutive actions are applied to the same entangled property using Alpine.

I have added a test for this which runs consecutive actions from the Alpine end, and just to ensure it's robust, it then runs consecutive actions from the Livewire end.

The cause was a Livewire properties' value wasn't being deep cloned when assigned to the Alpine property in the Livewire watcher, as such any Alpine changes were mutating the Livewire property, which was then blocking the Alpine watchers and causing checksum issues on the next request.

Hope this helps!


calebporzio:
Great call Josh! Can't believe we missed the deep clone. Things should be much more stable now.

Only tiny concern is if the deep clone technique (stringify and parse) changes array orders or something that would break the checksum, but in this case, this data is for Alpine ONLY and shouldn't really leak back into the Livewire data right? Only through updates?

Mmmkay, I'm going to test this locally and merge - I'm sure it's great.

Reminder once again that you are the man.

iAmKevinMcKee:
Thanks @joshhanley I can't believe this got resolved so quickly. THANK YOU! 👏

joshhanley:
@calebporzio yep same! And yeah Alpine->Livewire changes should go through the set function so should all be ok.

@iAmKevinMcKee no worries! I've fixed a few issues in this area of the code base in the last couple of weeks, so my head was still fresh on it 😁

PR #2264, opened 2020-12-25
Fix Unable to call component method

mtawil:
When I do something like this `wire:click="@if (true) method1() @else method2() @endif"`

I got `Unable to call component method. Public method [ method1] not found on component`

Clearly the issue was because there is a space in the method name, so we need to trim it first.


calebporzio:
Thanks! I'll merge if there is a test included.

mtawil:
@calebporzio done

calebporzio:
Nice! Thanks!

PR #2279, opened 2020-12-30
Add the new turbo eval tags to scripts

tonysm:
### Added

- Adds the new `data-turbo-eval` for Turbo Drive (see the [documentation here](https://turbo.hotwire.dev/handbook/building#working-with-script-elements)). This PR has a companion PR on the `livewire/turbolinks` plugin (see [here](https://github.com/livewire/turbolinks/pull/12))

---

I haven't created an issue for it, but since Turbolinks is supported, I figure adding support for the new Turbo Drive would be OK.

I tested it out locally and it works fine persisting the component state across turbo visits (with the new livewire/turbolinks package)

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -218,8 +218,8 @@ HTML;
         // because it will be minified in production.
         return <<<HTML
 {$assetWarning}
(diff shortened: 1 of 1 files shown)


calebporzio:
Perfect! Thank you!

PR #2280, opened 2020-12-30
remove trailing question mark

timrspratt:
The PR ensures there is no `?` left at the end of the href.

The occurrence I've been having for this is when you navigate back to page 1 which removes the `page` parameter from the query string, resulting in an empty query string. Instead of `example.com/posts` you'd end up with the URL `example.com/posts?`.

Apologies, I don't have experience in writing the tests for this or building the assets. I've popped it on a branch so that it can be modified by others. Thanks

### DIFF
--- a/js/component/SyncBrowserHistory.js
+++ b/js/component/SyncBrowserHistory.js
@@ -90,7 +90,7 @@ export default function () {
 
         let destination = new URL(url)
 
-        let afterOrigin = destination.href.replace(destination.origin, '')
+        let afterOrigin = destination.href.replace(destination.origin, '').replace(/\?$/, '')
 
         return window.location.origin + afterOrigin + window.location.hash
     }


marispro:
I had the same issue and idea before, but eventually solved this problem with except. `?` does not appear anymore.

`protected $queryString = [
        'page' => ['except' => ''],
    ];`

https://laravel-livewire.com/docs/2.x/query-string

timrspratt:
Thanks @marispro. On reflection of my use case, it was actually when a default is `null` that `except` wasn't working. I've managed to solve this by changing the default to `false`, but I'm leaving this open as I feel it's a sensible addition, since the `path` returned in the Livewire message doesn't contain the `?`, it feels a bug that it's the JS that is adding the `?`.

calebporzio:
Definitely a good addition. Thanks @timrspratt ! That's bugged me in the past too!

PR #2288, opened 2021-01-02
Fix entangle array defer issue

joshhanley:
When an array value is entangled and deferred and the request is finally sent, if the server updates that value, it's not reflected in Alpine.

A use case for this is if you have a list of roles in a select, when a user selects a role, and want it to be deferred to save a request.
When the server finally gets the request it uses the new value (ie to save the role to a user) and then you reset it back to it's original value.
But Alpine wasn't showing the new value.

This PR fixes this issue.

Hope this helps!


calebporzio:
Thanks Josh!

joshhanley:
No worries!

Issue #2158, opened 2020-12-03
Livewire doesn't work with Laravel Swoole


3tew:
### Description
Every time the input is updated (wire:model="..."), this error always occurs.
Usually, I test the project by using the command `php artisan serve` on my local machine, but when I take my project to test it again with Laravel Swoole in the Docker container, there will always be this problem.

### Stripped-down, copy-pastable code snippets
```
[2020-12-04 02:03:19] production.ERROR: Livewire encountered corrupt data when trying to hydrate the [product.create-product] component. 
Ensure that the [name, id, data] of the Livewire component wasn't tampered with between requests. {"userId":1,"exception":"[object] (Livewire\\Exceptions\\CorruptComponentPayloadException(code: 0): Livewire encountered corrupt data when trying to hydrate the [product.create-product] component. 
Ensure that the [name, id, data] of the Livewire component wasn't tampered with between requests. at /app/vendor/livewire/livewire/src/HydrationMiddleware/SecureHydrationWithChecksum.php:21)
[stacktrace]
#0 /app/vendor/livewire/livewire/src/LifecycleManager.php(70): Livewire\\HydrationMiddleware\\SecureHydrationWithChecksum::hydrate(Object(App\\Http\\Livewire\\Product\\CreateProduct), Object(Livewire\\Request))
#1 /app/vendor/livewire/livewire/src/Connection/ConnectionHandler.php(12): Livewire\\LifecycleManager->hydrate()
#2 /app/vendor/livewire/livewire/src/Controllers/HttpConnectionHandler.php(13): Livewire\\Connection\\ConnectionHandler->handle(Array)
#3 /app/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(48): Livewire\\Controllers\\HttpConnectionHandler->__invoke('product.create-...')
#4 /app/vendor/laravel/framework/src/Illuminate/Routing/Route.php(254): Illuminate\\Routing\\ControllerDispatcher->dispatch(Object(Illuminate\\Routing\\Route), Object(Livewire\\Controllers\\HttpConnectionHandler), '__invoke')
#5 /app/vendor/laravel/framework/src/Illuminate/Routing/Route.php(197): Illuminate\\Routing\\Route->runController()
#6 /app/vendor/laravel/framework/src/Illuminate/Routing/Router.php(692): Illuminate\\Routing\\Route->run()
[...]

PR #2292, opened 2021-01-03
Resolve Swoole Compatibility and Increase Readability

tanmuhittin:
This PR resolves the issue #2158 and increases code readability.
This PR might accelerate the app by removing the bypassTheseMiddlewaresDuringLivewireRequests method in boot

Also thank you for this rockstar package!

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -67,16 +67,6 @@ class LivewireServiceProvider extends ServiceProvider
         $this->registerBladeDirectives();
         $this->registerViewCompilerEngine();
         $this->registerHydrationMiddleware();
-
-        // Bypass specific middlewares during Livewire requests.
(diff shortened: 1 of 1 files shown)


calebporzio:
Woah, "excluded_middleware" with the route group is SOOO much better than what I had. THANKS!

calebporzio:
Hey @tanmuhittin  - So I actually reverted this PR merge because the test suite was broken.

Because this change will fundamentally effect everything I'm backing it out for now.

If you can get the test suite passing, I'd welcome a new PR. Thanks!

jihadismail8:
Swoole can dramatically change the future of Livewire , Swoole does increase speed of PHP apps by at least x5 , if Swoole will be supported in Livewire , the server side rendering will become faster and more smooth , i think it should be your next target to support Swoole in livewire apps , as it will make livewire shine <3.

PR #2299, opened 2021-01-04
fixed method call

### LINKED ISSUE #2286: Testing: Non-static method Facade\Ignition\Views\Engines\PhpEngine::handleViewException() should not be called statically
I have a test that used to work, that no longer works.  This seems to have started since upgrade my Livewire/Laravel to latest versions.

Here is my test:

    public function testLivewireCreateBusiness()
    {
        $livewire = Livewire::test(\App\Http\Livewire\Business\Create::class);
        $livewire->lastResponse->assertStatus(403);
    }

**Context**
- Livewire version:  dev-master
- Laravel version: 8.20.1
- Phpunit: 9.5.0

Gives error:

```
ErrorException : Non-static method Facade\Ignition\Views\Engines\PhpEngine::handleViewException() should not be called statically
```

austenc:
Related to https://github.com/livewire/livewire/pull/2160

This fixes the static method call (the actual method is not static).

Should also fix https://github.com/livewire/livewire/issues/2286.

### DIFF
--- a/src/CompilerEngineForIgnition.php
+++ b/src/CompilerEngineForIgnition.php
@@ -14,7 +14,7 @@ class CompilerEngineForIgnition extends CompilerEngine
     protected function handleViewException(Throwable $e, $obLevel)
     {
         if ($this->shouldBypassExceptionForLivewire($e, $obLevel)) {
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

PR #2301, opened 2021-01-04
Fix entangle equality check

joshhanley:
Due to the deep clone introduced in PR #2251 the current check to see if Alpine and Livewire properties values are the same, will no longer work as they are not referencing the same thing.

This is causing Alpine to send updates back to Livewire after Livewire makes changes.

This PR fixes that issue.

So it's a bit rough naming wise, but thought I'd get it up.

Hope this helps!


calebporzio:
Thanks so much @joshhanley !

joshhanley:
No worries!

PR #2302, opened 2021-01-05
Revisited Resolve Swoole Compatibility and Increase Readability

tanmuhittin:
related to #2292 

Looks like I forgot to exclude the middlewares from test endpoint /livewire-dusk/{component}

I hope this will pass the tests

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -108,29 +108,34 @@ class LivewireServiceProvider extends ServiceProvider
 
     protected function registerRoutes()
     {
+        // Bypass specific middlewares during Livewire requests.
+        // These are usually helpful during a typical request, but
+        // during Livewire requests, they can damage data properties.
+        $excludedMiddlewares = [
+            'middleware' => config('livewire.middleware_group', 'web'),
+            'excluded_middleware' => [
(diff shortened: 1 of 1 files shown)


joshhanley:
@tanmuhittin you can run the tests locally, they are using phpunit and dusk. Just run `vendor/bin/phpunit` to run the whole test suite, so you can ensure you fix any issues before you push it up.

You may run into an issue after you install dusk like this
> Facebook\WebDriver\Exception\SessionNotCreatedException: session not created: Chrome version must be between 70 and 73

Here is the fix for that, run:
```
./vendor/bin/dusk-updater detect --auto-update
```

This is the reference in the testbench-dusk docs for that fix:
 https://github.com/orchestral/testbench-dusk#chrome-versions

Hope this helps!

tanmuhittin:
I had hard time running browser tests in my docker image.

Can anyone advise an appropriate docker image for browser tests?

tanmuhittin:
I finally got the browser tests running. Looks like excluding middleware feature is not working for global middlewares yet. So, this should not be merged. I hope this feature would be added to Laravel in the future.

Ref : https://github.com/laravel/framework/pull/32404#issuecomment-614808543

PR #2305, opened 2021-01-05
Update phpunit.xml configuration format

mtawil:
PHPUnit 9.3.0 has changed coverage filtering from filter tag to coverage tag. So, this pull request migrates `phpunit.xml` to the new configuration format.


calebporzio:
Thanks!

PR #2306, opened 2021-01-05
Headless browser test

mtawil:
I am using this line manually every time I run browser tests, and I think we should stop watching a browser doing some magic automatically 😅

### DIFF
--- a/tests/Browser/TestCase.php
+++ b/tests/Browser/TestCase.php
@@ -158,6 +158,8 @@ class TestCase extends BaseTestCase
     {
         $options = DuskOptions::getChromeOptions();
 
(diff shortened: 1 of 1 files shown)


mtawil:
I forget to mention that the browser test with a `--headless` option is ~1.25x faster than without it.

nuernbergerA:
Hey @mtawil 

There is actually a manual toggle to test with or without UI [here](https://github.com/livewire/livewire/blob/master/tests/Browser/TestCase.php#L32)

During CI tests we run without UI (headless).

mtawil:
Hi @nuernbergerA

Okay, didn't notice this toggle line!

Thank you!

Issue #2308, opened 2021-01-05
500 error: This action is unauthorized.


AdvancedStyle:
In previous versions of Livewire authorize worked like this:

```
class MyClass extends \Livewire\Component
{
    use AuthorizesRequests;

    public function mount()
    {
        $this->authorize('some-permission');
    }
}
```

Would return a **403** code and show forbidden page.

It the current dev-master branch it instead throws an exception with "This action is unauthorized." and so is considered a 500 error.

To prevent it you'd have to do something like this:

```
class MyClass extends \Livewire\Component
{
    use AuthorizesRequests;

    public function mount()
    {
        try {
            $this->authorize('some-permission');
        }catch(Exception $e){
            abort(403);
        }
    }
}
```

This looks a mess and also prevents you through throwing 'real' exceptions (should they happen to occur).

**Context**
- Livewire version: dev-master (d2ed682)
- Laravel version: 8.20.1

## Note

It looks like this was previously addressed, but must have been removed at some point in the current repo: https://github.com/livewire/livewire/pull/843

joshhanley:
Yep can confirm this is happening. Before with app debug on or off, both would should this
![image](https://user-images.githubusercontent.com/882837/103695021-f9b37d80-4fef-11eb-84fd-a2dc12f36db9.png)

But running on dev-master:

It's interesting, with app debug on, it shows the correct exception
![image](https://user-images.githubusercontent.com/882837/103694915-d1c41a00-4fef-11eb-973c-83dad8919a7e.png)

But with app debug off, it's throwing a 500 error
![image](https://user-images.githubusercontent.com/882837/103694951-de487280-4fef-11eb-8045-dc78242816fc.png)

austenc:
Hey @AdvancedStyle @joshhanley does the route you're testing this with have the `auth` middleware applied by chance? Any other details that might help me to replicate this?

I haven't been able to replicate it, but hopefully I'm just missing something simple. I did see your PR @AdvancedStyle but I think that may break some of Ignition's data collection features in debug mode.

AdvancedStyle:
Yes the routes have `['auth:sanctum', 'verified']` middleware.

However the problem also occurs without using a route at all.  Using livewire tests also triggers the Exception inside of a 403 status code.

For example here is a test (`Transaction` class has an `authorize()` that causes this to be denied):

    public function testLivewireAccountTransactionsAuthorization()
    {
[...]

PR #2309, opened 2021-01-06
Use Illuminate\View\Engines\PhpEngine instead of Facade\Ignition\View…

AdvancedStyle:
…s\Engines\PhpEngine in order to correctly show authorization errors.

It fixes this problem: https://github.com/livewire/livewire/issues/2308

No, it only contains a fix for 2308

Does not include tests

It fixes a problem of authorization exceptions in Livewire not being handled.

### DIFF
--- a/src/CompilerEngineForIgnition.php
+++ b/src/CompilerEngineForIgnition.php
@@ -2,7 +2,7 @@
 
 namespace Livewire;
 
-use Facade\Ignition\Views\Engines\PhpEngine;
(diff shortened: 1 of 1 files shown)


austenc:
Hey @AdvancedStyle, thank you for the contribution and for catching this. 

Initially, I was concerned that making this change would disable some ignition features. However, after investigating I found that all the view exceptions were still behaving as expected with Ignition, so I approve of this change. 

Thanks again for the PR!

calebporzio:
Thanks!

PR #2314, opened 2021-01-06
redirect before even render the default layout

nuernbergerA:
As mentioned by @AdvancedStyle [here](https://github.com/livewire/livewire/pull/2223#issuecomment-751668193)

Redirect doesn't work if you use a `redirect()` within `mount()` in Full-Page components.

I don't think this is related, but I noticed that the target URL isn't always the same
| Call        | Url           |
| ------------- |-------------|
|  `$this->redirect('/foo')` | `/foo` |
|  `redirect('/foo')` | `http://localhost/foo` |

therefore I used `assertStringEndsWith()`

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -48,6 +48,10 @@ abstract class Component
             ->mount($componentParams)
             ->renderToView();
 
+        if ($this->redirectTo) {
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks Adrian!

PR #2315, opened 2021-01-06
Add set slot method for view object

GertjanRoke:
Hello Caleb,

**Short story:** I wanted one method to fill a slot of a component without also giving the same layout every time.

**Longer story:**
I was setting up a new Laravel project with Jetstream and the new layouts.app layout has this `$header` slot for setting the page title in the navbar, but if you want to set this slot you would need to do something like this: `->layout('layouts.app', ['header' => 'Users'])`.
But after building five components you now have five locations where you are setting the same layout file over and over again.  
This felt a bit nasty for me, so that why I want to add this `->setSlot('header', 'My Page')` to make this better looking and also not having to give the same layout file over and over again.

**Extra info:**
I also added tests for all methods in the `ViewMacros` class including the method I added because there were no tests for it yet.  
Also one failing test, I can't find out why the test about the custom slot can't pass, I followed the docs but maybe I'm doing something wrong or I found a "feature". See test: `ComponentLayoutTest -> can_set_custom_slot_for_a_layout`.

If there is something I should change or can make more clear please let me know and don't feel bad if you give this MR a #noPlansToMerge, no feelings will be hurt. ❤️


joshhanley (changes requested):
@GertjanRoke this looks useful! I can also imagine it being used to set the `$title` of the page on an initial render.

There are a couple of thing you could do to improve this PR.

First one is to rebase all of your changes to the latest master branch. Currently you are using a version from October and as such its ~170 commits behind.

The other thing, I'm not sure why are you calling `Livewire::component` and then using route get, you could just use `Livewire::test`.

So for example, the test that is failing, you could change it to

```php
/** @test */
public function can_set_custom_slot_for_a_layout()
{
    Livewire::test(ComponentWithCustomSlotForLayout::class)
              ->assertSee('baz');
}
```

Hope this helps!

GertjanRoke:
@joshhanley oh I have no idea why my code is so behind.. I will look into that.

I think I tried with the test but didn't seem to work but I will give it a try and see if it works.

Thanks for the feedback 🙏

GertjanRoke:
@joshhanley I tried your way of testing but it doesn't seem to load the layout file then.
Is there a way I can trigger it to also load the layout file?
I think this is done on purpose but not really handy for my tests.

joshhanley:
@GertjanRoke yeah you are right actually it doesn't load the layout which I thought it would, so your original method is fine 🙂

I have dug into why that one test fails, and I think the `->slot` method doesn't actually work 😂 clearly no-one actually uses that!

I was originally thinking it might be better to split this PR into two, one for all the views tests and then another (this one) for your set slot method and the tests just for that.

If you have the time, are you able to do that by creating a PR just with the view tests (and remove them from this one), so we can demonstrate that there is an issue with the `->slot` method and then we can get that fixed first before considering this one?

GertjanRoke:
@joshhanley good idea! I will create the new PR, that's no problem.

GertjanRoke:
@joshhanley I created that new PR to fix the slot method and removed all changes that are not related to this PR.

austenc:
@GertjanRoke if you update your branch with the latest master (and resolve the conflict), I think your tests will pass.

calebporzio:
Thanks for the PR @GertjanRoke, however this functionality already exists by passing a second param of arguments to the `layout` method.

```php
return view(...)->layout(..., ['foo' => 'bar'])
```

Because $slots in layouts are just view variables, a `$header` named slot would also just accept a `'header'` argument as the second param of the layout.

I tested that this works as I described.

I'm guessing this behavior is not well documented. I'm going to close this PR, but would you be willing to PR a writeup to the livewire/docs repo? Thanks!

GertjanRoke:
Hey @calebporzio, sure no problem! Thanks for taking your time to check for another way 👏🏻

GertjanRoke:
Hey @calebporzio,
Sorry I didn't read your message comment correctly.

But like I stated in my first message when making a new Laravel installation and using Jetstreams app layout this is getting a pain in the butt.
Because when I later decide to use a different layout for 50% of my Livewire routes I then need to update all those components just because I'm forced to also give the layout file.

So that's why I thought with adding this method I can avoid telling each component what the layout file is.
I also completely understand that you don't want to maintain something you don't feel it's useful.

On a side note, I think it's already pretty clear in the docs that you can set it from the layout function.

Issue #2145, opened 2020-12-02
Lazy binding does not work as expected for input, checkbox, select


ERFOLGswerkzeug:
### Description
Data binding does not work as expected for input, checkbox, and select multiple using the `.lazy` modifier.
Clicking away from a text input directly into a checkbox or a select multiple does not bind the value of the text input to the underlying property. The text input field is cleared instead.

### Exact steps to reproduce
1. Create a text input using `wire:model.lazy` modifier, bind it to a property
2. Create e.g. a checkbox or a select multiple with using `wire:model.lazy`, bind it to a property
3. Input text into the text input and check the checkbox immediately, without focusing any other element in between
4. The checkbox value is bound, the text input will be cleared or at least flicker

### Stripped-down, copy-pastable code snippets

Component:
```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;

class Form extends Component
{
    public $title;

    public $enabled;

    public $selected;

    public function mount()
    {
        $this->enabled = false;
        $this->selected = [];
    }

    public function render()
    {
        return view('livewire.form');
    }
}
```

Template:
```blade
<div>
    <div>
        Title:
        <input type="text" wire:model.lazy="title">
    </div>
    <div>
        Enabled:
        <input type="checkbox" wire:model.lazy="enabled">
    </div>
    <div>
        Options:
        <select wire:model.lazy="selected" multiple>
            <option value="A">Option A</option>
            <option value="B">Option B</option>
            <option value="C">Option C</option>
        </select>
    </div>
    <hr>
    <div>
        Title: {{ $title }} <br>
        Enabled: {{ $enabled }} <br>
        Options: {{ json_encode($selected) }}
    </div>
</div>

```

**Context**
- Livewire version: 2.3.5
- Laravel version: 8.17.1
- Alpine version: -
- Browser: Chrome, Firefox, Safari

jeffchown:
I can confirm.  Very strange.  When I clicked on the checkbox, Livewire fired two events but in the reverse order I would have expected.  I expected:
- 1st event to `syncInput` the `title` text input
- 2nd event to `syncInput` the `enabled` checkbox

Instead, they fired per the screenshots below - losing/clearing the `title` input in the process:
![image](https://user-images.githubusercontent.com/42819001/100904598-2b877080-3495-11eb-9700-e2990c99dc40.png)

[...]

PR #2340, opened 2021-01-12
Add failing test for lazy input and checkbox issue

joshhanley:
This is a failing test that demonstrates issue #2145.

What happens is if you enter text into an input that has `.lazy` on it, then select a checkbox, the checkbox request is sent first, then moments later another request is sent with the input update, which causes the input to flash.

If I get a chance later I will investigate it to see if I can work out what's going on.

But I suspect it should actually batch the two up together in the correct order, like the other lazy test (in this test file) when typing into an input and clicking submit, but it's not.

Hope this helps!


calebporzio:
Thanks for submitting this @joshhanley.

After digging in, we realized that the safari autofill hack we introduced a while back was causing weird issues like this one.

We've changed our autofill support strategy to something less invasive that made these tests pass and actually fixed some autofill in safari bugs.

Thanks a ton!

joshhanley:
No worries 🙂

Ah right!

leitom:
This introduced the issue in #2612

PR #2341, opened 2021-01-12
Slot method on view class is not working

GertjanRoke:
This PR was made because the `->slot()` method of the view is not working and this came to mind when I was adding a new feature and saw no tests were written for the view methods Livewire adds to the view class.

See this PR for more info: https://github.com/livewire/livewire/pull/2315


GertjanRoke:
Well @joshhanley found the fix! Thanks for helping!

joshhanley:
The test failure seems to be a GitHub actions issue, connection was refused on the WebDriver for some of the tests in PHP 7.3. The rest of the PHP versions ran fine. Another PR today had the same failure but for PHP 7.2.

calebporzio:
Thanks!

PR #2342, opened 2021-01-12
add small hint to generated view

thyseus:
add small hint to generated view of livewire component to not remove
the <div> element that has been generated.

### DIFF
--- a/src/Commands/livewire.view.stub
+++ b/src/Commands/livewire.view.stub
@@ -1,3 +1,6 @@
 <div>
+    {{-- Please ensure there is always only one root element inside
+         this view file for livewire to function properly. --}}
+
     {{-- [quote] --}}
 </div>


joshhanley:
I like the idea of the comment, my only issue with it is the root element can be any element, not just divs. So maybe something more generic, like "Ensure there is only one root element", short and to the point 🙂 

Also ignore the failing tests, not sure what is going on with GitHub actions, I saw the same failure on a different PR yesterday but for PHP 7.3 instead of 7.2 like on this PR.

thyseus:
Commit is amended and force pushed. Please check the wording again.

calebporzio:
I think we should handle this on the front-end in the form of a console warning, not in a comment in the component for aesthetic reasons.

Closing this one. Feel free to re-PR what I've suggested.

Issue #2344, opened 2021-01-12
temporarySignedRoute is hardcoded to 5 minutes and causes requests to return 401 


FrancisMawn:
### Description
When uploading large videos files, we noticed a strange behaviour where requests returned 401 instead of 200. After some investigation, it appears that the problem is caused by the handle logic in the `FileUploadHandler` class.

Digging a bit more, I found this code in the `GenerateSignedUploadUrl` class:
```
public function forLocal()
{
    return URL::temporarySignedRoute(
        'livewire.upload-file', now()->addMinutes(5)
    );
}
```

This invalidates any uploads that would require more than 5 minutes. There are currently no ways to override this behaviour (as far as I know). Manually changing the value to 30 minutes fixed the issue, but that's hardly a solution. Would it be possible to make this value configurable just like we have configurations for maximum file size? Maybe something like:

```
'temporary_file_upload' => [
   'max_upload_time' => 5,
];
```

### Exact steps to reproduce
1. Upload a file that would take more than 5 minutes to upload
2. Request returned a 401 status code. 

### Stripped-down, copy-pastable code snippets

**Context**
- Livewire version: 2.3.0
- Laravel version: 8.11.2
- Browser: Chrome

joshhanley:
@FrancisMawn yeah my opinion would be that it should be configurable. Any chance you could submit a PR for that config option?

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

PR #2347, opened 2021-01-13
Add max_upload_time configuration

FrancisMawn:
This PR addresses the issue #2344.

When uploading a large files, if the upload takes more than 5 minutes, the request gets invalidated and returns a 401. This PR adds a `livewire.temporary_file_upload.max_upload_time` configuration.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -90,7 +90,7 @@ return [
             'mov', 'avi', 'wmv', 'mp3', 'm4a',
             'jpg', 'jpeg', 'mpga', 'webp', 'wma',
         ],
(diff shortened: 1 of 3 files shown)


skoontastic:
+1 for this. I'm running into the same issue.

austenc:
@FrancisMawn thanks for making this PR. 

It might be better to specify the default (instead of `null`) in the config file.

Also, if you pass a second parameter to the `config` helper, you can avoid the ternary shorthand. For example:
`config('livewire.temporary_file_upload.max_upload_time', 5)`. 

If you specify the default in Livewire's main config file though, you probably don't need the second parameter fallback, just use what is configured.

calebporzio:
Perfect! Thanks!

PR #2350, opened 2021-01-13
initialize with an empy div to ensure initial render redirects for ne…

### LINKED ISSUE #2348: Redirect Not Working in 2.3.6
### Description

I just upgraded to 2.3.6 and my redirect doesn't work. The view does not render and the redirect does not happen.

### Exact steps to reproduce

Create a new laravel project, install livewire, create a component, put `$this->redirect('https://laravel-livewire.com')` in the mount method on the component. Put the component on a page and load the browser.

```
public function mount()
    {
        $this->redirect('https://google.com');
    }
```

One other thing to note is that I've tried to `dd` in the `PerformsRedirects` class and it never even gets there.

**Context**
- Livewire version: 2.3.6
- Laravel version: 8.22.1
- Alpine version: N/A
- Browser: Chrome

nuernbergerA:
Livewire sill needs a dom element to initially render a component.

Therefore we still skip rendering from the component but add an empty div so livewire can initialize the component properly to then redirect.

---
Fixes #2348

### DIFF
--- a/src/RenameMe/SupportRedirects.php
+++ b/src/RenameMe/SupportRedirects.php
@@ -40,6 +40,7 @@ class SupportRedirects
             }
 
             $response->effects['redirect'] = $component->redirectTo;
+            $response->effects['html'] = $response->effects['html'] ?? '<div></div>';
         });
     }
 }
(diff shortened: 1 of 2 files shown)


nuernbergerA:
Fun-fact its all green in my repo https://github.com/nuernbergerA/livewire/runs/1696292312

joshhanley:
@nuernbergerA not sure what's going on with GH actions, some PR's are passing on a few of the PHP version suites then breaks on one randomly and it's different every time.

calebporzio:
Thanks Adrian!

PR #2352, opened 2021-01-13
Make public component methods available in view

inxilpro:
Right now public properties are automatically made available to Blade (i.e. `$this->foo` is `$foo` in your Blade template). This PR adds public methods, as well, so that `public function isFooAllowed()` is available as `$isFooAllowed()` in your Blade template.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -11,7 +11,6 @@ use Illuminate\Support\Traits\Macroable;
 use Illuminate\Contracts\Container\Container;
 use Livewire\Exceptions\CannotUseReservedLivewireComponentProperties;
 use Livewire\Exceptions\PropertyNotFoundException;
-use ReflectionMethod;
 
 abstract class Component
 {
--- a/src/Component.php
+++ b/src/Component.php
@@ -11,6 +11,7 @@ use Illuminate\Support\Traits\Macroable;
 use Illuminate\Contracts\Container\Container;
 use Livewire\Exceptions\CannotUseReservedLivewireComponentProperties;
 use Livewire\Exceptions\PropertyNotFoundException;
+use ReflectionMethod;
 
 abstract class Component
 {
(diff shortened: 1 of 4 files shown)


ryangjchandler:
@inxilpro For the time being, you can call methods on `$this`, i.e. `$this->isFooAllowed()`.

calebporzio:
@inxilpro - this is so sick. Definitely something I've wanted, but didn't think it was possible. Nice work!

jameslkingsley:
This is a breaking change - if you have the following setup for example:

```php
public function resources()
{
    return [];
}

public function render()
{
    return view('template', ['resources' => $this->resources()]);
}
```

calebporzio:
@inxilpro - I had to revert this PR as the latest release broke people's apps. Happy to merge if you can make it non-breaking and add tests for those cases. Thanks!

imliam:
Honestly, even if it's non-breaking (by making properties take priority over public methods) IMHO I still don't like it. It leads to a confusing DX where you're not sure at a glance if a variable is a property value or an invokable method without looking it up.

Sure Blade components have this, but [they suffer from the same problems](https://github.com/laravel/ideas/issues/2129).

danharrin:
1+ to Liam's evaluation, I feel that the differentiation in syntax is very unclear.

calebporzio:
Ok cool, thanks for writing that up @imliam, I'll hold off on it then.

inxilpro:
I agree with @imliam that methods shouldn't be invoked without parentheses. This implementation doesn't bring that functionality over from Blade components.

I don't feel particularly strongly about this feature—it just felt right. If public class attributes are available without `$this` then it seems like public methods should be as well.

I think it'd be simple enough to PR a version that handles the conflicts better. Honestly, I would expect the following to work but it doesn't for the same reason:

```php
public $resources = [];

public function render()
{
    return view('template', ['resources' => $this->resources->map(/* do some transformation */)]);
}
```

It feels like view data explicitly set in the render method should take precedence over Livewire in general.

If @calebporzio agrees I'll submit the PR. Otherwise I think that using `$this` is probably fine. It just falls a little short of the "it just works the way you expect" ethos, IMHO.

danharrin:
> Otherwise I think that using $this is probably fine. It just falls a little short of the "it just works the way you expect" ethos, IMHO.

To be perfectly honest, as someone who is not familiar with the Blade alternative $call(), this is not the way I would "expect it to work". I'd expect to just call().

PR #2368, opened 2021-01-16
Resolve only valid page number or set default 1

p-andrey:
If you specify the 'page' parameter in the request with a string (for example 'foo'), livewire does not filter the parameter. And when paginating (i use $collection->forPage($this->page, 10)), an error appears.

Added check of the 'page' parameter in the resolvePage() method.
I also added a check for the 'page' parameter in the SupportBrowserHistory because it was overwriting the property set by the initialization of the WithPagination trait.

2 tests have been written that check the correctness of setting the page number.

Thanks for contributing! 🙌

### DIFF
--- a/src/RenameMe/SupportBrowserHistory.php
+++ b/src/RenameMe/SupportBrowserHistory.php
@@ -2,12 +2,13 @@
 
 namespace Livewire\RenameMe;
 
-use Livewire\Livewire;
-use Livewire\Response;
-use Livewire\Component;
-use Illuminate\Support\Arr;
 use Illuminate\Http\Request;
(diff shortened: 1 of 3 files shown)


p-andrey:
@calebporzio Hi.

Can you check why in my fork all tests was passed https://github.com/p-andrey/livewire/actions/runs/495943008 but in this PR tests fails? 
Maybe some problems in your check limits?

calebporzio:
Thanks for submitting @p-andrey - however, in the interest of keeping pagination related code from sprawling throughout the codebase I'm going to close this. I also haven't heard others mention this, AND I personally would expect a breakage if an expected value was manipulated. Thanks again!

PR #2372, opened 2021-01-17
Allow polling to be disabled outside viewport

ju5t:
This PR adds a polling modifier that limits polling when an element is not in the current viewport. It is used like this:

```blade
<div wire:poll.visible>
    {{ now() }}
</div>
```

The `visible` modifier tells Livewire to only poll visible elements. If the element is not visible, for example when a user has scrolled it out of view, the element is only being refreshed 5% of the time.

Another example, in case you want to change the polling frequency:

```blade
<div wire:poll.visible.1000ms>
    {{ now() }}
</div>
```

This solves discussion [2233](https://github.com/livewire/livewire/discussions/2233). It doesn't come with tests yet as I couldn't think of how to make them for this feature. Any help on tests (if possible) would be appreciated!

### DIFF
--- a/js/component/Polling.js
+++ b/js/component/Polling.js
@@ -56,8 +56,9 @@ function fireActionOnInterval(node, component) {
             if (Math.random() < .95) return
         }
 
-        // Don't poll if the element is not in the viewport, unless `keep-alive` is used
-        if (! inViewPort(directive.el) && ! directive.modifiers.includes('keep-alive')) {
+        // Only poll visible elements - effectively this prevents 95% of the requests
+        // if the element is not in the viewport
+        if (! inViewPort(directive.el) && directive.modifiers.includes('visible')) {
             if (Math.random() < .95) return
         }
 
--- a/js/component/Polling.js
+++ b/js/component/Polling.js
@@ -56,7 +56,7 @@ function fireActionOnInterval(node, component) {
             if (Math.random() < .95) return
         }
 
-        // Don't poll if the element is not in the viewport if "wire.poll.disable-hidden" is attached
+        // Don't poll if the element is not in the viewport, unless `keep-alive` is used
         if (! inViewPort(directive.el) && ! directive.modifiers.includes('keep-alive')) {
             if (Math.random() < .95) return
         }
--- a/js/component/Polling.js
+++ b/js/component/Polling.js
@@ -56,13 +56,13 @@ function fireActionOnInterval(node, component) {
             if (Math.random() < .95) return
         }
 
+        // Don't poll if the element is not in the viewport if "wire.poll.disable-hidden" is attached
+        if (! inViewPort(directive.el) && ! directive.modifiers.includes('keep-alive')) {
+            if (Math.random() < .95) return
+        }
+
         // Don't poll if livewire is offline as well.
         if (store.livewireIsOffline) return
-
-        // Don't poll if the element is not in the viewport if "wire.poll.disable-hidden" is attached
-        if (directive.modifiers.includes('disable-hidden') && ! inViewPort(directive.el)) {
-            return
-        }
 
         component.addAction(new MethodAction(method, directive.params, node))
     }, interval);
--- a/js/component/Polling.js
+++ b/js/component/Polling.js
@@ -1,5 +1,5 @@
 import MethodAction from '@/action/method'
-import { wireDirectives} from '@/util'
+import { wireDirectives, inViewPort } from '@/util'
 import store from '@/Store'
 
 export default function () {
@@ -58,6 +58,11 @@ function fireActionOnInterval(node, component) {
 
         // Don't poll if livewire is offline as well.
         if (store.livewireIsOffline) return
+
+        // Don't poll if the element is not in the viewport if "wire.poll.disable-hidden" is attached
+        if (directive.modifiers.includes('disable-hidden') && ! inViewPort(directive.el)) {
+            return
+        }
 
         component.addAction(new MethodAction(method, directive.params, node))
     }, interval);
--- a/js/util/inViewPort.js
+++ b/js/util/inViewPort.js
@@ -1,0 +1,10 @@
+export function inViewPort(element) {
+    var bounding = element.getBoundingClientRect();
+
+    return (
+        bounding.top >= 0 &&
+        bounding.left >= 0 &&
+        bounding.bottom <= (window.innerHeight || document.documentElement.clientHeight) &&
+        bounding.right <= (window.innerWidth || document.documentElement.clientWidth)
+    );
+}
--- a/js/util/index.js
+++ b/js/util/index.js
@@ -4,6 +4,7 @@ export * from './wire-directives'
 export * from './walk'
 export * from './dispatch'
 export * from './getCsrfToken'
+export * from './inViewPort'
 
 export function kebabCase(subject) {
     return subject.replace(/([a-z])([A-Z])/g, '$1-$2').replace(/[_\s]/, '-').toLowerCase()


ryangjchandler:
@ju5t Would something like `wire:poll.visible` sound better, removes the need to hyphenate the modifier name.

ju5t:
> @ju5t Would something like `wire:poll.visible` sound better, removes the need to hyphenate the modifier name.

@ryangjchandler `wire:poll.visible` sounds like it only polls when the element is visible to me. It does the opposite: it doesn't poll when the modifier is set and the element is not in the viewport. The current default behaviour is to always poll.

There might be another solution but I think that needs some more feedback. Livewire disables most polling requests if the tab is not active. You can disable this by doing `wire:poll.keep-alive`. This is an existing feature.

This could do the same. Disable polling if the element is not in the viewport by default and accept the same `keep-alive` parameter to enable it. That would make things more consistent, but comes at the price of changing default behaviour.

@calebporzio @joshhanley would you mind sharing your thoughts on this too?

ryangjchandler:
> > @ju5t Would something like `wire:poll.visible` sound better, removes the need to hyphenate the modifier name.
> 
> @ryangjchandler `wire:poll.visible` sounds like it only polls when the element is visible to me. It does the opposite: it doesn't poll when the modifier is set and the element is not in the viewport. The current default behaviour is to always poll.
> 
> There might be another solution but I think that needs some more feedback. Livewire disables most polling requests if the tab is not active. You can disable this by doing `wire:poll.keep-alive`. This is an existing feature.
> 
> This could do the same. Disable polling if the element is not in the viewport by default and accept the same `keep-alive` parameter to enable it. That would make things more consistent, but comes at the price of changing default behaviour.
> 
> @calebporzio @joshhanley would you mind sharing your thoughts on this too?

Oh right. Based on the discussion you mentioned in the PR description, it seems like the solution is to only poll when the element is visible, hence the modifier.

joshhanley:
I agree with Ryan in terms of naming, but maybe I'm missing something. What's the difference between only polling when the element is within the view port, vs disabling polling when the element is not in the view port?

But to me both names, disable-hidden or visible, to me, both suggest actual visibility. I'm not exactly sure how it works, but if an element is within the screen bounds but, for example, has tailwind hidden class applied, does getBoundingClientRect return a position? If so would it poll or not?

Maybe it needs to be something like `wire:poll.in-view` so it polls only when the element is in the view port?

ju5t:
> What's the difference between only polling when the element is within the view port, vs disabling polling when the element is not in the view port?

Its functionality is the same. However, the current default (before this PR) functionality is to always poll, with some exceptions:

- If the tab is in the background it prevents 95% of the poll requests, unless you specify `wire:poll.keep-alive`. With `keep-alive` it will also poll if the tab is in the background.
- If Livewire is offline.

From my perspective, `in-view` or `visible` imply that it is not polling at all unless you specify `in-view` or `visible`, as this is not the default behaviour of the current polling mechanism, it kind of feels wrong? Not sure.

@joshhanley thanks for pointing me to hidden elements. This PR would currently still poll those. I'm not sure if we should make that part of this feature. It's not that difficult to implement. All values of `getBoundingClientRect` will be `0` if the element is hidden. It does not matter if it's Tailwind or plain CSS.

I'm leaning towards rewriting this so its functionality is exactly the same as the background tabs. This means by default, if the element is not in the viewport, it should prevent 95% of the poll requests unless you specify `keep-alive`. I think this would be the best solution, even though it changes default functionality.

Thanks for thinking out loud @ryangjchandler and @joshhanley. Always appreciated 👍

boynet:
is it good idea to use getBoundingClientRect for that?
maybe  Intersection Observer API will be better for performance?


anyway I believe the polling method should not care about visibility, lets keep it simple.

the visibility detection should live has its own method, and it need a way to stop\start the polling.

maybe something like( I am new in livewire so not sure its the right way of stopping it)

```
<div @if(IAmVisible) wire:poll @endif  wire:visible="IAmVisible">
    {{ now() }}
</div>
```

the idea is that wire:visible updates a variable to the current state, that way you can use the functionality to other stuff like adding\removing class etc..

ju5t:
@boynet I'm open for code suggestions/benchmarks to improve viewport check.

> anyway I believe the polling method should not care about visibility

It kind of does already: if the tab is not active, polling only runs 5% of the time.

> the idea is that wire:visible updates a variable to the current state, that way you can use the functionality to other stuff like adding\removing class etc..

I have the same difficulty with `wire:visible` as I have with `wire:poll.visible`. It implies something that is not the current default in Livewire. Being that it would not be polling unless this attribute is set. 

About implementing this for other wire attributes too: this would quickly become problematic. Polling is the only mechanism that runs continuously. If you do not set a class on an element when it is not in the viewport, nothing will add it later on for you when the element does get within the viewport. This is beyond the scope of what this PR intends to do and would definitely make things waaay more complex. Sorry.

I will commit the changes I think make most sense considering the feedback I got so far.

joshuadegier:
Thanks for your efforts and time @ju5t - haven't been able to reply sooner.

I think we should first look at what the functionality does, as you already said, default behavior is to always poll except when the tab is in the background or Livewire is offline. So this feature would add "prevent polling when element is out of view".

In my eyes adding `.visible` implies that polling only occurs when the element is visible. It would still be an addition to `.keep-alive` which does the opposite: keep the polling alive no matter if the element is visible or in another tab. Giving a user the ability to choose one of the three options:

- `wire:poll`: always poll except when tab is closed or offline
- `wire:poll.keep-alive`: always poll
- `wire:poll.visible`: only poll when element is (partially) visible in the viewport

I think `poll.visible` is more in line with keep-alive compared to disable-hidden.

ju5t:
I think in the end the general consensus is to use `visible`. 

I would personally go for the 'breaking' change of changing the default so you can always use `keep-alive` to make sure it always polls. It makes more sense to me personally. This what the last commit in this PR does, but the majority wins.

The code for this would be:

```javascript
if (! inViewPort(directive.el) && directive.modifiers.includes('visible')) {
    if (Math.random() < .95) return
}
```

This would only poll if the item is visible. It will limit requests when it's not. This is similar to how `wire:poll` works for tabs that are inactive.

It ignores `keep-alive`, but `wire:poll.visible.keep-alive` makes no sense anyway. 

👍  or 👎

joshhanley:
I like the way @joshuadegier broke it down, they're all telling you when polling is happening.

I personally would lean away from a breaking change, for two reasons:
1. The less breaking changes for users the better in my opinion (unless there is a really good reason)
2. If you want this feature to be merged and useable now, it would have to be a non-breaking change, otherwise you would need to wait until the next major release.

If you wanted to make it a little more explicit (which I don't thinks necessary) but you could change it to `when-visible`.

So they would look like this (copying @joshuadegier's nice breakdown):
- `wire:poll`: always poll except when tab is closed or offline
- `wire:poll.keep-alive`: always poll
- `wire:poll.when-visible`: only poll when element is visible in the viewport

ju5t:
@joshhanley just to be sure, that's what the code snippet in my previous reply does too. If you use `wire:poll.visible` it will only poll the element when it's in the viewport. If it's not, it will block ~95% of the requests, exactly the same as the current functionality of the tab in the background.

In the end it doesn't matter much. It's also personal preference. I am 99% of the time with you on breaking changes, but I'm 40/60 on this one. As `keep-alive` is an existing modifier with the same functionality, I would've gone for the breaking change if it was up to me.

But it is not my feature, it's @joshuadegier. I just thought I'd spend some time on it as it is an easy feature to implement. I've just committed the last version. It does block 95% of the requests, I will leave it up to @joshuadegier if this should be 100%.

Thanks again for all the input.

joshhanley:
Ah sorry I read that wrong, I thought the code snippet was in relation to your breaking change option. 🤦‍♂️ Don't mind me!

Can I suggest putting and edit at the top of the original post explaining the final code/decision and then put a line between the edit and the original, so other's (read Caleb) don't need to read the whole thread to find out what the final ideas was?

ju5t:
All done!

joshuadegier:
All good as far as I'm concerned! Just curious to see @calebporzio's take on this one aswel. Since he's going to determine wether or not this ends up in a release or not. 

Currently my agenda is overbooked, but I'll try my hardest to fit in a few minutes anytime soon to check out your changes @ju5t in the project I need this for! Once again: thanks for your time. Let's hope this gets merged (or improved and merged) and I'm all set to payout the bounty I promised :-)

ju5t:
> I'm all set to payout the bounty I promised :-)

@joshuadegier I prefer it to be donated to a charity or to sponsor an open source project like Livewire.

joshuadegier:
@ju5t I respect that decision! If it's okay for you, I'd like to support that choice by already donating half of the bounty (125 EUR) just for your efforts to a charity or OS project of your choice <3

calebporzio:
Thanks for the PR @ju5t, I like the solution. Are you able to write a test for this?

Here would be my general approach off the top of my head.

You can reference the existing polling Dusk test here: tests/Browser/Polling/Test.php

- have a counter on the page that increments with every Livewire component refresh (at the top of the page)
- have a div the size of the viewport next
- have an element with `wire:poll.visible` at the bottom (hidden below the viewport)
- wait the poll time, assert the counter hasn't changed
- scroll into view with dusk, wait, assert the counter HAS changed
- scroll back to top, wait and assert it hasn't changed again

calebporzio:
Closing this PR for now. If someone wants to re-submit this code with a test, I'm open to it. Thanks!

ju5t:
Not sure why you closed this @calebporzio? I'm more than happy to make tests but my time is limited. It kind of feels like PR's aren't valued? Fair enough if it is a policy to close them but I find this a bit disappointing as you might notice by my tone of voice.

ju5t:
@joshuadegier I have no preference. It's just that I don't need to get paid for a small PR like this, even though it took more time than I anticipated in the end. I like Livewire and feel that people who make open source software in general should be much more rewarded for what they make.

It's up to you mate, I have no strong opinion about who you donate it to.

calebporzio:
Hey @ju5t - sorry you were put off by me closing this PR. When I close a PR like this, it's not me saying "no" or performing an act of aggression or disapproval or something. It's simple me keeping my never-ending-always-growing-non-stop github inbox from being messy as hell.

A lot of PRs that I comment on and ask for something never get addressed and then they are left out there to get stale and add to my mental load.

Again, thanks for the contribution, no harm on my part.

calebporzio:
Also, "It kind of feels like PR's as small as these aren't valued? " - where does that come from? I think I pay more attention to them honestly lol.

ju5t:
@calebporzio I might let myself go a bit, sorry about that. At the time I thought it was really unfair to close a PR that had been active until the solution was finished as contributors do spend their free time on it too. I'm not always able to respond quickly, so closing it within a week or so to me seemed pretty quick. To be fair, I still think that, but my response was a little disproportionate.

PR #2382, opened 2021-01-20
Automatic Component Routing

redbastie:
This PR will add auto-routing for full page Livewire components.

## Usage

Specify public `$route*` properties in your full page Livewire component:

    class Vehicle extends Component
    {
        public $routeUri = '/vehicle/{name}';
        public $routeDomain = null;
        public $routeMiddleware = ['guest'];
        public $routeName = 'vehicle';
        public $routeWhere = ['name' => '[A-Za-z]+'];
    
        public function render()
        {
            return view('livewire.vehicle');
        }
    }

A minimum of `$routeUri` is required.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -118,6 +118,10 @@ class LivewireServiceProvider extends ServiceProvider
 
     protected function registerRoutes()
     {
+        $this->loadRoutesFrom(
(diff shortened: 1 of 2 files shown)


calebporzio:
Hey @redbastie, this is pretty cool.

Unfortunately I'm not sure this belongs in Livewire core.

Based on your implementation it looks like it wouldn't be too difficult to create a composer package that provides this functionality.

I suggest doing that and keeping this as third-party behavior.

PR #2386, opened 2021-01-21
[Updated - loadMissing] Restore collection relations when hydrating

### LINKED ISSUE #2010: Eager Loading Relationships
I'm not sure if this is a bug or just a missing feature or maybe I'm just doing something wrong. The idea is that I want to eager load a relationship to resolve the n + 1 problem.

In the mount method I have a query where I load the relationship with the "with" method. If I make a "dd", the relationship is there and loaded:
`$query = OfferFile::select();

$query->where('offer_id', '=', $this->offer_id);

$query->with('File');

this->offerFiles = $query->get();`

image

But then in the Laravel Debug bar I still have all the queries running for each relationship one by one when it's rendered:
image

Then when I go back the Livewire Class for example with a button click, the relationship is missing. That means the relationship doesn't survive the trip to the blade?
image

Is that a bug? Or just not supported by Livewire? Or how is the recommended way to solve the n + 1 problem?

**Context**
- Livewire version: 2.2.9
- Laravel version: 8.9.0
- Browser: Chrome

joshhanley:
UPDATE
Use `loadMissing`. 

Thanks to @DanielCoulbourne, that method is perfect for this!

Reference: https://youtu.be/NjUj4oGGCbQ?t=4401

---

This PR adds eloquent collection relation restoration to the hydration process to stop N+1 issues (see #2010).

Ideally this is a short-term fix with this eventually being fixed upstream in the Laravel framework so it's no longer needed here.

I have included a check to ensure that, if this does get fixed upstream, and collection relations are loaded in Laravel's `SerializesAndRestoresModelIdentifiers` class, that Livewire won't load them again. Otherwise this would cause a double up of database requests for the relations (turns out `->load()` isn't smart enough to know when a relation is already loaded).

I've added tests for both collection with eager loaded relations, and without, to ensure all is working as expected.

I couldn't add a test for the upstream relation loaded check as I have no way to hook in and preload the relations after the collection has been restored but before this call. But I have manually tested it with a local version of Laravel with the patch applied and ensured that the check works as expected and that there wasn't a double of up database requests.

Hope this helps!

Fixes #2010


calebporzio:
This is so clean. Thanks for this @joshhanley 

I'm heading out of town until next week, will merge then. Thanks!

joshhanley:
@calebporzio no worries! 

Ok sounds good. Enjoy your trip 🙂

calebporzio:
Hey @joshhanley - I love the implementation, but like I suspected, if I comment out your addition that tests still pass. Are you seeing that as well? Would you mind updating the tests to fail without the feature? Thanks!

joshhanley:
@calebporzio bugger!

I'd accidentally left the manual hydration of relations in the test component. I have removed that and have now ensured the tests failed with fix commented out.

You can look at it again now 😁

joshhanley:
Tests failed for unrelated reasons, have merged master in to trigger them again, but still failed, not sure what is going on with them

calebporzio:
This is GREAT. Thank you so much for this @joshhanley !

joshhanley:
No worries! 🙂

PR #2408, opened 2021-01-26
base64 make return "/" and break the path

VakulenkoIhor:

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -124,7 +124,7 @@ class TemporaryUploadedFile extends UploadedFile
     public static function generateHashNameWithOriginalNameEmbedded($file)
     {
         $hash = Str::random(30);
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for this. Would love a test for this. Do you feel up for that?

VakulenkoIhor:
> Thanks for this. Would love a test for this. Do you feel up for that?

I've added test for that case.

calebporzio:
Thanks for the test. From a quick glance, it looks like in the implementation you are replacing forward slashes with underscores, but then in your test, you are testing a filename with spaces, not forward slashes. Can you update the test to reflect the usage you were originally trying to fix. Or if I'm miss-understanding, can you inform me? Thanks!

calebporzio:
Closing due to inactivity - feel free to re-submit. Thanks!

PR #2409, opened 2021-01-26
[1.x] Only return integers for pages

driesvints:
I've recently had an issue in Laravel.io where if you passed a non-integer value as the page query string parameter the internal page resolving in Laravel would crash and burn. At the moment as it stands, only an integer can be passed as the page number: https://github.com/laravel/framework/blob/8.x/src/Illuminate/Pagination/AbstractPaginator.php#L492

One could also argue we should type cast in the method I linked above. I'll also PR that to the framework as a companion PR.

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -18,7 +18,7 @@ trait WithPagination
         $this->page = $this->resolvePage();
 
         Paginator::currentPageResolver(function () {
-            return $this->page;
+            return (int) $this->page;
         });
 
         Paginator::defaultView($this->paginationView());
@@ -53,6 +53,6 @@ trait WithPagination
     {
         // The "page" query string item should only be available
         // from within the original component mount run.
-        return request()->query('page', $this->page);
+        return (int) request()->query('page', $this->page);
     }
 }


driesvints:
I have no idea why the tests are failing. Seems unrelated to this PR 🤔

joshhanley:
@driesvints yeah github actions tests have been unreliable on the Livewire repo of late, not sure why. I submitted #2412 and they failed, submitted a different one #2413 a few minutes later and all tests passed 🤷‍♂️

Do you think it should still cast this in Livewire even though the fix has now been merged into Laravel?

nuernbergerA:
Is 1.x even supported?

It's super easy to upgrade to v2

driesvints:
> Do you think it should still cast this in Livewire even though the fix has now been merged into Laravel?

Yeah, I think it's a good fix to have anyhow because you should only return integers anyway.

> Is 1.x even supported?

I checked with Caleb beforehand and he was okay with me sending this in. This can also just go into 2.x, I really don't mind 🙂

calebporzio:
Thanks for this @driesvints 

@nuernbergerA, you're right, however, the bigger hurdle for people is upgrading their Laravel version to one compatible with v2.

PR #2411, opened 2021-01-26
Add support for Collections when using $queryString

stancl:

It's a bugfix

It does not

A regression test could be useful, though I didn't have the time to write one and I wanted to submit this at least in some form.

When you have a typed `Collection` property, Livewire will treat it as an array when setting data from the query string. Resulting in an error saying that the object is uninitialized.


calebporzio:
Thanks for the contribution, I will need to see more demand in the community for this feature (or issues submitted by others) AND I'd need a test included with this PR. Also, I avoid messing with the browser history system if I can help it. Anyhow, that doesn't mean it's not a valid need, but those are my hangups for now. Closing this. Please re-submit if you have solutions to my hangups. Thanks!

Issue #2395, opened 2021-01-23
X-Amz-Expires must be non-negative while file upload since 2.3.7


stefro:
### Description
I've upgraded from Livewire 2.3.6 to 2.3.7 (also tried to 2.3.8, same result).
My file upload (to S3) suddenly stopped working. When I open de network tab in my browser I can this request with a 400 status:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error>
    <Code>AuthorizationQueryParametersError</Code>
    <Message>X-Amz-Expires must be non-negative</Message>
    <RequestId>784898D8C4E388F4</RequestId>
    <HostId>vPRsnfz9YMKe0FXpF4Nk513SP4VNff8hszxRdZArjhXrZM2k2p6QdyfXM1vT1t4Pwtfu8dKOEwY=</HostId>
</Error>
```
When I go back to Livewire 2.3.5 it works.

### Exact steps to reproduce
In my blade file I've stripped down all fancy styling stuff to a bare minimum:
```html
<input type="file" id="newImage" name="newImage" wire:model="newImage">
```
### Stripped-down, copy-pastable code snippets

**Context**
- Livewire version: 2.3.6
- Laravel version:8.24.0
- Alpine version: 2.3.5
- Browser: Safari & Chrome

tarikhagustia:
same result

eqxDev:
Same issue here

heruputra:
[This addition](https://github.com/livewire/livewire/commit/ca2ca53963de5574271af08021f329f44a1cae07) to the config file need to be added manually to your config file.

Looks like an undocumented breaking changes.

felixsc:
> [This addition](https://github.com/livewire/livewire/commit/ca2ca53963de5574271af08021f329f44a1cae07) to the config file need to be added manually to your config file.
> 
> Looks like an undocumented breaking changes.

Fixed it for me! Thank you @heruputra

stefro:
> [This addition](https://github.com/livewire/livewire/commit/ca2ca53963de5574271af08021f329f44a1cae07) to the config file need to be added manually to your config file.
> 
> Looks like an undocumented breaking changes.

Works for me too! Thanks!

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

[...]

PR #2412, opened 2021-01-26
Merge published config with package config

joshhanley:
This was intended to fix an issue #2395 caused by #2347 where the default max_upload_time wasn't being populated if it wasn't in the apps published config.

But this doesn't actually fix it, as per this note in the Laravel docs, as max_upload_time is a second level configuration option.
>This method only merges the first level of the configuration array. If your users partially define a multi-dimensional configuration array, the missing options will not be merged.

I will submit a separate fix for #2395, but I thought this merge should still happen to prevent this issue for any future first level config options.

Hope this helps!

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -53,6 +53,7 @@ class LivewireServiceProvider extends ServiceProvider
         $this->registerTestMacros();
         $this->registerLivewireSingleton();
         $this->registerComponentAutoDiscovery();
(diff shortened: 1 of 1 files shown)


gdebrauwer:
If you didn't publish the config, then `config('livewire')` will always be null. Using `mergeConfigFrom` also fixes that issue

calebporzio:
Thanks @joshhanley!

joshhanley:
No worries!

Issue #2401, opened 2021-01-23
Undefined index error when emitting event which no longer exists


mjerwin:
### Description
When using `getListeners()` to define dynamic listeners if one of those dynamic listeners is removed then you get an undefined index error.

**I've got a fix for this ready to PR if this deemed to be a valid issue.**
Suggested fix is to check for undefined indexes in `\Livewire\ComponentConcerns\ReceivesEvents::fireEvent`

### Exact steps to reproduce

1. Override `getListeners()` to add dynamic events
2. Call an action which would cause one of the dynamic events to be removed
3. Emit the event which has now been removed
4. See "Undefined index" error

### Stripped-down, copy-pastable code snippets
```php
class ItemList extends Component
{
    public $items = [];

    public function mount()
    {
        $this->items = [
            1 => ['name' => 'Item 1'],
            2 => ['name' => 'Item 2'],
            3 => ['name' => 'Item 3'],
        ];
    }

    public function render()
    {
        return view('livewire.item-list');
    }

    public function delete($id)
    {
        unset($this->items[$id]);
    }

    public function getListeners()
    {
        return collect($this->items)
            ->mapWithKeys(function ($item, $key) {
                return ['itemUpdated:' . $key => '$refresh'];
            })
            ->toArray();
    }
}
```

```javascript
Livewire.emit('itemUpdated:2');
```

```
ErrorException: Undefined index: itemUpdated:2

/Users/mje/Sites/packages/livewire/src/ComponentConcerns/ReceivesEvents.php:72
/Users/mje/Sites/packages/livewire/src/HydrationMiddleware/PerformEventEmissions.php:19
/Users/mje/Sites/packages/livewire/src/LifecycleManager.php:70
/Users/mje/Sites/packages/livewire/src/Connection/ConnectionHandler.php:12
/Users/mje/Sites/packages/livewire/src/Controllers/HttpConnectionHandler.php:13
/Users/mje/Sites/packages/livewire/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php:48
/Users/mje/Sites/packages/livewire/vendor/laravel/framework/src/Illuminate/Routing/Route.php:254
/Users/mje/Sites/packages/livewire/vendor/laravel/framework/src/Illuminate/Routing/Route.php:197
/Users/mje/Sites/packages/livewire/vendor/laravel/framework/src/Illuminate/Routing/Router.php:692
[...]

PR #2427, opened 2021-01-30
Fix undefined index error

mjerwin:

Fix for issue #2401 

Added test which fails without this fix and passed after the fix.

Added null coalescing operator when finding the event method to handle this and added tests.

**Exact steps to reproduce**
Override getListeners() to add dynamic events
Call an action which would cause one of the dynamic events to be removed
Emit the event which has now been removed
See "Undefined index" error

This will cause the following error:
ErrorException: Undefined index: itemUpdated:2
/Users/mje/Sites/packages/livewire/src/ComponentConcerns/ReceivesEvents.php:72

### DIFF
--- a/src/ComponentConcerns/ReceivesEvents.php
+++ b/src/ComponentConcerns/ReceivesEvents.php
@@ -69,8 +69,10 @@ trait ReceivesEvents
 
     public function fireEvent($event, $params)
     {
-        $method = $this->getEventsAndHandlers()[$event];
+        $method = $this->getEventsAndHandlers()[$event] ?? null;
 
-        $this->callMethod($method, $params);
+        if($method) {
+            $this->callMethod($method, $params);
+        }
     }
 }
--- a/tests/Unit/ComponentEventsTest.php
+++ b/tests/Unit/ComponentEventsTest.php
@@ -39,6 +39,18 @@ class ComponentEventsTest extends TestCase
     }
 
     /** @test */
+    public function receive_modified_dynamic_event()
+    {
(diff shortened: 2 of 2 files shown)


joshhanley:
I was assisting on an issue that had a similar error the other day.

I think there is two parts to this, this PR would be the first part, ensuring that the listener key exists before trying to access it.

The other would be making sure the the dynamic listener gets removed from the listener list on the front end (I haven't investigated yet to see how it handles this and whether or not it does get removed).

calebporzio:
Thanks for submitting this. Like @joshhanley mentioned, I believe changes to listeners at runtime aren't currently observed by Livewire on the front-end. I think Livewire sets the listeners up on initialization in JS, and doesn't change them with new listeners from the backend.

Your tests aren't showing this because they are Unit tests and not Dusk tests that excercise the actual JS.

Going to close this for now, feel free to re-open if you address the above. Thanks!

cby016:
We are currently getting these undefined array key errors on our production environment.

`Undefined array key "echo..." at .../vendor/livewire/livewire/src/ComponentConcerns/ReceivesEvents.php:73`

I thought this issue was closed and the changes merged in, but when I looked at the ReceivesEvents.php file, it seems [the changes](https://github.com/livewire/livewire/pull/2427/files) never made it into the code base.  @calebporzio @mjerwin Any idea why not?

joshhanley:
@cby016 currently dynamic listeners aren't supported. This PR was closed as per Caleb's comment above, about it being an incomplete fix and no one has re-submitted with a complete fix. I'm sure a PR with the fix would be welcome!

PR #2434, opened 2021-02-01
Fix action doubling - closes #763

### LINKED ISSUE #763: x-on:click called twice after livewire call under certain circumstances

austenc:
Reported in #763 and continued from PR in #1995. Sometimes Alpine event listeners are registered twice when inside a blade conditional. 

This PR is still missing one change, but it's dependent on https://github.com/alpinejs/alpine/pull/1058. Marking as a draft until the fate of the PR on the Alpine repo is determined. There will be no other changes required for Livewire since the main fix is in Alpine. 

The change is only in the tests, and it'll just be bumping the Alpine CDN version in the testing `app.blade.php` layout.

Thanks to @joshhanley for writing the original test for this! I've added an extra test case to ensure that a nested alpine component doesn't have event listener problems too.


calebporzio:
Thanks for this update Austen!

joshhanley:
Thanks @austenc! Tested it today and it works great!

PR #2436, opened 2021-02-01
only clear session flash data on subsequent dehydrate - closes #2294

### LINKED ISSUE #2294: Components clearing flash data set elsewhere (e.g. controller)
### Description

If a flash message is set in the session store from the controller, the data won't be available after rendering a LiveWire component.

### Exact steps to reproduce
Attached a playground snippet.

After opening the snippet, the view should show "Hello from component". If you comment out the rendering (`<livewire:hello-world />`), it will say "hello from session flash".

I also dumped the session data and you can see that the _flash.new still contains a key reference to the "danger" key, but the actual value is missing.

https://laravelplayground.com/#/snippets/3bd10c5e-6833-409f-89ce-08bc47499e66

austenc:

Original Issue: https://github.com/livewire/livewire/issues/2294

I tried to write a test for this, but it seems to be difficult to flash data to the session prior to dusk running. Hopefully someone can point out an easy way to do this and I'd be happy to include a test. 

Data flashed to the session is not available on the initial render like you'd expect. This PR changes it so that Livewire only  
clears flash data from the session on subsequent dehydrates.

🍻

### DIFF
--- a/src/RenameMe/SupportRedirects.php
+++ b/src/RenameMe/SupportRedirects.php
@@ -32,15 +32,21 @@ class SupportRedirects
             // Put the old redirector back into the container.
             app()->instance('redirect', array_pop(static::$redirectorCacheStack));
 
+            if (empty($component->redirectTo)) {
+                return;
+            }
+
+            $response->effects['redirect'] = $component->redirectTo;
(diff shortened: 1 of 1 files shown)


calebporzio:
I dig. Thanks!

PR #2437, opened 2021-02-01
fixed encoding when building method + param signature - closes #2390

### LINKED ISSUE #2390: Problem with encoding $set() function
### Description
When i try to send in javascript with $set() function some special char like "řěšč" i will get error 
`Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.` And livewire start sending every second new request to `http://webdo.test/livewire/message/...`, but what is interesting is that if i look into variable value i get a right value with special chars. So i think it will be only something with the nonending requests. 

There is no needed so long code to show. Just try in blade component add this into <script></script>
`@this.set('variable', 'žžž');`

**Context**
- Livewire version: [2.3.6]
- Laravel version: [8.3.1]
- Alpine version: [2.3.0]
- Browser: [Chrome, Firefox]

austenc:

Reported in #2390

No sir!

Yes, I added an additional chunk to the Alpine dusk test.

When an action happens we serialize the method and params with `btoa` in javascript. This results in an error when trying to set a variable to something that contains strange characters like `žžž`.

Other uses of `btoa` within Livewire use `encodeURIComponent`, so I added that to the params in this case. 

🍻


htulibacki:
People who use these strange characters on a daily basis as an integral part of their language will be very grateful for merging this pull request! 😄

mokhosh:
@htulibacki Yes, everyone who is using non Latin characters (Persian, Arabic, Hebrew, Hindi, Urdu, Japanese, etc.) will be very grateful, including myself.

@calebporzio Appreciate the frequent releases lately. Can you please merge this in one of these releases?

mityukov:
need this fix badly… :/

JunaidQadirB:
When is this PR going to be merged?

baceto90:
I have a problem with Cyrillic characters too. Hopefully this pull request will be merged as soon as possible.

mokhosh:
@austenc seems like you have spatie/ray in your composer.json and it's causing the tests to fail. It might help with this getting merged if the tests are passing when @calebporzio gets here :D

austenc:
@mokhosh thanks for the reminder -- updated with latest master and tests are now passing!

fawzanm:
Thanks for this PR. I badly need this PR to get merged. 😔

calebporzio:
Merged and tagged, thanks!

Issue #2414, opened 2021-01-27
Emitting an event beforeunload


redbastie:
I'm getting this error:

> Fetch API cannot load http://dio-ccg.test/livewire/message/partials.play due to access control checks.

Here is my JS:

    window.addEventListener('beforeunload', () => {
      console.log('leaveGameQueue');
      Livewire.emit('leaveGameQueue');
    });

Here is my component:

    class Play extends Component
    {
        public function getListeners()
        {
            return [
                'leaveGameQueue',
            ];
        }
  
        public function render()
        {
            return view('livewire.partials.play');
        }
  
        public function leaveGameQueue()
        {
            Auth::user()->gameQueue()->delete();
        }
    }

Is it possible to emit an event `beforeunload`? In this example, I'm making a simple game, and the user can "queue up" for matchmaking. When the person leaves the page, it should automatically remove them from the queue.

I just want to emit that event when they leave the page.

redbastie:
Looks like this only happens in Safari. Chrome is working fine. :/

calebporzio:
Because `Livewire.emit()` will submit an AJAX request for the listening Livewire components, this is not ideal.

"beforeunload" and "unload" event listeners should not be used for sending ajax requests.

The "Beacon" API is intended for cases like these.

I'm not sure how this fits into Livewire exactly, but I'd recommend reading up on the "send beacon" functionality here: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

Then maybe handle this behavior in a normal request outside of a Livewire component. Not really sure tbh.

Going to close this issue, but please report your findings here for others to reference.

Thanks!

joshhanley:
Closing this as per Caleb's comment above 🙂

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

27pchrisl:
REOPEN

Reopened for feedback on #2439

27pchrisl:
REOPEN

[...]

PR #2439, opened 2021-02-02
Support for emitting events as beacons

27pchrisl:

Requested in #2414 

No

Yes

Both @redbastie and myself have a requirement to be able to dispatch an event to a Livewire component as the browser unloads. In my case to send last-minute analytics data. As noted in #2414 the way to do that is with sendBeacon.

This PR adds support for emitting an event as a beacon. I have added `<component>.beacon()` in the same style as `<component>.emit()`, and `Livewire.beaconTo()` the same as `Livewire.emitTo()`.

Technically you could add beaconing all over the place in the Livewire API, but I think the most likely use case is where a component is listening to a visibilitychange or unload event, and can call its own beacon() handler in that callback to make a last method call to the server-side component.

Sure thing!

Will leave this here for @calebporzio to comment on whether this is the right way to add this functionality.


27pchrisl:
Ah-ha, you have to commit the build for the tests!

calebporzio:
Closing this as something that is better left to user-land implementations until there is greater demand. Thanks!

jerodion:
this is still a huge problem

why wasn't this merged?

this PR solves the issue of safari failing to emit events on unload etc

I'd even argue that this should actually be part of the native emit event functionality

Issue #2440, opened 2021-02-02
problem on ubuntu 18.04.05: Call to a member function name() on array


ravenshill:
### Upgrade to 2.3.9, 
run :   composer install --no-interaction --prefer-dist --optimize-autoloader

### Stripped-down, copy-pastable code snippets

**Context**
- Livewire version: [2.3.8] 
- is ok
- Laravel version: [2.3.9]
- @php artisan package:discover --ansi

   Error 

  Call to a member function name() on array

  at vendor/livewire/livewire/src/LivewireServiceProvider.php:129
    125▕     protected function registerRoutes()
    126▕     {
    127▕         RouteFacade::post('/livewire/message/{name}', HttpConnectionHandler::class)
    128▕             ->middleware(config('livewire.middleware_group'))
  ➜ 129▕             ->name('livewire.message');
    130▕ 
    131▕         RouteFacade::post('/livewire/upload-file', [FileUploadHandler::class, 'handle'])
    132▕             ->middleware(config('livewire.middleware_group'))
    133▕             ->name('livewire.upload-file');

      +8 vendor frames 
  9   [internal]:0
      Illuminate\Foundation\Application::Illuminate\Foundation\{closure}()

      +5 vendor frames 
  15  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()
Script @php artisan package:discover --ansi handling the post-autoload-dump event returned with error code 1

ravenshill:
This is the relevant diff

    protected function registerRoutes()
    {
##removed 
        RouteFacade::post('/livewire/message/{name}', HttpConnectionHandler::class)->name('livewire.message');
##replaced with 
        RouteFacade::post('/livewire/message/{name}', HttpConnectionHandler::class)
            ->middleware(config('livewire.middleware_group'))
            ->name('livewire.message');

calebporzio:
Yeah, the previous release contained this:
![image](https://user-images.githubusercontent.com/3670578/106674423-35a31980-6581-11eb-8ca0-a2510ae0d1a1.png)

And the current release (like you saw) adds a fluent `->name()` call at the end.

A can't see why this would break anything though? Why is `->middleware()` returning an array in your error message?

I tried `artisan package:discover` locally with success.

Is there something possibly custom about your setup that would cause this? What's your Laravel version?

Thanks!

cg0012:
Hi Caleb, I am seeing the same on the latest Laravel build. I ran composer update:

```
λ composer update
[...]

PR #2444, opened 2021-02-03
Fix middleware_group

mertasan:
It solves the problem here: #2440

### Edit:

It does not work when set to `null`.


lloricode:
oh, so just a deprecate config file, ok thanks :)

Issue #2456, opened 2021-02-04
Temporary Uploaded File preview Url not working


lyriad:
### Description
In the last commit to master, the preview mimes array was removed from the `temporaryUrl` method, instead fetching the mimes from the livewire config file. Since having custom values in the config file is optional this should be avoided, and should have predetermined mime values as a default in case they're not defined in the config file.

Line 70 from TemporaryUploadedFile should be reverted to as it was in this commit `3f967d4747`.

### Exact steps to reproduce
Try to create a temporary url from a TemporaryUploadedFile instance.

### Stripped-down, copy-pastable code snippets

**Context**
- Livewire version: 2.2.15
- Laravel version: 8.21.0
- Alpine version: 2.7.3
- Browser: Chrome

lyriad:
@calebporzio

calebporzio:
Good catch, thanks for PRing. Merged and tagged in v2.3.16

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

lyriad:
No problem, thanks.

PR #2457, opened 2021-02-04
fix: add default value to supported preview files

lyriad:

It's needed. Also just created the issue [here](https://github.com/livewire/livewire/issues/2456).

No.

No.

Mostly a fix. Preview Urls not working due to supported mimes array being null if user doesn't have the values set in config file.


calebporzio:
Ooof thanks for catching that!

PR #2476, opened 2021-02-09
Remove debug related data from <livewire:scripts /> output

marispro:
1️⃣  Is this something that is wanted/needed? Did you create an issue / discussion about it first?
**No. Can be discussed here?**

2️⃣  Does it contain multiple, unrelated changes? Please separate the PRs out.
**No**

3️⃣  Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have)
**I guess not needed**

4️⃣  Please include a thorough description of the improvement and reasons why it's useful.
**Decreases document output and excludes debuggable information on non-debug environment.**

**Maybe config('app.debug') isn't best for this, because this information might be important for those who starts exploring Livewire world. But any another option to disable console warnings would be great.**

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -206,12 +206,6 @@ HTML;
     {
         $jsonEncodedOptions = $options ? json_encode($options) : '';
 
-        $devTools = null;
-
-        if (config('app.debug')) {
-            $devTools = 'window.livewire.devTools(true);';
-        }
-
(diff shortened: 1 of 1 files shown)


austenc:
Could you provide a little more information on what problem this is solving please?

marispro:
> Could you provide a little more information on what problem this is solving please?

Decreases document output and excludes debuggable information on non-debug environment.
Reduces file size by removing unnecessary code.
That is all.

calebporzio:
Nice, I'm with it! Thanks!

PR #2488, opened 2021-02-10
Add before validation hook

mokhosh:
In Laravel you can manipulate your validator after it's created and before you run it on the input. Livewire makes validation super easy, but if you want to do complex things with the validator you have to go back to using a normal validator.

This PR adds a before validation hook so you can do anything you need to the validator before running the actual validation. For example https://laravel.com/docs/8.x/validation#complex-conditional-validation

You just add a `protected beforeValidation` and it's gonna get called right before validation and receive the validator as argument.

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -141,6 +141,11 @@ trait ValidatesInput
         return ! $this->hasRuleFor($dotNotatedProperty);
     }
 
+    protected function beforeValidation(\Illuminate\Validation\Validator $validator)
+    {
+        return $validator;
+    }
+
     public function validate($rules = null, $messages = [], $attributes = [])
     {
         [$rules, $messages, $attributes] = $this->providedOrGlobalRulesMessagesAndAttributes($rules, $messages, $attributes);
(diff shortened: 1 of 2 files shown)


mokhosh:
To give a little more context, in Laravel you don't usually reuse validators as much as you do with Livewire if you even create one in the first place. But in a Livewire component it's pretty normal to run validation at the top of a handful of your methods. You will usually have realtime validation as well. And all of this is very cumbersome to achieve when using a normal validator.

So, I don't think the current solution presented in Livewire docs is really Livewire's way. But with this hook you can keep all of the Livewire goodness and use your minimalistic magical validation everywhere AND add complex functionality to validation.

calebporzio:
Thanks for the PR. I haven't heard much demand for this, AND it adds to the public API of Component which I am always hesitant to do. For now, I'll say just stick to the custom validator if you have needs beyond what Livewire currently offers.

mokhosh:
@calebporzio #noplanstomerge Wouldn't you say the likes on the PR could count as a small portion of an existing demand?

I am using custom validator, but it has a very bad user experience. Because you can't merge the result of the two validators. The user gets the errors of the normal validator, then after fixing all the nags, they are presented with some new errors by the custom validator.

Issue #2491, opened 2021-02-11
Bug in File Upload cleanup on high traffic websites


bzarzuela:
### Description
We use Laravel Livewire on a very busy website. In some cases, the file upload code is triggered with a race-condition that causes errors like this to be logged.

File not found at path: livewire-tmp/paeevOyI4gyBawMcQeRfOT3NX5A6WhmetaU2VsZmllIHdpdGggZ292dCBJRC5qcGc=-.jpg

### Exact steps to reproduce
Hard to reproduce but the stack trace gives a good confirmation.

`{
    "class": "League\\Flysystem\\FileNotFoundException",
    "message": "File not found at path: livewire-tmp/nUOOk8NDKbXX1ugfIdaUmoFd6GOXzp-metacmVjZWl2ZWRfMTcxNjIwOTE3ODAwODY5LmpwZWc=-.jpg",
    "code": 0,
    "file": "/home/forge/psahelpline.ph/vendor/league/flysystem/src/Filesystem.php:389",
    "trace": [
        "/home/forge/psahelpline.ph/vendor/league/flysystem/src/Filesystem.php:298",
        "/home/forge/psahelpline.ph/vendor/laravel/framework/src/Illuminate/Filesystem/FilesystemAdapter.php:427",
        "/home/forge/psahelpline.ph/vendor/livewire/livewire/src/WithFileUploads.php:100",
        "/home/forge/psahelpline.ph/vendor/livewire/livewire/src/WithFileUploads.php:29",
        "/home/forge/psahelpline.ph/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php:36",`

Line 100 of WithFileUploads is within this code block.

```
foreach ($storage->allFiles(FileUploadConfiguration::path()) as $filePathname) {
                $yesterdaysStamp = now()->subDay()->timestamp;
                if ($yesterdaysStamp > $storage->lastModified($filePathname)) {
                    $storage->delete($filePathname);
                }
        }
```

My idea is to add an exists() check there so that the error is not triggered when lastModified is called.

Maybe it won't completely eliminate the race condition. 

**Context**
- Livewire version: v2.3.6
- Laravel version: v8.23.1
- Alpine version: 2.7.3
- Browser: Chrome

bzarzuela:
I have this fix running on our website for several days now and it has eliminated all of these error messages from our logs.

austenc:
[...]

PR #2492, opened 2021-02-11
Fixes a race condition when Livewire uploads is used on busy websites

bzarzuela:
Yes, issue #2491 2491

No

No, unfortunately I don't know how to test race conditions

On very busy websites, the cleanup code can run in multiple threads causing part of the output of allFiles() to have already been deleted by another thread.

### DIFF
--- a/src/WithFileUploads.php
+++ b/src/WithFileUploads.php
@@ -96,11 +96,13 @@ trait WithFileUploads
         $storage = FileUploadConfiguration::storage();
 
         foreach ($storage->allFiles(FileUploadConfiguration::path()) as $filePathname) {
-            if ($storage->exists($filePathname)) {
-                $yesterdaysStamp = now()->subDay()->timestamp;
-                if ($yesterdaysStamp > $storage->lastModified($filePathname)) {
-                    $storage->delete($filePathname);
-                }
+            if (! $storage->exists($filePathname)) {
+                continue;
+            }
+
+            $yesterdaysStamp = now()->subDay()->timestamp;
+            if ($yesterdaysStamp > $storage->lastModified($filePathname)) {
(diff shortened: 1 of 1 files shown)


src/WithFileUploads.php
```diff
@@ -96,9 +96,11 @@ protected function cleanupOldUploads()
         $storage = FileUploadConfiguration::storage();
 
         foreach ($storage->allFiles(FileUploadConfiguration::path()) as $filePathname) {
-            $yesterdaysStamp = now()->subDay()->timestamp;
-            if ($yesterdaysStamp > $storage->lastModified($filePathname)) {
-                $storage->delete($filePathname);
+            if ($storage->exists($filePathname)) {
```
nuernbergerA, line 99: you could reduce cyclomatic complexity by continue early
```php
if (! $storage->exists($filePathname)) {
    continue;
}

$yesterdaysStamp = now()->subDay()->timestamp;
if ($yesterdaysStamp > $storage->lastModified($filePathname)) {
    $storage->delete($filePathname);
}

nuernbergerA (changes requested):

src/WithFileUploads.php
bzarzuela: You're right, this is a better solution. How do I pull in your suggestion in a way that is properly attributed to you? I'm not that familiar with this feature.

nuernbergerA: Just commit it I have enough contributions.

And "my way" is just formating your code different 👍

nuernbergerA (approved):

calebporzio:
Good catch. Thanks! I understand the difficulty of testing this sort of thing, so I'm good with what you have.

PR #2495, opened 2021-02-12
Add assert count to tests

chrisrhymes:
I haven't created an issue or discussion for it, but it is something that I have thought would be useful to make it easier to test arrays or results of queries are what you expect

Single PR

Yes

This would be useful to count a property, such as the number of results returned in a data table before and then after a search or filter has been done by updating another Livewire component property. 

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -28,6 +28,13 @@ trait MakesAssertions
         return $this;
     }
 
+    public function assertCount($name, $value)
+    {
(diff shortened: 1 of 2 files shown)


calebporzio:
I like it. Thanks!

calebporzio:
@chrisrhymes - would you mind PRing this to the docs? 

Here are the two files to PR to (should be really easy):
- https://github.com/livewire/docs/blob/2.x/testing.blade.php
- https://github.com/livewire/docs/blob/2.x/reference.blade.php

Thanks!

Issue #1707, opened 2020-09-27
Undefined variable slot in app.blade.php


MUHAMMADSiyab:
I am using `livewire 2.2` with `laravel 7.24` in my laravel project. When i load component using 
`Route::get('/path', ComponentClassPath);`
it works, but when i load my livewire component inside a blade view. For instance, **test.blade.php** like this
```
@extends('layouts.app')

@section('content')
    @livewire("my-livewire-component")
@endsection
```
it doesn't work and shows the below error:

**Undefined variable: slot (View: E:\xampp\htdocs\lw\resources\views\layouts\app.blade.php)** .

Any solution ?

- Livewire version: 2.2
- Laravel version: 7.24

iksaku:
As [stated in Livewire's Documentation], when rendering a component, you need to provide the base template, be it using a `$slot

mblarsen:
If you want to use a template that uses @yield (sections) you can use the 'extended' method in your component instead:

```php
public function render()
{
    return view(...)
        ->extends('layouts.app') // the default is layout('layouts.app')
        ->section('content');
}
```

https://laravel-livewire.com/docs/2.x/rendering-components#custom-layout

MUHAMMADSiyab:
> If you want to use a template that uses @yield (sections) you can use the 'extended' method in your component instead:
> 
> ```
> public function render()
> {
>     return view(...)
>         ->extends('layouts.app') // the default is layout('layouts.app')
>         ->section('content');
> }
> ```
> 
> https://laravel-livewire.com/docs/2.x/rendering-components#custom-layout

I got it solved by using `extends` method. Thanks @mblarsen

RomainMazB:
For those who care, [I've submitted a PR](https://github.com/livewire/livewire/pull/1807) which describe the problem and could have solved it but had not been merged :/

keongbalap757:
thank you for the information I came to know about [slot online](http://128.199.202.211/)

keongbalap757:
hank you for the information I came to know about[ slot online](https://www.lisinopril0.com/)

beactriclaura:
<a href="https://bit.ly/slotpulsaonlineterbaik"<game-online-play</a>

thank you for the information

beactriclaura:
<a href="https://bit.ly/slotpulsaonlineterbaik
good

beactriclaura:
<a href="https://bit.ly/slotpulsaonlineterbaik
good

bolarolet88muh:
[...]

PR #2499, opened 2021-02-14
Fix x layout

RomainMazB:
[Yes](https://forum.laravel-livewire.com/t/layout-class-constructor-is-not-triggered-while-rendering/1985), there is some other issues related to this: #1707, https://forum.laravel-livewire.com/t/livewire-with-custom-layout/1666/8

No

No

This PR follows a [previous one](https://github.com/livewire/livewire/pull/1807), already discussed and closed due to the way I tried to solved it.

I'm trying to follow the @calebporzio guideline and re-implement this in an acceptable way.

To use this new implementation, we just need to define the layout class into the layout method:
```
<?php namespace App\Http\Livewire;

use Livewire\Component;
use App\View\Components\AppLayout;

{
    public function render()
    {
        return view('livewire.my-component')->layout(AppLayout::class);
    }
}
```

### DIFF
--- a/src/Macros/ViewMacros.php
+++ b/src/Macros/ViewMacros.php
@@ -21,6 +21,12 @@ class ViewMacros
     public function layout()
     {
         return function ($view, $params = []) {
+            if (is_subclass_of($view, \Illuminate\View\Component::class)) {
+                $layout = new $view();
+                $params = array_merge($params, $layout->data());
+                $view = $layout->resolveView()->name();
+            }
+
             $this->livewireLayout = [
                 'type' => 'component',
                 'slotOrSection' => 'slot',
--- a/tests/AppLayout.php
+++ b/tests/AppLayout.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Tests;
+
+use Illuminate\View\Component;
+
+class AppLayout extends Component
+{
+	public $foo = 'bar';
+
+    public function render()
+    {
+        return view('layouts.app-x');
+    }
+}
--- a/tests/Unit/ComponentLayoutTest.php
+++ b/tests/Unit/ComponentLayoutTest.php
@@ -47,6 +47,16 @@ class ComponentLayoutTest extends TestCase
 
         $this->withoutExceptionHandling()->get('/foo')->assertSee('baz');
     }
+
+    /** @test */
+    public function can_load_dynamic_layout_properties()
+    {
+        Livewire::component(ComponentWithDynamicLayout::class);
+
+        Route::get('/foo', ComponentWithDynamicLayout::class);
+
+        $this->withoutExceptionHandling()->get('/foo')->assertSee('bar')->assertSee('baz');
+    }
 }
 
 class ComponentWithExtendsLayout extends Component
@@ -86,5 +96,15 @@ class ComponentWithCustomSlotForLayout extends Component
     public function render()
     {
         return view('show-name')->layout('layouts.app-custom-slot')->slot('main');
+    }
+}
+
+class ComponentWithDynamicLayout extends Component
+{
+    public function render()
+    {
+        return view('null-view')->layout(\Tests\AppLayout::class, [
+            'bar' => 'baz'
+        ]);
     }
 }
--- a/tests/Unit/views/layouts/app-x.blade.php
+++ b/tests/Unit/views/layouts/app-x.blade.php
@@ -1,0 +1,7 @@
+@yield('content')
+
+{{ $slot }}
+
+{{ $foo }}
+
+{{ $bar }}


calebporzio:
Sorry @RomainMazB - I don't fully understand what this provides that is distinct and beneficial from the existing `->layout('name', [params])`

Can you expound clearly for me. Maybe show me something that is currently impossible? Thanks!

RomainMazB:
Nothing is technically impossible. This PR fix an issue every one using the new dynamic layout `<x-app-layout>` syntax introduced in L7 and studiously reading the doc will run into.

Plus, when installing Jetstream and using the default layout as a base, x-layout is used in both AppLayout and GuestLayout. Anybody who - like me - install Laravel+Jetstream with the Livewire stack right now will potentially face the issue that I solved in this PR.

Basically: when we are using the x-layout syntax, we are able to initialize the component file with extra data, probably a new private message counter, or a layout primary color for example. The problem is when we are running Livewire components as a full page component, Livewire search for a `layouts/app.blade.php` file as default - which is the same path when using x-layout - and run it. Livewire at the moment doesn't know that the blade file is a component, so it doesn't search for a potential `app/Views/component/Layouts/AppLayout.php` file. It just run the blade file, without initializing the message counter variable or the primary color one and we are facing an undefined variable error.

When I faced this issue, it was frustrating: I was using the layout in at least ten different pages, but only the Livewire ones were failing.

As I said, this issue is solvable without the PR, but the solution, even if it's pretty simple when you understand the Laravel's environment - is not that easy to find for the users who don't know how x-layout and Livewire work individually first. Plus, the way to solve is ugly because we need to manually instantiate the layout class file, get its data and pass it to the layout in every single full-page Livewire component, even if we are just using the default layout.

    // My Layout Class file
    <?php namespace App\View\Components;
    
    use Illuminate\Support\Facades\Auth;
    use Illuminate\View\Component;

    class AppLayout extends Component
    {
        public int $unreadMessagesCount;

        public function __construct()
        {
            /** @var User $user */
            $user = Auth::user();

            $this->unreadMessagesCount = $user->unreadMessagesCount();
        }
    }

   // My Livewire component
    <?php namespace App\Http\Livewire\Messenger;
    
    use App\View\Components\AppLayout;
    use Livewire\Component;
    
    class Messenger extends Component
    {
        public function render()
        {
            $layoutData = (new \App\View\\Component\AppLayout())->data();
           
            $view = view('livewire.messenger.index')->layout('layouts.app', $layoutData);
        }
    }

    // My routes file
    Route::prefix('/messenger')->name('messenger')->group(static function () {
        Route::get('/', 'Livewire\\Messenger\\Messenger')->name('.index');
    });

If we were using a basic - not dynamic - layout with its default name `layouts.app`, livewire would support it without even calling the `->layout()` method.

With my previous PR I tried to add the support of the dynamic layout syntax at the same level that livewire support not-dynamic layout: when using the default layout, we wouldn't need to call the `layout` method, and we just set the extra config `use-x-layout` to true (or it could be the default layout class path with an additional commit to support custom default layout)

This was to me the more appropriate way for the new livewire users to land in: just one line in the "Get started" documentation page could have done the job:
> If you are using dynamic layout, set the use-x-layout config to true into your `config/livewire.php`.

Then users could have read further to use custom layout using the `layout` method.

This PR wasn't merged so I thought about - at least - avoid to manually instantiate the layout class in every Livewire full-page component.

I still prefer my previous one because it's more idiomatic for the new users, easier to teach and learn and cleaner in our code but I'm not the one with the crown here :)

Feel free after this deep explanations if you could positively reconsider my previous PR with some additional stuff I could work on.

calebporzio:
Ok, thanks for explanation.

Because (I think) `<x-dynamic-component` isn't supported in Laravel 7, we can't rely on that for this behavior in Livewire.

The problem is, the following code is used to render layout components in Livewire:
```php
@component($view, $params)
    @slot($slotOrSection)
        {!! $manager->initialDehydrate()->toInitialResponse()->effects['html']; !!}
    @endslot
@endcomponent
```

Because we need a dynamic component, we have to use this syntax which does not support the class portion of component.

So the ideal fix here is that we massage the above code (taken from `src/Macros/livewire-view-component.blade.php`) into something that supports using the FULL class portion of a component.

calebporzio:
If I didn't say it above, I agree this should be supported. Thanks for bringing it up.

RomainMazB:
Indeed you are right, a deep search made me realize L7 doesn't support dynamic component, I was wrong about that.

My search also mentioned that the x-blade syntax is just a wrapper around the @component syntax, so it should be possible to handle this behavior with the basic syntax.

I'll do some investigations and get back here with new commits using only basic syntax.

bilfeldt:
I might be misunderstanding something here, if so then I apologise for bloating your PR @RomainMazB, but I think the use-case described below is also affected by this PR and wanted to just shed some light on it:

Imagine a default Laravel 8 Jetstream installation where you have a lot of pages each in some "group" all having the same header for example. Going with the `Posts` example from the docs, then imagine you build a component `resources\components\posts-layout.blade.php` which uses the default `x-app-layout` like so:

```php
@props(['post'])
<x-app-layout>
    <x-slot name="header">
        {{-- This is something all Posts pages has in the header, maybe the title?, for now let us just dump it --}}
        @dump($post)
    </x-slot>

    {{ $slot }}
</x-app-layout>
```

I would based on the docs expect I could render [a full page Livewire component](https://laravel-livewire.com/docs/2.x/rendering-components#page-components) using:

```php
{
    public Post $post; // This uses route model binding
    ...
    public function render()
    {
        return view('livewire.show-posts', ['post' => $this->post])
            ->layout('components.posts-layout');
    }
}
```

But the layout specified is as far as I can see not interpreted like a blade component.

The error if using an [anonymous component](https://laravel.com/docs/8.x/blade#anonymous-components) as above is:

```php
Undefined variable: attributes
```
while the error when implementing a component php class as well becomes:
```php
Undefined variable: post
```

Las option is to render using the component:

```php
    public function render()
    {
        return view('livewire.show-posts', ['post' => $this->post])
            ->layout(\App\View\Components\PostsLayout::class);
    }
```
which returns a `View [App\View\Components\PostsLayout] not found.` as expected.

Would this PR solve this as well - or am I simply missing something here and this does not match the PR? 😄

RomainMazB:
@bilfeldt Hi!

Yes, you are right, this PR would fix the issue described.

I have a basic workaround for now if you need it, but I haven't tested it with props/route binding:
```
public function render()
{
    // Initialize the posts layout to get its data
    // API Documentation here: https://laravel.com/api/8.x/Illuminate/View/Component.html#method_data
    $layoutData = (new \App\View\\Component\PostsLayout())->data();

    return view('livewire.show-posts', ['post' => $this->post])
        ->layout('components.posts-layout', $layoutData);
}
```

RomainMazB:
Ok, it sounds like I've lost myself in the initial PR, the needed changes was way simpler than what I thought.

We just need to detect the dynamic layout in the ViewMacros class, initialize the layout's class merge the passed data with the layout data.

I'd like to write test for this new feature but I can't run the test on my local machine.

RomainMazB:
Reading my code after a good night and I'm not sure this code should live into ViewMacro, maybe the same code into Component class would be more appropriate, just let me know your thoughts @calebporzio , I'll modify the PR.

calebporzio:
Thank you for this @RomainMazB - can you add tests for this functionality as well?

RomainMazB:
Yep, I had some trouble to run the tests on my machine but someone gave me the solution on the Discord canal.

Will write them asap.

calebporzio:
This is great @RomainMazB - thanks so much for adding those tests!

Liam-PVTL:
Great work @RomainMazB, thank you!

PR #2502, opened 2021-02-15
Dynamic livewire components

joshhanley:
This PR adds dynamic Livewire component support to new component syntax.

This will now be possible (the same syntax as blade dynamic components https://laravel.com/docs/8.x/blade#dynamic-components):
```html
<livewire:dynamic-component component="counter" />
```

It also supports component attribute as an evaluated expression:
```html
<livewire:dynamic-component :component="'counter-'.$loop->index" />
```

There were no tests for the Livewire tag compiler, so I have added them and added tests for dynamic component tag compilation.

Hope this helps!

### DIFF
--- a/src/Exceptions/ComponentAttributeMissingOnDynamicComponentException.php
+++ b/src/Exceptions/ComponentAttributeMissingOnDynamicComponentException.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Livewire\Exceptions;
+
+class ComponentAttributeMissingOnDynamicComponentException extends \Exception
+{
+    use BypassViewHandler;
+
+    public function __construct()
+    {
+        parent::__construct('Dynamic component tag is missing component attribute.');
+    }
+}
--- a/src/LivewireTagCompiler.php
+++ b/src/LivewireTagCompiler.php
@@ -3,6 +3,7 @@
 namespace Livewire;
 
 use Illuminate\View\Compilers\ComponentTagCompiler;
+use Livewire\Exceptions\ComponentAttributeMissingOnDynamicComponentException;
 
 class LivewireTagCompiler extends ComponentTagCompiler
 {
@@ -46,10 +47,25 @@ class LivewireTagCompiler extends ComponentTagCompiler
                 return [(string) str($key)->camel() => $value];
             })->toArray();
 
-            if ($matches[1] === 'styles') return '@livewireStyles';
-            if ($matches[1] === 'scripts') return '@livewireScripts';
+            $component = $matches[1];
 
-            return $this->componentString($matches[1], $attributes);
+            if ($component === 'styles') return '@livewireStyles';
+            if ($component === 'scripts') return '@livewireScripts';
+            if ($component === 'dynamic-component') {
+                if(! isset($attributes['component'])) {
+                    throw new ComponentAttributeMissingOnDynamicComponentException;
+                }
+
+                // Does not need quotes as resolved with quotes already.
+                $component = $attributes['component'];
+
+                unset($attributes['component']);
+            } else {
+                // Add single quotes to the component name to compile it as string in quotes
+                $component = "'{$component}'";
+            }
+
+            return $this->componentString($component, $attributes);
         }, $value);
     }
 
@@ -59,9 +75,9 @@ class LivewireTagCompiler extends ComponentTagCompiler
             $key = $attributes['key'];
             unset($attributes['key']);
 
-            return "@livewire('{$component}', [".$this->attributesToString($attributes, $escapeBound = false)."], key({$key}))";
+            return "@livewire({$component}, [".$this->attributesToString($attributes, $escapeBound = false)."], key({$key}))";
         }
 
-        return "@livewire('{$component}', [".$this->attributesToString($attributes, $escapeBound = false).'])';
+        return "@livewire({$component}, [".$this->attributesToString($attributes, $escapeBound = false).'])';
     }
 }
--- a/src/LivewireTagCompiler.php
+++ b/src/LivewireTagCompiler.php
@@ -51,7 +51,7 @@ class LivewireTagCompiler extends ComponentTagCompiler
 
             if ($component === 'styles') return '@livewireStyles';
             if ($component === 'scripts') return '@livewireScripts';
-            if ($component === 'dynamic-component') {
+            if ($component === 'dynamic-component' || $component === 'is') {
                 if(! isset($attributes['component'])) {
                     throw new ComponentAttributeMissingOnDynamicComponentException;
                 }
--- a/src/LivewireTagCompiler.php
+++ b/src/LivewireTagCompiler.php
@@ -53,6 +53,15 @@ class LivewireTagCompiler extends ComponentTagCompiler
             if ($component === 'scripts') return '@livewireScripts';
             if ($component === 'dynamic-component' || $component === 'is') {
                 if(! isset($attributes['component'])) {
+                    $dynamicComponentExists = rescue(function() use ($component, $attributes) {
+                        // Need to run this in rescue otherwise running this during a test causes Livewire directory not found exception
+                        return $component === 'dynamic-component' && app('livewire')->getClass('dynamic-component');
+                    });
+
+                    if($dynamicComponentExists) {
+                        return $this->componentString("'{$component}'", $attributes);
+                    }
+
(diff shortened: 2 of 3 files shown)


austenc:
Great PR dude!

ju5t:
This is exactly what I know I need, but hadn't looked into yet. Awesome job!

calebporzio:
This is great @joshhanley 

How do you feel about also adding `<livewire:is :component="...">` like Vue?

I'm cool with just dynamic-component for Blade parity, I just fancy `is` myself. Thoughts?

joshhanley:
@calebporzio yup I like it! Much shorter. Would you be happy to support both? Should just be a simple check. Let me know and I'll fix it up.

elfeffe:
And what about passing values for properties?
I have a project where I need to load dynamic components but also pass the values saved in database.
Is it possible?

joshhanley:
@elfeffe that still works that same as normal 🙂

ryangjchandler:
@calebporzio I'm all for this being in core, but there is the possibility of breaking changes if people have already created a component called `DynamicComponent`, I'm one of those people...

calebporzio:
Ah interesting @ryan

calebporzio:
@joshhanley - I'm down to support both "dynamic-component" and "is". That'd be great, thanks!

ryangjchandler:
@calebporzio My only objection would be upgrade reliablity - my opinion is that I can upgrade to a new patch version, or even a minor version, with no breaks in my code (assuming the project adheres to SemVer).

stancl:
FWIW if this should bring breaking changes, it might be worth targeting 3.x, since this just adds syntax and not new functionality (you can use `@livewire($component, $data)`).

joshhanley:
@calebporzio I've updated it so it also supports `<livewire:is :component="..." />` syntax.

I did have a quick look and Vue's syntax is actually `<component :is="..." />` not sure if that would be better, as no one should have a component called `component`. We could only use that for now, and then merge the blade syntax into V3 instead?

Another option is I could put a check in to see if a component of `dynamic-component` already exists, if so, use that as priority. That way we can essentially soft launch this feature and remove that check in V3.

Thoughts?

https://v3.vuejs.org/guide/component-dynamic-async.html#dynamic-components-with-keep-alive

stancl:
> Another option is I could put a check in to see if a component ... already exists

If Caleb wants to merge this in 2.x I think this is the right solution. It adds the feature to people for whom it can't cause any issues, and it won't affect people who have the component already. Since this is just syntactic sugar.

Dunno if LW uses some practice for this but in my larger OSS projects I just annotate these bits of code and then deal with them when making a new major version. Kinda the same purpose as the RenameMe namespace I guess

elfeffe:
@joshhanley 
I didn't explain.
It will be interesting if you can pass a bag of properties.

I have blocks, where the user can choose the block (which is a component), and where each component has its own properties. And you don't know that.
I'm using this now:

```
$livewireBlock = Livewire::mount('component-1', $properties);
$result[] = $livewireBlock->html();
```


Then I echo each $result using Blade::compileString().

This is not the best option, I know. But is the only way to load components when you don't know which properties they will have.
I don't know if the Blade components have the option to pass an array with properties, instead of each property by its name.
But it will be great if you add the option.
Did I explain it? My English is not perfect.

stancl:
> But is the only way to load components when you don't know which properties they will have.

Why not

```
@livewire('component-1', $properties)
```

elfeffe:
> > But is the only way to load components when you don't know which properties they will have.
> 
> Why not
> 
> ```
> @livewire('component-1', $properties)
> ```

I had that commented in my code, I don't really remember why it didn't work as expected. I will try to reproduce it and see if it works.
Thank you

jehadassaf:
Any plans to merge this in V2? We desperately need this functionality...

stancl:
@jehadassaf You desperately need pretty syntax?

calebporzio:
Hey @joshhanley - just getting back to this now, sorry for the lag on this.

I think adding the check for existing components named "dynamic-component" would be great to avoid breakages. Are you down for that?

Thanks again for such a well-formed PR as always!

stancl:
@calebporzio FWIW do you have some system to keep track of things that should be changed in the next major version? Like your `RenameMe` namespace.

Here Josh could add something like `// todo-major` and you could go through these things when you work on the next major release (since removing the dynamic-component existence check is a thing that'd probably be wanted in a major release).

calebporzio:
Good call @stancl , there is not.

The bottom of the readme is probably the best place for something like that

joshhanley:
@calebporzio yep happy to add that in. Should be straight forward as it will only be a check for a component called `dynamic-component` in the root of the Livewire folder.

calebporzio:
Awesome, thanks!

joshhanley:
@calebporzio have added a check for an existing component called `dynamic-component`.

calebporzio:
Thanks so much Josh!

elfeffe:
There is no way to pass a bag of $properties, or am I wrong?
You can pass an array of unknown properties if you use:

@livewire($component, $properties)

But there is no way to do the same with this component, one of the features of a dynamic component is that its properties can be unknown, or dynamic.

Is there any way to pass an array of $properties? 
If not, I think we can just use the normal @livewire($component, $data)
Anyway in some cases I'm getting a problem with a null fingerprint using @livewire($component, $data)

stancl:
Blade supports `<x-foo :attributes="$myCustomComponentAttributeBagInstance">`. Maybe this could support `<x-foo :props="$myArrayOfProps">` (unless I misread the code and it already does).

orware:
I also would be curious to see if there's an answer to @elfeffe's question above about being able to pass in a list of potentially dynamic properties as well, but overall this seems like a great addition :-)!

SamDeimos:
I have managed to generate a dynamic component that renders any component that we pass to it, Here an example.

https://laravelplayground.com/#/snippets/808ec92b-81cb-4cb7-92c4-bcd2b121258e

when I replicated it in a development it was necessary to add

`$wire.__instance.serverMemo.children = [];` 

listening to the `@showportal` event

```
<div x-data="{open: false}" @showportal.window="
    $wire.__instance.serverMemo.children = [];
    open = true;" @closeportal.window="open = false;">
```
I have been looking for a way to better optimize this logic but I do not achieve the same results.

ju5t:
Sorry to continue on a merged PR, but we have the following code:

```blade
@if ($currentComponent)
<livewire:is :component="$currentComponent" :state="$state" />
@endif
```

`$currentComponent` is dynamic. It can change. `{{ $currentComponent }}` changes; but the component itself doesn't. I'm not sure if this is intentional, I think it is, but I don't know enough on the internals of Livewire to be sure.

@joshhanley is this something that should work or is the current behaviour intentional?

joshhanley:
@ju5t Try a wire key to it `:wire:key="$currentComponent"`

ju5t:
🤯 

I could've/should've known. Thanks for the quick response!

joshhanley:
@ju5t heh yeah it's always wire keys 😬 No worries! 🙂

reppair:
Seems the docs are out of date. Should I create a docs PR to update references of `<livewire:dynamic-component :is="$current" />` to `<livewire:dynamic-component component="counter" />`? 

https://livewire.laravel.com/docs/nesting#dynamic-child-components

joshhanley:
@reppair both should work.

nuernbergerA:
@joshhanley maybe its related to this line https://github.com/livewire/livewire/blob/8c3eb912562248d89ca7f5de59de2b2ecdb25f2e/src/Mechanisms/CompileLivewireTags.php#L67

`app('livewire')->getClass()` should be in v3 `app(ComponentRegistry::class)->getClass()`

joshhanley:
@nuernbergerA ah yeah maybe. I'm surprised tests haven't picked it up 🤔

reppair:
> @reppair both should work.

I found this PR while looking for existing issues on the subject be cause it did not work as documented for me, the it worked with the new syntax.

reppair:
> @reppair both should work.

@joshhanley I found this PR while looking for existing issues on the subject be cause it did not work as documented for me, the it worked with the new syntax.

joshhanley:
@reppair it has now been fixed in V3 🙂 https://github.com/livewire/livewire/commit/7b3e98567033208175be84856d72b237f9c5821f

PR #2503, opened 2021-02-15
added missing use statement - closes #2497

### LINKED ISSUE #2497: Missing `use` operator
https://github.com/livewire/livewire/blob/1a1d43ff365301ae99ed7f0ccf02c553d4eeba03/src/Component.php#L192

`use Illuminate\Database\Eloquent\Collection as EloquentCollection;` is missing in this file, making this line check if value is an instance of `Livewire\EloquentCollection`, which does not exist.

austenc:
#2497 noted that a use statement is missing, this PR adds it in!


calebporzio:
Thanks!

PR #2505, opened 2021-02-15
Allow App namespace in config - closes #2107

### LINKED ISSUE #2107: Changing Default Class Namespace to "App" is getting error in livewire.php config
### Description
I want to change the Default Livewire Directory from `"App\\Http\\Livewire" ` to "App" but it is getting error

### Exact steps to reproduce
Go to livewire.php in config folder.
Change the default value of class_namespace to "App"
and execute the command below

`php artisan livewire:discover`

and this error will show

The "C:\laragon\www\my_project\app/App" directory does not exist.

if I change it to null or empty string, the generated class namespace will also be 

`namespace \Component\Component;`

any work around for this? Thank you.

**Context**
- Livewire version: [e.g. 2.3.1]
- Laravel version: [e.g. 8.3.1]
- Browser: [MS Edge]

austenc:
This was mentioned in #2107 and it seems like something that should work. 

If the configured namespace is the same as the app default, it should now work as expected.


calebporzio:
Thanks Austen! Can we go for more "inline" logic where we strip and add suffixes so we don't have to check for them? Thanks!

theonly27:
Thanks for resolving my issue, waiting for this to be merge in the master.

calebporzio:
Thanks Austen!

theonly27:
Hi, upon checking, when you change the namespace to App, it works in the component namespace but the livewire component finder couldn't  register the components in the cache.
it will throw an error, ``Unable to find component``
am I doing it wrong?

Issue #2482, opened 2021-02-09
 L7: Undefined property: Livewire\CompilerEngineForIgnition::$files 


underdpt:
### Description

On Laravel 7, `Illuminate\View\Engines\PhpEngine` doesn't have a `$files` property, so when Livewire handles an exception on `Livewire\CompilerEngineForIgnition` it throws another exception on line 17:

```php
14    protected function handleViewException(Throwable $e, $obLevel)
15    {
16        if ($this->shouldBypassExceptionForLivewire($e, $obLevel)) {
17            (new PhpEngine($this->files))->handleViewException($e, $obLevel);
18
19            return;
20        }
```

Possible solution, change that line adding a null coalesce to null:

```php
17            (new PhpEngine($this->files ?? null))->handleViewException($e, $obLevel);
```

### Exact steps to reproduce

Start a new Laravel7 project, install livewire and simply try to render a non-existan component.

You should see this:

![imagen](https://user-images.githubusercontent.com/8122137/107403297-3c8fd600-6b05-11eb-8640-9bafab0ca19b.png)

But you will get:

![imagen](https://user-images.githubusercontent.com/8122137/107403111-04889300-6b05-11eb-8dfe-d0e0219b4e68.png)

### Stripped-down, copy-pastable code snippets

**Context**
- Livewire version: v2.3.17
- Laravel version: v7.30.4

habibiazmi123:
i have same problem

andreich1980:
Are you guys sure you called you livewire component correctly? 
For me it was a messed up name: I typed `@livewire('profile-update')` instead of `@livewire('update-profile')`
Make sure your component is under `app\Http\Livewire`

@habibiazmi123 @underdpt

shstkvch:
I'm also encountering this problem

underdpt:
@andreich1980 The problem only arises when Livewire handles an exception. If no exception is thrown there's no issue at all, but if an exception is thrown, then the underlying error is masqueraded by this issue (another exception is thrown when handling the first exception).

austenc:
Thanks for letting us know about this one @underdpt, I've submitted a PR that will fix it.

PR #2510, opened 2021-02-16
Fixed compiler engine support for php 7

austenc:
In https://github.com/livewire/livewire/issues/2482 several users noted that Laravel 7 didn't have a `$files` property available on the `PhpEngine` and underlying `CompilerEngine` class. 

This PR adds a null coalescing operator to fix that problem.


calebporzio:
Thanks!

PR #2514, opened 2021-02-16
Support ->queryString() method in addition to $queryString property

calebporzio:
Support for:

```php
{
    public $foo = 'bar;

    public function queryString()
    {
        return ['foo'];
    }
}
```


calebporzio:
Thanks Caleb!

binaryweavers:
> Thanks Caleb!

Love Caleb! ❤

PR #2528, opened 2021-02-19
Add --test flag to make command

jasonlbeggs:
This PR adds a `--test` flag to the `php artisan make:livewire` (and aliased) commands. When the `--test` flag is passed, the make command will now scaffold a basic test for the component.

If this PR is merged, we should probably go through all the copy, delete, move, etc. commands to make sure they also perform their duties on the test file.

### DIFF
--- a/src/Commands/ComponentParser.php
+++ b/src/Commands/ComponentParser.php
@@ -19,11 +19,14 @@ class ComponentParser
     {
 
         $this->baseClassNamespace = $classNamespace;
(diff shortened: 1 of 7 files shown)


calebporzio:
Thanks @jasonlbeggs! Mind PRing docs for this?

calebporzio:
Also, I think we're good to not add this functionality for copy, move, and delete

jasonlbeggs:
I'll get the docs PR going shortly. Thanks @calebporzio!

PR #2530, opened 2021-02-19
resetExcept func added

hbakouane:
1️⃣ This function will reset all the properties of the subClass but the ones passed in the parameter

3️⃣ N

4️⃣ Sometimes I want to reset all my properties but some ones, and I don't find a way to do it


akiyamaSM (commented):
Ne need to check on the is array. just check on the case if is_string

calebporzio:
Thanks for the contribution. However, in the interest of keeping the API slim, I'm going to close this.

PR #2536, opened 2021-02-19
Fixed TestableLivewire to always return an instance of itself

### LINKED ISSUE #2507: Livewire assertSeeHtml test helper isn't stackable
I want to use this code:

```php
Livewire::test(Create::class, ['resourceSlug' => $resource->slug()])
    ->assertStatus(200)
    ->assertSee('Body')
    ->assertSee('textarea')
    ->assertSeeHtml('rows="3"')
    ->assertDontSeeHtml('rows="6"');
```

But I get this error message:

```
Method Illuminate\Http\Response::assertSeeHtml does not exist.
```

So the `assertSee` helper returns an `Illuminate\Http\Response` instance, so I can't stack `assertSeeHtml` on it.

Fortunately I found a simple workaround:

```php
Livewire::test(Create::class, ['resourceSlug' => $resource->slug()])
    ->assertStatus(200)
    ->assertSee('Body')
    ->assertSee('textarea');

Livewire::test(Create::class, ['resourceSlug' => $resource->slug()])
    ->assertSeeHtml('rows="3"')
    ->assertDontSeeHtml('rows="6"');
```

But it would still be nice if it gets fixed.

Thank you! Jeroen

austenc:
Fixes https://github.com/livewire/livewire/issues/2507 

Having `assertStatus` (or any other assertion not extended by Livewire) and then trying to use Livewire-specific assertions causes an error. This is because `assertStatus` returns an instance of `TestReponse` and not an instance of `TestableLivewire`.

### DIFF
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -247,6 +247,7 @@ class TestableLivewire
             return $this->macroCall($method, $params);
         }
 
-        return $this->lastResponse->$method(...$params);
+        $this->lastResponse->$method(...$params);
+        return $this;
     }
 }
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -248,6 +248,7 @@ class TestableLivewire
         }
 
         $this->lastResponse->$method(...$params);
+        
         return $this;
     }
 }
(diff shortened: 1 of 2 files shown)


austenc:
Tests should pass! Not sure what happened here.

src/Testing/TestableLivewire.php
```diff
@@ -247,6 +247,7 @@ public function __call($method, $params)
             return $this->macroCall($method, $params);
         }
 
-        return $this->lastResponse->$method(...$params);
+        $this->lastResponse->$method(...$params);
```
calebporzio, line 250: Can you add a line break after this and an accompanying test?

austenc: Done!

nuernbergerA (approved):

calebporzio:
Thanks Austen!

PR #2543, opened 2021-02-21
Require Facade\Ignition\IgnitionServiceProvider as a string instead of using ::class

kountouris7:
Hi,
This is a very simple change. 

We are simply requiring IgnitionServiceProvider as a string, because it offers the flexibility of not having to install Ignition to use Livewire.

Thanks for this great package!


calebporzio:
Cool, thanks!

PR #2554, opened 2021-02-23
Update `WithPagination` to allow  `queryString()`

rodrigopedra:
This PR updates `WithPagination` to account for  `queryString()` method introduced in 2.4.0

I didn't create an issue before. I just tested the new 2.4.0 against a project and noticed components using `WithPagination` could benefit from the introduction of `queryString()` method support to avoid using `array_merge` to account for the `page` attribute.

I just noticed the PR (#2514 ) which added `queryString()` support was merged an hour before.

Thanks @calebporzio 

No it doesn't. This is a single improvement to a single trait.

No it doesn't, as  I thought it will be tested by the current test suit. Currently `WithPagination` already merges with `$this->queryString`, so from my assessment of the test suit, no additional tests would be needed.

If I am wrong on this, please let me know and I will try adding some tests.

Please read description under question # 1.

As a code sample, I had this before:

~~~php
<?php

namespace App\Http\Livewire\System\Reports;

use Illuminate\Contracts\View\View;
use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
use Livewire\Component;
use Livewire\WithPagination;

class Sample extends Component
{
    use AuthorizesRequests;
    use WithPagination {
        getQueryString as withPaginationQueryString;
    }

    public string $fromDate = '';
    public string $toDate = '';

    protected array $rules = [
        'fromDate' => ['nullable', 'date_format:Y-m-d', 'before:toDate'],
        'toDate' => ['nullable', 'date_format:Y-m-d', 'after:fromDate'],
    ];

    public function clearDates()
    {
        $this->fromDate = '';
        $this->toDate = '';
    }

    public function render(): View
    {
        // ... not important for the example
    }

    public function getQueryString(): array
    {
        return array_merge([
            'fromDate' => ['except' => ''],
            'toDate' => ['except' => ''],
        ], $this->withPaginationQueryString());
    }
}
~~~

And now I can simplify it to this:

~~~php
<?php

namespace App\Http\Livewire\System\Reports;

use Illuminate\Contracts\View\View;
use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
use Livewire\Component;
use Livewire\WithPagination;

class Sample extends Component
{
    use AuthorizesRequests;
    use WithPagination;

    public string $fromDate = '';
    public string $toDate = '';

    protected array $rules = [
        'fromDate' => ['nullable', 'date_format:Y-m-d', 'before:toDate'],
        'toDate' => ['nullable', 'date_format:Y-m-d', 'after:fromDate'],
    ];

    public function clearDates()
    {
        $this->fromDate = '';
        $this->toDate = '';
    }

    public function render(): View
    {
        // ... not important for the example
    }

    public function queryString(): array
    {
        return [
            'fromDate' => ['except' => ''],
            'toDate' => ['except' => ''],
        ];
    }
}
~~~

After this PR I won't need to know the inner working of how `Component` or `WithPagination` handle the query strings.

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -10,7 +10,11 @@ trait WithPagination
 
     public function getQueryString()
     {
-        return array_merge(['page' => ['except' => 1]], $this->queryString);
+        $queryString = method_exists($this, 'queryString')
+            ? $this->queryString()
+            : $this->queryString;
+
+        return array_merge(['page' => ['except' => 1]], $queryString);
     }
 
     public function initializeWithPagination()


rodrigopedra:
One additional note:

In the project I tested the new setup I actually have a `DataTable` trait which uses `WithPagination` and adds `$search` and `$perPage` query parameters.

I then use this trait in some other components, some of which needs to add additional filters, that is why in my use-case I am overriding the trait's `getQueryString` instead of defining a `$queryString` property.

I tested those components with the code proposed in this PR, and the proposed solution works and simplifies the code in a fashion similar to the code sample above.

calebporzio:
Awesome, this is great. Thanks!

PR #2560, opened 2021-02-24
Disable poll when hidden

ju5t:
This PR adds a polling modifier that limits polling when an element is not in the current viewport. It is used like this:

```blade
<div wire:poll.visible>
    {{ now() }}
</div>
```

The `visible` modifier tells Livewire to only poll visible elements. If the element is not visible, for example when a user has scrolled it out of view, the element ~~is only being refreshed 5% of the time~~ is not refreshed at all.

Another example, in case you want to change the polling frequency:

```blade
<div wire:poll.visible.1000ms>
    {{ now() }}
</div>
```

This solves discussion [2233](https://github.com/livewire/livewire/discussions/2233). It is the same PR as https://github.com/livewire/livewire/pull/2372 but it comes with a Dusk test.

More than happy to make changes but please note my time is limited and I may not respond quickly.

Edit: tests are failing because I didn't commit the build of livewire.js. I am not used to do this because it almost always causes merge conflicts.


joshuadegier:
Once again thanks for your efforts! I wish I could be of more use besides just showing my gratitude.

ju5t:
@joshuadegier no worries.

@calebporzio just so you know, I discussed the issue of not having a valid build in a PR with @joshhanley a few days ago. Not sure if he will be looking at it, but what would be great is to have CI create a build prior to testing it all. Then, when the PR is merged, have another step that builds the latest version and commits it automatically in the main branch.

In this case, if you want me to commit the build in this PR, let me know.

calebporzio:
Great, thanks for the tests!

Question now that I'm looking at it, I think it makes sense to block 100% of polls when the element is not visible (instead of the current 95%).

That's at least what I would intuit for this behavior.

I like the CI asset build idea. Only gap I can see is the following scenario:
- someone PRs something to Livewire's JS without building
- CI runs the build and passes
- I merge, then tag the release because "master" is green and passing
- The new JS never gets shipped because the bundle didn't get committed

Feel free to open a new PR to continue that discussion. Thanks!

ju5t:
> Question now that I'm looking at it, I think it makes sense to block 100% of polls when the 
> element is not visible (instead of the current 95%).

I did that to have the same behaviour as when a tab is inactive. I don't think anyone has any objections to change that to 100%. Let me know what you'd like and I'll change it.

About CI, when you merge a PR that should trigger the build to create a bundle. I will talk to @joshhanley to see if he did some work on it yet and if not, I will set up a test repo to try a few things, so you can see how it would work. When that's finished I'll open a PR.

joshuadegier:
I'm with @ju5t at this one: blocking about 95% of the requests seem logic to me since that's what the inactive tab feature does as well. This makes this a solution that both problems have in common. But I would have nothing against making it 100% either.

ju5t:
@calebporzio your Laracon talk got me thinking about this PR again. Great talk by the way. Even for those who are all in like ourselves, it was useful. Made me think about using smaller components every now and then.

Ok, back to the PR. I agree that it is more intuitive to have it not poll at all if the keyword is set. 

I will commit the change tomorrow.

calebporzio:
Great, thanks for that adjustment.

2 notes:
A) The tests seem to be failing on your new test.
B) (this can just be a note for myself if you don't get to this) The `isInViewport()` check should be AFTER the `.includes('visible')` so we're only doing that expensive browser calculation when `.visible` is present.

Thanks for the contribution, and sorry for the lag on this.

ju5t:
@calebporzio the tests running in Github actions fail because I didn't commit the build. If you like I can include that, I didn't because it tends to cause merge conflicts :)

I know I promised to look into running those builds prior to the actual test in the action, I didn't get to it yet.

You're right about the if-statement, I switched them around.

joshhanley:
@ju5t yeah I would build the assets (after any changes) and then commit them. Or also merge in master, then build and commit (that's what I typically do). 

If I run into merge conflicts, what I typically do is stage all actual code changes and then run a build and stage the results, and merge it all in.

I haven't had a chance to look at GH actions for building yet either.

ju5t:
No problem. It's still on my list to run the build during CI and only commit the finished build on the master branch. I hope to get to it in the next few weeks.

Until then, everything's green!

calebporzio:
Great, thanks!

joshuadegier:
Nice! Thanks to everyone involved. Now it's my turn to keep my promise, still no recommendations for parties to receive the bounty @ju5t?

marispro:
Maybe it is possible to make this polling not like with 5% random chance, but with some underlay logic:
for example you have polling every second, that means it will be polled about every 20 seconds.

If poll time is 30 seconds - then request will happen in average once every 10 minutes?

ju5t:
@joshuadegier nope, I don't mind really.

@marispro the 5% has been removed in the final version. I have updated the description. Sorry for the confusion.

joshuadegier:
@ju5t I'm going to donate 250 euro's to www.edukans.nl. A charity to help children get better education chances. Thanks for being to generous!

ju5t:
@joshuadegier awesome! 👍🏽

calebporzio:
Super cool of you both @joshuadegier and @ju5t 👊

PR #2561, opened 2021-02-24
Macro actions

fourstacks:
This PR adds the ability to define action methods on components via macros or mixin classes.

**Use case**

I'm developing a package that will ultimately allow other developers to extend the functionality in a set of Livewire components.  Although the `Livewire\Component` class is `Macroable`, if you try to extend a Component class using `macro(...)` or `mixin(...)`, Livewire will currently throw a `Livewire\Exceptions\MethodNotFoundException` when you try to call this method via something like `wire:click="foo"`.

As the `Component` class is `Macroable` already it feels like this should be allowed behaviour and it would certainly open up options for third party component extension.

**Solution**

This PR takes a simple (hopefully not naive) solution to this by adding a check within the `callMethod` method within the `HandlesActions` concern.  This check ensures that the parent class has the custom `macroCall` trait method and that a macro with the method name exists.  If those conditions are met, the macro is called and the Livewire action is dispatched as it would be for a regular method.

This implementation supports closure-based macros (with or without parameters) as well as mixin class-based macros.  Tests are included.

**Questions/Concerns**

I'm not too sure I fully grok what's happening within the `ImplicitlyBoundMethod` class which is what is called for a regular component method.  Bit short on time to really dig into that but if there's some magic happening there that won't be applied if calling a macro then that might be something that makes this feature less simple.

I'm also slightly torn on whether it's OK to make methods set via a macro higher priority than those that might be 'native' to the component.  For example if a component has a method `foo` and then a `foo` method is set via a macro, the macro one will be called as this is evaluated first.  On balance I think this is fine (sharp knives and all that) but I wanted to flag it.

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -2,17 +2,16 @@
 
 namespace Livewire\ComponentConcerns;
 
-use Illuminate\Support\Traits\Macroable;
-use Livewire\Livewire;
-use Livewire\ImplicitlyBoundMethod;
+use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Database\Eloquent\Model;
+use Livewire\Exceptions\CannotBindToModelDataWithoutValidationRuleException;
 use Livewire\Exceptions\MethodNotFoundException;
+use Livewire\Exceptions\MissingFileUploadsTraitException;
 use Livewire\Exceptions\NonPublicComponentMethodCall;
 use Livewire\Exceptions\PublicPropertyNotFoundException;
-use Livewire\Exceptions\MissingFileUploadsTraitException;
-use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Livewire\HydrationMiddleware\HashDataPropertiesForDirtyDetection;
-use Livewire\Exceptions\CannotBindToModelDataWithoutValidationRuleException;
+use Livewire\ImplicitlyBoundMethod;
+use Livewire\Livewire;
 use function Livewire\str;
 
 trait HandlesActions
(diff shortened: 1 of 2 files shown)


austenc:
Hey @fourstacks, couldn't you use a trait to accomplish this sort of thing instead?

fourstacks:
Hey @austenc, so the use case this is trying to solve is to provide an extension point for components that the end user does not own.  

Primarily this will be cases where the Livewire component is in a package.  The end user of the package doesn't own that component class (so cannot add a trait).  They could extend the component class in their own app though this feels heavy handed when all they may want to do is add a single simple method (especially when the Livewire component base class is `macroable` anyway).  As they don't own the component class themselves, they cannot add a trait to it.

Also inheritance does not solve a scenario where the component is in Package A but both Package B and the consuming app both want to add new functionality to the component.  The end user would have to know to extend from the Package B version (and this breaks down entirely if more than one package wants to add to the Package A component).  Macros solve this very gracefully.

If it helps I can give you my specific use case which is that I'm building team software with an extension API over it that can be used to build addons.  For each given screen, I want to create points at which third party developers (or the consumer) can hook in and add new functionality.  Using macros as a basis for this is a perfect solution whereas inheritance would be pretty nightmarish/impossible!

calebporzio:
Hey @fourstacks, you're right that allowing macros would make Livewire more flexible, however, I think I want to stick to only allowing methods in the class hierarchy that don't exist on the base "Component.php" class (this includes allowing trait methods).

All of this is in the interest of not making methods that people only intend to be executed in a secure environment (the back-end), available for calling from the front-end.

Thanks for the effort and contribution, but going to close this.

fourstacks:
Hey @calebporzio, no problem and understood.  Bit of a shame for my particular project but as I'd not gone too far down the road with Livewire it's early enough to pivot to something like Hotwire instead.  Was having a blast with Livewire though so hopefully I'll have another opportunity to work with it again soon!

PR #2565, opened 2021-02-25
Removes hardcoded disk when in testing environment

kajgies:
When in a test environment, the disk is hardcoded in the code here and file uploads don't work unless you add this disk to your filesystems config.
This fixes that.

If we want custom disks per environment, maybe we should add env() to the config file?


kajgies:
Oh it seems the github testing bot relies on this, I will make an issue for this instead.

Issue #1341, opened 2020-07-30
Disk [tmp-for-tests] does not have a configured driver.


MarcelWeidum:
### Description
I have the `temporary_file_upload` => `disk` specified. But I still get the error with uploading:

`Disk [tmp-for-tests] does not have a configured driver.`.

### Steps to reproduce

**Context**
- Livewire version: [e.g. 1.3.0]
- Laravel version: [e.g. 7.22.4]
- Browser: x (server side)

axit-joost:
I had this same thing when deploying an app to a Windows server. I ended up defining the following as a workaround:

In `config/filesystems.php`:
```php
// below disks => local

        'tmp-for-tests' => [
            'driver' => 'local',
            'root' => storage_path('app/livewire-tmp'),
        ],
```

After which the file upload problems disappeared.

MarcelWeidum:
@axit-joost this is working for me

andreich1980:
@axit-joost I guess it would be `storage_path('app')`, `livewire-tmp` comes from `config/livewire.php` option `temporary_file_upload.directory`

vitoo:
Another workaround is to add this in your test : 

`Storage::fake('tmp-for-tests');`

PR #2571, opened 2021-02-26
Add config to define the temporary file upload disk using tests

### LINKED ISSUE #2566: Disk used for file uploaded is hardcoded in testing environments
The disk used in testing environments is currently hardcoded with no override possible:
FileUploadConfiguration.php Line 27:
```
    public static function disk()
    {
        if (app()->environment('testing')) {
            return 'tmp-for-tests';
        }

        return config('livewire.temporary_file_upload.disk') ?: config('filesystems.default');
    }
```
This means that in a testing environment you have no choice to add a 'tmp-for-tests' disk to your filesystems config, otherwise file uploads will give an error.

I think it would be a good idea to remove this hardcoded requirement and make the disk environment specific in the config using env(),

PhiloNL:
**Is this something that is wanted/needed? Did you create an issue / discussion about it first?**
This will resolve https://github.com/livewire/livewire/issues/2566 and https://github.com/livewire/livewire/issues/1341

**Please include a thorough description of the improvement and reasons why it's useful.**
This will make it easier for people to override the disk to be used when running tests.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -38,7 +38,7 @@ return [
     | the view returned by SomeComponent will be wrapped in "layouts.app"
     |
     */
-    
+
     'layout' => 'layouts.app',
 
     /*
@@ -92,6 +92,7 @@ return [
             'jpg', 'jpeg', 'mpga', 'webp', 'wma',
         ],
         'max_upload_time' => 5, // Max duration (in minutes) before an upload gets invalidated.
+        'test_disk' => null // Overwrite disk to be used during tests, defaults to 'tmp-for-tests'
     ],
 
     /*
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -92,7 +92,7 @@ return [
             'jpg', 'jpeg', 'mpga', 'webp', 'wma',
         ],
         'max_upload_time' => 5, // Max duration (in minutes) before an upload gets invalidated.
-        'test_disk' => null // Overwrite disk to be used during tests, defaults to 'tmp-for-tests'
+        //'test_disk' => 'tmp-for-tests' // Overwrite disk to be used during tests
     ],
 
     /*
--- a/src/FileUploadConfiguration.php
+++ b/src/FileUploadConfiguration.php
@@ -27,7 +27,7 @@ class FileUploadConfiguration
     public static function disk()
     {
         if (app()->environment('testing')) {
-            return 'tmp-for-tests';
+            return config('livewire.temporary_file_upload.test_disk', 'tmp-for-tests');
         }
 
         return config('livewire.temporary_file_upload.disk') ?: config('filesystems.default');


austenc:
@PhiloNL you can probably uncomment the line you added in the config file and just change it to 
`'test_disk' => null`

The config fallback should still work the same way I think.

PhiloNL:
Thanks for the comment @austenc 😄  I actually did so in the first commit https://github.com/livewire/livewire/pull/2571/commits/3b2cc69498d15a35fe7324eb74a2da0dc49b43aa but the tests failed when I did that, unfortunately.

calebporzio:
@PhiloNL,

Ok, I just tested this in a brand new app using the following code: https://laravel-livewire.com/docs/2.x/file-uploads#testing-uploads

I think I'm missing something here. Is the happy path for testing file uploads not working for people? Or is there a custom set up that makes things not work.

Looking for more information. Thanks!

kajgies:
Hello everyone ,

To clarify; The reason this was in the way for us is because we have a testing environment for all our apps, part of our DTAP approach. We use this testing environment to test our own app. We would like to use this testing environment with our own existing drives but currently we are required to create a new drive called "tmp-for-tests" just for livewire if we want file uploads to work, this seems a bit counterintuitive to me.
We currently don't use the Livewire::test tests at all.

Kind regards,

Kaj

joshhanley:
I ran into this issue the other day, what happens is the following error gets logged
> Disk [tmp-for-tests] does not have a configured driver.

But under the hood, Livewire tries to find the disk and if it can't find it, it will fake the `tmp-for-tests` storage disk.
See here https://github.com/livewire/livewire/blob/b647ac63378ea02bdac975f70a4e57aea84bda0b/src/FileUploadConfiguration.php#L15-L21

So even though the error was logged, the tests should still run.

If you don't want the error to appear in the log, you can add this to your tests `Storage::fake('tmp-for-tests');`.

I don't think there is a need to have the test disk be configurable. But maybe we should look at changing it so the error doesn't get thrown when it fails over to use the fake disk anyway.

@PhiloNL @kajgies can one of you see if your tests actually run regardless of the error that is logged? Or is your test actually failing? Because my upload tests pass even though that error is logged.

Hope this helps!

kajgies:
I am a bit confused about these "tests" you are talking about. The problem occurs in any "testing" environment, I don't know about these specific tests.
With the "testing" environment I mean a separate website with a separate database where we have this set in our .env file: APP_ENV=testing

I don't know exactly what kind of tests you are talking about and just wanted to clarify that that's not what I am talking about. We just use the "testing" environment to test our app internally with our team, then when it's ready we push it to an acceptance environment, not to test livewire or anything related to Livewire::test(), and file upload didn't seem to work unless we added the drive.

joshhanley:
@kajgies ah sorry, I've misunderstood.

A testing environment to me, is an environment that runs automated tests, such as GitHub actions. Hence why I was explaining how to get tests running.

I personally would not use `APP_ENV=testing` in the scenario you have described, as the testing env should be for automated tests.

Instead I would set `APP_ENV=staging` or something like that (can be whatever you like). That also means you won't run into the Livewire hardcoded issue.

Hope this helps!

austenc:
Agree with @joshhanley -- the usual convention is `staging` because Laravel assumes `testing` is for running automated tests.

I suggest not using `testing` as you'll be fighting the framework's default configuration for running unit tests. It sounds like what you want is an additional environment that mimics `production` as close as possible.

kajgies:
Oh I wasn't aware testing was used this extensively within Laravel. We actually use `local`, `testing`, `acceptance` and `production` environments, which I thought is quite standard.

austenc:
@kajgies Default Laravel apps make the assumption that `testing` will be used for automated tests, but like most things in the framework, you can configure it differently if you'd like. 

See here: https://github.com/laravel/laravel/blob/8.x/phpunit.xml#L21

However, Livewire is (currently) assuming that `testing` is the environment for running automated tests based on that default Laravel convention. Since this assumption is made within Livewire, the simplest solution would be for you to use another environment name. 

This may be configurable in a future Livewire version, but I wouldn't count on it since most folks reserve `testing` for phpunit.

bkilshaw:
Is anyone able to ELI5 what's happening here? In my scenario my default filesystem disk is s3, livewires temporary_file_upload.disk is s3, in my test I `Storage::fake('s3'), but when I run my tests I still receive: `Disk [tmp-for-tests] does not have a configured driver.`

Am I missing something or is this expected behaviour?

ohnotnow:
Your 'real' code will work.  Just for tests you're out of luck.  I only looked into it as far as needing to fake my actual storage disk - and that's not possible.  I've not tried faking livewire's own test disk.

bkilshaw:
I'm just confused why they give an example of faking a disk in the docs if it gets ignored during the test: https://laravel-livewire.com/docs/2.x/file-uploads#testing-uploads

The docs demo is:
```php
/** @test */
public function can_upload_photo()
{
    Storage::fake('avatars');

    $file = UploadedFile::fake()->image('avatar.png');

    Livewire::test(UploadPhoto::class)
        ->set('photo', $file)
        ->call('upload', 'uploaded-avatar.png');

    Storage::disk('avatars')->assertExists('uploaded-avatar.png');
}
```

Issue #2578, opened 2021-02-28
Possible solution for some diffing / patching issues


ncharalampidis:
I was experiencing strange behaviour when toggling a checkbox and I realised that the problem was because of multiple events attached to the element.

After playing around with the livewire's js files I found out that even when I had set an explicit wire:key to the checkbox, the morphdom's "onNodeAdded" event was called and that caused the duplicate event to be registered.

I found out this [bug fix](https://github.com/patrick-steele-idem/morphdom/pull/194/files) in the original morphdom repository and when I applied it to livewire the problems gone.

Can someone with deep knowledge on livewire check it out? It may be a bug fix that should be merged.

austenc:
Hi @ncharalampidis I'm curious if you were using Alpine as well when you saw this behavior?

ncharalampidis:
No it is not related to Alpine. Check the pull request #2580. I added a test that reproduces the issue.

austenc:
Was investigating another issue last week and wanted to double check. Thank you!

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

PR #2580, opened 2021-02-28
Remove onNodeAdded event call when matched keyed node moved to new parent

ncharalampidis:
This bug fix has been solved in the original morphdom project. You can see the original pull request [here](https://github.com/patrick-steele-idem/morphdom/pull/194).

It fixes multiple event registration in an keyed element mentioned in this issue #2578 

The pull request includes the test to reproduce the issue.


calebporzio:
Thanks a bunch for catching this and referencing the original fix!

Only thought here: why did you remove the `@livewireModification` - that's the only part of this that makes me nervous.

ncharalampidis:
The @livewireModification had the following comment:
`Otherwise, "curChild" (matched element by key) will be unattached when it is passed to "handleNodeAdded`

With this fix the matched element is never passed to handleNodeAdded event, which is exactly the fix. I think the @livewireModification was there to somehow skip the bug that morphdom had.

Also all the tests pass, so I guess its ok but please take a closer look yourself because I am new to the project and no so familiar.

markhuot:
Can confirm I am experiencing a similar bug and this fork resolves the issue. In my particular case I have nested components with an outer `PostList` and an inner `Post` component, for example. When I save changes to the DB inside the `Post` component it triggers a render of the `Post` (to end editing) and then an event triggers a render of the `PostList` too (to update ordering/grouping). The first render completes without issue but after the `PostList` catches the event and tries to re-render itself I end up with a "TypeError: initialData is null" JS error.

You can see an example here, https://laravelplayground.com/#/snippets/4b1273ab-3d73-4d6f-b9ab-5e90eabd300b. To trigger the bug edit "lindsey" and change the family to "stooges" then edit "lindsey" again and change the family back to "bluth."

calebporzio:
Awesome, thanks so much!

PR #2584, opened 2021-03-02
Make tolerant of public class-properties that are added or removed after the initial page-load

code-distortion:
Currently (as of ver 2.4,0), when visiting a page with a Livewire component in it; If a public PHP property is then added to the component,  the user will be shown an `ErrorException` if they continue to interact with it. The same occurs when removing a public property.

Screenshot from 2021-03-03 01-23-27

I believe this scenario is fairly likely to occur at some point in a project when features are worked on and deployed over time.

This PR updates Livewire to be tolerant of this situation. This would give the developer an opportunity to decide how to handle it (they could show the user a message or restart the process for example, or they could ignore it altogether).

Here's an example. Uncomment `$newProp`, or comment out `$initialProp` as you interact with it:

Screenshot from 2021-03-03 01-29-55

``` php
<?php
namespace App\Http\Livewire;

use Livewire\Component;

class TestLivewire extends Component
{
    public string $initialProp = 'val1';
//  public string $newProp = 'val2';
    public int $count = 1;

    public function doSomething(): void
    {
        $this->count++;
    }

    {
        return <<<EOT
<div>
    <p>initialProp: {{ $initialProp ?? '(unset)' }}</p>
    <p>newProp: {{ $newProp ?? '(unset)' }}</p>
    <p>count: {{ $count }}</p>
    <button wire:click="doSomething" class="border-2 px-2 py-0 rounded border-gray-400 bg-gray-100">Go</button>
</div>
EOT;
    }
}
```

> I considered adding further functionality to detect when this happens, and do something when it has (like let the component know so *it* can choose to continue or not). But this could be a separate PR,  and might not be needed anyway (as it can be addressed in user-land).


code-distortion:
I'm going to revisit the tests as I need to add one explicitly for this situation.

There's also broken test that I don't understand the significance of yet: `dependent_props_are_set_via_implicit_binding` which expects to see "*prop:via-route:robert:via-route:bobby*" (but "*prop:via-route:robert:via-parent:bobby*" is there instead).

I haven't been able to get Selenium working yet (requests are returning "*Whoops! The URL specified routes to this help page.*" instead) so I haven't checked the browser tests.

code-distortion:
I've added a browser test that checks the situation when a browser loads a page, the Livewire component is changed, and the browser then makes a subsequent request.

The unit test `dependent_props_are_set_via_implicit_binding` still needs looking in to. I'll have a go of fixing it up, any info that helps point me in the right direction will be appreciated.

calebporzio:
Thanks for being willing to contribute @code-distortion, this looks like a valid issue some might run into if a new property is deployed and existing users are on a site.

Personally this PR scares me as it tampers with parts of the codebase I'm generally cautious about messing with.

I would be open to discussing a completely different solution to problem where Livewire has some sort of system to track deploys and force reloads after a new deploy (kinda like cache busting assets with Laravel mix type of thing).

In the meantime, I'm going to close this. Thanks again!

code-distortion:
Hi Caleb.  Thanks for your thoughts about the problem.

I've [created an issue](https://github.com/livewire/livewire/issues/2718) to hopefully continue discussion about it.

PR #2589, opened 2021-03-03
Move "This page has expired..." message to a method

stancl:

No issue, but it makes LW more suitable for human-facing apps

No

No

This makes localizing the message easier. It's still a bit hacky, i.e. you need to replace the method in a JS file, but it seems like a nice solution that doesn't involve creating an abstraction for localizing frontend language strings.

### DIFF
--- a/js/connection/index.js
+++ b/js/connection/index.js
@@ -12,6 +12,12 @@ export default class Connection {
 
         return componentStore.onErrorCallback(status)
     }
+    
+    showExpiredMessage() {
+        confirm(
+            'This page has expired due to inactivity.\nWould you like to refresh the page?'
+        ) && window.location.reload()
+    }
 
     sendMessage(message) {
         let payload = message.payload()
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks @stancl - does this PR actually provide you value? Like are you using this refactor to enable something that wasn't previously possible?

stancl:
@calebporzio Yeah, you can do something like:

```js
Livewire.connection.showExpiredMessage = () => alert('Vaše relace vypršela, prosím načtěte stránku znova.')
```

Obviously not the perfect solution, but this is one of the few language strings in the JS codebase that the **users** will interact with, so it makes sense to provide _some_ way to customize it.

I went with this solution because a proper localization layer for the frontend would bring significantly more complexity. (Though if I may suggest a solution, having a a simple JS object like in e.g. `Livewire.lang` would work nicely, developers could change language strings in PHP and the `@livewireScripts` directive could add a line that sets `Livewire.lang` to a JSON-encoded version of the Laravel translation strings.)

calebporzio:
Ah gotcha. Ok, I'm good with this. I'm not going to consider this an official API or anything though - so it may be removed in a future patch or version. Thanks!

PR #2592, opened 2021-03-05
Add helpful entangle error message

joshhanley:
Currently if you try to entangle a Livewire property that doesn't exist, you get this error (and it happens multiple times) and Alpine won't run after.

Screen Shot 2021-03-06 at 7 53 58 am

This PR adds a helpful console error message to entangle, so you know which property can't be found, and returns rather than throwing an error so Alpine continues running.

Screen Shot 2021-03-06 at 8 02 33 am

Hope this helps!


calebporzio:
Thanks Josh!

joshhanley:
No worries 🙂

Issue #2591, opened 2021-03-05
validateOnly validates all matching fields in a collection


acurrieclark:
### Description
When using validate only, I would expect that only the specific field would be validated against the rules. For example, if there is a collection of 3 `Reference` models, each with a `name` and `email` field, and the validation rules are:

```php
protected $rules = [
    'references.*.name' => ['required', 'distinct'],
    'references.*.email' => ['required', 'email', 'distinct'],
];
```

then when `$this->validateOnly('references.0.name')` is called, we should only validate the first reference's name. Instead, all 3 reference names will be validated, and an error added for each.

if `$this->validateOnly('references.*.name')` is called, the same behaviour applies, but this would be the expected result, to my mind.

### Exact steps to reproduce

I have made a [repo](https://github.com/acurrieclark/livewire-validation-tests) to reproduce the behaviour. Instructions in the readme:
`git clone git@github.com:acurrieclark/livewire-validation-tests.git` to take a look

**Context**
- Livewire version: 2.4
- Laravel version: 8.12
- Alpine version: n/a
- Browser: Brave/Safari/Firefox

Happy to put together a pull request for this if wanted.

acurrieclark:
I have done a little more investigation into this in the last few hours and drew up a few tests to fit what I am suggesting.

In doing so, I found that there are actually 2 tests which already mark the current behaviour as correct. Specifically:

```
old_deeply_nested_specific_validation_only_is_cleared_if_new_validation_passes
// validates "items.0.baz" and then checks for an error on "items.1.baz"

old_deeply_nested_specific_validation_only_is_cleared_if_new_validation_fails
// validates "items.0.baz" and then checks that an error on "items.1.baz" has cleared.
```

What I am suggesting would mean that both of these tests would fail. So the above suggestion would be a breaking change.

zepfietje:
@acurrieclark, I don't think the tests you linked to should fail, since they call the following method.
https://github.com/livewire/livewire/blob/b647ac63378ea02bdac975f70a4e57aea84bda0b/tests/Unit/ValidationTest.php#L464-L472

This tests exactly what you described here:
> if $this->validateOnly('references.*.name') is called, the same behaviour applies, but this would be the expected result, to my mind.

Without diving into the source I'm not 100% sure what causes this bug, but it's possibly an unintended side effect of https://github.com/livewire/livewire/pull/1615.

[...]

PR #2594, opened 2021-03-06
Failing Test: Fix wildcard model validationOnly()

acurrieclark:
Added failing test for non-wildcard validateOnly
Added passing test for wildcard validateOnly

See discussion in #2591 

No, just the one requested change

Right now, it's only a test

At the moment, when calling `$this->validateOnly('foo.1.bar')`, the validation is actually run against `foo.*.bar`, which I don't think should be the expected behaviour.

I have some thoughts about a fix for this if it is deemed correct.

### DIFF
--- a/tests/Unit/EloquentModelValidationTest.php
+++ b/tests/Unit/EloquentModelValidationTest.php
@@ -185,6 +185,26 @@ class EloquentModelValidationTest extends TestCase
 
         $this->assertEquals([[['name' => 'arise']]], $foo->fresh()->zap);
     }
+
+    /** @test */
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for this PR. How does Laravel itself handle this? Is this something that should be fixed upstream inside Laravel's validation system? I try to keep parity with that as close as possible. Thanks.

acurrieclark:
A good point. I will look into it. I have pencilled a fix, but it adapts the current validation implementation rather than looking at the Laravel system. Will investigate.

calebporzio:
Awesome thanks!

calebporzio:

PR #2597, opened 2021-03-06
Reactive push and prepend directives [experimental feature]

ncharalampidis:
The original idea for this feature came from Vuejs teleport. The ability to push upwards content from a blade file already exists in Laravel (using push and stack directives) but they are not reactive when used inside a Livewire's component.

So I tried to extend the same directives and build on top of them the reactivity.

You need such a feature when you want to make components that sends content to body like this:
```
// dropdown.blade.php
<button wire:click="$toggle('active')"></button>

@if ($active)
@push('body')
<div>{{-- Dropdown content goes here --}}</div>
@endpush
@endif
```

This is a very common case in modals, dropdowns and selectboxes.

It works with nested components too.

I'm not sure if its useful to anyone else but I created the PR anyway.


calebporzio:
Hey @ncharalampidis, thanks for this PR.

This feature is definitely something I've needed myself. Very interesting stuff.

Also, the times I've considered an implementation I've thought along the lines of your backend implementation so good there.

I will say, the parts of this that make me nervous are the backend ViewFactory overriding and the JS stack handling stuff.

I imagine there are lots of ripple effects and implications of an addition like this.

Thank you for bravely charting out a course for this feature, but I think I'm going to close this until at least the next major version and then we can revisit this with maybe a different front-end architecture.

Thanks so much! (I wonder also if this is something you could put into a package and OSS it that way)

All the best!

Issue #2605, opened 2021-03-07
Implicit route-model parameters don't work in query strings


nihilsen:
### Description
The query string functionality supports named route parameters.

However, when these are implicitly bound route-model parameters, it breaks down and the following error is thrown: `ErrorException: "Array to string conversion"`.

This is because we don't convert implicitly bound route-model parameters back into their route key equivalents when we update the query string, so in `Illuminate/Routing/RouteUrlGenerator::replaceNamedParameters()` we try to pass the model directly, which fails.

### Exact steps to reproduce
- Use any implicit route model parameters in your Livewire component.
- Include that in your `$queryString` property.
- Attempt to update the property.

**Context**
- Livewire version: [e.g. 2.2.3]
- Laravel version: [e.g. 8.3.1]
- Alpine version: [e.g. 2.3.0]
- Browser: [e.g. Chrome, Safari]

nihilsen:
It turns out this issue affects only optional route parameters, and specifically going from changing an optional route parameter from null to not-null. Otherwise, history sync of route parameters appears to be working fine.

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

PR #2607, opened 2021-03-07
Fix route model parameters in query string

Persaeus:
I was surprised to discover that the query string functionality supports named route parameters seamlessly. Awesome!

However, it appears that implicitly bound route model parameters don't work in the query string. (#2605)

If you try to use such parameters in the query string, the following `ErrorException` will be thrown: `Array to string conversion` (`from vendor/laravel/framework/src/Illuminate/Routing/RouteUrlGenerator.php:220`; `Illuminate/Routing/RouteUrlGenerator::replaceNamedParameters()`).

The reason is that model parameters are not converted back into their route key equivalents for the query string.

This PR fixes that by making sure any query string parameters that are `UrlRoutable` are converted into their route key equivalents for the query string.

Let me know if tests are necessary, and I will look into adding those.

### DIFF
--- a/src/RenameMe/SupportBrowserHistory.php
+++ b/src/RenameMe/SupportBrowserHistory.php
@@ -40,7 +40,12 @@ class SupportBrowserHistory
         Livewire::listen('component.dehydrate.initial', function (Component $component, Response $response) {
             if (! $this->shouldSendPath($component)) return;
 
-            $queryParams = $this->mergeComponentPropertiesWithExistingQueryParamsFromOtherComponentsAndTheRequest($component);
+            $route = app('router')->current();
+            $queryParams = $this->mergeComponentPropertiesWithExistingQueryParamsFromOtherComponentsAndTheRequest($component)
+                // We need to filter any route parameters so they won't get embedded in the query string.
+                ->filter(function ($parameter, $key) use ($route) {
+                    return ! $route->hasParameter($key);
+                });
 
             $response->effects['path'] = url()->current().$this->stringifyQueryParams($queryParams);
         });
@@ -131,7 +136,7 @@ class SupportBrowserHistory
             )
         );
 
-        return app(UrlGenerator::class)->toRoute($route, $boundParameters + $queryString->toArray(), true);
+        return app(UrlGenerator::class)->toRoute($route, $boundParameters + $queryString->all(), true);
     }
 
     protected function mergeComponentPropertiesWithExistingQueryParamsFromOtherComponentsAndTheRequest($component)


calebporzio:
Thanks for the contribution @nihilsen.

Would you mind adding a Dusk test for this. It's very helpful for the future of the feature AND as a way to glance at the problem this actually solves. Thanks!

nuernbergerA:
Hey @nihilsen

Test can be found here https://www.github.com/livewire/livewire/tree/master/tests%2FBrowser%2FSyncHistory%2FTest.php

Routes are defined in the `TestCase`

nihilsen:
You got it. I will look into getting that added.

nihilsen:
I made a couple of changes from the original PR.

It occurred to me that it should not be necessary to convert `UrlRoutable` parameters into their route key equivalents, as that should already be taken care of by `Illuminate\Routing\UrlGenerator::toRoute()`, which we use in `SupportBrowserHistory::buildPathFromRoute()`.

https://github.com/livewire/livewire/blob/6bb2f45fe0f4a372ecc2e7bbb12d91ced63a9622/src/RenameMe/SupportBrowserHistory.php#L134

The problem, as I see it, is that for we use `$queryString->toArray()` when we should probably be using `$queryString->all()`, because `Collection::toArray` converts any Eloquent models in the collection to arrays which prevents `UrlGenerator::formatParameters()` from doing its job.

---

Additionally, on initial dehydration we use `url()->current().$this->stringifyQueryParams($queryParams)`. https://github.com/livewire/livewire/blob/6bb2f45fe0f4a372ecc2e7bbb12d91ced63a9622/src/RenameMe/SupportBrowserHistory.php#L45

This assumes that all parameters are query string parameters. Any route parameters, e.g. eloquent models, get converted to arrays and sent directly to the query string.

This is not an issue for subsequent dehydration, as this is where we use `buildPathFromRoute()` which does handle route parameters.

https://github.com/livewire/livewire/blob/6bb2f45fe0f4a372ecc2e7bbb12d91ced63a9622/src/RenameMe/SupportBrowserHistory.php#L58

I initially thought, why not use `buildPathFromRoute()` for the initial query string like we do for subsequent query string updates, as in something like the following:

```PHP
Livewire::listen('component.dehydrate.initial', function (Component $component, Response $response) {
    if (! $this->shouldSendPath($component)) return;

    $queryParams = $this->mergeComponentPropertiesWithExistingQueryParamsFromOtherComponentsAndTheRequest($component);

    $route = app('router')->current();

    $response->effects['path'] = $this->buildPathFromRoute($component, $route, $queryParams);
});
```

However, this caused a number of unit tests to fail for reasons I don't fully understand.

A different approach for which tests don't fail, is to filter any route parameters for the query string during initial dehydration:

https://github.com/livewire/livewire/blob/17e56dc1bf1d4772bf8d2800aade80128f790a5a/src/RenameMe/SupportBrowserHistory.php#L43-L49

nihilsen:
After deeper examination, I see now that this is not an issue affecting route parameters generally, but more specifically optional route parameters. I will close this PR and the accompanying issue. A better PR with tests is underway.

PR #2618, opened 2021-03-10
fix $wire.emit, $wire.emitUp and $wire.emitTo (fixes #2498)

### LINKED ISSUE #2498: $wire.emit gives TypeError: this.listeners.call is undefined
**Context**
- Livewire version: 2.3.17
- Laravel version: 8.27.0
- Alpine version: 2.8.0
- Browser: Firefox 85.0.1 (64-bit)]
 
### Description

When adding a component with:
```
<div x-data="{}">
    <button @click="$wire.emit('testEvent')">Test $wire.emit</button>
</div>
```
Clicking the element gives the following error:
```
Uncaught (in promise) TypeError: this.listeners.call is undefined
    emit Store.js:51
```

This issue is solved by using
```
<div x-data="{}">
    <button @click="Livewire.emit('testEvent')">Test $wire.emit</button>
</div>
```

The [events](https://laravel-livewire.com/docs/2.x/events#firing-events) documentation section indicate that you should use `Livewire.emit` in this case
But in the [alpine](https://laravel-livewire.com/docs/2.x/alpine-js) integration documentation you can read in the `$wire` API section:
```
// Emitting a Livewire event called "some-event" with two parameters
$wire.emit('some-event', 'foo', 'bar')
```

If `$wire.emit` cannot be used in this case, what is it made for ?

claytonrcarter:

Issue exists at https://github.com/livewire/livewire/issues/2498, but appears to be larger than what is reported there. I ran into this while playing w/ LW and trying to use `@click="$wire.emitUp('sortBy', 'column_name')`. This caused a `this.componentsListeningForEventThatAreTreeAncestors is undefined` error. Changing to `$wire.emit` changed the error to `this.listeners.call is undefined`, as noted in #2498, and `$wire.emitTo` was causing `this.getComponentsByName is undefined`. Basically, `this` was being bound incorrectly when forwarding these calls to the store. This change corrects that by removing the `.apply(component, args)` code and simply calling those methods directly on the store via `store[property](...args)`.

Also, it seems that `emitUp` should be receiving special handling like `emitSelf` (because it needs to receive the component element), but wasn't, so I have added a case to handle that.

Please note that this is my first PR here, and I've only been using LW for ~1 week, so I may just be me misunderstanding things, but this change was enough to get me up and running again w/ `$wire.emit`, `$wire.emitUp` and `$wire.emitTo`.

No, all of this code was needed to getting `$wire.emit*` working.

Yes. I have added a dusk test to do a basic check of each of the `$wire.emit*` methods. Before this change, only `$wire.emitSelf` appears to be working. W/ this change, all of the `$wire.emit*` methods are working.

Please note that I've only tested this in Safari as I can't seem to get Chrome to work w/ Dusk.

Check: see 1️⃣  😄 

Thank you for Livewire!


claytonrcarter:
For those playing along at home, it looks like other PRs here are committing the build artifacts, so I've just pushed a commit that includes them here as well. Hope that's correct; easy to revert if not.

calebporzio:
Thanks so much for this @claytonrcarter! And for the thorough tests!

PR #2625, opened 2021-03-13
Failing Test: Add failing loop clicks to alpine click handlers test

joshhanley:
The alpine click handlers within a blade if statement was fixed in PR #2434 but I have found that the same issue is still happening in foreach loops.

This PR adds loops to the test which now makes it a failing test.

When I get a chance I will look into fixing.

Hope this helps!


calebporzio:
Thanks Josh!

calebporzio:

PR #2628, opened 2021-03-15
Fix: Paginator footer text

mertasan:
Previous codes were not compatible for all languages.

For example, in my language (Turkish), the places of words are different.

English: Showing 1 to 2 of 4 results
Turkish: Toplam 4 sonuç arasından 1 ile 2 arasındaki sonuçlar gösteriliyor


calebporzio:
I believe this is basically a copy of what Laravel's Tailwind pagination template uses. Can you confirm this problem exists in both? I'd like to keep Livewire's template in-step with Laravel's as much as possible.

If Laravel has this problem as well, can you attempt to PR it there and if it goes through PR it here?

Also, for uncommon circumstances like yours, you can use a custom pagination template.

Thank you for putting in the work to help others by contributing, but I'm going to close this for now.

PR #2633, opened 2021-03-16
Make LivewireManager singleton injectable to other services

vdauchy:
This is something necessary to use dependency injection via constructors.

No, this is only about making LivewireManager injectable to other components/services.

No as this is already tested extensively.

Currently the `LivewireManager::class` is linked to 'livewire' key which is not a class. This means that we cannot use type to inject a LivewireManager singleton instance to an other service as this will fail:

```
class MyService {
    public function __construct(LivewireManager $livewire) {
        assert($livewire === app('livewire'), "Not the same manager");
    }
}
app()->make(MyService::class);
```

To correct this `LivewireManager::class` must first be set as singleton and then Aliased.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -82,7 +82,8 @@ class LivewireServiceProvider extends ServiceProvider
 
     protected function registerLivewireSingleton()
     {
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

PR #2634, opened 2021-03-16
Add getComponentAliases() to LivewireManager

vdauchy:
This is needed to check if an alias is already set / exists. It mimics the Laravel `BladeCompiler::getClassComponentAliases()` to get access to the current list of aliases.

No, this just a simple accessor.

No, this method is just an accessor to the protected property: `componentAliases`.

Currently to check if an Alias exists we need to do something like:
```
try {
    $exist = boolval(Livewire::getClass($alias));
} catch (Exception $error) {
    $exist = false;
}
```
this PR let you do:
```
$exist = isset(Livewire::getComponentAliases()[$alias]);
```
Mind that I on purpose did not make a more precise method like `aliasExists` to mimic Laravel's component function `BladeCompiler::getClassComponentAliases()` that return the list of aliases and they assigned classes.

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -43,6 +43,11 @@ class LivewireManager
         return $alias === false ? $default : $alias;
     }
 
+    public function getComponentAliases()
+    {
+        return $this->componentAliases;
+    }
+
     public function getClass($alias)
     {
         $finder = app(LivewireComponentsFinder::class);


calebporzio:
Great, thank you @vdauchy!

yusufonur:
When I call the method related to Livewire, I get an 'undefined method' warning. As I understand it is necessary to use the following way, if I am wrong, please correct it;

`(new LivewireManager ()) -> getComponentAliases ()`

vdauchy:
@yusufonur Yes you need to call that method on an instance, don't do 'new ...' as this is a singleton you will miss its configuration, use something like `resolve(LivewireManager::class)` or `app(LivewireManager::class)` or DI in your constructor... or use the facade `Livewire::...` to interact with the singleton like if it was a static method (ie the code in my PR).

yusufonur:
> @vdauchy Thanks for information

Issue #2445, opened 2021-02-03
Getting 404 response on all request on v2.3.11


tarikhagustia:
### Description
After upgrading from 2.3.8 to 2.3.11, all livewire component getting 404 response

**Context**
- Livewire version: [e.g. 2.3.11]
- Laravel version: [e.g.  7.30.4]

joshhanley:
@tarikhagustia is that happening when you go to any endpoint that has a Livewire component, or is it happening on subsequent requests, when you interact with any Livewire components?

Butochnikov:
@joshhanley same problem. this happens after on subsequent requests.

Butochnikov:
404 only if prefix is set `prefix' => LaravelLocalization::setLocale(),`

RJFares:
+1

joshhanley:
@Butochnikov thanks for the info!

## Setup

To replicate this issue, setup a Livewire app with a basic component with a button that triggers a request to the server.

Install `mcamara/laravel-localization` (I'm sure any package with prefixes will do the same).

Setup `mcamara/laravel-localization`  as per instructions, by registering route middleware in http kernel

Add prefix to route

![image](https://user-images.githubusercontent.com/882837/106723089-dc0f1d80-665a-11eb-9ff8-f7afe5459bda.png)

## Action

Browse to the route with a specific language specified like `/en`

Trigger Livewire action and you get the 404.

![image](https://user-images.githubusercontent.com/882837/106723598-60fa3700-665b-11eb-9aa5-fbfdee900e43.png)

## Troubleshooting

So it looks like in HttpConnectionHandler:64 Livewire is trying to do a match of the incoming route against the current listed routes.

So it's trying to match the original route `/en` with the list of routes, but it can't find it as it's a prefix. The route would actually be `/`.

Not sure how to match when prefixed are involved.

Hope this helps!

tarikhagustia:
> 404 only if prefix is set `prefix' => LaravelLocalization::setLocale(),`
> 
> on 2.3.8 version works fine

yes, i'm working with this library, any trick to running latest livewire with this package?

marispro:
Also getting this error, and can confirm that it happens even when you use prefix without any kind of packages.

calebporzio:
Ok I've zeroed in on the problem.

From what I can tell, @marispro, standard prefixes work fine.

It's dynamic prefixes that are iffy.

In this case, when using the localization package, the prefix for the original request route is dynamically set like this: 
[...]

PR #2661, opened 2021-03-24
Remove base url for correct route matching.

haringsrob:

There are a couple of similar reports regarding this issue, or might be related to this.

https://github.com/livewire/livewire/issues/2445
https://github.com/livewire/livewire/pull/1544
https://github.com/livewire/livewire/issues/1535

No, all changes done are just for fixing the issue.

No, I did not :(

We are running our application in a subdirectory. In our .env file our APP_URL looks like `APP_URL=http://localhost:8080/public`

On the url: `http://localhost:8080/public/manage/some-page`

When laravel does route matching, it manages to detect it correctly and says `/manage/some-page` is the path to use for route matching.

However in makeRequestFromUrlAndMethod, the request build is using `/public/manage/some-page` for the matching which will result in a 404.

I am not saying the provided solution is the best, because I am not sure about the implications on various setups. But it does solve our issue.

If you have alternative suggestions on how to handle it, let me know, happy to provide an updated or.

I hope I made all clear.

Thanks in advance and thank you for all the great work on Livewire to all the contributors!

### DIFF
--- a/src/Controllers/HttpConnectionHandler.php
+++ b/src/Controllers/HttpConnectionHandler.php
@@ -62,6 +62,8 @@ class HttpConnectionHandler extends ConnectionHandler
 
     protected function makeRequestFromUrlAndMethod($url, $method = 'GET')
     {
+        $url = str_replace(config('app.url'), '', $url);
+
         $request = Request::create($url, $method);
 
         if ($session = request()->getSession()) {


lanuma:
In my case this patch is working

our staging url: https://staging.example.com/project/some-page

let me know if there is another updates

calebporzio:
Thanks for contributing this @haringsrob. As it effects a major part of the app - would you mind attempting to add tests to this? Thanks!

haringsrob:
Hey @calebporzio, I will try to find some time to do this :)

calebporzio:
Hey @haringsrob  - any word on this? Thanks!

calebporzio:
I understand, have you had a chance to look into adding tests?

masakik:
if just change the default livewire config from `'asset_url' => null,`
to `'asset_url' => config('app.url')`?

jastreich:
Latest version with .env set correctly does indeed seem to make this PR irrelevant.

calebporzio:
Thanks @jastreich, closing this then.

masakik:
Anyway, if change the default value in config should  be good.

PR #2662, opened 2021-03-24
feat: make Livewire default routes configurable and publishable

kslimani:
Yes, i added some comments to #1674.

Nope, the commits only apply to Livewire routes.

Publish command has no unit test (yet ?)

 - It introduce the `routes/livewire.php` file, which contains Livewire default routes
 - It add the `--routes` option to `livewire:publish` artisan command (to publish routes file)
 - It add the `routes` configuration option, if set to `false`, to disable default routes (to use published one)
 - It add `routes_namespace`, `routes_domain` and `routes_prefix` configuration options (all default to null to keep current behaviour)
 - It should __not introduce any breaking changes__ (it only adds optional features)

__Example use case 1 :__

Default behaviour is to execute requests using `/livewire/` path. Let's say i want to add a prefix and execute request using `/acme/livewire` path, all i need is :

Publish configuration file :

```console
php artisan livewire:publish --config
```
Then change `asset_url` and add the `routes_prefix` option in application `config/livewire.php` file :

```php
// [...]
'asset_url' => '/acme'
// [...]
'routes_prefix' => 'acme',
```

Finally since we modified the assets url, don't forget to rebuild Laravel application assets (`npm run dev`).

__Example use case 2 :__

Let's says we want to modify the Livewire default routes : then publish also the routes file :

```console
php artisan livewire:publish --routes
```
Then add the `routes` option (and set value to `false`) in application `config/livewire.php` file :

```php
// [...]
'routes' => false, // Disable Livewire default routes
```

Finally require the published livewire routes file, for example,  in your application `routes/web.php` file :

```php
// [...]
require_once 'livewire.php';
```

At this stade you can modify application `routes/livewire.php` to suit your needs.

---

__Miscellaneous :__

 - The `routes*` options are a proposition, i can rename them if required
 - By default if you run the `livewire:publish` command without option, it __does not publish__ routes files. I made this choice to keep current behaviour. I can modify the command to also to publish routes file if required.

### DIFF
--- a/routes/livewire.php
+++ b/routes/livewire.php
@@ -1,0 +1,28 @@
+<?php
+
+use Illuminate\Support\Facades\Route;
+use Livewire\Controllers\FilePreviewHandler;
+use Livewire\Controllers\FileUploadHandler;
+use Livewire\Controllers\HttpConnectionHandler;
+use Livewire\Controllers\LivewireJavaScriptAssets;
+
+Route::group([
+    'namespace' => config('livewire.routes_namespace', null),
+    'domain' => config('livewire.routes_domain', null),
+    'prefix' => config('livewire.routes_prefix', null),
+], function () {
+    Route::post('/livewire/message/{name}', HttpConnectionHandler::class)
+        ->name('livewire.message')
+        ->middleware(config('livewire.middleware_group', ''));
+
+    Route::post('/livewire/upload-file', [FileUploadHandler::class, 'handle'])
+        ->name('livewire.upload-file')
+        ->middleware(config('livewire.middleware_group', ''));
+
+    Route::get('/livewire/preview-file/{filename}', [FilePreviewHandler::class, 'handle'])
+        ->name('livewire.preview-file')
+        ->middleware(config('livewire.middleware_group', ''));
+
+    Route::get('/livewire/livewire.js', [LivewireJavaScriptAssets::class, 'source']);
+    Route::get('/livewire/livewire.js.map', [LivewireJavaScriptAssets::class, 'maps']);
+});
--- a/routes/livewire.php
+++ b/routes/livewire.php
@@ -11,17 +11,16 @@ Route::group([
     'domain' => config('livewire.routes_domain', null),
     'prefix' => config('livewire.routes_prefix', null),
 ], function () {
-    Route::post('/livewire/message/{name}', HttpConnectionHandler::class)
-        ->name('livewire.message')
-        ->middleware(config('livewire.middleware_group', ''));
+    Route::middleware(config('livewire.middleware_group', ''))->group(function () {
+        Route::post('/livewire/message/{name}', HttpConnectionHandler::class)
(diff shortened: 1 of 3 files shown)


kslimani:
It may also help resolving routing issues like #2445.

calebporzio:
Thank you for this PR, going to hold off on this until there is more demand for such a thing. Seems like a reasonable feature though, so let's keep an eye on demand for it. Thanks for being willing to contribute!

denisdulici:
This is required for the [Default Values](https://laravel.com/docs/8.x/urls#default-values) feature such as locale in URL. In our case with [Akaunting](https://github.com/akaunting/akaunting), we add the tenant id as a prefix to all URLs.

@calebporzio any chance to consider it?

moxx:
This feature is absolutely needed, why was the pull request not accepted? It is a pure improvement!

SEUH:
@calebporzio We need this for Jetstream teams application. Because we use prefixed routes to determine the team, we also need to prefix livewire to ensure that the component does the action in the correct team.

This MR does not change that much code and enables livewire to be much more flexible so we can use it with prefixes and middlewares. We would love to see this in livewire! :)

calebporzio:
I hear you. This PR in its current form means we are locked into not changing the code around the Livewire routes because they would be published in people's apps. If all your after is a prefix, maybe there should be a PR to make that a dedicated config option or something

Arkhas:
Any news about that ? We are using a multi tenancy system with the tenant name in a prefix like : `myapp.com/tenant` and that would be very useful for us :)

schakko:
+1 for a PR. In the meantime I've extended the `LivewireServiceProvider` and excluded it from auto-discovery. With that, the routes can be registered manually in the `routes/web.php`.
I did a quick [write-up about that](https://dreitier.com/knowledge-base/multitenancy-architectures/using-livewire-in-a-multitenant-laravel-application) @Arkhas

MDGDesign:
+1 for this one!
I think this would solve my current problems https://github.com/livewire/livewire/discussions/5731.

MarllonGomes:
+1

gazzoy:
+1

PhillipMwaniki:
+1

joshhanley:
In V3, this is already possible. See here https://livewire.laravel.com/docs/installation#configuring-livewires-update-endpoint

denisdulici:
Thanks @joshhanley for the update. It is far better now. However, I still had to [override a class](https://github.com/akaunting/akaunting/blob/master/overrides/livewire/livewire/Mechanisms/HandleRequests/HandleRequests.php#L28) to make it work.

Do you know the reason for making the URL absolute and starting with a slash?

joshhanley:
@denisdulici hmm not sure. PRs welcome if you think it's not needed 🙂

denisdulici:
Got it. Will do when I find some time 😉

on-future:
+1

PR #2663, opened 2021-03-24
Implement custom page name pagination to deal with multiple component using pagination trait on the same page

RomainMazB:
I guess this could be the mid-step for multiple pagination in single component which is a wanted feature

Nope.

Yes.

When we are dealing with multiple paginated components in a single page, the page query string is failing because both component will share the same page attribute.

Imagine we have a Forum containing a paginated list of thread on the side navbar and in the main content we display a topic with paginated messages.

Both of the two component will share the `?page` attribute.

With this modification, a component could change its page property name to something else so there is no conflict, resulting in something like `?threadsListPage=3&page=2`
```php
class ThreadsList extends Component
{
    use WithPagination;

    public $threadsListPage = 1;

    public function getPageName()
    {
        return 'threadsListPage';
    }

    public function render()
    {
        return view('threads-list', [
           'threads' => Thread::paginate(5, ['*'], 'threadsListPage')
        ]);
    }
}

class MessagesList extends Component
{
    use WithPagination;

    public function render()
    {
        return view('messages-list', [
           'messages' => Message::paginate(5)
        ]);
    }
}
```

_The next step as I know it has already been purposed but closed due to API breaking change would be to handle an optional parameter into the `previousPage`, `nextPage` and `setPage` methods so a component could deal with multiple pagination, storing the page into different properties._

Ping @ahmedabdel3al as I felt he was in demand of this feature

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -14,7 +14,7 @@ trait WithPagination
             ? $this->queryString()
             : $this->queryString;
 
-        return array_merge(['page' => ['except' => 1]], $queryString);
+        return array_merge([$this->getPageName() => ['except' => 1]], $this->queryString);
     }
 
     public function initializeWithPagination()
@@ -22,7 +22,7 @@ trait WithPagination
         $this->page = $this->resolvePage();
 
         Paginator::currentPageResolver(function () {
-            return (int) $this->page;
+            return $this->{$this->getPageName()};
         });
 
         Paginator::defaultView($this->paginationView());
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for this PR @RomainMazB. This is a pretty non-invasive solution to the problem, so I like that.

Two thoughts:
- Is there a reason you removed the `(int)` cast?
- I think there might be a better API for setting the page name - I'm wondering if Laravel has a pagination convention for what this key might be.

I know @austenc has done some thinking on multi-pagination in Livewire. Any thoughts @austenc?

austenc:
@RomainMazB This looks good! It'd be even better if we didn't have to specify both `getPageName()` **and** the new public variable for the custom page name, but I think it's an acceptable tradeoff if we document this feature well.

This PR allows multiple paginators to coexist on a page while still keeping backwards compatibility, so I'm in favor of it. 
I also like that you used `getPageName()` which matches the Laravel `$paginator->getPageName()` method. Even though Livewire's paginator isn't the same, it's nice to have that consistency.

We should probably add the `(int)` casts back, though?

Assuming this PR is merged, we can discuss solutions and best practices for multiple paginators in your other PR (#2665).

calebporzio:
Thanks for PRing this. However, going to close this.

Here is the current decision and way forward for now: https://github.com/livewire/livewire/discussions/2727

PR #2665, opened 2021-03-24
Implement multi-pagination in a single component

RomainMazB:
Yes, it's a wanted feature, multiple PR has previously opened and closed due to API breaking change.

No, but it's an extension of a previous mandatory PR #2663 . I've spitted them as even if this one is not merged, the previous one is still a nice feature to have.

Yes, unit and browser

This PR gives us the ability to manage multiple pagination in a single component, using different properties to store the different page state and an array to store the name of each paginator:
```php
class ComponentWithMultiplePaginationStub extends Component
{
    use WithPagination;

    public $postsPage = 1;

    protected $paginatorNames = ['page', 'postsPage'];

    public function render()
    {
        return view('show-name'[
            'posts' => Post::paginate(3, ['*'], 'postsPage'),
            'messages' => Message::paginate(3), // Will use the default $page attribute
        ]);
    }
}
```

The initial API is not broken, we now can just pass an additional parameter to select the paginator we want to increment/decrement or change:
```html
// Old API still work, it still modify the $page attribute
<button wire:click="previousPage"></button>
<button wire:click="nextPage"></button>
<button wire:click="goToPage(3)"></button>

// New API for multiple paginator component:
<button wire:click="previousPage('postsPage')"></button>
<button wire:click="nextPage('postsPage')"></button>
<button wire:click="goToPage(3, 'postsPage')"></button>
```

Ping @mokhosh as I felt he was in demand of this feature.

Hopefully it will help some of us :)

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -14,15 +14,19 @@ trait WithPagination
             ? $this->queryString()
             : $this->queryString;
 
-        return array_merge([$this->getPageName() => ['except' => 1]], $this->queryString);
+        $queryParameters = array_fill_keys($this->getPaginatorNames(), ['except' => 1]);
+
+        return array_merge($queryParameters, $this->queryString);
     }
 
     public function initializeWithPagination()
     {
-        $this->page = $this->resolvePage();
+        foreach($this->getPaginatorNames() as $pageName) {
+            $this->{$this->getPageName($pageName)} = $this->resolvePage($pageName);
+        }
 
-        Paginator::currentPageResolver(function () {
-            return $this->{$this->getPageName()};
+        Paginator::currentPageResolver(function ($pageName = 'page') {
+            return $this->{$this->getPageName($pageName)};
         });
 
         Paginator::defaultView($this->paginationView());
@@ -39,40 +43,45 @@ trait WithPagination
         return 'livewire::simple-' . (property_exists($this, 'paginationTheme') ? $this->paginationTheme : 'tailwind');
     }
 
-    public function previousPage()
+    public function previousPage($pageName = 'page')
     {
-        $this->setPage(max($this->{$this->getPageName()} - 1, 1));
+        $this->setPage(max($this->{$this->getPageName($pageName)} - 1, 1), $pageName);
     }
 
-    public function nextPage()
+    public function nextPage($pageName = 'page')
     {
-        $this->setPage($this->{$this->getPageName()} + 1);
+        $this->setPage($this->{$this->getPageName($pageName)} + 1, $pageName);
     }
 
-    public function gotoPage($page)
+    public function gotoPage($page, $pageName = 'page')
     {
-        $this->setPage($page);
+        $this->setPage($page, $pageName);
     }
 
-    public function resetPage()
+    public function resetPage($pageName = 'page')
     {
-        $this->setPage(1);
+        $this->setPage(1, $pageName);
     }
 
-    public function setPage($page)
+    public function setPage($page, $pageName = 'page')
     {
-        $this->{$this->getPageName()} = $page;
+        $this->{$this->getPageName($pageName)} = $page;
     }
 
-    public function resolvePage()
+    public function resolvePage($pageName = 'page')
     {
         // The "page" query string item should only be available
         // from within the original component mount run.
-        return request()->query($this->getPageName(), $this->page);
+        return request()->query($this->getPageName($pageName), $this->{$this->getPageName($pageName)});
     }
-    
-    public function getPageName()
+
+    public function getPageName($pageName = 'page')
     {
-        return 'page';
+        return $pageName;
+    }
+
+    public function getPaginatorNames()
+    {
+        return ['page'];
     }
 }
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -14,7 +14,7 @@ trait WithPagination
             ? $this->queryString()
             : $this->queryString;
 
-        return array_merge(['page' => ['except' => 1]], $queryString);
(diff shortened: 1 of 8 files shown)


mokhosh:
Thanks. Yes I've resorted to multiple child components and a custom pagination trait for now.

I wish you didn't have to declare the variable and have it in an array too. It would be nice to just have the array.

That's what I tried to achieve haphazardly in this PR https://github.com/livewire/livewire/pull/1997

RomainMazB:
Hi, yes I've done this way instead of storing the pagination into an associative array, like you've done in your PR because :
- this would have make the `$page` property useless
- even if we could redefine a "page" index into the array, it would have been more verbose to increment it:
```
goToPage(3, 'page')
nextPage('page')
```

Now that the tests are passing, if Caleb feels good about this PR, we could talk about some syntax sugar to improve the DX.

Using magic method we could even imagine, without an array to have this kind of syntax:
```
// Old API still work, modify the $page property
<button wire:click="previousPage"></button>
<button wire:click="nextPage"></button>
<button wire:click="goToPage(3)"></button>

// New API modifying custom page property
<button wire:click="previousPostsPage"></button>
<button wire:click="nextPostsPage"></button>
<button wire:click="goToPostsPage(3)"></button>
```

These methods would proxy the modification to the wanted properties, without the need to define them into an array:
```php
class ComponentWithMultiplePaginationStub extends Component
{
    use WithPagination;

    public $postsPage = 1;

    public function render()
    {
        return view('show-name'[
            'posts' => Post::paginate(3, ['*'], 'postsPage'),
            'messages' => Message::paginate(3), // Will use the default $page attribute
        ]);
    }
```
_The only down side to this it the fact that it create a breaking change for those who've created custom pagination system or just `goTo[...]Page` methods like `goToLeftPage`_

Maybe this sugar syntax could be added into 3.0

mokhosh:
> even if we could redefine a "page" index into the array, it would have been more verbose to increment it

Couldn't that be done with default values for the `page` parameter?

RomainMazB:
You mean the default value to `['page']` for the `$paginatorNames` property?

Yes it could be done that way but this would create a breaking change IMO for those who rely on the `$page` property in custom code as the `$page` property will never get incremented anymore:

```php
class ComponentWithPagination extends Component
{
    use WithPagination;

    // protected $paginatorNames = ['page' => 1]; // Default from WithPagination trait, just here for clarity

    public function render()
    {
        return view('show-name', [ 'posts' => Post::paginate(3) ]);
    }
}
```

```html
<span>
You are now on the page {{$page}}
</span>
```

mokhosh:
We could have the default parameter values, plus a default `$paginatorNames = ['page' => 1]`.

RomainMazB:
In such case, how to know which property should be changed?

Another way to do would have to explicitly define the component to handle multiple pagination, or to not define the `paginatorNames` in the trait and only define it in the instantiated components which need multiple pagination.

Doing so we could proxy the method to the `$page` property if the `$paginatorNames` is not defined and to the `$paginatorNames['page']` if it is.

Doing so, the developers explicitly know they can't use the `{{ $page }}` variable into view.

mokhosh:
I'm not sure if I understood the question, but how about we change whatever they specify, and change `page` as fallback if they don't specify a page name?

RomainMazB:
The question was about the previous comment:
> We could have the default parameter values, plus a default $paginatorNames = ['page' => 1]

How do we know which "page storage" to increment if the `WithPagination` trait is defined like so:
```php
trait WithPagination {
    public $page = 1;

    public $pageNames = ['page' => 1];
}
```

> but how about we change whatever they specify, and change page as fallback if they don't specify a page name?

I think this is what I just exposed before, we could have something like this:

```php
trait WithPagination {
    public $page = 1;

    public function setPage($page, $pageName = 'page') {
        if ($pageName !== 'page' && property_exists($this, 'paginatorNames')) {
              $this->paginatorNames[$pageName] = $page;
        } else {
           $this->page = $page;
        }
    }
}
```

Doing so, the developer can still use the `{{ $page }}` into his view and can define `paginatorNames` to handle additional pagination.

mokhosh:
No you have this:
```php
trait WithPagination {
    public $pageNames = ['page' => 1];
}
```

and only change the array, and set public properties based on the array, and default to the name `page` in methods:
```php
public function goToPage($page, $pageName = 'page')
```

RomainMazB:
> We just need to find the right hook to initialize all the page number public properties based on the array before anyone tries to access them.
The array is only the blueprint for dynamically setting the public properties.

Ok, I think I got you.

This should work but I dislike to work with dynamic properties (and so does my IDE 😄), but I'm totally open to the conversation:

As you correctly said, we should search for the perfect hook to make sure the property is not accessed before its declaration, and actually we can't make sure of that: our best chance is probably the `initializeWithPagination`, but imagine someone use - for any reason - the `$page` variable into the component's constructor, it would suddenly fail because the `initializeWithPagination` method wouldn't have been called.

AFAIK, we cannot make sure someone didn't use the variable before any hook we could register from the trait.

mokhosh:
I understand this is a trade off.

Elegance + accounting for 99% of the usage. Or ugly duplicate code + 100% of the theoretically possible cases. (excuse the emphatic language)
About the IDE thing, we're already working with Laravel ecosystem so having 100% IDE integration is irrelevant.

RomainMazB:
No problem, we are here to debate and that's exactly why I pinged you :)

For now I'm working on something else, so I can't dive into this for few days. But I'll modify the PR if needed.

Maybe some feedback here @calebporzio to know what would be acceptable?

calebporzio:
Thanks for tackling this. Going to close this for now. More info here: https://github.com/livewire/livewire/discussions/2727

Issue #1765, opened 2020-10-06
Prefetch does not work when button with wire:click.prefetch has a dom element surrounding the button text (Copy paste simple example)


isaackearl:
### Description

This bug took me awhile to nail down... truly a strange bug.

If you have a button, and in that button the "slot" or "text" has a paragraph tag around it.. then prefetching doesn't work.  When I say it doesn't work, what I mean is that it STILL makes a request when you click instead of using the prefetched content.  The prefetching of the content itself seems to work fine, it just doesn't respect that it's already been requested.  It is hard to explain but very easy to show.  Look at the code below to see what I mean!

### Exact steps to reproduce
This example is basically copied straight out of the docs on prefetching so I'm not doing anything out of the ordinary here.  What you'll see is a hidden content with 2 buttons, one that works and one that doesn't.  The only different between the one that doesn't work and the example straight out of the documentation, is that there is a <p> tag in the button.

Copy this code:

```
class TesterLive extends Component
{

    public $contentIsVisible = false;

    public function toggleContent() {
        $this->contentIsVisible = ! $this->contentIsVisible;
    }

    public function render()
    {
        return view('livewire.tester-live');
    }
}
```
and the view:

```
<div>
    <button wire:click.prefetch="toggleContent"
            class="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded">
        Show Content
    </button>
    <button wire:click.prefetch="toggleContent"
            class="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded">
        <p>Show Content but prefetch doesn't work</p>{{--paragraph tag makes this not work--}}
    </button>
    @if ($contentIsVisible)
        <div class="text-lg">
            Some Content...
        </div>
    @endif
</div>
```

### Other Notes: 
This bug makes pre-fetching nearly impossible with anything that isn't a straightforward button.  If you have blade components that have if statements and other features (such as breadcrumbs etc), then any amount of nesting within the component causes prefetching to stop being respected (it pre-fetches, but then still makes a request when you click).. removing the nesting within the button or link and wa-la! it starts to work again... I've tried writing wire:key on the outmost element of both the content being fetched as well as the button... but it doesn't seem to help.

**Context**
- Livewire version: v2.2.9
- Laravel version: v8.8.0
- Browser: Chrome (probably any)

squishythejellyfish:
[...]

PR #2667, opened 2021-03-25
WIP - test(dusk): allow custom path for web drivers

kslimani:
I wanted to run browser tests locally and i had same error as [this issue comment](https://github.com/livewire/livewire/issues/1765#issuecomment-725331231). (chrome version error)

Nope, it only refactor web driver process creation code.

It is part of browser testing setup.
By default it will attempt to run the chromedriver packaged with current dev. requirements. I guess (but i may be wrong) it is working in github action because the running image has the previous versions of chrome available.

On your workstation, you may (like myself) have the lastest chrome installed with [latest chrome drivers](https://chromedriver.chromium.org/downloads); When i attempt to run locally a browser test; i get an exception like :

```
Facebook\WebDriver\Exception\SessionNotCreatedException: session not created: Chrome version must be between 70 and 73
  (Driver info: chromedriver=2.45.615279 (12b89733300bd268cff3b78fc76cb8f3a7cc44e5),platform=Linux 4.19.0-14-amd64 x86_64)
```

It adds a simple way to specify your web driver binary path. Currently it is commented in the `TestCase.php` file, in the `prepare()` static method. (i moved `prepare()` from SupportsSafari trait to TestCase class, i think it is a proper location)

You can specify custom Safari or Chrome driver respectivelly using `static::useSafaridriver('/path/to/bin/safaridriver')` and `static::useChromedriver('/path/to/bin/chromedriver')`.

---

This pull request is a work on progress, i need some suggestion on best/proper way to :

* Specify if browser tests are run using Safari or Chrome driver __without__ editing `TestCase.php`
* Specify Safari or Chrome driver custom path __without__ editing `TestCase.php`

Maybe using ENV variable ? add commented available ENV variables in a phpunit.xml.dist file ? (and add phpunit.xml in git ignore file ?), so you can copy phpunit.xml.dist to phpunit.xml and edit it.

Or should we keep it commented as is for now in the `TestCase.php` file ?

### DIFF
--- a/tests/Browser/SupportsSafari.php
+++ b/tests/Browser/SupportsSafari.php
@@ -2,31 +2,25 @@
 
 namespace Tests\Browser;
 
+use Symfony\Component\Process\Process;
+
 // Thanks to https://github.com/appstract/laravel-dusk-safari for most of this source.
 trait SupportsSafari
 {
-    protected static $safariProcess;
+    protected static $safariDriver = '/usr/bin/safaridriver';
 
-    /** @beforeClass */
-    public static function prepare()
-    {
-        if (static::$useSafari) {
-            static::startSafariDriver();
-        } else {
-            static::startChromeDriver(['port' => 9515]);
-        }
-    }
+    protected static $safariProcess;
 
     public function onlyRunOnChrome()
     {
         static::$useSafari && $this->markTestSkipped();
     }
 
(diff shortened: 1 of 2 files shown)


kslimani:
I just figured out i could also solve the issue by running :

```console
./vendor/bin/dusk-updater update
```

PR #2680, opened 2021-03-30
Avoid exception for unsupported HTTP methods

Namoshek:
Working with the browser history for dynamically loaded components is currently broken. Assume we have a route `GET foo/bar` rendering a view, which in return uses `POST biz/baz` (via ajax) to load Livewire components dynamically. Then actions of the component will send `foo/bar` as `Referer` in the header but `POST` and `biz/baz` as `fingerprint` in the body.

This seems totally right, as the referer is what the user sees in their browser and `POST biz/baz` is the route which generated the component in the first place. The Symfony router will throw a `MethodNotAllowedHttpException` though if the route exists and the method is not allowed. In my opinion, a not supported method is a sub-case of `NotFoundHttpException` - but the exception class inheritance says otherwise. Therefore the PR adds `MethodNotAllowedHttpException` to the `catch` clause.

I'm not sure if this fix could potentially cause issues in other areas.

## Background

I have a simple `GET /list` route which renders a [DataTable](https://datatables.net/). The DataTable uses `POST` to retrieve data, but from another route than `/list` (because there are multiple such tables on the same page). Some of the tables contain Livewire components in the response (which works totally fine by the way, all that's needed is a `window.livewire.rescan()` after drawing the table). If a component contains a button to trigger a server-side action, clicking the button will cause the issue described above.

### DIFF
--- a/src/RenameMe/SupportBrowserHistory.php
+++ b/src/RenameMe/SupportBrowserHistory.php
@@ -9,6 +9,7 @@ use Illuminate\Support\Arr;
 use Illuminate\Http\Request;
 use Illuminate\Routing\UrlGenerator;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException;
 use function Livewire\str;
 
 class SupportBrowserHistory
@@ -73,7 +74,7 @@ class SupportBrowserHistory
             return app('router')->getRoutes()->match(
                 Request::create($referer, Livewire::originalMethod())
             );
-        } catch (NotFoundHttpException $e) {
+        } catch (NotFoundHttpException|MethodNotAllowedHttpException $e) {
             // If not, use the current route.
             return app('router')->current();
         }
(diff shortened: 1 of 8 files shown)


calebporzio:
Thank you for this. Can you submit a test along with this @Namoshek?

Namoshek:
@calebporzio Would love to, but I'm not sure how to be honest. I think a test would require a custom route definition. But I'll look into it.

calebporzio:
No problem @Namoshek.

The SupportsBrowserHistory functionality IS tested. You can look in this directory for reference:
`tests/Browser/SyncHistory`

Also, `tests/Browser/TestCase.php` on line 77 is a good example of another test that added a route definition in Dusk.

With those two examples you should have lots to go off of.

Hopefully that helps! Thanks!

Namoshek:
@calebporzio I added a test which demonstrates that loading a component dynamically via `POST` (from a view loaded via `GET`) is not a problem anymore after the change. I locally verified that removing the `MethodNotAllowedHttpException` from the catch block in `SupportBrowserHistory` concern triggers the error as expected. In other words, seems like the fix works as expected.

Adding this test was actually a bit tricky because the whole test setup did not work for normal views, at least I couldn't make it work without adding a separate layout file for this case. Also the missing testbench directory `Http/Livewire` was a bit tricky to debug.

calebporzio:
Thanks for putting in the work to get the tests up and running!

PR #2699, opened 2021-04-06
Add docblock tags to provide auto-completion in IDE's

jameslkingsley:
When writing tests it's always slightly bothered me that it doesn't autocomplete the assertion methods, so I figured I'd make this PR that adds the necessary Docblock tags to fix that.


calebporzio:
Thanks!

PR #2700, opened 2021-04-07
Add helper method for making custom component assertions

jameslkingsley:
This is just a convenience method for this:
```php
$component = Livewire::test(Foo::class);

$response = $component->viewData('_instance')->someMethod();

$this->assertEquals('bar', $response);
```
So now you can keep the conga line going 😎 
```php
Livewire::test(Foo::class)
    ->assertComponent(function ($component) {
        $this->assertEquals('bar', $component->someMethod());
    })
    ->call('keepOnGoing');
```

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Testing\Concerns;
 
+use Closure;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the PR, however I'd say to prefer this currently available syntax:

```php
Livewire::test(Foo::class)
    ->tap(function ($testSubject) {
        $this->assertEquals('bar', $testSubject->instance()->someMethod());
    })
    ->call('keepOnGoing');
```

(not great variable names, but you get the idea)

PR #2704, opened 2021-04-07
make the test stub file publishable

Mecit:


calebporzio:
Great, thanks!

Mecit:
@calebporzio I'm sorry, I've just realized that I actually forgot to change the first parameter for the `file_put_contents` function. That'd actually overwrite the livewire.view.stub file. Would you please care to fix it? I think that'd be less troublesome for you than me reverting this commit and sending a new one.

calebporzio:
@Mecit, I'll revert this PR, but please submit a new one with the proper fix. Thanks!

PR #2708, opened 2021-04-08
Fixed wrong manifest path in Vapor

lkmadushan:

See the attachment.
Screenshot 2021-04-09 at 2 13 16 AM

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -104,7 +104,7 @@ return [
     | "<app_root>/bootstrap/cache/livewire-components.php)", but for specific
     | cases like when hosting on Laravel Vapor, it could be set to a different value.
     |
-    | Example: for Laravel Vapor, it would be "/tmp/storage/bootstrap/cache/livewire-components.php".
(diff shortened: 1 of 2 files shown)


calebporzio:
Hi @lkmadushan - I believe you that this is the right adjustment. But can you provide some further proof and context as to when/why this changed? Thanks!

jelleroorda:
Maybe it would be an idea to handle it the same way as the vapor core package is doing in it's [AppHandler.php#L30](https://github.com/laravel/vapor-core/blob/2.0/src/Runtime/Handlers/AppHandler.php#L30)?

calebporzio:
@jelleroorda, are you suggesting we make this dynamic and use `$_ENV['LAMBDA_TASK_ROOT']` to prefix `/bootstrap...` instead of hardcoding `/tmp...`?

That makes sense.

I wonder how to make this change without it being a breaking change and still supporting the config.

Would love someone to propose a solution. Thanks!

calebporzio:

PR #2709, opened 2021-04-09
Add support for history sync with optional route parameters

Persaeus:
History sync for full-page Livewire components with optional route-parameters appears to be broken  (#2710)

This PR fixes that.

Tests have been included to demonstrate the problem.


calebporzio:
Thanks for the PR and good tests!

PR #2711, opened 2021-04-10
Add nullable as a default validation rule when property is not in the rules array

narcisonunez:
**Issue**: You need to define all properties in the rules array even when that property is nullable when using `validateOnly` method.

In the following example we are missing the `company` property in the `$rules` array. If you have validation errors for the `email` field and then fill the `company` field in the form, all the errors will disappear.
```php
public $email;
public $company;

protected $rules = [
    'email' => 'required|email',
];
```

This will empty the error bag if the `$propertyName` is `company`.
```php
public function updated($propertyName)
{
    $this->validateOnly($propertyName);
}
```

HTML
```html
    <input type="text" wire:model="email">
    @error('email') <span class="error">{{ $message }}</span> @enderror

    <input type="text" wire:model="company">
    @error('company') <span class="error">{{ $message }}</span> @enderror

    <button type="submit">Save Contact</button>
</form>
```

This is an issue specially when you are using `debounce` in your `wire:model` and your user use `tab` to the next field and type something.

This can be fixed just adding the property to the `$rules` array as `nullable` but that's annoying :D

```php
protected $rules = [
    'email' => 'required|email',
   'company' => 'nullable'
];
```

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -138,7 +138,7 @@ trait ValidatesInput
 
     public function missingRuleFor($dotNotatedProperty)
     {
-        return ! $this->hasRuleFor($dotNotatedProperty);
+        return !$this->hasRuleFor($dotNotatedProperty);
     }
 
     public function validate($rules = null, $messages = [], $attributes = [])
@@ -168,6 +168,10 @@ trait ValidatesInput
         $rulesForField = collect($rules)->filter(function ($rule, $fullFieldKey) use ($field) {
             return str($field)->is($fullFieldKey);
         })->toArray();
+
+        if (empty($rulesForField)) {
+            $rulesForField[$field] = "nullable";
+        }
 
         $ruleKeysForField = array_keys($rulesForField);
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Hmmm this is a tricky one. I understand your point, but I wonder if making non-existing fields "nullable" by default has any other side effects or implications.

I DO think the behavior you describe is expected though. Help me understand why it's a hinderance?

Thanks!

narcisonunez:
Yes, I agree with you that is the expected behavior for Laravel. 

**Laravel Documentation.**
> By default, Laravel includes the TrimStrings and ConvertEmptyStringsToNull middleware in your application's global middleware stack. These middleware are listed in the stack by the App\Http\Kernel class. Because of this, you will often need to mark your "optional" request fields as nullable if you do not want the validator to consider null values as invalid.
(https://laravel.com/docs/8.x/validation#a-note-on-optional-fields)


Based on that is ok but Laravel does not allows real time validation. That's the reason I just added that behavior to `validateOnly` that is mostly used for Real time validation.

narcisonunez:
I will be closing this PR for now and leave the current functionality as it is. I'm not sure if this can cause any side effects.

You have a really great project man, keep the hard work.

PR #2714, opened 2021-04-12
Added Configuration: force_trailing_slash

schwindy:
>1️⃣ Is this something that is wanted/needed? Did you create an issue / discussion about it first?

Yes, this is necessary for me to use `livewire` in my project at all, which enforces all routes end with a trailing slash (`/`). I did not create an issue or any discussion about this prior to opening this PR.

>2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No, the changes in this PR are scoped to allowing Livewire to work in environments that require a trailing slash be used.

>3️⃣ Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have)

Nope 🙂 I am however using my `schwindy/livewire` fork in my project via `VCS` repository include in `composer.json` for `livewire/livewire@dev-master`, and the issue is solved 🎉 The issue returns if I set my project's `force_trailing_slash` value in  `config/livewire.php` to `false`, which I believe proves it works the same as it did before when `force_trailing_slash` is not enabled.

>4️⃣ Please include a thorough description of the improvement and reasons why it's useful.

I added a new configuration value to `config/livewire.php` named `force_trailing_slash`.

This configuration value will change the `livewire/message/example` AJAX request endpoint to `livewire/message/example/`.

Without this value being set to `true` in my project, all AJAX requests fail. By default, this value is set to `null` in `config/livewire.php` which will set `window.livewire_force_trailing_slash` to `false`. To enable it, we simply change that value to `true`, which in my case fixes AJAX `livewire/message` functionality.

Please let me know if there is anything else I can provide to get this merged. The sooner I can stop using my fork in my project, the better 🙏 I love the package though, great work!

>5️⃣ Thanks for contributing! 🙌


calebporzio:
That's an interesting constraints. Why is it required in your project?

I'd sooner explore a workaround for your project than introduce a new config item.

schwindy:
@calebporzio Why would your solution involve me changing how my project works to workaround your package? Rather than you adding a configuration to your package that would provide what I need.

This is the workaround for my project. Your library is incapable of handling my environment with my constraints. I don't even prefer to always force a trailing slash, but that's my job and that's what I have to do in this case. They are business requirements, so I don't know that I can give you a logic reason other than that is the requirement, end of story.

If you refuse to accept this addition, then I'll have no choice but to either continue to use my fork or just abandon usage of livewire entirely. The AJAX requests do NOT work after being redirected, the requests just 404. Happy to send you a link (privately, in a DM) if you want to see the behavior.

After spending roughly 4 hours reading through the many issues in this repo and other resources on Google trying to solve my problem yesterday, I came to the conclusion that cannot possibly be the only person who needs this. There are so many comments saying that people are getting 404 responses for `message`, and nearly every response is talking about livewire.js being a 404, which had absolutely nothing to do with the core issue.

calebporzio:
Hmmm some strong language and opinions there.

Let me start by saying this is an open source project, we're all pitching in on.

Surely you can understand me wanting an addition to the core being justified? If I merged everyone's PRs without knowing the underlying problem and being sure that others experience it as a pain point, this package would be scattered, bloated, and much more difficult to maintain.

I didn't say people don't experience this problem. I didn't say you don't have good reasons.

I just want to see those examples, and hear the reasons. 

We're all pitching in and trying our best to make this project a good experience for people.

Would you mind posting some references to the problems people are experiencing and helping me understand how this change can help alleviate those?

schwindy:
I am full of strong opinions 🙂 I mean well, hopefully didn't come off too strong...truly. Sorry if I did. 😇 

I think the change can be justified by the `livewire/message` AJAX requests no longer working in if they have been 301 redirected to a new route, such as middleware adding a trailing slash to the request route. This is not an uncommon application behavior, and could be tested by adding a simple middleware that does this and testing any `livewire/message` functionality which should result in a `301` -> `404`.

If you want a link to see the behavior in a dev environment, let me know how I can send a private link to you, as it's a client so I'd prefer it be private. All `livewire/message` requests receive a `301` redirect which results in a `404`, so a "lightbox/iframe" popup appears showing my website's 404 page on any Action. The only difference the redirect caused was adding a `/` to the end of the URL, but that is enough to make it fail it appears 😞 

I believe this optional configuration will:
1. Make the package compatible with more applications/environments
2. Not change any existing behavior unless this configuration is explicitly enabled

Unfortunately, I spent too long fixing this for myself yesterday to go back and find all of the issues I read... on top of submitting this PR. I mean no disrespect but I truly have no time left 😞 

Of course, feel free to discard this solution if you can't reason it for yourself @calebporzio. But I hope it's more clear why this could be **necessary** in some projects, like mine, given the requests fail in this case. It would be VERY useful to me so that I could continue to use livewire without needing to maintain a fork for my project 😞

austenc:
@schwindy how are you hosting this project? 

I'm wondering if it's something strange in your web server configuration that is causing this sort of url rewriting you are experiencing?

schwindy:
We use Apache to force trailing slashes via .htaccess. I suppose maybe it is a little strange, but I would argue it's a relatively common thing to do. I think the same would happen using a middleware too correct?

I just added an exception to my .htaccess rewrite to configure the application for Livewire, which is what we do for Nova, and all works fine now. 🙂 

I think it would be nice to not need to edit my .htaccess and instead a configuration option was used to do the same thing. They are both configuration, but apache is a lot more complicated. I also didn't come across documentation stating the these requests cannot be redirected to force a trailing slash, ultimately sending me down a large rabbit hole 🤦 I feel silly now, and certainly would have preferred to avoid this.

I'll just use this apache rewrite exception though. Thanks.

austenc:
Although I'm not as familiar with Apache as I used to be, I think this might be a common thing for most Laravel apps?

Looks like Laravel's `.htaccess` is removing trailing slashes, see here:
https://github.com/laravel/laravel/blob/8.x/public/.htaccess#L12

calebporzio:
@austenc , because Laravel has that in .htaccess, you're staying they anyone that is running livewire on Apache would experience these errors?

So the reason most people don't is that forge and valet use nginx I'm guessing?

austenc:
@calebporzio No, sorry, I should've been more clear. Since Laravel does that and @schwindy mentioned this:
> We use Apache to force trailing slashes via .htaccess

This seems to be an uncommon server configuration, and since @schwindy had to make an exception for Nova too, I think this may just be a difference for his setup?

calebporzio:
Gotcha, yep thanks. Ok I'm going to close this until there's more demand for it.

Thanks @austenc and @schwindy !

schwindy:
While I am fine with my custom .htaccess hack to make this package work, nothing about adding trailing slashes is "weird", "uncommon", or "strange". I see the Nova exclusion as a hack too, to be fair. But just because one package requires a hack doesn't mean others should follow suit.

It may be that none of you have done it before, but this is a rather common practice. There are a lot of projects that do this in the world, so just imagine how many would experience the same issue I did. I would think that a lot of users would have given up fixing it long before finding the cause to be trailing slashes or opening an issue, given there is no documentation regarding it and it took a bit to debug. But I digress.

Any kind of trailing slash redirection on `/livewire/message/*` will break this package. It must be excluded from the trailing slash redirection in whatever custom way works in your project.

Maybe this will help someone in the future. Thanks all, I appreciate the information.

joshhanley:
@schwindy if you don't mind me asking, you said that is common practice, but can you explain why it is done and what it is useful for? I'm unclear why you'd need to force a trailing slash at the web server level (or at all for that matter) as I've never seen it done. Genuinely curious, so would love to hear your thoughts 🙂

schwindy:
@joshhanley I believe the reason we do it for this client is to ensure that any given link can only be accessed from a single url `path`, so it's an SEO thing ("canonical"). It prevents Google from marking identical pages as `duplicates` where the only difference in route path is a trailing slash.

In the case of Laravel Routing, by default if you create a route as so:
```
Route::get('/blog/', 'BlogController@index')->name('blog.index');
```

Then naturally, that route can be accessed from `both` of the following as `200` responses:
1. `/blog`
2. `/blog/`

Enforcing trailing slashes ensures that even though a page "could" be loaded as `/example` and `/example/`, that it will only ever be loaded as `/example/` (trailing slash). Any requests to `/example` will be `301` redirected to its trailing slash counterpart. Which prevents Google from marking `/example` and `/example/` as duplicate pages. That's as far as I understand, I am certainly no SEO expert but that is our use-case at least.

Regardless of how a UX may link to a given route, with or without trailing slash, it will always resolve to the same path before being a `200`, with a trailing slash, ensuring that a given page can only be accessed from a singular path.

This is obviously **a lot less important for API routes**, like this `livewire/message` request, but that is why I believe this should be a package configuration option. To allow the package to work using only its own config and not require users implement custom solutions/exceptions for cases like this.

See: https://developers.google.com/search/blog/2010/04/to-slash-or-not-to-slash (first result, it's old too 🤷)
>don't both return a 200 response code, but that one version redirects to the other.

> If only one version can be returned (i.e., the other redirects to it), that's great! This behavior is beneficial because it reduces duplicate content. In the particular case of redirects to trailing slash URLs, our search results will likely show the version of the URL with the 200 response code (most often the trailing slash URL) -- regardless of whether the redirect was a 301 or 302.

Which presents a problem if trying to use Livewire without adding a custom redirect exception specifically for `*/livewire/*` which I did via this addition to my `.htaccess` 🙂 
```
RewriteCond %{REQUEST_URI} !^/(livewire)/ [NC]
```
And is also why I cannot buy that I am the only person who has ever struggled with this. It is **not** an uncommon practice, as far as I know at least, for any application concerned with their SEO. 🐒

I feel this would be a great addition to the Troubleshooting docs (https://laravel-livewire.com/docs/2.x/troubleshooting), at least. It could have saved me hours of head scratching, mainly 😇

PR #2723, opened 2021-04-14
fix the publish command for test files

Mecit:
Fixes the mistake made in #2704, added tests this time.


calebporzio:
Perfect, thanks for resubmitting!

PR #2724, opened 2021-04-14
Ignore blade files for Github linguist

reziamini:
I guess it's better not to show blade language in GitHub as the first language.
With this one line GitHub will ignore *.blade.php files to calculate the first lang!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,6 @@
 /tests              export-ignore
 /.editorconfig      export-ignore
 /github_banner.png  export-ignore
+
+# Ignore blade files
+*.blade.php linguist-vendored


joshhanley:
@rezaamini-ir not sure why this matters, but I would think you'd actually be better off ignoring the tests directory completely rather than blade files (as some are good), as that would the give a clearer picture of what this repo consists of.

It I think it'd actually be interesting to see how much php vs javascript there is if tests are not included in the calculation.

rezaamini-ir:
@joshhanley
Anyway, I feel better when seeing the first lang isn't Blade and It's PHP :)

calebporzio:
@joshhanley - I too would be interested in knowing that sans-tests language split.

I'm going to close this as it doesn't really add value to the project. Thanks for contributing though @rezaamini-ir!

PR #2725, opened 2021-04-14
Fixes to ensure Livewire can run in a sessionless Laravel application

ralphschindler:

I believe so, primarily for deploying livewire on websites (not webapps per-se). Brief discussion in Discord.

No, all changes are specific to running livewire in a session-less application.

No tests, minimal backwards compatible changes, all existing tests pass.

Since Livewire bundles its component state and sends it via POST request to the /livewire endpoint, it is especially well suited to being used in a website environment (dynamically generated pages that are heavily CDN cached). Websites of this nature generally don't have a need for sessions as they are not user centric.

The initial page load may come from a CDN (a static page), as such, there might not be a valid CSRF meta tag.  Additionally, to further enforce sessionless development, the session stuff may be removed:

from config/app.php, for example:

```php
    'providers' => [
        // ... commenting out the session service provider
        // Illuminate\Session\SessionServiceProvider::class,
```

and removing the session/stateful specific middleware:

```php
    protected $middlewareGroups = [
        'web' => [
            // \App\Http\Middleware\EncryptCookies::class,
            // \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
            // \Illuminate\Session\Middleware\StartSession::class,
            // \Illuminate\Session\Middleware\AuthenticateSession::class,
            // \Illuminate\View\Middleware\ShareErrorsFromSession::class,
            // \App\Http\Middleware\VerifyCsrfToken::class,
            \Illuminate\Routing\Middleware\SubstituteBindings::class,
        ],
```

The suggested changes effectively check to see if sessions are enabled before opting into session specific features (like session clearing and redirector url persistence.  Additionally, the front end will no longer throw an Error if the csrf meta tag is no longer present, if it is not present, it will simply not include a `X-CSRF-TOKEN` header in its http calls.

I don't believe these changes don't impact anyone using sessions, it merely makes it more clear to consumers of Livewire that wish to run Livewire in a sessionless application.

No, thank you! 😃


calebporzio:
This makes sense to me. Why enforce CSRF if we don't need to. Thanks!

PR #2730, opened 2021-04-16
poll if any part is visible, not if every part is visible

mokhosh:
The https://github.com/livewire/livewire/pull/2560 PR added ability to poll only when the element is visible, but it currently checks if all of the element is inside the boundaries of the viewport. I've changed it so it checks if any part of the element is visible.

For example if you have a section that is slightly longer than viewport's hight, it will never do the poll because the bounding box will never fully fit within the page.

### DIFF
--- a/js/component/Polling.js
+++ b/js/component/Polling.js
@@ -73,9 +73,9 @@ function inViewport(el) {
     var bounding = el.getBoundingClientRect();
 
     return (
-        bounding.top >= 0 &&
-        bounding.left >= 0 &&
-        bounding.bottom <= (window.innerHeight || document.documentElement.clientHeight) &&
-        bounding.right <= (window.innerWidth || document.documentElement.clientWidth)
+        bounding.top < (window.innerHeight || document.documentElement.clientHeight) &&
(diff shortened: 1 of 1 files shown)


calebporzio:
Good call @mokhosh , thanks for the contribution!

ju5t:
Perhaps a bit late to the party, but the current tests don’t cover this change.

mokhosh:
> Perhaps a bit late to the party, but the current tests don’t cover this change.

You're right. I don't have any experience with testing js. But I tested both codes manually.

Can you please help by adding tests for this?

ju5t:
@mokhosh it's pretty easy once you see it I guess, I didn't knew anything about it either when I started writing this test. The current tests are in [PollingViewport](https://github.com/livewire/livewire/tree/master/tests/Browser/PollingViewport).

In short: the component increases a variable by 1 each time it renders. The test will wait until it sees the second round trip:

```php
->assertSeeIn('@output', '2')
```

It will the scroll to the bottom, where it will wait 2 seconds until it scrolls back up. When Livewire runs, it expects to see 3. If not, something went wrong.

Testing your change would be pretty similar I guess.

PR #2736, opened 2021-04-18
Add params method to view macros

reziamini:
In some cases we just need to pass some data like the title to the base layout and now to do that in Livewire we have to use `layout` method then pass the params as the second param like this : 

```php
return view('login')->layout('layouts.app', [
    'title' => 'Login Page'
]);
```

With these changes, we will be able to use a syntax like this:
```php
return view('login')->params([
    'title' => 'Login Page'
]);
```
Also, all tests have been passed and new tests have been written!

### DIFF
--- a/src/Macros/ViewMacros.php
+++ b/src/Macros/ViewMacros.php
@@ -38,6 +38,18 @@ class ViewMacros
         };
     }
 
+    public function params()
+    {
+        return function ($params = []) {
+
+            $this->livewireLayout = [
+                'params' => $params,
+            ];
+
+            return $this;
+        };
+    }
+
     public function section()
     {
         return function ($section) {
--- a/tests/Unit/ComponentLayoutTest.php
+++ b/tests/Unit/ComponentLayoutTest.php
@@ -59,6 +59,28 @@ class ComponentLayoutTest extends TestCase
             ->assertSee('bar')
             ->assertSee('baz');
     }
+
+    /** @test */
+    public function can_show_the_params()
+    {
+        Livewire::component(ComponentWithCustomParams::class);
+
+        Route::get('/foo', ComponentWithCustomParams::class);
+
+        $this->withoutExceptionHandling()->get('/foo')
+            ->assertSee('foo');
+    }
+
+    /** @test */
+    public function can_show_params_with_custom_layout()
+    {
+        Livewire::component(ComponentWithCustomParamsAndLayout::class);
+
+        Route::get('/foo', ComponentWithCustomParamsAndLayout::class);
+
+        $this->withoutExceptionHandling()->get('/foo')
+            ->assertSee('foo');
+    }
 }
 
 class ComponentWithExtendsLayout extends Component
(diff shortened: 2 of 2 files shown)


rezaamini-ir:
@joshhanley Yes you are right about the name and use cases.
I think it's better to wait what does @calebporzio think about name and PR !

zepfietje:
I've also encountered this. What I found is that you can pass `null` as the `view` to the `layout()` method as a workaround.

That being said, I do feel like a dedicated method without having to specify the layout does make sense. In my own project I created a macro `layoutData()`, since it best describes what its purpose is. However, I feel like there could possibly be an even better name.

rezaamini-ir:
@zepfietje I'm agree with `layoutData` name for method

calebporzio:
I could be down with this. Thanks for the input everyone.

I prefer layoutParams though to stay consistent.

Would you be able to PR this to the docs as well?

zepfietje:
@calebporzio wouldn't `layoutData` be more consistent with the Laravel core than `layoutParams` though?

calebporzio:
Interesting, what are the references to Laravel core?

zepfietje:
The Laravel docs mention passing **data** to views: https://laravel.com/docs/8.x/views#passing-data-to-views. I couldn't find any references where it's being referred to as **params**.

rezaamini-ir:
@calebporzio So are agree to change the method name to `layoutData`?

calebporzio:
Yep! I dig!

joshhanley:
@calebporzio I think this hadn't been updated yet, so `params()` got merged instead of `layoutData()`

rezaamini-ir:
@joshhanley Yes you are right.
@calebporzio has merged `params` method

calebporzio:
Ooops, adjusted now. Thanks!

binaryweavers:
Not sure if changes are still being considered. I see that `params` implementation is not right and will override everything in 
`$this->livewireLayout`. Like `slot` and `section` macro, shouldn't it be `$this->livewireLayout['params'] = $params;` ?
[https://github.com/rezaamini-ir/livewire/blob/77e8b4b93d6da95dd219d4ade79dafb18dc49065/src/Macros/ViewMacros.php#L45](https://github.com/rezaamini-ir/livewire/blob/77e8b4b93d6da95dd219d4ade79dafb18dc49065/src/Macros/ViewMacros.php#L45)

Second test is passing because it checks for `main` param set to `foo` but not for `layout` which is being set to default from `config`.
[https://github.com/rezaamini-ir/livewire/blob/77e8b4b93d6da95dd219d4ade79dafb18dc49065/tests/Unit/ComponentLayoutTest.php#L82](https://github.com/rezaamini-ir/livewire/blob/77e8b4b93d6da95dd219d4ade79dafb18dc49065/tests/Unit/ComponentLayoutTest.php#L82)

rezaamini-ir:
> Not sure if changes are still being considered. I see that `params` implementation is not right and will override everything in
> `$this->livewireLayout`. Like `slot` and `section` macro, shouldn't it be `$this->livewireLayout['params'] = $params;` ?
> https://github.com/rezaamini-ir/livewire/blob/77e8b4b93d6da95dd219d4ade79dafb18dc49065/src/Macros/ViewMacros.php#L45
> 
> Second test is passing because it checks for `main` param set to `foo` but not for `layout` which is being set to default from `config`.
> https://github.com/rezaamini-ir/livewire/blob/77e8b4b93d6da95dd219d4ade79dafb18dc49065/tests/Unit/ComponentLayoutTest.php#L82

Thanks! 
It was fixed in a new PR which has been mentioned.

PR #2743, opened 2021-04-20
Update LivewireBladeDirectives.php

wuwx:
remove space from html output

```
x-data="{ data:  window.aLivewire.find('hi1VPZYqmvBpBJySHe0G').entangle('data') .defer }"
                s                                                              s
```


calebporzio:
Thanks!

Issue #2603, opened 2021-03-07
Redirect & Browser Back Form Interaction Bugs


M-Porter:
### Description

There are actually 2 bugs here related to form behaviors after navigating back. Not opening two separate tickets as I believe they are related.

* **Bug 1** – When navigating back in the browser (clicking the back button) after doing an action which causes a redirect, forms are no longer editable. When inspecting the DOM, a `readonly=""` attribute remains on the input element. 

* **Bug 2** –When navigating back in the borwser (clicking the back button) after doing an action which causes a redirect, the form input is editable but clicking submit does nothing. The form goes into a read-only and disabled state but no xhr request gets sent.

Only a full page refresh (as expected) fixes this.

Webapp of bug can be found at https://livewire-redirect-back-ro-bug.herokuapp.com/
Code repository of bug can be found at https://github.com/M-Porter/livewire-redirect-back-readonly-bug

### Exact steps to reproduce

To make it easier i setup a heroku instance to run this bug:

**Bug 1**
1. Go to https://livewire-redirect-back-ro-bug.herokuapp.com/
2. Enter a value in the input field and click submit.
3. After being redirected to the new page, click the back button in your browser.
4. See your original input value in the input field. The input field is not editable.

**Bug 2**
1. Go to https://livewire-redirect-back-ro-bug.herokuapp.com/
2. Enter a value in the input field and click submit.
3. After being redirected to the new page, click the back button in your browser.
4. See your original input which is editable.
5. Click submit, nothing happens (no xhr).

### Stripped-down, copy-pastable code snippets

I'm not really sure which code snippets to include (which is why i made the repo to follow, but here are the files i changed from the base laravel install:
* Routes – https://github.com/M-Porter/livewire-redirect-back-readonly-bug/blob/main/routes/web.php
* Controller – https://github.com/M-Porter/livewire-redirect-back-readonly-bug/blob/main/app/Http/Controllers/Controller.php
* Livewire Component – https://github.com/M-Porter/livewire-redirect-back-readonly-bug/blob/main/app/Http/Livewire/SomeForm.php
* Views – https://github.com/M-Porter/livewire-redirect-back-readonly-bug/tree/main/resources/views

**Context**
* Livewire version: v2.4.0
* Laravel version: v8.31.0
* Alpine version: N/A
* Browser: Chrome, Firefox, Safari

M-Porter:
[...]

Issue #2334, opened 2021-01-11
Unable to bust cache on a component that has updated data


roni-estein:
### Description
I have a series of global components that related to user status, that work fine on page, but "ghost reappear" if you use the back button. When I say ghost re-appear, the functions that would be re-performed like submitting a form, or updating a queried object will fail if you actually attempt to submit them or refresh them. In my case, I need to pull notifications from the database, as they are stored and logged.

Here is a video demo

https://user-images.githubusercontent.com/8517475/104217679-50122900-5401-11eb-9dc9-ee6d0a05a567.mov

### Exact steps to reproduce

Stripping this down but keeping it sufficiently realistic actually turned out incredibly long, if you don't mind, I'll just make a public repository, that can be pulled in, and edit this. I'll post it shortly.

This may be similar to issue https://github.com/livewire/livewire/issues/1971

**Context**
- Livewire version: [e.g. 2.3.6]
- Laravel version: [e.g. 8.21.0]
- Alpine version: [e.g. 2.8.0]
- Browser: [e.g. Chrome, Safari]

roni-estein:
test project created and working and simply demonstrates the above

https://github.com/roni-estein/livewire-2334

roni-estein:
I think the behavior that would solve this issue might be to to allow an opt in method of letting a component bust the cache on a "Full Page" reload. It's an edge cases that represents a small fraction of the traffic livewire handles and if it's opt in, you would only include it in troublesome components. Perhaps a setting that could be turned on or a trait included that might help middleware detect when to bust cache.

stevebauman:
Hi @roni-estein,

Unfortunately there isn't much we can do in terms of cache busting, because the cache control header isn't even used (to my knowledge) when loading a page from the back button. The browsers back button is a tricky beast to deal with.

The only way we could manage this is by hooking into the browsers navigation and detecting if the page has been loaded from a back/forward request.

Can you try placing this inside of your `app.js` file?

```js
require('./bootstrap');

if (window.performance && window.performance.navigation.type === window.performance.navigation.TYPE_BACK_FORWARD) {
    // Restart livewire on back or foward.
    Livewire.restart();
}

require('alpinejs');
```

Let me know if that works!

I'm going to do some more experimenting with this myself and see what I can come up with.

joshhanley:
I've been experimenting with this too, and if you create a middleware like described here https://stackoverflow.com/a/60671771/14909455 

```php
<?php

class DisableRouteCache
{
[...]

PR #2747, opened 2021-04-21
[Fix] back button cache issues

calebporzio:
This is a WIP based on problems outlined here: https://github.com/livewire/livewire/issues/2603


bastinald:
you rock dude

joshhanley:
@calebporzio I have done testing of this today as a follow up from the other day.

First off, I have updated the test so it is a more realistic test of what is happening. I have pushed the changes to my own branch here if you want to pull the changes in https://github.com/joshhanley/livewire/commit/2d59b564aa2fa73b21de1bb6b3548731bf939e44

The reason we weren't seeing the changes reflected after moving the redirect call was because (as you said) no HTML is being returned in the response.

We have to allow render by removing `$this->shouldSkipRender = true;` from the `ComponentConcerns/PerformsRedirects.php` methods (which I haven't done, I just updated the test)
https://github.com/livewire/livewire/blob/2e07fe08a792a6bbad8804ae6aeca1aa7b10b9f0/src/ComponentConcerns/PerformsRedirects.php#L9-L25

I did that, and I can see foo change to bar in the test before the redirect kicks in (have left the 500ms delay for now and added a delay to the test to compensate).

But when the browser redirects back, it is still pulling the old HTML from the browser cache, even after morphdom has had time to update the page (and you can see the changes). It seems like modifying the HTML on the page won't update the browser cache. I did a quick search and I couldn't see a way to use javascript to tell the browser to update its cache.

Which I think brings me back to my original idea on https://github.com/livewire/livewire/issues/2334#issuecomment-759114918 which is that browser cache should probably be disabled for Livewire components. Well it should be disabled for components that are not idempotent at least, just like Laravel disables browser cache for POST/PATCH/PUT/DELETE routes.

The hard part is, how do you tell which components modify data and which don't. Do you just disable browser cache for all Livewire components? What about scenarios where the route is a Laravel controller and the Livewire component is just added into the view?

calebporzio:
Yeah, these are great questions. Thanks for looking into this josh.

It's possible there's more digging to be done on why the browser isn't caching the latest html. it seems really of that or wouldn't.

As for disabling the cache, that would be ideal in a lot of ways, but would be a bold move. We'd want to make sure we've understood all the implications and only disabled it when necessary.

Thanks again for driving into this Josh

bastinald:
https://stackoverflow.com/questions/1195440/ajax-back-button-and-dom-updates

calebporzio:
Good reference @bastinald, can you please sum up what information is relevant from that post?

rginnow:
@calebporzio In regards to the link @bastinald posted, it seems it has to do specifically with browser caches only invalidating if there is an "onunload" event -- otherwise the browser caches the previous DOM state and the state of the page is reloaded nearly instantly by the browser. According to the [accepted answer](https://stackoverflow.com/a/1195934), only some browsers do this caching, and each browser handles it differently. A hacky method, provided by the [second answer](https://stackoverflow.com/a/14428894), would be to set the Cache-Control header as `cache-control: no-store, no-cache, must-revalidate`

I have no way to test this at the moment, but is there any way that Livewire could possibly inject an "onunload" event that will allow the browser to invalidate the page and not cache the page so it forces a fresh call from the server?

joshhanley:
Have marked this as a draft for now, but wanted to push up what I had so far.

How this will work is Livewire will have two new config options:
- `disable_back_button_cache` which will be true by default
- `should_skip_render_on_redirect` which will be false by default

And two new helper methods in components:
- `$this->enableBackButtonCache()`
- `$this->disableBackButtonCahce()`

The reason for needing the two different config options above is listed below.

## disable_back_button_cache

With the first config option, for Chrome to work currently (until it has a full implementation of BFCache) then browser cache needs to be disabled for any Livewire component, otherwise anyone using the back button in Chrome will get stale data. 

Disabling browser cache will be set to true by default for all Livewire components. 

This can be set to `false` in the config by setting `disable_back_button_cache` to `false`. 

It can also be set to `false` on the component level by calling `$this->enableBackButtonCache()`.

If disabled browser cache is set to `false` at a config level, then it can be set to `true` at the component level by calling `$this->disableBackButtonCache()`

## should_skip_render_on_redirect

The second config option `should_skip_render_on_redirect` is for `bfcache` in Firefox and Safari to work correctly. 

Currently when a redirect is fired, Livewire doesn't render the view before doing the redirect. But if that happens with a browser that has `bfcache` and the user goes back, they will get a stale view.

By making this a config option, rendering on redirect can be skipped if necessary, but by default the render will happen, ensuring the latest changes get written to the page before the redirect happens, which in turn means `bfcache` has the latest view.

Livewire already has a `$this->skipRender()` helper method, which can be called to override the config when `should_skip_render_on_redirect` is set to `false`, to set skip render to `true`.

I think there needs another helper method for the reverse of the above situation, where skip render is set to true by default in the config, but you want to allow render to happen for a specific component.

What should this be called? `$this->dontSkipRender()`?

Or should we change both and have `$this->enableSkipRender()` and `$this->disableSkipRender()`?

## Things left to do

- [ ] Change `ComponentConcerns\PerformsRedirects.php` to use config option `should_skip_render_on_redirect` and component override
- [ ] Consider adding the opposite of `$this->skipRender()`
- [ ] Clean up and update tests
- [ ] Merge latest changes from master and fix any conflicts (RenameMes moved to Features)

## Note

Something to note, I had to ensure the `DisableBrowserCache` middleware is disabled by default (even if the config is set to enabled), so that other routes that don't have a Livewire component in them aren't impacted by it. 

I then enable it in the constructor of `Component.php` if the config option is set to true, so it gets set if ever a Livewire component is constructed. Not sure if that is the best place for this or not.

Hope this helps!

joshhanley:
Ok, this PR is basically good to go. Will need reviewing though.

One thing to note is, the `setTimeout` still has a time of `50ms`. Ideally this should be `0ms`, but I couldn't get the tests to pass. So not sure how to halt the redirect in the test until the test has run some assertions.

Main two that need it are in `tests/Browser/Redirects/Test.php` on lines 39 and 60.

Hope this helps!

joshhanley:
Not sure what is going on with the tests. They were running fine earlier and the change I made hasn't impacted it as you can see some of them still passing.

nuernbergerA:
@joshhanley have the same issue https://github.com/livewire/livewire/pull/3554

calebporzio:
Latest decisions:
* 1 config item: `back_button_cache => false` (defaults to false)
* render before redirect ALL the time, UNLESS a person calls `$this->skipRender()`
* No per-component back button cache option (only config)

bastinald:
any idea when this will be released? need it super bad

thanks for all your work

joshhanley:
@bastinald this has now been merged, so will be in the next release 🙂

bastinald:
> @bastinald this has now been merged, so will be in the next release 🙂

has this been released yet? was playing around with the latest version and notice components still aren't preserving their mutated state when you go back to the page.

just curious, thanks again for all of this.

calebporzio:
This has been released, yes. 

Can you verify that when you hit the back button, you are getting a full page refresh? And that your browser isn't using it's back button cache to load the page

bastinald:
nevermind, i just tried the most recent version and it seems to be working great.

i made a public repo where i tested it out: https://github.com/bastinald/bas-live

thanks a million times!

calebporzio:
Awesome! Glad to hear it!

joshhanley:
@bastinald ah thanks for testing! Great to hear it's working 🙂

killgt:
We're having a serious issue with this middleware. Octane keeps the DisableBrowserCache middleware active, so we're having a no-cache header on the whole page, even if the request isnt a livewire one, and it's overwriting the our middleware cache headers. Any ideas? We've set the 
`
    'back_button_cache' => true,
`
to fix the problem for now

PR #2748, opened 2021-04-22
Add dots to make `the-tao.php` more consistent

dmitry-ivanov:
Just a small cosmetic fix to make `the-tao.php` more consistent - all sentences end with dots.


calebporzio:
Thanks!

PR #2754, opened 2021-04-22
Support using class names with assertSeeLivewire

jasonlbeggs:
In this PR, I added a few lines to support passing a component class string when calling `assertSeeLivewire()` and `assertDontSeeLivewire()`. I also added some additional test coverage for these two methods. Let me know if these tests would be more appropriate in the `LivewireTestingTest` class or somewhere else.


calebporzio:
Thanks @jasonlbeggs!

PR #2756, opened 2021-04-22
Sensitive Component Name

reziamini:
Now we can use names for our component which has been reserved by PHP and when we make these components we will get syntax error!
With this PR Livewire creates components more carefully.


joshhanley:
@rezaamini-ir great PR! My only feedback is in the future to provide some more information. For example, for this PR, add a screenshot of the syntax error and links to the documentation where you got the keywords, would really help support your submission.

I wasn't sure about this, so without the screenshot and docs link I had to test this myself and investigate, which it looks like you have already done, so adding that info here would have saved me having to do it too 🙂

This is the error you get if you try to create a Livewire component called `List`

image

See the full list of reserved keywords here, which even though some of them may be function names, they still can't be used as class names https://www.php.net/manual/en/reserved.keywords.php
> Some of them represent things which look like functions, some look like constants, and so on - but they're not, really: they are language constructs. You cannot use any of the following words as constants, class names, function or method names.

Good find though @rezaamini-ir.

Hope this helps!

rezaamini-ir:
@joshhanley Thanks for your opinion
But I think it's clear and doesn't need screenshot.
Also I have used Laravel Generator Class for these words and I believe it's sync with PHP reserved names.

calebporzio:
Thanks for the PR, and thanks for the nudge @joshhanley.

@rezaamini-ir - this type of thing DEFINITELY warrants more info. That screenshot was all I needed to convince me this is a good PR to merge. Otherwise, I wouldn't know the exact user experience of the problem without going and digging in myself. And because I haven't heard many others talking about this, I'd have probably closed it.

Thanks again to both of you!

PR #2765, opened 2021-04-25
Fix layoutData method bug

reziamini:
`layoutData` method had a problem and override other values.
reference: https://github.com/livewire/livewire/pull/2736#issuecomment-826234208


calebporzio:
Thanks!!

PR #2774, opened 2021-04-26
Make paginationTheme Configable

reziamini:
There was no way to set the default theme for all components, and we should have defined `paginationTheme` property for every component.
So with this PR and these changes, we will be able to define the default theme in Livewire config.


calebporzio:
Thanks for this. Can we instead resolve the default paginator (tailwind or bootstrap) using the same mechanism as Laravel? So that users who place this in their service providers:
image

Will just automatically get the right pagination from Livewire?

Going to close this. Please Re-open this PR with the above change if you desire. Thanks!

https://laravel.com/docs/8.x/pagination#using-bootstrap

Issue #2380, opened 2021-01-19
Livewire loosing $this with nested components in a blade component slot.


christophheich:
### Description
If a livewire component is used within a blade component within a livewire component there is no way to access a variable anymore, and an exception will be thrown. Using $this when not in object context.

https://github.com/livewire/livewire/discussions/2313#discussioncomment-293525
https://github.com/livewire/livewire/discussions/2313#discussioncomment-285714

### Exact steps to reproduce

Create a livewire component and a nested livewire component as well as blade component with a named or default slot.

Add the blade component into the livewire component and add the nested component into the blade component.

Trying to access a livewire variable (declared / initialized in the parent livewire component) in the blade component should now throw the exception, if the nested component in the blade slot will be removed, it works as expected.

```html
(Livewire Component)
<div>
    <h1>Random</h1>
    (Blade Component)
    <x-random-blade-component>
        (Nested Livewire Component in a blade slot)
        @livewire('random-nested-livewire-component')
    </x-random-blade-component>
</div>
```

### Stripped-down, copy-pastable code snippets

```html
(Livewire Component)
<div>
    <h1>Random</h1>
    (Blade Component)
    <x-random-blade-component>
        (Nested Livewire Component in a blade slot)
        @livewire('random-nested-livewire-component')
    </x-random-blade-component>
</div>
```

**Context**
- Livewire version: 2.3.6
- Laravel version: 8.22.1
- Alpine version: 2.8.0
- Browser: Chrome, Firefox

redbastie:
You can't use `$this` inside of an included view. You have to use `$this` directly inside the parent view.

redbastie:
What you can do for computed properties, is pass them to the include as a variable. For example

    @include('includes.infinite-scroll', ['infiniteScroll' => $this->infiniteScroll])

christophheich:
> What you can do for computed properties, is pass them to the include as a variable. For example
> 
> ```
> @include('includes.infinite-scroll', ['infiniteScroll' => $this->infiniteScroll])
> ```

Okay, i'll try to explain it on another way.

If i am using a livewire component, which has a blade component (https://laravel.com/docs/8.x/blade#components) inside of it, i can call $this in the blade component and it will work just fine. The blade component can access $this from the parent livewire component.

```
(Livewire Component)
<div>
    <h1>Random</h1>
    <x-random-blade-component /> ($this will be called inside the blade component and it just works)
</div>
```

[...]

Issue #1653, opened 2020-09-20
$this looses its context with nested component in advanced layout


tanthammar:
### Description
I have a livewire component blade layout, with a **VERY large amount** of `@include` and `<x-blade-components>`
When I add a livewire nested (child) component to the layout, I am unable to use `$this->someMethod()` after, or in the declaration of that component.

- If I remove the child component all is working.
- If I use the child component on its own, it is working.
- If I use the child component nested in the parent, but only 2 levels deep, all is working

I have had this issue before in another context #1401 

### Exact steps to reproduce
It is hard to give an exact example, because the layout has a lot of nested elements. In short:
```blade
<div>//start parent component
    @include('some-blade-file') // multiple levels of @includes
      <x-some-blade-component> // a lot of blade components
          <livewire:child-component 
                     :some-prop="$this->someMethod()" // here unable to use $this
         />  
      </x-some-blade-component>
    @include('some-other-blade-file') // everything after the child component cannot use $this
</div>
``` 

### Error
````
Call to undefined method Livewire\LivewireViewCompilerEngine::someMethod()
````

### Stripped-down, copy-pastable code snippets
I have a private repo that i do not wish to share in public but I can send a link in private PM.

**Context**
- Livewire version: v2.2.5
- Laravel version: v7.28.3

calebporzio:
The code provided is not copy-pastable.

Closing this for now.

Please resubmit with a Laravel Playground link demonstrating this issue. Thank you.

tanthammar:
Solved with  #2786. THANK YOU @joshhanley

Issue #1401, opened 2020-08-21
$this looses its context depending on where a file input is placed in the component view


tanthammar:
livewire v1.3.2
 `$this` looses its context depending on where the input is positioned in the form. 
If the field is in the top of the form, $this = the livewire component.
If it is positioned at the end of the form `$this` becomes the LivewireViewCompilerEngine.

Example:
```blade
@error('photos.*')
   <pre>@json($photos)</pre>
   @php $this->deleteAllTempFiles('photos') @endphp
@enderror 
```
Error:
 `Call to undefined method Livewire\LivewireViewCompilerEngine::deleteAllTempFiles()`.

Here you can see the @json($this) in the middle of the form = the livewire component

![](https://cdn.discordapp.com/attachments/698231501501890600/746480557092896858/unknown.png)

When the field is in the end of the form it becomes the LivewireViewCompilerEngine.
![](https://cdn.discordapp.com/attachments/698231501501890600/746481056655474718/unknown.png)

Please note that it is an extremely bad idea :) to call a method, that sets a component property, if the form has errors. It causes a loop with ajax requests. (But that is not the topic of this question.)

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

tanthammar:
Solved with  #2786. THANK YOU @joshhanley

PR #2786, opened 2021-04-29
Fixed: Livewire subcomponent clears render context after it is finished

joshhanley:
## Issue

This PR adds a failing test for the issue raised in #2313 and #2380.

If you make use of `$this` within a blade component to access the main Livewire component, it works fine. 

But when you add a Livewire subcomponent to the page and then try to add a blade component after the subcomponent that tries to access `$this` to access the main Livewire component, you get the following error
> Using $this when not in object context

This test demonstrates the issue.

## Cause

I have dug into what causes the issue, and the problem exists in Livewire's blade compiler engine.

What happens is in ComponentConcerns/RendersLivewireComponents.php the Livewire component is bound to the `$this` context.

At the top of the same function it does a check for `isRenderingLivewireComponent`, which if true, enables the render context to be set to the Livewire component, which is also applied to blade components.

https://github.com/livewire/livewire/blob/33101c83b75728651b9e668a4559f97def7c9138/src/ComponentConcerns/RendersLivewireComponents.php#L36-L51

The issue is though, that after a sub component is finished rendering it calls the `endLivewireRendering` method which sets `isRenderingLivewireComponent` to false.

https://github.com/livewire/livewire/blob/33101c83b75728651b9e668a4559f97def7c9138/src/ComponentConcerns/RendersLivewireComponents.php#L24-L27

This then stops the blade components from having this context set, throwing the above error as there is no `$this` set at all.

## Temporary Fix

As a temporary solution (note this is a hack as a word around for the moment), I have added a blade directive to my `AppServiceProvider`

```php
Blade::directive('livewireResetContext', function() {
  return '<?php app("view.engine.resolver")->resolve("blade")->startLivewireRendering($this); ?>';
});
```

Then in my Livewire component blade view, I add the call to that blade directive after any sub component
```blade
<livewire:subcomponent />
@livewireResetContext
```

## Possible Solution

To fix this, I think we need to change the boolean to an array and "stack" each components name into the array and use the last one as the context. Then when a component has finished it can pop itself off the end of the stack. The check can then see if there is anything left in the stack as to whether to bind a Livewire context or not.

The regular blade compiler makes use of a similar concept with `lastCompiled` where each view is pushed on to the "stack" (array) and removed once it has finished compiling.

Happy to give it a crack if you think it is suitable.

Hope this helps!

### DIFF
--- a/src/ComponentConcerns/RendersLivewireComponents.php
+++ b/src/ComponentConcerns/RendersLivewireComponents.php
@@ -12,28 +12,26 @@ use Throwable;
 
 trait RendersLivewireComponents
 {
-    protected $livewireComponent;
-    protected $isRenderingLivewireComponent = false;
+    protected $livewireComponents = [];
 
     public function startLivewireRendering($component)
     {
-        $this->livewireComponent = $component;
-        $this->isRenderingLivewireComponent = true;
+        $this->livewireComponents[] = $component;
     }
 
     public function endLivewireRendering()
     {
-        $this->isRenderingLivewireComponent = false;
+        array_pop($this->livewireComponents);
     }
 
-    public function setLivewireComponent($component)
+    public function isRenderingLivewireComponent()
     {
-        $this->livewireComponent = $component;
+        return ! empty($this->livewireComponents);
     }
 
     protected function evaluatePath($__path, $__data)
     {
-        if (! $this->isRenderingLivewireComponent) {
+        if (! $this->isRenderingLivewireComponent()) {
             return parent::evaluatePath($__path, $__data);
         }
 
@@ -48,7 +46,7 @@ trait RendersLivewireComponents
             \Closure::bind(function () use ($__path, $__data) {
                 extract($__data, EXTR_SKIP);
                 include $__path;
-            }, $this->livewireComponent ? $this->livewireComponent : $this)();
+            }, end($this->livewireComponents))();
         } catch (Exception $e) {
             $this->handleViewException($e, $obLevel);
         } catch (Throwable $e) {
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks @joshhanley , all of the above makes sense. I think you're right on with needing to convert to a stack. Of you're up for it, I'd put course love a PR. Thanks man!

joshhanley:
@calebporzio have pushed up the fix 🙂

tanthammar:
🎉  Confirmed! 
This solves #1653, #1401, #2380, #2313

@joshhanley 
🌟  You are a lifesaver!!!!!!
Amazed by your perseverance to fix this and other issues with Livewire as well as your politeness to the community.

@calebporzio 
This fix is actually a game changer for Livewire. 
**I hope you give @joshhanley a real treat** for solving this, (and other) issues :)
It held me and others back to create more advanced Livewire components that requires nesting mixed with blade components.

Now I'll go out on my balcony and do some crazy jumping while shouting out to the world that the lid finally popped off the Livewire jar 💖  Then I'll remove some nasty hacks from the tall-forms package that worked around this issue.

Sending LOVE!

src/ComponentConcerns/RendersLivewireComponents.php
```diff
@@ -12,28 +12,26 @@
...
     {
-        $this->isRenderingLivewireComponent = false;
+        array_pop($this->livewireComponents);
     }
 
-    public function setLivewireComponent($component)
+    public function isRenderingLivewireComponent()
     {
-        $this->livewireComponent = $component;
+        return ! empty($this->livewireComponents);
```
tanthammar, line 29: Maybe it's just me but I prefer:
```php 
return filled($this->livewireComponents);
```
My brain get's confused by negations.

calebporzio:
Thanks Josh! You're the man!

joshhanley:
No worries 🙂 I'm glad it helps!

SalatielMG:
Thanks @joshhanley!, that solved my prroblem

PR #2811, opened 2021-05-03
add config option for detecting serverless

timrspratt:
Hi, hopefully this is ok to add? I'm using bref to serve my application serverless using AWS Lambda. Currently, during the deploy process, I'm having to `sed` replace `vapor` with `bref` and it just feels dirty 🙈

My thinking behind this, is that the default livewire config will still retain the reference to `vapor`, and allow developers to specify the value in their application's overridden livewire config. I've added an "alias" method too since if this value is changeable `isRunningServerless` felt more natural. I haven't changed the 2 references to the `isOnVapor` but if you like this idea I can happily change them to reference the new method. I've left the `isOnVapor` to prevent breaking changes.

Thanks

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -110,4 +110,20 @@ return [
 
     'manifest_path' => null,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Serverless Identifier
+    |--------------------------------------------------------------------------
+    |
+    | This value is used to check the $_ENV['SERVER_SOFTWARE'] to determine
+    | the assets and manifest paths.
+    |
+    | Examples:
+    |       - for Laravel Vapor, it would be "vapor".
+    |       - for Bref, it would be "bref".
+    |
+    */
+
+    'serverless_identifier' => 'vapor',
+
 ];
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -110,20 +110,4 @@ return [
 
     'manifest_path' => null,
 
-    /*
-    |--------------------------------------------------------------------------
-    | Serverless Identifier
-    |--------------------------------------------------------------------------
-    |
-    | This value is used to check the $_ENV['SERVER_SOFTWARE'] to determine
-    | the assets and manifest paths.
-    |
-    | Examples:
-    |       - for Laravel Vapor, it would be "vapor".
-    |       - for Bref, it would be "bref".
-    |
-    */
-
-    'serverless_identifier' => 'vapor',
-
 ];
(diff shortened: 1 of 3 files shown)


stancl:
Public static properties are a good way to add configuration for things that 99.9% of users won't change, but should be changeable *somehow*.

Then you could just do

```php
LivewireManager::$serverlessIdentifier = 'bref';
```

in a service provider.

I'd go with that instead of config, to avoid making the config file unnecessarily long when most people won't need this.

timrspratt:
@stancl I'd be happy with either option. Thoughts @calebporzio? The only thought I have with the static property is that it isn't (to my knowledge) used anywhere else in Livewire (other than a testing flag), and so is a "foreign" concept.

https://github.com/livewire/livewire/pull/2708 relates to vapor, I wonder if there could be a nested config for both the `identifier` and `manifest_path`? Whilst `$_ENV['LAMBDA_TASK_ROOT']` makes the manifest path dynamic, it also limits use to Lambda too. Providing a nested `serverless` array in the livewire config would provide a dedicated location to configure the serverless options.

calebporzio:
I'm very hesitant to add or change config items.

Also, I'm open to a static property, however it would then feel like serverless config is sperated and belongs together.

Can we introduce support for bref without config at all? Rename isOnVapor, and use an "or" statement to check for bref?

stancl:
Caleb has a good point that it can just be hardcoded in. If no one needed bref until now, it's unlikely that someone would need yet another option. This is < 1% of users and configs are an issue-prone thing. Worst case scenario is just someone opening a PR to add another option.

timrspratt:
Agreed, thanks for the input guys! I've adjusted this PR to reflect what has been discussed. Please let me know if it needs tweaking to coding standards. I've re-named the `isOnVapor`, but with it being `public` I'm concerned others may be using it? Is this an issue, and should I add it back in as a alias, or is that overthinking?

stancl:
I think you should call `isRunningServerless()` from `isOnVapor()` as you did before in case someone made a change, e.g. to add bref support, by replacing the class in the service container. And like you said, someone may be calling it from other places too. Then in the next major version the alias should be dropped.

Not 100% sure, but I think I was using the `Livewire::isOnVapor()` thing in an app at some point.

calebporzio:
Yeah, we need to add it back as an alias

timrspratt:
Sure thing, @calebporzio I've added in the alias

calebporzio:
Thanks!!

PR #2816, opened 2021-05-04
File upload refactor — supports S3 multi-upload

jameslkingsley:
This is a work-in-progress PR that refactors the file upload system. The reason for this PR is to support uploading multiple files to S3, and to offer more control over the individual files being uploaded from the client-side.

With this refactor, you can now do this:
```html
<div x-data="{ uploads: {} }">
    <template x-for="[id, upload] in Object.entries(uploads)">
        <div>
            <img :src="URL.createObjectURL(upload.file)" />
            <span x-text="upload.file.name"></span>
            <button @click="upload.abort()">Abort</button>
        </div>
    </template>

    <input
        x-on:livewire-upload-start="uploads[$event.detail.pendingUpload.id] = $event.detail.pendingUpload"
        x-on:livewire-upload-abort="delete uploads?.[$event.detail.pendingUpload.id]"
        wire:model="files"
        type="file"
        multiple
    />
</div>
```

Of course this is an optional way of displaying the files, but if you wanted to you could now show the preview image, and even abort the upload request. If you'd rather keep things server-side, that works too:
```blade
<div>
    <span wire:loading wire:target="files">Uploading...</span>

    <div>
        @foreach ($files as $file)
            <div>
                <img src="{{ $file->temporaryUrl() }}" />
                <span>{{ $file->getClientOriginalName() }}</span>
            </div>
        @endforeach
    </div>

    <input
        wire:model="files"
        type="file"
        multiple
    />
</div>
```

Biggest change to the code that you'll notice is the bulk of the work is being done in `PendingUpload`, as this offers more control as seen above.

**Benefits of this refactor:**
- Multi-file uploads to S3 now supported
- All files are uploaded individually
    - This might be up for debate — uploading individually results in more Livewire requests and _things happening_ but it also means uploading many files will be faster (as they're sent in parallel). We can also handle errors individually, so if a single file fails to upload, it doesn't affect the rest.
- Access to the `PendingUpload` instance from event handlers
- Should be simpler to maintain

**Dependencies:**
- [UUID](https://www.npmjs.com/package/uuid) — used to assign a unique ID to each file upload (safer than relying on file names or array indexes)

**To do:**
- [x] `wire:loading wire:target="files"` should target the upload bag, not each file
- [ ] Handle upload errors and removing uploads
- [ ] Events for upload bag _and_ each file
- [ ] Backdoor hooks
- [ ] Tests
- [ ] Handle `wire:model.defer`?
    - Not sure if this would introduce issues, but it would be cool if `.defer` worked on file inputs. It would not upload anything until you perform an action like "submit". Potential issue is if said action removes the file input unexpectedly.


jameslkingsley:
Reason for the backdoor in `TemporaryUploadedFile` is to do something like this:
```php
class TemporaryUploadedFileDecorator
{
    public function __invoke(TemporaryUploadedFile &$instance)
    {
        $instance->cache = Cache::remember($instance->getPath().'_metadata', now()->addMinutes(10), fn () => [
            'file_name' => $instance->getClientOriginalName(),
            'mime_type' => $instance->getMimeType(),
            'size' => $instance->getSize(),
        ]);
    }
}
```

This is especially useful if you're using S3 and don't want to slow down the app with multiple calls to S3. Might be a better way of handling this, but I don't think it's Livewire's responsibility to do so.

jameslkingsley:
@calebporzio is there a reason why `syncInput` is only called for singular files, and not arrays?

https://github.com/livewire/livewire/blob/6c45c16952a23447474079613d87af38669b13f8/src/WithFileUploads.php#L88

jameslkingsley:
Waiting on any feedback as to whether this is wanted / changes in direction.

joshhanley:
@jameslkingsley it's definitely something I'd make use of!

I'd assumed that there was an issue with S3 as to why there wasn't already multiple uploads, but I haven't looked into it.

calebporzio:
I'm closing this as it's not finalized. Please reopen separate smaller PRs. I think the ideas in this PR are valid, however, it should not be tackled all in one single PR. Thanks!

Issue #1672, opened 2020-09-23
Carbon parsing error on datetime cast when binding directly to model properties


richbkuk:
### Description
When a model has an attribute cast as datetime and you're binding directly to model properties in Livewire, if the user submits a value that Carbon cannot parse, Carbon returns an error when ideally, validation should have picked up it up first before the value is applied to the model.
I believe the issue is that the value submitted is being applied to the model before validation takes place?

### Exact steps to reproduce
- Create a model with an attribute cast as datetime
- 'Make' the empty model in the mount() Livewire method
- Add a validation entry for it, as per the docs
- From the view (front end form), submit an invalid date format such as 'non-date'
- Receive error: Could not parse 'non-date': DateTime::__construct(): Failed to parse time string (non-test) at position 0 (n): The timezone could not be found in the database

### Stripped-down, copy-pastable code snippets
Model:
```php
class Event extends Model
{
    protected $casts = [
        'start' => 'datetime'
    ];
}
```

Livewire Component:
```php
class CreateEvent extends Component
{
    public $event;
	
    protected $rules = [
        'event.start' => 'required|date',
    ];
	
    public function mount() {
        $this->event = Event::make();
    }
	
    public function submit() {
        $this->validate();

        $this->event->save();
    }
	
    public function render() {
        return view('livewire.create-event');
    }
}
```

Livewire View:
```html
<form wire:submit.prevent="submit">
    <input type="text" wire:model.defer="event.start" />
    <button type="submit">Submit</button>
</form>
```

**Context**
- Livewire version: 2.2.5
- Laravel version: 8.6.0
- Browser: Chrome 85.0.4183.121
- PHP: 7.3

uteq:
What I did to resolve this is added the start date to the models $dates property and removed it from the $casts. This prevented throwing the error. This is more a workaround I guess than an actual solution. Hopefully this helps someone.

JuanRangel:
After running into the same issue I did some digging and noticed that when trying to set my date attribute with Livewire it was never reaching my livewire method. The error seems to happen HasAttribute trait in Laravel. 

Even if you remove the date validation the error still exists

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

[...]

PR #2827, opened 2021-05-06
Add a way to bail out of property updates

jameslkingsley:
This is a backdoor for cancelling a property update by just returning `false` in the `updating()` hook.

Addresses https://github.com/livewire/livewire/discussions/2349

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -75,7 +75,9 @@ trait HandlesActions
 
         $name = $name->__toString();
 
-        $this->updating($name, $value);
+        if ($this->updating($name, $value) === false) {
+            return;
+        }
 
         if (method_exists($this, $beforeMethod)) {
             $this->{$beforeMethod}($value, $keyAfterFirstDot);
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -91,7 +91,9 @@ trait HandlesActions
             }
         }
 
-        Livewire::dispatch('component.updating', $this, $name, $value);
+        if (Livewire::dispatch('component.updating', $this, $name, $value) === false) {
+            return;
+        }
 
         $callback($name, $value);
 
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -91,7 +91,7 @@ trait HandlesActions
             }
         }
 
-        if (Livewire::dispatch('component.updating', $this, $name, $value) === false) {
+        if (Livewire::dispatchStrict('component.updating', $this, $name, $value) === false) {
             return;
         }
 
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -80,11 +80,15 @@ trait HandlesActions
         }
 
         if (method_exists($this, $beforeMethod)) {
-            $this->{$beforeMethod}($value, $keyAfterFirstDot);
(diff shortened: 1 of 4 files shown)


joshhanley:
I think this could be useful as there are situations where validation can't even run due to type casting errors.

For example, in issue #1672 if you have an input that is bound to a model property that is cast to a carbon instance, and a user has entered a string into the input field, then you get a carbon parse error.

It would be nice to find a way to run validation against the value of a property to be updated before actually gets applied, and cancel the update if it fails, like in this PR.

I think this PR is only part of the story though, as the example in #2349 validation wouldn't actually work, as the validation runs against the value of the property, not the incoming change, hence why validation is typically run in an `updated` hook rather than `updating`.

I know a manual validator could be used in updating instead, but it would be handy if there was a way to use the components in rule/s but run against the update value.

There are also some other instances where this might be useful if there are multiple updates coming from the front end in the same request, whether it be due to defer or alpine triggering a couple of updates simultaneously, that you may want to cancel some of the changes. I can provide some examples of this if need be.

Hope this helps!

jameslkingsley:
> It would be nice to find a way to run validation against the value of a property to be updated before actually gets applied

The way I've handled this so far:
```php
public function updating($attribute, &$value)
{
    // $value is passed by reference so we can modify it before it's used by Livewire
    $value = 'something else';
}
```

stancl:
Does this work with `updatingFoo()`?

jameslkingsley:
Now it does 😉

calebporzio:
This makes a lot of sense to me, is this a breaking change though?

stancl:
I think it should be fine with `=== false`. Sure, someone might return `null` from an updating hook to return early, but if you're specifically returning `false` from it, you're probably looking for this behavior.

calebporzio:
Yeah, I think I agree @stancl, @jameslkingsley ?

jameslkingsley:
Yeah I agree. I'm gonna push some tests btw.

joshhanley:
@jameslkingsley yeah I'm not sure about the traits handling, but the rest looks good 🙂

I'd imagine if one of the traits returned false, that the rest of the update is cancelled (including other traits), instead of continuing to loop through the other traits before finally returning false. Same for the dispatch of the listeners.

Keep in mind, that if you "cancel" an update, it will only be for that particular property that is being updated. So if you had batched updates, and say returned false for the first property update (as such was cancelled), the second property update will start from the beginning, running through all update methods and traits.

Hence why I think if a trait returns false, then just to stop the update there instead of running through the rest of the traits.

What do you think?

jameslkingsley:
Yeah that makes a lot of sense, also simplifies the implementation

jameslkingsley:
Made it an opt-in thing internally. `dispatchStrict()` will return early if one of the listeners returns `false`. Open to better method name.

joshhanley:
> Does it matter that the early exit will happen for other events though? `LivewireManager::dispatch` might not run all listeners if one of them returns `false`. Quite a hard to spot bug.

@jameslkingsley I had that first thought too, but actually when you follow it through, it will only run through listeners with that exact event name `'component.updating'` for example.

```php
foreach ($this->listeners[$event] ?? [] as $listener)
```
So it shouldn't block any other events 🙂  does that sound right to you?

Probably don't need the `dispatchStrict` in that instance if I'm following it correctly.

jameslkingsley:
I mean for other events that are dispatched with the `dispatch` method.

calebporzio:
Sorry for the lag on this. I like this addition. Not sure about adding "dispatchStrict" - is there an implementation that doesn't add any extra API, but adds this behavior? Closing this, but please resubmit with the changes and I'll pull in. Thanks!

PR #2841, opened 2021-05-07
[Fix] FileUploadHandler didn't handle long filenames

### LINKED ISSUE #2834: FileUploadHandler didn't handle long original filenames.
### Description
- Related discussion #2238 #2813 
> When uploads with a long filename, it throws an exception:
> `fopen(<path with long filename>): failed to open stream: File name too long`

- In the process of encoding the original filename to base64.
- Increases filename length compared to the original filename.
https://github.com/livewire/livewire/blob/b0f4026509604f5239813a605ba7172f4bb5fa0e/src/TemporaryUploadedFile.php#L126-L133
- FileUploadHandler isn't catching that increases filename could be over the maximum length of filename.
https://github.com/livewire/livewire/blob/b0f4026509604f5239813a605ba7172f4bb5fa0e/src/Controllers/FileUploadHandler.php#L32-L42
### Exact steps to reproduce
- Select a file with a long filename in the file input form.

UploadTest.php
```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;
use Livewire\WithFileUploads;

class UploadTest extends Component
{
    use WithFileUploads;

    protected $rules = [
        'file' => 'required',
    ];

    public $file;

    public function render()
    {
        return view('livewire.upload-test');
    }
}
```

upload-test.blade.php
```html
<div>
    <input wire:model="file" type="file">
</div>
```

**Context**
- Livewire version: 2.4.3
- Laravel version: 8.39.0
- Alpine version: 2.8.2
- Browser: Chrome

ZerglingGo:
Fixes #2834 

* Add checking the filename lengths
* Add test
* Add skip renaming in test syncUploadedFiles() when a blank file

### DIFF
--- a/src/Controllers/FileUploadHandler.php
+++ b/src/Controllers/FileUploadHandler.php
@@ -2,9 +2,11 @@
 
 namespace Livewire\Controllers;
 
+use function Livewire\str;
 use Livewire\TemporaryUploadedFile;
 use Livewire\FileUploadConfiguration;
 use Illuminate\Support\Facades\Validator;
+use Illuminate\Validation\ValidationException;
 
 class FileUploadHandler
 {
@@ -35,6 +37,10 @@ class FileUploadHandler
 
         $fileHashPaths = collect($files)->map(function ($file) use ($disk) {
             $filename = TemporaryUploadedFile::generateHashNameWithOriginalNameEmbedded($file);
+
+            if (str($filename)->length() >= 256) {
+                throw ValidationException::withMessages(['files.0' => 'File name too long']);
+            }
 
             return $file->storeAs('/'.FileUploadConfiguration::path(), $filename, [
                 'disk' => $disk
--- a/src/Testing/Concerns/MakesCallsToComponent.php
+++ b/src/Testing/Concerns/MakesCallsToComponent.php
@@ -112,10 +112,20 @@ trait MakesCallsToComponent
         // We are going to encode the file size in the filename so that when we create
         // a new TemporaryUploadedFile instance we can fake a specific file size.
         $newFileHashes = collect($files)->zip($fileHashes)->mapSpread(function ($file, $fileHash) {
+            // Skip appending file sizes in the file name of blank file
+            // Because testing the maximum length of file name cause that throwing "File name too long" errors.
+            if ($file->getSize() === 0) {
+                return $fileHash;
+            }
+
             return (string) str($fileHash)->replaceFirst('.', "-size={$file->getSize()}.");
         })->toArray();
 
         collect($fileHashes)->zip($newFileHashes)->mapSpread(function ($fileHash, $newFileHash) use ($storage) {
+            if ($fileHash == $newFileHash) {
+                return;
+            }
+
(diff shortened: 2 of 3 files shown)


calebporzio:
Thanks for this! Just so I'm clear, where is the 256 restriction coming from? The browser? Server?

I'm wondering because it's it something that can be configured to allow for more, making this a bad idea to hard code?

joshhanley:
At a quick look the 255 is the character limit for most common disk formats (APFS, exFAT, NTFS, etc) https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

I think the issue isn't necessarily in the filename length of the original file, but also how deep the tmp livewire folder is stored.

For example on #2238 the error has the full path and file name (see error below).

> "fopen(/var/www/html/multiup/storage/app/livewire-tmp/LYCKtev7fgBxuTDjG2EJcNsSbs6bKD-metaMjAyMC0xMi0wOS3guKPguLLguKLguIfguLLguJnguKrguKPguLjguJvguJzguKXguIHguLLguKPguJTguLPguYDguJnguLTguJnguIHguLLguKPguIjguLHguJTguIvguLfguYnguK3guIjguLHguJTguIjguYnguLLguIfguYPguJnguKPguK3guJrguYDguJTguLfguK3guJkg4Lie4Lik4Lio4LiI4Li04LiB4Liy4Lii4LiZ4LieLuC4qC4yNTYzLnBkZg==-.pdf): failed to open stream: File name too long"

I think actually base_64 encoding the original file name in the hash is what is actually causing the issues, as the results gets longer depending on the length of the file name. I wonder if we could use something else to encode the filename which is of fixed length, no matter the length of the string passed in?

ZerglingGo:
I referenced the same page linked by @joshhanley.

I can't think of anything about fixed length decryptable string.
Another way that comes to me is to store the file with fixed length hash name, and create a metadata file for each file, and store the original filenames.
```
// upload file name
longfilenames.png

// In livewire tmp folder
520f639156198cf82b96767c695081dde6ccb6c7ebcf7bb73e2d750897da61e7 // longfilenames.png
520f639156198cf82b96767c695081dde6ccb6c7ebcf7bb73e2d750897da61e7.meta // contains original filename
```

calebporzio:
Yeah, I'm pretty rusty on the internals of that system, but it makes sense to explore not adding the original name to the filename.

calebporzio:
Sorry, can someone refresh me on what's the best way to move forward here?

ZerglingGo:
Sorry, I was busy working for business.
There has two solutions getting original filename:
1. Create metadata file
2. Store in property (I couldn't proof it because I didn't know logic of how can that part worked.)

I'm curious about other people's opinions to 2nd solution.

phattarachai:
Hi, I am still facing this issue? Is there a way to workaround this? 

Thanks.

Mattia-Marchiorato:
Hello, any update for this issue? What about the pull request?

calebporzio:

DanielGSoftware:
We're running into the same issue. I'm not sure if I'm the right person to try to fix this, but I'll have a quick look.

DanielGSoftware:
Nevermind, I found this open [PR](https://github.com/livewire/livewire/pull/8690) to solve this issue hopefully!

Issue #2425, opened 2021-01-30
wire:loading doesn't work with arrays


stancl:
### Description

`wire:loading` doesn't react to changes _inside_ arrays. It only runs listeners for e.g. `foo`, but not `foo.bar`.

### Exact steps to reproduce

Create a component that has a property like `public $foo = ['bar' => 'hello']`.

And use `wire:loading="foo.bar"`.

### Stripped-down, copy-pastable code snippets

I think this can be solved by adding this:

```js
.filter(action => {
    return action.type === 'syncInput'
})
.map(action => action.payload.name)
.map(model => {
    // Model is an array
    if (model.includes('.')) {
        parts = model.split('.')
        parts.pop(); // Remove last item

        return parts.reduce((result, part, index) => {
            if(parts[index-1] !== undefined) {
                parent = parts[index-1] + '.'
            } else {
                parent = ''
            }

            return result.concat([parent + part])
        }, [])
    }

    return model;
})
.flat();
```

Somewhere between this: https://github.com/livewire/livewire/blob/2ff70fdd893ecf9860fb70105efbc81aee93a02a/js/component/LoadingStates.js#L44-L50

It makes sure that when `foo.bar.baz` is updated, it will fire listeners for `foo` and `foo.bar` as well.

**Context**
- Livewire version: 2.3.8
- Laravel version: latest
- Alpine version: 2.8.0
- Browser: Chrome

PR #2843, opened 2021-05-08
Add loading state helpers to component class

jameslkingsley:
This is a small addition to the client-side component class, that'll let you control the loading state for a component. Particularly useful in custom directives or very specific use-cases where you just wanna say "hey this component is processing this thing".

Note: there's no special handling for resuming a loading state when Livewire re-renders the page. This is something the developer would need to cater for in their own implementation.


joshhanley:
@jameslkingsley I think could come in handy! 

Can you give an example of how that would look in code to use?

Would this only work for Livewire actions/models or could you trigger Livewire's loading state say from Alpine while it was making an external API call or something?

jameslkingsley:
It's available anywhere within a Livewire component.
All you need is the client-side component instance, which you can get a couple of ways:

**Using `$wire`**
```html
<div x-data>
  <button x-on:click="$wire.__instance.setLoading('foo')">Load</button>
  <span wire:loading wire:target="foo">Loading...</span>
</div>
```

**Using `el.__livewire`**
```html
<div wire:id="..." id="componentRootNode">
  <!-- document.getElementById('componentRootNode').__livewire.setLoading('foo') -->
  <span wire:loading wire:target="foo">Loading...</span>
</div>
```

joshhanley:
@jameslkingsley thanks for that! It helps to be able to see possible implementations of it.

calebporzio:
This is interesting. Because it adds extra maintenance and API surface area, I'm going to hold off until there's more demand for it. Thanks!

stancl:
@calebporzio Maybe this could help with https://github.com/livewire/livewire/issues/2425?

Right now Livewire doesn't expose *any API at all* that would let the developer affect the loading states. If it did, I think devs could do:

```js
Livewire.hook('message.sent', (component, message) => {
    if (message.something) {
        component.setLoading('foo');
   }
})
```

PR #2864, opened 2021-05-12
Add support for assertNotRedirected

leganz:
There is currently no discussion or issue for that. But i guess a test helper should be ok?

No

Yes 2 test. 1 positiv and 1 negativ test

For example, if a livewire componente has a click event in which a redirect only occurs under certain conditions, there is currently no way to test this. This assert helps to fill the gap.

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -2,9 +2,9 @@
 
 namespace Livewire\Testing\Concerns;
 
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
-use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks for this, as this assertion is not available in Laravel, I'm wondering if assertNotRedirected is the best opposite of "assertRedirect"?

Like, should it be "assertNoRedirect"? Instead?

I know it doesn't read as well, but it's called "assertRedirect" not "assertRedirected" you know?

Thoughts? Other examples for precedence here? Thanks!

leganz:
@calebporzio thanks for the feedback! I completely agree that "assertNoRedirect" would fit better. I had initial doubts about "assertNotRedirected" myself.

I personally don't care what the method is called in the end though, as long as I can test the desired behavior.

I updated the commit.

calebporzio:
Thanks!

PR #2868, opened 2021-05-12
Allow overwrite phpunit.xml locally

nuernbergerA:
phpunit looks for `phpunit.xml` then fallbacks to `phpunit.xml.dist`

So if we rename `phpunit.xml` to `phpunit.xml.dist` it allows us to have some variants in local development (its only relevant for livewire contributors).

### Why is this a thing?
To run dusk tests with UI on my Linux maschine i need to set this in my `phpunit.xml`:

```xml
<phpunit>
    ...
    <php>
        <env name="DISPLAY" value=":1" force="true"/>
    </php>
</phpunit>
```

We could use this also to set the preferred way to run dusk tests (with/without UI)

---
https://stackoverflow.com/questions/44764023/is-there-any-difference-in-naming-the-phpunit-configuration-file-phpunit-xml-dis#answer-44764078


calebporzio:
I'm down, thanks @nuernbergerA

Issue #677, opened 2020-03-04
Reactive nested component parent parameters.


kdion4891:
As per the docs:

> Nested components CAN accept data parameters from their parents, HOWEVER they are not reactive like props from a Vue component.

Is the only way to accomplish this via listeners so you have a pseudo parent-child reactive property?

calebporzio:
I imagine there is a way to make this sort of thing possible, but there would be lots of issues I can see off the bat.

For example:
- would we re-fire the mount method when data changes? (we can't refire mount after the component has already been mounted)
- we would need to somehow track incoming state so we know when it changes, we could do this with a hash or something but would need to store it somewhere, it would either add size to the payloads or we could store it in the backend cache or session but I'm trying to avoid both of those things where possible.

For now, I'm going to close this as it's a little too much ATM.

If anyone has any other clever ideas or thinks this really should exist, chime in here so we can gauge interest.

Thanks!

philipboeken:
Hi @calebporzio, I've created a [PR](https://github.com/livewire/livewire/pull/2905) which (partly) implements reactive properties by firing events from a HydrationMiddleware. Any thoughts?

PR #2905, opened 2021-05-26
Entangle attributes of nested components

philipboeken:

The issue of making nested components reactive has been brought up [here](https://github.com/livewire/livewire/issues/677). This PR is merely a suggestion for new functionality.

No.

No(t yet).

This pull requests (partly) implements functionality of entangling attributes of nested components, through events that are submitted in a `HydrationMiddleware`. Listening to these events could possibly be done by a trait on the `Component` class. The goal is to provide a simple api for entangling attributes between parent- and child-components, without any boilerplate.  For instance, if one considers the following components:

Parent.php:
```
{
    public $fooParent;

    public function render()
    {
        return view('parent');
    }
}
```
parent.blade.php:
```
<input type="text" wire:model="fooParent" />

<livewire:child wire:entangle:foo-child="fooParent"></livewire:child>
```

Child.php:
```
class Child extends Component
{
    public $fooChild;

    public function render()
    {
        return view('child');
    }
}
```

child.blade.php:
```
<input type="text" wire:model="fooChild" />
```

Then updating the `fooChild` attribute would be applied to `fooParent` and vice-versa.

This PR currently only facilitates entangling `fooChild -> fooParent`, but not (yet) the other way around.  Also, the implemented syntax for entangling these properties is
`<livewire:child :foo-child="$fooParent" :entangled="['fooChild' => 'fooParent']">`
to prevent from having to dive into the parser and add the `wire:entangle:` attribute.

I was mainly wondering whether something like this could eventually be added to livewire, as this functionality would eliminate much work on boilerplate input events and refreshing child components.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -30,6 +30,15 @@ abstract class Component
     protected $initialLayoutConfiguration = [];
     protected $shouldSkipRender = false;
     protected $preRenderedView;
+    
+    // TO make entanglement (child->parent) work, copy this part to your child and parent components.
+    // ----------------------------------------------------------------------------------------------
+    public $entangled = [];
+    public $entangledByChildren = [];
+    protected $listeners = ['fillEntangled', 'registerEntangled'];
+    public function fillEntangled($data) { $this->fill($data); }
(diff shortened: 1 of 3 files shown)


calebporzio:
Going to hold off on this addition. Thanks!

PR #2935, opened 2021-06-04
Resolve `ParseError ` due to excess semicolon. Affects Laravel 8.45.0 [Urgent?]

ibrasho:
The extra semicolon causes the Blade compiler in the latest Laravel 8.45.0 release to throw a `ParseError` exception.

Anyone who upgrades to the latest Laravel release will break their application, so I'm not sure if this needs an urgent patch release?

### DIFF
--- a/src/Macros/livewire-view-component.blade.php
+++ b/src/Macros/livewire-view-component.blade.php
@@ -1,5 +1,5 @@
 @component($view, $params)
     @slot($slotOrSection)
-        {!! $manager->initialDehydrate()->toInitialResponse()->effects['html']; !!}
+        {!! $manager->initialDehydrate()->toInitialResponse()->effects['html'] !!}
     @endslot
 @endcomponent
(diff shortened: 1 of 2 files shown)


joshhanley:
@ibrasho hmm that's a tough one, but I'd argue that a fix should actually be applied to Laravel as this is a breaking change from their end (albeit unintentional).

My reasoning is, someone may be upgrading their Laravel version and not their Livewire version and it would cause this error no matter what version of Livewire 2 they are on, unless they also upgrade Livewire to whatever version include this patch.

I do still agree that this could be changed in the Livewire codebase as the semicolon isn't needed.

Thoughts?

@calebporzio what do you think?

calebporzio:
Totally agree @joshhanley - thanks for this @ibrasho !

ibrasho:
I'm going to debate that semicolons are not explicitly supported in Blade since I can't find any documentation saying it's supported. 😓 There are some discussions in https://github.com/laravel/framework/pull/37603 are some new issues in Blade compiler after 8.45.0 stringable feature.

PR #2957, opened 2021-06-12
Support missing() for route model binding

### LINKED ISSUE #2951: route model binding missing function won't fire when using full-page component
### Description
Hi 
I have a problem with the route model binding in Full-Page Components.
when the data is missing its always return 404 error page and won't run the missing closure.

### Exact steps to reproduce
create a Full-Page Component Route like this :
```php
Route::get('/category/edit/{currentCategory}', \App\Http\Livewire\Admin\Category\Edit::class)
    ->missing(fn() => redirect(\route('admin.category')))
    ->name('category.edit');
```

and bind it to the livewire component:
```php
<?php

namespace App\Http\Livewire\Admin\Category;

use App\Models\Category;
use Livewire\Component;

class Edit extends Component
{
    public Category $currentCategory;

    public function render()
    {
        return view('livewire.admin.category.edit')
            ->layout('layouts.admin.admin');
    }
}
```
now if u access to a missing ID it will always show you 404 error page instead of redirecting according to that missing closure.

**Context**
- Livewire version: 2.4.4
- Laravel version: 8.46.0
- Alpine version: 2.8.1
- Browser: Chrome

nuernbergerA:
Laravel recently added `missing()` to [customize missing model behavior](https://laravel.com/docs/8.x/routing#customizing-missing-model-behavior)

This PR adds support invoked `Components`.

image

@calebporzio hope you like my test :joy: 

---
Closes #2951

### DIFF
--- a/tests/Unit/RouteMissingTest.php
+++ b/tests/Unit/RouteMissingTest.php
@@ -17,6 +17,10 @@ class RouteMissingTest extends TestCase
         if (PHP_VERSION_ID < 70400) {
             $this->markTestSkipped('Only applies to PHP 7.4 and above.');
         }
+
+        if (! method_exists(\Illuminate\Routing\Route::class, 'missing')) {
+            $this->markTestSkipped('Need Laravel >= 8');
+        }
     }
 
     /** @test */
--- a/tests/Unit/RouteMissingTest.php
+++ b/tests/Unit/RouteMissingTest.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Tests\Unit;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Illuminate\Http\Request;
+use Livewire\Component;
+use Illuminate\Support\Facades\Route;
+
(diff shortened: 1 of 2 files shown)


nuernbergerA:
Didn't noticed there is already a PR open #2957

tests/Unit/RouteMissingTest.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $class = <<<'PHP'
+namespace Tests\Unit;
+use Livewire\Component;
+class ComponentWithModel extends Component
+{
+    public FrameworkModel $framework;
+}
+PHP;
+        eval($class);
```
stancl, line 37: Why not just add the class below?

nuernbergerA: because the test will fail with PHP 7.3 and below

we test multiple PHP version with a [matrix](https://github.com/livewire/livewire/blob/master/.github/workflows/test.yml#L16)

calebporzio:
Thanks @nuernbergerA! - yeah, like you said, there's already a PR for this. Basically the same implementation - so I'll go with the one added first. Thanks!

Issue #2715, opened 2021-04-13
Issues with 'nonce' attribute


navindex:
v.2.4.2.

Nonce attribute is missing for `livewire.js`. Additionally, a space missing when 'nonce' attribute is present.
A possible solution is a the following change in `/src/LivewireManager.php` file (lines 270-271).

```
<script src="{$fullAssetPath}" data-turbo-eval="false" data-turbolinks-eval="false"></script>
<script data-turbo-eval="false" data-turbolinks-eval="false"{$nonce}>
```
could be
```
<script src="{$fullAssetPath}" data-turbo-eval="false" data-turbolinks-eval="false" {$nonce}></script>
<script data-turbo-eval="false" data-turbolinks-eval="false" {$nonce}>
```
Note the space prior {$nonce}.

austenc:
Hey @navindex, could you elaborate on your need of the `nonce` property on that tag? Is it just the lack of a space that's causing you problems, or are you trying to satisfy some sort of CSP?

This has been mentioned before -- [see this comment from Caleb](https://github.com/livewire/livewire/pull/2485#issuecomment-784355989)

navindex:
Hi @austenc,
I'm using Livewire with [laravel-csp](https://github.com/spatie/laravel-csp) by calling it like
```
@livewireScripts(['nonce' => csp_nonce()])
```
from Blade templates. Their `csp_nonce()` helper generates the nonce and takes care of adding it to the header.
Having said that, it would be nice to have the same syntax for `@livewireStyles`, too.

I've tried different ways to create a strict CSP, however, the easiest way is when I set `script` to `strict-dynamic` and use nonce everywhere.

Oha-you:
I'm having the same problem. Strict CSP is enabled on the website and Livewire inserts a raw `<script>` tag without a `nonce`: https://github.com/livewire/livewire/blob/master/src/LivewireManager.php#L275

An example of the current output is:
```html
<script src="/vendor/livewire/livewire.js?id=124876dasd2767" data-turbo-eval="false"
        data-turbolinks-eval="false"></script>
<script data-turbo-eval="false" data-turbolinks-eval="false"
        nonce="HkjAjSJHjas==">window.livewire = new Livewire({'nonce': 'HkjAjSJHjas=='});
    window.Livewire = window.livewire;
    window.livewire_app_url = '';
...
```
[...]

PR #2969, opened 2021-06-15
CSP and Nonce

alnutile:
This came up here https://github.com/livewire/livewire/issues/2715 and adding this one `{$nonce}` fixes the error you can see here 

image

Not sure why it is not include? Maybe I messed up something on my LiveWire settings?

Without this I had the one failing error out of all the other assets. After this the error was gone and the page worked.

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -297,7 +297,7 @@ HTML;
         if (! started) {
             window.livewire.start();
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks! Can you resolve the merge conflicts? Then I'll merge. Thanks again

PR #2970, opened 2021-06-15
Fixes Starting Livewire with Alpine 3 when debug=false

cubemastercodes:
- Added semicolons for JavaScript (Adding semicolons for this JavaScript is important, because it will be minified in production.)

- Removed '// Support for Alpine V3.' as it causes an issue (see below)

image

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -289,22 +289,21 @@ HTML;
         });
     };
 
-    let started = false
+    let started = false;
 
-    // Support for Alpine V3.
     window.addEventListener('alpine:initializing', () => {
         if (! started) {
             window.livewire.start();
 
-            started = true
+            started = true;
         }
-    })
+    });
 
     document.addEventListener("DOMContentLoaded", function () {
         if (! started) {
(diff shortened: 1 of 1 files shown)


joshhanley:
@cubemastercodes are you minifying it yourself or is that an issue in the built JS that is committed to Livewire's repo?

Because I've been testing this today and didn't have any issues. Would be great to know your process.

Thanks!

nuernbergerA:
@joshhanley the [minification](https://github.com/livewire/livewire/blob/be2450a4971479b89183248aaf1b56634115c3b1/src/LivewireManager.php#L179) is done with PHP if `app.debug` is `true`

joshhanley:
@nuernbergerA ah right! TIL. Thanks, good to know. I was just testing locally, so that makes sense 🙂

nuernbergerA:
Closes #2971

nuernbergerA (approved):

iksaku (approved):

calebporzio:
Ugh, thanks!

PR #2973, opened 2021-06-15
Added echo channel type encryptedPrivate

JuanRangel:
Needed. I did not create an issue.

No

No

While building a messaging app we need to bass information along echo's private encrypted channel. 


calebporzio:
Thanks!

PR #2988, opened 2021-06-19
Allow TemporaryUploadedFile to use storage defined temporaryURL function

TechhDan:
I know we need this and I hope this is needed. I created a discussion.
No
No
The TemporaryUploadedFile->temporaryUrl function does not work for other storage drivers such as Google Cloud Storage (only works with S3). This change allows the function to use the storage defined temporaryUrl.

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -67,7 +67,7 @@ class TemporaryUploadedFile extends UploadedFile
             );
         }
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

calebporzio:
Hi @TechhDan - I reverted this PR as it caused all the temporary file upload tests to fail. Can you please adjust your PR and resubmit. Thanks!

PR #3151, opened 2021-06-19
[Failing Test] Entangle breaks when using Turbo.js

tonysm:
Hey, folks.

I've been chasing this error for a while. I've created a discussion in the [Alpine repository](https://github.com/alpinejs/alpine/discussions/1534) but looks like not many people are using this combo (Alpine+Livewire+Turbo).

I couldn't find a fix for it yet but was able to write a failing Dusk test to reproduce it.

As I mentioned in the Alpine discussion issue referenced above, the issue seems to happen when making a Turbo Visit to a page that has a Livewire&Alpine component using entangle. These can be seen on any settings page of Jetstream right now (since the modals all use entangle).

I wish I was able to identify the root cause (there were some attempts in the Alpine issue if you want to check those out).

(tests were failing weirdly in the [previous branch](https://github.com/livewire/livewire/pull/3150))

### DIFF
--- a/tests/Browser/Alpine/Entangle/Test.php
+++ b/tests/Browser/Alpine/Entangle/Test.php
@@ -207,4 +207,16 @@ class Test extends TestCase
             ;
         });
     }
+
+    public function test_entangle_works_with_turbo()
+    {
+        $this->browse(function ($browser) {
+            $browser->visit(route('entangle-turbo', [], false))
+                ->assertSeeIn('@page.title', 'Testing Entangle with Turbo')
+                ->click('@turbo.link')
+                ->waitForTextIn('@page.title', 'Showing Livewire&Alpine Component after a Turbo Visit')
+                ->assertSeeIn('@output.livewire', 'false')
+                ->assertSeeIn('@output.alpine', 'false');
+        });
+    }
 }
--- a/tests/Browser/Alpine/Entangle/ToggleEntangledTurbo.php
+++ b/tests/Browser/Alpine/Entangle/ToggleEntangledTurbo.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Tests\Browser\Alpine\Entangle;
+
+use Livewire\Component as BaseComponent;
+
+class ToggleEntangledTurbo extends BaseComponent
+{
+    public $active = false;
+    public $title = 'Showing Livewire&Alpine Component after a Turbo Visit';
+
+    public function render()
+    {
+        return view()->file(__DIR__ . '/view-toggle-entangled-turbo.blade.php')
+            ->layout('components.layouts.app-for-turbo-views');
+    }
+}
--- a/tests/Browser/Alpine/Entangle/view-toggle-entangled-turbo.blade.php
+++ b/tests/Browser/Alpine/Entangle/view-toggle-entangled-turbo.blade.php
@@ -1,0 +1,11 @@
+<div>
+    <h1 dusk="page.title">{{ $title }}</h1>
+
+    <div x-data="{
+        active: @entangle('active')
+    }">
+        <div dusk="output.alpine" x-text="active"></div>
+        <div dusk="output.livewire">{{ $active ? 'true' : 'false' }}</div>
+        <button dusk="toggle" x-on:click="active = !active">Toggle Active</button>
+    </div>
+</div>
--- a/tests/Browser/TestCase.php
+++ b/tests/Browser/TestCase.php
@@ -137,6 +137,12 @@ class TestCase extends BaseTestCase
                 return app()->call(new $class);
             })->middleware(['web', 'auth', 'can:update,post']);
 
+            Route::middleware('web')->get('/entangle-turbo', function () {
+                return view('turbo', [
+                    'link' => '/livewire-dusk/' . urlencode(\Tests\Browser\Alpine\Entangle\ToggleEntangledTurbo::class),
+                ]);
+            })->name('entangle-turbo');
+
             app('session')->put('_token', 'this-is-a-hack-because-something-about-validating-the-csrf-token-is-broken');
 
             app('config')->set('view.paths', [
--- a/tests/Browser/views/components/layouts/app-for-turbo-views.blade.php
+++ b/tests/Browser/views/components/layouts/app-for-turbo-views.blade.php
@@ -1,0 +1,39 @@
+<html>
+<head>
+    <meta name="csrf-token" content="{{ csrf_token() }}">
+
+    <script src="https://unpkg.com/alpinejs@3.0.6/dist/cdn.min.js" defer></script>
+
+    <script type="module">
+        import hotwiredTurbo from 'https://cdn.skypack.dev/@hotwired/turbo';
+    </script>
+
+    @livewireStyles
+</head>
+<body>
+    {{ $slot }}
+
+    @livewireScripts
+    <script src="https://cdn.jsdelivr.net/gh/livewire/turbolinks@v0.1.x/dist/livewire-turbolinks.js" data-turbolinks-eval="false" data-turbo-eval="false"></script>
+    <script data-turbo-eval="false">
+        document.addEventListener('turbo:before-render', () => {
+            let permanents = document.querySelectorAll('[data-turbo-permanent]')
+
+            let undos = Array.from(permanents).map(el => {
+                el._x_ignore = true
+
+                return () => {
+                    delete el._x_ignore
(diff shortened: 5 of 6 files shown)


tonysm:
Not sure what is going on with the tests here. I ran the actions in my fork (see [here](https://github.com/tonysm/livewire/runs/2866394979?check_suite_focus=true)), and it fails correctly.

tonysm:
Folks, let me know if there is anything else I can do to help out here. Also, I don't think I need to say this, but I'm saying it anyway: feel free to close the PR if you think it's not worth it. It feels wrong to even open this PR with nothing but a broken test for an edge case at the intersection of 3 libs. 😅

joshhanley:
@tonysm I had a little play with it the other day, but haven't had the time to dig in properly. A failing test is good, it will help with diagnosing the issue (even if the test doesn't end up in the Livewire codebase).

What I suspect is happening, is Turbo is loading the new page, and Alpine's mutation observer is initialising any new Alpine instances, which I think is happening either before Livewire is ready, hence the `$wire` undefined error. 

I tried just calling `Livewire.start/restart/etc` in `turbo:render` but that didn't work. Then I experimented with halting the mutation observer during `turbo:before-render` and the resuming it on `turbo:render` but that also didn't help.

So I'm not sure what the best way to approach that would be. Probably will need some guidance from @calebporzio as to how to approach it.

binaryweavers:
> @tonysm I had a little play with it the other day, but haven't had the time to dig in properly. A failing test is good, it will help with diagnosing the issue (even if the test doesn't end up in the Livewire codebase).
> 
> What I suspect is happening, is Turbo is loading the new page, and Alpine's mutation observer is initialising any new Alpine instances, which I think is happening either before Livewire is ready, hence the `$wire` undefined error.
> 
> I tried just calling `Livewire.start/restart/etc` in `turbo:render` but that didn't work. Then I experimented with halting the mutation observer during `turbo:before-render` and the resuming it on `turbo:render` but that also didn't help.
> 
> So I'm not sure what the best way to approach that would be. Probably will need some guidance from @calebporzio as to how to approach it.

@joshhanley I gave it a try again keeping your feedback in mind as per `Turbo v7.0.0-rc.1` and `Alpine v3.2.4` I got it working if we make `startObservingMutations` &     `stopObservingMutations,` public on `Alpine` object.
I cloned alpine locally and tried that. in `turbo:before-render` I stop the mutation observer and then in `turbo:load` or `turbo:render` events I started it again. 
I am not get `$wire of undefined` error anymore. Now as we stopped the observer during `Turbo` replacing the body so alpine components were behaving weird and of course new components will not be initialized. For that already available `Alpine.initTree` API comes in handy where I grabbed all elements based on `[x-data]` and also `[x-init]` standalone (if using), looping through each and checking if `el._x_dataStack` is `undefined` then `Alpine.initTree(el)` 🤔
As of now I haven't faced any issues. Sharing it so that you could give it a try again and see if I am missing something. 
```js
document.addEventListener('turbo:before-render',()=>{
    Alpine.stopObservingMutations();
});
```
```js
let skipFirstRender = true;

document.addEventListener('turbo:load', () => { //turbo:render or turbo:load both works fine with same implementation

    if (skipFirstRender) {
        skipFirstRender = false
        return
    }
    let els = Array.from(document.querySelectorAll('[x-data]'));
    els.forEach((el) => {
        if (el._x_dataStack == undefined) {
            Alpine.initTree(el)
        }
    })
   Alpine.startObservingMutations();

})
```
I also tried to hack in to [this ignore mutation condition](https://github.com/alpinejs/alpine/blob/ffcf82a115b9739e68c802290e8c8fc5576dd66a/packages/alpinejs/src/mutation.js#L94) as well like so
```js
document.addEventListener('turbo:before-render',()=>{
        let els = Array.from(document.querySelectorAll('[x-data]'));
    els.forEach((el) => {
        if (el._x_ignoreMutationObserver== undefined) {
            el._x_ignoreMutationObserver = true
        }
    })
});
```
though it sets the property on elements and application started working fine without breaking but this didn't suppress `$wire of undefined` error on console. `Alpine` and `livewire` components are working fine though and communicating with each other with `$wire` as well. 
Hope this help to analyze the issue further and we'll have better support for `Turbo` etc  🙏🏼

joshhanley:
@binaryweavers oh nice good work! 🙂 If I get a chance over the weekend I'll try it out!

binaryweavers:
Leaving it for anyone else who might find it helpful. If you are using only `Turbo Drive` just to make you application to feel like an SPA and avoid full page reloads, not taking full advantage of `Turbo` echo system like `Turbo Frames` & `Turbo Streams`. I found a good alternative to that [SwupJs](https://swup.js.org/getting-started). It doesn't bother `AlpineJs` at all and has a [Livewire Plugin](https://swup.js.org/plugins/livewire-plugin) as well.  No `@entagle` issues 👍

calebporzio:
Ok @tonysm - I did some digging on this. Thanks for the good failing test.

Like you said, the problem is that Alpine 3's mutation observer is picking up the DOM mutations from Turbo and initializing itself BEFORE Livewire gets a chance to. Because Alpine is happening before Livewire, @entangle doesn't have a Livewire instance to look at and use and that's what's throwing the error.

Like I've mentioned in this discussion: https://github.com/alpinejs/alpine/pull/1995, simply "stopping" the mutation observer while Turbo is doing it's thing, works, but it goes against the way Alpine is built: to take advantage of the mutation observer and tear down listeners and such when DOM changes.

It's possible that I'm wrong in my thinking, but I've added a new mechanism to Alpine in version v3.4.1 that consists of two new public functions on the global Alpine object:

* `Alpine.`
* `Alpine.`

These functions are a better alternative than actually stopping the mutation observer. Instead, they stop handling the mutations, and queue up all mutations for later evaluation when the observer is restarted.

I've tested this and it all works as well as I can tell. Here is the new code to make Livewire, Alpine 3, and Turbo play nicely:

```html
@livewireScripts
<script src="https://cdn.jsdelivr.net/gh/livewire/turbolinks@v0.1.x/dist/livewire-turbolinks.js" data-turbolinks-eval="false" data-turbo-eval="false"></script>
<script data-turbo-eval="false">
    document.addEventListener("turbo:before-cache", () => {
        Alpine.deferMutations()
    });

    document.addEventListener("turbo:load", () => {
        Alpine.flushAndStopDeferringMutations()
    })

    document.addEventListener('turbo:before-render', () => {
        let permanents = document.querySelectorAll('[data-turbo-permanent]')
        let undos = Array.from(permanents).map(el => {
            el._x_ignore = true
            return () => {
                delete el._x_ignore
            }
        })
        document.addEventListener('turbo:render', function handler() {
            while(undos.length) undos.shift()()
            document.removeEventListener('turbo:render', handler)
        })
    })
</script>
```

The next question is: where does this code belong?

I have a few thoughts:
A) A new Alpine plugin called "Turbo" (my hesitation with this is being on the hook officially for integration - also I don't have experience with this integration beyond Turbo Drive (don't know how this would work with frames and such))
B) An "integration" in the upcoming Alpine Component Patterns
C) Add this to livewire/turbolinks (my hesitation here is that currently there is no Alpine dependency in this codebase, and it supports both Turbo AND Turbolinks - idk)
D) Leave the code as a snippet for power users to copy/paste

A is the best case scenario I think in a perfect world. B has the fewest drawbacks. C is probably a good idea, I just feel icky about it. D has already happened in the above snippet and stopping there would leave users without official guidance.

Hope this helps!

calebporzio:
Ok, I've modified the test code to use the latest versions of livewire-turbolinks and Alpine and it now passes!

Thanks @tonysm for all your work on this!

tonysm:
@calebporzio I've tested this out today. Works like a charm. Thank you so much! 🎉🎉🎉

calebporzio:
You got it!

bep:
I was pointed here by someone, so I might as well add my 50 cents:

* The above patch may work great if all you have is `[data-turbo-permanent]` components (but even then I'm not sure).
* But you cannot navigate away before you have cleaned up the _non-permanent_ Alpine components (not doing that leads to memory leaks and a non-working application/site).

See this Gist for my current best approach: https://gist.github.com/bep/a9809f0cb119e44e8ddbe37dd1e58b50

joshhanley:
@bep have you tested your gist with the latest `deferMutations` API Alpine now exposes?

If so and it works, then PR's are welcome, I think you would need to PR the change to https://github.com/livewire/turbolinks.

Hope this helps!

PR #3152, opened 2021-06-20
Monkey patch setAttribute to allow @ symbols in attribute names

calebporzio:
Browser's `setAttribute` method doesn't allow `@` symbols in the attribute name, so people using Livewire and Alpine together and conditionally adding attributes to an element like `@click="..."` are experiencing the following error:

`Uncaught (in promise) DOMException: Failed to execute 'setAttribute' on 'Element': '@click' is not a valid attribute name.`

The only way around this is to override (monkey patch) the Element prototype's setAttribute method with the following shim:

```js
function monkeyPatchDomSetAttributeToAllowAtSymbols() {
    // Because morphdom may add attributes to elements containing "@" symbols
    // like in the case of an Alpine `@click` directive, we have to patch
    // the standard Element.setAttribute method to allow this to work.
    let original = Element.prototype.setAttribute

    let hostDiv = document.createElement('div')

    Element.prototype.setAttribute = function newSetAttribute(name, value) {
        if (! name.includes('@')) original.call(this, name, value)

        hostDiv.innerHTML = `<span ${name}="${value}"></span>`

        let attr = hostDiv.firstElementChild.getAttributeNode('@click')

        hostDiv.firstElementChild.removeAttributeNode(attr)

        this.setAttributeNode(attr)
    }
}
```

This is not referenced from anywhere, I wrote this myself, so hopefully it is robust and won't cause any unintended interactions with other JavaScript code on a given site.

I perf tested it and it seems pretty good. 

If this causes unintended side effects in any way, we could add this code directly to morphdom rather than monkey patch, but I'd rather start here.


orware:
Thanks Caleb for looking into this one earlier this evening!

I was hoping this would do the trick so I gave it a quick test here on my end (switched my `x-on:click` adjustment back to `@click` after updating with composer to `dev-master` as you mentioned).

After doing so and reloading the page on my end I received the following (different) error than before:

```
Uncaught (in promise) TypeError: Failed to execute 'removeAttributeNode' on 'Element': parameter 1 is not of type 'Attr'.
```

And Chrome mentions the issue is on line 147 of index.js which is this line:

```
    hostDiv.firstElementChild.removeAttributeNode(attr)
```

I then went back to revert to using `x-on:click` again and still received the same error so it seems that the monkey patch may not be good to go just yet so we might need a little more Flex Tape on this one still ;-).

Also, I don't want to muddy this issue at all (I know how important it is to focus on one problem at a time as much as possible), but I did want to bring some attention to the other problem I came across in that same discussion thread https://github.com/livewire/livewire/discussions/2989#discussioncomment-893357 (for the `Uncaught TypeError: Cannot read property 'startsWith' of undefined` error). since I don't think that one has an equivalent just yet (and if I switch to Alpine v2 the issue goes away, so it seems specific to Alpine v3 and when running a search on it in the Discussions I didn't quite find something that sounded overly similar: https://github.com/alpinejs/alpine/discussions?discussions_q=undefined although this one seems potentially similar: https://github.com/alpinejs/alpine/discussions/1568 I also started to ask a question about it in the Alpine Discord). I'm sure you're dealing with a lot of these currently after the v3 update so hopefully it hasn't been too overwhelming and thank you for taking the time to look into these (and other) issues!

mixdd:
This breaks when upgrading from 2.5.1 to 2.5.2. for other code using setAttribute to set the value of an attribute to something like 
`"{\"data\":\"name=John\"}"`

I'm not sure if the intent was to only execute the code for elements where `name.includes('@')` (it kind of looks like it). If so, something like the following would fix the issue:

```
    Element.prototype.setAttribute = function newSetAttribute(name, value) {
        if (! name.includes('@')) {
            original.call(this, name, value)
        }
        else {
            hostDiv.innerHTML = '<span ${name}="${value}"></span>'

            let attr = hostDiv.firstElementChild.getAttributeNode('@click')

            hostDiv.firstElementChild.removeAttributeNode(attr)

            this.setAttributeNode(attr)
        }
    }
```

joshhanley:
@mixdd can you share a code sample of what you are trying that isn' working?

mixdd:
Sure, in my case this is happening when using the fetchify.com address completion library (but would happen in all instances where you're using `JSON.stringify` to set the value on an element).

The code tries to use `setAttribute` which then uses the MonkeyPatch version, resulting in the attributes getting mangled.

Code from fetchify (cc_c2a.min.js)
```
    clickToAddress.prototype.tools.ccData = function(target, attr, value) {
    if (typeof target === "undefined" || typeof attr === "undefined") {
        return
    }
    if (typeof value !== "undefined") {
        **target.setAttribute("data-" + attr, JSON.stringify({
            data: value
        }));**
        return true
    } else {
        return JSON.parse(target.getAttribute("data-" + attr)).data
    }
}
```
Before version 2.5.2 the element being set would look as follows after setting the attribute's value:

```
<li data-id="{&quot;data&quot;:&quot;postal_code_partial=04930|town=Dexter|street=Number 10 Rd&quot;}" data-count="{&quot;data&quot;:&quot;23&quot;}" title="04930, Number 10 Rd, Dexter" class=" cc-filter">
  <div><span>04930</span><span class="light">Number 10 Rd, Dexter</span><span class="light">(23 more)</span></div>
</li>
```
and after:

```
<li data-id="{" data-count="{" title="00682, Mayaguez" class=" cc-filter">
  <div><span>00682</span><span class="light">Mayaguez</span><span class="light">(175 more)</span></div>
</li>
```

joshhanley:
@mixdd a fix has been tagged for this, please upgrade to Livewire 2.5.5 and let us know if you are still having any issues 🙂

PR #3175, opened 2021-06-23
Enable other traits to build on pagination trait

daniel-zahariev:
I need this in order to build on top of the pagination trait. Haven't created issue/discussion.

Only one file is changed

No tests, sorry

It's useful to extend the pagination trait functionality and build on top but the `paginationView` and `paginationSimpleView` functions in `WithPagination` are written in a way that doesn't allow extending. So they're now updated to follow a similar approach to the one used in `getQueryString`.

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -25,17 +25,25 @@ trait WithPagination
             return (int) $this->page;
         });
 
-        Paginator::defaultView($this->paginationView());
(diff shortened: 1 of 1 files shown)


calebporzio:
@daniel-zahariev - this makes sense, however if someone was overriding these methods in there component class, and calling "parent::paginationView()" - this would be a breaking change.

My recommendation would be to pull this trait into your project and manipulate it as you need. Thanks!

PR #3198, opened 2021-06-27
Google Cloud Storage driver compatibility for temporary file uploads

TechhDan:
I know we need this and I hope this is needed. I created a discussion.
No
This PR should pass a tests
The TemporaryUploadedFile->temporaryUrl function does not work for other storage drivers such as Google Cloud Storage (only works with S3). This change allows the function to use the storage defined temporaryUrl.

### DIFF
--- a/src/FileUploadConfiguration.php
+++ b/src/FileUploadConfiguration.php
@@ -45,13 +45,6 @@ class FileUploadConfiguration
         return config('filesystems.disks.'.strtolower($diskBeforeTestFake).'.driver') === 's3';
     }
 
-    public static function isUsingGCS()
-    {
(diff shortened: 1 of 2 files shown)


TechhDan:
This is a modification of a previous pull request https://github.com/livewire/livewire/pull/2988. This one should pass all tests. Thanks!

calebporzio:
Rather than including driver specific code inside core - is there a way you can currently allow this via extension? Or a way that Livewire can make that easier for individuals?

TechhDan:
This worked out for me what do you think?

calebporzio:
Thanks!

PR #3207, opened 2021-06-29
Add test flag to all commands

ariaieboy:
No
No
Yes
added --test flag for move copy and delete command
since we have this flag with make command it is nice to have the same flag for move copy and delete command so we can manage livewire tests like how we manage the components using the cli 


calebporzio:
This makes sense - THANKS!

PR #3214, opened 2021-07-01
Fix: non-typed initialized properties were not begin hydrated when setting value to null

jonneroelofs:
Initialized properties were not begin hydrated when setting value to null.
Therefore livewire considered the value dirty and reverted the changed value on the next hydration.
The provided test fails without the change to the HydratePublicProperties test.

This fix actually requires php7.4 or higher I think because I used the ReflectionProperty .

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -49,12 +49,16 @@ class HydratePublicProperties implements HydrationMiddleware
 
                 // If the value is null and the property is typed, don't set it, because all values start off as null and this
                 // will prevent Typed properties from wining about being set to null.
-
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @jonneroelofs!

joshhanley:
@jonneroelofs there seems to be an unintentional breaking change caused by this PR, any change you could check out this discussion and see if there is a way we can work around it? https://github.com/livewire/livewire/discussions/3305

jonneroelofs:
@joshhanley I see the persons who started the discussion already has a PR fixing it. If there are more/other issues caused by my PR I am of course willing to help.

joshhanley:
@jonneroelofs yeah just saw that! No worries, thanks for checking it out 🙂

PR #3216, opened 2021-07-01
[Feature] 'upload:cancel' Event

MoamenEltouny:
It's an simple event to handling XMLHttpRequest of uploading of some field.
it works on cancelling the Request.

I hope it can be helpful or even be the base of the idea.

Thank you

### DIFF
--- a/js/component/UploadManager.js
+++ b/js/component/UploadManager.js
@@ -128,6 +128,7 @@ class UploadManager {
             this.component.call('uploadErrored', name, errors, this.uploadBag.first(name).multiple)
         })
 
+        this.uploadBag.first(name).request = request
         request.send(formData)
     }
 
--- a/js/component/UploadManager.js
+++ b/js/component/UploadManager.js
@@ -28,6 +28,7 @@ class UploadManager {
         this.component.on('upload:finished', (name, tmpFilenames) => this.markUploadFinished(name, tmpFilenames))
         this.component.on('upload:errored', (name) => this.markUploadErrored(name))
         this.component.on('upload:removed', (name, tmpFilename) => this.removeBag.shift(name).finishCallback(tmpFilename))
+        this.component.on('upload:cancel', (name) => this.cancelUpload(name)
     }
 
     upload(name, file, finishCallback, errorCallback, progressCallback) {
@@ -157,6 +158,15 @@ class UploadManager {
         this.uploadBag.shift(name).errorCallback()
 
         if (this.uploadBag.get(name).length > 0) this.startUpload(name, this.uploadBag.last(name))
+    }
+    
+    cancelUpload(name) {
+        if (this.uploadBag.first(name) != undefined) {
+            this.uploadBag.bag[name][0].request.abort();
+            return this.uploadBag.shift(name);
+        }
+        
+        return;
     }
 }
 


calebporzio:
Thanks @RaggiTech - The PR looks ok, just would like more information on the use case (and maybe some example code) - as well as a Dusk test if that's possible. Thank you!

RaggiTech:
Thank you @calebporzio for your reply 

### Just i found that Livewire uploading methodology depends on uploading files directly to the server as a tmp files but there is no Cancel option for this uploading process.

**In case i chose a wrong file  i need to cancel this uploading process because there's no need to full upload then delete it from tmp directory.**

**Example** :  photos input
```html
<input type="file" wire:model="photos" multiple>
<button type="button" wire:click="$emit('upload:cancel', 'photos')">Cancel</button>
```

**The idea still need more developing and study .. i hope you can help me in that.**

_I think Livewire now allow uploading multiple files in a single request but i think it gonna be great if that be optional .. i mean we can allow it uploading multiple files all together in a single request or split them in multiple requests to can cancel a specific file and keep uploading the remain files._ 

**Thank you for your efficient efforts.**
    Raggi

calebporzio:
I can see the use case for something like this - I wonder what the best API is though? And before exploring this further I'd like to see more evidence of demand for this. Thanks

joshhanley:
@RaggiTech do you know of any issues or discussions where others have discussed needing this feature? If so can you link them here? Thanks!

joshhanley:
Have just come across one discussion where a couple of people are requesting this #3607

I think this is a valid use case

> I would also like to know this, my application displays a modal for uploading files and I would like to cancel any ongoing uploads if the user decides to close the modal.

piersgill312:
Topic starter from #3607 here, I would very much like this implemented.

My Use case in this siutation for a project I'm using Filepond (Front facing File Upload JS library) + Livewire to upload files in a file uploading system.

Filepond provides a cancel upload output function when a user decides to cancel the current upload from the UI but Livewire has no current cancel upload/abort receive function. Like a mistakenly inputted huge file etc.

My two cents here but right now I would agree with the PR in using request.abort() API from looking at these documents.

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort

It's currently used in the Filepond docs at the moment using pure JS. 
https://pqina.nl/filepond/docs/api/server/#process-1

theChaosCoder:
My use case: Uploading (3-8mb) images from a smartphone. If the mobile connection is currently not that fast and you pick the wrong image you have to wait several minutes for the upload to complete.

So there is definitely demand for this.

smares:
Would really like to see this merged ASAP. 👍

smares:
Any ETA for this before I fork and use my own repo?

js/component/UploadManager.js
```diff
@@ -28,6 +28,7 @@ class UploadManager {
         this.component.on('upload:finished', (name, tmpFilenames) => this.markUploadFinished(name, tmpFilenames))
         this.component.on('upload:errored', (name) => this.markUploadErrored(name))
         this.component.on('upload:removed', (name, tmpFilename) => this.removeBag.shift(name).finishCallback(tmpFilename))
+        this.component.on('upload:cancel', (name) => this.cancelUpload(name)
```
smares, line 31: There is a syntax error on this line, a closing bracket is missing

calebporzio:
Hey everyone, going to close this one for now. Not because it's not a good idea and a valid use-case, but because of the following:
* The PR doesn't include any tests
* This PR doesn't include code snippets displaying how a user of Livewire would use this event to achieve a specific result

Feel free to re-submit this PR when those things have been addressed, and we can continue the discussion there.

Thanks again!

PR #3224, opened 2021-07-02
Update README.md

jonneroelofs:
updated the readme to reflect that the current livewire philosophy is to open discussions not issues


calebporzio:
Thank you!

PR #3263, opened 2021-07-08
Fix Livewire not running in production

joshhanley:
Livewire v2.5.2 reintroduced errors when trying to run Livewire in production (aka `APP_DEBUG=false`), due to the missing semicolons in `LivewireManager`. For some reason the fix applied in this commit a4ffb135693e7982e5b982ca203f5dc7a7ae1126 got reverted.

I've added a test which simulates a production env by setting `APP_DEBUG=false` in the mount method of the component in this new test.

I've submitted this PR before the fix, to ensure the test fails in CI. Then I will push the fix up.

Hopefully this will mean tests catch these issues incase they happen in the future.

Hope this helps!


nuernbergerA (approved):

nuernbergerA:
@joshhanley thanks for the PR and test to ensure we don't hit this error anymore

Restingo (approved):

calebporzio:
Thanks @joshhanley, and thanks for the tests - finally this won't bite us again.

PR #3264, opened 2021-07-08
add multiple pagination support on single component

ariaieboy:
yes its discussed here https://github.com/livewire/livewire/discussions/3145 and here https://github.com/livewire/livewire/pull/3215#issuecomment-875787631
No
No

ok so i add a PR for the Cursor Pagination Support here https://github.com/livewire/livewire/pull/3215 and @danharrin asked about multiple pagination in single component support so i make this Pr to add multiple pagination support.
its only work on cursor pagination but if @calebporzio approve the design and workflow i can add support for regular pagination too.

so lets see how it's work :
image
as you can see in the picture i add a `protected $paginations` to the trait that contains the name of the querystring that we want to use and i will create the query strings based on it.

now if anyone want to add multiple pagination in 1 component they can use the `setCursorName()` or pass the key name to the `cursorPaginate()` function like this:

image

ok so we have `allCursor` here as key now we need to add a public property and add the key to `paginations` property :
```php
    protected $paginations = [
        'cursor',
        'allCursor'
    ];
    public $allCursor = '';
```
now we have multiple pagination in our single component with `cursor` and `allCursor` queryString.
the good thing here is that we can chage default `cursor` key for single pagination too. 

### DIFF
--- a/src/WithCursorPagination.php
+++ b/src/WithCursorPagination.php
@@ -16,7 +16,7 @@ trait WithCursorPagination
             ? $this->queryString()
             : $this->queryString;
 
-        return array_merge(['cursor' => ['except' => null]], $queryString);
+        return array_merge(['cursor' => ['except' => '']], $queryString);
     }
 
     public function initializeWithCursorPagination()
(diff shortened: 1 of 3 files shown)


stevebauman:
Hi @ariaieboy,

Can you please insert markdown'ed code instead of photos for those examples?

calebporzio:
Thank you @ariaieboy - however, this has been added by @joshhanley  in #3684

PR #3289, opened 2021-07-10
Detect multiple root elements

joshhanley:
This is a draft proof of concept for detecting whether there are multiple root elements in a component.

It will output the below error in the console if multiple root elements are detected
image

How it works is it adds a html comment before and after the components rendered html. 
The comments contain `wire-start:<component-id>` and `wire-end:<component-id>` respectively.

**Update:** Not sure if the start comment and walking back up the tree is actually required, as Livewire gets embedded on the first element anyway. What do you think?

## Things to note:

1. There was an issue when I added the comments that contains the blade echo of the ID into the views, threw the below error, hence the changes in the `src/Component.php` to fix that error.
image

2. This didn't fix the issue of a html comment being the first element in a component, still causes the below output onto the screen and Livewire not running. 
image

**I have updated the regex to be stricter.**

It currently has an issue due to the `*` at the end being check for "zero or more" where I believe it should be "1 or more" of any alphanumeric character or dash `-`. Which is what is currently allowing `<!--` to be matched.

There is also an issue though of, if a comment has a html element inside of it, it will also not work. So this comment would be matched
```html
<!-- Some comment with a <div>inner element</div> -->
```
I've added a check at the front of the regex to ensure the match element is at the start of a line (with optional spaces before it).

This doesn't cover the use case where it is a multiline comment with html inside of it. That is something we can explore if required, but this currently covers the use case where a user puts a descriptive comment above their html like `<!-- Main Section -->` or comments out a line of code.

I wasn't sure if this should go into this PR or into a separate one, as it is related.

## Things to do

- [x] Add tests that cover each change
- [ ] Work out whether there will be any unintentional side-effects


calebporzio:
Fantastic. Thanks @joshhanley!

PR #3290, opened 2021-07-11
Testing Components can generate false positives from wire:id

mertasan:
1️⃣  Is this something that is wanted/needed? Did you create an issue / discussion about it first? : No

2️⃣  Does it contain multiple, unrelated changes? Please separate the PRs out. : No

3️⃣  Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have) : Yes

---

Details on this change are available here #62

Some tests also pass with `wire:id`. Please watch the video.

https://user-images.githubusercontent.com/13007665/125180269-d668c580-e200-11eb-9cf9-2997c6c1b411.mov

Additionally, the current pattern was not removing the trailing space.

**before:**
```blade
<div wire:id="123Abc" wire:initial-data="{...}">...</div>

<div wire:id="123Abc" >...</div>
```
**after:**
```blade
<div wire:id="123Abc" wire:initial-data="{...}">...</div>

<div>...<div>
```

## Demo/Test: 
https://regex101.com/r/UhjREC/3

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -119,7 +119,7 @@ trait MakesAssertions
 
     protected function stripOutInitialData($subject)
     {
-        return preg_replace('(wire:initial-data=\".+}")', '', $subject);
+        return preg_replace('/((?:[\n\s+]+)?wire:initial-data=\".+}"\n?|(?:[\n\s+]+)?wire:id=\"[^"]*"\n?)/m', '', $subject);
     }
 
     public function assertEmitted($value, ...$params)
--- a/tests/Unit/LivewireTestingTest.php
+++ b/tests/Unit/LivewireTestingTest.php
@@ -96,9 +96,12 @@ class LivewireTestingTest extends TestCase
     }
 
     /** @test */
-    public function assert_see_doesnt_include_json_encoded_data_put_in_wire_data_attribute()
+    public function assert_see_doesnt_include_wire_id_and_wire_data_attribute()
     {
-        // See for more info: https://github.com/calebporzio/livewire/issues/62
+        /*
+        * See for more info: https://github.com/calebporzio/livewire/issues/62
+        * Regex test: https://regex101.com/r/UhjREC/2/
+        */
         app(LivewireManager::class)
             ->test(HasMountArgumentsButDoesntPassThemToBladeView::class, ['name' => 'shouldnt see me'])
             ->assertDontSee('shouldnt see me');


calebporzio:
Thanks!

mertasan:
@calebporzio

This way I think it would be more appropriate to completely remove the wrapper div. How idea?

```php
protected function stripOutInitialData($subject)
{
    $content = preg_replace('/((?:[\n\s+]+)?wire:initial-data=\".+}"\n?|(?:[\n\s+]+)?wire:id=\"[^"]*"\n?)/m', '', $subject);

    if (Str::startsWith($content, '<div>')) {
        $content = preg_replace('/^<div>\n?([\S\s]*)\n?<\/div>\n?$/m', '$1', $content);
    }

    return $content;
}
```

### Regex: https://regex101.com/r/XKeZYf

If the wrapper has no other attributes:

```blade
<!-- before -->
<div wire:id="foo" wire:initial-data="bar">
content
</div>

<!-- after -->
content
```

If the wrapper has another attribute:

```blade
<!-- before -->
<div class="hello" wire:id="foo" wire:initial-data="bar">
content
</div>

<!-- after -->
<div class="hello">
content
</div>
```

Additionally,

https://github.com/livewire/livewire/blob/b095ad6c5fa9be326a0a36854af68a322d62d7b1/src/Testing/TestableLivewire.php#L97

Something like this could be added below this line.

```php
$this->lastRenderedCleanDom = $this->stripOutInitialData($this->lastRenderedDom);
```

The benefit of this is:

For example, when testing with Pest, clean results can be obtained with such use.

```php
// PestPHP

expect($component->lastRenderedCleanDom)
      ->toStartWith('foo')
      ->toEndWith('bar');
```

The `assertSee` function produces results. 
But we can use the filtered data multiple times with `$this->lastRenderedCleanDom`.

I sampled this with Pest. But actually I think this will be useful for everyone.

I would like to get your opinion on this matter before creating a PR on this.

joshhanley:
@mertasan personally, I don't think thats a good idea. The wrapper element is considered part of the html of the component and as such the structure and layout.

Imagine a scenario where that element is a `<tr>` for example, you wouldn't want to remove that. 

I don't see any benefit to removing it. Maybe you can explain better your use case? I don't think the sample test you have provided above is something we would actually test for.

But I could imagine it being nice to access the rendered html though for testing, so maybe the `->lastRenderedDom` could be documented?

Thoughts?

mertasan:
@joshhanley There are several ways to test data. I know. But in most cases it is necessary to see the last data displayed in the generated dom.

According to Livewire requirements, the component must have a wrapper div.

https://github.com/livewire/livewire/blob/b095ad6c5fa9be326a0a36854af68a322d62d7b1/tests/Unit/views/fillable-view.blade.php#L1-L5

But sometimes it may be necessary not to see any html elements other than the data we need to test.

So we can get this data and transform it as we test it.

For example:
https://pestphp.com/docs/expectations#expect-json

The following test fails because of `<div>` elements.

```blade
<div>
{!! json_encode($column['value']) !!}
</div>
```

```php
test('example test', function() {

  $component = livewire(Component::class);

  expect($component->lastRenderedDom)
        ->json()
        ->toHaveCount(2)
        ->name->toBe('foo')
        ->credit->toBeFloat();
});
```

But this might be possible without `<div>`.

If wrapper div are needed, `->lastRenderedDom` is used. But if a clean content is required `->lastRenderedCleanDom` can be used.

```php
test('example test', function() {

  $component = livewire(Component::class);

  expect($component->lastRenderedCleanDom)
        ->json()
        ->toHaveCount(2)
        ->name->toBe('foo')
        ->credit->toBeFloat();
});
```

joshhanley:
@mertasan so just to clarify, although the docs say wrapper `<div>` it doesn't actually matter what the element is, as long as there is a single root element that wraps everything within the component. So support would be needed for all the different wrapper element types.

And don't get me wrong, I understand what you are saying, and what your testing in your example, and if that were how my Livewire components were constructed, then I'd definitely see the benefit.

My question is, do you actually have Livewire components that only return a string such as `{!! json_encode($column['value']) !!}` in your apps or is this a theoretical idea at this stage?

I ask, because in my experience, using Livewire components down to that granular of a level can cause issues with cross component communication and keep state in sync. In my opinion it's better to use blade components for splitting things up like this.

Please understand, I'm only pushing back to make sure it's use-cases and need are solid, as I know Caleb is hesitant to adding things into the core until there is genuine need with real world usages. By clarifying all this will help supporting your PR and increases your chances of it getting in 🙂

mertasan:
@joshhanley I don't get you wrong. But my English is not very good. So I may not be able to express exactly what I mean. I think this suggestions will be useful for everyone. If I'm wrong, I'd like to know. And thank you very much for your answers.

I'm developing a data tables package. The data of the sub-columns added to the columns is stored as an array in the data of the columns. I'm testing this data. The code in the example is not theoretical.

The content of one of my sample files is as follows.

```php
// SubAvatarValuesTest.php

test('values → from data', function () {
    $values = [
        'name'     => [
            [
                'left' => [
                    'value'    => 'avatar.png',
                    'type'     => 'avatar',
                    'isDB'     => false,
                    'typeInfo' => [
                        'imageSize' => 'md',
                    ],
                ],
            ],
            [
                'left' => [
                    'value'    => 'avatar.png',
                    'type'     => 'avatar',
                    'isDB'     => false,
                    'typeInfo' => [
                        'imageSize' => 'md',
                    ],
                ],
            ],
        ],
        'gender'   => [
            [
                'left' => [
                    'value'    => 'foo.png',
                    'type'     => 'avatar',
                    'isDB'     => true,
                    'typeInfo' => [
                        'imageSize' => 'sm',
                    ],
                ],
            ],
            [
                'left' => [
                    'value'    => 'bar.png',
                    'type'     => 'avatar',
                    'isDB'     => true,
                    'typeInfo' => [
                        'imageSize' => 'sm',
                    ],
                ],
            ],
        ],
        'username' => [
            [
                'left' => [
                    'value'    => 'foo-avatar.png',
                    'type'     => 'avatar',
                    'isDB'     => true,
                    'typeInfo' => [
                        'imageSize' => 'xl',
                    ],
                ],
            ],
            [
                'left' => [
                    'value'    => 'bar-avatar.png',
                    'type'     => 'avatar',
                    'isDB'     => true,
                    'typeInfo' => [
                        'imageSize' => 'xl',
                    ],
                ],
            ],
        ],
    ];
    $newArray = [];
    foreach ($values as $key => $data) {
        foreach ($data as $i => $iValue) {
            data_set($newArray, $i . '.' . $key, $iValue);
        }
    }

    for ($i = 0; $i < 2; $i++) {
        $columns = tdt()->result($i)->toCollection(true)->pluck('sub', 'name');
        foreach ($columns as $name => $sub) {
            $this->assertSame(
                $sub,
                $newArray[$i][$name]
            );
        }
    }
});

test('values → from view', function () {
    $expected = <<< 'HTML'
{"value":"avatar.png","type":"avatar","isDB":false,"typeInfo":{"imageSize":"md"}}
{"value":"foo.png","type":"avatar","isDB":true,"typeInfo":{"imageSize":"sm"}}
{"value":"foo-avatar.png","type":"avatar","isDB":true,"typeInfo":{"imageSize":"xl"}}
{"value":"avatar.png","type":"avatar","isDB":false,"typeInfo":{"imageSize":"md"}}
{"value":"bar.png","type":"avatar","isDB":true,"typeInfo":{"imageSize":"sm"}}
{"value":"bar-avatar.png","type":"avatar","isDB":true,"typeInfo":{"imageSize":"xl"}}
HTML;

    expect(
        seeComponent() // <<< See screenshot to see what this does.
    )->toEqual($expected);
});

beforeEach(function () {
    migrate();

    factoryCreate('user', [
        'name'   => 'foo',
        'gender' => 'one',
        'avatar' => 'foo.png',
    ]);

    factoryCreate('user', [
        'name'   => 'bar',
        'gender' => 'two',
        'avatar' => 'bar.png',
    ]);

    factoryCreate('profile', [
        'user_id'     => 1,
        'cover_photo' => 'foo-avatar.png',
    ]);

    factoryCreate('profile', [
        'user_id'     => 2,
        'cover_photo' => 'bar-avatar.png',
    ]);

    tdtCreateComponent(SubAvatarValuesTest::class); // This is same as Livewire::test(...)
});

beforeAll(function () {
    class SubAvatarValuesTest extends Yediyuz\TallDataTables\Tests\Support\TableComponent
    {
        public function columns(): array
        {
            return [
                // tdtColumn() is a helper function. Same as Column::avatar('avatar')->size('md)
                tdtColumn()::string('name')
                           ->avatar('avatar.png'),

                tdtColumn()::string('gender')
                           ->avatar('{%avatar%}', size: 'sm'),

                tdtColumn()::string('username')
                           ->avatar('{%profile.cover_photo%}', function (Yediyuz\TallDataTables\Columns\Sub\Avatar $avatar) {

                               $avatar->size('xl');

                               return $avatar;
                           }),

            ];
        }

        public function model()
        {
            return makeModel('user');
        }

        public function render()
        {
            return view('row-sub-data');
        }
    }
});
```

```blade
<!-- tests/resources/views/row-sub-data.blade.php -->
<div>
@foreach ($this->rows as $index => $row)
@foreach($columns as $column)
@include('row-sub-data-picker', ['column' => $this->getColumnData($row, $column)])
@endforeach
@endforeach
</div>
```

```blade
<!--tests/resources/views/row-sub-data-picker.blade.php -->
@foreach ($column['sub'] as $position => $subData)
@include('sub-columns.'.$subData['type'], ['subData' => $subData])
@endforeach
```

```blade
<!-- tests/resources/views/sub-columns/avatar.blade.php -->
{!! nl2br(json_encode($subData, JSON_THROW_ON_ERROR | JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP)) !!}
```

TestCase.php

screenshot 66

Also, the wire data is still exists in `->lastRenderedDom`. This causes the tests to pass for the same reasons as merging this PR.

Example:

```php
expect($component->lastRenderedDom)->toContain('1');
```

In summary, the wrapper div is necessary for Livewire. But this is necessary for requests etc.

If we don't want to see it in the test results, there should be an alternative. `->lastRenderedCleanDom`

Thank you very much again.

joshhanley:
@mertasan ah right, thanks for the detailed explanation, that makes sense.

I'm not sure if it would be accepted or not, but if you have already written the code, it can't hurt to submit the PR 🙂

If it doesn't get in though, `src\Testing\TestableLivewire.php` is macroable, so you could create your own package that adds that functionality to the testable livewire class, so you and others can use it nicely!

mertasan:
@joshhanley thank you very much for your support and ideas! 🎉

joshhanley:
@mertasan no worries 🙂

PR #3320, opened 2021-07-13
Avoid reflection exception on virtual properties

Dimonka2:
This is a fix for the latest changes on hydration that making impossible to use virtual properties in Livewire components. Discussion is here: https://github.com/livewire/livewire/discussions/3305

Unfortunately this PR does not have tests, but it fixes the issue..

The virtual properties are work around for missing aliases in query string. Aliases are needed to resolve pagination issue, search, filter etc. Just to have an ability to use the same code base for components, but change only query strings aliases. So virtual properties are kind of complicated work around to have aliases. 

Thank you!


calebporzio:
Thanks @Dimonka2 - do you feel able to add a test for this case so we don't re-introduce this bug?

Dimonka2:
Hi @calebporzio, I added a test that checks read/write of virtual property and it passes on latest version.

calebporzio:
Thanks!

PR #3347, opened 2021-07-16
Fix eloquent on event emitting when the model uses other route key name distinct to id

luischavez:
I'm working a lot with Livewire on a new project, when trying to improve the SEO of my app using slugs on eloquent models I found a bug when I use that models in event emitting, example:

```php
class Book extends Model
{
    public function getRouteKeyName(): string
    {
        return 'slug';
    }
}
```

```php
$this->emit('bookTracked', $book, $track);
```

```php
class Header extends Component
{
    protected $listeners = [
        'bookTracked',
    ];

    public function bookTracked(Book $book, bool $tracked): void
    {
        ...
    }
```

After emitting the event an error is throw indicating that the model can't be found, after some research I detect the **ImplicitlyBoundMethod@getImplicitBinding** method is passing the model as array in **resolveRouteBinding($value)**, internally the default implementation of resolveRouteBinding do: 
```php
   /**
     * Retrieve the model for a bound value.
     *
     * @param  mixed  $value
     * @param  string|null  $field
     * @return \Illuminate\Database\Eloquent\Model|null
     */
    public function resolveRouteBinding($value, $field = null)
    {
        return $this->where($field ?? $this->getRouteKeyName(), $value)->first();
    }
```
so I think internally the where method is fetching by default the id from the model array, and this causes a mismatch when the model use other route key name distinct to the id.

### DIFF
--- a/src/ImplicitlyBoundMethod.php
+++ b/src/ImplicitlyBoundMethod.php
@@ -91,7 +91,8 @@ class ImplicitlyBoundMethod extends BoundMethod
 
     protected static function getImplicitBinding($container, $className, $value)
     {
-        $model = $container->make($className)->resolveRouteBinding($value);
+        $class = $container->make($className);
+        $model = $class->resolveRouteBinding($value[$class->getRouteKeyName()]);
 
         if (! $model) {
             throw (new ModelNotFoundException)->setModel($className, [$value]);
--- a/src/ImplicitlyBoundMethod.php
+++ b/src/ImplicitlyBoundMethod.php
@@ -4,6 +4,7 @@ namespace Livewire;
 
 use Illuminate\Container\BoundMethod;
 use Illuminate\Contracts\Routing\UrlRoutable as ImplicitlyBindable;
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use ReflectionClass;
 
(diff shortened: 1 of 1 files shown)


luischavez:
issue #3349

nuernbergerA:
@luischavez if you could a a test it would be perfect.

calebporzio:
@luischavez - thanks for this, like @nuernbergerA  said, can you add a test for this case?

calebporzio:
Ok, after looking into this, what you are after makes sense from a usability perspective.

Currently, you can achieve what you want by explicitly passing a slug instead of the entire model
```
$this->emit('bookTracked', $book->slug, $track);
```

I see how your implementation is more intuitive, allowing people to pass entire models as params and have them injected as params into the corresponding method. 

Does anyone have any opinions on whether this is a good idea or not? My hesitancy is that users may falsely think that they don't have to authorize ownership of the injected model because they explicitly passed the model into `emit()`. That scares me a tiny bit.

Thanks for the PR - I'm pretty hesitant on it though - waiting to hear other opinions

luischavez:
> Ok, after looking into this, what you are after makes sense from a usability perspective.
> 
> Currently, you can achieve what you want by explicitly passing a slug instead of the entire model
> 
> ```
> $this->emit('bookTracked', $book->slug, $track);
> ```
> 
> I see how your implementation is more intuitive, allowing people to pass entire models as params and have them injected as params into the corresponding method.
> 
> Does anyone have any opinions on whether this is a good idea or not? My hesitancy is that users may falsely think that they don't have to authorize ownership of the injected model because they explicitly passed the model into `emit()`. That scares me a tiny bit.
> 
> Thanks for the PR - I'm pretty hesitant on it though - waiting to hear other opinions

Sorry for the late response I'm working on a library based on Livewire with a set of the most used utilities and components for  a common projects and I'm spending a lot of time on it, btw I agree with you with the security concern, after looking in the network traffic I notice that when you pass an entire model on the emit method all the properties on the model are visible on the response, so I decide to pass only the id and not the entire model for security reasons.

calebporzio:
Yeah, sounds good. If you like, you can PR a bit to the docs repo about this for others.

Thanks for pitching in, but going to close this one for now. Thanks again!

PR #3362, opened 2021-07-18
Support Wireable Objects (DTOs)

danharrin:
This PR finally introduces support for custom DTO public properties in Livewire! 🚀 

> Additional credit to @ryangjchandler who I hacked on this with yesterday,

Any object may be `Wireable`, which allows it to be persisted between Livewire requests. Some use case ideas for this feature...
- Custom DTOs, commonly used in PHP development in general.
- Persisting unsaved Eloquent models between requests (implementing the interface on the model, and then serializing / unserializing from the `$attributes` property).
- Any other class you might want to persist and use in Livewire.

This feature even supports `wire:modelling` to a property in the custom object, using model-style dot syntax!

### How to use it

1. Implement the `Livewire\Wireable` interface on the custom class.
2. Set up `toLivewire()` and `fromLivewire()` methods on the class, which serializes it to and from an array.
3. Add the type of the custom class to the component's public property, like you would for a model.

### Example

```php
class Component extends \Livewire\Component
{
    public DTO $dto;

    public function mount()
    {
        $this->dto = new DTO('Message');
    }
}
```

```php
use Livewire\Wireable;

class DTO implements Wireable
{
    public $message;

    public function __construct($message)
    {
        $this->message = $message;
    }

    public function toLivewire()
    {
        return [
            'message' => $this->message,
        ];
    }

    public static function fromLivewire($value): self
    {
        return new self($value['message']);
    }
}
```

You can now use the `$dto` in your Livewire view, and even bind to it with `wire:model`...

```html
<div>
    <span>{{ $dto->message }}</span>

    <input wire:model="dto.message" />
</div>
```

This pattern is incredibly flexible, and (in theory) you can now use *any* PHP class in *any* state with Livewire!

### Considerations

Like other typed-property powered features in Livewire, this only supports PHP 7.4+.

### DIFF
--- a/src/Wireable.php
+++ b/src/Wireable.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Livewire;
+
+trait Wireable
+{
+    public function toLivewire()
+    {
+        return serialize($this);
+    }
+
+    public static function fromLivewire($value): self
+    {
+        return unserialize($value);
+    }
+}
--- a/src/Wireable.php
+++ b/src/Wireable.php
@@ -2,15 +2,9 @@
 
 namespace Livewire;
 
-trait Wireable
+interface Wireable
 {
-    public function toLivewire()
-    {
-        return serialize($this);
-    }
+    public function toLivewire();
 
-    public static function fromLivewire($value): self
-    {
-        return unserialize($value);
-    }
+    public static function fromLivewire($value): self;
 }
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -12,6 +12,7 @@ use Illuminate\Support\Carbon as IlluminateCarbon;
 use Illuminate\Contracts\Queue\QueueableCollection;
 use Illuminate\Queue\SerializesAndRestoresModelIdentifiers;
 use Livewire\Exceptions\PublicPropertyTypeNotAllowedException;
+use Livewire\Wireable;
 use ReflectionProperty;
 
 class HydratePublicProperties implements HydrationMiddleware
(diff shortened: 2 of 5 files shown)


ryangjchandler:
I'd love to see this introduced in v2. The overhead is minimal and is more of a power user feature but the functionality it provides can simplify lots of code I have in a dozen projects.

ryangjchandler:
I think it's also worth mentioning that this is basically impossible to implement in user land due to Livewire's strict type checks in the hydration middleware.

danharrin:
The failing test now is not related to this feature, the tests just need rerunning AFAIK.

src/Wireable.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Livewire;
+
+interface Wireable
+{
+    public function toLivewire();
```
nuernbergerA, line 7: in all your examples it returns an array, is there anything against a return type?

danharrin: Originally there was an array return type, but I did not want to close off the possibility of serializing to a string as well, especially if the DTO only holds one value 👍 

Once LW drops PHP <8.x support, maybe `array|string` would be good.

roni-estein:
Harrin ... so hot right now. Harrin. Sorry, I have Zoolander on my mind. You are so hot right now too @ryangjchandler! Love this! @danharrin

joshhanley:
@danharrin looks good!

Really could have used this, this week. Using event sourcing with Livewire, so would be really handy to bind to a DTO to pass through to the event system!

My only query is around the use of the dedicated `Wireable` interface. Does Laravel have an interface already for serialising/deserialising that could be used instead? But maybe there is benefit to Livewire providing and maintaining its own interface.

danharrin:
Yeah, we could use PHP's `__serialize()` and `__unserialize()` methods, but then the Livewire-specific serlialization and binding behaviour is not supported. This method makes sense to me, it also allows us to check for objects that implement this interface strictly and still throw an exception for any unsupported objects.

danharrin:
> Harrin ... so hot right now. Harrin. Sorry, I have Zoolander on my mind. You are so hot right now too @ryangjchandler! Love this! @danharrin 

Haha thanks Roni 😘

joshhanley:
Yep makes sense. Yeah thats a good point about being able to verify specifically if something is a Livewire serialisable object or not!

ryangjchandler:
> Harrin ... so hot right now. Harrin. Sorry, I have Zoolander on my mind. You are so hot right now too @ryangjchandler! Love this! @danharrin

Thanks Roni! ❤️

calebporzio:
This is great - thanks @danharrin & @ryangjchandler !

PR #3363, opened 2021-07-18
Better support for blade component layouts.

JoshMoreno:
1️⃣ Is this something that is wanted/needed? Did you create an issue/discussion about it first?
Yes, it's wanted/needed. Issues and discussions already existed so I didn't create any new ones. Plus, it's hard to discuss it without seeing code, at least for me anyway.
- #2499 
- #1807 
- #1707 
- https://forum.laravel-livewire.com/t/layout-class-constructor-is-not-triggered-while-rendering/1985
- https://forum.laravel-livewire.com/t/livewire-with-custom-layout/1666/8

No, all related

Yes! 
I tried my best to keep it consistent with how I saw the rest of the tests being written however, all the foo, bar, baz stuff was a little hard to keep track of. Mostly cause the slot names and the data rendered in them were mixing them. Also, naming the test classes was challenging since they're all in the same namespace so need to be unique. So please feel free to refactor to whatever makes sense to you.

So @RomainMazB had added support for using class-based component layouts in #2499. Thanks, @RomainMazB 🙏!
This builds on top of that and adds support for:
- Setting class-based component layouts in the config file
- Setting anonymous (aka class-less) component layouts in the config file
- Using anonymous component layouts via `Livewire\Macros\ViewMacros::layout()`
- `Illuminate\View\ComponentAttributeBag $attributes` is always available to the layout components now
- can fill `$attributes` with `Livewire\Macros\ViewMacros::layout()` or `Livewire\Macros\ViewMacros::layoutData()`

No prob. 🙌

### Removed protected `Livewire\Component::$initialLayoutConfiguration`

This property didn't seem needed. We pass the view in the `component.rendered` event and it was only being read from in `Livewire\Component::__invoke()`. But idk what your policy is for supporting undocumented things that were purely intended for internal use. It's possible that people were creating livewire components and using `$initialLayoutConfiguration` as an alternative to calling the various view macros that populate `$initialLayoutConfiguration`. But I removed it to make the code simpler and overall have less "public" code to support.

### `$view->livewireLayout['params']['attributes']` is handled uniquely.
If the user didn't call `layout()` or `extends()` then we'll end up calling `layout()` which will call `Illuminate\View\Component::data()` which returrns a `Illuminate\View\ComponentAttributeBag`  as the value for the `attributes` key. This will end up replacing the one from the user's livewire component when we merge with `array_merge($params, $layout->data())`.

However, we do pass it to the layout component like with `$layout->withAttributes($params['attributes'] ?? []);`.  So this should only really be an issue if they're not using `$attributes` in their view in a way that's compatible with `Illuminate\View\ComponentAttributeBag`. 

Hope all that makes sense.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -60,22 +60,11 @@ abstract class Component
             return redirect()->response($this->redirectTo);
         }
 
-        $layout = $this->preRenderedView->livewireLayout ?? [];
-        $hasCustomLayout = isset($layout['view']);
+        $this->ensureViewHasValidLivewireLayout($this->preRenderedView);
 
-        if (!$hasCustomLayout) {
-            $this->preRenderedView->layout(config('livewire.layout'), $layout['params'] ?? []);
+        $layout = $this->preRenderedView->livewireLayout;
 
-            if (isset($layout['slotOrSection'])) {
-                $this->preRenderedView->livewireLayout['slotOrSection'] = $layout['slotOrSection'];
-            }
-
-            $layout = $this->preRenderedView->livewireLayout;
-        }
-
-        $layoutType = $layout['type'] ?? 'component';
-
-        return app('view')->file(__DIR__."/Macros/livewire-view-{$layoutType}.blade.php", [
+        return app('view')->file(__DIR__."/Macros/livewire-view-{$layout['type']}.blade.php", [
             'view' => $layout['view'],
             'params' => $layout['params'],
             'slotOrSection' => $layout['slotOrSection'],
@@ -149,6 +138,18 @@ abstract class Component
         Livewire::dispatch('component.rendered', $this, $view);
 
         return $this->preRenderedView = $view;
+    }
+
+    protected function ensureViewHasValidLivewireLayout(View $view)
+    {
+        $layout = $view->livewireLayout ?? [];
+
+        $isValid = isset($layout['view'], $layout['type'], $layout['params'], $layout['slotOrSection']);
+
+        if (!$isValid) {
+            $view->layout($layout['view'] ?? config('livewire.layout'), $layout['params'] ?? []);
+            $view->slot($layout['slotOrSection'] ?? $view->livewireLayout['slotOrSection']);
+        }
     }
 
     public function output($errors = null)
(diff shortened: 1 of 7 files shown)


calebporzio:
Thank you for addressing this - please pose code examples of what this PR enables. 

You listed 5 bullet points of what this PR enables, but I would much prefer full code snippets of what each point will look like for a user of Livewire. Thank you!

JoshMoreno:
No problem. Ah, sorry! Should have provided examples. Here ya go

### Setting class-based component layouts in the config file
```php
'layout' => App\View\Components\AppLayout::class,
```

### Setting anonymous (aka class-less) component layouts in the config file
```php
'layout' => 'components.layouts.app',
```

### Using anonymous component layouts via Livewire\Macros\ViewMacros::layout()
```php
public function render()
{
    return view('livewire.show-posts')
        ->layout('components.layouts.app');
}

// works with data
public function render()
{
    return view('livewire.show-posts')
        ->layout('components.layouts.app', ['title' => 'Show Posts']);
}
```

### Illuminate\View\ComponentAttributeBag $attributes is always available to the layout components now
Before `$attributes` was only available if the layout component was a subclass of `\Illuminate\View\Component::class`. We now fallback to rendering the layout as a `\Illuminate\View\AnonymousComponent` so `$attributes` is always going to be available.
```blade
<div {{ $attributes }}></div>
```

### can fill $attributes with Livewire\Macros\ViewMacros::layout() or Livewire\Macros\ViewMacros::layoutData()
```php
public function render()
{
    return view('livewire.show-posts')
        ->layout('components.layouts.app', [
            'attributes' => [
                'class' => 'bg-gray-200'
            ]
        ]);
}

public function render()
{
    return view('livewire.show-posts')
        ->layoutData([
            'attributes' => [
                'class' => 'bg-gray-200'
            ]
        ]);
}
```

```blade
<div {{ $attributes->merge(['class' => 'container']) }}></div>
```

calebporzio:
Great work @JoshMoreno!

JoshMoreno:
Always missing these github notifications.

Thanks @calebporzio 
And thanks for making and maintaining this! 🙌

PR #3374, opened 2021-07-19
Added custom view path configuration

denizozturk:
1️⃣ A feature needed to avoid complex directory structure in "resources". I haven't created a topic about this before.

2️⃣ It only allows defining custom path for views.

3️⃣ No, it does not include testing.

4️⃣ This configuration ensures that the created Livewire component does not depend on the "views/livewire" directory of the blade file in "resources". The dependency of livewire components on the "views/livewire" directory creates directory clutter. In this way, a more controllable directory structure can be created, provided that the "resources" directory is adhered to.

Otherwise, "livewire:make" command has to change the path of the render method in the class file of the component and repeatedly change the path of the blade file in "resources/views/livewire"

img


3alampro:
Love it 🎉

denizozturk:
I tried but couldn't do it. If it's possible to do this without this feature, I'd love to learn how to do it. Positioning of the components class and blade file is an important issue, to avoid confusion

calebporzio:
Hmm interesting - there must be a way to make the current config options achieve what you want? It doesn't make sense to me to have a `view_path` config and then a `custom_view_path` config directly under it.

denizozturk:
> Hmm interesting - there must be a way to make the current config options achieve what you want? It doesn't make sense to me to have a `view_path` config and then a `custom_view_path` config directly under it.

Got it, I updated it as you said. I hope it was successful :)

Test:
`'view_path' => resource_path(),`

screen

joshhanley:
@denizozturk I may be misunderstanding, so please correct me if I'm wrong, but the default config option for `view_path` is actually this

https://github.com/livewire/livewire/blob/396c5c910ecf747a55ad4bc1f8d07790068154df/config/livewire.php#L30

So if you want your Livewire components to create its views in say the `resources/views/components` directory, all you would need to do is set

```php
'view_path' => resource_path('views/components'),
```

I have just done a test and it seemed to work fine for me.

image

Can you confirm that is what you are after?

denizozturk:
> @denizozturk I may be misunderstanding, so please correct me if I'm wrong, but the default config option for `view_path` is actually this
> 
> https://github.com/livewire/livewire/blob/396c5c910ecf747a55ad4bc1f8d07790068154df/config/livewire.php#L30
> 
> So if you want your Livewire components to create its views in say the `resources/views/components` directory, all you would need to do is set
> 
> ```
> 'view_path' => resource_path('views/components'),
> ```
> 
> I have just done a test and it seemed to work fine for me.
> 
> image
> 
> Can you confirm that is what you are after?

No, my goal is to be able to freely create the livewire component inside the "resources" folder without being tied to the "views" folder. I tried hard to do this, but I couldn't manage to do it except the commit I sent.

`php artisan make:livewire component.child`

When this command runs, `resources/component/child.blade.php` file should be created and this blade file should be defined correctly file path in the created livewire class file.

I want to define livewire components freely in resource, but it doesn't seem possible to do this with existing build.
It is very important that the folder structure is organized and understandable in our projects. In this way, I can make the backend and frontend components of the projects I develop more organized. It is debatable whether my method is the most correct method, but I think this arrangement is needed :)

joshhanley:
@denizozturk ah ok, I understand now.

But Livewire's render methods make use of the `view` helper, which by default generates a view from the `resources/views` directory. 

Adding this PR, so the views get generated anywhere in the resources folder won't fix that issue though. 

So I'm curious how you intend to load the views properly in the render method if they're not in the views directory?

joshhanley:
@denizozturk so I just did a test of setting the `'view_path' => resource_path()` and it seemed to work fine for generating

image

But when you look at the components render method, it's not happy.

image

So the resource path generation works fine from my perspective but it's actually the render method that has issues.

Can you confirm?

denizozturk:
Yes you are right. I did not get this problem locally, but when I now tested it on the server, I got a similar path error. I will send a commit again to fix this problem, as soon as possible.

joshhanley:
@denizozturk can you screenshot the path error?

denizozturk:
> @denizozturk can you screenshot the path error?

I didn't get such error on local server. When I checked after you said on the server, such a way was created. It should be `test.child`

screenshot

joshhanley:
@denizozturk ah ok cool, no worries 🙂 does that component render properly?

denizozturk:
> @denizozturk ah ok cool, no worries 🙂 does that component render properly?

Yes everything is fine now, render path is working correctly

calebporzio:
@denizozturk - Glad you're getting closer. However, I don't like an implementation that splits logic for a one specific folder setting (resource_path).

Is there a way to restructure Livewire's logic so that there are no conditionals but it still supports your use case?

If not we may have to say no to this addition as I think wanting to store views in `resource_path` root and not under views is an unpopular approach.

Thank

denizozturk:
> @denizozturk - Glad you're getting closer. However, I don't like an implementation that splits logic for a one specific folder setting (resource_path).
> 
> Is there a way to restructure Livewire's logic so that there are no conditionals but it still supports your use case?
> 
> If not we may have to say no to this addition as I think wanting to store views in `resource_path` root and not under views is an unpopular approach.
> 
> Thanks

This preference is of course yours, but class files can be managed very well in the `Http/Livewire` directory, but I cannot say this for view files.

This PR can be rejected, but I believe there should be a solution to this problem, it doesn't seem very healthy to me to have view components in different folders, inconsistent from each other. As in the example, if we are creating a component with livewire, we should be able to define it freely. The default path remains `views/livewire`, such a preference can be offered as an option only for those who want to define a custom route.
68747470733a2f2f69322e70617374652e706963732f63383564306637396139323930666130353465613464336439333833326230332e706e67

Thanks

joshhanley:
> Yes everything is fine now, render path is working correctly

Ah good to know! How did you manage to get the `view` helper to render a file that's not in the `resources/views` directory? Did you have to change your Laravel config?

denizozturk:
> > Yes everything is fine now, render path is working correctly
> 
> @denizozturk ah good to know! How did you manage to get the `view` helper to render a file that's not in the `resources/views` directory? Did you have to change your Laravel config?

Yes, you can try updating `view_path` to `resource_path()` in your config file. 
If you are not going to use this feature, you can continue to use `resource_path('views/livewire')' as the default value.

screenshot-work

joshhanley:
@denizozturk ah yep, great thanks!

Have you had a chance to have a look to see if this can be implemented in a cleaner way like Caleb was wondering?

calebporzio:
Ok, thanks @denizozturk

Issue #3357, opened 2021-07-17
Full page components not accepting route parameters


AmG2000:
I noticed this issue after I upgraded from version 2.4.4 to version 2.5.2 (because of the google cloud storage disk support for temporary file uploads). Problem is, now all my full page components that make use of route model binding now fail with a `BadMethodCallException,
Method Illuminate\Routing\Route::getMissing` does not exist.  

Here's an example of one of my components and the corresponding route definition. 

When I comment out the mount method and the Recipe variable and all its references, everything works well. 

Component class

`
class SubmitRecipeFiles extends Component
{
    use WithFileUploads, AuthorizesRequests;

    public $images = [];

    public $recipe;

    public function mount(Recipe $recipe)
    {
        $this->recipe = $recipe;
    }

    public function updatedImages()
    {
        $this->validate([
            'images' => ['required','array'],
            'images.*' => ['required', 'image', 'mimes:jpeg,jpg,png', 'max:10000'],
        ]);
    }

    public function saveImages()
    {
        $this->authorize('update',$this->recipe);
        $unsavedImages = [];
        foreach ($this->images as $image) {
            $filePath = Storage::disk('gcs')->putFile('recipeImages', $image);
            $image = new Image(['url' => Storage::disk('gcs')->url($filePath)]);
            array_push($unsavedImages, $image);
        }
        $this->recipe->images()->saveMany($unsavedImages);

        return redirect('/chef');
    }
    public function render()
    {
        return view('livewire.chef.submit-recipe-files')
        ->extends('layouts.dashboard_master')
        ->section('main');
    }
}
`

Route
`Route::get('/recipes/{recipe}/files', [\App\Http\Livewire\Chef\SubmitRecipeFiles::class, '__invoke'])->middleware('auth:chef');`

Laravel 7.22.2
php 7.3.1
livewire 2.5.2

krokovicsmate:
@AmG2000 Your route is not required the `__invoke`, maybe thats the problem. 
`Route::get('/recipes/{recipe}/files', \App\Http\Livewire\Chef\SubmitRecipeFiles::class)->middleware('auth:chef');`

AmG2000:
It does require the __invoke since i'm using laravel 7.

joshhanley:
@AmG2000 it should still work ok in Laravel 7 the way @krokovicsmate suggested. The only thing that changed between 7 and 8 was removal of prefixing the controller path.

[...]

PR #3401, opened 2021-07-23
add check for getMissing function

ariaieboy:
https://github.com/livewire/livewire/issues/3357
no
no
i've just add a method_exists check to resolve the issue that i mentioned


calebporzio:
Thanks!

PR #3419, opened 2021-07-27
Immutable datetimes

rubenvanassche:
Adds support for immutable versions of DateTime and Carbon, `DateTimeImmutable` and `CarbonImmutable` respectively. At the moment using such objects as a public property within a component will  throw a `PublicPropertyTypeNotAllowedException` although they are perfectly valid to be used.

Tests were updated and this PR does not have breaking changes.


calebporzio:
This is great - thanks!

PR #3426, opened 2021-07-27
Fix deep nesting relation persistence

joshhanley:
There was an issue with the deep nesting not persisting relations correctly to the database. 

I've updated a tests to include persistence and fixed the problem.


ryatkins:
I've just tested this on my code base and it is working now to save models with relations.

Nested models hierarchy
`'groups.*.forms.*.parts.*.parts_desc'`

Saving of models works with either of the following when ran inside `updated()`

`$this->groups->each->push();`

OR long form:

```
foreach ($this->groups as $group) {

    $group->save();

    foreach ($group->forms as $form) {

        $form->save();

        foreach ($form->parts as $part) {

            $part->save();

        }
    }
}
```

PR #3439, opened 2021-07-29
Removed the wrapper div for assertSee

mertasan:
1️⃣ Is this something that is wanted/needed? Did you create an issue / discussion about it first? : #3290

2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out. : No

3️⃣ Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have): Yes

---

This PR relates to: #3290 #62

This way I think it would be more appropriate to completely remove the wrapper div.

If the wrapper has no other attributes:

```blade
<!-- before -->
<div wire:id="foo" wire:initial-data="bar">
content
</div>

<!-- after -->
content
```

If the wrapper has another attribute:

```blade
<!-- before -->
<div class="hello" wire:id="foo" wire:initial-data="bar">
content
</div>

<!-- after -->
<div class="hello">
content
</div>
```

### Regex: https://regex101.com/r/XKeZYf

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -119,7 +119,13 @@ trait MakesAssertions
 
     protected function stripOutInitialData($subject)
     {
-        return preg_replace('/((?:[\n\s+]+)?wire:initial-data=\".+}"\n?|(?:[\n\s+]+)?wire:id=\"[^"]*"\n?)/m', '', $subject);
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks for the submission, however, I'm going to hold off on this for the reasons Josh states in the referenced PR and advise what he advised:

> @mertasan ah right, thanks for the detailed explanation, that makes sense.
> 
> I'm not sure if it would be accepted or not, but if you have already written the code, it can't hurt to submit the PR 🙂
> 
> If it doesn't get in though, `src\Testing\TestableLivewire.php` is macroable, so you could create your own package that adds that functionality to the testable livewire class, so you and others can use it nicely!

Thanks!

PR #3440, opened 2021-07-29
Add lastRenderedCleanDom

mertasan:
1️⃣ Is this something that is wanted/needed? Did you create an issue / discussion about it first? : https://github.com/livewire/livewire/pull/3290#issuecomment-879553472

2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out. : No

3️⃣ Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have) : No

This PR relates to: #3439 

**Details**: https://github.com/livewire/livewire/pull/3290#issuecomment-879553472

This returns a DOM with `wire:id` and `wire:initial-data` data removed from lastRenderedDom.

( If #3439 is merged, it returns a DOM with the wrapper div removed as well )

For example, when testing with Pest, clean results can be obtained with such use.

```php
// PestPHP

expect($component->lastRenderedCleanDom)
      ->toStartWith('foo')
      ->toEndWith('bar');
```

The `assertSee` function produces results. 

But we can use the filtered data multiple times with `$this->lastRenderedCleanDom`.

I sampled this with Pest. But actually I think this will be useful for everyone.

### DIFF
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -26,6 +26,7 @@ class TestableLivewire
     public $lastErrorBag;
     public $lastRenderedView;
     public $lastRenderedDom;
+    public $lastRenderedCleanDom;
     public $lastResponse;
     public $rawMountedResponse;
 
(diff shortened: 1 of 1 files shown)


mertasan:
For the failed action ( https://github.com/livewire/livewire/pull/3440/checks?check_run_id=3196922871 ):

screenshot 78

calebporzio:
Thanks for the PR, although, this too, I really don't see the need and would advise you use macros and such to achieve what you want for your own projects. Thanks!

PR #3454, opened 2021-08-01
Add support for Laravel `FormRequest` inside the components 

HajMo:
This PR will add the functionality to support using Laravel `FormRquest`s.

All what you need to do is set your rules, messages and attributes inside a `FormRquest` that generated by Laravel then define a `$formRequest` property inside your component that will be hold the full class name, Just like this:
```php
    protected $formRequest = AttributesRequest::class;
```
And it will extract the rules, messages and attributes from the `FormRquest` class and handed to Livewire to handle it as the `$rules` property.

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -75,6 +75,11 @@ trait ValidatesInput
     {
         if (method_exists($this, 'rules')) return $this->rules();
         if (property_exists($this, 'rules')) return $this->rules;
+        if (property_exists($this, 'formRequest')) {
+            $formRequestInstance = $this->formRequest::createFrom(request());
+
+            return (new \ReflectionMethod($this->formRequest, 'rules'))->invoke($formRequestInstance);
+        }
 
         return [];
     }
@@ -83,6 +88,11 @@ trait ValidatesInput
     {
         if (method_exists($this, 'messages')) return $this->messages();
         if (property_exists($this, 'messages')) return $this->messages;
+        if (property_exists($this, 'formRequest')) {
+            $formRequestInstance = $this->formRequest::createFrom(request());
+
+            return (new \ReflectionMethod($this->formRequest, 'messages'))->invoke($formRequestInstance);
+        }
 
         return [];
     }
@@ -91,6 +101,11 @@ trait ValidatesInput
     {
         if (method_exists($this, 'validationAttributes')) return $this->validationAttributes();
         if (property_exists($this, 'validationAttributes')) return $this->validationAttributes;
+        if (property_exists($this, 'formRequest')) {
+            $formRequestInstance = $this->formRequest::createFrom(request());
+
+            return (new \ReflectionMethod($this->formRequest, 'attributes'))->invoke($formRequestInstance);
+        }
 
         return [];
     }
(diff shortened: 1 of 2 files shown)


danswiser:
At the end of the [validation page](https://laravel-livewire.com/docs/2.x/input-validation) of the docs it states:

> You might be wondering if you can use Laravel's "FormRequest"s. Due to the nature of Livewire, hooking into the http request wouldn't make sense. For now, this functionality is not possible or recommended.

However, I support this PR as I've already found myself copying or importing the rules() method defined in existing form requests I've already created.

My only question with this would be dealing with binding with a model. For example, I have a livewire `CreateUserForm` with a `public $user;` property referencing the User to which the form relates and each input has `wire:model="user.name"`. What would be the suggested way to bind to properties when using the related FormRequest?

devcircus:
Seems we'd still not have access to most of the functionality of FormRequests, and it would basically be a wrapper around validation attributes. Is there anything gained except a way to use a separate object to store validation rules, messages, etc.?

danswiser:
I suppose not much else is gained but I'd much rather have a single place where model validation is stored. Then for custom rules, manual validation is still available in the component class if needed.

HajMo:
@devcircus Is there anything that the FormRequests provide rather than the basic operations(rules, messages, etc.)? What you suggest adding more? Do you have any use case except the model binding?

danswiser:
In addition to the rules and messages, the FormRequest class also grants the ability to prepare input for validation as well as authorize the request. I definitely see extra benefits to fully utilizing the FormRequest class but for me even just reusing already defined rules prevents having to store identical validation rules in more than one location.

https://laravel.com/docs/8.x/validation#form-request-validation

HajMo:
@danswiser I've added ability to use `authorize` function from FormRequets. After more investigations, there's no clear way to add Laravel validation 'hooks' (prepareForValidation, passedValidation) to Livewire. Am I missing any features?

calebporzio:
Hi @HajMo - thanks for the contribution, however, I think for now I don't want to support FormRequests inside Livewire.

Like you mentioned in this thread, for most people the main reason for wanting this, is to use the rules from the FormRequest.

In these cases, you can use the following to achieve that:

```php
```

HajMo:
@calebporzio The main reason for this PR is the reusability. You can write all your rules in one place and use it everywhere. It's get harder to keep maintain/changing rules in more than one place

nuernbergerA:
@HajMo you can make a trait and use it

danswiser:
As an example and also a request for feedback in case there's a better way to handle this, here's my current process of importing rules and also binding to a model:

```
// app/Http/Livewire/Forms/CreateUserForm.php

class CreateUserForm extends Component
{
    public User $user;

    public function mount(?User $user = null)
    {
        $this->user = $user ?? new User;
    }

    public function rules()
    {
        foreach((new UserRequest)->rules() as $key => $value) {
            $rules['user.' . $key] = $value;
        }
        return $rules;
    }

    public function save()
    {
        $this->user->save();
    }

    public function render()
    {
        return view('livewire.forms.create-user-form');
    }
}
```

```
// resources/views/livewire/forms/create-user-form.blade.php

<form wire:submit.prevent="save" ..>
    <input wire:model.lazy="user.name" name="name"/>
    <input wire:model.lazy="user.email" name="email"/>
</form>
```

x7ryan:
Interesting enough I just wrote this same thing for my app. So I can share the rules from a FormRequest class to a Livewire component without actually having to copy/paste the rules into the Livewire component and having to maintain two copies (as I still have to maintain the FormRequest class for traditional Laravel controllers serving API endpoints).

My implementation is as a simple trait that defines a rules() and messages() method that return the response from that method of the FormRequest. Not sure if it is as robust as this. But I mention this to say there are use-cases particularly for those maintaining a mix of full page Livewire components and Laravel Controllers that need to share rules either due to an in-progress refactoring to Livewire or in my case to also support API endpoints.

PR #3481, opened 2021-08-05
[1.x] Add nonce to livewire script tag

ckovey:

Yes, for CSP.   Created issue https://github.com/livewire/livewire/discussions/3444

No

No

Using a CSP rule of `script-src 'nonce-xxx' 'strict-dynamic' https:;` for Laravel 6.x site almost works with livewire 1.x

A nonce is added to the inline script, but not the script source.  Changing https://github.com/livewire/livewire/blob/1.x/src/LivewireManager.php to use

    <script src="{$fullAssetPath}" data-turbolinks-eval="false"{$nonce}></script>
    <script data-turbolinks-eval="false"{$nonce}>

Resolves the CSP warning and the master branch already has this change https://github.com/livewire/livewire/blob/master/src/LivewireManager.php#L277

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -243,7 +243,7 @@ HTML;
         // because it will be minified in production.
         return <<<HTML
 {$assetWarning}
-<script src="{$fullAssetPath}" data-turbolinks-eval="false"></script>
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

PR #3485, opened 2021-08-07
Add `Content-Type` to file download info in effects

mertasan:
1️⃣ Is this something that is wanted/needed? Did you create an issue / discussion about it first? No

2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out. No

3️⃣ Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have) Yes

As you know, the file download response can sometimes have a different `Content-Type` header.

For example, after changing the content-type with the method **[here](https://docs.laravel-excel.com/3.1/exports/export-formats.html#csv).**, it was not easy to test with Livewire.

In addition to these, I separated the assertions so that only the content or content-type of the downloaded file can be tested.

And in `assertFileDownloaded` method I changed the `$filename` to optional. This way it can only be tested that the file download has started.


calebporzio:
Hey @mertasan, I'm open to supporting this, but I think we should actually set the content-type of the file that get's produced from the blob - not just capture the content-type for assertions.

Are you open to exploring / adding that. You are already sending it down, you would just need to pass it in as a separate parameter to: base64toBlob(...):

https://github.com/livewire/livewire/blob/1974981049bb171b1ec46dd3e3dc371254071970/js/component/FileDownloads.js#L13

calebporzio:
And ideally also add Dusk tests for that case

mertasan:
Hi @calebporzio, I did what you said. 

Thanks.

calebporzio:
Thanks a bunch!

This won't affect the current behavior of triggering a download without specifying content-type right?

Before the contentType param in the JS file defaulted to `''` so it was always an empty string. Will that be the same behavior now?

If not, is that OK? Or maybe it's even better this way? I'm not sure, just want to make sure we are preserving behavior while enhancing downloads for this use case.

Thanks for the solid PR!

mertasan:
@calebporzio Honestly, I wasn't sure after you wrote it.

I tested it in many different ways after you wrote it. All tests passed. I only included a few of them in the PR so as not to add unnecessary tests.

Just in case, I forced the null content-type to the empty string for forward compatibility.

As always, thank you very much for your interest and sincere communication.

calebporzio:
Thanks!

mertasan:
You're welcome! 🎉

PR #3495, opened 2021-08-10
Fix octane issue with unresolved route params

dimitri-koenig:
If laravel octane is used, and a livewire component gets called again after the initial request, the following expection is thrown:

Illuminate\Contracts\Container\BindingResolutionException
Unresolvable dependency resolving [Parameter #0 [ <required> $methods ]] in class Illuminate\Routing\Route

Since laravel octane does not resolve the route binding propertly, the following PR utilizes another way of resolving that class.

This should fix the open issue https://github.com/laravel/octane/issues/267

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -39,8 +39,10 @@ abstract class Component
         $this->ensureIdPropertyIsntOverridden();
     }
 
-    public function __invoke(Container $container, Route $route)
+    public function __invoke(Container $container)
     {
+        $route = request()->route();
+
         try {
             $componentParams = (new ImplicitRouteBinding($container))
                 ->resolveAllParameters($route, $this);


brocorp-dev:
Hi @dimitri-koenig, 

Thanks to you! 

We have deployed your PR to a development version and so far we are no longer seeing a 500 error, although it is still too early to be sure.

Octane is started with the following parameters, via Supervisor:
php artisan octane: start --workers=2 --task-workers=4 --max-requests=100

joshhanley:
@dimitri-koenig thanks for the PR!

Can I just confirm, it has been discussed in this issue #3043 that components should actually be destroyed at the end of its request cycle, which fixes some other issues with Octane and caches.

If we were to destroy the component, so it is invoked properly from scratch on each request, would we still need this PR or would it solve this issue too?

I haven't had a chance to play with Octane yet, as such haven't looked into it, so would appreciate your thoughts on it.

Hope this helps!

dimitri-koenig:
@joshhanley I've just tested the described issue in #3043 and I can't reproduce it with the most current versions of laravel and livewire. The targeted livewire component is created anew on every request, no persistence there.

joshhanley:
@dimitri-koenig thanks for checking it out, much appreciated! I wonder why they were having issues with computed properties then.

Can you clarify then, for this PR, if each Livewire component is created a new, then how come invoke isn't getting the correct instance of the container using DI, but using the app helper to get the router and component instance works?

Thanks again!

dimitri-koenig:
Great question. Don't know. I was already deep into the rabbit hole and didn't dig even deeper into why it wasn't resolved properly.

michael-rubel:
I've tested it on my full-page components project.
Looks like it's going okay. Will it be merged?

calebporzio:
Thanks for this @dimitri-koenig - happy to merge it.

2 hesitations:
A) By changing the signature of `__invoke` are we introducing breaking changes for anyone?
B) I would like a little bit of a deeper explanation to some of the questions posed by @joshhanley before I feel comfortable merging - not sure if it's possible to add a test (probably not) but that would help ensure this doesn't get reintroduced and would help us understand the nature of the problem.

Again, thanks a lot for this contribution - very helpful!

dimitri-koenig:
Hi guys

Tracing this bug is really time consuming and tedious. Hope we can find a solution without a total indepth analysis.

@calebporzio, concerning A) honestly, could be, but only in really rare cases. When someone needs to override __invoke, then that could require some minor changes. But whoever wants to override __invoke usually is in some deep water anyway, so that risk could be tolerable.

Concerning B) I cannot offer more explanation on some specifics. There are two questions to be answered:
1) Why did some people had issues with computed properties, which I simply could not replicate, no matter what.
2) Why does the laravel core not properly inject \Illuminate\Routing\Route ??

I only have the answer for question 2 and that's a rather surprising one: it's a lucky case because \Illuminate\Routing\Route has parameters which cannot be resolved by injection, no way. But it works because: way before the livewire component is dispatched with __invoke, the Laravel Router attempts to find the first fitting route (method "findRoute"). If a route has been found it will, the route will be manually bound, though it's not a singleton. So next time laravel needs to inject the Route class into the livewire component, it takes a look at all the bound instances and finds one for Route, and voila: injecting goes smoothly.

It becomes tricky with octane and here I'm stopping to dig deeper since I'm already deep in the rabbit hole with octane.
The findRoute method binds the Route class in its instances property, but a couple of methods later the instances array has no Route bound to it. That's why the laravel service container cannot resolve the Route class properly, because usually the Route class is manually called via Route::get/post methods, which leads creating an instance of Route with already proper parameters.

Hope that helps and resolves any remaining questions.

emaia:
I faced this issue and this PR fixed it. No side effects, apparently. Thanks @dimitri-koenig

dimitri-koenig:
I’ll rework this PR a little to make sure we introduce no breaking changes…

dimitri-koenig:
@calebporzio I reworked this PR so that it should not have any breaking changes at all, no risk there.

PR #3499, opened 2021-08-10
Fix Google Cloud Storage preview URL in Temporary file upload

nidhalkratos:
Yes, this is a bub that we have been facing, issue is at https://github.com/livewire/livewire/discussions/3483
NO, one change to add support for google cloud storage preview upload
No
The change lets livewire correctly preview uploaded files (eg:images) when using google cloud storage instead of S3. Both GCS and S3 have a very similar storage engine and API, but not identical implementation. 


calebporzio:
Thanks!

PR #3519, opened 2021-08-13
Fix php8 ReflectionUnionType::isBuiltin()

tanthammar:
In php8 the `ReflectionType` has become abstract and `ReflectionType::isBuiltin()` has been moved to `ReflectionNamedType::isBuiltin()`.

The `$type->isBuiltin()` results in an error:

```php
Call to undefined method ReflectionUnionType::isBuiltin()
```

https://www.php.net/manual/en/class.reflectiontype.php

Discussed in #3403

### DIFF
--- a/src/ImplicitlyBoundMethod.php
+++ b/src/ImplicitlyBoundMethod.php
@@ -104,17 +104,9 @@ class ImplicitlyBoundMethod extends BoundMethod
     {
         $type = $parameter->getType();
 
-        if ($type == null){
-            return null;
-        }
-
-        if ($type instanceof \ReflectionNamedType){
-            return !$type->isBuiltin() ? $type->getName() : null;
-        }
-
-        if ($type instanceof \ReflectionUnionType) {
-            return static::getReflectionUnionTypeName($type);
-        }
+        if ($type == null) return null;
+        if ($type instanceof \ReflectionNamedType) return !$type->isBuiltin() ? $type->getName() : null;
+        if ($type instanceof \ReflectionUnionType) return static::getFirstReflectionUnionTypeName($type);
 
         return !$type->isBuiltin() ? $type->getName() : null;
     }
@@ -122,26 +114,27 @@ class ImplicitlyBoundMethod extends BoundMethod
     public static function implementsInterface($parameter)
     {
         $type = $parameter->getType();
-        if ($type instanceof \ReflectionUnionType) {
-            $name = static::getReflectionUnionTypeName($type);
-        } else {
-            $name = $type->getName();
-        }
-        if ($name !== null) {
-            return (new ReflectionClass($name))->implementsInterface(ImplicitlyBindable::class);
-        } else {
-            return false;
-        }
+
+        $name = ($type instanceof \ReflectionUnionType)
+            ? static::getFirstReflectionUnionTypeName($type)
+            : $type->getName();
+
+        return $name !== null
+            ? (new ReflectionClass($name))->implementsInterface(ImplicitlyBindable::class)
+            : false;
     }
 
-    public static function getReflectionUnionTypeName($type)
+    public static function getFirstReflectionUnionTypeName($type)
     {
         $union_types = [];
+
         foreach ($type->getTypes() as $union_type) {
             if(!$union_type->isBuiltin()) $union_types[] = $union_type->getName();
         }
+
         if (count($union_types) == 1) return $union_types[0]; //string
         if (count($union_types) > 1) throw new \Exception('Livewire does not support multiple, non-builtin union types');
+
         return null;
     }
 }
--- a/src/ImplicitlyBoundMethod.php
+++ b/src/ImplicitlyBoundMethod.php
@@ -99,10 +99,24 @@ class ImplicitlyBoundMethod extends BoundMethod
 
         return $model;
     }
-
+    
     public static function getParameterClassName($parameter)
     {
         $type = $parameter->getType();
+
+        if ($type === null){
+            return null;
+        }
+
+        if ($type instanceof \ReflectionNamedType){
+            return $type->isBuiltin() ? null : $type->getName();
+        }
+
+        if ($type instanceof \ReflectionUnionType){
+            foreach ($type->getTypes() as $union_type){
+                return $union_type->isBuiltin() ? null : $union_type->getName();
+            }
+        }
 
         return ($type && ! $type->isBuiltin()) ? $type->getName() : null;
     }
--- a/src/ImplicitlyBoundMethod.php
+++ b/src/ImplicitlyBoundMethod.php
@@ -105,7 +105,7 @@ class ImplicitlyBoundMethod extends BoundMethod
         $type = $parameter->getType();
 
         if ($type == null) return null;
-        if ($type instanceof \ReflectionNamedType) return !$type->isBuiltin() ? $type->getName() : null;
+
         if ($type instanceof \ReflectionUnionType) return static::getFirstReflectionUnionTypeName($type);
 
         return !$type->isBuiltin() ? $type->getName() : null;
--- a/src/ImplicitlyBoundMethod.php
+++ b/src/ImplicitlyBoundMethod.php
@@ -104,21 +104,22 @@ class ImplicitlyBoundMethod extends BoundMethod
     {
         $type = $parameter->getType();
 
-        if ($type === null){
+        if ($type == null){
             return null;
         }
 
         if ($type instanceof \ReflectionNamedType){
-            return $type->isBuiltin() ? null : $type->getName();
+            return !$type->isBuiltin() ? $type->getName() : null;
         }
 
         if ($type instanceof \ReflectionUnionType){
             foreach ($type->getTypes() as $union_type){
-                return $union_type->isBuiltin() ? null : $union_type->getName();
+                if(!$union_type->isBuiltin()) return $union_type->getName();
             }
+            return null;
         }
 
-        return ($type && ! $type->isBuiltin()) ? $type->getName() : null;
+        return !$type->isBuiltin() ? $type->getName() : null;
     }
 
     public static function implementsInterface($parameter)
--- a/src/ImplicitlyBoundMethod.php
+++ b/src/ImplicitlyBoundMethod.php
@@ -112,11 +112,8 @@ class ImplicitlyBoundMethod extends BoundMethod
             return !$type->isBuiltin() ? $type->getName() : null;
         }
 
-        if ($type instanceof \ReflectionUnionType){
-            foreach ($type->getTypes() as $union_type){
-                if(!$union_type->isBuiltin()) return $union_type->getName();
-            }
-            return null;
+        if ($type instanceof \ReflectionUnionType) {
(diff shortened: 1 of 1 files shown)


fabio-ivona:
It could be a good idea to add tests to it

jrmajor:
I think this method is used for dependency injection, so it would make sense to return `null` for `ReflectionUnionType`.

fabio-ivona:
> I think this method is used for dependency injection, so it would make sense to return `null` for `ReflectionUnionType`.

I agree with this

My first implementation would better suit our needs

@tanthammar wouldn't be worth to update your code to somethings similar to the other code you find in my answer for the discussion?

joshhanley:
I think I agree with @jrmajor and @fabio-ivona that the if the union reflection type is detected that it should return null.

As this is used for dependency injection, if we were to try and support DI for a union type, then in theory all types in a union should be returned and each attempted to be pulled from the container.

But then that leads to questions of does union type order then matter? So do we just get the first that can be resolved? What if multiple can actually be resolved? 

I have found a good discussion on that here https://www.reddit.com/r/PHP/comments/kvbnao/union_types_and_php_8/

Personally, I were using unions, it would be so I could support passing both the ID of a model or an instance of a model in, for example `int | User`, or allowing two primitive types like `int|string`. Rarely would I pass two class names to a union, as that should be an interface instead, in my opinion.

For reference, `ReflectionNamedType` has been supported since PHP7.1 so changing to that shouldn't cause any backwards compatibility issues for Livewire's currently supported PHP versions. But there should probably be an existence check put before the use of `ReflectionUnionType`.

Hope this helps!

fabio-ivona:
@joshhanley I'm not 100% sure, but if I remember well, using ReflectionUnionType only in an `instanceof` statement should not trigger an error in php versions lower than 8. A good test set would ensure this will work. 

@tanthammar in the ReflectionUnionType check block, returning `null` would prevent to check DI for the second type and would ever stop the iteration at the first element. Could you fix this by returning a value only if is not a builtin type? 

If you have any issue, let me know and I will send you my code. Hope this could help

tanthammar:
@joshhanley @fabio-ivona 
How about if we add ReflectionClass::isInstantiable()?

fabio-ivona:
> @joshhanley @fabio-ivona
> How about if we add ReflectionClass::isInstantiable()?

Which line are you referring to?

tanthammar:
@fabio-ivona @joshhanley 

I updated. (kept the `!` negations throughout the function for code readability.)

The solution is not complete. This version will only support multiple builtin types and the first Instantiable, but not multiple Instantiables, just like @joshhanley mentioned.

```php
mount(array|User $parameter)// ok
mount(array|Foo|Bar $parameter) //will work if $parameter = Foo but not if $parameter = Bar
```

Regarding: `if ($type instanceof \ReflectionUnionType)`
I think we have three options?
* return `null` would mean Livewire won't support union types, but a quick fix.
* return first, `!isBuiltin()`, and cross fingers that devs don't do like `array|object|Foo|Bar`
* return an array with Instantiables and update all functions that calls this, to expect a return type of `string|array`

The third option requires that other functions checks if the `$parameter` is an instance of the returned `$union_type->name`.

**If we returned an array:**
```php
public static function getParameterClassName($parameter)
     //...
     if ($type instanceof \ReflectionUnionType) {
            $union_types = [];
            foreach ($type->getTypes() as $union_type){
                if(!$union_type->isBuiltin()) $union_types[] = $union_type->getName();
            }
            return (count($union_types) > 0) ? $union_types : null;
        }
     //...
}
```

**I don't know all the places where  this function is used, but I found some places that would need updating.**

ImplicitRouteBinding
```php
//row 82
return collect($component->getPublicPropertiesDefinedBySubClass())
            ->map(function ($value, $name) use ($component) {
                return Reflector::getParameterClassName(new \ReflectionProperty($component, $name));
            })
            ->filter();
```

ImplicitlyBoundMethod
```php
//row 76
protected static function getClassForDependencyInjection($parameter)
{
    if (! is_null($className = static::getParameterClassName($parameter)) && ! static::implementsInterface($parameter)) {
        return $className;
    }
}

protected static function getClassForImplicitBinding($parameter)
{
    if (! is_null($className = static::getParameterClassName($parameter)) && static::implementsInterface($parameter)) {
        return $className;
    }

    return null;
}

//row 43
// stop if this one is due for dependency injection
if (! is_null($className = static::getClassForDependencyInjection($parameter)) && ! $parameters[$paramIndex] instanceof $className) {
    return;
}

//row 63
// check if we have a candidate for implicit binding
if (is_null($className = static::getClassForImplicitBinding($parameter))) {
    return;
}
```

Personally, I would prefer to return an array, not to leave things "half" implemented.
What do you think?

tanthammar:
We could also throw an error to get a quick fix now, while still informing about the constraints.

If we returned an array:

public static function getParameterClassName($parameter)
     //...
     if ($type instanceof \ReflectionUnionType) {
            $union_types = [];
            foreach ($type->getTypes() as $union_type){
                if(!$union_type->isBuiltin()) $union_types[] = $union_type->getName();
            }
            if (count($union_types) == 1) return $union_types[0]; //string
            if (count($union_types) > 1) throw new Exception('Livewire does not support multiple non-builtin union types');
            return null;
        }
     //...
}

tanthammar:
After some testing I found that `implementsInterface()` also needed updating to support php8.
Did a new commit where I created a separate `getFirstReflectionUnionTypeName()` because it was needed in multiple places.
Single source to edit is always nice.

tanthammar:
Cleaned up the code.
Latest version:

```php
mount(array|User $parameter)// ok
mount(array|Foo|Bar $parameter) //throws "Livewire does not support multiple, non-builtin union types"
```

@fabio-ivona, @joshhanley, @jrmajor - comments?

fabio-ivona:
> Cleaned up the code.
> Latest version:
> 
> ```
> mount(array|User $parameter)// ok
> mount(array|Foo|Bar $parameter) //throws "Livewire does not support multiple, non-builtin union types"
> ```
> 
> @fabio-ivona, @joshhanley, @jrmajor - comments?

I need a PC to look thorough the code, will look at it in the next days

joshhanley:
@tanthammar I think some tests need to be added, because I believe Livewire should support union types but just not try to resolve them when using DI.

I'm concerned that by throwing the error message, that union types won't be able to be used at all.

I may be wrong though (as I'm only thinking through it in my head), and may not be fully appreciating the scenario, hence why I think some tests would be good to simulate the different instances where this would be used/ effected so we can see how it all actually runs.

I'd expect that if a user has type hinted a union type (even two class names), that it would work fine for the instance where they manually populate that property, but currently it seems like it would throw the error (again I may be wrong though 🙂).

If it does throw an error when just trying to use manually, then my preference would be to just return `null` if a `ReflectionUnionType` has been detected, so it does not try and resolve any of the the union classes.

But if it only throws the error when Livewire tries to resolve the union from the container (and all the other scenarios work fine), then I'm all for it, and think it's a great idea as it would be better than it just failing silently and the dev is left scratching their head wondering why something doesn't work.

Again, I may be misunderstanding, so please correct me if I am!

I will trigger the existing tests to run so we can see if anything is impacted by this PR.

I will also have a dig through all the other usages of this that you listed and see if there will be any other areas that need changing.

Thank you for efforts on this! 🙂

tanthammar:
@joshhanley 
I am not fond of the idea to return `null`.
* If we do, the developer will be confused why the union class is unresolved.
* If union types is not to be supported, then we should throw a helpful error, to give instant feedback to the developer.

With this PR we can do this:
```php
public function mount(string|int|User $mixed){}

// If we return null, the developer will have to spend time on finding out that he/she must separate the union types to multiple parameters.
public function mount(string $string, int $int, User $user){}
```
But not
```php
public function mount(Foo|User $object){}

///The developer would have to separate it to multiple parameters, which I think is acceptable, and a rare use case.
public function mount(Foo $foo, User $user){}
```

Regarding tests:
I don't know Livewire deep enough to cover all scenarios.
Have tests in my project for the cases I could think of, and only tested on php8...
I'll gladly drop this PR if someone reopens and takes it from here.

fabio-ivona:
I could open my PR with the code and test that were ready, but cannot work on it until the first of September, I'm currently out of office for vacation

If you can wait, I'll be glad to help

tanthammar:
@fabio-ivona 
Maybe we can take it in two steps?
This PR at least solves the current php8 problem.
You can reopen a new PR that takes it all the way?
Solving multiple instantiable types, is a much more extensive code change that requires a lot of testing...

Would be nice if @calebporzio could bless us with his brain on this ;)

joshhanley:
@tanthammar no worries, if I get a chance this week I will have a look at what tests will be needed, and work with @fabio-ivona to make sure we have it all covered 🙂

If you can leave this PR open for the moment, that'd be great, so we can keep all the discussion in the one place.

But yeah I do agree about things happening silently isn't ideal as it can be hard to debug what is going on.

And yep, the best person to decide how to handle it would be Caleb, so hopefully he gets a chance to have a look at this.

joshhanley:
@fabio-ivona and @tanthammar I have tried to summarise this issue as best as I can in this discussion #3635.

I have locked that discussion, so if you have any feedback or comments, please add them here and I will edit the discussion to keep it as clear as possible.

Hopefully I have captured it well enough 🙂

fabio-ivona:
@joshhanley seems fine to me, thanks!

calebporzio:
Thanks a ton for the awesome writeup @joshhanley - seriously can't thank you enough. It made the problem clear and the potential solutions clear as well. Saved me so much time.

My first question as always is: What does Laravel do? Like if you have a union type in a controller method, does Laravel inject anything? (it's possible I missed this part of the discussion) - but I would say we should do whatever Laravel does.

That question aside, passing null seems like the best option to me as we can't know what to inject and we don't want to disallow people the polymorphic goodness of the param if they really want the typehint.

So in short:
We should return null for DI'd union types, but first consult what Laravel itself does.

Thanks for everyone's hard work on this!

fabio-ivona:
@calebporzio and @joshhanley 

I've checked and Laravel doesn't apply dependency injection if there is a Union Type

I'm preparing a PR with this solution and tests

calebporzio:
Thank you so much @fabio-ivona !

joshhanley:
@calebporzio no worries 🙂

Thanks @fabio-ivona let me know if you need anything!

fabio-ivona:
> @calebporzio no worries 🙂
> 
> Thanks @fabio-ivona let me know if you need anything!

I'm currently working on writing good tests for it. It would be useful a test that currently fails for union types, in order to make it succeed with my implementation. thanks!

joshhanley:
@fabio-ivona only thing I experimented with was a component that had a union type the mount method.

I think I just use `int|string` in my experimentation that caused the error to throw 🙂

fabio-ivona:
@calebporzio ,  @joshhanley I've opened a PR with the fix and some test to ensure everything works

https://github.com/livewire/livewire/pull/3663

please feel free to comment and point out any issue

joshhanley:
Thanks @fabio-ivona! I will close this PR now. Thank you @tanthammar for getting the ball rolling on this 🙂

PR #3524, opened 2021-08-14
Custom query params & clean query strings

nullthoughts:

Yes, there are multiple discussions for cleaner/customizable queryStrings: #2036, #3030 and https://forum.laravel-livewire.com/t/ability-to-use-computed-properties-for-updatequerystring/236

No, the changes are related

No new tests (yet), WIP. Could use a hand with them.

This PR cleans up the encoding/readability of query strings. 

For these examples, a public `$filters` property is utilized and specified as a `$queryString` property:
```php
public $filters = [
    'make' => [],
    'model' => [],
    ...
];

public $queryString = ['filters'];
```

### Query strings with fresh Livewire install: 
`?filters%5Bmake%5D%5B0%5D=Tesla&filters%5Bmake%5D%5B1%5D=VW`

### Query strings with url decoding in PR:
`?filters[make][0]=Tesla&filters[make][1]=VW`

(Originally detailed in https://github.com/livewire/livewire/discussions/2036#discussioncomment-1168681)


joshhanley:
@nullthoughts can you rebase your changes onto the latest master and then do a build of the assets as a new commit? There have been some changes in the last couple of days that are causing merge conflicts.

Thanks!

calebporzio:
Hey @nullthoughts, thanks for the contribution. I agree with most of the API proposals in #2036 - great explanations!

Question: I'm tempted to remove the config that you proposed and just make this the default. Do you see any issues with that? Either backwards compatibility issues for people who upgrade their apps, but users still have bookmarked URLs from before the update? Or maybe some cases where people would still want the current query param encoding treatment?

Thanks!

joshhanley:
I agree re just making it the default. I've run into the same issue, but worked around it just by splitting up the properties instead of using an array, which isn't ideal.

I'm not sure on the BC implications, I mean technically yeah, it would be a breaking change. How big of an impact that would be, I can't tell. Bookmarked URL's would be the main issue I believe.

Also I wonder if instead of just the `queryParams` method, there should be two, one for hydrating from a query string into properties and another to dehydrate properties to construct the query string. How to name them is another challenge haha.

There was also an issue #2553 raised about modification of queryString after the component has mounted. I know @austenc had a need for this, maybe he can expand on it.

nullthoughts:
> Hey @nullthoughts, thanks for the contribution. I agree with most of the API proposals in #2036 - great explanations!
> 
> Question: I'm tempted to remove the config that you proposed and just make this the default. Do you see any issues with that? Either backwards compatibility issues for people who upgrade their apps, but users still have bookmarked URLs from before the update? Or maybe some cases where people would still want the current query param encoding treatment?
> 
> Thanks!

Thanks @calebporzio! When I originally implemented the config variable, it was to provide some backwards compatibility for existing applications (that won't be making use of the new features) - however the JS decoding will impact that either way.

> I agree regarding making it the default. I've run into the same issue, but worked around it just by splitting up the properties instead of using an array, which isn't ideal.
> 
> I'm not sure on the BC implications, I mean technically yeah, it would be a breaking change. How big of an impact that would be, I can't tell. Bookmarked URL's would be the main issue I believe.
> 
> Also I wonder if instead of just the `queryParams` method, there should be two, one for hydrating from a query string into properties and another to dehydrate properties to construct the query string. How to name them is another challenge haha.
> 
> There was also an issue #2553 raised about modification of queryString after the component has mounted. I know @austenc had a need for this, maybe he can expand on it.

Bookmarked URLs would be a concern. I'll see if a good solution comes to mind (other than adding additional logic just for it), open to any creative suggestions.

I really like the idea of two methods for hydrating and dehydrating the query string (with good names).

It was a beyond busy weekend, but I'll pull in the latest commits and rebuild the assets asap.

Thank you both!

calebporzio:
Cool, yeah I think I'd be fine with an extra conditional for backwards compatibility - like something that detects the encoding and properly decodes it or something. (I'm actually not fully immersed in the proposed solution so I'm going off partial knowledge)

I see the need for hydration/dehydration filter, but like you said, having two separate methods/properties wouldn't be elegant. In a perfect world we could assign a function as the value of a queryParam key and intercept the getting/setting, but silly PHP won't allow that.

We could allow classes to be set or provide a ->queryParams method that users can return an array with function values for interceptors.

Or we could expand on the Wireable trait and somehow use that for interception?

Idk, just tossing stuff out there

calebporzio:
But that stuff can be a separate PR - we can stick to the clean strings here

joshhanley:
> But that stuff can be a separate PR - we can stick to the clean strings here

Yeah agreed on this. This PR also adds a new method `$component->queryParams()` though which was for a custom construction of the query string from properties, hence the discussion. I think that should be removed for now and refine this down to the encoding issue only.

nullthoughts:
> > But that stuff can be a separate PR - we can stick to the clean strings here
> 
> Yeah agreed on this. This PR also adds a new method `$component->queryParams()` though which was for a custom construction of the query string from properties, hence the discussion. I think that should be removed for now and refine this down to the encoding issue only.

I agree, let's tackle the clean query strings now (and custom parameters later). I'll work to clean up the code and see if I can figure out a good solution to keep bookmarked urls functioning properly.

austenc:
My use case is also a `$filters` array. Looks like the changes in this PR will be very helpful in cleaning things up. 

To expand on what @joshhanley mentioned, the reason I'd need to modify the query string after a component has been mounted is for the below filters. They're dynamic, so I can't just hardcode defaults into the `$queryString` param.
CleanShot 2021-08-18 at 08 17 55

Anyway, I'll be happy to chime in on a second PR if you're planning on tackling that @nullthoughts. Really nice work on this one!

nullthoughts:
I went ahead and removed the custom `queryParams()` functionality, focusing the PR on just cleaning up the URL encoding.

Backwards compatibility: added is functionality to check for encoded characters (`%20` spaces, `%2C` commas, `%5B` brackets), and to skip decoding if they are present. Running through a few test scenarios involving mixed string and array query string properties that contained commas, hyphens and spaces, it appears we'd be able to hard-code the PHP url encoding value in the `SupportBrowserHistory` without issue.

A bit of a busy day, but I'll try to do some more extensive testing later.

joshhanley:
@nullthoughts thanks for refining the PR down 🙂 I have approved the tests to run so we can see what happens.

nullthoughts:
New tests added for bookmarked URL, tests against encoded characters in both RFC 1738 & 3986 (for backwards compatibility). 

I'm seeing some (what I believe to be unrelated) Dusk testing glitches locally (even on fresh clones of the master branch), trying to sort those out. Let me see if I can those tackled before we run the workflow tests :)

joshhanley:
@nullthoughts before you dig too much, the dusk tests on master are green, so I'd look at merging master into your branch again first and then re-running your tests locally to see what they do.

Also dusk can be a bit fickle, so if you are seeing some tests failing that you think are unrelated to your changes, post up which tests here, as there are a couple of problematic tests which fail occasionally.

nullthoughts:
All green on tests (woohoo!), and everything appears to functioning correctly in my application that utilizes Livewire.

Backwards compatibility: previously bookmarked URLs are being decoded and properties are being hydrated. In other words:
`?filters%5Bcondition%5D%5B0%5D=new&filters%5Bmake%5D%5B0%5D=Alfa%20Romeo` will transform to `?filters[condition][0]=new&filters[make][0]=Alfa+Romeo` in the browser

I've tried to think of edge case scenarios to add to the tests, however the query string tests are pretty comprehensive and cover the scenarios that have so far come to mind.

@joshhanley Should we re-run the workflow tests?

joshhanley:
@nullthoughts nice, good work! Yep done 🙂

nullthoughts:
> @nullthoughts nice, good work! Yep done 🙂

Thank you! The 7.3 and 7.4 Laravel 7 workflow tests: Dusk glitches possibly?

calebporzio:
All tests passing. Merging this. Thanks!

PR #3525, opened 2021-08-14
[Fix] Tag compiler to support snake cased attributes

joshhanley:
This PR ensures that Livewire's tag compiler doesn't convert snake_cased attributes to camelCase.

Currently if you have a public property on a component, such as `$item_no` and you use the blade directive syntax, to pass in data to it, it will work fine.

```blade
@livewire('sample-component', ['item_no' => $item_no'])
```

But that won't work using the blade tags as it is converting the attribute name to camel case.

```blade
<livewire:sample-component :item_no="$item_no" />
```

This PR fixes that by checking if the attribute name contains a `-` and if so, then convert to camel case.

Discussed in #3521.

Hope this helps!


nuernbergerA (approved):

calebporzio:
Thanks for the fix @joshhanley

I checked and Laravel's blade components camelCase all parameters. I think Livewire currently aligns with Laravel itself with this behavior.

As a user, I would almost intuit that Livewire would convert underscores to camelCased data properties automatically.

But I understand that folks can run into this and the error messaging isn't great.

What are your thoughts? Have we heard of others besides Ryan experiencing this? 

Would this also be a breaking change for anyone passing in snake_cased properties and using them as camelCased params? At first glance it seems like it.

joshhanley:
@calebporzio oh ok, I didn't know that's how blade components handle it. 

I guess the main issue from my perspective is the using `@livewire` supports underscores, but the tag syntax doesn't. They should at least be consistent.

I've seen it a couple of times, but it's pretty rare. I personally use snaked_cased on my eloquent models, but my properties in my LW class are all camelCased.

I couldn't imagine someone doing that, but that's not to say that there isn't someone doing that. In the scenario though, that'd definitely be a breaking change for them, which I hadn't considered.

Not sure then how we'd go about parity between the `@livewire` directive and the tag syntax, without a breaking change one way or another.

I guess the only way I could imagine getting this in now, would be as a config option, which by default leaves it how it is, but if enabled, allows the use of `snake_cased` properties.

We could also look at the error message and see if we can make that clearer instead, like if it said "property $itemNo" not found, then it would make it easier to debug what's going on.

Open to ideas, also understand bringing these into parity may have to wait for v3 if you don't want another config item. I will have a look at clearing up the error messaging if we don't proceed with this.

@ryatkins what's your thoughts?

calebporzio:
Good points @joshhanley 

Yeah, and I don't think perfect parity between the tag syntax and the blade directive is a must. I would guess that Laravel doesn't have that parity itself. I think there's a general expectation that HTML attributes shouldn't be camelCased, so there's also an expectation that there is a system to convert kebab to camelCase and I would imagine the same expectation applies to snake cased.

I think we should leave the functionality as-is and provide a better error message if you're up for it. Maybe also documentation to clarify - but again, I don't think we should touch the current behavior. Thanks!

ryatkins:
Thanks both for your time on this.

I 100% agree that better error messaging will solve this issue.  The current error in no way leads a user to believe that there was a mistake in the naming of their variable, but more in how they were trying to mount it.

calebporzio:
Ok, I think we should actually just support both snake_cased and snakeCased attributes.

(I committed this change to this PR)

For example, if you pass a parameter like this: `<livewire:... some-param="..."`, then both `someParam` and `some_param` will get passed in, so you can use either.

This way we don't break people's apps relying on the current behavior, but we provide something sensible and fix this for people encountering this in the future.

What do we think?

joshhanley:
@calebporzio nice! Now thats a solution, why not both! 🤷‍♂️ 😁

Is there any implications to having both there, other then there being two lots of properties available (only if snaked_cased is being used)?

I'm guessing the blade compiler will just ignore anything that isn't an actual property on the class when it populates it?

Otherwise I think it's a good solution!

calebporzio:
Yeah, that was my only thought too. The passing tests give me a lot of confidence though.

PR #3529, opened 2021-08-16
Add wire:duration for wire:loading

joshhanley:
**Update**

After looking at previous issues/ PR's on this, such as issue #69 and PR #153, the proposal was to have it be `.after.500ms` to make sure it's clear what it is doing.

What I proposed below with `wire:duration.500ms` might be too ambiguous, so I think it might actually be better to call it `wire:delay-duration.500ms` or something like that.

Let me know what you think, and I will update the PR.

---

**Original**

Currently `wire:loading.delay` is fixed at `200ms`.

The main reason I can see for this is, to set the default CSS styles to `display: none` for any element with `wire:loading`, we need to have a fixed attribute name so a css selector like `[wire:loading.delay]` works, as wildcard selectors don't work for attribute names.

See this line of code.

https://github.com/livewire/livewire/blob/4720f5abe8a9d1f89cec752497acbff09422a2d2/src/LivewireManager.php#L190

## Proposed workaround

I propose that we instead add a `wire:duration` directive that takes a duration, like `wire:duration.500ms`.

For any element that has `wire:loading.delay` on it, if you add `wire:duration.500ms` it would change the delay to `500ms`.

## Example

```blade
<div wire:loading.delay wire:duration.500ms>My loading div will show after 500ms</div>
```

if there was a way so we can support `wire:loading.delay.500ms` that would be preferred.

But this is the only way I can see around the attribute selector issue.

This PR adds the `wire:duration` functionality.

Open to other ideas/ names, etc.

Hope this helps!


js/component/LoadingStates.js
```diff
@@ -235,10 +236,20 @@ function doAndSetCallbackOnElToUndo(el, directive, doCallback, undoCallback) {
...
     if (directive.modifiers.includes('delay')) {
+        let directives = wireDirectives(el)
+
+        let duration = 200
+
+        if(directives.has('duration')) {
+            duration = directives.get('duration').durationOr(duration)
+        }
+
+        console.log(duration)
```
nuernbergerA, line 247: Leftover debug statement

calebporzio:
Yeah good call @joshhanley. You are right in the attribute name constraint and I totally see how just having the default with no customization isn't ideal.

Introducing `wire:duration` is kind of a big step however.

If I as a newcomer saw `wire:loading.delay` and then `wire:duration` directly after it - I wouldn't intuit that the duration applies to the delay. I would intuit that the duration is like a minimum loading spinner time.

Also, I would expect `wire:duration` to be able to be applied to `wire:dirty` and other similar attributes.

So like you mentioned, seeing that `duration` is too ambitious, `wire:delay-duration` would be specific enough to work for this case.

My issue here is that it feels like a divergence from the existing established patterns in the API.

A new user would expect to be able to chain the time after `wire:loading.delay` using dot notation and this would be a surprising unintuitive API for them.

Also it's a compound word and I believe all core directives so far are single words (which I oddly care about)

So all this to say. You're right, it's a shortcoming of Livewire's API caused by a technological constraint. In an ideal world, we think of some clever way to allow for the `.delay.250ms` - otherwise, I think we should punt on this until there is more demand for it?

Hit me with your thoughts? Am I being to idealistic?

js/component/LoadingStates.js
joshhanley: @nuernbergerA thanks! Have removed

joshhanley:
@calebporzio yeah honestly, I don't like it either haha. I'd much prefer it to be consistent with the established patterns.

And yeah agreed re the compound word, it did seem like a long directive which isn't ideal. I hadn't considered that all the rest are single, but do think it'd be nice to keep to that as much as possible if we can.

Also fun fact, I think you and I have different definitions of "punt" haha. Here that means to take a gamble, aka do it anyway. But when I double checked alternative meanings it can also mean to push down the road, aka look at it later. 😆

**New Idea**

So I have an idea, and I actually think it's a better solution, what if we had a `wire:cloak` which its only duty is to set display to none, while Livewire is booting.

Then we don't need to worry about `wire:loading.delay.500ms` not working as an attribute name, and when `loading` is processed on the element, then the `wire:cloak` would be removed and the loading stuff would control visibility.

Now so there isn't any breaking changes in V2, we would still leave the existing `wire:loading` styles, so people don't just have their loading elements randomly appear. Whether these styles should be removed in V3 would be debatable. 

But if a user wants to specify a custom loading duration, then that element must also contain `wire:cloak`.

So it would look like this

```blade
<div wire:loading.delay.500ms wire:cloak>This will show after 500ms but be hidden on render by wire:cloak</div>
```

I like it because it is then also consistent with Alpine.

Thoughts?

calebporzio:
Interesting. Yeah `wire:cloak` would solve the problem - however, I would like that behavior to be the default (not showing loading spinner blips). In other words, I wouldn't want to force everyone to add wire:cloak in addition to wire:loading every time they want to use it - that seems like a heavy tax all to allow for a more obscure API (the delay times).

Again, tough problem.

Like I'd sooner just support only a certain set of durations in the CSS than do the other options. Tough stuff - and I like where your heads at.

joshhanley:
Yep I don't want people to have to put `wire:cloak` on each time either, hence why leaving the existing `wire:loading` styles would be a good idea I think for when using any of the existing loading variations.

`wire:cloak` would then only be for when using something like the duration modifier.
 
Supporting a fixed set of durations might be a simpler solution though. If we went that route, I'd propose that we don't have any specific duration set (like `.500ms`) otherwise people my think other durations would be possible.

I personally think we'd pick 3-4 defaults and maybe put names against them instead like:
- `wire:loading.delay.xs` - 50ms
- `wire:loading.delay.short` - 100ms
- `wire:loading.delay` - 200ms // Existing one
- `wire:loading.delay.long` - 300ms
- `wire:loading.delay.xl` - 400ms

Something like that maybe?

I have thrown two different options above using like `.xl` but that makes me think that'd be on an `xl` screen due to Tailwind.
The other, using a name like `.long` sounds better, but what would you call the `xs` and `xl` versions, `extra-long` (not a fan of the hyphenation) or something else?

Personally, after typing the above out, I think I'd prefer just a few fixed options, as trying to get around the attribute name constraint to get full customisable durations isn't work the overhead. 

This would be simple to maintain, but gives some optionality. 

Just have to settle on some decent names 🙂

ingoldsby:
> * `wire:loading.delay.xs` - 50ms
> * `wire:loading.delay.short` - 100ms
> * `wire:loading.delay` - 200ms // Existing one
> * `wire:loading.delay.long` - 300ms
> * `wire:loading.delay.xl` - 400ms
> 
> Something like that maybe?
>  
> Personally, after typing the above out, I think I'd prefer just a few fixed options, as trying to get around the attribute name constraint to get full customisable durations isn't work the overhead.
> 
> This would be simpler to maintain, but gives some optionality.
> 
> Just have to settle on some decent names 🙂

What if you went with something like Tailwind CSS has with letter spacing, terminology-wise (https://tailwindcss.com/docs/letter-spacing#class-reference), so maybe:
* `wire:loading.delay.shortest` - 50ms
* `wire:loading.delay.short` - 100ms
* `wire:loading.delay` - 200ms
* `wire:loading.delay.long` - 300ms
* `wire:loading.delay.longer` - 400ms
* `wire:loading.delay.longest` - 500ms

joshhanley:
@ingoldsby thanks! I knew there was something there haha. I searched the TW docs a bit but couldn't remember what used those names!

Yep I like those names, keeps it clean and descriptive🙂

I think my preference (and for symmetry) would be this (have removed `longer` as there was no `shorter`):

* `wire:loading.delay.shortest` - 50ms
* `wire:loading.delay.short` - 100ms
* `wire:loading.delay` - 200ms
* `wire:loading.delay.long` - 300ms
* `wire:loading.delay.longest` - 400ms

@calebporzio let us know what you think and if you're happy with it, which options you want to support, and I can update the PR/ submit a new one for that.

calebporzio:
Great thinking here. I think the solid words are our savior here - and using direct Tailwind conventions makes it a much easier decision.

The question is to include all of them (`shorter/longer`) or just `short/shortest/long/longest`.

I'll leave that decision up to you. The only con of supporting all of them I think is the feature just seems larger.

If we can just get away with `.short` and `.long` for now then I'd say go with that, but I think that's probably not great because the scales would change if we had more words so we're not setting ourselves up to add more.

Ok, let's just go all the way on this.

Let's just make sure that `longest` is long enough - I think that's the only one we might regret not making longer. But I trust your information gathering on it so you judge for yourself.

joshhanley:
Ok so I have implemented this and it works great. 

I went with:
* `wire:loading.delay.shortest` - 50ms
* `wire:loading.delay.shorter` - 100ms
* `wire:loading.delay.short` - 150ms
* `wire:loading.delay` - 200ms
* `wire:loading.delay.long` - 300ms
* `wire:loading.delay.longer` - 400ms
* `wire:loading.delay.longest` - 500ms

Yeah, my only concern would the the "longest" one too as it locks that in as the upper limit.

But it's half a second, I can't imagine wanting to hold out showing the loading icon that long.

I think it should be ok, but I will do some research to see if I can find anything on ideal loading delays, before it's merged.

joshhanley:
I have found a few different articles on this, I will list two below.

## Research

https://www.smashingmagazine.com/2016/12/best-practices-for-animated-progress-indicators/#always-give-some-type-of-immediate-feedback

This article suggests two things:

Provide immediate feedback
> Any action, such as clicking a button or pulling to refresh, should have an immediate reaction. It’s essential to give some visual feedback immediately after receiving the request from the user to indicate that the process has initiated.

Use progress indicator for actions longer than one second
> When the app takes more than 0.1 second but less than one second to respond to user input, it feels like the app is causing the result to appear. Although users notice a short delay, they stay focused on their current task. After one second, the user’s attention begins to wander, and they notice that they’re waiting for a slow app to respond.
> 
> In order to reduce a user’s uncertainty, offer a reason to wait with a progress indicator for any action that takes longer than about one second (Note: the use of animation isn’t recommended for anything that takes less than one second to load, because users might feel anxious about what just flashed on the screen).

The **key thing to note** about that is the potential issue with providing a loading indicator for an action that takes less than one second.

This next article also supports that

https://pencilandpaper.io/articles/user-experience/ux-pattern-analysis-loading-feedback/#loading-patterns

> For anything that takes less than 1 second to load, it is distracting to use a looped animation, because users cannot keep up with what happened and might feel anxious about whatever flashed on the screen

> As soon as you pass the 1 second mark, the wait becomes noticeable to the user. Consider your options.

## Findings

So it may be worth offering a 1 second option.

If that's the case, I suggest bumping the long ones up to this:

* `wire:loading.delay.shortest` - 50ms
* `wire:loading.delay.shorter` - 100ms
* `wire:loading.delay.short` - 150ms
* `wire:loading.delay` - 200ms
* `wire:loading.delay.long` - 300ms
* `wire:loading.delay.longer` - ~~400ms~~ 500ms
* `wire:loading.delay.longest` - ~~500ms~~ 1000ms

Longer term (maybe V3), we can provide a loading JS hook, so developers can determine their own loading times, then it's not restricted by the CSS attribute selector.

ingoldsby:
I like your longest being 1000ms, then it does give a bit more variety to people

joshhanley:
@ingoldsby yep, I've thought more about it over the weekend, and I agree.

calebporzio:
I agree with this. Thanks for doing such good research @joshhanley !

PR #3530, opened 2021-08-16
Target model loading states

joshhanley:
This PR adds support for using `wire:target` with a property name, and it will show even when an attribute/ nested relation of that property is being updated.

For example, if we had a property `$user` with a keys `firstName` and `lastName` like this

```php
public $user = [
    'firstName' => 'Bob',
    'lastName' => 'Roberts',
];
```

And we had an input bound to `user.firstName` and another bound to `user.lastName`.

We could then setup an element  with `wire:loading` and add `wire:target="user"` to show loading state for any updates to `$user` property.

```blade
<input wire:model="user.firstName" />
<input wire:model="user.lastName" />

<div wire:loading wire:target="user">This will show for any modifications to $user property.</div>
```

Hope this helps!


amiranagram:
Hi @joshhanley, I was actually waiting for this, so thanks. Also can you check out this feature idea I posted a while ago? Is this something that other devs want, because I really do? Thanks again.

calebporzio:
Nice @joshhanley, I like this 👌 

Just to make sure this feature is grounded in real needs.

When is a common case for needing this?

Of course this is useful for having multiple forms in the same Livewire component for different models and wanting a loading indicator for each of them.

However, having not really encountered a good need for it before myself, I'd love a specific example of compelling use case if you have one.

Thanks

joshhanley:
@calebporzio yeah fair enough, happy to give some examples!

I've had need for this before when doing inline editing on table rows, I had a loading indicator at the end of each row.

Currently to work around it I had to extract the rows to a row component, so each row then could have its own loading state.

But my preference is to keep everything contained in a single component as much as possible, especially when dealing with tables.

It would also work in a scenario where if you had multiple filters all bound to a filters array, that you could just then have a single loading icon for when it's filtering and set it to `wire:target="filters"` then it won't matter which filter you are using, it'd still show.

I know @stancl also had a need for it and bought it up in #2425 but it doesn't have a use case.

Let me know what you think 🙂 

Also if you get a chance, can you have a look at this #3248 feature request raised above by @amiranagram. It's a very similar use case to my last one but a little different.

I'd be happy to look at implementing it if it's something you'd be willing to support.

Hope this helps!

stancl:
I think my use case was something like `wire:model="formData.foo.bar"` where essentially all of the user input is stored in that array, and other LW properties are used for different things.

calebporzio:
@joshhanley, different rows in a table is a good example.

My hesitancy with this behavior is adding more API constraints for future Livewire versions, however I think I'm probably good with this one.

As far as the except use case, that makes sense, but again, I would want to know there's sufficient demand before expanding the API.

"No is temporary, Yes is forever" - said someone smart

If we add behavior, we're locked into supporting it forever. If we reject feature additions we can always go back and add them when the demand is present, clear, and significant.

Both of these additions make me a little hesitant, but I'm definitely more keen on the first one you originally proposed in this PR.

stancl:
I think that in real world apps, array binding is actually very very common, so this specific feature just makes `wire:target` do what the developer intuitively expects it to do.

I agree with the scope creep thing, supporting the wrong things forever is definitely a risk.

joshhanley:
Yeah both have merit as they improve DX.

I think, of the two options, my preference would be this one as we aren't adding another directive modifier (like `.except`). 

As @stancl said, I think this would add something that seems intuitive.

But I understand if you want to hold off on it 🙂

calebporzio:
I'm down with this

joshhanley:
No worries! 🙂

stancl:
Thanks for merging Caleb!

PR #3556, opened 2021-08-22
Support collection re-ordering

calebporzio:
Currently, if you have a collection instance set as a public property on a Livewire component AND that collection has numeric keys AND those keys are out of numerical order in the array (wew), Livewire will throw one of those dreaded corrupt payload error things that are crazy hard to debug.

This PR fixes that.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -199,20 +199,6 @@ abstract class Component
         return $output;
     }
 
-    public function normalizePublicPropertiesForJavaScript()
-    {
-        foreach ($this->getPublicPropertiesDefinedBySubClass() as $key => $value) {
-            if (is_array($value)) {
-                $this->$key = $this->reindexArrayWithNumericKeysOtherwiseJavaScriptWillMessWithTheOrder($value);
-            }
-
-            if ($value instanceof EloquentCollection) {
-                // Preserve collection items order by reindexing underlying array.
-                $this->$key = $value->values();
-            }
-        }
-    }
-
     public function forgetComputed($key = null)
     {
         if (is_null($key)) {
--- a/src/HydrationMiddleware/NormalizeComponentPropertiesForJavaScript.php
+++ b/src/HydrationMiddleware/NormalizeComponentPropertiesForJavaScript.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\HydrationMiddleware;
 
+use Illuminate\Support\Collection;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 
 class NormalizeComponentPropertiesForJavaScript extends NormalizeDataForJavaScript implements HydrationMiddleware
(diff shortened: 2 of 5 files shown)


mertasan:
A very important/useful fix for those new to using Livewire. A long time ago I almost got cancer because of this exception.

Thanks @calebporzio ! :)

Those who want to see exactly what these fixes are can read @joshhanley's tweets.

https://twitter.com/_joshhanley/status/1420919521628475393?s=20

ingoldsby:
Just wanted to check if this is what is intended or whether it's a side-effect or I've missed something really simple. Using a new setup, with v2.5 and the following data as an example of a result after manipulating various pieces to get my desired result:
```
$this->data = collect([
    [
        'id' => 1,
        'total_points' => 44
    ],
    [
        'id' => 2,
        'total_points' => 182
    ],
    [
        'id' => 3,
        'total_points' => 13
    ]
])->sortByDesc('total_points');
```

Using v2.5 I get the desired output, with total points being from largest to smallest:
* Team - Total points
* 2 - 182
* 1 - 44
* 3 - 13

Changing composer.json to use master instead (to take into account this commit) and running the exact same code/view I get:
* Team - Total points
* 1 - 44
* 2 - 182
* 3 - 13

Now sorted by the team number (aka id) rather than the order I had set in my Livewire component.

Is this something simple I'm overlooking or with this change does it mean that sorting of the fields I do in the component will not be reflected in the blade? Will I need to do some sorting of the data in the blade instead to display it how I want it sorted?

mertasan:
@ingoldsby 

https://github.com/livewire/livewire/blob/6f7bd96ba48ab530c5deceeb81842081983dbbd6/src/HydrationMiddleware/NormalizeDataForJavaScript.php#L37

You can use `->values()`

```php
$this->data = collect([
    [
        'id' => 1,
        'total_points' => 44
    ],
    [
        'id' => 2,
        'total_points' => 182
    ],
    [
        'id' => 3,
        'total_points' => 13
    ]
])->sortByDesc('total_points')->values(); // with ->values()
```

Output:

```php

// Before:

Illuminate\Support\Collection {#698
  #items: array:3 [
    0 => array:2 [
      "id" => 2
      "total_points" => 182
    ]
    1 => array:2 [
      "id" => 1
      "total_points" => 44
    ]
    2 => array:2 [
      "id" => 3
      "total_points" => 13
    ]
  ]
}

// After:

Illuminate\Support\Collection {#695
  #items: array:3 [
    0 => array:2 [
      "id" => 2
      "total_points" => 182
    ]
    1 => array:2 [
      "id" => 1
      "total_points" => 44
    ]
    2 => array:2 [
      "id" => 3
      "total_points" => 13
    ]
  ]
}
```

ingoldsby:
Perfect, thank you @mertasan, glad it was something simple and confirmed it did resolve this for me.

Thank you! A straightforward update for me that will/has saved considerable time debugging going forward.

mertasan:
You're welcome 🎉 @ingoldsby 

@calebporzio Does this also need to be supported? Is it a expected behavior or bug?

PR #3572, opened 2021-08-23
Add with method to the redirect() helper

ariaieboy:
https://github.com/livewire/livewire/discussions/3068
No
Yes
in laravel request() global helper we can use with() method to add flash message to the redirect response this PR aim to add this feature to livewire request() helper


calebporzio:
Thanks! I've been wanting this for a while and never got around to supporting it.

PR #3575, opened 2021-08-24
Replace "testing" env check with native runningUnitTests check

roelofr:

I've encountered issue #2566 and read PR #2571, which caused confusion as to why the "testing"-env is forbidden.  That "testing"-magic-string is quickly mentioned in [the Laravel docs][1]) but it wasn't all that clear to me.

[1]: https://laravel.com/docs/8.x/testing#environment

Nope, just one change across several files.

No, it doesn't change anything functionally and should be covered by existing tests.

I've encountered the "disk not found" error on our DTAP-testing server (we used the names from the DTAP-pattern, which is wrong, apparently). 
Of course I grepped into the vendor folder looking for the disk name, and found the
disk-definition with the magic `app()->environment('testing')` string. 
This confused me, as I thought this was a convention thought up by Livewire.

By using a Laravel-native method which is very explicit about what it checks, the code quickly shifts the "blame" to the Laravel framework, where (if all goes well) the encountering dev will find the aforementioned paragraph.

Functionally, it doesn't do anything other than `app()->environment('testing')`, it just words it differently so one can quickly draw conclusions.

Thank you for your time 💜

### DIFF
--- a/src/Commands/MakeCommand.php
+++ b/src/Commands/MakeCommand.php
@@ -51,7 +51,7 @@ class MakeCommand extends FileManipulationCommand
                 $test && $this->line("<options=bold;fg=green>TEST:</>  {$this->parser->relativeTestPath()}");
             }
 
-            if ($showWelcomeMessage && ! app()->environment('testing')) {
+            if ($showWelcomeMessage && ! app()->runningUnitTests()) {
                 $this->writeWelcomeMessage();
             }
         }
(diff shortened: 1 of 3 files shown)


calebporzio:
Sounds good! Thanks!

calebporzio:
It's a little cleaner anyways than depending on a hardcoded string

PR #3585, opened 2021-08-24
Correctly get the S3 adapter when it is cached

ricki-io:

Yes, [discussion here](https://github.com/livewire/livewire/discussions/3573)

No

No

Livewire crashes when uploading a temporary file to s3 if cache is defined in `filesystems.php`


calebporzio:
Thanks!

PR #3596, opened 2021-08-27
validateOnly should only validate specific fields in a collection/array

### LINKED ISSUE #2591: validateOnly validates all matching fields in a collection

acurrieclark:
This pull request builds on an initial "failing test only" pull request which I submitted some months ago to attempt to resolve #2591

The current behaviour for the `validateOnly` method (when handling nested arrays) is that if we ask to validate eg. `foo.1.bar` then `foo.2.bar`, `foo.3.bar` etc will also be validated.

In my mind, this should be the expected behaviour if `foo.*.bar` is requested. If `foo.1.bar` is validated, then only that one field should be returned.

This pull request now means that:

`foo.1.bar` will validate **only** `foo.1.bar`
`foo.*.bar` will validate every `bar` field within the `foo` array
`foo.1` will validate every field within `foo.1`, including any `foo.*` rules
`foo.*` validates every item in the foo array, including nested items

Tests have been updated where necessary and new tests added.

I look forward to your feedback.


calebporzio:
Thank you for bringing this back up with a PR.

My head isn't fully wrapped around the problem being solved - just surface-level.

I agree about the premise that if you validate a specific nested field you wouldn't expect all it's siblings to be validated as well with `->validateOnly()`

That said, what kind of breaking changes would this cause for people if we pushed this?

calebporzio:
Merging, thanks @acurrieclark !

PR #3604, opened 2021-08-27
fix #3603

domthomas-dev:
don't replace sequentially but each occurrences

https://github.com/livewire/livewire/issues/3603


moxerox (approved):
tested, and it works as intended. it replaces each occurrence of <code>[view]</code> <code>[class]</code> <code>[namespace]</code>

calebporzio:
Awesome! Thanks for contributing the fix!

PR #3628, opened 2021-08-30
add toggle method in testable livewire

ph7jack:
I think it can be helpful for more users
Add a toggle method to change bool values in tests

Ex: 
`Livewire::test(MyComponent::class)->toggle('activated');`


calebporzio:
I'm in! Can you PR a docs addition for this too? https://github.com/livewire/docs

PH7-Jack:
@calebporzio of course. Thanks

Issue #1536, opened 2020-09-11
Pagination display error


nguyentranchung:
### Steps to reproduce
![livewire bug 1](https://user-images.githubusercontent.com/9611224/92830436-44d9cd00-f410-11ea-8395-527d258f343c.gif)

![image](https://user-images.githubusercontent.com/9611224/92830705-8b2f2c00-f410-11ea-8f2c-fc2bdd93207b.png)

**Context**
- Livewire version: 2.x
- Laravel version: 7.x
- Browser: Chrome

iksaku:
I'm experiencing the same display issue with pagination links

![image](https://user-images.githubusercontent.com/4632429/92974986-fdbdfd00-f44c-11ea-93aa-08ff9d56359d.png)

calebporzio:
Ok bootstrap pagination is now fixed in v2.1.1

Re-open if this is not the case for you

Thanks!

iksaku:
My issue was with the Tailwind one, but was fixed as well :+1:

nguyentranchung:
@calebporzio 
![image](https://user-images.githubusercontent.com/9611224/92997125-dbf76f80-f54b-11ea-86fa-0b31445a9aa3.png)
![image](https://user-images.githubusercontent.com/9611224/92997155-1a8d2a00-f54c-11ea-9c21-c8aba263d968.png)
This error still appears

nguyentranchung:
![image](https://user-images.githubusercontent.com/9611224/92997393-fa5e6a80-f54d-11ea-902c-fdff6816072b.png)

because render has span tag

nguyentranchung:
![image](https://user-images.githubusercontent.com/9611224/92997455-5e812e80-f54e-11ea-89fa-05852a4bd983.png)

nguyentranchung:
@calebporzio 
```blade
<div>
    @foreach ($posts as $post)
        <h1 wire:key="post-{{ $post->id }}">{{ $post->title }}</h1>
    @endforeach

    {{ $posts->links() }}
</div>
```
When testing pagination, you can try rendering both above and below, like: 
```blade
<div>
    {{ $posts->links() }}

    @foreach ($posts as $post)
        <h1 wire:key="post-{{ $post->id }}">{{ $post->title }}</h1>
    @endforeach

    {{ $posts->links() }}
</div>
```

PR #3631, opened 2021-08-31
[Fix] Two sets of links for one paginator morphdom issue

joshhanley:
This PR adds a fix for if you display two sets of pagination links for the one paginator on a page, for example if you have links above and below your table, raised in #1536 and #1582.

What is currently happening is morphdom is tripping up when there are two of them on the page due to them having the same `wire:key`, so each set of links needs its own identifier to pass into `wire:key`.

Laravel's paginator supports passing extra data to the view through the links method.

So I have added support for a property called `paginationKey` which can be passed in through the links method, like the below, and it will add it to the existing wire key in the pagination view.

```php
$posts->links(null, ['paginationKey' => 1]);
```

This ensures then if `$posts->links` is called twice, you can pass a different `paginationKey` into each one, which will keep morphdom happy.

Hope this helps!

Below is a screencast of the current issue with morphdom glitching.

recording (33)


calebporzio:
Good stuff @joshhanley - I like this. However, is there a way to do this automatically? So that a dev doesn't have to experience a bug, dig around and eventually find this "paginationKey" solution.

Like can we assign a unique key or something?

This may lead to a bigger question: We should have a way to generate a random key for this blade stuff in Livewire that is only random for the first request, then subsequent requests it is the same.

I think I may hack on this.

calebporzio:
After thinking about this a bit - providing a "Livewire aware random string generator" system would be pretty complex. I'd still like to explore it, but we should move on with this as if it doesn't exist.

Any ideas?

joshhanley:
@calebporzio yep, I'm not a fan if it either.

I couldn't think of an elegant way to populate an automatic key that was different for each instance of links, but that didn't trigger morphdom replacing of each instance every time a request is made.

Simple solution is to just put `uniqid()` into the `wire:key` but that'd cause it to be wiped out each time due to the change of key.

A random string generator that runs on first request also wouldn't work if the paginated data/links aren't displayed on first render either.

So one possible thing is, for the multiple paginators on a page PR, you can manually pass the different pagination names into the paginator and get it back out using `getPageName`, so I was going to use that to ensure keys are unique for different paginator links.

Maybe there might be a way to override the links method so it internally tracks how many times links has been called or something as uses that as part of the key? Not ideal either though.

calebporzio:
Mmmm yeah this is really tough. Decent ideas you've suggested The `->links()` is the most ideal from an API perspective, but not ideal from a maintenance one.

joshhanley:
@calebporzio yeah exactly. So the paginator classes are not macroable, so it I think it would be too much maintenance nightmare to try and support overloading the links method.

That's why my thought was just to pass a key through links provided they need it. It would be no different that needing to put `wire:key` on things, which if you don't, can cause odd things to happen.

I also didn't want to call it `wire:key` as that is predetermined inside the paginator view as it sets different keys for different buttons, hence this was more just an identifier to distinguish different sets of pagination links.

calebporzio:
@joshhanley 

I changed the implementation for this PR. If you hate it, just revert my commits.

I've switched to a strategy that doesn't require passing additional params to `->links(...)`

Instead, because every single render will render every single pagination link set, we can reliably store a property on the Livewire component instance that we can reference in `wire:key` AND increment before we use it each time.

I think this accomplishes our goal without developers having to do extra work.

I'll merge it for now, but again, feel free to back this out if you disagree.

PR #3640, opened 2021-09-01
Bind component view to component and make non-public properties and make methods available in view only

dimitri-koenig:
Currently only public properties and methods are accessable in view via $this.
With this PR the components scope to the view will be extended so that non-public properties and methods are available in the view also, without access from outside the view to still main security.

That way big chunks of render view parameters can be reduced, and some special logic can be extracted into usage of protected methods.

Could make https://github.com/archtechx/livewire-access obsolete


calebporzio:
@dimitri-koenig, this is a great PR. Makes total sense. If you are using `$this`, you would expect to be able to access protected stuff. Thanks for adding such good tests as well!

ryangjchandler:
@calebporzio This is technically a breaking change. Explicit property declarations will have precedence over computed properties, so something like this won't work:

```php
class Users extends Component
{
    protected $users;

    public function getUsersProperty()
    {
        return User::all();
    }

}
```

See #3768 for an example - I've also run into this myself.

calebporzio:
I see that the behavior has changed, but my gut says the new way is how you would actually expect Livewire to work and the old way was just a side effect of our bad `$this` implementation.

2 things support this way of thinking:
- In Vue, an actual component property would take precedence over a computed - I think it's generally assumed that making properties with the same name as a computed is a bad idea because of conflicts.
- Before 2.6, you would be able to access the computed on the view, but still not from the component class - which would feel like a weird decision.

Anyhow, you're right, it could break people's apps, but I almost consider this addition more of a bug fix. You expect that you can access public properties on a variable called `$this`.

If you disagree, I'm open to reverting - just don't want to jerk people around with adding this and then removing it you know?

PR #2977, opened 2021-06-16
Fix for undefined array key (500 error, ErrorException) when there is a property mismatch on livewire requests

ralphschindler:
1️⃣ Is this something that is wanted/needed? Did you create an issue / discussion about it first?

This is a bug fix. It attempts to be true to the spirit of Livewire in the fix delivered (permissive request responses, few errors unless a critical and unrecoverable error is encountered).

2️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

There is a single bug fix in this PR.

3️⃣ Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have)

There are currently no tests directory for the Response::toSubsequentResponse() method, but I would be willing to add one if needed.

4️⃣ Please include a thorough description of the improvement and reasons why it's useful.

In certain situations, the state of the initial render of a livewire component may become out of sync with the endpoint delivering the components subsequent state. This may be due to the initial page being full page cached at the edge, or could be the result of a browser aggressively caching certain pages, or could be a user initial rendered a page and the server software was updated.

In situations where properties were added to or removed from the Livewire component, the Response::toSubsequentResponse() method would be checking for a non-existent key which would in turn produce a PHP error of `ErrorException: Undefined array key "xxxx"`. This can, and arguably should, be mitigated in the Response method when checking for dirty state, this is the route this particular fix has taken.

5️⃣ Thanks for contributing! 🙌

https://user-images.githubusercontent.com/76674/122300940-a1997500-cec5-11eb-948e-7cb1b3887eed.mov

PR #3646, opened 2021-09-02
Don't error out if $key not found in memo

seliquity:
This error happens in instances where a property is added to a component, but a user still has an old copy of the page active. The solution would be to log everyone out, but this is tedious and disruptive in a production environment, especially when the actual change is not material enough that they should actually need to.

Given the error seems unintentional (this wasn't a problem in Livewire 2), I propose just guarding the comparison which seems to work fine and I don't think will have and unintended consequences.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -91,7 +91,7 @@ class Response
         // If 'data' is present in the response memo, diff it one level deep.
         if (isset($dirtyMemo['data']) && isset($requestMemo['data'])) {
             foreach ($dirtyMemo['data'] as $key => $value) {
-                if ($value === $requestMemo['data'][$key]) {
+                if (isset($requestMemo['data'][$key]) && $value === $requestMemo['data'][$key]) {
                     unset($dirtyMemo['data'][$key]);
                 }
             }


calebporzio:
Ah yeah, this is an issue with no simple solution for sure. It's been talked about before.

So this addition fixes your problem? This is the only place that throws an error when a piece of data is missing that was there before? or visa-versa?

seliquity:
This fixes our immediate problem yes, I'm not sure if it's the only place that throws an error as there's presumably multiple things that could happen (e.g. changing mount signature), but this fixes the main one we keep running into regarding adding new properties to components, and with this change we don't see regular errors like we see without it.

joshhanley:
@seliquity there is also another open PR #2977 also dealing with the same issue. Can you have a look at that PR and see if you have any ideas of how we can deal with it?

calebporzio:
Thanks @seliquity - like @joshhanley mentioned, there is another PR opened for this. I merged that one because it covers the other case as well (adding a new property after deploy), and it has tests. So this should be all set in the next release. Thanks!

PR #3663, opened 2021-09-06
fix Call to undefined method ReflectionUnionType::isBuiltin() exception

fabio-ivona:
This PR  fixes an exception thrown when union types are used in component method parameters

A component like this, would throw a `Call to undefined method ReflectionUnionType::isBuiltin()` exception

```
class ComponentWithUnionTypes extends Livewire\Component
{
    public $foo;
    public $bar;

    public function mount(Illuminate\Routing\UrlGenerator $generator, string|int $id = 123)
    {
        $this->foo = $generator->to("/some-url", $id);
        $this->bar = $id;
    }
   
   // ...
}
```

A detailed explanation was kindly made by @joshhanley  in #3635.

I used a trick to make the tests for this code to run only for php 8.0 and above. The test class is defined dynamically in the test case, otherwise would break the code as php7 doesn't parse union types at all


tests/Unit/ComponentDependencyInjectionTest.php
```diff
@@ -118,6 +118,51 @@ public function component_render_method_with_dependency()
...
     }
+
+    /** @test */
+    public function component_mount_action_with_primitive_union_types()
+    {
+        $php_version = phpversion();
+        $php_major_version = explode('.', $php_version)[0];
+        if ($php_major_version < 8) {
+            $this->markTestSkipped("Feature not available in php version < 8.0");
+        }
```
jrmajor, line 129:
```suggestion
    /**
     * @test
     * @requires PHP 8.0
     */
    public function component_mount_action_with_primitive_union_types()
    {
```

tests/Unit/ComponentDependencyInjectionTest.php
```diff
@@ -118,6 +118,51 @@ public function component_render_method_with_dependency()
...
+
+        eval('
+                class ComponentWithUnionTypes extends Livewire\Component
```
jrmajor, line 132: If you put this class in a separate file and run this test only on PHP 8.0, this file won't be loaded on lover versions.

jrmajor (changes requested):

tests/Unit/ComponentDependencyInjectionTest.php
fabio-ivona: @jrmajor you're right! thanks!

fabio-ivona: @jrmajor it doesn't seem to work, with `@requires php 8.0` the code is still parsed and throws an `ParseError: syntax error, unexpected '|', expecting variable (T_VARIABLE)` exception

any idea?

fabio-ivona: just found that the problem is with the `@requires php 8.0` part, so I managed to extract the class definition to an external file, but the php version checking is still there

jrmajor: Have you tried with proper casing (uppercase “PHP”)?

fabio-ivona: ok, I think it's fixed now

fabio-ivona:
@jrmajor workflows are failing because of dusk test failures: 
`JavaScript expression [return document.querySelector('[dusk="inner"]').style.opacity] mismatched.`

are there issues with dusk tests?

joshhanley:
@fabio-ivona have triggered the tests to run again so we can see 🙂 sometimes they are flaky

fabio-ivona:
> @fabio-ivona have triggered the tests to run again so we can see  sometimes they are flaky

nope, dusk keeps failing :(

joshhanley:
@fabio-ivona yep tests were being flaky for a few different PRs earlier. Seems ok now

fabio-ivona:
I see, great!

calebporzio:
Thanks so much @fabio-ivona!

tanthammar:
@fabio-ivona, @joshhanley Thank you for following through, though I admit being a bit sad that it ended up in returning "null". 
But I understand and respect the decision.

fabio-ivona:
> @fabio-ivona, @joshhanley Thank you for following through, though I admit being a bit sad that it ended up in returning "null".
> But I understand and respect the decision.

This is only a starting point to address the main issue, I think a followup could be taken into account ;)

again, thanks for your contribution!

PR #3666, opened 2021-09-06
Use strict comparison on childHasBeenRendered Fix #3665

allgood:
Yes

Only a single change

Not needed

It fixes a small #3665 

Thank you for this great tool!


calebporzio:
Good call! Thanks!

PR #3679, opened 2021-09-07
[Feature] Customize query params via component hydrate/dehydrate methods

nullthoughts:
Yes, this extends PR #3524 (which provided clean URLs) with the ability to customize how the parameters are generated from the component itself. Related discussions: #2036, #3030 and https://forum.laravel-livewire.com/t/ability-to-use-computed-properties-for-updatequerystring/236

No, this just focuses on giving users the ability to customize query params from within the component

No additional tests (yet), however current query-string related tests (and all tests at the time of writing) are passing.

Current query strings with an array:
`?filters[make][0]=Tesla&filters[make][1]=VW`

Query strings with PR after implementing `dehydratePropertiesToQueryParams()` & `hydratePropertiesFromQueryParams()` methods:
`?make=Tesla,VW`

This PR does not automatically flatten arrays, rather simply provides the ability to extend logic for defining the parameters. The above example was achieved by appending the following methods to the Livewire component:

```php

// Create query parameters from component properties (specified in $queryString)
public function dehydratePropertiesToQueryParams()
{
    $params = collect($this->queryString)->map(function ($param) {
        return collect($this->{$param})
            ->map(function ($value) {
                if (is_array($value)) {
                    return implode(',', $value);
                }

                return $value;
            })->reject(function ($value) {
                return empty($value);
            });
    });

    return $params->collapse();
}

// Fill component properties with query string values
public function hydratePropertiesFromQueryParams()
{
    collect($this->queryString)->each(function ($param) {
        if (is_array($this->{$param})) {
            return collect($this->{$param})->each(function ($value, $key) use ($param) {
                if (request()->has($key)) {
                    $this->{$param}[$key] = explode(',', request()->input($key));
                }
            });
        }
        
        if (request()->has($param)) {
            $this->{$param} = request()->input($param);
        }
    });
}
```

**This PR is a WIP** and opened for collaboration purposes. I believe there might be better placement options (in the lifecycle) for checking-for and calling both methods.

### DIFF
--- a/src/Features/SupportBrowserHistory.php
+++ b/src/Features/SupportBrowserHistory.php
@@ -149,8 +149,18 @@ class SupportBrowserHistory
         $this->mergedQueryParamsFromDehydratedComponents = collect(request()->query())
             ->merge($this->mergedQueryParamsFromDehydratedComponents)
             ->merge($componentParams)
-            ->reject(function ($value, $key) use ($componentParams) {
-                return empty($componentParams[$key]);
+            ->reject(function ($value, $key) use ($component, $componentParams) {
+                if (! isset($component->queryString)) {
+                    return false;
+                }
+
+                foreach ($component->queryString as $param) {
+                    if (is_array($param) && isset($param[$key])) {
+                        return empty($componentParams[$key]);
+                    }
+
+                    return empty($componentParams[$key]);
+                }
             })
             ->reject(function ($value, $key) use ($excepts) {
                 return isset($excepts[$key]) && $excepts[$key] === $value;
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for this @nullthoughts, for now I don't see a really clean API and the work involved and also the overhead of another custom API doesn't justify the need in my oppinion.

It is possible to achieve what you want I believe by setting a new public property who's sole purpose is providing the right data to the query string, and watching that property for changes in an `updated` hook in Livewire and then reacting accordingly.

I'm going to close this one, but if you think I'm wrong or have a better proposal, please re-submit with more context.

Thanks!

PR #3685, opened 2021-09-09
[Fix] query string on nested component

joshhanley:
Currently if you use query strings in a conditionally shown nested component, such as pagination, but don't have query string usage in the parent component, then the displayed URL is changed to `/livewire/message...` when the nested component is loaded.

This PR fixes that.

Hope this helps!


calebporzio:
Thanks @joshhanley !

PR #3689, opened 2021-09-09
Add component "boot" method

calebporzio:


francoism90:
@calebporzio Is this a hybrid between `mount` + `hydrate`? What exactly is the usage? :)

joshhanley:
@francoism90 yep, we found people were using the constructor to inject things like services that they wanted available on every request into protected properties that don't get sent to the front end. 

But it's not recommended to use the component constructor as Livewire uses it under the hood for other stuff.

So by adding a boot method, it is a method that will run on every requests, where as mount is only initial and hydrate is subsequent requests.

Hope this helps!

francoism90:
@joshhanley Seems Laravel News covered this nicely:
https://laravel-news.com/livewire-v2-6-0

Thanks for your reply!

joshhanley:
@francoism90 yep Caleb wrote it 😁 no worries 🙂

PR #3693, opened 2021-09-10
[Fix] property [$] not found on component

joshhanley:
Currently, if there is an empty `wire:model` on an input like `<input type="text" wire:model />` the below error appears.

Screen Shot 2021-09-10 at 1 48 43 pm

This PR adds a nice warning message in the browser console with the element listed for easy debugging.

Screen Shot 2021-09-10 at 2 09 26 pm

Hope this helps!


calebporzio:
Thanks @joshhanley!

PR #3684, opened 2021-09-09
[Feature] multiple paginators support

joshhanley:
This PR adds support for multiple paginators in one component.

Need I say more? 😁

Hope this helps!

PR #3697, opened 2021-09-10
[Feature] Query string hooks

amiranagram:
Not that I'm aware of. Also apologies for not creating a discussion for it myself. I recently saw many improvements and additions to Livewire's query string feature so I decided to contribute a small feature.

No.

Yes. Also all previous test pass. Note: I ran query string and pagination tests only.

First of all I'd like to acknowledge how far has Livewire gone in it's ability to allow modular development of Livewire components using traits. I specifically love the lifecycle hooks.

Anyway, this is a small feature that allows traits to have query string hooks as well. Currently, in components that use traits and those traits have some property that should sync to the query string, you'd have to collect all those properties and redefine them in each component. Specifically I don't like how `WithPagination` trait completely overrides `getQueryString` method, and I think I'm not the only one.

With this addition `getQueryString` method in `Component` merges query strings from each trait that has `queryStringTraitName` property or method (keeping it consistent with `queryString` that you define in your component).

That's why I renamed method `getQueryString` to `queryStringWithPagination` in `WithPagination` trait and removed the merging of query strings because now the `Component` class does that.

Going forward with modular development, and considering this feature we could have components like this one:

```php
class DataTable extends Component
{
    use WithPagination;
    use WithSearch;

    public $editingRow;

    protected $queryString = [
        'editingRow' => ['as' => 'row']
    ];
}

trait WithPagination
{
    protected $queryStringWithPagination = [
        'page' => ['except' => 1],
    ];
}

trait WithSearch
{
    protected $queryStringWithSearch = [
        'search' => ['except' => ''],
    ];
}
```

Livewire now merges all of them together into one nice array.

```php
[
    'editingRow' => ['as' => 'row'],
    'page' => ['except' => 1],
    'search' => ['except' => ''],
];
```

No, thank you for giving us Livewire ♥

P.S. I'm willing to contribute to Livewire docs as well if PR is warranted.


nuernbergerA:
I like that feature but it's a breaking change

amiranagram:
> 
> 
> I like that feature but it's a breaking change

What are we breaking, maybe I missed something?

nuernbergerA:
> What are we breaking, maybe I missed something?

If someone has `$queryString` in his class now it would overwrite everything from `Traits`.
After this PR it's not the case, so I would consider this as a breaking change.

But still :heart: this feature :wink:

amiranagram:
> 
> 
> > What are we breaking, maybe I missed something?
> 
> If someone has `$queryString` in his class now it would overwrite everything from `Traits`.
> After this PR it's not the case, so I would consider this as a breaking change.
> 
> But still ❤️ this feature 😉

Oops! Yes, that's true. I'll see if I can do something about it. Thanks.

amiranagram:
Hey @nuernbergerA, I've looked into this and #3684 already broke this feature. 

For example if you add `page` from pagination in your component class' `$queryString` property you aren't overriding it anymore.

```php
class DataTable extends Component
{
    use WithPagination;

    protected $queryString = [
        'page' => ['except' => 2]
    ];
}
```

In `2.5.5` it would be overridden, but not in `^2.6.0`.

My PR actually reverts it to the old behavior, and I've got tests to prove it. Or maybe the change was intentional. @joshhanley will tell us.

joshhanley:
@amiranagram or @nuernbergerA I'll be honest, I've been watching this PR, but I have no idea what you guys are talking about re the breaking change. 😂 

Could one of you please explain it to me?

amiranagram:
> If someone has $queryString in his class now it would overwrite everything from Traits.
> After this PR it's not the case, so I would consider this as a breaking change.

@nuernbergerA raised a concern that with this PR you aren't able to override query strings in `$queryString` property from your component class.

```php
class DataTable extends Component
{
    use WithPagination;

    protected $queryString = [
        'page' => ['except' => 2]
    ];
}
```

In this little snippet for example, `DataTable` overrides `page`. Now `page` query string won't show in URL if we are on page 2 of the paginator. After #3684 it is not the case anymore. `'except' => 1` from `WithPagination` trait stays `'except' => 1`. And that happens here:

https://github.com/livewire/livewire/blob/ad9af07a9c623c1aefbbcfcbd178adcac8ebb8ac/src/WithPagination.php#L25-L27

With this PR tho, it is not the case since I'm merging `$queryString` from component class after it collects all the query strings from Traits, like `WithTrait` used to do before #3684, here:

https://github.com/livewire/livewire/blob/97b8f3b157874ae209d646a03369c283d00ac3d9/src/WithPagination.php#L13-L20

joshhanley:
Thanks @amiranagram! Much appreciated 🙂 Yeah there is no way I would have seen that, spent too long staring at it trying to understand haha.

Yep that's a breaking change. I've got to patch a couple of things anyway, so I will amend my changes, so it behaves in the same way.

Thanks again!

P.s. if I'm being really honest, a lot of the code I just took from here #2727 😉

amiranagram:
No problem @joshhanley, with query hooks or not, at least we found a breaking change, so I'll mark this day as productive 😂

joshhanley:
@amiranagram haha yep exactly! Have tagged a release 2.6.4 with this fix 🙂

calebporzio:
Glad you guys found and identified a breaking change, I still feel like this is a good addition: Adding the ability for traits to add their own queryString items (without breaking parent component queryString items of course).

So if everything else is resolved I'm game for this addition.

joshhanley:
Yep agreed! 

@amiranagram  I think you've fixed the breaking change in the PR right? So it should be all good to go yeah?

amiranagram:
@joshhanley Yes, I merged master already and resolve everything on my end.

joshhanley:
Ok great thanks! 🙂

calebporzio:
Solid stuff, thanks!

calebporzio:
When you get a chance, can someone PR documentation for this as well? https://github.com/livewire/docs

amiranagram:
> When you get a chance, can someone PR documentation for this as well? https://github.com/livewire/docs

I can do that, no problem @calebporzio. 🙂

calebporzio:
Thanks!!

PR #3777, opened 2021-09-16
[Feature] Support 'withValidator' method

gdebrauwer:
Laravel's FormRequest has the option to [define a 'withValidator' method](https://laravel.com/docs/8.x/validation#adding-after-hooks-to-form-requests). This PR makes it possible to access the validator instance before the validation is run, so you can call any of its methods, just like the 'withValidator' method of Laravel's FormRequest.

```php
class FooBar extends Component
{
    protected function save(Validator $validator)
    {
        $this->withValidator(function (Validator $validator) {
            // acccess the validator instance

            // e.g. $validator->after(...);
            // e.g. $validator->sometimes(...)
        })->validate();
    }
}
```

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -19,6 +19,8 @@ trait ValidatesInput
 {
     protected $errorBag;
 
+    protected $withValidatorCallback;
+
     public function getErrorBag()
     {
         return $this->errorBag ?? new MessageBag;
@@ -144,6 +146,13 @@ trait ValidatesInput
         return ! $this->hasRuleFor($dotNotatedProperty);
     }
 
+    public function withValidator($callback) : self
+    {
+        $this->withValidatorCallback = $callback;
+
+        return $this;
+    }
+
     public function validate($rules = null, $messages = [], $attributes = [])
     {
         [$rules, $messages, $attributes] = $this->providedOrGlobalRulesMessagesAndAttributes($rules, $messages, $attributes);
@@ -158,8 +167,8 @@ trait ValidatesInput
 
         $validator = Validator::make($data, $rules, $messages, $attributes);
 
-        if (method_exists($this, 'withValidator')) {
-            $this->withValidator($validator);
+        if ($this->withValidatorCallback) {
+            call_user_func($this->withValidatorCallback, $validator);
         }
 
         $this->shortenModelAttributesInsideValidator($ruleKeysToShorten, $validator);
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -158,6 +158,10 @@ trait ValidatesInput
 
         $validator = Validator::make($data, $rules, $messages, $attributes);
 
+        if (method_exists($this, 'withValidator')) {
+            $this->withValidator($validator);
+        }
+
         $this->shortenModelAttributesInsideValidator($ruleKeysToShorten, $validator);
 
         $validatedData = $validator->validate();
@@ -193,7 +197,7 @@ trait ValidatesInput
             })->toArray();
 
         $ruleKeysForField = array_keys($rulesForField);
-        
+
         $data = $this->prepareForValidation(
             $this->getDataForValidation($rules)
         );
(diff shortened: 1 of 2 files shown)


joshhanley:
@gdebrauwer can you please give an example of how/ why you would use this in Livewire?

Thanks!

gdebrauwer:
In a component, I needed to add the `lte` validation rule to a key only if the other value was not null. This is only possible using the `$validator->sometimes(...)` method

```
$validator->sometimes('targets.*.min_build', 'lte:targets.*.max_build', function($input, $item) {
    return ! empty($item->max_build);
});
```

calebporzio:
Thanks @gdebrauwer, however I think to fully see the benefit we will need a full example component class. Please post an example component with properties, withValidation, and $this->validate() in it - thanks!

gdebrauwer:
Imagine you have an array of targets. A target has an optional min build and max build. You want to validate that min build is always less or equal to max build, and max build has to be greater than or equal to min build.

Because the min build and max build of a target can be null, the `lte` or `gte` validation rules will fail if either min or max build is null.

```php
class WithValidatorExampleComponent extends Component
{
    public array $targets = [
        [
            'min_build' => 100,
            'max_build' => null,
        ],
        [
            'min_build' => null,
            'max_build' => 100,
        ],
    ];

    public function runValidation()
    {
        $this->validate();
    }

    protected function rules()
    {
        return [
            'targets.*.min_build' => [
                'nullable',
                'integer',
                'lte:targets.*.max_build'
            ],
            'targets.*.max_build' => [
                'nullable',
                'integer',
                'gte:targets.*.min_build'
            ]
        ];
    }
}
```

To achieve this validation, you need to use the `$validator->sometimes()`, but then you need access to the `$validator`. The `withValidator` method makes this possible:

```php
class WithValidatorExampleComponent extends Component
{
    public array $targets = [
        [
            'min_build' => 100,
            'max_build' => null,
        ],
        [
            'min_build' => null,
            'max_build' => 100,
        ],
    ];

    public function runValidation()
    {
        $this->validate();
    }

    protected function rules()
    {
        return [
            'targets.*.min_build' => [
                'nullable',
                'integer',
            ],
            'targets.*.max_build' => [
                'nullable',
                'integer',
            ]
        ];
    }

    protected function withValidator(Validator $validator)
    {
        $validator->sometimes('targets.*.min_build', 'lte:targets.*.max_build', fn ($data, $item) => ! empty($item->max_build));
        $validator->sometimes('targets.*.max_build', 'gte:targets.*.min_build', fn ($data, $item) => ! empty($item->min_build));
    }
}
```

abordage:
another example:

```php
protected function withValidator(\Illuminate\Validation\Validator $validator): void
{
    $validator->after(
        function (\Illuminate\Validation\Validator $validator) {
        
            if ($validator->errors()->isNotEmpty()) {
                $this->reset(['password', 'password_confirmation']);
            }

            if ($validator->errors()->isNotEmpty()) {
                Log:info('form [xxx] validation errors', $validator->getData());
            }
            
            // something else
        }
    );
}
```

calebporzio:
Ok, I think I see. What about making it chainable like:
```php
$this->withValidator(function ($validator) { ... })->validate();
```

Then if you want it in a dedicated method you can do this:
```php
$this->withValidator([$this, 'doSomethingWithValidator])->validate();
```
???

gdebrauwer:
@calebporzio That is an excellent idea! I will update the PR

calebporzio:
Thanks @gdebrauwer, I think we're almost there. Two lil things:
- Can you add this for the `->validateOnly()` method as well? I think you will need to add the callback running bit to that method for it to work with that - right?
- Also, can you add a test for using ->withValidator with ->validateOnly as well?
- Ok, three things lol: can you remove the `: self` return type to stay consistent with the rest of Livewire?

After those things are fixed, we'll be good to merge. Thanks! 

- Ok, fourth (optional, but nice) thing: if you can PR documentation for this feature that'd be great too: https://github.com/livewire/docs

gdebrauwer:
@calebporzio Done; There's just one thing that I'm wondering suddenly: maybe we should unset the `$withValidatorCallback` property when it has been used? 🤔 In case multiple `validate`/`validateOnly` methods are run that might not all use the `withValidator` method

calebporzio:
@gdebrauwer - yeah I think that's a good call. When using it, it feels like `->withValidator` is linked to the subsequent `->render()` and wouldn't apply to others. Mind adding that? Thanks!

PR #3793, opened 2021-09-17
DisableBrowserCache support symfony response & skip otherwise

cappuc:
Right now `DisableBrowserCache` throw an exception if `$response` is not an instance of `Illuminate\Http\Response`.

With this change it also support the base `\Symfony\Component\HttpFoundation\Response` and it will skip the headers set if the response is something else.


joshhanley:
Thanks!

PR #3815, opened 2021-09-20
[Fix] PHP 8.1 compatibility

kylekatarnls:
reset() on object is deprecated (so is head())

I guess being PHP 8.1 compatible before it's released as stable is a must-have.

Single change.

Behavior does not change, maybe in a different PR, PHP 8.1 should be enabled for CI.

Running this provider with PHP 8.1 output:
```
reset(): Calling reset() on an object is deprecated
```
and it's indeed ambigous, here we're not actually picking the "head" of the object.

`iterator_to_array()` would be a relevant alternative.

You're welcome.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -243,7 +243,7 @@ class LivewireServiceProvider extends ServiceProvider
         // Early versions of Laravel 7.x don't have this method.
         if (method_exists(ComponentAttributeBag::class, 'macro')) {
             ComponentAttributeBag::macro('wire', function ($name) {
-                $entries = head($this->whereStartsWith('wire:'.$name));
+                $entries = $this->whereStartsWith('wire:'.$name)->first();
 
                 $directive = head(array_keys($entries));
                 $value = head(array_values($entries));
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -243,13 +243,12 @@ class LivewireServiceProvider extends ServiceProvider
         // Early versions of Laravel 7.x don't have this method.
         if (method_exists(ComponentAttributeBag::class, 'macro')) {
             ComponentAttributeBag::macro('wire', function ($name) {
-                $entries = iterator_to_array($this->whereStartsWith('wire:'.$name));
+                foreach ($this->whereStartsWith('wire:'.$name) as $directive => $value) {
+                    return new WireDirective($name, $directive, $value);
+                }
 
-                $directive = head(array_keys($entries));
-                $value = head(array_values($entries));
-
-                return new WireDirective($name, $directive, $value);
-            });
+                throw new InvalidArgument('Missing wire:'.$name.' attribute');
+            });;
         }
 
         View::mixin(new ViewMacros);
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -248,7 +248,7 @@ class LivewireServiceProvider extends ServiceProvider
                 }
 
                 throw new InvalidArgument('Missing wire:'.$name.' attribute');
-            });;
+            });
         }
 
         View::mixin(new ViewMacros);
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -243,7 +243,7 @@ class LivewireServiceProvider extends ServiceProvider
         // Early versions of Laravel 7.x don't have this method.
         if (method_exists(ComponentAttributeBag::class, 'macro')) {
             ComponentAttributeBag::macro('wire', function ($name) {
-                $entries = $this->whereStartsWith('wire:'.$name)->first();
+                $entries = iterator_to_array($this->whereStartsWith('wire:'.$name));
 
                 $directive = head(array_keys($entries));
                 $value = head(array_values($entries));


src/LivewireServiceProvider.php
```diff
@@ -243,7 +243,7 @@ protected function registerViewMacros()
         // Early versions of Laravel 7.x don't have this method.
         if (method_exists(ComponentAttributeBag::class, 'macro')) {
             ComponentAttributeBag::macro('wire', function ($name) {
-                $entries = head($this->whereStartsWith('wire:'.$name));
+                $entries = $this->whereStartsWith('wire:'.$name)->first();
```
kylekatarnls, line 246: Actually it looks like `head()` did not have the effect to take the first but just convert to array.
```suggestion
                $entries = iterator_to_array($this->whereStartsWith('wire:'.$name));
```

calebporzio:
Thanks for this, are you sure that `iterator_to_array` is a replacement for `head`? It seems that it just converts the object to an array rather than getting the first item of that array. I'm probably missing something - please explain - it seems like `->first()` would be a better alternative

kylekatarnls:
It's what I tried in my first commit, but then when trying it in a real project `$entries` was a `string` so I guess, `head` was actually never used to get the first item but to convert the bag into an array.

Then anyway, `head` is redundant as we do it again on the result array:
```php
                $directive = head(array_keys($entries));
                $value = head(array_values($entries));
```

There is very likely a cleaner (and more easy to understand) way to get the key-value pair from the `'wire:'.$name` param.

calebporzio:
Ok, mind refactoring this PR then? Thanks!

kylekatarnls:
You mean squash both commits? Can't you use the "Squash" merge option of GitHub?

kylekatarnls:
It could be:
```php
            ComponentAttributeBag::macro('wire', function ($name) {
                foreach ($this->whereStartsWith('wire:'.$name) as $directive => $value) {
                    return new WireDirective($name, $directive, $value);
                }

                // Should actually not happen
                throw new InvalidArgument('Missing wire:'.$name.' attribute');
            });
```

calebporzio:
Looks good to me!

src/LivewireServiceProvider.php
```diff
@@ -243,13 +243,12 @@ protected function registerViewMacros()
...
-                $directive = head(array_keys($entries));
-                $value = head(array_values($entries));
+                foreach ($this->whereStartsWith('wire:'.$name) as $directive => $value) {
+                    return new WireDirective($name, $directive, $value);
+                }
 
-                return new WireDirective($name, $directive, $value);
-            });
+                throw new InvalidArgument('Missing wire:'.$name.' attribute');
+            });;
```
kylekatarnls, line 251:
```suggestion
            });
```

kylekatarnls:
I tested this last version and it works fine in my Laravel project.

calebporzio:
Thank you!

joshhanley:
@kylekatarnls we've had to revert this PR for now as it was causing issues, see discussion #3864.

Upon further investigation, the `reset()` method will be marked as deprecated in php 8.1 and removed in php 9 https://wiki.php.net/rfc/deprecations_php_8_1#key_current_next_prev_reset_on_objects

But as we are making use of `head()` which is a Laravel method, we'd expect that they will potentially look into how to handle that closer to php 9 being released.

As such this should still all run on php 8.1 for now 🙂

If Laravel doesn't update the use of `head()`, we could also just call `->getAttributes()` on the component attribute bag which will return an array rather than an object, so then we could still make use of `head()`.

Hope this helps!

kylekatarnls:
@joshhanley be prepared for the possibility that Laravel will simply consider to propagate the deprecation as is: e.g. enforce user to pass an `array` to `head()`

The reason for the deprecation is good and the version at which it will be dropped does not matter. The same way we should not call `reset()` on an object, then we should not call `head()` on an object. If Laravel would "fix" it, it would actually defeat the purpose of this deprecation.

The code now (after the revert) is also sub-optimal as it proceed the whole iteration and call additional 2 `head()`, `array_keys()` and `array_values()`

So we should check the template code for #3864 and really fix it, rather than hoping that Laravel team will arrange every problems.

There is few possibilities:
1. simply fix the exception import
2. `return null`
3. `return new WireDirective($name, null, null)` (which should be equivalent to current behavior)

I think we can start with 3. to avoid regressions but I feel like having a void `WireDirective` is just putting something under the carpet.

kylekatarnls:
I started #3869 I will also try to add tests for this.

joshhanley:
@kylekatarnls yeah they could well do nothing. I guess to be clearer on what I was thinking is, Laravel are currently working on php 8.1 support, see this PR for example https://github.com/laravel/framework/pull/38961

I was just thinking wait until Laravel has full support for 8.1, then see what changes they have/have not made and work from there. That will then allow us to update github actions and work through any issues.

Regarding the return, the reason why adding the exception is a regression and causes issues, is because currently you can have the attributes `wire` call in a component, but not necessarily pass a `wire` attribute into the component and it works happily. If we add an exception though, it forces those attributes to be present, which would be a breaking change for some people.

For example this currently works, but would break with the exception:

input.blade.php
```blade
<input {{ $attributes->wire('model') }} />
```

```blade
<x-input />
```

Hope this helps!

PR #3856, opened 2021-09-24
Put Wireable at the top of the dehydration ladder

mabdullahsari:
Closes #3839

Defining anything as `Wireable` means "At this point, I got complete control. Leave everything to me."
It does not make sense to ignore this property on Eloquent models and all of the others that were checked before the `Wireable` check.


calebporzio:
Agreed

calebporzio:
Thanks!

PR #3869, opened 2021-09-26
[Fix] PHP 8.1 compatibility

kylekatarnls:
Alternative to #3815 without throwing exception (keep the current behavior)


calebporzio:
Thanks for being willing to contribute - however, the addition doesn't provide a tangible benefit to anyone right now and without a test (and given the issues with the last one) I don't feel comfortable merging this. Thanks again!

PR #3893, opened 2021-09-28
[Feature] Adds $readOnly reserved property

ryanshipswell:
I quickly discussed this on discord and confirmed no existing method to block client updates to a property exists short of using an updating hook.  At least one other person mentioned having had to use updating hook in the past to accomplish this.

This PR adds the ability to easily define one or more properties which should not be editable by the client and will block requests to do so with a simple conditional early return.

If you have ever wanted to use a property on a Livewire component to store a piece of state between requests that the user should not be able to edit you must block the edits for example by using a 'Updating' lifecycle hook. This adds the ability to instead define an array of properties that should be considered readonly to the client. 

Example usage:

```
    protected $readOnly = ['foo'];
    public $foo = 'Bar';
    public $bar = 'Baz';
```

By using this the client cannot update $foo, even if the user would intentionally send a `$set()` request. It can only be edited in the backend php.

Matching docs PR will be provided to document this new feature.

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -24,6 +24,8 @@ trait HandlesActions
             ($this->{$propertyName} instanceof Model || $this->{$propertyName} instanceof EloquentCollection) && $this->missingRuleFor($name),
             new CannotBindToModelDataWithoutValidationRuleException($name, $this::getName())
         );
+        
+        if (isset($this->readOnly) && in_array($propertyName, $this->readOnly)) return;
 
         $this->callBeforeAndAfterSyncHooks($name, $value, function ($name, $value) use ($propertyName, $rehash) {
             throw_unless(
--- a/tests/Unit/ComponentReadOnlyPropertyTest.php
+++ b/tests/Unit/ComponentReadOnlyPropertyTest.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Tests\Unit;
+
+use Livewire\Livewire;
+use Livewire\Component;
+
+class ComponentReadOnlyPropertyTest extends TestCase
+{
+    /** @test */
+    public function only_readonly_properties_on_a_livewire_component_cannot_be_set_by_the_client()
+    {
+        Livewire::test(ComponentWithReadOnlyProperty::class)
+            ->set('foo', 'Baz')
+            ->set('bar', 'Foo')
+            ->assertSet('foo', 'Bar')
+            ->assertSet('bar', 'Foo');
+    }
+}
+
+class ComponentWithReadOnlyProperty extends Component
+{
+    protected $readOnly = ['foo'];
+
+    public $foo = 'Bar';
+
+    public $bar = 'Baz';
+
+    public function render()
+    {
+        return view('null-view');
+    }
+}


amiranagram:
I've been using computed properties so far to achieve the read-only effect. I don't see a real need for this, but maybe I'm wrong.

x7ryan:
I'm not sure what benefit a computer property gives you that just using a protected property doesn't.  Either way it's not state that will persist. By using a real public property the state will be part of the payload and will be rehydrate on subsequent requests. This just makes it simpler to prevent client updates to those properties rather than having to use an updating hook.

joshhanley:
@x7ryan thanks for the PR 🙂

So I've had a couple of thoughts on it since you mentioned it on discord, I think this could be beneficial for a few things.

First is, does it work for model properties, like `$readOnly = [ 'user.name'];`? I think that's essential for it to work fully.

I'd also like to see it become a white list for allowing model properties to be sent to the front end.

I frequently send a collection of results to the front end for an Alpine powered select dropdown. Currently, I have to whitelist the properties of the collection that I want available to the front end using the `rules` property.

For example:

```php
$rules = [
    'users.*.id' => 'nullable',
    'users.*.name' => 'nullable',
];
```

But in reality they should be read only, therefore shouldn't be in the rules array, so by adding it as readonly properties, would not only clean up the rules array, but make the intent of that data clearer.

So I should be able to do:

```php
$readonly = [
    'users.*.id',
    'users.*.name',
];
```

I also think that maybe an exception should be thrown if they try to modify a readonly property, instead of it just silently returning.

Couple of little details on the PR, my first thought was it should be `$readonly` not `$readOnly`, but upon looking up the word in english it's spelt as "read-only", so now not sure haha. I think I prefer it without the uppercase "O" but that's open for discussion.

Also generally when you PR, it's best to create a separate branch in your fork of Livewire to do the PR on, as maintainers can then contribute to the PR. Where as when it's on your fork's master branch, they cannot contribute. But that can't be changed now without closing the PR and reopening a new one.

Hope this helps!

x7ryan:
Interesting thoughts. 

I believe looking back at it the property name I am grabbing to compare against `$readOnly` is it just looks at what is before the first dot. So you could for example make every part of `user` readonly but not say just `user.id`. It would be trivial to change it to grab the full name with all the dots but then I'm not sure that it would work with wildcards like `*` without having to get deep into parsing the property names which is beyond my comprehension of livewire atm as I haven't doven that deep in the source.

As far as becoming "a white list for allowing model properties to be sent to the front end" I am not sure I get what you mean here, Because afaik livewire sends every public property to the frontend. My goal here is to say I have this property which is tracking some state I want to save between requests so its a public property but I want to explicitly block any update requests to it so it can only be modified in the backend with php, not by a `$set()` request from the frontend (since that effectively can be done with a javascript request in the devtools console by a user).

I haven't considered use of an exception because my thinking was its not that we cannot do anything with the request because we can just ignore it but I guess I could see the point of providing that feedback though an exception.

As far at the capitalization goes I guess I can see it go either way I'm not particularly married to one over the other I cjust chose what felt right in the moment.

And sorry about not doing it in a branch I didn't think that would matter like that (GH did show an option to allow maintainers to edit the RP and that was checked so I figured that would work regardless of the branch) and since its a fork instead of the same repo a branch seemed redundant at that point but I will keep that in mind in the future. I think as a workaround for this PR though I suspect I could just add you or anyone else who wants to the forked repo manually to edit it there.

eric-hansen:
Just a thought, but why not follow other property conventions and have a getter as well?

I.e.: `protected $listeners == getListeners()`, so for this `protected $readOnly == getReadOnly()`

I'm also wondering if `$readOnly` is safe.  Another possible term (since "readOnly" can be used in various ways outside of this) is `$changeable` or something like that.

nuernbergerA:
I use read-only as a property name very often so this would be a breaking change to introduce a new reserved property.

x7ryan:
> Just a thought, but why not follow other property conventions and have a getter as well?
> 
> I.e.: `protected $listeners == getListeners()`, so for this `protected $readOnly == getReadOnly()`
> 
> I'm also wondering if `$readOnly` is safe. Another possible term (since "readOnly" can be used in various ways outside of this) is `$changeable` or something like that.

Honestly this is just an initial pass at the concept. I've never used getListeners myself so I never even realized that was a thing. So at this point I'm looking for feedback like that and I'm also looking to see what @calebporzio  has to say about the concept before putting much more effort into building this out.

As far as the name I'm not sure what would be a better option though I'm not opposed to changing it `$readOnly` seems thr most simplest and straight forward. `$changeable` dosent seem right because this is really the opposite this defines what is not changeable and imo `$readOnly` is a better naming convention than something like `$notChangeable`.

> I use read-only as a property name very often so this would be a breaking change to introduce a new reserved property.
> 
> In my case I toggle interactivity with the property.
> 
> I also wonder what's the use case to have a public property that should not be updated.
> 
> Why not using a protected property?
> 
> If the use case is to keep a state but prevent it from being modified we should find a better solution to separate property:
> 
>     * available in JS to modify (public)
> 
>     * keep server-side state (protected)
> 
> 
> Isn't that already the case?

Unfortunately thats not how that works. Protected or private properties are never even sent to the frontend so they cannot be used like that. And that would be an even bigger breaking change as currently Protected is used to define essentially component level config. Ie `protected $listeners` and `protected $rules` ect. 

Unfortunately being a breaking change to someone is a risk that is run by introducing a new reserved property if someone is already using that property name . The naming of it could change if someone has a better idea. Otherwise I'm not sure what to tell you about that.  That might make this have to wait for v3 which ive heard rumors is coming but idk.

joshhanley:
@nuernbergerA have a look at my use case I explained above. I frequently want to share data to the front end (for Alpine or straight JS), that I want accessible through `$wire` but not be writable.

A collection/array of data is a good example of this, I want it in Alpine to loop through, but that data may be variable, so accessing it through `$wire` is better than echoing it out into `x-data`. But I don't want that list to be editable, it's a reference list for a select/autocomplete dropdown.

But I agree regarding the breaking change. That's one area that's difficult with features like this, is anyone could create any property in their component which would collide with a future feature, that's the trade off for having a clean api for devs to use (rather than say using a verbose name like `$livewire_readonly`).

I think a trait would solve this issue the best as @x7ryan said, then people can opt in using `WithReadonly` for V2, and if it's popular, then can be potentially merged as a reserved property in V3.

Hope this helps!

calebporzio:
I see the need for the concept. I am however on the fence about the API for it. Maybe a `$readOnly` property is best, but it's something I think we should really think outside the box on before settling on that.

A property annotation could work:
image

It wouldn't cover the use-case of the nested properties though.

Tough one, maybe the property is the best thing here.

But to be clear, I'm not certain I'll merge this right now, just want to keep exploring.

x7ryan:
An annotation is an interesting idea, Admittedly I am not sure how to even go about detecting the annotation to apply it but I am sure I could figure it out. 

My thinking was this would mirror how you define listeners so it feels familiar. And while I don't personally have a use case for this on nested properties I could see how that could really be useful so I'm really not sure how else the API could look on this and still work like that. 

It would be nice though to not have to use a property because then it requires hiding behind a trait to avoid breaking changes.

walirazzaq:
> I see the need for the concept. I am however on the fence about the API for it. Maybe a `$readOnly` property is best, but it's something I think we should really think outside the box on before settling on that.
> 
> A property annotation could work: image
> 
> It wouldn't cover the use-case of the nested properties though.
> 
> Tough one, maybe the property is the best thing here.
> 
> But to be clear, I'm not certain I'll merge this right now, just want to keep exploring.

I am using a package by @stancl and API is something similar. It uses `PHP8 Attributes`. It even blocks or allows access to public methods on demand. Here is the link to package
[Livewire Access](https://github.com/archtechx/livewire-access)

stancl:
Yeah, my package solves this very easily.

To make a property readonly you just need to do:

```php
{
    use WithImplicitAccess;

    #[BlockFrontendAccess]
    public $open = false;
}
```

And any attempts to change the property from the frontend will be blocked, but the value will be accessible in the frontend payload. Also it'll be changeable from PHP logic, but just never *directly* from the frontend.

stancl:
I believe I talked about attributes with @calebporzio and he said that at least for this major version he wouldn't introduce them since they're a PHP 8 feature. But maybe in the next release Caleb can add some first-party solution inspired by my implementation.

Agree that this is a pretty common use case (I very often want some things to be stateful, but not changeable form the frontend), but I just use that package for mine because it's very low level and allows you to do anything you need.

A property like `$readOnly` only solves some use cases, but then you maybe want to use a whitelist not a blacklist solution, or you want to make methods only callable from PHP and not JS (`readonly` doesn't work for that anymore semantically, but the implementation is similar, both are `#[BlockFrontendAccess]` with my approach), etc.

x7ryan:
> 
> 
> I believe I talked about attributes with @calebporzio and he said that at least for this major version he wouldn't introduce them since they're a PHP 8 feature. But maybe in the next release Caleb can add some first-party solution inspired by my implementation.
> 
> Agree that this is a pretty common use case (I very often want some things to be stateful, but not changeable form the frontend), but I just use that package for mine because it's very low level and allows you to do anything you need.
> 
> A property like `$readOnly` only solves some use cases, but then you maybe want to use a whitelist not a blacklist solution, or you want to make methods only callable from PHP and not JS (`readonly` doesn't work for that anymore semantically, but the implementation is similar, both are `#[BlockFrontendAccess]` with my approach), etc.

The problem I see with an annotation is it only works with PHP 8+, idk that it has wide-enough adoption yet to force that as the minimum version for Livewire, in fact I would suspect Livewire would probably just defer to the current minimum required by Laravel itself.

I am not sure I understand how the property approach "only solves some use cases" what exactly is it not solving? On the contrary even if Livewire were to require a min on PHP 8, the annotation approach doesn't work for nested properties which it sounds like there may be some interest in the functionality of this specifically. 

As for the naming Something like `$FrontendReadOnly` is perhaps more apt but shortening it to `$readOnly` is just a convenience. IMO it's better than `BlockFrontendAccess` because that would imply the frontend can't see it at all which isn't the case.  I would rather have `$frontendReadOnly` or just simply `$readOnly` myself.

The more I think about it the more I am personally convinced this API is a better solution and also more consistent with the Livewire way, since there are other reserved properties to define things such as listeners.

stancl:
> I am not sure I understand how the property approach "only solves some use cases" what exactly is it not solving?

I meant the:
- method access
- whitelist approach instead of a blacklist approach

Note that I'm not saying that this shouldn't be implemented, I haven't had much time to look at the PR anyway. Just sharing why I like the solution I use, and that if this were merged it still wouldn't change how Lean controls access to properties.

x7ryan:
> 
> 
> > I am not sure I understand how the property approach "only solves some use cases" what exactly is it not solving?
> 
> I meant the:
> 
>     * method access
> 
>     * whitelist approach instead of a blacklist approach
> 
> 
> Note that I'm not saying that this shouldn't be implemented, I haven't had much time to look at the PR anyway. Just sharing why I like the solution I use, and that if this were merged it still wouldn't change how Lean controls access to properties.

I still don't understand what you mean. It seems to me that whether you use an attribute or property they would work 100% the same, it only changes the semantics of how you define it.

Issue #2878, opened 2021-05-16
Intermittent unresolvable dependency when used with Laravel Octane


georgesa:
### Description

I get this error from time to time only with Livewire components. Not sure how to reproduce it or what causes it.

```
Illuminate\Contracts\Container\BindingResolutionException
Unresolvable dependency resolving [Parameter #0 [ <required> $methods ]] in class Illuminate\Routing\Route
```

**Context**
- Livewire version: 2.4.4
- Laravel version: 8
- Octane version: 1.0.0
- Browser: All

richan-fongdasen:
I also get the same intermittent error. At first, switching between Livewire routes didn't produce any error. The error occurs mostly when I move to a different Livewire route after updating the Livewire component's state a few times on the previous page.

**Context**

- Livewire version: 2.4.4
- Laravel framework version: 8.41.0
- Octane version: 1.0.0 (swoole)

PR #3899, opened 2021-09-29
[bug] Octane has issues with injecting $route

rennokki:
I have a component that got updated and on the next request, the following error was thrown:

```
Unresolvable dependency resolving [Parameter #0 [ <required> $methods ]] in class Illuminate\Routing\Route
```

There are a [lot of references](https://github.com/livewire/livewire/search?q=Octane), but it seems like [one commit removed the binding](https://github.com/livewire/livewire/commit/da335efeb41c004411685709d7733303177ac88e), but [it got reverted](https://github.com/livewire/livewire/commit/ad9af07a9c623c1aefbbcfcbd178adcac8ebb8ac).

There are injection issues in general with Octane if you're not careful, and even if it is useless, should not be injected. 🐱 

References: https://github.com/livewire/livewire/issues/2878, https://github.com/livewire/livewire/pull/3495

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -42,7 +42,7 @@ abstract class Component
         Livewire::setBackButtonCache();
     }
 
-    public function __invoke(Container $container, Route $route)
+    public function __invoke(Container $container)
     {
         // For some reason Octane doesn't play nice with the injected $route.
         // We need to override it here. However, we can't remove the actual
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -86,7 +86,7 @@ class LivewireServiceProvider extends ServiceProvider
 
     protected function registerLivewireSingleton()
     {
-        $this->app->bind(LivewireManager::class);
+        $this->app->singleton(LivewireManager::class);
 
         $this->app->alias(LivewireManager::class, 'livewire');
     }
@@ -102,12 +102,12 @@ class LivewireServiceProvider extends ServiceProvider
             ? '/tmp/storage/bootstrap/cache/livewire-components.php'
             : app()->bootstrapPath('cache/livewire-components.php');
 
-        $this->app->bind(LivewireComponentsFinder::class, function ($app) use ($defaultManifestPath) {
+        $this->app->singleton(LivewireComponentsFinder::class, function () use ($defaultManifestPath) {
             return new LivewireComponentsFinder(
                 new Filesystem,
-                $app['config']->get('livewire.manifest_path') ?: $defaultManifestPath,
+                config('livewire.manifest_path') ?: $defaultManifestPath,
                 ComponentParser::generatePathFromNamespace(
-                    $app['config']->get('livewire.class_namespace')
+                    config('livewire.class_namespace')
                 )
             );
         });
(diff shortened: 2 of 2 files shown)


joshhanley:
@rennokki the original PR was reverted as it was a breaking change, so we can't change the signature of `__invoke`, hence why Caleb changed it back, so I don't think this will be merged.

Does the new update of using the route from the request helper not work? Any other ideas how we can solve it without removing the parameter from the method signature?

I've not used Octane, but does this issue happen a lot? Can you provide some detail around when the error is triggered?

Thanks!

rennokki:
I currently have installed Livewire through composer via the patched fork I made to make it work.

The issue here is not really clear, but so far what I tracked, is related to the Component class. Whenever the component gets updated with `wire:click`, changing the route will trigger this issue. 

I still have to run some tests to get below the surface and find this bug. 🤔

joshhanley:
@rennokki ah ok, yeah I was wondering whether it was actually Livewire's internal route checker that might be causing the issue. 

Have a look at https://github.com/livewire/livewire/blob/master/src/Controllers/HttpConnectionHandler.php

It dummy loads the original route to get the middleware applied to it, so the persistant middleware can be applied. Maybe it's causing the issue.

If you have an example repo that demonstrates the issue, with some instructions to get it installed and running with Octane, I'd be happy to pull it down and have a play.

Hope this helps!

rennokki:
Currently, the repo I work on is closed source, but I can make up a demo for it. In the meanwhile, I suspect it's about registering the route managers through the app. Somehow, Octane has issues with this when it comes to booting up the application due to really fast wind-up.

[Octane warns about that and this is causing the issue](https://laravel.com/docs/8.x/octane#container-injection). Perhaps the Route instance gets modified somehow.

rennokki:
I'm dying 💀 [Persistent middleware](https://github.com/livewire/livewire/blob/master/src/Controllers/HttpConnectionHandler.php#L16) is causing this issue.

rennokki:
Not sure what this persistent middleware does, but I assume that we can ignore it in Octane? 🤔

rennokki:
This is the final message from this monologue. [App doesn't like this pipeline](https://github.com/livewire/livewire/blob/master/src/Controllers/HttpConnectionHandler.php#L55-L60).

joshhanley:
@rennokki yeah I thought that might be the issue.

Have a look at the persistant middleware section on this docs page, hopefully it clears it up a bit what it's for https://laravel-livewire.com/docs/2.x/security

I'm not sure if we can ignore it with Octane, but that code is not public api (unlike the `__invoke` method on the component), so we might be able to modify it to support Octane.

rennokki:
Honestly, I don't know how to approach the problem. It seems like piping through the `app()` (which is called correctly here) is causing the issues.

joshhanley:
@rennokki yeah I'm not sure. If you get the time to put a small demo repo together, I'd be happy to have a look and see if I can find anything 🙂

joshhanley:
@rennokki I had a play around with Octane today, and I couldn't seem to replicate the issue.

I tried all sorts of things including using `wire:click` to modify components, dynamically add and remove nested components, trigger redirects and event dynamically change the uri.

Nothing I tried triggered the error. So yeah if you can provide an example that would be great 🙂

luischavez:
Fix: #3936

joshhanley:
@rennokki this has been fixed in #3946, so I'm going to close this PR. If you have any further issues with it, please let us know.

Thanks!

PR #3901, opened 2021-09-29
[Feature] Allow filling a model using a factory when testing

chrisrhymes:
I created a discussion about this. The idea is to make it easier to use factories to populate many properties from a factory at the same time for testing. 
https://github.com/livewire/livewire/discussions/3892

Single change

Yes, single test

Rather than having to set each property manually you can fill the model properties with a single `fillFactory()` call.

```php
//Before
$user = User::factory()->make();

   ->set('user.name', $user->name),
   ->set('user.email', $user->email);

//After
$user = User::factory()->make();

   ->fillFactory($user, 'user');
```

It also offers the ability to specify fields you don't want to exclude from the factory, by passing in a third argument (an array).
```php
$user = User::factory()->make();

   ->fillFactory($user, 'user', ['password'])
   ->assertNotSet('user.password');
```

### DIFF
--- a/src/Testing/Concerns/MakesCallsToComponent.php
+++ b/src/Testing/Concerns/MakesCallsToComponent.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Testing\Concerns;
 
+use Illuminate\Support\Arr;
 use function Livewire\str;
 use Illuminate\Support\Str;
 use Illuminate\Http\UploadedFile;
@@ -46,6 +47,17 @@ trait MakesCallsToComponent
     {
         foreach ($values as $name => $value) {
             $this->set($name, $value);
+        }
+
+        return $this;
+    }
+
+    public function fillFactory($modelFactory, $prefix, $except = [])
+    {
+        $factory = Arr::except($modelFactory->toArray(), $except);
+
+        foreach ($factory as $name => $value) {
+            $this->set("$prefix.$name", $value);
         }
 
         return $this;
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for the PR. I don't think I'm going to merge this though. I think I'd suggest the following:

```php
->fill($user->toArray())
```

or if you want a subset, something like:
```php
->fill($user->pluck(['name', 'email']))
```

chrisrhymes:
Thanks for your feedback. I have tried a few different ways of doing this but not making much progress. Sorry if there is something really obvious I am missing. 

When I try the below I get a PropertyNotFoundException as the property [$name] is not found on the component, as it is `user.name` rather than `name` due to [binding directly to model properties](https://laravel-livewire.com/docs/2.x/properties#binding-models)

```php
->fill($user->toArray())
```

If I try to fill using the array key of 'user' I get a CannotBindToModelDataWithoutValidationRuleException as there isn't a validation rule for user. 

```php
->fill(['user' => $user->toArray()])
```

If I add 'user' to the validation rules and try and fill user, it then errors when I try to save the user with `no such column: user` as it adds `user.user` to the user property.

```php
// In the component class
protected $rules = [
     'user' => '',
     'user.name' => 'required',
     'user.email' => 'required|email',
];
```

```php
->fill(['user' => $user)])
```

`dd($this->user)` in the component gives the following with the `user` added to the end of the attributes. 

```
#attributes: array: 9 [
    "id" => 1
    "name" => "Example User"
     ... etc.
    "user" => null
]
```

Thanks.

PR #3903, opened 2021-09-29
[Feature] Removing the limit on uploads directly to S3

coleshirley:
Hi I am not sure if this is something that is already in the works for a future update but I couldn't find it anywhere on the roadmap or projects so I thought I would contribute. I need the functionality to upload multiple files to our S3 storage system for my projects so if this pull request doesn't get accepted then it wasn't wasted time for me.

### Summary
The gist of the changes I made are that now when a user adds multiple files to be uploaded and their temporary upload directory is configured to an S3 storage system livewire will create multiple temporary upload links and send them back to the users browser. The js implementation then uploads a single file per temporary upload URL and waits for all of those requests to finish before firing off the `uploadFinish` or `uploadError` events.

### Limitations
One limitation of my implementation is that if only one of the uploads fails then the whole set of files will be discarded instead of just the single failed file. However this is how multiple uploads directly to any other storage backend work so it doesn't break the event api. Ideally I would like to have a system in place where each file could fail or succeed individually but that would require several breaking changes to the project.

### Tests
I have not modified or added any test cases in the project because I figured I would submit this pull request first before doing so. If my implementation is appreciated then I am happy to modify and add test cases to the package before finalizing this pull request


calebporzio:
Just chiming in here because this PR has been left unattended for so long.

My overall take on this is:
* Yes, seems like a valid problem and valid fix
* There are no tests for it however, which makes sense because testing S3 is non-trivial and we don't have any tests for it for anything else. This means that I can't just merge this with confidence without testing everything myself.
* On top of that, there are enough deep code changes that it's difficult to understand just by reading it

All this to say, I'm not comfortable merging this PR as is, so I'm going to close it with the following thoughts:
* To re-submit this PR, you will need to submit a branch from your fork, not `master` so that we can run the tests, contribute ourselves to it, and resolve merge conflicts
* After I close this, if you've since come up with a local workaround for others in the meantime, please comment on this PR with that.

This is a deep one, thanks for putting in the effort.

PR #3916, opened 2021-09-30
[Feature] add emit down option for events

coleshirley:
My use case for this is pretty simple: `ParentA` and `ParentB` are sibling livewire components on the same page. Each of them has a nested `Uploader` component. When `Uploader-A` finishes it emits an event upwards to whichever parent component that allows that parent to register the upload as done. Then when I want to save the files I hit a button on `ParentA` that emits an event to the `Uploader` component to run through the save process. Currently in this workflow the event emitted from `ParentA` triggers `Uploader-A` and `Uploader-B` when it should only trigger `Uploader-A`.

This is how the [spatie/laravel-medialibrary-pro](https://spatie.be/docs/laravel-medialibrary/v9/handling-uploads-with-media-library-pro/handling-uploads-with-livewire) package works. To fix this and avoid unnecessary requests I propose a `emitDown()` function that can then be utilized in that package. This feature was also requested here in this discussion: https://github.com/livewire/livewire/discussions/1912

This will provide more robust event handling options for the users of livewire


calebporzio:
Hi @coleshirley, thanks for the well done PR! I understand your use case with this.

Here's my only hangup: this would be the first emit functionality that couldn't be mimicked by native DOM events (DOM events can only dispatch up, never down a tree like this does). So if we rewrite the JavaScript portion of Livewire (which we plan to), our creativity in solving the event transmission stuff is limited to the current strategy.

Probably not a big deal, just something that I want to think about before expanding this scope.

Are there other ways you can think of to accomplish this right now?

@joshhanley do you have thoughts on this?

coleshirley:
> Are there other ways you can think of to accomplish this right now?

@calebporzio I have seen some other issues/PRs that reference updating the `emitTo` handler or adding a similar handler to allow events to be dispatched to a specific instance of a component on a page instead of to all components of one type. That would allow for my use case to work I would just scope the event from `ParentA` to target `Uploader-A`. I'd be willing to take a look at implementing that functionality. Before I would love some ideas on the best way to go about it. i.e. should I update the `emitTo` event or add some other function like `emitToWithSelector`. Either way I image the developers implementation could be something like so `$this->emitTo('uploader', '[id="uploader-a"]', $data)` where you pass a dom selector as the condition.

coleshirley:
@joshhanley That is a good idea. I will make a pull request to the uploader project that incorporates your "scoped event" suggestion.

@calebporzio Is there any interest in my scoped `emitTo` dom selector suggestion? I'm happy to submit another pull request if there is.

calebporzio:
Ok, I'm going to close this in favor of making `emitTo` accept a component ID.

If someone wants to PR that, here's what I think you should do:

* Method signature: `->emitTo([component name | comonent id], ...args)`
* in the implementation, first loop through all Livewire components rendered on the page (by looking through `Store.js` and the `componentsById` property)
* if there is no match for the passed in `[name | id]` then we know it's not an ID and is instead a component name, in this case just resume the current implementation

In addition to PRing that functionality, please PR code snippets demoing how a user would consume this to make sure it's a pleasant experience to pass in another component's ID from a different component.

Thanks!

PR #3929, opened 2021-10-02
[Fix] Added app_url config value - fixing CDN url problem

Lukasz93P:

I think yes, it fixes problem with app url when serving livewire scripts from CDN.
https://github.com/livewire/livewire/discussions/3928
https://github.com/livewire/livewire/discussions/3108

No.

It includes required changes in existing tests and one new test case.

Like described in point 1 - without this fix livewire served from CDN tries to communicate with application (ex POST requests) using CDN url instead of app url.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -58,20 +58,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Livewire App URL
-    |--------------------------------------------------------------------------
-    |
-    | This value should be used if livewire assets are served from CDN.
-    | Livewire will communicate with an app through this url.
-    |
-    | Examples: "https://my-app.com", "myurl.com/app".
-    |
-    */
-
-    'app_url' => null,
-
-    /*
-    |--------------------------------------------------------------------------
     | Livewire Endpoint Middleware Group
     |--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Lukasz93P:
Hi, @joshhanley could you please look at it in the near future? It's priority for me.

Lukasz93P:
@calebporzio can somebody finally look at this? Newer request has been commented while this is still waiting.

Lukasz93P:
@joshhanley Here it is - with app.url (Laravel config) usage.

Lukasz93P:
@joshhanley @calebporzio how long can it take? My app production release is affected by that because I also use other packages which requires livewire and I don't want to fork all those repos.

Lukasz93P:
> @joshhanley yes, if `assets_url` points to `https://cdn.myapp.com/` it won't work, because LiveWire will try to communicate with routes under this domain (CDN domain) instead of app server domain.
> 
> Why I'm not used Laravel `app_url`? - I could it it would work but I wanted to preserve backward compatibility and not change settings of existing apps without package users knowledge.

@calebporzio  I was even talking about this...

Now I reverted last commits and it's just like you have described it - now backward compatibility is preserved.

Issue #2720, opened 2021-04-13
Octane - computed properties don't get reset between requests


lalov:
### Description
When using computed properties on a component, the method that computes the property only seems to execute once per swoole instance. 

### A fix?
This seems to fix it, but I don't know where to put it in a pull request: 

<pre>
        Livewire::listen('component.hydrate', function ($component, $request) {
            $component->forgetComputed();
        });
</pre>

### Exact steps to reproduce
Create a component with a computed property, and load the component from a route. 
Open the page so that the computed property gets different values, and reload the page multiple times. 
The page will show "old" values in random order.

**Context**
- Livewire version: 2.4.2
- Laravel version: 8.37.0

calebporzio:
Perfect fix, thanks! 

If you could PR this that'd be great.

I'd put it in a dedicated hydration middleware and reference in the service provider.

Hope that helps, thanks again!

joshhanley:
@calebporzio interesting that the computed properties aren't cleared.

I haven't played with it yet, but does that mean the components don't get destroyed at the end of a lifecycle? 

I wonder what other side effects there are if thats the case? Should the new middleware actually deconstruct a component when the lifecycle is finished, or is there a benefit that it doesn't get destroyed?

lalov:
@joshhanley I checked, and the constructor for the component only runs for the first requests, so it doesn't get destroyed. 
I think it matters that the component is run from as a full-page-component.

lalov:
Another way to fix it is to reset computed properties in the __invoke() method in Livewire/Component

<pre>

public function __invoke(Container $container, Route $route)
{
    $this->computedPropertyCache = [];

    $componentParams = (new ImplicitRouteBinding($container))
    ->resolveAllParameters($route, $this);

</pre>

joshhanley:
@lalov ah right, thanks for the info 🙂 

Hmm so yeah I do wonder then whether it would be better to add a middleware that actually deconstructs the components rather than leave them running. I could imagine issues if data is assigned to any protected/ private properties on a component.

That would then also deal with the computed property issue.

calebporzio:
I agree Josh. Components should be destroyed after we're done with them.

lalov:
I agree, it would be good to deconstruct the components. I dont think I'll be able to make the PR...

ju5t:
I can help with the PR if someone could give me some pointers where to look for :)

This might solve an issue I'm having with query strings leaking to another component.

joshhanley:
[...]

PR #3936, opened 2021-10-03
Better Octane support

luischavez:
Hi, everyone I'm very excited with octane so I'm trying to get a stable release of Livewire that works perfectly with octane, this MR contains a few changes to solve some of the common octane problem that I found, here you have a brief description:

1. Fixed properties and computed properties in full page component that preserves along requests (Moved to #3947).
2. Fixed __invoke Route parameter binding (Moved to #3946).
3. Fixed memory issues with some of the clases in Features folder (Moved to #3940).

### Fixed properties and computed properties in full page component that preserves along requests.
The Illuminate/Routing/Route keeps in memory an instance of the component:
[https://github.com/laravel/framework/blob/8.x/src/Illuminate/Routing/Route.php#L271](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Routing/Route.php#L271)
```PHP
  /**
     * Get the controller instance for the route.
     *
     * @return mixed
     */
    public function getController()
    {
        if (! $this->controller) {
            $class = $this->parseControllerCallback()[0];

            $this->controller = $this->container->make(ltrim($class, '\\'));
        }

        return $this->controller;
    }
```
so we need to create always a fresh instance of the component on the __invoke method.

### Fixed __invoke Route parameter binding.
This was really hard to found and to be honest I really don't know why is this causing the issue, but if you not set the container in the route created in the method makeRequestFromUrlAndMethod then the routing system can't resolve any Route parameter binding anymore, so after set the container in the route all works good.
```PHP
    {
        $request = Request::create($url, $method);

        if ($session = request()->getSession()) {
            $request->setLaravelSession($session);
        }

        $request->setUserResolver(request()->getUserResolver());

        $route = app('router')->getRoutes()->match($request);

        // For some reason without this octane breaks the route parameter binding.
        $route->setContainer(app());

        $request->setRouteResolver(function () use ($route) {
            return $route;
        });

        return $request;
    }
```

### Fixed memory issues with some of the clases in Features folder.
We need to be careful when working with octane and Livewire::listen, because Livewire instance is a singleton it will keep in memory any reference inside the listener, for example in SupportComponentTraits:
```PHP
    protected $componentIdMethodMap = [];

    function __construct()
    {
        Livewire::listen('component.hydrate', function ($component) {
            $component->initializeTraits();

            foreach (class_uses_recursive($component) as $trait) {
                $hooks = [
                    'hydrate',
                    'mount',
                    'updating',
                    'updated',
                    'rendering',
                    'rendered',
                    'dehydrate',
                ];

                foreach ($hooks as $hook) {
                    $method = $hook.class_basename($trait);

                    if (method_exists($component, $method)) {
                        $this->componentIdMethodMap[$component->id][$hook][] = [$component, $method];
                    }
                }
            }

            $methods = $this->componentIdMethodMap[$component->id]['hydrate'] ?? [];

            foreach ($methods as $method) {
                ImplicitlyBoundMethod::call(app(), $method);
            }
        });
```
the $componentIdMethodMap variable referenced in the listener will be the same on each request, so the listener will add more and more callables to the variable and it grows without limit, my solution is keep all the "state" in the component so I added a few methods on the component to store the state. 

@calebporzio could you take a look on this?


luischavez:
Related issues:
#3495 
#3899
#2878
#2720

calebporzio:
Thanks for submitting this @luischavez, it's a huge help!

Is it possible to separate out these changes into separate PRs?

My reasoning is this:
- I'd like to merge the stuff with the `$route->setContainer` and `__invoke()` changes right away.
- I think the `->getState()` strategy is good and their should be a mechanism like that, however, I'm not sure about putting it on the component class itself. Sometimes we need cross-component data when components render nested components. So separating it out into a different PR allows us to talk more about that solution specifically.
- @joshhanley has some ideas about the "Fixed memory issues" part that I think he wants to PR to Octane itself.
- It looks like there might be a bug with you removing the array of hooks whitelist entirely (`update`, `hydrate`, `mount` etc...)

Thanks again, but I'm not comfortable merging this all at once.

luischavez:
@calebporzio no problem, I put the changes related to the state into this PR: #3940

luischavez:
@joshhanley Thanks, as you suggest I split the changes on 3 PR:
#3947 
#3946 
#3940 

Also I reverted the changes on this PR as I original submit.

joshhanley:
@luischavez great, thanks! 🙂

calebporzio:
Thanks @luischavez - I'm closing this one

PR #3945, opened 2021-10-04
Add support for after validation hook

Arkitecht:
Laravel provides an after hook for validation, that you can use to perform other validation logic outside of the rules. Currently with Livewire's implementation, there is no way to hook into that, since the validate method returns the validatedData but not the validator itself. This extends the validate method to take an optional after argument, which when provided, is sent to the validator after hook. I added an accompanying test as well.

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -144,7 +144,7 @@ trait ValidatesInput
         return ! $this->hasRuleFor($dotNotatedProperty);
     }
 
-    public function validate($rules = null, $messages = [], $attributes = [])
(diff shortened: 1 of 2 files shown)


joshhanley:
@Arkitecht thanks for the PR! There is a similar PR already open that will allow you to tap into all the different validator methods like `after`.

See it here https://github.com/livewire/livewire/pull/3777

Hope this helps!

calebporzio:
Good call @joshhanley - thanks for linking to that. Closing this one for now as we should be ready to merge the other one soon enough.

Arkitecht:
Thanks all, that PR does look to cover it, and in a more elegant and comprehensive way. Looking forward to that update.

PR #3946, opened 2021-10-04
Better Octane support - route binding fix

luischavez:
Moved from #3936

### Fixed __invoke Route parameter binding.
This was really hard to found and to be honest I really don't know why is this causing the issue, but if you not set the container in the route created in the method makeRequestFromUrlAndMethod then the routing system can't resolve any Route parameter binding anymore, so after set the container in the route all works good.
```PHP
    {
        $request = Request::create($url, $method);

        if ($session = request()->getSession()) {
            $request->setLaravelSession($session);
        }

        $request->setUserResolver(request()->getUserResolver());

        $route = app('router')->getRoutes()->match($request);

        // For some reason without this octane breaks the route parameter binding.
        $route->setContainer(app());

        $request->setRouteResolver(function () use ($route) {
            return $route;
        });

        return $request;
    }
```


calebporzio:
Thanks for breaking this out!

PR #3947, opened 2021-10-04
Better Octane support - full page component fix

luischavez:
Moved from #3936

### Fixed properties and computed properties in full page component that preserves along requests.
The Illuminate/Routing/Route keeps in memory an instance of the component:
[https://github.com/laravel/framework/blob/8.x/src/Illuminate/Routing/Route.php#L271](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Routing/Route.php#L271)
```PHP
  /**
     * Get the controller instance for the route.
     *
     * @return mixed
     */
    public function getController()
    {
        if (! $this->controller) {
            $class = $this->parseControllerCallback()[0];

            $this->controller = $this->container->make(ltrim($class, '\\'));
        }

        return $this->controller;
    }
```
so we need to create always a fresh instance of the component on the __invoke method.


calebporzio:
Thanks!

PR #3966, opened 2021-10-07
[FIX] Check all the values in dehydratePropertyFromWithFileUploads before calling serializeMultipleForLivewireResponse

some-daan:
Referring to this issue: https://github.com/livewire/livewire/discussions/2243

I'm using a file upload (using Livewire fileuploads) functionality with sorting (using Livewire/sortable). I ran into the same issue as the creator of the issue mentioned above. Once you start dragging a freshly added image to the first position in the list an error pops up. There's [a video](https://drive.google.com/file/d/10sVkwguH_b0xWuzdpPNvwqeaCof_8rJp/view) in the mentioned issue that shows you how to reproduce this bug.

This issue only happens when a newly uploaded image gets dragged to the first position. This is because the $value array would look like this:

image

The _dehydratePropertyFromWithFileUploads_ function only checks in the third if-statement if the first item of this array is an instance of _TemporaryUploadedFile_ while that doesn't necessarily mean that this is a valid array.

The _serializeMultipleForLivewireResponse_ calls the _getFilename_ on all of the items in the array, which will fail for items that aren't an instance of _TemporaryUploadedFile_. 

This pull request fixes this issue by not allowing invalid arrays to call the _serializeMultipleForLivewireResponse_ function, but instead let's them pass on to the fourth if-statement of the _dehydratePropertyFromWithFileUploads_ function. An invalid array meaning that not all items in the array are an instance of _TemporaryUploadedFile_

### DIFF
--- a/src/Features/SupportFileUploads.php
+++ b/src/Features/SupportFileUploads.php
@@ -45,8 +45,19 @@ class SupportFileUploads
             return  $value->serializeForLivewireResponse();
         }
 
-        if (is_array($value) && isset(array_values($value)[0]) && array_values($value)[0] instanceof TemporaryUploadedFile && is_numeric(key($value))) {
-            return array_values($value)[0]::serializeMultipleForLivewireResponse($value);
+        if (is_array($value) && isset(array_values($value)[0])) {
+            $isValid = true;
+
+            foreach (array_values($value) as $key => $arrayValue) {
+                if (!($arrayValue instanceof TemporaryUploadedFile) || !is_numeric($key)) {
+                    $isValid = false;
+                    break;
+                }
+            }
+
+            if ($isValid) {
+                return array_values($value)[0]::serializeMultipleForLivewireResponse($value);
+            }
         }
 
         if (is_array($value)) {


jhonoryza:
i have this problem too

some-daan:
@jhonoryza Thanks for letting me know that you're having the same problem. Do you want to try my change and let me know if that works for you? I'd highly appreciate it. 😄 

Since there isn't any interest from the project maintainers for this fix this might catch their attention.

jhonoryza:
@some-daan in my case i am using multiple dynamic input with file upload and yes i try your implementation with overriding supportFileUploads class, and it works thanks 👍🔥

some-daan:
@jhonoryza That's great to hear and thanks for taking the time to test this. Let's hope this gets merged soon.

joshhanley:
@some-daan thanks for the PR! Any chance you could add tests for this?

Also something strange is going on with the PR, it's not letting me run the existing test suite. Would you be able to close this PR and open a new one, so we can run all the existing tests to make sure nothing breaks?

Thanks!

PR #3971, opened 2021-10-08
[Fix] Add fix for skip render on redirect

joshhanley:
PR #2747 released in v2.6 allowed the render method to run on redirect, to ensure that browser's bfcache had the latest render before the redirect was performed, so users weren't presented with stale data when they clicked on the back button after a redirect.

But that caused issues with redirecting in mount or an action method if authorisation failed (or something similar) and you don't want the render method to run.

This PR reverts the default back to skipping render on redirect and adds a config option `render_on_redirect` which is set to `false` by default. 

This config option allows developers to set this to `true` if they are concerned about users seeing stale data when they click the back button.

If the config option has been set to `true`, a developer can still tell Livewire to skip render on a specific redirect by calling `$this->skipRender()` right before the redirect call.

```php
public function mount() {
    $this->skipRender();
    return $this->redirect('/some-path');
}
```

Hope this helps!

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -126,4 +126,19 @@ return [
 
     'back_button_cache' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Render On Redirect
+    |--------------------------------------------------------------------------
+    |
+    | This value determines whether Livewire will render before it's redirected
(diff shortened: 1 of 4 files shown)


joshhanley:
No worries 🙂

PR #3987, opened 2021-10-10
[Fix] Add flush state and feature listeners to support Octane

joshhanley:
Currently some of Livewire's feature classes are retaining state between requests when using Octane, see #2859,#3492, #3509.

This PR adds a method to the `LivewireManager` class called `flushState` that fires an internal Livewire event, that the feature classes can listen to and reset their own state.

A corresponding PR will be submitted to Octane, with an Octane listener that triggers this `flushState` method as required to prepare for new requests.

This will allow us to add new features to Livewire without needing to add a corresponding PR to Octane.

Hope this helps!

Edit: Here is the Octane PR https://github.com/laravel/octane/pull/400

### DIFF
--- a/src/Features/OptimizeRenderedDom.php
+++ b/src/Features/OptimizeRenderedDom.php
@@ -29,5 +29,9 @@ class OptimizeRenderedDom
                 $response->effects['html'] = null;
             }
         });
+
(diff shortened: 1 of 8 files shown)


calebporzio:
Love it. Thanks @joshhanley - great work on this!

PR #3992, opened 2021-10-11
Support pushing and prepending to stacks from Livewire components

calebporzio:


joshhanley:
@vlados we've had to revert this feature for now.

PR #4001, opened 2021-10-12
[Fix] Override `getPath` to return temp path set by Livewire

amiranagram:
I do believe it is needed since Livewire overrides temp path but method `getPath` returns path set by system.
No but other devs have raised concerns in #3055 and #3064.

No.

Yes.

When working with file uploads in Livewire we should be able to return file paths with `getPath`.

When instantiating `TemporaryUploadedFile` we are creating a temporary file then giving it's path to the parent constructor, which goes all the way to the first ancestor `SplFileInfo`. Now the object has path defined as the default temporary path defined by the system.

https://github.com/livewire/livewire/blob/ba99d0d17267f22bdee34a159a6921c772ef5820/src/TemporaryUploadedFile.php#L15-L24

Livewire does a good job of overriding tons of other methods, but `getPath` is left hanging, which is IMHO very useful as tons of third-party packages rely on it.

One example from the top of my head being [laravel-medialibrary](https://github.com/spatie/laravel-medialibrary) where you can upload files by just passing an `\Symfony\Component\HttpFoundation\File\UploadedFile` instance (which extends `SplFileInfo`, extended by `TemporaryUploadedFile`) into it's `addMedia` method.

```php
class Form extends Component
{
    use WithFileUploads;

    public $avatar;
    
    public function saveAvatar()
    {
        Auth::user()->addMedia($this->avatar)->toMediaCollection('avatars');
    }
    
    public function render()
    {
        return view('livewire.form');
    }
}
```
```html
<form wire:submit.prevent="saveAvatar">
    <input type="file" wire:model="avatar">

</form>
```

This little snippet will not work in the latest release, but this PR fixes it.

P.S. I exposed `FileUploadConfiguration::directory()` as a public method since my fix relies on it but if you have any better way to do this that I missed let me know.

🙌

### DIFF
--- a/src/FileUploadConfiguration.php
+++ b/src/FileUploadConfiguration.php
@@ -52,7 +52,7 @@ class FileUploadConfiguration
         return config('filesystems.disks.'.strtolower($diskBeforeTestFake).'.driver') === 'gcs';
     }
 
-    protected static function directory()
+    public static function directory()
     {
         return Util::normalizeRelativePath(config('livewire.temporary_file_upload.directory') ?: 'livewire-tmp');
     }
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -23,6 +23,11 @@ class TemporaryUploadedFile extends UploadedFile
         parent::__construct(stream_get_meta_data($tmpFile)['uri'], $this->path);
     }
 
+    public function getPath()
+    {
+        return $this->storage->path(FileUploadConfiguration::directory());
+    }
+
     public function isValid()
     {
         return true;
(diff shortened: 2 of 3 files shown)


hazem-taha:
Hello,

I'm getting 

```
File `livewire-tmp` does not exist
```

Here is how I'm trying to upload the file..

```
                auth()->user()->addMedia($file->getPath())->toMediaCollection('docs');
```

amiranagram:
@hazem-taha You can pass your file object directly into `addMedia` function, which is possible with this fix. Like this:

```php
auth()->user()->addMedia($file)->toMediaCollection('docs')
```

hazem-taha:
@amiranagram 
That is what I was already using but it wasn't working and getting this error

image

amiranagram:
@hazem-taha I don't have this issue with media library. Could you provide some code (maybe in a discussion, then reference this PR)?

Also you should be getting an absolute file path, not what you're getting. What filesystems driver are you using?

hazem-taha:
@amiranagram 
I used `addMediaFromUrl` on `temporaryUrl` and it worked fine, but still don't know why the normal way does not work.

image

PR #4018, opened 2021-10-14
Add custom exception handling when performing actions calls

lgmworks:
In a traditional Laravel application if the flow is broken by throwing an exception, you can react to this by rendering an appropriate response to the exception type and request type(html or json) when using the App\Exceptions\Handler class, but in Livewire there is no equivalent, when a call to a component method fails, unless it is a ValidationException it is thrown, and in the front end there is no way to capture the message and type of error to show something more suitable to the user. It is desirable to have this functionality to be able to react to exceptions in personalized ways and not just as it is currently default in Livewire.

**Example:**

When in a hypothetical component "Checkout", some class method is called that makes an http request to the payment gateway, if this request fails, you do not want to show any dialog to the user saying that there was a generic error, what you really want is to show a message saying that:
"_Communication with the payment processor failed, try again later, or contact the email: support@example.com_"
 and in addition to that send the failure to the application's error reporting system(eg: Bugsnag), the part of sending the failure and the writing of the message can be done in the nested class, no matter the depth, and the message is sent upwards by throwing a custom exception class, and the Livewire component must be able to display the message, but for that instead of having a try catch block for each possible custom exception, you could add methods in the component or in a trait to handle each type of exception that you want to handle, and show the user the appropriate message for each failure.

### DIFF
--- a/src/HydrationMiddleware/PerformActionCalls.php
+++ b/src/HydrationMiddleware/PerformActionCalls.php
@@ -3,6 +3,7 @@
 namespace Livewire\HydrationMiddleware;
 
 use Livewire\Livewire;
+use Throwable;
 use function Livewire\str;
 
 use Illuminate\Validation\ValidationException;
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for this submission, going to close this for now. I don't think there's enough demand for this and I'd rather not add a separate API dedicated to this functionality that we'd have to adhere to going forward.

Thanks again!

calebporzio:
Also, I think using a try/catch is the best course of action for people wanting this functionality currently

PR #4026, opened 2021-10-16
Fix boot so it runs before all mount and hydrate methods

joshhanley:
Currently on the initial request the `boot` method runs before `mount`. But on a subsequent request, it runs after `hydrateFoo` but before `hydrate`. See discussion #3875 for details.

This PR fixes the boot method to ensure it runs first before any `mount`, `hydrate`, `hydrateFoo`, etc. methods.

There was an issue with the tests, where component property hydration was wiping out any changes pushed into the test `$memo` property by the boot method (but that's a good thing as it now means it's working as it should 😁). 

So made a protected `$_memo` to build up the string before assigning to the public property in render to enable checks to be run against it. I've added comments into the test components that explains that.

As such it means the protected `$_memo` is empty at the beginning of each request, allowing me to explicitly check results only from that request instead of a full string of from the initial request too.

Hope this helps!

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -66,6 +66,7 @@ abstract class Component
         }
 
         $manager = LifecycleManager::fromInitialInstance($instance)
+            ->boot()
             ->initialHydrate()
             ->mount($componentParams)
             ->renderToView();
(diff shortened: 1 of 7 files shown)


calebporzio:
Thanks Josh

PR #4027, opened 2021-10-16
Adding an "except" and "all" methods into InteractsWithProperties.php

nihad-ziyatkhanli:
Yes, I need it whenever I need to process public properties. I created a [discussion](https://github.com/livewire/livewire/discussions/3704) about it first.
No.
No.

`all` is just short alias of `getPublicPropertiesDefinedBySubClass` method which already existed. It is compatible name with laravel's array helpers and collection methods. Both `all` and `except` methods are often used when there are many public properties and you don't want to type each name. For example, when you want to process (for example, trimming) public properties other than password, you can pass `$this->except(['password'])` to whatever method that will process it. It is way less typing than `$this->only(['first_name', 'last_name', 'mid_name', 'email', 'phone_number', /* and so on */])`.
For my case, I first needed it for calling `Validator::make`, when `$this->validate()` wasn't suitable for the logic:

```
/* $flag is a boolean that doesn't need to be validated */
$validator = Validator::make($this->except(['flag']), $this->rules);
if ($validator->fails()) {
     $this->reset();
}
```
The exact case of mine was more complex, so I put simplified version here.
I would have to list all properties manually without the `except` method.


calebporzio:
Thanks!

PR #4036, opened 2021-10-18
Validating DTOs

stevenmond:
Yes - I created a discussion and was recommended to make a PR for it - https://github.com/livewire/livewire/discussions/4032

No

Yes

I added check in unwrapDataForValidation of ValidatesInput for a Wireable object. If an object matches, it uses the Wireable object's toLivewire() function to serialize it for validation purposes.

Right now from my testing it appears as though when trying to validate a Wireable object it doesn't work and errors. The issue looks to be the unwrapDataForValidation() function noop-ing it as it didn't match any of the objects for explicit serialization. The end result was unwrapDataForValidation() would return a class::__set_state(array()) object rather than an array to be used in validation.

I used the toLivewire() function as the simplest solution (since it already returns an array), but if there's a better way to do it (like an explicit serialization function like model and collection classes have) I'm all for it.


joshhanley:
Tests all look good, that one that failed is unrelated, so should be good to merge!

stevenmond:
As a follow-up it looks like this fix breaks with images since they need to resolve differently based on the context.

When serializing for Livewire, it needs to be done via a manual call to serializeForLivewireResponse(). However, this serialization isn't recognized by the validator/unserialized by the validator when checking file/image-related properties.

If explicit serialization is ignored (like just trying to return the file) it just results in it returning {"disk": "public"} to Livewire since that's the only public property.

I'm trying to think through a solution to this.

joshhanley:
@stevenmond can you also make sure you add a test for `validateOnly` as it has its own code. You will need to rebase on master as I recently fixed a bug in `validateOnly`.

Thanks!

stevenmond:
@joshhanley Added in the requested tests.

On a side note, I think I've figured out the other issue in regards to image serialization. What's going on is that ```SupportsFileUploads``` isn't checking for Wireables to see if it had any images that needed to be serialized. Adding the below to the end of ```dehydratePropertyFromWithFileUploads``` looks to have resolved the issue. Want me to open another pull request for this?

```
        if ($value instanceof Wireable) {
            $keys = array_keys((array) get_object_vars($value));

            foreach ($keys as $key) {
                $value->{$key} = $this->dehydratePropertyFromWithFileUploads($value->{$key});
            }
        }
```

joshhanley:
@stevenmond thanks! I've triggered the tests to run again.

Yep if you can, that'd be great, thanks!

stevenmond:
Will do. Fixed an issue with an extra comma being added to a test as well.

joshhanley:
@stevenmond yeah that gets me every time haha, I'm so used to trailing commas on everything now!

calebporzio:
Thanks @stevenmond  and @joshhanley !

reppair:
@stevenmond @joshhanley @calebporzio seems this breaks comparison rules like Greater Than / Less Than since the returned data from the `unwrapDataForValidation` no longer contains all the component attributes as explained in Discord Help channel.

joshhanley:
@reppair thanks for reporting! Have a look a discussion #4278 for more details what is going on. I'm currently investigating a fix.

Hope this helps!

PR #4037, opened 2021-10-18
[Failing Test] Duplicate `wire:key` issues

joshhanley:
This adds a failing tests for the scenario where there are two lists on a page with each element using the same `wire:key` structure like `$loop->index`.

If a new element is added to one list with the same `wire:key` as an element in the other list, when morphdom tries to morph them, it moves the existing one instead of creating a new one.

See screen capture below, where there are two components each displaying the same list. If we add a new user to the child component and refresh the parent, we can see the new user gets moved.

 I've added a sleep to the parent component so we can clearly see the user in the child list before the parent refreshes.

I suspect it has to do with Livewire's custom morphdom look ahead.

Hope this helps!

recording (46)

### DIFF
--- a/tests/Browser/Morphdom/ChildComponent.php
+++ b/tests/Browser/Morphdom/ChildComponent.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Tests\Browser\Morphdom;
+
+use Livewire\Component as BaseComponent;
+
+class ChildComponent extends BaseComponent
+{
+    public $numbers = [
+        1,
+        2,
+    ];
+
+    public function addNumber()
+    {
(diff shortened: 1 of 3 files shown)


nuernbergerA:
@joshhanley is this a real issue? because I always prefix my keys with the model name or something similar or is it just to illustrate the problem?

calebporzio:
Note for ourselves: The solution here is to use Alpine's new "Morph" plugin instead of morphdom - its algo resolves this issue

joshhanley:
Closing as this shouldn't be an issue in V3 as keys will be scoped to within the parent element.

PR #4055, opened 2021-10-20
Fix Content-Disposition for RFC 5987

morodomi:
I want it but not discussed.

No.

No.

When an upload filename is too long, it's useful.


joshhanley:
@morodomi thanks for the PR!

Can you please explain in some more detail why this is useful when the filename is too long? Are you getting any errors? 

Also there is another PR open with discussions about fixing the filename length you may want to check out #2841.

Thanks!

calebporzio:

PR #4064, opened 2021-10-21
Check the class name is valid before creating it

mrsimonbennett:

No, just a error I found by been an idiot

None, just one command/test

Tests pass and added test to prove the fix

Just stops you from running the `livewire:make` command with an invalid class name like **1Class**


joshhanley (approved):
Good call! Looks good!

PR #4085, opened 2021-10-24
fix wire:key issue containing commas when using on livewire components

walirazzaq:
Yes / No 🙊
No
No (Actually I am bad at tests atm, learning though)
Current implementation of wire:key extraction based on comma( , ) and causes error when there is comma in key value either explicitly or when using some function/method and passing more than one params like
```php
   @livewire('componen-name',key(some_helper('foo','bar')))
   //Or as in component syntax
   <livewire:component-name wire:key="{{some_helper('foo','bar')}}"
```
Owing to comma inside expression string, it bypasses the current implementation and compiles to 
```php
   key('.e(some_helper('foo','bar')).')
```
This causes following TypeError as `key()` accepts an array
image

Pleasure is always mine. It's in fact my first ever contribution (though a small one) to any OSS and to my luck it is to my favorite one 🎉

### DIFF
--- a/src/LivewireBladeDirectives.php
+++ b/src/LivewireBladeDirectives.php
@@ -45,10 +45,10 @@ EOT;
 
     public static function livewire($expression)
     {
-        $cachedKey  = "'" . Str::random(7) . "'";
+        $cachedKey = "'" . Str::random(7) . "'";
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @walirazzaq!

PR #4093, opened 2021-10-25
Allow response effects html to be an html node

godismyjudge95:
It is needed to be able to properly fix Livewire's integration with Vue 3 (separate pull request coming soon on the livewire/vue repo).

No, just a single change.

Sorry no, but all included tests should still work.

I added a simple condition to check that the `response.effects.html` is a string before running `trim()` on it.  This prevents an error if the `response.effects.html` is an html node - which can happen when the response is hooked into via the `message.received` hook.

Basically, this code should function the same in all use cases except when the `message.received` hook is used to change it to an html node.

Thank you for such a great library.  This is my first pull request to Livewire, hopefully many more in the future.

### DIFF
--- a/js/component/index.js
+++ b/js/component/index.js
@@ -288,7 +288,14 @@ export default class Component {
             // If we get HTML from the server, store it for the next time we might not.
             this.lastFreshHtml = response.effects.html
 
-            this.handleMorph(response.effects.html.trim())
+			let html = response.effects.html;
+
+			// If the html is a string trim it, otherwise it should be a html node
+			if (typeof response.effects.html === 'string') {
+				html = response.effects.html.trim();
+			}
+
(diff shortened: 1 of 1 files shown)


godismyjudge95:
Re-opened this due to a more solid example on why it is needed:  https://github.com/livewire/livewire/pull/4127

calebporzio:
Hey @godismyjudge95, I just commented on the livewire/vue PR.

My main hesitancy with adding this to core is now the future of Livewire is locked into allowing dom nodes as morph targets, which currently that isn't a constraint we have to honor.

However, if it is the only way to allow you're (hopefully) third-party Vue 3 support plugin I'm willing to consider.

Let me know if this is the only way and I'll seriously consider merging this. Thanks!

PR #4098, opened 2021-10-26
New method "resetExcept"

SebastianSchoeps:
If you have a component with many public properties and want to reset all except some, you would have to write all properties into the reset method - and if you add more public properties, you would always have to remember to add them. This PR makes this a little easier - you can name all the properties you want to keep.

I don't know if that's of use for enough people, but I needed it. I have it in a base component anyway, so no hard feelings if this PR gets rejected :-).

https://github.com/livewire/livewire/discussions/4097

No.

No, but would be easy to add at `ResetPropertiesTest.php`.

See above.


calebporzio:
Thanks!

PR #4103, opened 2021-10-26
bugfix/paginatio-type :bug: Changes the page type from int to double.

oyepez003:
It's a very small change, if in the PR discussion the most of you guys consider the change is not required, the PR can be rejected.

No, It doesn't.

Some tests have been added to ComponentPaginationTest.php. 

`vendor/bin/phpunit tests/Unit/ComponentPaginationTest.php`

When the user tries to paginate a component using an integer overflow value livewire returns a 500 error in the pagination response.

Example: 
http://localhost/my-component-with-pagination?page=9223372036854775807
Returns error 500

Values equal to **999999999999999999** or lower works perfectly.

Taking into account the final user is able to change the page param value in the browser URL, livewire should catch this error.

### DIFF
--- a/src/WithPagination.php
+++ b/src/WithPagination.php
@@ -36,7 +36,7 @@ trait WithPagination
                 $this->paginators[$pageName] = request()->query($pageName, 1);
             }
 
-            return (double) $this->paginators[$pageName];
+            return (int) $this->paginators[$pageName];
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @oyepez003!

PR #4108, opened 2021-10-27
js() helper for blade component $attributes

tanthammar:
Using the `@js` directives within component tags is not supported by Laravel. 
For example, `<x-alert :errors="@js($errors->all())"/>` will not be compiled.
This PR adds a `js()` helper to use in blade components or anywhere in php.

### Why we need this:

Lets say we have a blade component; button.blade.php
```blade
<button {{ $attributes->merge(['class' => 'classes']) }}>
{{ $slot }}
</button>
```

If we want to add Alpine interactivity we cannot use the Livewire `@js` directive, it won't compile.
```blade
<x-button x-data
   x-on:click="$dispatch( 'notify-errors', @js($errors->all()) )"
   type="submit"> Submit </x-button>
```

With the `js()` helper we will be able to work around it.
```blade
<x-button x-data
   x-on:click="$dispatch( 'notify-errors', {{ Livewire\js($errors->all()) }} )"
   type="submit"> Submit </x-button>
```

I added `$options` and `$depth`, following the Laravel `@json()` directive. Adds usability that the Livewire `@js()` directive is missing.

Use the `@js()` blade directive with _optional_ flags and depth
```blade
<div 
    x-data="{ config: @js( $complexNestedArray, JSON_NUMERIC_CHECK | JSON_FORCE_OBJECT, 8 ) }" 
></div>
```

Where it is not possible to use the directive, we use the new `js()` method.
```blade
<x-foo 
    x-data="{ config: {{ Livewire\js( $complexNestedArray, JSON_NUMERIC_CHECK | JSON_FORCE_OBJECT, 8 ) }} }" 
></x-foo>
```

I like to keep the namespace `Livewire\js` to make it clear where the helper comes from and to avoid naming conflicts.

Discussed in #4083

### DIFF
--- a/src/LivewireBladeDirectives.php
+++ b/src/LivewireBladeDirectives.php
@@ -20,17 +20,7 @@ EOT;
 
     public static function js($expression)
     {
-        return <<<EOT
-<?php
-    if (is_object({$expression}) || is_array({$expression})) {
-        echo "JSON.parse(atob('".base64_encode(json_encode({$expression}))."'))";
-    } elseif (is_string({$expression})) {
-        echo "'".str_replace("'", "\'", {$expression})."'";
-    } else {
-        echo json_encode({$expression});
-    }
-?>
-EOT;
+        return '{!! Livewire\js('.$expression.') !!}';
     }
 
     public static function livewireStyles($expression)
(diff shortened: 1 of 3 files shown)


nuernbergerA (approved):

tanthammar:
The PR was accepted. Should I close this one?

calebporzio:
Thanks for this @tanthammar - as this in now in Laravel, Livewire will likely deprecate our own directive for the versions of Laravel that have this, so I'm going to close this. Thank you!

PR #4127, opened 2021-10-29
Allow response effects html to be an html node - part 2

godismyjudge95:
Re-opening this:  https://github.com/livewire/livewire/pull/4093

>If this is for a change to support Vue3 - then maybe you can convert the node to a string (.outerHTML) inside the plugin and not in core? I'm hesitant to allow a different type into the handleMorph method as we have to then support that type ongoing.
>
>Going to close this. If you disagree and have a reason it needs to be here, please resubmit the PR.

**Reason for re-opening:**  After further testing, Vue 3 is incompatible with using the .outerHTML method described.
Here are two examples to illustrate the issue:

- Using outerHTML - Vue 3 click events do not work after clicking onto a new page:
https://laravelplayground.com/#/snippets/bb262148-ab44-4c14-b900-54fcc84792b3

- Using html node - Vue 3 click events work properly even after clicking onto a new page:
https://laravelplayground.com/#/snippets/caf93e10-223d-479a-aa2c-175620bc647c

If you click the increment or decrement button on either example it works fine.  As does changing the text in the text boxes.  However, once you use the paginate buttons (which causes the whole livewire div to refresh), the increment and decrement buttons will only work on the html node version.

Please let me know if there is anything I can clear up.


calebporzio:
Closing this due to the following comment on the livewire/vue repo: https://github.com/livewire/vue/pull/38#issuecomment-965980323

PR #4135, opened 2021-11-01
Fix for model properties set to an empty array returning NULL after a request

joshhanley:
Fixes #3840 

If you have an eloquent model and set one of its properties to an empty array, on the next request it was being hydrated as null instead of the empty array.

```php
$model = new Model();
$model->list = [];  // After a request this was returning NULL
```

Hope this helps!


calebporzio:
Thanks Josh!

PR #4141, opened 2021-11-01
Add custom headers to ajax message request

naabster:

Add custom headers to message ajax requests. Useful for example to add an Authorization Header if no cookie session is available.
Also discussed here: #1320

no

yes

LiveWire sends ajax messages currently without the option to add custom headers. In more complex setup scenarios, custom headers for these requests might be crucial, to allow LiveWire to play nicely with anything that relies on these headers.

Examples might be:
* Custom **X-Foo-Something** Headers, for example multi tenancy with identification through headers (see #1320)
* **Authentication** Headers (Bearer Token), when cookies are not an option (for example in iFrames, when third party cookies are blocked as in shopify applications)

Custom Headers are set like this:

```html
    <script type="text/javascript">
        document.addEventListener("livewire:load", function(event) {
            window.livewire.connection.headers = {
                'X-Foo-Header': 'Bar'
            }
        });
    </script>
```


calebporzio:
Thanks for this, and thanks for adding tests.

I see the need for this, but I think the public API could be a bit better.

What do you think about instead of manually adding to `livewire.connection.headers`, doing:

```js
Livewire.addHeaders({
    'X-Foo-Header': 'Bar'',
})
```

If that's cool with you, update the PR and resolve the conflicts and we'll merge. Thanks!

naabster:
Makes absolute sense and is way nicer. I just was not sure, how guarded you are with your namespace :)

calebporzio:
looks great! Thanks!

zepfietje:
@naabster is there any reason why you're setting the custom headers before the `'X-CSRF-TOKEN'` header? For https://github.com/livewire/livewire/discussions/4332 I want to set the `'X-CSRF-TOKEN'` token manually, but I can't since it will be overwritten.

In my opinion, the user should be able to override all headers (so the custom headers line will need to be moved down a bit). Thoughts?

PR #4144, opened 2021-11-02
Allow Responsable objects download

fabio-ivona:
Hi all!

this PR will add support for Responsable objects download

my main concern is to be able handle a download from a Responsable object like this one (it may be, for example, a **spatie/media-library Media** object)

```php
class DownloadableResponse implements Responsable
{
    public function toResponse($request)
    {   
        return  Storage::download('test-file.txt');
    }
}
```

as for now, in order to allow the download, the Livewire component method should be 

```php
 public function download()
{
    $response = new DownloadableResponse();

    return $response->toResponse(request());
}
```

with this PR, you will be able to do:

```php
 public function download()
{
    return new DownloadableResponse();
}
```

and the `->toResponse()` conversion will be handled in SupportFileDownloads Trait

### DIFF
--- a/src/Features/SupportFileDownloads.php
+++ b/src/Features/SupportFileDownloads.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features;
 
+use Illuminate\Contracts\Support\Responsable;
 use Livewire\Livewire;
 use Illuminate\Support\Str;
 use Symfony\Component\HttpFoundation\StreamedResponse;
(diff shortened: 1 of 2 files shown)


joshhanley:
@fabio-ivona ah nice! I've had that issue before, so this will definitely be helpful!

calebporzio:
Perfect! Thanks @fabio-ivona!

PR #4153, opened 2021-11-03
[Fix] Prevent wire end comment from disappearing #3887 #4015

hpieterse:
This fixes issues #3887 and #4015

### Livewire wire:end, nested component, not re-generated = Error, multiple root elements #3887
The code changes include an update to the morph dom implementation. The logic that transverse the dom simply skips any occurrences for the wire-end comments. This prevents them from being removed. This in turn prevents incorrect console warnings about multiple root elements.

### Multiple root element check fails after web performance optimization strips comments in HTML #4015
The changes include adding conditional comments is to prevent minification services like  Cloudflare's Auto Minify from stripping the comments 
(https://community.cloudflare.com/t/omit-formatted-comments-from-minification/18572/21)


calebporzio:
Thank you @hpieterse! Mind resolving the conflicts and adding a test? Thanks again!

joshhanley:
Another issue related to this is #3814, just putting it here for reference.

joshhanley:
I wonder whether we should change to using `<template>` tags instead of using conditional HTML comments for the root detection, incase other minifiers don't make use of them like Cloudflare does.

Then it wouldn't be impacted by any minifiers, and still accessible as a node for when we walk down the DOM.

hpieterse:
> Thank you @hpieterse! Mind resolving the conflicts and adding a test? Thanks again!

Hello @calebporzio, I've been getting stuck on running the browser tests. I'm not sure what I am missing. I installed the chrome drivers with './vendor/bin/dusk-updater` and when I run the test, I can see a chromium instance starting up, but the test keeps failing.

For example,
```
./vendor/bin/phpunit --filter test  tests/Browser/Morphdom/Test.php
``` 

Fails with
```
PHPUnit 9.5.10 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 02:31.520, Memory: 28.00 MB

There was 1 error:

1) Tests\Browser\Morphdom\Test::test
Facebook\WebDriver\Exception\WebDriverCurlException: Curl error thrown for http POST to /session with params: {"capabilities":{"firstMatch":[{"browserName":"chrome","goog:chromeOptions":{"w3c":false,"prefs":{"download.default_directory":"/home/hein/repo/livewire/tests/Browser/downloads"}}}]},"desiredCapabilities":{"browserName":"chrome","platform":"ANY","chromeOptions":{"w3c":false,"prefs":{"download.default_directory":"/home/hein/repo/livewire/tests/Browser/downloads"}}}}

Operation timed out after 30005 milliseconds with 0 bytes received

/home/hein/repo/livewire/vendor/php-webdriver/webdriver/lib/Remote/HttpCommandExecutor.php:333
/home/hein/repo/livewire/vendor/php-webdriver/webdriver/lib/Remote/RemoteWebDriver.php:131
/home/hein/repo/livewire/tests/Browser/TestCase.php:244
/home/hein/repo/livewire/vendor/laravel/dusk/src/Concerns/ProvidesBrowser.php:219
/home/hein/repo/livewire/vendor/laravel/framework/src/Illuminate/Support/helpers.php:234
/home/hein/repo/livewire/vendor/laravel/dusk/src/Concerns/ProvidesBrowser.php:220
/home/hein/repo/livewire/vendor/laravel/dusk/src/Concerns/ProvidesBrowser.php:98
/home/hein/repo/livewire/vendor/laravel/dusk/src/Concerns/ProvidesBrowser.php:66
/home/hein/repo/livewire/tests/Browser/TestCase.php:264
/home/hein/repo/livewire/tests/Browser/Morphdom/Test.php:58

```

It seems like the browser is not loading the content that is expected. Is there a different way I should be running the browser tests, or am I missing some dependency?

joshhanley:
@hpieterse I've started writting up a contribution guide, you can have a look at the draft here to see if it might help you get the tests running https://github.com/joshhanley/docs/blob/contribution-guide/contribution-guide.blade.php

Hope this helps!

hpieterse:
@joshhanley thanks for the info. It helped.

@calebporzio I added a test and resolved the conflicts.

> I wonder whether we should change to using `<template>` tags instead of using conditional HTML comments for the root detection, incase other minifiers don't make use of them like Cloudflare does.
> 
> Then it wouldn't be impacted by any minifiers, and still accessible as a node for when we walk down the DOM.

I like the conditional comments more. It's a smaller change, and I think comments in the dom is better than empty template tags. But that's just my opinion.

I would expect most minifiers to keep the conditional comments, but if you guys prefer using template tags, then I could look into it and update the PR.

joshhanley:
@hpieterse thanks for adding the test and fixing the conflict! 

So we're actually making use of template tags for other things in Livewire to denote start/end of things, hence why I suggested it.

But I think we should wait to see what Caleb thinks before changing anything else 🙂

calebporzio:
Thanks for this, couple thoughts:

* This is actually two PRs I think:
  * A) A PR to prevent Cloudflare from stripping out the "end" comment
  * B) A fix to ensure that Livewire doesn't throw the warning on nested components after the initial request

B seems like a much more urgent fix to me, so I'll first talk about that:
* I'd love to see a fix for that in a separate PR, however, I'd rather not modify core morphdom in this case, and instead just make sure Livewire always sends an "end" comment for subsequent renders of a parent component that contains children

A feels a little iffy, it's just not clear that that's why we're adding that `<!-- if...` bit, and it also feels a bit hacky.
* Is there some way to tell Cloudflare in the settings to not strip comments when minifying?

Going to close this PR for the above reasons. I think restructuring this PR for the solution I described in "A" is the best way forward, thank you!

PR #4154, opened 2021-11-03
Add public property manager

jonassiewertsen:
1️⃣  Is this something that is wanted/needed? Did you create an issue / discussion about it first?
The discussion has been pretty active: https://github.com/livewire/livewire/discussions/3812

2️⃣  Does it contain multiple, unrelated changes? Please separate the PRs out.
No.

3️⃣  Does it include tests, if possible? (Not a deal-breaker, just a nice-to-have)
Yes, I did add some tests.

This PR does add the option to register custom properties. There is already the option to implement the Wireable interface, but you can't use this approach if you're fx using a package and can't access the class you want to support in Livewire. 

This is one example I did test with a statamic.de Entry, which is not supported by Livewire.

First we need to register the custom resolver:
```php
// AppServiceProvider.php
public function boot()
    {
        LivewireProperty::register(CustomProperty::class, CustomPropertyResolver::class);

       // There should also be the option to register multiple resolvers
       LivewireProperty::register([
           CustomProperty::class => CustomPropertyResolver::class,
           AnotherCustomProperty::class => AnotherCustomPropertyResolver::class,
       ]);
    }
 ```
 
 We need the resolver itself:
 ```php
// Our Example Entry Resolver for a Statamic entry.

use Livewire\PropertyHandler;
use Statamic\Facades\Entry;

class EntryResolver extends PropertyHandler
{
    public static function hydrate($value)
    {
        $entry = Entry::find($value['id']);

        foreach($value as $key => $item) {
            $entry->set($key, $item);
        }

        return $entry;
    }

    public function dehydrate($value)
    {
        return array_merge([
          'id' => $value->id(),
          'slug' => $value->slug(),
          'title' => $value->title,
        ], $value->values()->all());
    }
}
```

That's the way it can be used. 

A Livewire component would look as expected
```php
class Test extends Component
{
    public Entry $entry;

    public function mount()
    {
        $this->entry = Facades\Entry::find('30cf66b5-2c0a-4061-b29f-0ed240126897');
    }

    public function save()
    {
        $this->entry->save();
    }
 }
```

A little video from my Statamic Demo.
https://user-images.githubusercontent.com/38906163/140199176-c5dd3550-5b7a-45c5-8978-5314e5567af1.mov

Needs to be done before the PR is ready:
- [x] Add a Facade
- [x] Add the option to add an array of multiple resolvers
- [x] Add validation
- [x] Add a test to make sure that a registered class can be overwritten
- [x] Integration of the update method in the Property Handler class.


joshhanley:
Looking good! 

Can you also add validation in as well - see this PR for Wireable #4036 - as the validation system needs the objects in certain format (array I think) to work properly, so we will need a way to allow custom properties to implement a `validate` method (even if that is just deferring to the `dehydrate` method).

https://github.com/livewire/livewire/blob/4a349459fd9c8e2bbeae2306d7fd97ba424de006/src/ComponentConcerns/ValidatesInput.php#L305-L310

---

I'm personally not a fan of the abstract class and don't think it should be enforced, instead just doing a method exists check on the resolver would be better. That way, if update is not implemented for example, then there is no issues and just Livewire's default update process will be followed.

My thoughts were that if needed then you could just implement a hydrate method, or just an update method, whatever is required for that specific property. Any of the other methods would just fall through and keep working down the conditionals in `HydratePublicProperties.php`.

I think we can still have an interface though that has all 4 methods (hydrate, validate, update, dehydrate) if people want to use it, but that would mostly be for IDE benefit, so it you can just click implement and it will pre-populate the method stubs for you.

---

Re: the eloquent model stuff, maybe it would be good, when doing the check for existance and returning the resolver, that we check the array in reverse order. That way later bindings have preference over earlier ones. That way if you have a custom resolver for eloquent models bound first, and then a specific one for `PostResolver` then the post resolver is checked first.


Great job! Loving this feature and your implementation looks so clean. Looking forward to being able to use it! 😁

src/Hydration/PublicPropertyManager.php
```diff
@@ -9,13 +9,13 @@ class PublicPropertyManager
 {
     public $properties;
 
-    public function register($class)
+    public function register($class, $resolver)
     {
-        if (! $class instanceof Wireable) {
+        if (! $resolver instanceof Wireable) {
             throw new CannotRegisterPublicPropertyWithoutImplementingWireableException(get_class($class));
```
edalzell: `throw_if`?

jonassiewertsen: Yep, that would be even nicer.

src/Hydration/Middleware/AddAttributesToRootTagOfHtml.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace Livewire\HydrationMiddleware;
+namespace Livewire\Hydration\Middleware;
```
joshhanley: Can you move all these back to their original directory?

src/Hydration/PublicPropertyManager.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Livewire\Hydration;
+
+use Livewire\Exceptions\CannotRegisterPublicPropertyWithoutImplementingWireableException;
+use Livewire\Wireable;
+
+class PublicPropertyManager
```
joshhanley: Can you move this to the base Livewire directory `src` and rename to `LivewirePropertyManager`?

jonassiewertsen:
I will add the validation as soon as possible but wanted to show this early so we can discuss a few things and I think it might be easier here than in the discussion (#3812) itself. 

I thought it was a great idea to enforce the abstract class, as you would get a nice and clear error message if not extending that class. I totally see your point, but am not sure how to change it that way inside the `HydratePublicProperties` class. 

Eloquent:
Let me add a test to make sure that later bindings have preference over earlier ones, as described in your use case.

src/Hydration/PublicPropertyManager.php
jonassiewertsen: This is pretty old and already refactored.

jonassiewertsen: I think it's the way you do like it Erin :-)

https://github.com/livewire/livewire/pull/4154/files#diff-f5ec0a6e40cd9166462c2bc3c7c033aeab3ad2486b87901108c9ec202ab7f004R19

joshhanley (changes requested):
Edit: Github was showing old code. Has refreshed now so this comment was irrelevant. Have removed it's contents.

joshhanley:
> I will add the validation as soon as possible but wanted to show this early so we can discuss a few things and I think it might be easier here than in the discussion (#3812) itself.

Yep makes sense and yeah it's good because we can use the review feature of a PR, and hopefully get some input from Caleb.

> I thought it was a great idea to enforce the abstract class, as you would get a nice and clear error message if not extending that class. I totally see your point, but am not sure how to change it that way inside the `HydratePublicProperties` class.

I'll have a look (it may not make sense if we do end up removing that conditional completely if we decide to refactor the existing properties implementations).

> Eloquent: Let me add a test to make sure that later bindings have preference over earlier ones, as described in your use case.

Ok great 🙂

src/Hydration/Middleware/AddAttributesToRootTagOfHtml.php
jonassiewertsen: I think you are seeing an old version. It has already been moved.

src/Hydration/PublicPropertyManager.php
jonassiewertsen: This seems to be outdated as well, but I'll take a look.

joshhanley: Github was showing old code. Has refreshed now so this comment is irrelevant.

src/Hydration/Middleware/AddAttributesToRootTagOfHtml.php
joshhanley: Github was showing old code. Has refreshed now so this comment is irrelevant.

joshhanley (commented):
Should we also add a facade like the Livewire one https://github.com/livewire/livewire/blob/master/src/Livewire.php?

Maybe the facade could be just `LivewireProperty`, so the in a service provider it would look like

```php
LivewireProperty::register(CustomProperty::class, CustomPropertyResolver::class);
```

jonassiewertsen:
Yes, we should! I'll add it to my to do list :-)

jonassiewertsen:
I will as well add the option to add multiple resolvers at once, so that you can register your properties like this.

```php
public function boot()
    {
        LivewireProperty::register(CustomProperty::class, CustomPropertyResolver::class);

       // There should also be the option to register multiple resolvers
       LivewireProperty::register([
           CustomProperty::class => CustomPropertyResolver::class,
           AnotherCustomProperty::class => AnotherCustomPropertyResolver::class,
       ]);
    }
```

jonassiewertsen:
@joshhanley 
- I added a Facade which is much nicer.
- Multiple properties can be registered at once.
- Validation has been implemented.
- Refactored some method names:
  - `LivewireProperty::exists($class)` instead of `LivewireProperty::has($class)`.
  - `LivewireProperty::properties()` does stay, but I did add `LivewireProperty::all()` as an alias.
  
I do need help on how we should call the update method inside the class PropertyHandler.
  
I did not change that the abstract class `PropertyHandler` is being forced if registering a property. Personally I do like it, but that's not the reason for leaving it as it is.
This adjustment would need some changes in the `src/HydrationMiddleware/HydratePublicProperties.php` and would mean to touch how we determine on how to hydrate and dehydrate all properties. I couldn't find a clean way of implementing it, without making this PR pretty huge. I am open to ideas of course!

joshhanley:
@jonassiewertsen nice, great work!

Really like the new changes 🙂

Ok, no worries, I will have a look at that hopefully tomorrow and see how to handle it.

Will also have a quick look at the abstract class and the hydrate methods. It may be that it's the best way.

joshhanley:
@jonassiewertsen so the update happens in `ComponentConcerns/HandlesActions.php` inside the `syncInput` method.

So I think you would need to add a check for the existence of a resolver and then use it if it exists and has an `update` method on it.

What to pass into the update method, will probably need to be the full key, the current data and the update value. Then the update method then returns a value and that value gets assigned to the component property.

Something like this

```php
$propertyType = (new \ReflectionClass($this))
    ->getProperty($propertyName)
    ->getType()
    ->getName();

if ($propertyHandler = LivewireProperty::get($propertyType) && method_exists($propertyHandler, 'update')) {
    $originalData = data_get($this, $propertyName);

    // Name and value are what is passed into the syncInput method
    $updatedData = $propertyHandler->update($name, $originalData, $value); 

    data_set($this, $propertyName, $updatedData);
}
```

Hopefully that helps, let me know if you need any more info.

I personally preferred your `->has` method over `->exists` (but I guess `has` only makes sense when calling it like `LivewirePropertyManager::has`, where as maybe exists makes more sense when calling `LivewireProperty::exists`).

Also one thing with how the manager resolves a handler, I think it might be better to change to `is_a()` function inside the `contains` check, so that way we can check for specificity, so like

```php
use Illuminate\Database\Eloquent\Model;

LivewireProperty::register(Model::class, ModelResolver::class);

// Returns true as it `is_a` subclass of Model - and get would return ModelResolver
LivewireProperty::has(User::class);

// But if we also had this registered after ModelResolver
LivewireProperty::register(User::class, UserResolver::class);

// Returns true as it `is_a` class of User - and get would return UserResolver if registered after ModelResolver
LivewireProperty::has(User::class);
```

Maybe it would be better to change has/exists to make use of the get method for it's check. So the get method will iterate through the registered handlers in the reverse order from latest registered to first, and return the first one that matches using `is_a`, and returns null if nothing found.

Then the has/exists method can just be `return (bool) $this->get($class);`

Hope this helps!

joshhanley (commented):
@jonassiewertsen so the update happens in `ComponentConcerns/HandlesActions.php` inside the `syncInput` method.

So I think you would need to add a check for the existence of a resolver and then use it if it exists and has an `update` method on it.

What to pass into the update method, will probably need to be the full key, the current data and the update value. Then the update method then returns a value and that value gets assigned to the component property.

Something like this

```php
$propertyType = (new \ReflectionClass($this))
    ->getProperty($propertyName)
    ->getType()
    ->getName();

if ($propertyHandler = LivewireProperty::get($propertyType) && method_exists($propertyHandler, 'update')) {
    $originalData = data_get($this, $propertyName);

    // Name and value are what is passed into the syncInput method
    $updatedData = $propertyHandler->update($name, $originalData, $value); 

    data_set($this, $propertyName, $updatedData);
}
```

Hopefully that helps, let me know if you need any more info.

I personally preferred your `->has` method over `->exists` (but I guess `has` only makes sense when calling it like `LivewirePropertyManager::has`, where as maybe exists makes more sense when calling `LivewireProperty::exists`).

Also one thing with how the manager resolves a handler, I think it might be better to change to `is_a()` function inside the `contains` check, so that way we can check for specificity, so like

```php
use Illuminate\Database\Eloquent\Model;

LivewireProperty::register(Model::class, ModelResolver::class);

// Returns true as it `is_a` subclass of Model - and get would return ModelResolver
LivewireProperty::has(User::class);

// But if we also had this registered after ModelResolver
LivewireProperty::register(User::class, UserResolver::class);

// Returns true as it `is_a` class of User - and get would return UserResolver if registered after ModelResolver
LivewireProperty::has(User::class);
```

Maybe it would be better to change has/exists to make use of the get method for it's check. So the get method will iterate through the registered handlers in the reverse order from latest registered to first, and return the first one that matches using `is_a`, and returns null if nothing found.

Then the has/exists method can just be `return (bool) $this->get($class);`

Hope this helps!

jonassiewertsen:
Your Input has been great and implemented.

iksaku (approved):
Looks awesome!

joshhanley:
@jonassiewertsen awesome! Will have a look 🙂

src/ComponentConcerns/HandlesActions.php
```diff
@@ -20,6 +21,28 @@ public function syncInput($name, $value, $rehash = true)
...
+        if (version_compare(PHP_VERSION, '7.4', '>=')) {
+            try {
+                $propertyType = (new \ReflectionClass($this))
+                    ->getProperty($propertyName)
+                    ->getType();
+            } catch (\Exception $e) {
+                $propertyType = null;
+            }
+
+            $property = $propertyType ? $propertyType->getName() : null;
```
jameslkingsley, line 33: Since PHP 8, `getType()` can either return a `ReflectionNamedType` or `ReflectionUnionType`. The latter containing multiple types. I think it's worth supporting union types, and defaulting the chosen type to the first in the list. Alternatively you could throw an exception if this shouldn't try to guess which type to use?

src/HydrationMiddleware/HydratePublicProperties.php
```diff
@@ -36,9 +36,12 @@ public static function hydrate($instance, $request)
         $modelCollections = data_get($request, 'memo.dataMeta.modelCollections', []);
         $stringables = data_get($request, 'memo.dataMeta.stringables', []);
         $wireables = data_get($request, 'memo.dataMeta.wireables', []);
+        $customProperties = data_get($request, 'memo.dataMeta.customProperties', []);
 
         foreach ($publicProperties as $property => $value) {
-            if ($type = data_get($dates, $property)) {
+            if (in_array($property, $customProperties) && version_compare(PHP_VERSION, '7.4', '>=')) {
+                data_set($instance, $property, LivewireProperty::hydrate($instance, $property, $value));
+            } else if ($type = data_get($dates, $property)) {
```
jameslkingsley, line 44: Maybe this method can be simplified to just use Reflection and defer to the property manager? Something like:

**A) The property has a declared type:**
1. Resolve the property type using Reflection
2. Check if the property manager has this type
    1. Yes — hydrate/dehydrate
    2. No — Check if the incoming value is `null` and whether the property supports `null`
        1. Yes — set as `null`
        2. No — Throw exception

**B) The property _does not_ have a declared type:**
1. Resolve the property type using the value's type
2. Go to A.2

src/PropertyHandler.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Livewire;
+
+abstract class PropertyHandler
```
jameslkingsley, line 5: Could this just be an interface instead if it only contains abstract methods?

tests/Unit/PublicPropertyManagerTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+namespace Tests\Unit;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Str;
+use Livewire\Exceptions\CannotRegisterPublicPropertyWithoutExtendingThePropertyHandlerException;
+use Livewire\LivewireProperty;
+use Livewire\Livewire;
+
+class PublicPropertyManagerTest extends TestCase
```
jameslkingsley, line 11: I think another test that's needed is ensuring that a property of:
```php
public ?Foo $foo = null;
```
can receive `null` correctly. `ReflectionProperty` has a `allowsNull()` method I think.

src/PropertyHandler.php
```diff
@@ -0,0 +1,29 @@
...
+     *  When an update is triggered, a request is sent to the server INCLUDING the last-known component state.
+     *  The server "hydrates" or "deserializes" the component from that state and performs any updates.
+     */
+    abstract public static function hydrate($value);
+
+    /**
+     * In addition to render HTML, Livewire "dehydrates" or "serializes" the component's state
+     * of a public property, so it can be passed to the front-end.
+     */
+    abstract public function dehydrate($value);
```
jameslkingsley, line 17: Just had a thought.. maybe these methods should also be passed the `ReflectionProperty` instance? This way the consumer can leverage PHP 8 attributes in their handler. Being able to do something like this would be awesome:
```php
#[DateFormat('H:i, d M Y')]
public ?Carbon $date = null;
```

and then the handler can check this:
```php
public function dehydrate($value, ReflectionProperty $property)
{
    $property->getAttributes();
    // ...
}
```

src/PropertyHandler.php
Raziqbangulzai: Could this just be an interface instead if we only contains abstract methods?

src/HydrationMiddleware/HydratePublicProperties.php
```diff
@@ -94,6 +97,10 @@ public static function dehydrate($instance, $response)
                 is_bool($value) || is_null($value) || is_array($value) || is_numeric($value) || is_string($value)
             ) {
                 data_set($response, 'memo.data.'.$key, $value);
+            } else if (version_compare(PHP_VERSION, '7.4', '>=') && LivewireProperty::has($value)) {
+                $response->memo['dataMeta']['customProperties'][] = $key;
+
+                data_set($response, 'memo.data.'.$key, LivewireProperty::dehydrate($value));
```
jameslkingsley, line 103: This is only checking the value type to determine if it's a custom property. This means that the following scenario would not work:
```php
public ?Carbon $cancellationDate = null;
```
```blade
<x-input.date wire:model="cancellationDate" />
```

Ideally a custom property should be allowed to be `null` and Livewire should check the defined type using Reflection to know if it's a custom property. The value type is really a last resort if the Reflection check fails.

src/LivewirePropertyManager.php
```diff
@@ -0,0 +1,161 @@
...
+        if ($this->hasNot($class)) {
+            return null;
+        }
+
+        $resolver = $this->properties[$class];
+
+        return $resolver::hydrate($value);
+    }
+
+    private function containsMultipleInstances($args)
```
joshhanley, line 143: Maybe make all these `private` methods `protected`

src/PropertyHandler.php
joshhanley: @jonassiewertsen I'm not sure if Caleb will agree to an interface/abstract class.

I know we've discussed it before, but I think changing over to `method_exists` checks to see if the method exists on a resolver would be better (Caleb does that in his experiment).

Then people can implement as many or little methods as it required for their property type.

I'm ok if you don't change this yet though, until we're ready for Caleb to provide feedback and he can decide which he would prefer.

src/PropertyHandler.php
joshhanley: @jameslkingsley yeah I like it. I think in the interest in keeping the PR focused @jonassiewertsen, maybe we will hold off on this and do it in a seperate PR later once this is merged, where we can flesh out a detailed description of how it can be useful and how it will work with clear examples.

tests/Unit/PublicPropertyManagerTest.php
joshhanley: @jameslkingsley yeah agreed, @jonassiewertsen can you add a test for the above as requested.

tests/Unit/PublicPropertyManagerTest.php
```diff
@@ -0,0 +1,118 @@
...
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        if (version_compare(PHP_VERSION, '7.4', '<')) {
+            $this->markTestSkipped('Skip public property tests if the version is below PHP 7.4');
+        }
+
+        require_once __DIR__.'/PublicPropertyManagerStubs.php';
```
joshhanley, line 21: Maybe move the classes to the end of this test file, instead of including the stubs file here, as that keeps it consistent with Livewire's other tests.

joshhanley (changes requested):
Hey @jonassiewertsen sorry for taking so long!

PR is looking really good to me and it's nearly ready I think🙂

I've added my comments and also commented on some of @jameslkingsley's comments.

I also reviewed Caleb's experimental implementation ([see here](https://github.com/livewire/livewire/tree/refactor-property-hydration))  and @jameslkingsley PR #4565 alongside this and added some ideas from both into my feedback.

Let me know if you want me to clarify anything!

Hope this helps!

src/ComponentConcerns/HandlesActions.php
jameslkingsley: Yeah I agree, union types have no place here. Although the code above needs some tweaking to check if it's union and then throw a useful exception - otherwise it may try to use an array as an object.

src/PropertyHandler.php
jameslkingsley: I changed this idea in my PR to pass the component instance and property name, eg:
```php
public function hydrate($instance, $key, $value)
```
This gives the option to build the reflection instance manually or anything else the developer might find useful.

jonassiewertsen:
Thanks for the feedback Josh and James! I will take a closer look soon to make those changes.

src/PropertyHandler.php
silnarm: +1 for component instance and property name, even better would be to use the same signatures that hydrateModel, hydrateModels, dehydrateModel and dehydrateModels currently use:
```php
function hydrate($value, $property, $request, $instance);
function dehydrate($value, $property, $response, $instance);
```
I have a piece of work in progress that is currently heavily hacking HydratePublicProperties.php, it would be nice to re-implement it with this, but it needs the instance and property name, along with the request/response (for hydration and dehydration respectively).

jameslkingsley:
Is there anything I can do to help move this along? The lack of proper property hydration is a daily struggle at the moment 😬

joshhanley:
@jameslkingsley I think @jonassiewertsen was working on the changes I requested, so hopefully he can give us an update 🙂

jonassiewertsen:
Sorry for the delay. 

For the first reviews Josh has been really busy, now am I. Bad luck I guess :-D

I am trying to find some time as fast as possible.

joshhanley:
No worries! Yeah it happens 🙂

joshhanley:
We're going to close this PR as this has now been built into V3. Thanks everyone for their input and @jonassiewertsen for working on this PR! 🙂

PR #4155, opened 2021-11-04
[Fix] Correct behavior for ampersand (&) characters in query string property values

nullthoughts:

Fixes a bug when property values contain an ampersand character (character is not encoded properly, and breaks query strings). Originally documented by @trev33 in Discussion #4072 

No

Yes

Query strings with property values containing `&` characters will behave property


calebporzio:
Good catch, thanks!

totov:
I'm seeing a similar issue with the `#` character in query strings (v2.7.2). Our setup uses the `#` character as part of an advanced wildcard search, but the `#` character isn't encoded in the URL so when the user navigates back to the form any `#` characters aren't being retained in the parameter.

I'm wondering if this PR fixes this issue too or if I should open a new discussion/issue?

nullthoughts:
> I'm seeing a similar issue with the `#` character in query strings (v2.7.2). Our setup uses the `#` character as part of an advanced wildcard search, but the `#` character isn't encoded in the URL so when the user navigates back to the form any `#` characters aren't being retained in the parameter.
> 
> I'm wondering if this PR fixes this issue too or if I should open a new discussion/issue?

The fix also applies to other characters as well with the switch to the `decodeURI()` method. 

Just tested the `#` character in a property value, and it's getting encoded as `%23` in the query string.

PR #4167, opened 2021-11-07
Feature/transitions

ph7jack:
https://github.com/livewire/livewire/discussions/4133

NO

Yes, but I can't add tests to removed elements 😞 

This feature allows adding wire transitions

Add the transition attribute passing the CSS class
```
wire:transition.enter="scale-up-center"
wire:transition.leave="scale-down-center"
```
See working
https://www.loom.com/share/d5896501f52a4df1a8d8027ee5bad665


PH7-Jack:
I think that this functionality can be very useful when adding transitions.

calebporzio:
Thanks for this PR. I DO see the desire for it for sure.

Unfortunately getting transitions to work dependably is a VERY deep topic, and at first glance I don't think this code covers every scenario:
- allowing the transition to be interrupted, cancelled and replayed if the element is added back in before it's done transitioning out
- Because this code doesn't account for the duration of the transition, it appears it won't even properly remove the element at the very end of the transition (requestAnimationFrame won't simply suffice)

Take a look at Alpine's transition system to get a glimpse of what's involved in doing this properly: https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/directives/x-transition.js

Sorry, but I'm going to close this one. Maybe in the future Livewire can have morphdom transitions back.

studiowizjo:
@calebporzio Have you considered including this feature in the new version since it heavily utilizes Alpine?

PR #4178, opened 2021-11-09
[Feature] Query string aliases

amiranagram:
This was discussed in #2496, #2036 and #3305.

No.

Yes.

This PR gives developer the ability to alias (customize/rename) query strings in URL with `as` modifier.

I can think of 3 valid reasons why this is useful.

1. Shortening the URL.

```php
class LivewireComponent extends Component
{
    protected $queryString = [
        'page' => ['except' => 1, 'as' => 'p'],
        'search' => ['except' => '', 'as' => 's'],
    ];
}
```
```
?p=2&s=qwerty
```

2. Completely renaming for semantic purposes.

```php
class LivewireComponent extends Component
{
    public $user;

    protected $queryString = [
        'user' => ['as' => 'user_id'],
    ];
}
```
```
?user_id=42
```

3. Following your standards. For example we can make properties appear in snake case although they are defined in camel case inside the component.

```php
class LivewireComponent extends Component
{
    public $userId;

    protected $queryString = [
        'userId' => ['as' => 'user_id'],
    ];
}
```
```
?user_id=7
```

🙌

Let me know if I have missed something. Cheers.

### DIFF
--- a/src/Features/SupportBrowserHistory.php
+++ b/src/Features/SupportBrowserHistory.php
@@ -178,6 +178,7 @@ class SupportBrowserHistory
                 return isset($value['except']);
             })
             ->mapWithKeys(function ($value, $key) {
+                $key = $value['as'] ?? $key;
                 return [$key => $value['except']];
             });
     }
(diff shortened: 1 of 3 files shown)


calebporzio:
This is great, thanks @amiranagram! Mind PRing the documentation for it to: livewire/docs? Thanks!

amiranagram:
For sure @calebporzio.

Lupennat:
I suppose this pull request contains a bug, everything works only if the query string alias is not valued at the mount of the component. I opened a discussion about this [Here](https://github.com/livewire/livewire/discussions/4475)

Issue #1214, opened 2020-07-03
Can't use server returned error message inside onError callback 


lucasromanojf:
### Description

Hello!

First of all thanks @calebporzio for the onError callback!

The reason of this ticket is that currently onError passes only the status code. But we have situations where we return the error message from backend, and we can have different messages for the same status code.

My suggestion is pass the entire response as argument to onError callback instead of passing only the status code. So we can customize the error message based on what the backend returned.

I can do a PR for this if it is ok. This seems to be simple to fix.

Thanks!

PR #4179, opened 2021-11-09
Pass response as second argument to onError callback

lucasromanojf:

Yes, #1214, #3351 and #3730 

No.

Didn't see it as necessary since it is a very simple PR.

Currently, the onError callback only receives the HTTP status code as argument.

This PR adds the entire response too as the second argument (could be the first argument, but would be a breaking change), so you can use it to show messages returned by the backend, for example.

Here is an usage example:

```
window.livewire.onError(async (status, response) => {
    const json = await response.json();
    alert(json.error);
})
```


calebporzio:
Sounds good. Thank you @lucasromanojf!

PR #4185, opened 2021-11-10
Revert back wire:loading.remove on request error.

lucasromanojf:

Yes, see #3728

No

Yes

Elements with wire:loading.remove were correctly hidden on request start, but it the request throws an exception, the element wasn't being shown again. It was happening because, somehow, when the element had wire:loading.remove, it was being processed twice by the LoadingStates.js logic. This PR fixes this bug by just removing duplicates from the elements array.

🙌


calebporzio:
Thanks for the fix! I test would be ideal here so that if we re implement wire:loading with another core (which we very well may for the next Livewire version) we don't introduce a regression.

Mind adding a test for this? Thanks!

lucasromanojf:
Hey @calebporzio, sure, gonna try to make a test for this today

lucasromanojf:
hey @calebporzio, just updated the PR with the tests.

calebporzio:
Thanks @lucasromanojf!

PR #4188, opened 2021-11-11
Fix validate only custom attributes

joshhanley:
This PR fixes custom validation attribute names (both Livewire ones and Laravel's validation translation file ones), see discussion #3824.

Since 2.6, custom validation attributes were no longer being displayed for nested items when using `validateOnly`.

For example with the following custom attribute

```php
protected $validationAttributes = [
        'users.*.email' => 'email address'
];
```

When running `$this->onlyValidate('users.1.email');`, the error message was still showing "users.1.email" instead of "email address".

The issue was introduced in PR #3596 as that PR was filtering the rules to match the exact property to validate, and changing them from `users.*.email` to `users.1.email`.

The problem with that is the custom attributes still need the rule names to be `users.*.email` to work properly.

To solve this I've change it so it filters which data to pass to the validator and leaves the rules intact.

I started writing the logic for this, then realised we already have it in `HydratePublicProperties`, and so you will notice that I'm making use of the data processing methods from there as it is the same logic. With some small modifications to it, I got it all running.

Not sure if you would prefer to move these methods to another class which then hydrate and validate can make use of.

Hope this helps!

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -14,6 +14,7 @@ use Illuminate\Support\Facades\Validator;
 use Illuminate\Validation\ValidationException;
 use Livewire\Exceptions\MissingRulesException;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
+use Livewire\HydrationMiddleware\HydratePublicProperties;
 
 trait ValidatesInput
 {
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks Josh!!!

mnabialek:
For my app it now causes bug. I have following validation rules:

```
[
  'files' => ...,
  'files.* => ...,
]
```


How can I now run validation only for `files` key (but not `files.*`) ? Previously I could run:

```
$this->validateOnly('files')
```

but it doesn't work anymore - because validation for `files.*` is also running now.

joshhanley:
@mnabialek thanks for reporting! Have a look a discussion #4278 for more details what is going on. I'm currently investigating a fix.

Hope this helps!

PR #4191, opened 2021-11-11
Update html after download

fabio-ivona:

discussed here: https://github.com/livewire/livewire/discussions/4057

added a failing test before fixing it

This PR will allow the UI html to be updated when a called method triggers a download

Suppose you want to display a thanks message or hide the download button after the user has downloaded the file, maybe by setting a `$downloaded = true` property in order to show something like this in the UI:

```blade
@if($downloaded)
    Thanks!
@endif
```

the `Thanks!` message won't show up because `SupportFileDownloads` feature has a `$component->skipRender();` call at line 39

This PR wants to allow the html to update by removing that `$component->skipRender();` call

Thanks to you for this wonderful tool! 🙌


calebporzio:
Yeah, this checks out. Thanks for the addition!

calebporzio:
Sorry, spoke too soon. The reason that's in their is:

If a file download is sent in the response it's presumably a pretty big response. To keep the response from over-bloat by having the full HTML in it as well, Livewire on the backend "skips render":

image

So even with this PR, I think Livewire skips the render on the backend right? So this PR wouldn't even allow a "Thank You" message?

fabio-ivona:
> Sorry, spoke too soon. The reason that's in their is:
> 
> If a file download is sent in the response it's presumably a pretty big response. To keep the response from over-bloat by having the full HTML in it as well, Livewire on the backend "skips render":
> 
> image
> 
> So even with this PR, I think Livewire skips the render on the backend right? So this PR wouldn't even allow a "Thank You" message?

Sorry for the delay, it seems to work on my end and it renders the message 🤔

Isn't that `$component->skipRender()` the one I removed?

Even from the tests I added, the Html seems to update with the "Thank You" message

Am I missing something?

calebporzio:
Ooof you're right, you did remove it duh. Thanks.

@joshhanley - before I merge this, is there something I'm missing?

(Summary of change for easy consumption)

Currently: When you trigger a download in Livewire it sends the download data but skips rendering the view.

After this update: there is no more skip of render and the backend will re-render the HTML and send it down.

It makes sense to me just want to check with you and make sure I'm not making an oversight.

Going to merge this in the meantime.

Thanks @fabio-ivona!

joshhanley:
@calebporzio haha I was wondering...

Nope, I think it's all good. If the only reason was to stop HTML being sent with the file, then yeah i think it will be ok.

People can always manually call `skipRender()` if they need.

PR #4203, opened 2021-11-12
Discover index components in subfolders

Naoray:
This PR adds the functionality to discover `index` components in sub-folders if the initial `$alias` for the component could not be resolved. It follows the same logic as Laravel does for finding [Anonymous Index Components](https://laravel.com/docs/8.x/blade#anonymous-index-components).

## Use Case
We have a `Registration` component folder, because in this case a registration consist of multiple steps (username & password in step 1, address data in step 2, etc.). With the current implementation we would have to reference the specific registration component with `registration.index` with this PR we can reference the component by `registration` and also place the related steps (components) inside the same folder.

```blade
<!-- before -->
<livewire:registration.index />

<!-- after -->
<livewire:registration />
```

```php
namespace App\Http\Livewire\Registration;

use Livewire\Component;

class Index extends Component
{
    public function render()
    {
        return view('livewire.registration.index');
    }
}
```

## Advantages
Grouping components that belong to the an "index" component together in a folder without referencing it with a ".index" suffix.

## Tests
I've not added any tests because I am not sure if the idea of this PR is acceptable nor where to place the tests for this small feature. If you point me to a specific file I ll gladly add the missing tests.

### DIFF
--- a/src/LivewireComponentsFinder.php
+++ b/src/LivewireComponentsFinder.php
@@ -23,7 +23,9 @@ class LivewireComponentsFinder
 
     public function find($alias)
     {
-        return $this->getManifest()[$alias] ?? null;
+        $manifest = $this->getManifest();
+
+        return $manifest[$alias] ?? $manifest["{$alias}.index"] ?? null;
     }
 
     public function getManifest()
--- a/tests/Unit/LivewireDirectivesTest.php
+++ b/tests/Unit/LivewireDirectivesTest.php
@@ -190,6 +190,22 @@ class LivewireDirectivesTest extends TestCase
     }
 
     /** @test */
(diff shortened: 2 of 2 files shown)


amiranagram:
@Naoray Check out `TagCompilerTest` in unit tests directory.

Naoray:
@amiranagram I don't think this is related to the `TagCompilerTest` as this does not modify the rendered tag in any way. It only adds another way to receive the given `$alias` from the manifest file.

I have added a test to the `LivewireDirectivesTest`. Let me know if I need to do anything else.

calebporzio:
@Naoray thanks for adding this AND for the tests. @amiranagram thanks for providing guidance here ❤️

I like this feature of course (I think I PR'd it to Laravel)

Let's doooo it!

PR #4211, opened 2021-11-14
Deterministic "cached keys" for nested components

zakini:
No issue or discussion, but this is an issue I ran into personally, [as have a few other people](https://twitter.com/cssgareth/status/1391473632627642369?s=20).

Nope, one single change

Yes, this includes a test that fails without this change

Currently, rendering nested Livewire components when running multiple instances of a Laravel app behind a load balancer can cause Livewire to lose track of the state of all child components.

This is caused by Livewire using random strings to track each child component and storing this string in Laravel's view cache. Since the view cache isn't typically shared between app instances in this setup, Livewire can end up rendering child components from scratch if requests don't consistently reach the same instance of the app.

By making the generation of these strings deterministic, it doesn't matter if the view cache isn't shared. Each instance of the app will generate the same string for a given child component, so child components will be correctly tracked across app instances.

🙇

### DIFF
--- a/src/LivewireBladeDirectives.php
+++ b/src/LivewireBladeDirectives.php
@@ -45,7 +45,7 @@ EOT;
 
     public static function livewire($expression)
     {
-        $cachedKey = "'" . Str::random(7) . "'";
+        $cachedKey = "'" . sha1($expression) . "'";
 
         $pattern = "/,\s*?key\(([\s\S]*)\)/"; //everything between ",key(" and ")"
         $expression = preg_replace_callback($pattern, function ($match) use (&$cachedKey) {
--- a/src/LivewireBladeDirectives.php
+++ b/src/LivewireBladeDirectives.php
@@ -45,11 +45,12 @@ EOT;
 
     public static function livewire($expression)
     {
-        $cachedKey = "'" . sha1($expression) . "'";
+        $userDefinedCachedKey = 'null';
+        $expressionHash = "'" . sha1($expression) . "'";
 
         $pattern = "/,\s*?key\(([\s\S]*)\)/"; //everything between ",key(" and ")"
-        $expression = preg_replace_callback($pattern, function ($match) use (&$cachedKey) {
-            $cachedKey = trim($match[1]) ?: $cachedKey;
+        $expression = preg_replace_callback($pattern, function ($match) use (&$userDefinedCachedKey) {
+            $userDefinedCachedKey = trim($match[1]) ?: $userDefinedCachedKey;
             return "";
         }, $expression);
 
@@ -57,15 +58,18 @@ EOT;
 <?php
 if (! isset(\$_instance)) {
     \$html = \Livewire\Livewire::mount({$expression})->html();
-} elseif (\$_instance->childHasBeenRendered($cachedKey)) {
-    \$componentId = \$_instance->getRenderedChildComponentId($cachedKey);
-    \$componentTag = \$_instance->getRenderedChildComponentTagName($cachedKey);
-    \$html = \Livewire\Livewire::dummyMount(\$componentId, \$componentTag);
-    \$_instance->preserveRenderedChild($cachedKey);
 } else {
-    \$response = \Livewire\Livewire::mount({$expression});
-    \$html = \$response->html();
-    \$_instance->logRenderedChild($cachedKey, \$response->id(), \Livewire\Livewire::getRootElementTagName(\$html));
+    \$cachedKey = $userDefinedCachedKey ?: implode('-', [$expressionHash, count(\$_instance->getRenderedChildren())]);
+    if (\$_instance->childHasBeenRendered(\$cachedKey)) {
+        \$componentId = \$_instance->getRenderedChildComponentId(\$cachedKey);
+        \$componentTag = \$_instance->getRenderedChildComponentTagName(\$cachedKey);
+        \$html = \Livewire\Livewire::dummyMount(\$componentId, \$componentTag);
+        \$_instance->preserveRenderedChild(\$cachedKey);
+    } else {
+        \$response = \Livewire\Livewire::mount({$expression});
+        \$html = \$response->html();
+        \$_instance->logRenderedChild(\$cachedKey, \$response->id(), \Livewire\Livewire::getRootElementTagName(\$html));
+    }
 }
 echo \$html;
 ?>
--- a/tests/Unit/NestingComponentsTest.php
+++ b/tests/Unit/NestingComponentsTest.php
@@ -113,6 +113,25 @@ class NestingComponentsTest extends TestCase
         $this->assertStringNotContainsString('foo', $component->payload['effects']['html']);
         $this->assertStringNotContainsString('bar', $component->payload['effects']['html']);
     }
+
+    /** @test */
+    public function parent_tracks_subsequent_renders_of_duplicate_children()
+    {
+        app('livewire')->component('parent', ParentComponentForNestingDuplicateChildrenStub::class);
+        app('livewire')->component('child', ChildComponentForNestingStub::class);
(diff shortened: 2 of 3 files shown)


calebporzio:
Ah nice, this makes sense. My first thought is:

Because the deterministic seed is: `$expression`, what happens if you have two Livewire children that have the exact same expression?  For example, two `<livewire:foo />` right next to each other.

Where before they had unique keys, they are now the exact same right?

Zakini:
I figure that situation is effectively a loop, which already requires you to specify a key for each component.

I imagine you could count the number of child components being rendered and include that in the seed, which might completely remove the need to specify keys manually in loops too, but I wanted to keep this PR simple.

calebporzio:
Yeah, I hear you. Unfortunately, that case isn't considered the same as a loop. Officially Livewire only recommends keying components when they are within a loop, not when they are simply siblings.

This is a tough problem, but I think it really needs to be addressed.

Are there any out-of-the-box ideas to get multiple app instances on the same page?

Obviously sharing cached views would be great, but that would require infra stuff.

Outside of that, it's hard for me to imagine how we can do this effectively.

Maybe by doing something like you're saying that instead of the keys being randomly generated, they are incremented as they are rendered? Not sure

Zakini:
The only options I can see for this are either generating keys deterministically like in this PR, or sharing these keys across app instances using Laravel's cache somehow. I reckon deterministic keys are the better option if we can crack this remaining issue though, since generating a key is likely to be faster than getting/setting things in the cache, and maybe a bit simpler.

I'll have another look at this when I get a chance. I've got an idea of how I'd implement this fix, but I'm less sure on how I'd test it. Maybe rendering a child component twice in the same parent component and checking that the component IDs are different?

Sammyjo20:
Just wanted to post here to say that we've also been experiencing this problem with round-robin load balancers.

Zakini:
@calebporzio I think including the index of the child component has sorted the issue of duplicate child components. I've just added that fix and another test for it.

simple-hacker:
I like @Zakini idea of deterministic cached keys.

> Yeah, I hear you. Unfortunately, that case isn't considered the same as a loop. Officially Livewire only recommends keying components when they are within a loop, not when they are simply siblings.

With this feature you could always make a footnote in the docs explaining on a load balanced environment that all components need to be keyed, and leave it down to the developers?

joshhanley:
In my experience you need to `wire:key` all Livewire components when using multiple servers behind a load balancer.

Doing this will make them deterministic. I think simply adding a note to the docs as suggested by @simple-hacker could be sufficient.

It does mean though that if you have two sibling components (not in a loop), that they will need to be keyed (and differently) but I don't see that as a problem.

calebporzio:
Good call @joshhanley, this IS a solvable problem currently, but like you said there's no documentation for it.

I guess it just pains me to make people add verbosity to their Livewire code because they are on a load balancer (as that's a pretty common case).

I'm open to docs in the meantime, but I'd love to explore solutions, although like I said, my gut says it won't be easy and there's a part of me that thinks it's nearly impossible without explicit systems in place to talk between server instances.

Zakini:
Did you see the new commits I've added @calebporzio? I'm pretty sure this PR now provides deterministic cached keys _and_ doesn't break identical sibling components

bilfeldt:
> I guess it just pains me to make people add verbosity to their Livewire code because they are on a load balancer (as that's a pretty common case).

@calebporzio Also, a lot of 3rd party packages will not remember to put these keys explicitly (which is not currently recommended) so taking an existing application and putting it behind a load balancer can cause stuff to break even if you yourself as the application developer have put explicit keys everywhere (says the guy who just broke my app 😆 )

@Zakini That seems like a neat way of doing it 👍  Maybe we could add a separate test confirming that a views cache key is deterministic? Not sure if this is already implicitly covered by the new tests you have created though.

```php
\$cachedKey = $userDefinedCachedKey ?: implode('-', [$expressionHash, count(\$_instance->getRenderedChildren())]);
```

What a great PR 🥳

Zakini:
@bilfeldt I think the tests cover that implicitly. Testing the key generation directly could definitely be done (by inspecting the keys of the array returned by `Component::getRenderedChildren()`), but the existing tests focus on whether a component has been fully rendered or replaced with a dummy element like `<div wire:key="..."></div>`, so I did the same in these tests.

I agree on your point about 3rd party packages though. There'll be a lot of work for the maintainers of those packages if Livewire goes the route of expecting keys to always be manually specified in load balanced environments 😬

src/LivewireBladeDirectives.php
```diff
@@ -45,27 +45,31 @@ public static function livewireScripts($expression)
...
-} elseif (\$_instance->childHasBeenRendered($cachedKey)) {
-    \$componentId = \$_instance->getRenderedChildComponentId($cachedKey);
-    \$componentTag = \$_instance->getRenderedChildComponentTagName($cachedKey);
-    \$html = \Livewire\Livewire::dummyMount(\$componentId, \$componentTag);
-    \$_instance->preserveRenderedChild($cachedKey);
 } else {
-    \$response = \Livewire\Livewire::mount({$expression});
-    \$html = \$response->html();
-    \$_instance->logRenderedChild($cachedKey, \$response->id(), \Livewire\Livewire::getRootElementTagName(\$html));
+    \$cachedKey = $userDefinedCachedKey ?: implode('-', [$expressionHash, count(\$_instance->getRenderedChildren())]);
```
MrRio, line 62: This is great 👍

calebporzio:
Ok, I reviewed the new strategy (as well as I could just scanning the code) @Zakini - thanks for the updates there.

2 things:
1) Is it possible to document in a comment on this PR what specific blade input would output in cached PHP files. A "pure blade", "pure cached PHP" comparison of a few cases would make it easier for me and others probably to grasp.
2) I like your child-count strategy for the keys if nothing user-generated was passed in. Would this break if a component conditionally rendered/removed a component on a subsequent request? Would it throw off the count and break all the following keys? Or am I miss-understanding something.

Thanks!

Zakini:
Yeah I can expand on this. I'm not sure what you mean by "pure blade" or "pure cached PHP" though. Are you wanting to see how the blade code you write gets transformed when Laravel caches the view?

As for child counting breaking when conditionally rendering components, I'd have to try it out but yes I would imagine that would be an issue. Maybe instead of counting _all_ of the child components we should just count the components with the same hash? That wouldn't completely solve the issue, but at least it'd be limited to child components with the same `$expression`.

There might not be a good solution to this duplicate child component problem though. As I mentioned, it's effectively the same as a loop and there isn't a good solution for tracking components in a loop without specifying keys. I'm happy to try and find one, but it might come down to deciding between supporting load balanced environments or supporting duplicate, un-keyed, child components.

bilfeldt:
Just a thought in regards to

> it might come down to deciding between supporting load balanced environments or supporting duplicate, un-keyed, child components.

Is this problem also relevant for serverless deployments like Laravel Vapor or do they have a shared view cache?

Zakini:
I haven't tried running Livewire in a serverless environment, but it probably would be an issue there too. Laravel doesn't provide any means of changing how/where views are cached so the only option is to have a shared filesystem like [EFS](https://aws.amazon.com/efs/), which AFAIK isn't a 'standard' thing to do in serverless or load balanced deployments

MrRio:
This issue would happen in serverless deployments as well.

What's needed to get this merged?

Zakini:
Did you see [this comment](https://github.com/livewire/livewire/pull/4211#issuecomment-974634810) @calebporzio?

Also, should I @ you in every comment directed at you? I didn't want to spam you too much but maybe I'm not spamming you enough now? 😅

calebporzio:
Thanks, but closing this in favor of #4354

PR #4247, opened 2021-11-18
Fix issue with autofill in Safari.

codywoolsey:
This is a necessary bug fix to resolve issues with handling autofill in Safari. I did not create a discussion about it, but there is currently an open discussion [here](https://github.com/livewire/livewire/discussions/3082) that references the issue.

Nope

Difficult one to add tests for due to the issue only affecting certain versions of Safari, but if you can point me in a direction where tests could be added I'm happy to do so.

Currently, there's an issue with Safari autofill and the `wire:model` attribute. Safari fires a CustomEvent when an input is autofilled, but rather than storing the value in `event.detail` like most browsers, Safari stores the value in `event.target.value`. This is a simple change to attempt to read the value from `event.target.value` before falling back to reading the "value" attribute of the element (which will sometimes be undefined in Safari).

Alpine made a similar fix recently here: alpinejs/alpine#1969 and it seemed appropriate to apply the same fix to Livewire for people who don't wish to use Alpine with Livewire.

I was able to consistently reproduce this issue with the following steps on a simple login form using Safari's autofill feature:

1. Autofill your credentials, this may work properly the first time.
2. Clear out one of the values (username or password).
3. Autofill your credentials again. Whichever input that you didn't initiate the autofill from will be recognized as an undefined value. The input will be cleared, and the property of the component will erroneously be set to "null". 


zepfietje:
Finally a fix for this long-standing issue.
Thanks @codywoolsey!

calebporzio:
Seconding what @zepfietje said, BIG thanks for digging on this one, I'm sure it wasn't painless @codywoolsey. ❤️ 

I understand adding tests for this one is gnarly

codywoolsey:
> @codywoolsey Does this also fix the select issue from #3029?

I believe it does. I'm having trouble actually testing it because Safari really doesn't want to autofill select boxes for me 🙃 But based off your description, I'm fairly certain it's the same issue that this PR addresses.

PR #4239, opened 2021-11-17
[FIX] Check all the values in dehydratePropertyFromWithFileUploads before calling serializeMultipleForLivewireResponse

some-daan:
Referring to this issue: https://github.com/livewire/livewire/discussions/2243

I'm using a file upload (using Livewire fileuploads) functionality with sorting (using Livewire/sortable). I ran into the same issue as the creator of the issue mentioned above. Once you start dragging a freshly added image to the first position in the list an error pops up. There's [a video](https://drive.google.com/file/d/10sVkwguH_b0xWuzdpPNvwqeaCof_8rJp/view) in the mentioned issue that shows you how to reproduce this bug.

This issue only happens when a newly uploaded image gets dragged to the first position. This is because the $value array would look like this:

image

The _dehydratePropertyFromWithFileUploads_ function only checks in the third if-statement if the first item of this array is an instance of _TemporaryUploadedFile_ while that doesn't necessarily mean that this is a valid array.

The _serializeMultipleForLivewireResponse_ calls the _getFilename_ on all of the items in the array, which will fail for items that aren't an instance of _TemporaryUploadedFile_. 

This pull request fixes this issue by not allowing invalid arrays to call the _serializeMultipleForLivewireResponse_ function, but instead let's them pass on to the fourth if-statement of the _dehydratePropertyFromWithFileUploads_ function. An invalid array meaning that not all items in the array are an instance of _TemporaryUploadedFile_

PR #4248, opened 2021-11-19
Wireable Image Dehydration in SupportFileUpload

stevenmond:
This was mentioned as part of https://github.com/livewire/livewire/pull/4036

No, although I incorporated the code change from https://github.com/livewire/livewire/pull/4239 into this and added a test for the fix it added.

Yes

```SupportFileUploads``` isn't checking for ```Wireable``` objects when ```dehydratePropertyFromWithFileUploads``` is called. As a result, any images that are on ```Wireable``` objects are not serialized at that time. Later on, images serialized as ```{"disk":"public"}``` rather than as ```livewire-file[s]:``` since the disk property is the only public property on ```TemporaryUploadedFile```. This is the same result as running a ```json_encode``` on ```TemporaryUploadedFile```.

 This fix iterates through all the keys on the ```Wireable``` object checking for ```TemporaryUploadedFiles``` and updating the value as appropriate.

Hydration seems to work out of the box from what I've seen and didn't require any changes.


joshhanley:
@stevenmond should we mark this as a draft for now, and if #4239 gets merged then we will rebase this onto master and then mark it as ready?

joshhanley:
@stevenmond yeah that'd be preferred if you can keep it independent.

But if you could share your test code on that PR so they can add it, that would be much appreciated! 🙂

stevenmond:
@joshhanley Updated as requested. Will share test code to #4239's PR.

joshhanley:
@stevenmond thanks!

calebporzio:
Ok, merged the other PR - will merge this once the conflicts are resolved and everyone's happy with it. Thanks!

stevenmond:
@joshhanley I've resolved the conflicts with the test that were committed as part of #4239. Let me know if there is anything else you need me to do on my end.

joshhanley:
@stevenmond thanks! All looks good to go.

calebporzio:
Great, thanks!

PR #4253, opened 2021-11-21
Add query string sort mode

ariaieboy:
https://github.com/livewire/livewire/discussions/2722
no
no
as described in the discussion that i mentioned google have some rules about duplicate pages and urls and its contains the urls with query strings for example if we add 2 pagination in a single page we can create 2 url for it :
```
/posts?page=2&comment=4

/posts?comment=4&page=2
```
in the example above we have 2 url for a posts page with that paginate posts and comments and it have the same content and google dont like this duplicate urls.

I've added a simple boolean flag in livewire config file so users can forcing the query string sort option.

it's use alphabetic sorting because in a page we may have several livewire component and maybe alot of user defined or non-livewire defined query strings so the best choice here is to sort the keys. 


ariaieboy:
in my repo all tests pass without any problem.
https://github.com/ariaieboy/livewire/actions/runs/1485897101

calebporzio:
Thanks, can you add a test for this? Something that would fail before this addition is made?

ariaieboy:
> Thanks, can you add a test for this? Something that would fail before this addition is made?

actually i've added a test for my PR.
but i couldn't change the config on run time :| i will add the test but it will not pass because the default config is set to false

joshhanley:
@ariaieboy looks like there is an issue with the script assertion.

image

ariaieboy:
> > Thanks, can you add a test for this? Something that would fail before this addition is made?
> 
> actually i've added a test for my PR. but i couldn't change the config on run time :| i will add the test but it will not pass because the default config is set to false

@joshhanley i've said in this reply that i tried to fix this test but i couldn't 😑
i think my laravel knowledge is not enough for that 🤔
i've used a config for this so users can enable/disable it if they want.
i added this line of code `Config::set('livewire.force_querystring_sort',true);` to my test that usually works in my phpunit tests but here it not changing currectly.
if you check the config in the test function its set currectly but in the code its not set and the test fail because of that.
if you change the config to true this test will pass but the next test will not.

amiranagram:
@ariaieboy `config()` has no effect because the Dusk test and the app are running in separate processes. A couple things that you can try and see if it works. Though, through my research, unfortunately, the only solution is modify the `.env`, because modifying config does nothing.

1. Modify `.env` on runtime. Though I'm not sure how this will work on a package level.
2. Modify `$_ENV` before running test.

Both of these solutions include telling your `force_querystring_sort` config key to look for an env variable first. I'm not sure if @calebporzio will like that.

joshhanley:
@ariaieboy ah sorry, I just saw the failing test and investigated it, I missed your comment!

Yeah you can't change the config inside the test method, as Dusk launches a whole new request. But you should be able to change it in the component.

joshhanley:
@ariaieboy have pushed a commit to the PR, it adds a dedicated component for this test, so we can set the config in the mount method.

So hopefully the tests should run and pass now.

Hope this helps!

ariaieboy:
@joshhanley yeah it fixed the test 
i really need start learning browser testing 😊

joshhanley:
@ariaieboy great! Yeah I only started learning it when I started contributing to Livewire 🙂 definitely worth learning as it's great for testing things that have javascript and php like Livewire.

calebporzio:
Thanks for this, however I'm going to close this as we hopefully will be removing a lot of the Livewire management of query strings into Alpine when we get Livewire 3 going.

Thanks for putting in the effort on this though!

PR #4279, opened 2021-11-24
Adds prefix to the `payload id`

mertasan:
Yes - No

No

No

I have test macros that check the return values of the methods. After this #3510 change it was not possible to test the return values. If the prefix is added, it will be easy to parse.

This PR makes it possible to find the names of the methods.

### DIFF
--- a/src/Testing/Concerns/MakesCallsToComponent.php
+++ b/src/Testing/Concerns/MakesCallsToComponent.php
@@ -139,6 +139,10 @@ trait MakesCallsToComponent
     {
         $payload['id'] = Str::random(4);
 
+        if (array_key_exists('method', $payload)) {
+            $payload['id'] = $payload['method'] . '-' . $payload['id'];
+        }
+
         $this->lastResponse = $this->pretendWereSendingAComponentUpdateRequest($message, $payload);
 
         if (! $this->lastResponse->exception) {


joshhanley:
@mertasan thanks for the PR! Can you please provide some code examples of what your test macros do and how this fixes it?

Thanks!

mertasan:
Of course. @joshhanley 

I was already to make a new PR with the code example below. (After improving the method for Livewire.) I'm waiting for this PR result.

If merged I will make a PR for `assertReturnEquals`. What is your opinion about this?

```php

public function assertReturnEquals(): Closure
{
    return function (string|array $method, $expected, $message = '') {
        $method = Arr::wrap($method);
        $methodName = $method[0];
        $this->call(...$method);

        // If the same method is executed more than once, sorting to find the last one.
        $returns = array_reverse($this->payload['effects']['returns'] ?? []);
        $found = false;
        $actual = null;
        foreach ($returns as $key => $val) {
            if (Str::startsWith($key, $methodName . '-')) {
                $actual = $val;
                $found = true;
                break;
            }
        }

        if ($found) {
            PHPUnit::assertEquals($expected, $actual, $message);
        } else {
            PHPUnit::fail('Method [' . $methodName . '] not executed.');
        }

        return $this;
    };
}
````

ss2

ss

**Comment by mertasan** [comment]
Thank you! @joshhanley 😇

Yes, this is a good alternative.

Actually, I fixed this problem by using as in screenshots.

I can change the macro as you said.

But not everyone has to do this. Or no need to use extra variables instead of chained methods. 

Rather than using temporary variables, I find `assertReturnEquals` more useful.

Less complicated :)

```php
Livewire::test(SomeComponent::class)->assertReturnEquals('getName', 'foo');
Livewire::test(SomeComponent::class)->assertReturnEquals(['getUser', 1], 'bar');
// or
Livewire::test(SomeComponent::class)->assertReturn('getName', 'foo'); // assertSet's cousin :)
```

joshhanley:
@mertasan cool!

Yeah a method that make it easier could be nice! Maybe there is a better way to achieve that then adding a prefix to the payload id?

joshhanley:
@mertasan hmm yeah not sure, might have to wait for Caleb to weigh in on this, to see if it's something he is willing to support first.

calebporzio:
Thanks for the contribution, however, I think I'm going to pass on this one for now. Thanks for being willing to contribute!

PR #4283, opened 2021-11-25
@js fix for escape character

tanthammar:
For cases like:
```blade
@js('\\nested\\path')
```
currently we get
```
'\nested\path'
```
but with this fix we get a string that  avoids JavaScript syntax errors.
```
'\\nested\\path'
```
creds: @derekmd

### DIFF
--- a/src/LivewireBladeDirectives.php
+++ b/src/LivewireBladeDirectives.php
@@ -25,7 +25,7 @@ EOT;
     if (is_object({$expression}) || is_array({$expression})) {
         echo "JSON.parse(atob('".base64_encode(json_encode({$expression}))."'))";
     } elseif (is_string({$expression})) {
-        echo "'".str_replace("'", "\'", {$expression})."'";
+        echo "'".str_replace("'", "\'", str_replace('\\', '\\\\', {$expression}))."'";
     } else {
         echo json_encode({$expression});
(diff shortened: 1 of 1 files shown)


joshhanley:
@tanthammar thanks for the PR!

I think rather than fixing Livewire's one, we should probably just make use of Laravel's one by adding a version compare as described here https://github.com/laravel/framework/pull/39522

If anyone runs into this bug then, we can just advise them to update their Laravel version.

What do you think?

tanthammar:
@joshhanley 
I submitted this PR because Laravel `Js::from` doesn't work. They changed it before the release to something unusable.
I am sure it is going to be resolved but atm Livewires `@js` helper is the only thing working.

tanthammar:
@joshhanley my bad w the Laravel Js::from. It's working. 
But, it wouldn't hurt if this was fixed anyway for Ither Laravel versions + I prefer @js() syntax. :)

joshhanley:
@tanthammar all good 🙂 yeah fair enough re other Laravel versions.

calebporzio:
Thanks @tanthammar, I'd rather just add support for Laravel's `@js` by detecting the Laravel version in Livewire.

If you'd like to PR that that'd be great!

PR #4316, opened 2021-11-30
Avoid 500 server error when component gets request with empty body

marispro:
Avoid 500 server error when component gets request with empty body 
https://github.com/livewire/livewire/discussions/4262

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -369,7 +369,7 @@ HTML;
                 return str(request('fingerprint')['url'])->after(request()->root());
             }
 
-            return request('fingerprint')['path'];
+            return request('fingerprint.path');
         }
 
         return request()->path();
@@ -386,7 +386,7 @@ HTML;
                 return 'GET';
             }
 
-            return request('fingerprint')['method'];
+            return request('fingerprint.method', 'POST');
         }
 
         return request()->method();


calebporzio:
Can you add a test for this. I'm unclear

marispro:
Need a bit assistance with it - following method does not allow to make request without defined data,
it assuming that every request always have fingerprint, serverMemo & updates variables. 
I am not sure how to correctly make request test to the component.

https://github.com/livewire/livewire/blob/3d0ce6c320bbfdfb6488f43033525bf607a5e613/src/Testing/TestableLivewire.php#L176-L187

However in reality everyday I am getting many empty requests to the component.

joshhanley:
@marispro yeah I'm not sure how we can test it exactly. Have you been able to find out where the empty requests are coming from? Do you know if they are coming from logged in users or do you think it's just someone randomly hitting the Livewire endpoint?

marispro:
@joshhanley I have no idea but relying on IP seems the come from some crawlers.

joshhanley:
@marispro ah right makes sense. I wonder then if you can create a test with a standard Laravel get request to the `/livewire/message` endpoint and assert it doesn't throw the 500 error?

ralphjsmit:
I'm also getting errors like this and they are certainly not caused by a bot. I'm maintaining a time-tracking application where employees can clock-in every day (for greenhouses etc.). Personally on my own devices everything goes smoothly. However, since I did a new deploy last Saturday, another person did get errors.

I still personally can't reproduce it, but one of the errors is exactly the same as described here.

I can share a few Honeybadger links via email with you @joshhanley (or @calebporzio), would that be helpful?

joshhanley:
@ralphjsmit interesting, I wonder what could be causing it!

Yeah that'd be great, do you use discord? If so sending there would be better than email - my username is joshhanley#6022 or you can find me though the Livewire discord if that's easier.

calebporzio:
Thanks!

PR #4338, opened 2021-12-03
Add pest flag to commands

ariaieboy:
NO
NO
YES
https://github.com/pestphp/pest is an awesome test framework for php and i use it in all my projects. 
since the file structure of pestphp is diferrent from phpunit i've added a new flag to livewire commands so we can create-move-copy-delete test files for pest. 
laravel added this flag too :
https://github.com/laravel/framework/pull/38966


ariaieboy:
https://github.com/ariaieboy/livewire/actions/runs/1533149783

src/Commands/livewire.pest.stub
```diff
@@ -0,0 +1,9 @@
+<?php
```
calebporzio, line 1: There should be a line-break after between <?php and "use"

src/Commands/MakeCommand.php
```diff
@@ -44,6 +45,8 @@ public function handle()
 
         if ($test) {
             $test = $this->createTest($force);
+        }elseif($pest){
```
calebporzio, line 48: there should be more standard spacing here: `} elseif() {`

src/Commands/MakeCommand.php
```diff
@@ -122,11 +129,29 @@ protected function createTest($force = false)
         return $testPath;
     }
 
+    protected function createPest($force = false)
```
calebporzio, line 132: I'm OK with this, but I think it would be better to add a flag as a second param to `createTest()` because there is only one small difference between these two methods. I think abstracting the change is the best way here.

src/Commands/MakeCommand.php
```diff
@@ -60,6 +63,10 @@ public function handle()
                 $test && $this->line("<options=bold;fg=green>TEST:</>  {$this->parser->relativeTestPath()}");
             }
 
+            if ($pest) {
```
calebporzio, line 66: I think `if ($pest) {` is redundant/unneeded here, we already have `$pest &&` below it.

src/Commands/ComponentParser.php
```diff
@@ -53,9 +53,9 @@ public function component()
     public function classPath()
     {
         return $this->baseClassPath.collect()
-            ->concat($this->directories)
-            ->push($this->classFile())
-            ->implode('/');
+                ->concat($this->directories)
```
calebporzio, line 56: Please don't add these spaces here.

src/Commands/MakeCommand.php
calebporzio: Ah I see there is also a difference in calling `->pestContests()` instead of `testContents`. I'm more OK with this then.

src/Commands/MoveCommand.php
```diff
@@ -30,7 +30,7 @@ public function handle()
         $class = $this->renameClass();
         if (! $inline) $view = $this->renameView();
 
-        $test = $this->option('test');
+        $test = $this->option('test') || $this->option('pest');
```
calebporzio, line 33: I see you're adding support for pest here, but does this actually support renaming the pest test? Have you tested this yourself locally? It would appear that this triggers the normal test renaming code later on in this file. Can you confirm? Otherwise, support the rename?

calebporzio: This same comment applies to the other non-make commands.

calebporzio: I'm guessing it's because the pest test that is generated is the same in terms of the file name, however, I just want to make sure you've tried all of these locally and can confirm first. Thanks.

calebporzio (commented):
Ok, thanks for this - I've added some comments for you to review before I go further on this PR. Thanks!

calebporzio:
Closing this due to inactivity to my response. Pleases re-open if you've addressed the suggestions. Thanks!

PR #4354, opened 2021-12-07
Fix Livewire requests across load balancers

calebporzio:
# The Problem:

Currently, if a Livewire app is served across load balancers, the following situation may occur:
* A user loads a page with a Livewire component (that contains child Livewire components) on it
* That user does something on the page that triggers a new request to the server
* The load balancer picks up the request and routes it to a DIFFERENT server than the one that originated the page load
* The child components that were rendered on the page are re-rendered (and potentially other odd behavior occurs)

The reason this happens is because Livewire uses Laravel's view cache files as a sort of backend store to set unique keys on each child component in a Blade view. These "unique keys" are generated as random strings. This is how Livewire tracks whether to render, or leave alone a child Livewire component on subsequent renders. In a load-balancer situation, there are two different sets of Blade view cache files, and therefore two totally different sets of keys, creating a mis-match.

## The Solution:

The solution is to deterministically generate those keys for tracking child components rather than randomly generate them. This way, they will be the same across different application instances (across servers in this case).

The new strategy for key generation is this:
-> `[a hash generated from the blade view path]-[a counter that is incremented every time @livewire appears in a view]`

Because the blade files are the same across app instances, these keys will be deterministic.

## Notes:
It's actually fairly tough to get access to the current Blade view path from a custom Blade directive (@livewire), so I had to overload a core function in Laravel's compiler engine: `->get()`.

I don't think this method changes very much so we should be safe, but just to be sure, I forward the method calls to the parent when not inside a Livewire component, so the effects of this change should be minimal outside Livewire components in people's apps. Not my favorite thing to do, but I don't see another, simpler, way.

### DIFF
--- a/src/ComponentConcerns/RendersLivewireComponents.php
+++ b/src/ComponentConcerns/RendersLivewireComponents.php
@@ -6,6 +6,7 @@ use Illuminate\Auth\Access\AuthorizationException;
 use Illuminate\View\Engines\CompilerEngine;
 use Illuminate\View\Engines\PhpEngine;
 use Livewire\Exceptions\BypassViewHandler;
+use Livewire\LivewireManager;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Throwable;
@@ -27,6 +28,39 @@ trait RendersLivewireComponents
     public function isRenderingLivewireComponent()
     {
         return ! empty($this->livewireComponents);
+    }
+   
+    public function get($path, array $data = [])
+    {
+        if (! $this->isRenderingLivewireComponent()) return parent::get($path, $data); 
+
+        $this->lastCompiled[] = $path;
+
+        // If this given view has expired, which means it has simply been edited since
+        // it was last compiled, we will re-compile the views so we can evaluate a
+        // fresh copy of the view. We'll pass the compiler the path of the view.
+        if ($this->compiler->isExpired($path)) {
+            // @note: this is the only modification of this overladed Laravel method:
+            // We are globally setting the current view path being compiled for
+            // reference from the @livewire Blade directive.
+            LivewireManager::$currentCompilingViewPath = $path;
+            LivewireManager::$currentCompilingChildCounter = 0;
+            
+            $this->compiler->compile($path);
+
+            // Here, we'll reset them back, for the next view to be compiled.
+            LivewireManager::$currentCompilingViewPath = null;
+            LivewireManager::$currentCompilingChildCounter = null;
+        }
+
+        // Once we have the path to the compiled file, we will evaluate the paths with
+        // typical PHP just like any other templates. We also keep a stack of views
+        // which have been rendered for right exception messages to be generated.
+        $results = $this->evaluatePath($this->compiler->getCompiledPath($path), $data);
+
+        array_pop($this->lastCompiled);
+
+        return $results;
     }
 
     protected function evaluatePath($__path, $__data)
--- a/src/LivewireBladeDirectives.php
+++ b/src/LivewireBladeDirectives.php
@@ -52,7 +52,7 @@ EOT;
         // Livewire children are properly tracked across load balancers.
         if (LivewireManager::$currentCompilingViewPath !== null) {
             // $cachedKey = '[hash of Blade view path]-[current @livewire directive count]'
-            $cachedKey = "'" . crc32(LivewireManager::$currentCompilingViewPath) . "-" . LivewireManager::$currentCompilingChildCounter . "'";
+            $cachedKey = "'l" . crc32(LivewireManager::$currentCompilingViewPath) . "-" . LivewireManager::$currentCompilingChildCounter . "'";
 
             // We'll increment count, so each cache key inside a compiled view is unique.
             LivewireManager::$currentCompilingChildCounter++;
--- a/src/LivewireBladeDirectives.php
+++ b/src/LivewireBladeDirectives.php
@@ -47,6 +47,17 @@ EOT;
     {
         $cachedKey = "'" . Str::random(7) . "'";
 
+        // If we are inside a Livewire component, we know we're rendering a child.
+        // Therefore, we must create a more deterministic view cache key so that
+        // Livewire children are properly tracked across load balancers.
+        if (LivewireManager::$currentCompilingViewPath !== null) {
+            // $cachedKey = '[hash of Blade view path]-[current @livewire directive count]'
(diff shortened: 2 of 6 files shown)


Zakini:
Hey @calebporzio, I spotted this PR and figured I'd weigh in on it since it tackles the same problem as #4211.

Child components are tracked on their parent component, so the keys used don't need to be globally unique, they just need to be unique within that parent component. If I'm reading these changes right, the blade file path used in the key is the path to the blade file of the parent component. This means that every child component for a given parent component will have the same hash at the start of their key. This leaves only the counter part of the key to make these keys unique.

This effectively means these changes only track child components based the order in which they're rendered. This could causes issues when conditionally rendering child components, since this can change the render order. For this reason, I don't think the key generation strategy is as robust as the one in #4211.

calebporzio:
Thanks for you PR @Zakini, however, this PR takes a different approach and as such actually protects against things like messing up because of conditionally rendered children.

This PR generates the key during Blade's compilation phase which only regexes for Blade syntax and doesn't actually run PHP. Because of this, conditionals like `@if` are ignored and the Livewire components inside of them receive a key no matter what.

I think this is the most robust way to move forward, thank you for putting effort into this, but I think this PR will be the one we'll go with.

Thanks again!

Zakini:
Ah I see, yeah that's a better way of counting components. I think my point about all sibling components having the same hash at the start of their key is still valid, but that's probably a moot point since counting components like this would work on its own.

bberlijn:
Could it be that this issue has been re-introduced with v3?
When I have Stickiness turned off on my load balancer, I get "component not found" errors in the browser console. When I then turn Stickiness on, everything seems to work fine and I don't get any errors.

> Uncaught (in promise) Component not found: Q0IbLqWpgKL3bT5HB0LR [livewire.esm.js:5911]

joshhanley:
@bberlijn yes, possibly. The request handling was changed in V3, so maybe it was broken.

potsky:
> Could it be that this issue has been re-introduced with v3? When I have Stickiness turned off on my load balancer, I get "component not found" errors in the browser console. When I then turn Stickiness on, everything seems to work fine and I don't get any errors.
> 
> > Uncaught (in promise) Component not found: Q0IbLqWpgKL3bT5HB0LR [livewire.esm.js:5911]

Same problem with mailcoach package from Spatie which used LW3…

hackerESQ:
Seeing this on Vapor using Mailcoach v7

seabass86:
I am experiencing the same issue. Has anyone found a fix for this? Or does a new issue need to be opened? @calebporzio 
Thanks!

hackerESQ:
@seabass86  The only editor that's working for me is textarea and creating templates using Unlayer. This is _very_ frustrating. 

I don't believe it's a livewire issue either. There would be WAY more people having similar issues with Livewire + Vapor which I'm just not seeing.

I've narrowed this down to the autosave function and Livewire entangle but beyond that I'm stumped.

freekmurze:
Seems like this bug has been reintroduced in v3. We're having multiple Mailcoach users reporting this issue again.

hackerESQ:
> Seems like this bug has been reintroduced in v3. We're having multiple Mailcoach users reporting this issue again.

Several cases where this is affecting Livewire v3, Mailcoach v7 users on Vapor too. Here's the errors I'm seeing:

```
Uncaught (in promise) Component not found: Ndf8mrr4TxLensZrja7i [livewire.js:4439:39](https://[mailcoach-domain]/livewire/livewire.js?id=f121a5df)
Uncaught Component not found: Ndf8mrr4TxLensZrja7i [livewire.js:4439:39](https://[mailcoach-domain].com/livewire/livewire.js?id=f121a5df)
Uncaught Component not found: Ndf8mrr4TxLensZrja7i [livewire.js:4439:39](https://[mailcoach-domain]/livewire/livewire.js?id=f121a5df)
Uncaught Component not found: Ndf8mrr4TxLensZrja7i [livewire.js:4439:39](https://[mailcoach-domain]/livewire/livewire.js?id=f121a5df)
```

potsky:
Just taking a look on the old PR changes and the actual code...

If you take a look on file `src/Mechanisms/RenderComponent.php`, it seems that the trick with `crc32` to make the deterministic blade file name has been commented and the culprit is then `$key = "'" . Str::random(7) . "'";` because each server generates a different  `key` as in 2021. But I cannot know why it has been done like this.

We really need you @calebporzio.

Alpine and Livewire are awesome (I spend my time to tell that you are totally crazy to have written these both packages) but it is too complicated for me, I don't have the big picture to make a PR without destroying everything, it is really out of my capabilities 😬 

```php
<?php

namespace Livewire\Mechanisms;

...

class RenderComponent extends Mechanism
{
        $key = "'" . Str::random(7) . "'";
...
        // If we are inside a Livewire component, we know we're rendering a child.
        // Therefore, we must create a more deterministic view cache key so that
        // Livewire children are properly tracked across load balancers.
        // if (LivewireManager::$currentCompilingViewPath !== null) {
        //     // $key = '[hash of Blade view path]-[current @livewire directive count]'
        //     $key = "'l" . crc32(LivewireManager::$currentCompilingViewPath) . "-" . LivewireManager::$currentCompilingChildCounter . "'";

        //     // We'll increment count, so each cache key inside a compiled view is unique.
        //     LivewireManager::$currentCompilingChildCounter++;
        // }
...
        return <<<EOT
...
\$__html = app('livewire')->mount(\$__name, \$__params, $key, \$__slots ?? [], get_defined_vars());
...
?>
EOT;
```

nuernbergerA:
If you put a key on every livewire component (also not in loops) on the rendered page it should work without any problems.

potsky:
@freekmurze I think that @nuernbergerA is right. The beginning of the function is

```php
    public static function livewire($expression)
    {
        $key = "'" . Str::random(7) . "'";

        $pattern = "/,\s*?key\(([\s\S]*)\)/"; //everything between ",key(" and ")"

        $expression = preg_replace_callback($pattern, function ($match) use (&$key) {
            $key = trim($match[1]) ?: $key;
            return "";
        }, $expression);
```

So if there is a defined key when calling components, it should be taken into account via `$key = trim($match[1]) ?: $key;`

I will check this this morning on my mailcoach fork, fix the _Alpine is not defined_ bug (mailcoach broken for 1 month now 🤯) and renew my licence for the 3rd time, it expires tomorrow 🤗

calebporzio:
Thanks for the report, I'll look into this stat

calebporzio:
Here's the fix: https://github.com/livewire/livewire/pull/7751

Will be in next release. Thanks!

potsky:
Thanx @calebporzio and @nuernbergerA ❤️

PR #4362, opened 2021-12-08
Set initial property values from query string aliases

CraftLogan:
This is a bug myself and a colleague found while using the query string aliases. If query string aliases params are set during an initial request then they don't set their respective properties unless done manually in the mount function. This pr will fix that:)

### DIFF
--- a/src/Features/SupportBrowserHistory.php
+++ b/src/Features/SupportBrowserHistory.php
@@ -25,8 +25,8 @@ class SupportBrowserHistory
 
             $queryParams = request()->query();
 
-            foreach ($properties as $property) {
-                $fromQueryString = Arr::get($queryParams, $property);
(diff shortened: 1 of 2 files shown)


joshhanley:
@CraftLogan thank for the PR! It looks like this may be impacting some of the other tests.

Any chance you can review them and see if you can work out what is going on?

Also we've just published a contribution guide, so that should help you get the tests running locally https://laravel-livewire.com/docs/2.x/contribution-guide

Let me know if you have any issues getting them running 🙂

Thanks!

CraftLogan:
Hey @joshhanley , thanks for catching that hey I need to filp the keys when  when query string array does not have an array of params as a value. I think all tests should run fine now. Thanks

calebporzio:
Great, thanks @CraftLogan and @joshhanley !

PR #4376, opened 2021-12-10
Suppress type warning

telkins:
I'm not sure that this is the best solution, but since `parse_str()` expects a string, it made sense to me to cast whatever **this particular call to** `parse_url()` returns to a string.

It seems that `parse_url()`, when specifying the `PHP_URL_QUERY` component, will always return a string or `null`.  So, this seems like a safe solution.

FYI...this is only generating a warning:
```bash
parse_str(): Passing null to parameter #1 ($string) of type string is deprecated in /path/to/vendor/livewire/livewire/src/Features/SupportBrowserHistory.php on line 130
```

So, it's not fixing anything that's broken, but just silencing some of the noise. 🤓 

Please let me know if anyone has any questions/concerns.

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportBrowserHistory.php
+++ b/src/Features/SupportBrowserHistory.php
@@ -127,7 +127,7 @@ class SupportBrowserHistory
     {
         if (empty($referer = request()->header('Referer'))) return [];
 
-        parse_str(parse_url($referer, PHP_URL_QUERY), $refererQueryString);
+        parse_str((string) parse_url($referer, PHP_URL_QUERY), $refererQueryString);
 
         return $refererQueryString;
     }


joshhanley:
@telkins thanks for the PR!

Any chance you could add some tests for this? Please follow our new [contribution guide](https://laravel-livewire.com/docs/2.x/contribution-guide) for more details.

Thanks!

calebporzio:
If all the other tests still pass, I'm good on this one - just an error silencer. Thanks!

joshhanley:
RFC which caused this deprecation notice for future reference - https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg

PR #4382, opened 2021-12-13
Fix checkboxes and integers

joshhanley:
This PR fixes an issue #3069 where an array of checkboxes values are pre-filled with integers, aren't being unselected correctly and causing a duplicate when re-selected with both the integer value and a string version of that value in the array.

image

Livewire already had fuzzy matching to take the properties values and check/uncheck the checkbox inputs as required, hence why an array of integers were checked properly on initial load.

But it had strict matching when doing the reverse and taking the values of the checkboxes, which were checked, and updating the properties values.

This PR changes the input -> property updating to also be fuzzy matched.

Hope this helps!


calebporzio:
Perfect, thanks Josh!

PR #4398, opened 2021-12-16
Fix Calling reset() on an object is deprecated

reefki:
The Laravel `head()` helper is an alias of PHP `reset()`, after upgrading to PHP 8.1, Ray keeps showing depreciation notice:

```
reset(): Calling reset() on an object is deprecated in .../vendor/laravel/framework/src/Illuminate/Collections/helpers.php on line 158
```

This fix ensures that `ComponentAttributeBag` is treated as an `array` to remove the deprecation notice.


calebporzio:
Perfect! Thanks!

PR #4406, opened 2021-12-17
Fix cursor pagination in bootstrap

nshiro:
Yes, it is needed. I first discussed it because I couldn't run test. #4387 

Yes, I did.

No.

No.

In v2.8.2, cursor pagination feature was added #3215 but it's not working when using bootstrap theme.

### Steps to reproduce
ShowUsers.php
```php
{
    use WithPagination;

    protected $paginationTheme = 'bootstrap';

    public function render()
    {
        $users = User::query()
            ->orderByDesc('id')
            ->cursorPaginate(10);

            'users' => $users,
        ]);
    }
}
```
show-users.blade.php
```
{{ $users->links() }}
```
I get an error like so.
```
BadMethodCallException
Method Illuminate\Database\Eloquent\Collection::getPageName does not exist. (View: /___MY_PATH__/vendor/livewire/livewire/src/views/pagination/simple-bootstrap.blade.php)
```

When compared with 'simple-tailwind.blade.php', I think this is the right solution.
I didn't run test because I got errors while dusk testing.
But of course I tested with my brower manually.

Thanks for contributing! 🙌
🙌

### DIFF
--- a/src/views/pagination/simple-bootstrap.blade.php
+++ b/src/views/pagination/simple-bootstrap.blade.php
@@ -10,7 +10,7 @@
                 @else
                     @if(method_exists($paginator,'getCursorName'))
                         <li class="page-item">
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

PR #4437, opened 2021-12-21
[PHP 8.1] Fix deprecation warning reset(): Calling reset() on an object is deprecated

geisi:

No

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed
Yes

Only this fix

No tests are needed for deprecation warnings

Thanks for contributing! 🙌

Hello,

on a PHP 8.1 installation our deprecation.log was spammed by following message:

```bash
reset(): Calling reset() on an object is deprecated in /home/swwv-iob/swwv-iob.test.ecoplan-crm.com/vendor/laravel/framework/src/Illuminate/Collections/helpers.php on line 158
```

Digging deeper into this problem I found the particular code in the *LivewireServiceProvider.php* when the wire *ComponentAttributeBag* macro is registered. I'm not quite sure why this head() helper was used in this case indeed I think it was not necessary because the head() helper is called on the attributes array afterwards. So the result should always be the same.

These deprecation warnings occur every time a livewire component with registered query string vars is rendered.

All unit tests are still passing so I think this fix does not break anything.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -243,7 +243,7 @@ class LivewireServiceProvider extends ServiceProvider
         // Early versions of Laravel 7.x don't have this method.
         if (method_exists(ComponentAttributeBag::class, 'macro')) {
             ComponentAttributeBag::macro('wire', function ($name) {
-                $entries = $this->whereStartsWith('wire:'.$name)->getAttributes();
+                $entries = $this->whereStartsWith('wire:'.$name)->attributes;
 
                 $directive = head(array_keys($entries));
                 $value = head(array_values($entries));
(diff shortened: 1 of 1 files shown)


joshhanley:
@geisi thanks for the PR! We actually just merged a fix for this yesterday in PR #4398.

Can you confirm if pull down master into your project that it fixes the deprecation issue?

Thanks!

geisi:
@joshhanley Yeah it seems that the existing pr already fixes the problem shame on me I did not recognise it. So im gonna close this one. Thank you for your fast reply.

joshhanley:
@geisi no worries! Thanks for your effort and for confirming 🙂

PR #4450, opened 2021-12-24
FIX keep child components even when a parent component is skipped rendering.

nshiro:
Yes, I did. #4447

Yes, I did.

No.

Yes, it does.

When you use 'skipRender' in a parent component, child components will get initialized next time you don't skip rendering. I described it in detail in #4447

Thanks for contributing! 🙌
Thank you & Merry Christmas! 🙌


calebporzio:
Thanks for this!

PR #4460, opened 2021-12-29
Undefined array key "value" for syncInput

danharrin:
I had a chat with Josh about this on Discord.

I've had this exception in my error logs across several Livewire sites:

image

I've never hit this exception myself before, neither in development nor production.

I checked all instances in the JavaScript codebase where `syncInput` is sent as an update, and both times `value` is sent as a payload item:

Screenshot 2021-12-29 at 11 33 29

If I check the request payload, the `value` key is not present:

I understand that this PR is purely hiding the symptom of a problem, but I don't know how to write a failing test for this internal behaviour, and I can't replicate the issue on my own machine.

## Alternative solution

My immediate thought would be to add `?? null` onto the `$data['value']`, but in this case the value of the property would be completely lost - instead of being retained. If you think this is a better solution, let me know :)

### DIFF
--- a/src/HydrationMiddleware/PerformDataBindingUpdates.php
+++ b/src/HydrationMiddleware/PerformDataBindingUpdates.php
@@ -14,6 +14,8 @@ class PerformDataBindingUpdates implements HydrationMiddleware
                 if ($update['type'] !== 'syncInput') continue;
 
                 $data = $update['payload'];
+                
+                if (! array_key_exists('value', $data)) continue;
 
                 $unHydratedInstance->syncInput($data['name'], $data['value']);
             }


calebporzio:
Ok, I'm good with this, not sure what's happening there, but this addition isn't dangerous. Ideally there's a test included with this, so now that I merged, there's no guarantee this won't crop up again later, but it's small and we're probably fine. Thanks!

adriaanzon:
Any chance you are using `$dispatch('input', null)` from an x-on listener, to set the value of a wire:model to `null`? In my app, I noticed that some of my "Clear" buttons stopped working after updating Livewire.

Since commit https://github.com/livewire/livewire/commit/3de014b150ca55a16ea6ad01d2e44ad2a595db2d (or Livewire version 2.8.0), falsy values, including `null`, get overridden by the value of `e.target.value`, even when using `$dispatch('input', ...)` where `e.target` does not make sense. 

Here is an example, with an onclick handler on a `span` element:

```blade
<span role="button" x-on:click="$dispatch('input', null)">Clear</span>
```

In this case, the value of `e.target.value` is `undefined`, because `span` elements do not have a `value` attribute. Now, the dispatched `null` value gets overridden with `undefined`, causing the "value" attribute to be removed from the syncInput payload when it is converted to JSON.

A similar problem happens when using a `button` element instead of a `span` element, but the value of `e.target.value` will be an empty string instead of `undefined`, because buttons *can* have a value attribute.

PR #4485, opened 2022-01-07
handle SPLFileInfo deprecation notices

fabio-ivona:
This PR will add `#[\ReturnTypeWillChange]` to `TemporaryUploadedFile.php` in order to suppress deprecation notices thrown because of SPLFileInfo new deprecations

The fix would be to make the methods compatible with SPLFileInfo, but it would mean a breaking changes, so I've not done it yet

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -23,6 +23,7 @@ class TemporaryUploadedFile extends UploadedFile
         parent::__construct(stream_get_meta_data($tmpFile)['uri'], $this->path);
     }
 
+    #[\ReturnTypeWillChange]
     public function getPath()
     {
         return $this->storage->path(FileUploadConfiguration::directory());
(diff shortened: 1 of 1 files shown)


joshhanley:
@fabio-ivona thanks for the PR!

Will having attributes in the code cause any issues for any older versions of PHP? I assume not as all the tests have passed, but thought I'd check 🙂

If there is not issues, then yeah I think this is the better solution as it's not a breaking change. Plus I'm not sure how Caleb would feel about having return types 😬

fabio-ivona:
Afaik attributes don't break old code, they should be interpreted as comments

joshhanley:
Ok sweet!

samirmhsnv:
why this PR not getting merged?

calebporzio:
Ok, this has been fixed in #4544 - closing. Thanks for the contribution!

PR #4511, opened 2022-01-12
Missing dispatched events with assertDispatchedBrowserEvent() returns in undefined array key

ahinkle:

Probably so - helps with assertion testing.

Yes.

No, one singular change.

Yes.

This PR resolves a small bug with `assertDispatchedBrowserEvent()`. If you return a component with no dispatched events, an Error Exception from the framework will occur (vs. a failed PHPUnit assertion)
```
ErrorException: Undefined array key "dispatches"

src/Testing/Concerns/MakesAssertions.php:228
```

If other browser events are fired, this will be fine, but this PR allows the failed assertion to return.


calebporzio:
Good with me. Thanks!

PR #4517, opened 2022-01-13
Allow validation data to be set from the validate/validateOnly methods

mikemand:

Yes, #4459.

Yes.

No.

Yes, but a bunch of unrelated tests fail locally. Hoping they don't fail when Github Actions run, since I don't know why they fail.

This PR adds an additional, optional, parameter to pass data directly to the validate functions.

For normal validation, I can just use [Custom Validators](https://laravel-livewire.com/docs/2.x/input-validation#custom-validators). As far as I know, however, if I want to use real-time validation I would have to pluck the attribute's validation rules from the array and pass that to the `Validator::make` method (while also providing any messages and attributes - if applicable). While this is not necessarily the wrong way to do things, I feel that it duplicates code that the `validate()` and `validateOnly()` methods provide.

Thanks for contributing! 🙌

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -163,12 +163,12 @@ trait ValidatesInput
         return $this;
     }
 
-    public function validate($rules = null, $messages = [], $attributes = [])
+    public function validate($rules = null, $messages = [], $attributes = [], $data = null)
     {
         [$rules, $messages, $attributes] = $this->providedOrGlobalRulesMessagesAndAttributes($rules, $messages, $attributes);
 
         $data = $this->prepareForValidation(
-            $this->getDataForValidation($rules)
+            $data ?? $this->getDataForValidation($rules)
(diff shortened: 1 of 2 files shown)


mikemand:
The test that failed has to do with Dusk. As far as I know, I didn't change anything that would mess with Dusk? Any help would be appreciated.

joshhanley:
@mikemand yeah some of the dusk tests can be flaky at times. I'll run the tests again

joshhanley:
@mikemand yeah don't worry about the tests, most suites are passing, and I've checked the tests that fail and it's a common one that does fail.

Yeah I don't think he has had a chance to go through PR's recently, but I'm sure he will give feedback once he does!

mikemand:
No worries, thanks Caleb. If you don't mind, I'll publish my solution on #4459.

joshhanley:
@mikemand yep go for it! Or create a new discussion and put it in the show and tell category 🙂

PR #4544, opened 2022-01-18
Laravel 9 support

joshhanley:
This PR adds Laravel 9 support.

At the time of the PR, there are some upstream issues with sessions, so L9 tests will probably fail until that is fixed.

Once that's all working, I will work through any tests if they fail.


danharrin:
Just to note, I think this PR replaces some of #4485?

joshhanley:
@danharrin yeah it does 🙂 I haven't decided the best strategy yet, as waiting to be able to run the rest suites to see what happens

joshhanley:
Ok after merging in Nuno's changes, tests started to run, so started working through some of the issues. Most of them were due to Flysystem changes.

I have fixed most of them, but there is one change in Flysystem that I can't get around, some of the tests fail that have a faked file which is a pdf, which you will see in the failing tests above.

One of the tests that fails is this one, due to the `->temporaryUrl()` call.

https://github.com/livewire/livewire/blob/3017bab78bcaf4ed34bbb26802ff54d7d4da7ac7/tests/Unit/FileUploadsTest.php#L458

Flysystem V1 tries to get the mimetype from file info and failing that would try to guess it based on the file extension. See this function here 

https://github.com/thephpleague/flysystem/blob/094defdb4a7001845300334e7c1ee2335925ef99/src/Adapter/Local.php#L319-L330

Where as V2+ only tries to get the mimetype from the file info, and if that doesn't work ends up returning a generic file type. See here

https://github.com/thephpleague/flysystem/blob/912ba3e5595280999b9fa99a3c20c5ecb7de6870/src/Local/LocalFilesystemAdapter.php#L366-L377

Due to the generic file type, it then doesn't return an extension when Livewire requests it in the `->isPreviewable()` method in `TemporaryUploadedFile`.

https://github.com/livewire/livewire/blob/3017bab78bcaf4ed34bbb26802ff54d7d4da7ac7/src/TemporaryUploadedFile.php#L93

So not sure how to get around that yet. Will do some more testing to see if we can work around it. I think it's only an issue for the tests when using a fake file. But will need to test on a live app to confirm.

calebporzio:
Thanks Josh, without really putting my head deep in this, the two options as I see them are:

A) do something deeper with that actual fake file content to force the proper mime type guessing for V2 Flysystem
B) override the `guessExtension()` method in Livewire's TemporaryUploadedFile.php class to use the legacy strategy of checking the file name if the file contents check fails 

image

B is probably the way to go here.

joshhanley:
Ok I have patched Livewire's `TemporaryUploadedFile::getMimeType()` method to implement the mimetype detector from file extension that was present in Flysystem V1.

Have fix up any other small issues and now all tests are passing. 

So I think this is now all good to go!

calebporzio:
Thanks @joshhanley !

PR #4559, opened 2022-01-21
[2.x] Fixes the way Livewire sets the session on Livewire internal requests

nunomaduro:
This pull requests fixes the way Livewire sets the current Laravel 9.x session on internal requests. Added a test, to ensure we cover this in the future.


calebporzio:
Thanks @nunomaduro!

afrizal423:
Thanks @nunomaduro 🚀

PR #4565, opened 2022-01-23
Overhaul public property hydration

jameslkingsley:
This PR changes the way public properties are hydrated/dehydrated to support custom types and offer more flexibility when using typed properties in PHP. I've taken inspiration from #4154, which first introduced the idea of a property type manager. My implementation handles this in a similar way, although it offers some more control (eg. property attributes), as well as refactors the existing type handlers from `HydratePublicProperties`.

### Firstly, why?

Using typed properties with Livewire has always been a bit of a headache, and usually involved setting up update hooks to convert a JavaScript friendly value into a PHP value. I'm sure we've all done something like this:
```php
public $dateString;
public ?Carbon $date = null;

public function updatedDateString($value)
{
    $this->date = Carbon::parse($value);
}
```

There is _some_ support for dates already, providing your value is initiated as a date instance. However, I think in most scenarios these sorts of properties will be `null` until some user input has been made.

Ideally what should be possible is this:
```php
public ?Carbon $date = null;
```

### How?

The `LivewireManager` now holds an array of supported types. These types map to their own handlers which are responsible for hydration and dehydration. During the hydration lifecycle, Livewire will determine the type for a property. It does this by using `Reflection` _and_ the type of the value (if Reflection wasn't useful). Once it has the type and has resolved the handler, it simply calls on the handler to hydrate.

### Attributes

One thing that I thought would be really useful is the ability to use PHP 8 attributes to attach metadata to a property. This is particularly useful on date properties where you want to specify the format to use on dehydration.
```php
#[DateFormat('Y-m-d')]
public ?Carbon $date = null;
```

Doing this allows you to `wire:model` directly to a native date input, and you'll still receive a `Carbon` instance on the back-end. Specifying the date format ensures the front-end can understand it (eg. `<input type=date>` or some other date picker).

----------------------------------

Another attribute that's still a work in progress (and maybe a stretch too far), is the `ModelKey`. This would allow you to build a simple select input like this:
```html
<select wire:model="cheese">
    <option value="1">Edam</option>
    <option value="2">Cheddar</option>
    <option value="3">Blue</option>
</select>
```

And model it directly to the property like this:
```php
#[ModelKey('id')]
public ?Cheese $cheese = null;
```

As I said this attribute is a work in progress, and I'm not sure if this has some security concerns. Although it would make a common pattern extremely concise and easy to understand.

### DIFF
--- a/src/Attributes/DateFormat.php
+++ b/src/Attributes/DateFormat.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Attribute;
+
+#[Attribute]
+class DateFormat
+{
+    public $format;
+
+    public function __construct($format)
+    {
+        $this->format = $format;
+    }
+}
--- a/src/Attributes/ModelKey.php
+++ b/src/Attributes/ModelKey.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Attribute;
+
+#[Attribute]
+class ModelKey
+{
+    public $key;
+    public $label;
+    public $strict;
+
+    public function __construct($key, $label = null, $strict = false)
+    {
+        $this->key = $key;
+        $this->label = $label;
+        $this->strict = $strict;
+    }
+}
--- a/src/Attributes/ModelKey.php
+++ b/src/Attributes/ModelKey.php
@@ -8,13 +8,9 @@ use Attribute;
 class ModelKey
 {
     public $key;
-    public $label;
-    public $strict;
 
-    public function __construct($key, $label = null, $strict = false)
+    public function __construct($key)
     {
         $this->key = $key;
-        $this->label = $label;
-        $this->strict = $strict;
     }
 }
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -12,6 +12,7 @@ use Livewire\Exceptions\MissingFileUploadsTraitException;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Livewire\HydrationMiddleware\HashDataPropertiesForDirtyDetection;
 use Livewire\Exceptions\CannotBindToModelDataWithoutValidationRuleException;
+use Livewire\LivewirePropertyManager;
 use function Livewire\str;
 
 trait HandlesActions
@@ -19,6 +20,8 @@ trait HandlesActions
     public function syncInput($name, $value, $rehash = true)
     {
         $propertyName = $this->beforeFirstDot($name);
+
+        $value = Livewire::hydrate($this, $propertyName, $value);
 
         throw_if(
             ($this->{$propertyName} instanceof Model || $this->{$propertyName} instanceof EloquentCollection) && $this->missingRuleFor($name),
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -12,16 +12,17 @@ use Livewire\Exceptions\MissingFileUploadsTraitException;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Livewire\HydrationMiddleware\HashDataPropertiesForDirtyDetection;
 use Livewire\Exceptions\CannotBindToModelDataWithoutValidationRuleException;
-use Livewire\LivewirePropertyManager;
 use function Livewire\str;
 
 trait HandlesActions
 {
-    public function syncInput($name, $value, $rehash = true)
+    public function syncInput($name, $value, $rehash = true, $request = null)
     {
         $propertyName = $this->beforeFirstDot($name);
 
-        $value = Livewire::hydrate($this, $propertyName, $value);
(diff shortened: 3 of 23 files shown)


jameslkingsley:
There's a couple failing tests atm that I could do with a second pair of eyes on. Not sure why exactly they're failing, maybe it'll make sense tomorrow 🤷

joshhanley:
@jameslkingsley thanks for the PR!

Not to knock the effort you put into this, but can you explain why you submitted a PR that's similar to #4154? And if you can give us the pros/cons of this over the other PR?

Only reason I ask it because that other PR is nearly done, it's waiting on me to go through it (which I haven't had the time yet), as I've been working with Jonas to develop it.

One thing is though, Caleb had request in the original discussion that the feature be implemented first so it can be tested out with custom properties in peoples apps before we refactor the existing code to make use of it. So for now, can you revert the changes to the existing properties and save the code for a future PR?

Also Livewire supports PHP 7.2+, so I'm not sure if that impacts the usage of attributes as they weren't introduced until PHP 8. Can you explain how that will work for older versions?

All that being said, if this is a better implementation, then I'm not against it, so I'm looking forward to your thoughts on the above.

Thanks!

jameslkingsley:
Thanks for taking a look Josh,

Both PRs are quite similar, however after looking into #4154 I noticed that it doesn't actually solve one of the more common use cases for custom properties. I may not have been clear enough in the comments I left on it, my apologies, I'll update there as well. But essentially, if a property starts as `NULL` then it won't be considered a "custom property" and then on hydration it won't go through the intended property handler.

This is the scenario that I find most common in almost all Livewire components I write, where you have some properties that are `null`, and you update them with user input. Eg:
```php
public ?Carbon $cancellationDate = null;
```
```blade
<x-input.date wire:model="cancellationDate" />
```

This scenario doesn't work with #4154 as it's only checking the value on dehydration, whereas it should also factor in the property's type using Reflection.

That's the difference between the two PRs, where in my one it will do a bunch of checks to determine the property type (using the defined type, or value type etc.) and then serialize accordingly. I've taken steps to ensure this will work on properties without any type declared as it will simply use the value type (as it did before).

Another benefit of this approach is that in the custom property handlers they will receive enough information to resolve the property's Reflection instance. This gives the developer the option to use PHP 8 Attributes to decorate their properties and pass along metadata, such as a date format, as demo'd in this PR.

The attributes added in this PR are purely optional for the developer, so if they're on PHP 7.2 then they can simply ignore them (it'll skip the attribute check if not on version 8). I think it'd be nice if Livewire supported the most common workflows out of the box - like specifying a date format on dehydration - but I understand if this is out of scope. All of the built-in property type handlers can of course be overriden by a developer, simply by registering their own.

If you're still happy to keep this PR up then I can find some time to revert the changes made to the original property hydration. Hope this helps 👍

joshhanley:
@jameslkingsley yeah fair enough. That's one of the reasons why I need to review it because I want to make sure all possibilities are covered (such as starting from NULL) and I want to test it out in a small demo app to make sure it works as expected (just haven't had to time spare to do it justice).

Yeah it originally used reflection on dehydration (I think?) but I said we can't rely only on that incase they aren't using types, so need to also use the values type if no reflection found. It's possible he changed too much. But yeah it needs to support both.

Ok good to know about the attributes. If I'm being completely honest, as a personal preference I don't like them (in general) as such haven't ever used them and hence don't understand how they work nor the benefit of them. But if they add value, I've got no issue with the functionality allowing devs to use them if they wish 🙂

Yeah I think I'd like to see this PR without all the original property changes, if you don't mind updating it that is, that will make it easier to review it against the other PR, so maybe we can get the best of both of them!

jameslkingsley:
Cool, I'll update it and ping you when it's ready 👍

joshhanley:
Ok thanks

jameslkingsley:
@joshhanley I'm looking at how I can revert the changes to the existing hydration code, but not sure whether it's going to be feasible given how I've implemented my version. Would something like this suffice? These middleware would be resolved from the container, so we can swap the old/new version if the user opts in. Not sure whether this would impede your review process?
```php
$useNewHydrationMiddleware = config('livewire.use_new_property_hydration', false);

$this->app->bind(
    PublicPropertyHydrator::class,
    $useNewHydrationMiddleware
        ? HydratePublicProperties::class
        : LegacyHydratePublicProperties::class
);
```

joshhanley:
@jameslkingsley yeah I can see why it would take a fair bit of work to revert the changes.

I'm not a fan of the config flag and I know Caleb is hesitant in adding new ones.

Also if you look at Caleb's experiment which I linked to on the other PR, he makes use of a separate property manager class, which Jonas' PR has but yours doesn't.

Considering both of the above things and the changes required to get this PR there, I think we should close this PR and continue working on Jonas' PR.

What do you think?

Thank you for all your hard work on this though, and for going through and reviewing the other PR! It helps having another set of eyes that understand the depths of these changes 🙂

jameslkingsley:
@joshhanley Yeah I'm happy for this PR to be closed in favour of #4154 👍

joshhanley:
@jameslkingsley ok great, thanks!

PR #4568, opened 2022-01-25
fix array on query string

Lupennat:
Yes, I created a discussion https://github.com/livewire/livewire/discussions/4473
Yes
No
Yes
This PR will fix wrong behaviour on browser history when querystring contains square bracket parameters

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportBrowserHistory.php
+++ b/src/Features/SupportBrowserHistory.php
@@ -94,6 +94,10 @@ class SupportBrowserHistory
             $path = $this->buildPathFromReferer($referer, $queryParams);
         }
 
+        // Referer query string sent by the browser can have different encoding when parameters are bracket formatted
+        // the query string is rebuilt to avoid pushing undesired entries on the browser history
+        $referer = $this->buildPathFromReferer($referer, collect($this->getQueryParamsFromRefererHeader()));
+
         if ($referer !== $path) {
             $response->effects['path'] = $path;
         }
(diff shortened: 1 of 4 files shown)


Lupennat:
I believe that the failed tests are not due to the changes related to this pull request ...

joshhanley:
@Lupennat yeah some of the dusk tests can be flaky at times. I can see a few of the test suites have passed, so should be ok.

calebporzio:
Thanks for submitting this.

I agree that seeing plain brackets in the query string is preferred.

2 issues with this PR however:
- I don't see a test that actually asserts the usage of plain brackets
- The Dusk test that was added is very big compared to the feature added. Is it possible you copy and pasted it from somewhere and there are unnecessary assertions and scenarios in there?


Thanks for contributing!

Lupennat:
Hi @calebporzio,
thanks for reply.

For the first point, I didn't change the encoding of the path, but I made the encoding of the referrer consistent with the path generated by livewire before checking the two strings.
This way the path string generated by livewire remains unchanged, but the path key is not added to the response json when it is not needed because of the different encoding.
This allows both the javascript and the PHP_QUERY_RFC1738 encoding to remain unchanged, which is why there is no test to verify the use of plain brackets.

As for the tests, I created them myself (no copy and paste), it's actually quite a long test, but I think it's necessary since I didn't find any browser history checks in the other tests.
It can probably be split into multiple tests if needed, in detail:
- the first block only checks that the browser history is identified correctly
- the second block checks the first fix about the extra '?' in the path when there is an array of nested nullables in the default queryString
- the third and fourth blocks check that input changes generate entries in the history and that navigation in the browser history works properly.
- the fifth block verifies the second fix (referrer vs path) and checks that the wrong entry points are not generated in the history.

joshhanley:
@Lupennat Caleb is currently off on paternity leave, so might be a little while before he is back and reviewing PRs 🙂

joshhanley:
@Lupennat thanks for being willing to contribute this, however we're going to close this for now as the query string handling is being rewritten for V3.

Lupennat:
@joshhanley sure, no problem

PR #4570, opened 2022-01-25
Fix localization url in HttpConnectionHandler.php

dennisvandalen:
This pull request fixes wrong order of the arguments supplied the `Str::replaceFirst($search, $replace, $subject)` helper function in the `HttpConnectionHandler`. This introduced a bug where when you were not logged in and using locale in the url you were redirected to the login page in the livewire message. This bug is introduced in Livewire v2.3.14

https://laravel.com/docs/8.x/helpers#method-str-replace-first 

I started having issues after adding https://github.com/mcamara/laravel-localization to my project.

### DIFF
--- a/src/Controllers/HttpConnectionHandler.php
+++ b/src/Controllers/HttpConnectionHandler.php
@@ -33,7 +33,7 @@ class HttpConnectionHandler extends ConnectionHandler
             );
         } catch (NotFoundHttpException $e) {
             $request = $this->makeRequestFromUrlAndMethod(
-                Str::replaceFirst(Livewire::originalUrl(), request('fingerprint')['locale'].'/', ''),
+                Str::replaceFirst(request('fingerprint')['locale'].'/', '', Livewire::originalUrl()),
                 Livewire::originalMethod()
             );
         }


joshhanley:
@dennisvandalen thanks for the PR!

Do you think there is anyway we can add a test for this, so it doesn't happen again? It may be to complex to test, but would be nice if we could!

dennisvandalen:
@joshhanley I tried to write a test but I'm not familiar enough with the internals to write a decent test. Sorry

joshhanley:
@dennisvandalen no worries! Any chance you can create a small repo for me, that demonstrates the issue, so I can then use it to test out your fix? Hopefully then I should be able to come up with a test for it then.

Thanks!

dennisvandalen:
@joshhanley anything else I can do here? A quick glance at the [documentation](https://laravel.com/api/9.x/Illuminate/Support/Str.html#method_replaceFirst) of `Str::replaceFirst` should make it clear that the current implementation is a typo and my 1 change should put the parameters in the right order. 

`static string replaceFirst(string $search, string $replace, string $subject)`

Thanks

joshhanley:
@dennisvandalen thanks for providing that repo, I finally got a chance to have a look and it clearly demonstrated the issue. We didn't have the time to work out a test for it, so just merged the fix anyway. Thanks for the PR! 🙂

lam0819:
This fix seems crash my livewire when working with https://github.com/mcamara/laravel-localization together. I currently downgrade my version to 2.10.7

dennisvandalen:
> This fix seems crash my livewire when working with https://github.com/mcamara/laravel-localization together. I currently downgrade my version to 2.10.7
> 
> Further, try in our own code base , this will fix. I am not sure will this affect your code base @dennisvandalen
> 
> change `Str::replaceFirst(request('fingerprint')['locale'].'/', '', Livewire::originalUrl()),` to `Str::replaceFirst('/'.request('fingerprint')['locale'], '', Livewire::originalUrl()),`

Should be fine for me. I just kept most of the original code intact, only flipped the arguments

Shtier:
The problem seems to be that the route is not defined here:
```
    {
    ...
    $route = app('router')->getRoutes()->match($request); // 404
    ...
    }
```
    
if use `$route = app('router')->match($method, $url);` everything works for me

Simian-m:
@Shtier Yes I've been trough the same trail and it leads to this particular fix, the current fix here works only if the subdirectory is a localization string, if not ( in our cases ) this breaks the entire livewire application when using subdirectories.

lalov:
@Shtier Your solution solves the problem for me too when I have the project in a subdirectory

jochensengier:
We're experiencing a similar issue with the latest release of Livewire when using the https://github.com/ARCANEDEV/Localization package.

Livewire components on the 'root/home' page no longer work (example: 'https://example.com/nl' or 'https://example.com/fr')
However, when the components are loaded on any other page than the 'root/home' page, it works (example: https://example.com/nl/test).

Both solutions from @lam0819 and @Shtier mentioned in this thread work for us.

Perhaps @calebporzio can have a look for a permanent fix/solution?

Thanks!

stack74:
This change also causes a problem / 404 error when calling an action in a nested component

joshhanley:
There was a bug released in v2.10.8. I have submitted a PR #5490 with a fix. Hope this helps!

@lam0819 thanks for that! Your bug is slightly different to what others are experiencing (they're having issues in a subdirectory), but I have added your fix to the PR too, as it's related 🙂

@Shtier thanks for the input! You're proposed fix isn't ideal as calling `->match` directly on the router actually adds a new route, not "match" an existing one 😅 see here for details https://laravel.com/docs/9.x/routing#available-router-methods

The fix I ended up with was thanks to @Eighke here https://github.com/livewire/livewire/discussions/5464#discussioncomment-4603754

Issue #1356, opened 2020-08-05
CorruptComponentPayloadException when properties contains specific UTF-8 characters


huyby:
### Description
When component properties contains strings with specific UTF-8 characters that are "normalized" by the browser's JS engine, the checksum fails because the requests's payload will differ from the data set through the `wire:initial-data` attribute. 

We bumped into this issue when data contained an alternative character combination for the character (combination) `ë`. Apparently there's two ways of creating the character:

1. A single UTF-8 character `ë` with decimal code 235 and hex format `\u00eb` http://www.ltg.ed.ac.uk/~richard/utf-8.cgi?input=235&mode=decimal
2. A combination of two characters `e`and `¨` with latter decimal code 776 and hex format `\u0308` http://www.ltg.ed.ac.uk/~richard/utf-8.cgi?input=776&mode=decimal

The issue is with the second presentation, the combination of two characters. The JS engine of WebKit based browsers, JavascriptCore, will automatically "normalize" this combination into a single `\u00eb` character. Probably this happens when the `wire:initial-data` attribute is parsed, and can be simulated through Safari's Web Inspector console:

```
const test = 'e\u0308'
test.normalize('NFD').charCodeAt(0) // 101
test.normalize('NFD').charCodeAt(1) // 776
test.normalize('NFC').charCodeAt(0) // 235
```

When the parsed data is sent back to the server with the normalized character, the checksum will not match resulting in an `CorruptComponentPayloadException`.

#### Possible solution(s)?

1. We need to normalize our user input, and normalize all UTF-8 strings in our database. Probably most application are not doing this normalize step?
2. When Livewire package is creating the checksum, the normalization is done on all strings using Normalization Form C (NFC) which is default when using [PHP's Normalizer class](https://www.php.net/manual/en/class.normalizer.php)
3. ~~The Livewire JS needs to ensure normalization is consistent in all browsers when parsing the `wire:initial-data` attribute~~ Not an option, as checksum is created on back-end side. This would only emulate JavascriptCore behaviour in V8 (and maybe other) JS engine(s).

References:
https://withblue.ink/2019/03/11/why-you-need-to-normalize-unicode-strings.html
https://www.php.net/manual/en/class.normalizer.php
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize

### Steps to reproduce

[...]

PR #4613, opened 2022-02-04
Attempt at fixing the safari special character encoding checksum issue

jonneroelofs:
see https://github.com/livewire/livewire/discussions/3018

I did, this time :)

Out of curiosity I went down the rabbit hole of why Safair/Ios causes a corrupt data error. I found out that safari encodes (if that is the right word here) special characters differently from chrome. 

I saw a suggestion somewhere in and old issue https://github.com/livewire/livewire/issues/1356 that this could be solved by normalizing the data. So I gave it a go. 

It solved the problem for me for safari on a mac.

Don't know if this is the correct solution, but hopefully a step in the right direction.

Thanks for contributing! 🙌
Thank you for Livewire!

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -91,9 +91,11 @@ class HydratePublicProperties implements HydrationMiddleware
         array_walk($publicData, function ($value, $key) use ($instance, $response) {
             if (
                 // The value is a supported type, set it in the data, if not, throw an exception for the user.
-                is_bool($value) || is_null($value) || is_array($value) || is_numeric($value) || is_string($value)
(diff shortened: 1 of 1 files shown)


joshhanley:
Thanks @jonneroelofs for digging into this and for the PR!

Any chance you could create a test for this? When running tests, in the `tests/Browser/TestCase.php` there is a flag `$useSafari` that can be enabled to check it with.

Thanks!

calebporzio:
Thanks for this @jonneroelofs, I do think this is a valuable contribution, however we will need a test to make sure we don't re-introduce this bug in future versions. I'm closing because of this, please re-open if you add a test (btw, you can probably just do a PHPUnit test (not dusk) to simulate the mutation safari does)

PR #4618, opened 2022-02-05
Allow frontend wire: actions to span multiple lines

claytonrcarter:

Oh, uh, no, I didn't. 😞 

Yes

No.

Yes

Currently, `wire:` actions are only allowed to be on a single line. This is fine, but *sometimes* it would be nice to break up some of the funkier actions across several lines for readability. This has bitten me several times, and each time I'm left scratching my head b/c the error message is not very intuitive. This 1 character change makes this go away by allowing the JS to parse methods & params on multiple lines.

**Example:**
```html
<button type="button" wire:click="
        setOutputTo(
            'fizz',
            'fuzz'
        )" 
        dusk="fizzfuzz"
>Baw</button>
```

Currently, this will generate the error: `Unable to call component method. Public method [setOutputTo( 'fizz', 'fuzz' )] not found on component: [crop-plan.planting-editor] `. This message is hard to understand, though, because it *sure* looks right as it's printed in the message. I find myself double checking my spelling for the method name, comfirming that the params are correct, etc. On a whim, I might squish it back onto 1 line and *voila* it starts working again. This example is simplistic, of course, but consider if the those parameters were blade directives or `{{ $myEloquentlyNamedVariable }}`, etc.

By changing the JS regex that parses methods and params to use the `s` flag, this just goes away and allows us to format our actions a bit more naturally.

**Alternate Approach**  
Instead of allowing actions to span multiple lines, we could instead embrace the single line restriction and clarify the error message in this case. Perhaps something like "Unable to call component method. Actions must be written on a single line."?

**Tests**  
I had repeated issues running the dusk tests via Safari. (I don't use Chrome or have it installed.) Each failure was *different* though, and none of them seemed related to this change. CI will be the final judge, I suppose, though.


joshhanley:
@claytonrcarter thanks for the PR!

I think having multiline support is nicer than just adding an error message.

I have merged master in to resolve the conflicts with the built assets so the tests can run 🙂

Yeah I recommend using chrome when running the dusk tests (only because I know they all work 😬)

claytonrcarter:
Thank you @joshhanley!

Hmm, I fixed the first failure\*\*, but now I'm seeing [a failure that I can't repro](https://github.com/livewire/livewire/runs/5178557245?check_suite_focus=true). I haven't run the whole test Browser test suite at home (will do tomorrow), but that file is passing for me at home. (In Chrome no less! I thought I had uninstalled it, but I guess ... 🤔)

\*\* - Is there something magic about `wire:click="$refresh"` for the assertion on `tests/Browser/Actions/Test.php:53`? If it's just confirming that the last value is still displayed ... doesn't that mean that that action (ie the `click`) could just be doing nothing and that assertion will still succeed? I'd be happy to change it to something else, I just don't want to step on any toes and that assertion is talking about `wire:click.self` but I don't see that in the markup ... ?

joshhanley:
@claytonrcarter thanks! 

Ah yeah we have some flaky tests which fail in CI sometimes. If you have a look you can see a few test suites have actually passed, so that's generally what we look for, so all should be good!

calebporzio:
Thanks for this!

claytonrcarter:
Thanks. Happy to help!

Issue #2489, opened 2021-02-10
LivewireManager::$isLivewireRequestTestingOverride causing subsequent tests to fail


fuelingtheweb:
### Description
With Livewire v2.3.9 and above, running my entire test suite causes tests that rely on the ConvertEmptyStringsToNull middleware to fail when they run after a Livewire test that sets data.

```
        Livewire::actingAs($admin)
            ->test(FilterUsers::class)
            ->set('sort', 'first_name')
            ->assertViewHas('users');
```

All tests are green when `->set()` is omitted, but tests that rely on ConvertEmptyStringsToNull fail when `->set()` is included.

When I remove `LivewireManager::$isLivewireRequestTestingOverride = true;` from `TestableLivewire@pretendWereSendingAComponentUpdateRequest()`, all tests pass. I'm not sure if anything else is relying on that code. Can that safely be removed?

**Context**
- Livewire version: >=2.3.9
- Laravel version: 8.27.0

austenc:
Can you please provide more context so we can reproduce this without guessing? The test that fails, the livewire component, etc...

An example repository would be super helpful here. Thanks!

fuelingtheweb:
@austenc Here's a sample Laravel application that replicates the issue: https://github.com/benholmen/livewire-test-issue

If you run the entire test suite, `tests/Feature/OtherTest.php@testShouldPassButFailsWhenAlsoTestingLivewire` fails because `Livewire@set` in `tests/Feature/LivewireTest.php@testSettingPropertyBreaksOtherTests` is disabling middleware. When run in isolation, all tests pass.

If Livewire testing needs to disable middleware, could it just reset the middleware when the Livewire test is finished?

aleahy:
I've been finding the same thing. As @fuelingtheweb said, it only happens if the livewire version installed is 2.3.9 or greater.

The livewire component is not complicated - it just has a single property.
I've also created a separate, unrelated Post model with a controller to store new posts.

Here is an example of a livewire test that causes the fail:
```
public function test_it_can_update_value()
{
    Livewire::test(Something::class)
        ->set('textValue', 'Some value')
        ->assertViewHas('Some value');
}
```

Here is an example of a test that fails:
```
public function test_a_post_can_be_created_with_no_url()
{
    $this->post(route('posts.store'), [
            'title' => 'Some title',
            'url' => '',
            'message' => 'Some message',
        ])->assertSuccessful();

        $this->assertDatabaseHas('posts', [
[...]

PR #4619, opened 2022-02-05
Ensure Livewire request testing override is false by default

telkins:
It seems that during normal testing, this gets set to `true` and stays that way for subsequent tests...even if they're not Livewire tests.

I shared a little more information here (https://discord.com/channels/698229985755791471/698231501501890600/939586804200382484) and asked @calebporzio about it directly.

I haven't proven everything, but the various tests I've run seem to confirm my suspicions and forcing this setting to `false` appears to resolve the issues I've experienced.  I'm still not 100% sure of everything, however, nor am I sure that this is the best place to put this code...assuming it's the right solution in the first place.  🤓 

I'm happy to answer any questions or to share more information.  Please don't hesitate to ask. 🚀

Thanks for contributing! 🙌

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -59,6 +59,8 @@ class LivewireServiceProvider extends ServiceProvider
 
     public function boot()
     {
+        LivewireManager::$isLivewireRequestTestingOverride = false;
+
         $this->registerViews();
         $this->registerRoutes();
         $this->registerCommands();


telkins:
I should note...I know that the variable is set to `false` by default in `LivewireManager`:
```php
    public static $isLivewireRequestTestingOverride = false;
```

My **guess** is that this is how it starts during a normal set of tests...but because it's static, once it is changed it remains that way for subsequent tests until it's changed again.  As already noted, I'm not 100% sure of this, but when I force it to be `false` as I've done in this PR, the `TrimStrings` middleware is no longer skipped for my non-Livewire tests (that I'm looking at).

telkins:
FYI....here's a semi-related PR for Laravel framework: https://github.com/laravel/framework/pull/40829

It also has to do with static variables "persisting" from test to test.

And, yes, I understand one can run tests in separate processes.  I'm not sure that this would help.  (I haven't tried it for this situation.)  I've heard, however, that this slows tests down considerably, which I vaguely recall experiencing when trying out years ago.  In any case, this may not be 100% black and white.  😜

joshhanley:
@telkins thanks for this!

I've tried a couple of times to try to find a way to do this cleanly when a test is finished, I like this solution though.

eric-hansen:
> And, yes, I understand one can run tests in separate processes. I'm not sure that this would help. (I haven't tried it for this situation.) I've heard, however, that this slows tests down considerably, which I vaguely recall experiencing when trying out years ago. In any case, this may not be 100% black and white. 😜

From experience doing this causes undesirable side-effects that can make failing tests harder to debug.  PHPUnit also doesn't make this default for a reason (one of which is this I believe).  So it's best to try not running tests separately like this.

Part of the problem though with statics is that PHPUnit keeps running the class in the same runner instance, which is why the variables don't get reset as desired.  If PHPUnit would create a new runner for each test (class) that would solve this issue I believe.

I've ran into a semi-related issue with routes and Livewire that I'm hoping to submit a PR for soon.

JayBizzle:
I presume this would fix #2489?

Further discussion here - https://github.com/livewire/livewire/discussions/3120

calebporzio:
Thanks!

PR #4650, opened 2022-02-11
Fix S3 uploads with Flysystem v2+

joshhanley:
This PR fixes issues with uploading files to S3 when using Flysystem v2+. See discussion #4631.

Livewire requires access to the underlying client and bucket inside the Flysystem adapter. But as of v2+ the `getClient()` and `getBucket()` methods have been removed, so we've had to use invade to reflect those properties.

I've tested this code with an upload to S3 to both the Livewire temp folder and saving to a folder. I've also tested to ensure the lifecycle rule gets set up correctly when running the console command.

This has been tested manually on both Laravel 8/ Flysystem 1 and Laravel 9/ Flysystem 2+ to ensure everything is working as expected.

Hope this helps!

### DIFF
--- a/src/Commands/S3CleanupCommand.php
+++ b/src/Commands/S3CleanupCommand.php
@@ -3,7 +3,9 @@
 namespace Livewire\Commands;
 
 use Illuminate\Console\Command;
+use League\Flysystem\Cached\CachedAdapter;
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks Josh!

PR #4651, opened 2022-02-11
Update multiple root warning link

joshhanley:
A new troubleshooting section was added to the docs for handling multiple root elements https://github.com/livewire/docs/pull/411

This PR updates the warning console message that Livewire emits when it detects multiple root elements so that it points to this new section in the docs.

image

Hope this helps!


calebporzio:
Thanks Josh!

PR #4668, opened 2022-02-14
Extend events

mikield:

Yes. Discussion link: https://github.com/livewire/livewire/discussions/4580#discussioncomment-2170018

Yes.

No. The code is backward compatible.

Yes. New `ComponentEventInstanceTest.php` test file presented.

This code is most useful to be more strict with events using class names. 
Eventually when my page components are listening to global events of other components, instead of remembering the string representation of the event - I can use the event's class name.

Livewire `Event.php` class was renamed as proposed in the discussion.

Thanks for contributing! 🙌

### DIFF
--- a/src/ComponentConcerns/ReceivesEvents.php
+++ b/src/ComponentConcerns/ReceivesEvents.php
@@ -74,6 +74,11 @@ trait ReceivesEvents
 
     public function fireEvent($event, $params, $id)
     {
+        if($event instanceof Event){
+            $params = $event->getParams();
+            $event = $event->getName();
+        }
+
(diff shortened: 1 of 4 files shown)


joshhanley:
@mikield oh I had forgotten Livewire had an event class already! I like that is makes use of it, but I wouldn't change the name of it like I suggested. So I would change that back to just `Event`.

Thanks!

mikield:
@joshhanley have updated the code.

joshhanley:
@mikield great thanks!

mikield:
@joshhanley as I see one of test case has not passed. But that test was not affected by my changes.

PR #4669, opened 2022-02-14
Set array input one by one in tests

mnabialek:
Today I've encountered the following problem in tests:

```
        ->set([
                'files' => [
                    'type_1' => [
                         TemporaryUploadedFile::fake()->image('sample.jpg', 521, 732),
                         TemporaryUploadedFile::fake()->image('sample2.png', 1001, 1702),
                         TemporaryUploadedFile::fake()->image('sample3.png', 1510, 1920),
                      
                    'type_2' => [],
                ],
                'settings' => [
                    'some_setting' => 'setting value',
                ],
            ])

```

In Blade templates I had usage of `$file->temporaryUrl()` that failed (no method found) and when I commented using this method in Blade I got:

```
InvalidArgumentException : Type is not supported
 /usr/share/nginx/html/vendor/laravel/framework/src/Illuminate/Http/JsonResponse.php:88

```

The problem is that `syncInput` makes changes for UploadedFile using some assumptions and is not implemented to handle above case. Implementation of it looks like this:

```
    public function syncInput($name, $value)
    {
        if ($value instanceof UploadedFile) {
            return $this->syncUploadedFiles($name, [$value]);
        } elseif (is_array($value) && isset($value[0]) && $value[0] instanceof UploadedFile) {
            return $this->syncUploadedFiles($name, $value, $isMultiple = true);
        }

        return $this->sendMessage('syncInput', [
            'name' => $name,
            'value' => $value,
        ]);
    }
```

For above array it wouldn't sync uploaded files and it finally causes problem because probably json_encode is not able to deal with uploaded files. Probably `syncInput` method (used for tests) could be also refactored (now value should not be array anymore) but for now I think it's safer to leave it without change.

Just in case, there is possible workaround to make this working (without any Livewire code changes), when in tests you use:

```
    ->set('files.type_1.0', TemporaryUploadedFile::fake()->image('sample.jpg', 521, 732))
    ->set('files.type_1.1', TemporaryUploadedFile::fake()->image('sample2.png', 1001, 1702))
    ->set('files.type_1.2', TemporaryUploadedFile::fake()->image('sample3.png', 1510, 1920))
```

then everything is working as expected

### DIFF
--- a/src/Testing/Concerns/MakesCallsToComponent.php
+++ b/src/Testing/Concerns/MakesCallsToComponent.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Testing\Concerns;
 
+use Illuminate\Support\Arr;
 use function Livewire\str;
 use Illuminate\Support\Str;
 use Illuminate\Http\UploadedFile;
@@ -64,7 +65,7 @@ trait MakesCallsToComponent
     public function updateProperty($name, $value = null)
     {
         if (is_array($name)) {
-            foreach ($name as $key => $value) {
+            foreach (Arr::dot($name) as $key => $value) {
                 $this->syncInput($key, $value);
             }
 
--- a/tests/Unit/DataBindingTest.php
+++ b/tests/Unit/DataBindingTest.php
@@ -4,6 +4,8 @@ namespace Tests\Unit;
 
 use Livewire\Component;
 use Livewire\Livewire;
+use Livewire\TemporaryUploadedFile;
+use Livewire\WithFileUploads;
 
 class DataBindingTest extends TestCase
 {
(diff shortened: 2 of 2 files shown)


joshhanley:
@mnabialek thanks for the PR! Can you add tests for this?

Thanks!

mnabialek:
@joshhanley To be honest I'm not sure how to write test for it as test would be for functionality that is used only in tests

joshhanley:
@mnabialek essentially you would just write a small component and test that makes use of the change (basically the code you shared as a test), which passes with your changes but fails without 🙂

joshhanley:
@mnabialek thanks! I have triggered the tests to run

calebporzio:
I think I'm going to hold off on this update because it changes existing Livewire behavior. By flattening the array using `Arr::dot()` you would be calling "syncInput" many more times for deeply nested data than is currently happening.

Thanks for the contribution, but I'm going to hold off for now!

mnabialek:
@calebporzio So I believe `syncInput` should be fixed then. Also keep in mind that from what I understand this method is only used for tests (it's in `Livewire\Testing\Concerns`) so it doesn't affect anything but tests

mnabialek:
@joshhanley I don't see any easy workaround. Using `Arr::dot` caused that every value was processed as it should be, and without it, uploaded files are processed as expected only if they have expected structure (so simple array) or if you manually set such files using dot notation as I showed, otherwise you get exception

joshhanley:
@mnabialek yeah fair enough. So doing it manually works (albeit not as clean)?

mnabialek:
@joshhanley Yes, exactly, when manually using dot notation to upload files in tests it works fine, but using array notation with custom keys exception is thrown because Livewire built-in helper test methods cannot handle custom array with uploaded files

PR #4670, opened 2022-02-14
Fix multiple trait method executions

jelib3an:
Fixes #4666 

Each time we call `set()` in a test, it boots the component which creates an array of methods for the hooks. The problem is, it is appending to the existing array instead of starting with an empty array each time. And so each time you call `set()`, it is re-running all the previous hooks.

In non-test environment, this is fine because the component only ever gets booted once. But we expect our tests to work as they would in production. Tests are failing if they are only expecting hooks to run once over the course of the test.


joshhanley:
@jelib3an thanks for the PR!

Can you please update your test to match the style of the other test that is there rather than adding the `Counter` class?

Thanks!

jelib3an:
@joshhanley Sure, I can do that. I also didn't realize I should stick with php7.2 syntax. If I created a third trait and class for the test, would that be okay? The only thing is I have to use a static variable to keep track of the counts since it has to be "global". Thanks!

joshhanley:
@jelib3an thanks! Yeah we support from PHP 7.2+ probably until the next major release of Livewire.

In terms of keeping track of the counts, if you have a look at the other test there it just adds properties to an array each time a method is called (like a trait hook), so you should be able to follow the same pattern and make sure the resulting array doesn't have a duplication of say "traitUpdated".

Hope that makes sense!

jelib3an:
I did try that, unfortunately it didn't work. Otherwise the existing test would have caught this.

The issue is this line here. https://github.com/jelib3an/livewire/blob/fix-multiple-trait-method-execution/src/Features/SupportComponentTraits.php#L35

It's saving the instance of the component. Each time the test boots a new component, it's a different instance. But they all have the same id. So you're actually re-running hooks from the previous instances. Which explains why I'm seeing old values during debug (screenshot in #4666)

jelib3an:
@joshhanley I modified the test, would you please see if it looks okay? I had some trouble figuring out what to name the trait and class, lol.

Actually... do you think it would make sense to modify the existing tests to use static property instead of adding a new trait and class? :thinking:

calebporzio:
Thanks for this! I reworked it a bit, but your test and implementation were very helpful!

PR #4678, opened 2022-02-15
Pass layout data to layout component constructor

brizzo:
When rendering full-page components that use a layout component class, the data passed as the second argument to the layout function is not passed along to the layout component's constructor. For example:

```
public function render()
{
    return view( 'livewire.my-component' )
        ->layout(\App\Views\Components\AppLayout::class, [
            'title' => 'Meta Title Set Here',
            'heading' => 'Page Title Here',
        ]);
}
```

In the `layout` function defined in `Livewire\Macros\ViewMacros`, notice that when the layout component is instantiated, `$layout = new $view();`, the data in the `$params` variable is not passed to the layout constructor. This PR fixes that by using the `makeWith` function from Laravel's service container to create the layout instance and pass along the arguments in the `$params` variable to the constructor of the layout component class.

This is useful for cases where layout components manipulate or initialize data based on other arguments in their constructor. Consider this simple example where the AppLayout component constructor sets `$title` to the value of `$heading` if not explicitly set.
```
public function __construct($title = null, $heading = null)
{
    $this->title = $title ?? $heading;
    $this->heading = $heading;
}
```

### DIFF
--- a/src/Macros/ViewMacros.php
+++ b/src/Macros/ViewMacros.php
@@ -27,7 +27,7 @@ class ViewMacros
             unset($params['attributes']);
 
             if (is_subclass_of($view, \Illuminate\View\Component::class)) {
-                $layout = new $view();
+                $layout = app()->makeWith($view, $params);
                 $view = $layout->resolveView()->name();
             } else {
                 $layout = new AnonymousComponent($view, $params);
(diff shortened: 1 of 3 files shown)


joshhanley:
@brizzo thanks for the PR! Any chance you could add test for this?

Thanks!

joshhanley:
@brizzo thanks for adding tests! I have triggered the tests to run

calebporzio:
Makes sense, thanks!

PR #4685, opened 2022-02-16
Add assertBrowserEventIsNotDispatched() method.

reziamini:
There are some situations which we need to check the browser event isn't dispatched in our tests.

With this method, it will be accessible.

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -245,6 +245,20 @@ trait MakesAssertions
 
         return $this;
     }
+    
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for this!

calebporzio:
(renamed the event btw 😉 )

joshhanley:
@rezaamini-ir any chance you could PR this to the docs? Thanks!

rezaamini-ir:
@joshhanley Yes!
I’ll add the docs also.

joshhanley:
@rezaamini-ir thanks!

PR #4688, opened 2022-02-17
bug: respect modified properties by `prepareForValidation` during checking rules

comes:
Uh... no.. sorry 🙃

of course

no

no, not needed I think

Imagine you have a component with a computed property; nothing special, nothing unusual. 

```php
namespace App\Http\Livewire;

use Livewire\Component;

class SampleComponent extends Component
{
    protected $rules = [
        'foo' => 'gt:5'
    ];

    public function getFooProperty(): int
    {
        return mt_rand();
    }

    public function render()
    {
        return view('livewire.sample-component');
    }

    public function something()
    {
        $this->validate();
    }
}
```

When we try to call `$this->validate()` we will get an exception `new \Exception('No property found for validation: [foo]')`, which makes actually a lot of sense. In `Livewire\ComponentConcerns\ValidatesInput::getDataForValidation()` is a loop, which checks if there is a public property in the SubClass which is represented by a rule. We got the announced exception if there is a rule without a property.

By default, livewire components respect the function `protected function prepareForValidation($attributes)`. If we modify our Component above like follows, we modify the attribute set, which should be validated. In theory, this works like a charm, but livewire tries to figure out, is there a public property regarding our rules.

```php
namespace App\Http\Livewire;

use Livewire\Component;

class SampleComponent extends Component
{
    protected $rules = [
        'foo' => 'gt:5'
    ];

    protected function prepareForValidation($attributes) {
        $attributes['foo'] = $this->getFooProperty();
        return $attributes;
    }


    public function getFooProperty(): int
    {
        return mt_rand();
    }

    public function render()
    {
        return view('livewire.sample-component');
    }

    public function something()
    {
        $this->validate();
    }
}
```

My PR moves the rules-validation-loop one step later to respect attribute manipulation by `prepareForValidation()`.

Let's discuss, please.

greetings, J


joshhanley:
@comes thanks for the PR! Any chance you could add tests for this?

Thanks!

comes:
Hej @joshhanley,

I'd implemented a small test and fix the PHP <8.0 compatibility.

have a great day, J

calebporzio:
Please add this to ->validateOnly as well before we consider - thanks!

comes:
Hej @calebporzio, 

as mentioned, I implemented in `->validateOnly()` aswell. needed to move some logic around. Every tests runs well (on my end)

all the best, j

joshhanley:
@comes thanks for that! I have triggered the tests to run.

comes:
@joshhanley, I can not see in which way my changes are responsible for the failing tests.

joshhanley:
@comes oh it looks like you have submitted a GitHub workflow as part of your PR. Can you remove that. Maybe that is causing issues?

comes:
that makes sense. deleted it.

comes:
Seems to be a general problem, other PRs failing as well with nearly the same messages.

Let me please know, how I can solve it, @joshhanley. I am a bit lost on my end. :(

joshhanley:
@comes yeah there was an unrelated failing test due to an upstream change in Laravel. I have fixed it and merged it in here.

joshhanley:
@comes all looking good now!

comes:
puh! thanks @joshhanley. Nice journey and my first experiences in collaboration with open source stuff.. great work, all together.

enjoy your week

joshhanley:
@comes no worries! Awesome! It can be very rewarding 🙂

comes:
Hej there, I don't want to create unnecessary pressure, but are there any chances, that this PR will be merged in the near future?

all the best and take care!

joshhanley:
@comes not sure, Caleb will probably review it when he next goes through PR's 🙂

calebporzio:
Thanks a lot for this @comes!

PR #4699, opened 2022-02-17
Handle custom routing requests when present

eric-hansen:

Created a discussion: https://github.com/livewire/livewire/discussions/4698

Yes

No

No, if issue let me know as not sure exactly where/how this should be tested nor does a simple `composer install` make the whole test-suite pass.  Using Docker so unsure of what packages are missing if any.

I'd like to say everything that needs to be said is in the discussion, but if not then let me know and I can edit this to elaborate more.

Thanks for contributing! 🙌

### DIFF
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -10,6 +10,7 @@ use Livewire\GenerateSignedUploadUrl;
 use Illuminate\Routing\RouteCollection;
 use Illuminate\Support\Traits\Macroable;
 use Facades\Livewire\GenerateSignedUploadUrl as GenerateSignedUploadUrlFacade;
+use Illuminate\Support\Str;
 use Livewire\Component;
 use Livewire\Exceptions\PropertyNotFoundException;
 use Livewire\LivewireManager;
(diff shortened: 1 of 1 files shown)


calebporzio:
Closing this because no test has been provided. By providing a test, we can ensure this fix is fixed forever and this bug isn't re-introduced. Please re-submit after adding a test. Thanks!

eric-hansen:
> Closing this because no test has been provided. By providing a test, we can ensure this fix is fixed forever and this bug isn't re-introduced. Please re-submit after adding a test. Thanks!

2 questions:

1) Since this seems to only be affected when using `app()`/`request()`, is this testable?
2) If tests are a pre-req, then why was #4619 merged without one?

joshhanley:
@eric-hansen yeah you should be able to test it, you said you had a test that fails without this change, so maybe try and replicate that test and component.

Somethings can be too difficult to test with Livewire due to the nature of the front end interaction, so occasionally (rarely), will a PR be approved without tests. But as it was, that was an oversight, I have just added tests for it 🙂

> nor does a simple composer install make the whole test-suite pass.
In regards to this comment above, a composer install isn't enough to get Livewire's tests running due to using Laravel Dusk. Please have a look at our contribution guide as to the extra steps required https://laravel-livewire.com/docs/2.x/contribution-guide

I don't know if you can run dusk in docker or not, as I don't use docker (I use valet), so that is something you will need to work out.

Hope this helps!

eric-hansen:
@joshhanley I'll see if I can actually write a test for it.  The test I wrote was specific to the application I use Livewire for, not a test for Livewire itself.  So if the helpers mentioned a few times now are exposed as well then it may be doable.  If not then I hope we can assess whether a test is necessary to move forward.

As for Dusk, et. al. I'll read deeper and see what happens.

Thanks for the feedback.

joshhanley:
@eric-hansen yeah generally that is how I come across bugs too.

So what I like to do is copy the code from my main app, into a small test app with Livewire, then I try to delete as much code as I can until I can replicate the issue with as little code as possible.

That then become the code for the test which I submit to Livewire. Most of Livewire's tests are components that look and act exactly the same way they would in an app (which Dusk is really great for simulating).

No worries!

eric-hansen:
Should we continue any discussion regarding testing and such here or in the discussion thread?

Not usually contributing to projects that enforce the Discussion feature of GitHub so not sure what is preferred here.

PR #4706, opened 2022-02-20
Add new computed property declaration

hebinet:
This PR adds a new method to declare computed properties for livewire components.

Old way:
```php
    public function getFooBarProperty()
    {
        return strtolower($this->upperCasedFoo);
    }
```
New way:
```php
    public function fooBar(): Property
    {
        return new Property(fn() => strtolower($this->upperCasedFoo));
    }
```
or

```php
    public function fooBar(): Property
    {
        return Property::get(fn() => strtolower($this->upperCasedFoo));
    }
```

This is the same approach which laravel takes with the new Attribute Accessor/Mutator ( https://laravel.com/docs/9.x/eloquent-mutators#defining-an-accessor )

I like this approach, because I don't have to write a function like `get{Attribute}Property` and with the return type, it is clear that this is an computed property.

This feature can be taken as an additional method and is not meant to be a replacement.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -303,7 +303,7 @@ abstract class Component
         if (method_exists($this, $computedMethodName = Str::camel($studlyProperty))
             && ($returnType = (new ReflectionMethod($this, $computedMethodName))->getReturnType())
             && $returnType instanceof ReflectionNamedType
-            && $returnType->getName() === ComputedProperty::class) {
+            && $returnType->getName() === Property::class) {
             if (isset($this->computedPropertyCache[$property])) {
                 return $this->computedPropertyCache[$property];
             }
--- a/src/Component.php
+++ b/src/Component.php
@@ -14,6 +14,8 @@ use Livewire\Exceptions\PropertyNotFoundException;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Livewire\Exceptions\CannotUseReservedLivewireComponentProperties;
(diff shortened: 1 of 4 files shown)


joshhanley:
@hebinet thanks for the PR! I've triggered the tests to run.

joshhanley:
@hebinet tests were failing due to an unrelated error. I have merged master in and will run the tests again.

eric-hansen:
I just wonder why not use the `Attribute` flow that Laravel 9 has?

hebinet:
> I just wonder why not use the `Attribute` flow that Laravel 9 has?

Yeah, you are right, I considered using the Laravel Attribute class too, but after a little Deep-Dive into both Laravel and Livewire code it just didn't feel right.

- The first thing was, that the Laravel Attribute class has a getter and setter which is totally unnecessary for a computed property which only has a getter.

- The second thing was the naming itself. In Livewire the terminology `get{field}Property` is used and to replace it with the : Attribute Typehint seems confusing to me. So I came up with the new Property class.
This seemed more appealing to me as an adequate alternative.
- The new Property class is also only a simple DTO so no duplicate logic here.

hebinet:
@joshhanley Any updates on this?

joshhanley:
@hebinet not from me, Caleb will look at it when he goes through PR's next 🙂

calebporzio:
Interesting idea, but because computed's are only "getters" I don't think the pattern maps very well - also I don't want to introduce a new syntax like this without thinking deeper through its implications. Thanks for the novel idea and being willing to contribute!

PR #4712, opened 2022-02-22
Fix file is deleted but property still references deleted file if filename mismatches.

hasyirin:

No, I did not create any discussion about it.

Yes: fix-remove-upload

No

I need to explore how to code tests first...

If we were to trigger the upload:remove event to a property with a different filename, the file is deleted however it is still being referenced. This will throw exceptions if we were to use the value.

I'm still not sure whether it's better to match both property name and filename as my PR or ignore filename entirely if it's a single file.

Thanks for contributing! 🙌

### DIFF
--- a/src/WithFileUploads.php
+++ b/src/WithFileUploads.php
@@ -87,12 +87,12 @@ trait WithFileUploads
 
                 return true;
             })));
-        } elseif ($uploads instanceof TemporaryUploadedFile) {
+        } elseif ($uploads instanceof TemporaryUploadedFile && $uploads->getFilename() === $tmpFilename) {
             $uploads->delete();
 
             $this->emit('upload:removed', $name, $tmpFilename)->self();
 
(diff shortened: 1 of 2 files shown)


joshhanley:
@AhrimFakhriy thanks for the PR! Can you please add tests? Thanks!

AhrimFakhriy:
@joshhanley Hi, I added a test. Still not sure whether to ignore the filename or ensure filename matches when removing a single file property.

joshhanley:
@AhrimFakhriy tests were failing due to an unrelated error. I have merged master in and will run the tests again.

AhrimFakhriy:
That’s odd. Another failed test that I dont think is related?

joshhanley:
@AhrimFakhriy ah yeah we have some flaky tests which fail in CI sometimes. If you have a look you can see a few test suites have actually passed, so that's generally what we look for, so all should be good!

calebporzio:
Looks great and makes sense. Thanks!

PR #4722, opened 2022-02-24
Dont override route when defined

eric-hansen:

I did create a discussion, and submitted a previous PR which I addressed the feedback here (#4699 )

Yup

Nope

Yup

See discussion #4698 and previously mentioned PR.

Thanks for contributing! 🙌


joshhanley:
@eric-hansen thanks for the PR!

I've triggered test to run but there seems to be an issue with compatibility in PHP 7.3. Can you have a look at it?

Also there is an unrelated issue on master causing other tests to fail that has been fixed, so if you could merge master in while you are at it, that'd be great!

Thanks!

eric-hansen:
@joshhanley sure thing, I fixed the error for that (thought arrow funcs were in 7.0+ but guess it was later).  I also merged in the latest master as of current.

joshhanley:
@eric-hansen thanks! I've triggered the tests to run

eric-hansen:
@joshhanley I'm kind of wondering if the failed jobs are even related to my change.  But if you feel it is then I hope we can figure out a way for me to properly run the GitHub actions or something to properly ensure they are fixed.

I don't really use GitHub Actions myself so not super familiar with being able to trigger them or anything.  But it seems like I can't run them on my repository (probably due to the `on` structure set).

joshhanley:
@eric-hansen normally we do have the odd flaky test, but if a couple of the suites pass then it generally all good.

But it looks like there is an issue possibly with your PR as this test is consistently failing.

image

Have you checked out our [contribution guide](https://laravel-livewire.com/docs/2.x/contribution-guide) and are you able to run the tests locally? They should all pass locally first.

Regarding GitHub actions, tests will get triggered to run automatically, unless you are a first time contributor (haven't had a PR merged), in which case maintainers need to manually trigger the tests to run for you.

calebporzio:
Thanks for the contribution, going to close it however. You should be able to accomplish what you want by instead of retrieving the request parameters through the request helper, you can access them through mount parameters and access like so: `Livewire::test(ShowFoo::class, ['foo' => 'bar'])`

Hope that helps, thanks!

eric-hansen:
If the component isn't storing the route parameter as a property then that doesn't work.  I.e.: if I want to get the `client_id` param from the route simply to get the full `Client` object and store the `Client` object as a property, passing the ID in the way you describe still leads to the business logic failing.

PR #4733, opened 2022-02-25
Fix assertion problem

reziamini:
Refer to #4707.


joshhanley:
@rezaamini-ir thanks!

PR #4765, opened 2022-03-05
Allow dispatching CustomEvent with falsy value

adriaanzon:
It relates to #4617. It's not a fix for using `$dispatch('input', null)`, but at least it allows you to dispatch other falsy values such as `''`, `0` or `false`. I don't know a fix for `$dispatch('input', null)`, because you would have to distinguish the CustomEvent from the one Safari sends, where the value of `e.detail` is also `null`.

Yes

No

Yes

See my demo of the problem in the original post of #4617.


joshhanley:
@adriaanzon thanks for the PR! I've triggered the tests to run

adriaanzon:
The tests were failing. I saw some fixes for the test suite on the master branch so I have rebased my changes.

joshhanley:
@adriaanzon thanks! I have triggered them again

calebporzio:
Thanks! Makes sense!

PR #4772, opened 2022-03-08
Add logic for echo presence listen channel #3071

davidpiesse:
This PR allows for using the Listen event within an Echo presence channel.

@joshhanley this is for you ! :) 

The syntax now is:
```php
"echo-presence:presence-ping,PresencePong" => 'presencePong', //listen
// And the others that stay the same
"echo-presence:presence-ping,here" => 'presencePong', //here
"echo-presence:presence-ping,joining" => 'presencePong', //joining
"echo-presence:presence-ping,leaving" => 'presencePong', //leaving
```

The code now checks if the event name is one of the reserved three (here, joining, listening) and if not uses the value as the event name (usually the Events class name).

This should fix #3071 and the related issues and discussions.

If merged in I will then update the docs repo with the new syntax and improve the examples.

Sorry if I built / updated things wrong in the PR


joshhanley:
@davidpiesse thanks for the PR! I will pull it down when I get the chance and test it out 🙂

calebporzio:
This is perfect, thanks @davidpiesse!

PR #4593, opened 2022-01-30
fix exception in dynamic listener

mahdi4187:
fix undefined array key in dynamic listener

Review the contribution guide first at: https://laravel-livewire.com/docs/2.x/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

4️⃣ Does it include tests? (Required, where possible)

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

Thanks for contributing! 🙌

PR #4778, opened 2022-03-08
Support dynamic listeners

silnarm:

There exist previous discussions (#3056 & #3116) and two previous PRs (#2427 & #4593).
One of those PRs is still open, and perhaps I am jumping the gun here, but it has been a little while and looks like it is going the way of the other one.

Y.

N.

Y.

Adds support for dynamic component listeners. 
The previous two PRs (mentioned above) attempt to treat the worst symptom by ignoring an error condition on the server side.
This one fixes the actual problem, by adding support for listeners being added/removed after the initial render, so the frontend doesn't send any message if you emit an event that is no longer being listened for, and the fontend will send a message if you emit an event that is now being listened for, but wasn't on the initial render.

Thanks for contributing! 🙌


joshhanley:
@silnarm thanks for the PR! I've triggered the tests to run.

Don't worry about the other open PR, I don't suspect they will be finishing it as requested, so happy to go with your implementation.

Can you fix up your code so that all the docblock type hints and unused imports are not removed as well as the `component.initialised` listener, so that way this PR is only your additions?

Also not sure why `scrollTo` was removed, please add that back in. If it's unused, can you move that to a seperate PR?

Thanks!

silnarm:
Hi @joshhanley,

Re: scrollTo, removing things that the IDE says are unused in a dynamic language like PHP is of course not wise, the fact that it was calling `scrollTo` on an imaginary `browser` property was what caught my eye, because `$this` is a Browser object, which has no such prop. It does have a `scrollTo()` though, and being macros, that can't be overridden here, so it is inaccessible as well as erroneous.

You are most correct in that it has nothing to do with this PR though, will fix that up (and the other removals), and fix up that silly assert that was always going to fail with different PHP versions.

I am also intending to add a test for embedded components, hence the draft status.

Cheers.

joshhanley:
@silnarm no worries, thanks for fixing it up!

Have triggered the tests 🙂

calebporzio:
Here are my two hesitations with this:
1) This will send the full list of component listeners in every single Livewire response for that component. No other effect currently does this. (thinking specifically about payload size affecting speed)
2) Even though people might assume Livewire handles this by default, it currently doesn't and by supporting this, we are forced to keep that functionality in future versions - which could pose problems if we choose a different front-end listener strategy.

Thanks for the contribution, however for now I'm going close this.

PR #4803, opened 2022-03-14
[2.x] Add support for nested `$wire` property updates

ryangjchandler:

No discussion, but I've been wanting this myself for a little while now.

Yes.

No.

Rudimentary - will be expanded once I've received approval for continuation from maintainers.

At the time of writing this, there are multiple ways to interact with a Livewire component from JavaScript (specifically Alpine.js):

- You can use the `@entangle` directive to store a mutable copy of a Livewire property on an Alpine.js component and interact with the "local" data property.
- You can choose to use `$wire.entangle()` instead of the Blade directive. This is generally more consistent with regular Alpine.js usage since it involves zero server-side processing.
- You can use `$wire` like an object to retrieve properties on your Livewire component.

The first 2 options in the list allow you to mutate data from Alpine with ease, but if you have a few mutable properties you end up with an odd mixture of Alpine.js state as well as Livewire state that requires continuity.

The 3rd option in the list also supports mutating data but only on the top-level, i.e. `$wire.foo = 'bar'` will work perfectly, but `$wire.foo.bar = 'baz'` will not.

I'm opening this PR now to see how much interest there is in supporting deeply-reactive `$wire` property access and mutation through the use of recursive `Proxy` objects.

The code in this PR so far implements a rough working version, as well as a very simple test (replicates my local testing scenario).

With these changes, you're able to use `$wire` to modify nested properties and array keys from Alpine.js which is great for usage with `x-model`, `x-bind:y`, etc.

If there is enough interest and a nod from maintainers, I'll get this to a merge-ready state. Thanks for reading, sorry for the essay! 😬

> **EDIT (2022-03-14)** - I think it's worth noting that these changes shouldn't introduce any breaking changes since `Proxy` objects are transparent and typical `typeof` checks or `Array.isArray` checks still return the same value.


joshhanley:
@ryangjchandler I think that this is a nice addition! We can currently use `$wire.set('foo.bar') = 'baz'` to set nested properties values anyway, so this is essentially just syntactic sugar!

ryangjchandler:
@joshhanley Is it correct that workflows/tests are failing on `master` btw?

joshhanley:
@ryangjchandler I've run the tests again, and there is something strange going on with your new test in older php versions

image

ryangjchandler:
Hm, this is strange since I haven't touched anything PHP related besides the tests... maybe I need to add a longer delay between updating the inputs and making assertions.

joshhanley:
@ryangjchandler maybe try using `->waitForLivewire()` instead of the pause like `->waitForLivewire()->type('@foo-input', 'car')`

ryangjchandler:
@joshhanley Ah, good shout. Think I nabbed the delay from another test, will update now!

calebporzio:
Hey @ryangjchandler, yeah, I see the need for something this.

My main hesitancy is the maintenance overhead involved in introducing a recursive proxy into the codebase.

My vision for v3 is to have an actual Alpine reactive object to represent the state of each component. This way when that data is changed everything just reacts. And when you change that data, Livewire reacts with the proper network requests.

That is the perfect world to me.

Doing something like this in the meantime might be helpful, but I just want to make sure merging this doesn't spawn many future issues with that part of the codebase.

Thoughts either of you?

ryangjchandler:
@calebporzio I can't think of an easier way to do this without a recursive proxy (until Alpine.reactive and interceptors can be used). 

Do you have a timeline for Alpine in the package core? I know @joshhanley has been working on using Morph, but I imagine using Alpine itself will be a v3.x thing since there's likely breaking changes.

joshhanley:
Hmm yeah I have run into this before a few times. I can see the potential issue of maintaining the recursive proxy. Maybe there might be a way to use an object which represents the property data that has a setter on it, but each layer would need to be that object, so then you may as well just use the proxy.

Yeah using Alpine will be V3, not sure what the timeline is on that. Morph has had to be pushed back to V3 due to breaking changes.

calebporzio:
Ok, I've thought about this and I think I want to hold off on it for this version. I hope to solve this in v3, but I'd say for these deeper cases, just stick with `$wire.set()` for now. Thanks for the contribution!

ryangjchandler:
Shame, will look forward to a more consistent API in v3. Thanks!

PR #4806, opened 2022-03-15
Fix #4789 query-string-with-array-property

nicolaferrara82:
Yes

Yes

No

Yes

This fix #4789

### DIFF
--- a/src/Features/SupportBrowserHistory.php
+++ b/src/Features/SupportBrowserHistory.php
@@ -40,7 +40,9 @@ class SupportBrowserHistory
                     ? json_decode(json_encode($fromQueryString), true)
                     : json_decode($fromQueryString, true);
 
-                $component->$property = $decoded === null ? $fromQueryString : $decoded;
+                $result = $decoded === null ? $fromQueryString : $decoded;
+
+                $component->$property = is_array($component->$property) ? array_merge($component->$property, $result) : $result;
             }
         });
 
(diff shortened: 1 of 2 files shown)


silnarm (commented):
This seems like it is probably a breaking change.

Your use case certainly seems reasonable, but I imagine there are people doing things that rely on the current behaviour. I am far from the best positioned to judge this though.

Regarding the fix, the use of `array_merge()` is going to do horrible things for people using numeric indices, you probably want `array_replace()`.
eg:
```php
    // ...
    /** @test */
    public function it_doesnt_spam_a_numerically_indexed_array_from_query_params()
    {
        Livewire::withQueryParams(['foo' => [0 => 'foo', 1 => 'bar']])
                ->test(QueryParamsWithIndexedArrayPropertyComponent::class)
                ->assertSet('foo', [0 => 'foo', 1 => 'bar']); // fail.
    }
    // ...
}

// ...

class QueryParamsWithIndexedArrayPropertyComponent extends Component
{
    public $foo = [
        0 => 'foo',
        1 => 'bar',
    ];

    protected $queryString = ['foo'];

    public function render()
    {
        return app('view')->make('null-view');
    }
}
```

joshhanley:
@nicolaferrara82 thanks for the PR! I've triggered the tests to run. Can you please update your description with details of what this PR does/solves, so that others don't need to go through the discussion to find out? Thanks!

calebporzio:
Sorry, I'm closing this because there is no description of the problem this PR solves in this PR's description. Re-open if you've provided a clear and concise description of the problem this solves. Thanks!

PR #4822, opened 2022-03-19
Allow Livewire to work from a subdirectory

BassemN:
Yes

Yes

No

I don't think it needs a test

I'm running my Laravel app from a subdirectory on my domain so the Livewire javascript file doesn't load. please check the before and after code

before
`<script src="/livewire/livewire.js?id=940557fc56b15ccb9a2d" data-turbo-eval="false" data-turbolinks-eval="false" ></script>`

after
`<script src="https://example.com/subdirectory/livewire/livewire.js?id=940557fc56b15ccb9a2d" data-turbo-eval="false" data-turbolinks-eval="false" ></script>`

if the Laravel app is placed on the main domain then the code will be
`<script src="https://example.com/livewire/livewire.js?id=940557fc56b15ccb9a2d" data-turbo-eval="false" data-turbolinks-eval="false" ></script>`

### DIFF
--- a/src/LivewireManager.php
+++ b/src/LivewireManager.php
@@ -223,10 +223,10 @@ HTML;
     {
         $jsonEncodedOptions = $options ? json_encode($options) : '';
 
-        $assetsUrl = config('livewire.asset_url') ?: rtrim($options['asset_url'] ?? '', '/');
+        $assetsUrl = config('livewire.asset_url') ?: rtrim($options['asset_url'] ?? url(''), '/');
 
         $appUrl = config('livewire.app_url')
-            ?: rtrim($options['app_url'] ?? '', '/')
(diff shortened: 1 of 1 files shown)


joshhanley:
@BassemN thanks for the PR! Doesn't updating your `asset_url` or `app_url` in the config allow for this already?

BassemN:
Hello @joshhanley
Yes, updating the `app_url` will make it work but I think adding plain URL `''` to the core is not good practice and the most suitable is to use the `url('')` function.
This issue was also on the Laravel framework error pages and I submitted a PR (https://github.com/laravel/framework/pull/25725) and got approved.
If you need more details please let me know.

calebporzio:
Thanks for submitting this, but like @joshhanley said, this is already possible and I want to avoid breaking existing apps with a change like this - thanks!

PR #4824, opened 2022-03-21
remove test flag from MoveCommand

Naoray:
This PR removes the `--test` flag from the `livewire:move` command. As discussed in #4820 we will now check whether a test file is present and if it is rename the test.

```php
php artisan livewire:make Foo --test

// moving with tests before
php artisan livewire:move Foo Bar --test
`
// moving with tests now
php artisan livewire:move Foo Bar
```

### DIFF
--- a/src/Commands/MoveCommand.php
+++ b/src/Commands/MoveCommand.php
@@ -6,7 +6,7 @@ use Illuminate\Support\Facades\File;
 
 class MoveCommand extends FileManipulationCommand
 {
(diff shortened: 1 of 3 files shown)


calebporzio:
Makes sense, thanks!

calebporzio:
Is there feedback when a user runs this so that they know the test was moved as well?

Naoray:
@calebporzio yes

calebporzio:
Ok great, thanks!

PR #4836, opened 2022-03-26
Ci test hardening

nuernbergerA:
* removed `pause` because `waitUsing` will run every 25ms for 5s so there is no need to initially wait
* check every 25ms instead of 50ms so we check more frequently if we received a livewire response
* replaced random value from a random number between 0 and 1000 to a random string with 16 chars (so the chances are super low to overlap)


calebporzio:
Thanks @nuernbergerA!

PR #4838, opened 2022-03-28
Add support for custom query string

theupriser:
**Description**

I did not create a discussion thread about this but it's a feature I would like to have within livewire. With this change it possible to change the way the querystring works using a trait.

I've made changes to the SupportBrowserHistory class in which it checks if the component has access to the methods in the trait. If not, the standard will be used.

It does not include tests an extension within livewire which does nothing on itself.

**How to use this?**

In your component you include the WithCustomizedQueryString trait. This trait ensures your component inherits 2 functions: 
image
With these functions you are able to compose the url pushed to the browser (formatQueryString) and with the other function (formatQueryParameter) you are able to format the values of the query parameters to a usable string or array.

That works as following.
Normally a queryString URL looks like: ?array[0]=123&array[1]=456&string=abc
Now I want that same URL to look like this:  ?array=1,2,3&string=abc
In my Component class I added the WithCustomizedQueryStringTrait and filled the functions like this:

image

image

Now the URL formatting is like I would like to have it without changing the functionality within livewire.

I hope this is something other members/users also like to use. I thought of this as a usable and optional way of implementing this.

### DIFF
--- a/src/Features/SupportBrowserHistory.php
+++ b/src/Features/SupportBrowserHistory.php
@@ -36,6 +36,10 @@ class SupportBrowserHistory
                     continue;
                 }
 
+                if(method_exists($component, 'formatQueryParameter')) {
+                    $fromQueryString = $component->formatQueryParameter($fromQueryString);
+                }
+
                 $decoded = is_array($fromQueryString)
                     ? json_decode(json_encode($fromQueryString), true)
                     : json_decode($fromQueryString, true);
@@ -52,7 +56,11 @@ class SupportBrowserHistory
 
                 $queryParams = $this->mergeComponentPropertiesWithExistingQueryParamsFromOtherComponentsAndTheRequest($component);
 
-                $response->effects['path'] = url()->current().$this->stringifyQueryParams($queryParams);
+                if(method_exists($component, 'formatQueryString')) {
+                    $response->effects['path'] = url()->current().$component->formatQueryString($queryParams);
+                } else {
+                    $response->effects['path'] = url()->current().$this->stringifyQueryParams($queryParams);
+                }
             }
         });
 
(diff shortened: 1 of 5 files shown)


joshhanley:
@theupriser thanks for the PR! Can you please add tests? Thanks!

theupriser:
> @theupriser thanks for the PR! Can you please add tests? Thanks!

I will if I've got some time!

joshhanley:
@theupriser thanks! We require tests for the PR to be merged 🙂 See our contribution guide for details https://laravel-livewire.com/docs/2.x/contribution-guide#add-failing-tests

theupriser:
@joshhanley I've added a meaningful dusk test

theupriser:
The existing tests did not run from what I can see.

joshhanley:
@theupriser thanks! Yep, I had to approve them. Have triggered them now 🙂

theupriser:
> @theupriser thanks! Yep, I had to approve them. Have triggered them now 🙂

worked like a charm I see 👍

calebporzio:
Thanks for being willing to contribute this, however in the interest of keeping the API minimal I'm going to pass on this contribution. Thanks again!

theupriser:
Ahh that's a shame! what do you recommend on this then? I kind of need this very badly. Should I now keep my own fork with this addition?

zepfietje:
I'm not sure if the API introduced in this PR is the way to go. But I do want to say formatting arrays in the query string like `array=1,2,3` instead of `array[0]=1&array[1]=2&array[2]=3` is something that has come up as an issue multiple times already. I run into it myself time and again, and currently there seems to be no easy solution if any at all.

Reference:
- https://github.com/livewire/livewire/discussions/1252
- https://github.com/livewire/livewire/discussions/2036
- https://github.com/livewire/livewire/pull/3679
- https://github.com/livewire/livewire/discussions/4840


Maybe we should just find the right way to implement this behavior?

RealMrHex:
so what is the solution? query strings are extremely bad practice, there is no even a single way to customize or override the way that livewire use to manage them.

any solution or idea?

@calebporzio @zepfietje @theupriser @joshhanley

PR #4876, opened 2022-04-11
Skip @js Blade directive registration if already defined

lupinitylabs:
As discussed in https://github.com/livewire/livewire/discussions/4875, this PR skips registration of the @js custom blade directive unless the framework version is < 8.71.0 or running in the testing environment.

I am not a big fan of conditional execution based on the testing environment, but in the current case I don't see an option that will make sure that the tests will still always cover the custom implementation of the directive.

Also, I do not see an easy way to test the conditional registration itself, as this would require bootstrapping with a mocked App::version(). Not sure we want to go down that rabbit hole. Let me know if you think otherwise.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -57,7 +57,7 @@ class LivewireServiceProvider extends ServiceProvider
      *
      * @var string[][]
      */
-    static $bladeDirectivesToRegisterIfMissing = [
+    protected $bladeDirectivesToRegisterIfMissing = [
         'js' => [LivewireBladeDirectives::class, 'js'],
     ];
 
@@ -293,7 +293,7 @@ class LivewireServiceProvider extends ServiceProvider
 
     protected function registerBladeDirectives()
     {
-        foreach (static::$bladeDirectivesToRegisterIfMissing as $name => $callable) {
+        foreach ($this->bladeDirectivesToRegisterIfMissing as $name => $callable) {
             $this->registerBladeDirectiveIfNotRegistered($name, $callable);
         }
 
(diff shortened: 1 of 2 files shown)


lupinitylabs:
I went for the original approach, checking the framework version... but I see it could also be useful to go for your suggestion of looking up if the directive already exists. Should be easy enough to check with something like `method_exists(Illuminate\View\Compilers\BladeCompiler::class, 'compileJs')`...

Let me know if you prefer that option, and I'll fix something up for all of the definitions, and not just @js.

lupinitylabs:
I've been busy, so it took me a while to get back to this.
I have changed the implementation to only register the @ js directive if it's not yet registered.

I've only included the js directive for now, since I suppose the Livewire specific directives should always be overwritten.
If the directives should be replaceable as well, we could load them conditionally, too - but I'd be afraid that this might just result in a lot of support tickets if people accidentally overwrite the Livewire directives.

I kept the code for checking already registered Blade directives universal, so it should work not only for Blade directives defined in the Laravel core, but also for any custom directives (if any are registered before Livewire is bootstrapped).

lupinitylabs:
I was just having a discussion with @Gummibeer about app environments and had to think about the blade directive registration behaving differently on testing environment. While this makes sure that our tests run and the Livewire Js directive can be tested in Laravel >= 8.71.0, this also means that the directive will be overwritten when the app that uses Livewire is running tests, which will lead to slightly different results that will be unexpected for the developer (e.g., the use of the atob() function).

What do you think, should we instead overwrite the Blade directive manually during the test setup? I guess that should be working.

lupinitylabs:
It works... I am much happier with that implementation, so I pushed it. Let me know what you think.

lupinitylabs:
Duh, it doesn't make sense to register all directives in a test that's only for the js directive 😂 
I changed that accordingly.

PR #4916, opened 2022-04-25
Allow mixed parameters in only and except methods

jaxramus:
I want it, no discussion was made.

Yes, its on `patch-1`

No.

No. It's just a couple of changed lines, will work with existing tests.

Allow us to specify `only` and `except` properties the same way we can with Eloquent `only`, either mixed or as an array.

With this PR the following will work the same:

```php
$properties1 = $this->only('name', 'email');
$properties2 = $this->only(['name', 'email']);
```

It just makes the API a bit nicer and more flexible.


joshhanley:
@jaxramus thanks for the PR! Could you please add tests for this change? Thanks!

PR #4926, opened 2022-04-28
Support loading validation rules from traits.

emil-nasso:
:wave:

Traits are a great (and recommended) way to introduce some reusability into your livewire components.

To use lifecycle hooks inside your traits, but still be able to use them in your component, you can suffix them with the traits name as such:

```php
trait IsAnExample
{
    public function mountIsAnExample()
   {
       // Do stuff in addition to what is done in mount() in the component
   }
}
```

Hooks for query strings are also supported:

```php
trait IsAnExample
{
    protected $queryStringIsAnExample = [
        'example'
    ];
 
    // or as a method
 
    public function queryStringIsAnExample()
    {
        return ['example'];
    }
}
```

This PR adds the same functionality for validation rules. I use this for traits that extracts common functionality shared among a bunch of similar (but different :) ) forms.

```php
trait HasUsername
{
    protected ?string $username = null;

    public function rulesHasUsername()
    {
        return [
            'username' => 'required|email',
        ];
    }
}
```

If there is any interest in this I will:

* Add support for `protected $rulesIsAnExample`.
* Add support for validation messages.
* Add documentation.
* Add more tests?

Please let me know if you would be willing to accept this, if I work on it further.

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -77,36 +77,43 @@ trait ValidatesInput
         );
     }
 
+    protected function loadConfig(string $field)
+    {
+        $config = [];
+
+        foreach (class_uses_recursive($class = static::class) as $trait) {
+            if (method_exists($class, $method = $field . class_basename($trait))) {
+                $config = array_merge($config, $this->{$method}());
+            }
+            if (property_exists($class, $property = $field . class_basename($trait))) {
+                $config = array_merge($config, $this->{$property});
+            }
+        }
+
+        if (method_exists($this, $field)) $config = array_merge($config, $this->{$field}());
+        elseif (property_exists($this, $field)) $config = array_merge($config, $this->{$field});
+
+        return $config;
+    }
+
     protected function getRules()
     {
-        if (method_exists($this, 'rules')) return $this->rules();
-        if (property_exists($this, 'rules')) return $this->rules;
-
-        return [];
+        return $this->loadConfig('rules');
     }
 
     protected function getMessages()
     {
-        if (method_exists($this, 'messages')) return $this->messages();
-        if (property_exists($this, 'messages')) return $this->messages;
-
-        return [];
+        return $this->loadConfig('messages');
     }
 
     protected function getValidationAttributes()
     {
-        if (method_exists($this, 'validationAttributes')) return $this->validationAttributes();
-        if (property_exists($this, 'validationAttributes')) return $this->validationAttributes;
-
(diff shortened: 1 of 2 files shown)


danharrin:
This would be really cool, means that we can separate out the rules defined by the [Filament form builder](https://filamentphp.com/docs/forms) with those defined by the user outside of Filament.

calebporzio:
Makes sense - mind throwing an actual use-case out there? Even if it's pseudo-code?

I don't see any down-sides to this, you? As long as the actual component takes precedence over rules that exist in the trait seems pretty ok.

danharrin:
> mind throwing an actual use-case out there

Multiple forms on a Livewire component. Each form is contained within a trait. You can reuse forms between components. Each form has its own validation rules that belong inside the trait.

emil-nasso:
> Makes sense - mind throwing an actual use-case out there? Even if it's pseudo-code?
> 
> I don't see any down-sides to this, you? As long as the actual component takes precedence over rules that exist in the trait seems pretty ok.

I have multiple use-cases but a big one is basically a common behavior for forms. These forms creates/documents "services belonging to customers" and they have a `BelongsToCustomer` trait.

The trait:

* provides a property ("customerId")
* loads some initial data/suggestions of customers to choose from (mount hook for the trait)
* loads information about the customer of the customerid, to show in the form, when the property changes
* provides validation that the customer exists

This is also paired with a template that I simply `@include`  to show the input that has suggestions, autocomplete and that shows information about the customer that you have entered an id for/searched for.

It is a neat way to add as rich "the customer this thing belongs to" field to any form. All you have to do is use the trait and include the template. In the forms `create()` method I can `$this->customerId` and know that I have a validated customerId that I can use to create/document the service or do whatever the form is made for.

emil-nasso:
* Support for rules, custom messages, custom attribute names, and custom attribute values.
* Supports loading all the above from both methods and properties in traits.
* If something is defined in both the component and a trait, the component has precedence, allowing you to "override" the trait in your component.
* More tests.

Would love some feedback on the code, naming, tests etc. :pray:

joshhanley:

emil-nasso:
@joshhanley Would you be interested in this for v3?

joshhanley:
@emil-nasso we're hoping the new nested component structure will reduce the need for traits. But we're not against this idea, so feel free to PR it once V3 has been released 🙂

KSneijders:
@emil-nasso @joshhanley Hey, does either of you know if this got adopted in V3 already (one way or another)?

PR #4933, opened 2022-05-02
Fix wireable order validation

joshhanley:
To validate data, Livewire converts objects (like collections and models) to arrays first, so that way it can be validated.

But if you have a custom collection (say that contains DTOs) that implements the `Wireable` interface, currently it won't validate properly as calling `toArray` on the custom collection won't return the correct data structure.

To fix this, this PR changes the order that `upwrapDataForValidation` checks what data type the object is, nows checks if the object implements the `Wireable` interface first and if so, uses that method.

This also makes it consistent with the dehydration process in `HydratesPublicProperties`, which checks whether an object implements `Wireable` first and uses that method if it does.

https://github.com/livewire/livewire/blob/98b83e96128c30956edc8bd032270b7b7c605f35/src/HydrationMiddleware/HydratePublicProperties.php#L100-L103

To add the test for this, I had to add an example custom collection and custom DTO (mainly as that what how I found this issue). If you can think of a simpler way to test this change, I'd be happy to refine it.

Hope this helps!

### DIFF
--- a/src/ComponentConcerns/ValidatesInput.php
+++ b/src/ComponentConcerns/ValidatesInput.php
@@ -371,8 +371,8 @@ trait ValidatesInput
     protected function unwrapDataForValidation($data)
     {
         return collect($data)->map(function ($value) {
-            if ($value instanceof Collection || $value instanceof EloquentCollection || $value instanceof Model) return $value->toArray();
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks Josh!

PR #4934, opened 2022-05-03
Add ability to exclude querystring using an array of values

masterix21:
Add the ability to filter querystring params by an array of values like so:

```php
protected $queryString = [
    'search' => ['except' => ['black', 'white']],
];
```


joshhanley:
@masterix21 thanks for the PR! Any chance you could add tests for this? Thanks!

masterix21:
Done :)

joshhanley:
While we're developing V3, we're trying to keep V2 stable and limit API changes/additions. In that spirit, PR's like this that may otherwise make sense to merge, we're unfortunately closing. Feel free to PR this once V3 has been released. Thank for your effort and input!

PR #4938, opened 2022-05-03
Do not disable caching of pages that do not use livewire

tabuna:
When using livewire in only a small part of a project, the middleware that is being used prevents the whole project from being cached. For example this will not work:

```php
Route::middleware('cache.headers:public;max_age=2628000;etag')
    ->group(function (){
        Route::get('/', 'HomeController@index')->name('home');
    });
```

This change suggests limiting caching to only routes with the mask `livewire.`

I'm also curious if it's possible to transfer the registration of this middleware from the global level, to the configuration file.

### DIFF
--- a/src/DisableBrowserCache.php
+++ b/src/DisableBrowserCache.php
@@ -18,6 +18,10 @@ class DisableBrowserCache
     {
         $response = $next($request);
 
+        if (!$request->routeIs('livewire.')) {
+            return $response;
+        }
+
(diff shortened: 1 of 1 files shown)


joshhanley:
@tabuna thanks for the PR! When I wrote this feature, I made sure that any routes that don't contain a Livewire component don't have the caching disabled.

The way it is meant to work is `Livewire::shouldDisableBackButtonCache()` will only return true if a Livewire component has initialised, otherwise it returns false, and therefore should skip the no cache headers block.

Are you experiencing something different?

PR #4955, opened 2022-05-12
Add "HMR"/Live reload

haringsrob:
Hey,

This small pr (needs docs update as well) enables the usage of HMR with Laravel livewire.

It relies on Laravel mix/webpack.

The strategy is simple. When you run `npm run mix watch--hot` a `hot` file is created in `public` and also on the webpack js side a new function comes available.

We listen on the websocket, which is only injected with debug mode.

This greatly improves the workflow as you no longer need to refresh and go through steps again in some cases.

Then every time a hmr happens (for example when tailwind reparses), the event is emitted and will refresh all livewire components and keep its state.

Screen Recording 2022-05-12 at 08 52 10

This is of course only active with debug mode on. And will not trigger if webpack is not in hot reload mode.

### DIFF
--- a/src/ComponentConcerns/ReceivesEvents.php
+++ b/src/ComponentConcerns/ReceivesEvents.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\ComponentConcerns;
 
+use Illuminate\Support\Facades\File;
 use Livewire\Event;
 use Livewire\Livewire;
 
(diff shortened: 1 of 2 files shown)


joshhanley:
@haringsrob thanks for the PR! We're going to close this as this has been built into V3 🙂

nikolaimoiseev01:
@joshhanley How can I implement this behavour in Livewire 3? I now only see that with vite I do not need to refresh full page manually (it refreshes automaticaly), but @haringsrob shows livewire force reloads without full page refresh (even automaticaly).

PR #4957, opened 2022-05-12
Fix: Typed property must not be accessed before initialization

alexandre67fr:
`HandlesActions` trait checks if `$this->{$propertyName} instanceof Model`.  
However, prior to this check, we need to make sure that the property is initialized.

Added one browser test, which allows to catch the error and make sure that the MR fixes the issue.  
Couldn't reproduce this issue with unit tests.

More details in discussion https://github.com/livewire/livewire/discussions/4956

### DIFF
--- a/src/ComponentConcerns/HandlesActions.php
+++ b/src/ComponentConcerns/HandlesActions.php
@@ -21,7 +21,7 @@ trait HandlesActions
         $propertyName = $this->beforeFirstDot($name);
 
         throw_if(
-            ($this->{$propertyName} instanceof Model || $this->{$propertyName} instanceof EloquentCollection) && $this->missingRuleFor($name),
+            isset($this->{$propertyName}) && ($this->{$propertyName} instanceof Model || $this->{$propertyName} instanceof EloquentCollection) && $this->missingRuleFor($name),
             new CannotBindToModelDataWithoutValidationRuleException($name, $this::getName())
         );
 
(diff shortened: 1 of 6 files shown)


alexandre67fr:
I'll have a look to fix tests compatibility with PHP <=7.3

joshhanley:
@alexandre67fr yep have a this test to fix the <= 7.3 issue https://github.com/livewire/livewire/blob/612f591fee9a899955e58779a92e2ac9fd07c3eb/tests/Unit/WireablesCanBeSetAsPublicPropertiesTest.php

You can see we are using require_once to only include the class if the version check has passed.

That fixes the issues of typed properties throwing errors in older versions of php.

Hope this helps!

joshhanley:
Whoops, didn't mean to close!

alexandre67fr:
Perfect, thanks for the tip, @joshhanley , I'll have a look at this

alexandre67fr:
Done :)

Without the proposed fix, the test fails with message

```
Error : Typed property Tests\Unit\ComponentWithTypedProperty::$message must not be accessed before initialization
src/ComponentConcerns/HandlesActions.php:24
tests/Unit/TypedPropertiesTest.php:21
```

With the fix applied, this error doesn't occur.

joshhanley:
@alexandre67fr thanks! Looks like GH actions is failing on something unrelated. I will trigger the failed tests to re-run to make sure 🙂

alexandre67fr:
Hi @joshhanley , just letting you know that all tests have passed successfully :)

joshhanley:
@alexandre67fr thanks! Caleb is currently off on paternity leave, so might be a little while before he is back and reviewing PRs 🙂

PR #5043, opened 2022-06-18
Support Shadow DOM

luttje:
This modifies the walker to include shadow DOM. So when Livewire is "reset" (using tearDown and a subsequent initialize) all elements on the page (including those in Shadow DOM) are recognized and their events (e.g: wire:click) are attached.

As we're discussing in #5021 (more information there)


joshhanley:
@luttje thanks for the PR!

GitHub has a 15 day limit for running tests for new contributors and as I didn't trigger them in time, I've had to push an empty commit to get them to run. I have also done the same on your other PR #5044.

joshhanley:

luttje:
Understandable. Will you support shadow dom (as described above) in v3?

joshhanley:
@luttje possibly, that would be up to Caleb to decide. I can't remember whether Alpine already supports shadow dom or not, but as Livewire will be using Alpine, I think it would need to support it too.

PR #5044, opened 2022-06-18
Fix HTMLTemplateElement not updating

luttje:
Currently Livewire does not recognize changes to `<template>` (HTMLTemplateElement) elements. This is in part because it uses `isEqualNode` to compare and two HTMLTemplateElement tags are always the same, regardless of their contents. This is because they contain a DocumentFragment which can be different, but that isn't checked. [jsfiddle demo](https://jsfiddle.net/vryfm2sh/)

The changes in this PR have Livewire consider if a morphing element is an HTMLTemplateElement OR contains one. If that is the case it asks the browser to compare the fromEl to the toEl. It uses innerHTML for that, since that does look into the DocumentFragment and sees the differences between fromEl and toEl. 
If a morphing element is a HTMLTemplateElement it will compare it's content (DocumentFragment).

I discussed this in #5043 where you can find some more information.


joshhanley:

PR #5121, opened 2022-07-22
Fixes bug where null computed property is considered set

aidan-casey:
This pull request fixes a bug where a null return from a computed property is considered set. This is because no check was actually performed on the computed property value, but instead it was just checking if the computed property value could be retrieved.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -274,15 +274,17 @@ abstract class Component
     {
         return $this->forStack;
     }
-    
+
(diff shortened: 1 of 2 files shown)


joshhanley:
@aidan-casey thanks for the PR! I've triggered the tests to run 🙂

aidan-casey:
Thanks, @joshhanley!

PR #5129, opened 2022-07-26
Fix select input option clearing

joshhanley:
This PR fixes an issue with select inputs (see #2992, #4826 and #5081), where if `selected` is manually output on an `<option>` tag, then after the user selects a different option, the selected value disappears from the select input box.

See screen capture of what currently happens:
Recording 2022-07-26 at 16 09 37

Why is this needed? Well as Livewire can be used to progressively enhance sites, it can be desirable to output `selected` on the selected option, so that way if javascript isn't loaded or there is a delay in it being loaded, then the correct value is shown to the user from the beginning and then there is no "blip" once Livewire has loaded.

As far as I'm aware all other inputs work like this, just it is select that breaks.

The option I choose to fix this, is to ensure that all select inputs have the value reset from the Livewire property value after each request.

The reason for this is, due to the nesting nature of `<select>` and `<option>` tags, morphdom actually resets the `selectedIndex` of the select input when it processes the individual option tags. To fix this, morphdom has a `specialElHandlers` for fixing this. But this doesn't run on objects that have `wire:model`.

I could change Livewire's morphdom implementation to allow the select special handler to run (see screenshot), but it doesn't solve the issue for if the `selected` is option on a different option to the one Livewire is tracking with `wire:model` (see incorrect test).

Screen Shot 2022-07-26 at 4 01 11 pm

That's why I've chosen to process the select inputs value from Livewire's properties on each request instead. I also limited it to only single selects for now (not multiple selects as they are fine - as far as I'm aware).

Open to other solutions if there is a better way. It was quite a challenge to work out what morphdom was doing, but I can demonstrate it if needed.

Hope this helps!


calebporzio:
Dude, thank you Josh. As always - amazing work ❤️ ❤️ ❤️

joshhanley:
No worries! 🙂

PR #5173, opened 2022-08-09
Change from nowdoc to heredoc on asset warning output

joshhanley:
Currently the asset warning script `$nonce` variable isn't being evaluated due to nowdoc syntax, as such that breaks turbolinks/turbo.

Currently when it renders, it looks like this:

Screen Shot 2022-08-10 at 8 49 26 am

Which causes this error in turbolinks:

image

So have changed it to heredoc, so when the nonce is empty is doesn't render anything:

Screen Shot 2022-08-10 at 8 49 44 am

Hope this helps!


joshhanley:
No worries! 🙂

PR #5192, opened 2022-08-21
Add assertions for FileDownloaded contents

colinmackinlay:
@joshhanley here's the PR for #4981

This adds:
- an assertion `assertFileDownloadedContains `to check for some specific partial content in the downloaded file.
- an assertion `assertFileDownloadedNotContains `to check for missing specific partial content in the downloaded file.

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -400,4 +400,28 @@ trait MakesAssertions
 
         return $this;
     }
+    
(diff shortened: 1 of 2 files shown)


joshhanley:
@colinmackinlay thanks!

joshhanley:
While we're developing V3, we're trying to keep V2 stable and limit API changes/additions. In that spirit, PR's like this that may otherwise make sense to merge, we're unfortunately closing. Feel free to PR this once V3 has been released. Thank for your effort and input!

PR #5205, opened 2022-08-26
add method to get the preRenderedView

ph7jack:
1️⃣ **Is this something that is wanted/needed? Did you create a discussion about it first?**
Yes, I want to modify the livewire behavior, but I need access to the rendered view in many cases
https://github.com/livewire/livewire/discussions/5190

2️⃣ **Did you create a branch for your fix/feature? (Master branch PR's will be closed)**
Yes

3️⃣ **Does it contain multiple, unrelated changes? Please separate the PRs out.**
No

Yes

5️⃣ **Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.**
It will be really useful if we can get access to the preRenderedView property in the livewire component, I need to use a trait to expose it.

Thanks for contributing! 🙌

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -156,6 +156,11 @@ abstract class Component
             ->toArray();
     }
 
+    public function getPreRenderedView()
+    {
+        return $this->preRenderedView;
+    }
+
     public function skipRender()
     {
         $this->shouldSkipRender = true;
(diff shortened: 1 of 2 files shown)


joshhanley:

PR #5253, opened 2022-09-20
spread attributes for dynamic components

nuernbergerA:
Currently dynamic-components are not that dynamic as you need to hardcode the attributes
```blade
<livewire:dynamic-component component="alert" :type="$type" :message="$message" />
```

Or pass them via a "magic" attribute
```blade
<livewire:dynamic-component :component="$component" :attributes="$attributes" />
```
But then you have to parse them in the `Component` and assigned them yourself:
```php
public function mount(array $attributes) {
     $this->type = $attributes['type'];
     $this->message = $attributes['message'];
}
```

With this PR we will have `wire:spread`
```blade
<livewire:dynamic-component :component="$component" :wire:spread="$attributes" />
```
so we don't have to monkey around with arrays in mount

This will only be supported if you run PHP 8.1.

Below PHP 8.1 nothing will be manipulated, you could retrieve attributes via `mount`
```php
public function mount() {
     $attributes = func_get_args()[0];
     $this->type = $attributes['type'];
     $this->message = $attributes['message'];
}
```

### DIFF
--- a/src/LivewireTagCompiler.php
+++ b/src/LivewireTagCompiler.php
@@ -112,6 +112,15 @@ class LivewireTagCompiler extends ComponentTagCompiler
                                 ? "'{$attribute}' => \Illuminate\View\Compilers\BladeCompiler::sanitizeComponentAttribute({$value})"
                                 : "'{$attribute}' => {$value}";
                 })
+                ->when(PHP_VERSION_ID >= 80100, function($collection) {
+                    return $collection->map(function (string $value, string $attribute) {
+                        if(! str($attribute)->startsWith('wire:spread')) {
+                            return $value;
+                        }
+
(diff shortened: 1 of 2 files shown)


joshhanley:
@nuernbergerA makes sense to me! Any reason it needs to be 8.1 only?

nuernbergerA:
@joshhanley 

> @nuernbergerA makes sense to me! Any reason it needs to be 8.1 only?

From [php-watch](https://php.watch/versions/8.1/spread-operator-string-array-keys)
> Prior to PHP 8.1, the spread operator only supported arrays with numeric keys. Using an array with string key resulted in an error prior to PHP 8.1

joshhanley:
@nuernbergerA ah right yep! It's amazing how we quickly get used to new features 😅

joshhanley:

PR #5369, opened 2022-11-18
Add common OS generated files to .gitignore

jukkakoskinen:
I noticed there's a .DS_Store file in the root of the repo, and thought that ignoring common OS generated files would be a good idea.


joshhanley:
Thanks!

PR #4942, opened 2022-05-06
Fix Safari special character encoding issue

bobwurtz:
Review the contribution guide first at: https://laravel-livewire.com/docs/2.x/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

Yes - there is an existing [pull request](https://github.com/livewire/livewire/pull/4613) that was closed because it did not include a test. I could not contribute to that branch so I copied the fix for the special character issue provided by [jonneroelofs](https://github.com/jonneroelofs) and added a test.

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)

Yes

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No

4️⃣ Does it include tests? (Required)

Yes

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

This would fix a common issue with Safari handling special characters differently than Chrome.

```
            if (
                // The value is a supported type, set it in the data, if not, throw an exception for the user.
                is_bool($value) || is_null($value) || is_array($value) || is_numeric($value)
            ) {
                data_set($response, 'memo.data.'.$key, $value);
            } else if(is_string($value)) {
                data_set($response, 'memo.data.'.$key, \Normalizer::normalize($value));
            }
```

The `is_string` check is moved to a separate if statement and normalized.

This is my first time contributing to an open-source project so please let me know if I did anything wrong or if there is something I should do differently in the future. Thanks!

PR #5379, opened 2022-11-22
Add `StringNormalization` support to arrays and collections to fix Safari checksum issues

Balsakup:
Update `StringNormalization` test to prove that special chars are not normalized and break Safari checksum.  
https://github.com/livewire/livewire/pull/4942 don't fix all cases.

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -98,9 +98,11 @@ class HydratePublicProperties implements HydrationMiddleware
             ) {
                 data_set($response, 'memo.data.'.$key, $value);
             } else if(is_array($value)) {
+                // The data here needs to be normalised, so that Safari handles special charaters properly without throwing a checksum exception.
                 data_set($response, 'memo.data.'.$key, static::normalizeArray($value));
             } else if(is_string($value)) {
                 data_set($response, 'memo.data.'.$key, \Normalizer::normalize($value));
+                // The data here needs to be normalised, so that Safari handles special charaters properly without throwing a checksum exception.
             } else if ($value instanceof Wireable && version_compare(PHP_VERSION, '7.4', '>=')) {
                 $response->memo['dataMeta']['wireables'][] = $key;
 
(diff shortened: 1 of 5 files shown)


Okipa:
@Balsakup Nice 👍🏻

Seems OK to me, what do you think about this @joshhanley?

joshhanley:
@Balsakup thanks for the PR! All looks good for collections/ arrays/ strings.

What about the scenario where there is an eloquent model (or eloquent collection) which has special chars in attributes, does that break too?

Balsakup:
@joshhanley, I added some tests for the Eloquent collections and models, and everything is working as expected

joshhanley:
@Balsakup thanks! I've been running some performance checks on this to make sure we don't increase the processing time too much. Just waiting on Caleb to have a look too before actioning further 🙂

joshhanley:
So we've run some benchmarks on this PR in comparison to the current master branch, and the results are as follows:

As this PR only affects collections and arrays, so that is what has been tested.

Benchmark 1 - array with 10,000 items and also a collection of that array - running 100 iterations of dehydrate method:

Master:
Total time: 3.125s
Average: 31.25ms

This PR:
Total time: 3.848s
Average: 38.48ms


Benchmark 2 - array with 10 nested arrays of 10,000 items (100,000 items total) and also a collection of that array - running 100 iterations of dehydrate method:

Master:
Total time: 2.692s
Average: 26.92s

This PR:
Total time: 12.956s
Average: 129.56ms

As you can see in the benchmark 1, running a combined total of 20,000 items between the array and the collection, the difference between master and this PR is nominal.

There is a noticeable difference in benchmark 2, between master and this PR, approx 5 times slower. But, that is essentially testing a combined total of 200,000 items between the array and the collection and it's still only ~130ms.



Our conclusion is, that those times are not ideal, but that is likely due to the recursion through the arrays/collections. It's also unlikely that someone will have that much data loaded into a single array.

As such, we are going to merge this, to ensure Safari is properly supported and doesn't throw the checksum issue.

There might be a better place to do this processing, like somewhere where we are already traversing the property data, which could make this more efficient. That's something we can look into in the future.

joshhanley:
Thanks for the PR @Balsakup!

Balsakup:
@joshhanley Thank you so much for accepting my contribution. I am grateful for the opportunity to work with you and contribute to the project.

PR #5427, opened 2022-12-09
Add BackedEnum as possible type for property

lina-rodrigues:
Since backed enums can be used as a type since commit 8dd340ea07da5c5eb33afa3fce258046b4a6938a, is good to have it mentioned on PublicPropertyTypeNotAllowedException.

I was misled thinking that enums were not supported since I didn't see enums on the list provided in the exception message, but the problem was just that my enums were not backed.

### DIFF
--- a/src/Exceptions/PublicPropertyTypeNotAllowedException.php
+++ b/src/Exceptions/PublicPropertyTypeNotAllowedException.php
@@ -9,7 +9,7 @@ class PublicPropertyTypeNotAllowedException extends \Exception
     public function __construct($componentName, $key, $value)
     {
         parent::__construct(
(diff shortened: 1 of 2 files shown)


joshhanley:
@sirbaconjr thanks for the PR! Makes sense to me. I don't think that's where the test should live though. I actually don't think this needs a test at all. Any chance you can remove it? Thanks!

joshhanley:
Oh also `Wireable` properties  should also be added too - see https://laravel-livewire.com/docs/2.x/properties#wireable-properties

joshhanley:
@sirbaconjr we're going to hold off on this for now, to see if we can come up with a better solution in V3, due to there being so many more currently supported property types (in V2) that aren't listed in that exception. V3 has a whole new property management system so this exception may not even be needed.

As such we're going to close this, but thanks for being willing to contribute! 🙂

PR #5501, opened 2023-01-09
Bugfix for #5420 - allow multiple `wire:dirty` directives on a single element

samlev:
Original discussion here: https://github.com/livewire/livewire/discussions/5420

This update fixes a bug where only the first `wire:dirty` directive on an element would be processed, which was inconsistent with other directives like `wire:loading`.

Given the following code:

```html
<input wire:model.defer="search">
<button
    class="text-gray-300"
    wire:click="$refresh"
    wire:target="search"
    wire:dirty.class="text-green-900"
    wire:dirty.class.remove="text-gray-300"
>
  Search
</button>
```

Before this change, only `wire:dirty.class="text-green-900"` would be processed leaving the button with `class="text-gray-300 text-green-900"`.

After the change, it will correctly update to `class="text-green-900"`, and revert back to `class="text-gray-300"` when the field returns to a "clean" state.


joshhanley:
@samlev thanks! Failing tests are unrelated. There was a breaking change in Laravel 9.46 which they are keeping, so we need to update our tests to match. Once I've done that, I will merge master into here are re-run the tests.

joshhanley:
@samlev all looking good! Thanks for the PR 🙂

PR #5511, opened 2023-01-13
Add `assertReturned` method to test return value of a method

gdebrauwer:
Currently there is no easy way to assert the return value of method:

```php
{
    public function someMethod()
    {
        return 'foo'; // How to test that 'foo' is returned when this method is called?
    }
}
```
This PR adds an   `assertReturned` method to make that possible.
```php
   ->call('someMethod')
   ->assertReturned('foo'); //  OR  ->assertReturned(fn ($value) => $value === 'foo');
```

### DIFF
--- a/src/Testing/Concerns/MakesAssertions.php
+++ b/src/Testing/Concerns/MakesAssertions.php
@@ -400,4 +400,15 @@ trait MakesAssertions
 
         return $this;
     }
+
+    public function assertReturned($value)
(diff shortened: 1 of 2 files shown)


joshhanley:
@gdebrauwer thanks for the PR! We're going to close this for now as we're feature freezing V2. Please feel free to reopen it after V3 has been fully released 🙂

PR #5525, opened 2023-01-24
Fix multiword deep nesting relations

joshhanley:
This PR fixes an issue #3803 with deep model binding of eloquent models and collections, where any relations, that have multiple words in the relation name, don't function at all currently with Livewire. Text inputs for example show up blank rather than with the data populated.

So if you had a relation like `activeTasks()` on the `$user` model, and tried to bind to it using Livewire like the below, then it wouldn't work at all
```php
public $user;

protected $rules = [
    'user.activeTasks.*.name' => 'required',
];
```

```html
<div>
    @foreach($user->activeTasks as $key => $activeTask)
        <input type="text" wire:model="user.activeTasks.{{ $key }}.name" />
    @endforeach
</div>
```

The reason this wouldn't work, is due to the way the data is processed from the `$user` model when it is compared with the listed `$rules` on the component during dehydration.

For security purposes, Livewire will only send to the front-end any properties, of an Eloquent model or Eloquent collection, that have been "whitelisted" by rules added to the `$rules` array.

To achieve all this, Livewire calls `->toArray()` on the `$user` model to ensure all the data has been parsed and serialised by Laravel, ready to be sent to the front end. 

Then to filter the data from the `$user` model, Livewire takes the returned `$user` data array from `->toArray()` and then compares it with the `$rules` array, and removes anything that isn't whitelisted.

But in the `toArray()` method, Laravel converts all relation names to `snake_case`, so the outputting array would look like:

```php
'user' => [
    'active_tasks' => [
        0 => [
            'title' => 'Active Task 1',
        ],
    ],
];
```

Due to this Livewire can't match the rules. For example, to match the rule in the example at the top, Livewire looks for `$data['user']['activeTasks'][0]['title']`, which doesn't exist. As such Livewire would not send any of the data for `activeTasks` to the front end.

One way this problem could be solved, would be to try and work out which rules have relations in them and `snake_case` the relations in the rules. That would potentially solve the issue for dehydrating the `$user` data to send to the front-end.

But it then causes new issues when rehydrating the `$user` data. When it tries to reapply the serialised data to the `$user` mode, it will try and look for a relation (or attribute as Livewire can't distinguish between them) that is `active_tasks`, but not be able to find it. It will then instead set an attribute called `active_tasks` on the `$user` model and assign the data as an array.

As such, the above scenario would then also require further work to get hydration working correctly.

Instead as a simpler solution, this PR makes use of a static property on Eloquent models called `$snakeAttributes`, which is used for controlling how relations are serialised and is set to `true` by default.

We can memoize the current value of `$snakeAttributes` and set it to `false` before calling `->toArray()` on any eloquent model or collection. This ensures that any relations retain their current casing in the returned data array. Finally we can restore the original value of `$snakeAttributes` .

As this data is only used by Livewire for dehydration and hydration, I believe it shouldn't have any impact on anything else.

Hope this helps!

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -241,7 +241,12 @@ class HydratePublicProperties implements HydrationMiddleware
     }
 
     public static function filterData($instance, $property) {
+        $cacheSnakeAttributes = Model::$snakeAttributes;
+        Model::$snakeAttributes = false;
+
         $data = $instance->$property->toArray();
+
+        Model::$snakeAttributes = $cacheSnakeAttributes;
 
         $rules = $instance->rulesForModel($property)->keys();
 
--- a/tests/Browser/DataBinding/Models/Component.php
+++ b/tests/Browser/DataBinding/Models/Component.php
@@ -14,11 +14,13 @@ class Component extends BaseComponent
         'author.posts.*.title' => '',
         'author.posts.*.comments.*.comment' => '',
         'author.posts.*.comments.*.author.name' => '',
+        'author.posts.*.otherComments.*.comment' => '',
+        'author.posts.*.otherComments.*.author.name' => '',
     ];
 
     public function mount()
     {
-        $this->author = Author::with(['posts', 'posts.comments', 'posts.comments.author'])->first();
+        $this->author = Author::with(['posts', 'posts.comments', 'posts.comments.author', 'posts.otherComments', 'posts.otherComments.author'])->first();
     }
 
     public function save()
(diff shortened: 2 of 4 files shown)


calebporzio:
Great writeup and fantastic solution! Thanks Josh!

joshhanley:
No worries! 😁

AlexisSerneels:
This PR produced side effects as mentioned here: https://github.com/filamentphp/filament/issues/5568

joshhanley:
@AlexisSerneels thanks! There is an open PR #5536 to fix the issue. I will have a look at it in detail tomorrow.

joshhanley:
We've decided to revert this PR as currently there are too many edge cases. See my comment on PR https://github.com/livewire/livewire/pull/5536#issuecomment-1407813069 for more details on why it was reverted.

PR #5536, opened 2023-01-28
Fix snake case attributes when using the Attribute casting

realslimsutton:
Issue is described in #5535 and was introduced in #5525.

This PR had an unintentional knock-on effect with snake case attributes that are appended to the model json when serializing. [This behaviour is documented in the latest version of Laravel](https://laravel.com/docs/9.x/eloquent-serialization#appending-values-to-json).

I've proposed a solution that caches the model attributes before disabling snake case attributes. #5525 disables snake case so the relations can be properly accessed by Livewire, this aims to keep the fixes introduced in that PR, whilst also keeping support for snake case attributes.

Additionally, I have added a test for this, but since snake case attributes were introduced in Laravel 8+ they're only added if the `Attribute` class exists, though I have also added the fallback methods in older versions.

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -241,6 +241,8 @@ class HydratePublicProperties implements HydrationMiddleware
     }
 
     public static function filterData($instance, $property) {
+        Model::cacheMutatedAttributes(get_class($instance->$property));
+
         $cacheSnakeAttributes = Model::$snakeAttributes;
         Model::$snakeAttributes = false;
 
--- a/tests/Browser/DataBinding/Models/Test.php
+++ b/tests/Browser/DataBinding/Models/Test.php
@@ -152,7 +152,7 @@ if(class_exists(Attribute::class)) {
             return $this->hasMany(OtherComment::class);
         }
 
-        public function firstName(): Attribute
+        public function firstName()
         {
             return Attribute::make(
                 function() {
@@ -161,7 +161,7 @@ if(class_exists(Attribute::class)) {
             );
         }
 
-        public function lastName(): Attribute
+        public function lastName()
         {
             return Attribute::make(
                 function() {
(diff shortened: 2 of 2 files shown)


joshhanley:
@andreascreten thanks for reporting and submitting a fix so quick! I’ve triggered the tests to run. I will have a look at the PR in detail tomorrow 🙂

realslimsutton:
> @andreascreten thanks for reporting and submitting a fix so quick! I’ve triggered the tests to run. I will have a look at the PR in detail tomorrow 🙂

You’re welcome 😊 I’m not too familiar with Dusk though I believe the failed tests are unrelated to this change.

joshhanley:
@realslimsutton we've decided to revert the original PR #5525 that caused this bug for now.

I was testing your PR this morning and the `cacheMutatedAttributes()` was a great solution when dealing with a single model. There were some issues with it when dealing with Eloquent Collections or deep nesting of models.

The problem when dealing with an Eloquent Collections is that it doesn't have a `cacheMutatedAttributes()` method. So for us to successfully get this running with Eloquent Collections, we'd have to iterate through the collection and ensure the collection contains models all of the same class, then we could call `cacheMutatedAttributes()`. If this were the only issue, we probably could have got it running.

The bigger issue though is when dealing with deep nesting of models, which is the scenario that the original PR #5525 was working within. Due to the nature of deep nesting, we would have to ensure that whenever a model's relations are serialised, we would need to make sure each level of relation has `cacheMutatedAttributes()` called before the `$snakeAttributes` property is changed.

But currently, we're using Laravel's `toArray()` method when serialising a model, which in turn iterates though all the relations and calls `toArray()` on them. As such we have no way to interject into that process to call `cacheMutatedAttributes` before `toArray()` is called on each relation.

There are ways we could work around that by iterating through the tree of relations and calling `cacheMutatedAttributes()` before `toArray()` is called on the base model, but that could potentially cause performance issues.

Due to the above and a number of other edge cases that would need to be worked through, that is why we decided to revert the original PR. We'll probably not try to fix this now in V2, so we can continue to focus our efforts on building V3 and hopefully we can provide a full fix in V3.

As such, we're going to close your PR as it's no longer needed.

Thank you again for being willing to contribute and trying to address the issue so quickly, it's much appreciated. 🙂

PR #5567, opened 2023-02-20
Support anonymous class component registration

calebporzio:
This PR makes the following possible:

```php
// In something like a service provider:

Livewire::component('foo', new class extends \Livewire\Component {
    // ....
});
```

Now you can use the component as normal:

```html
<livewire:foo />
```


elmarzouguidev:
Amazing thanks @calebporzio

PR #5569, opened 2023-02-20
Query string sorting

ariaieboy:
As described here https://github.com/livewire/livewire/discussions/2722 for SEO reasons we must force a sorting for the query params. 

I've introduced a `sort` option for the QueryParams that will set the order of the queryParams.

here is an example:
```
    protected $queryString = [
        'page' => ['sort' => 2],
        'search' => ['sort' => 1],
    ];
```

The only problem I have is that in the initial request, it's not updating the URL to the correct order but when you update the value of one queryParams it will enforce the order based on the sort option.

I am not that familiar with browser testing and it might work in the initial request too but I couldn't write a test that passes on initial request

### DIFF
--- a/src/Features/SupportBrowserHistory.php
+++ b/src/Features/SupportBrowserHistory.php
@@ -161,15 +161,19 @@ class SupportBrowserHistory
         }
 
         $excepts = $this->getExceptsFromComponent($component);
-
+        $sorts = $this->getSortsFromComponent($component);
         $this->mergedQueryParamsFromDehydratedComponents = collect(request()->query())
             ->merge($this->mergedQueryParamsFromDehydratedComponents)
             ->merge($this->getQueryParamsFromComponentProperties($component))
             ->reject(function ($value, $key) use ($excepts) {
                 return isset($excepts[$key]) && $excepts[$key] === $value;
             })
+            ->map(function ($property,$key) use ($sorts){
+                return ['value'=>$property,'sort'=>$sorts[$key]??0];
+            })
+            ->sortBy('sort')
             ->map(function ($property) {
-                return is_bool($property) ? json_encode($property) : $property;
+                return is_bool($property['value']) ? json_encode($property['value']) : $property['value'];
             });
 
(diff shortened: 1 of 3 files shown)


joshhanley:
@ariaieboy thanks for the PR!

While I agree that this is something that is needed, I'm not sure if it will be accepted at the moment due to how close V3 is, and the fact that you'd submitted a similar PR #4253 in the past that was closed for the same reason.

Can you provide some insight as to why you've resubmitted it now?

ariaieboy:
> @ariaieboy thanks for the PR!
> 
> While I agree that this is something that is needed, I'm not sure if it will be accepted at the moment due to how close V3 is, and the fact that you'd submitted a similar PR #4253 in the past that was closed for the same reason.
> 
> Can you provide some insight as to why you've resubmitted it now?

I've just totally forgotten about that old PR of mine.

the main reason I opened this PR again is that I need this feature in one of my projects.

And since these changes are backward compatible we can have this feature in both versions 2 and 3

ariaieboy:
In the previous PR, I introduced a new config option that forces all query strings to be sorted alphabetically, in this new PR, users have more control over the orders of the query params using the `sort` option since we already have `expect` and `as` in the query params configuration I think this approach is closer to the current API of the query string in livewire.

joshhanley:
@ariaieboy yeah fair enough and thanks!

YasinAshorian (approved):

lrljoe:
As a suggestion, perhaps add the capability to order the components as well?

For example:
* You have three components on a page.  
* Each of these components has one or more items that interacts with the Query String. 
* Each component has orders for the query string items (let's say ordered 1-4)

When you trigger any of the items, the order will become (let's say confused).

What I think would be useful is something like this:
Component (ORDER 1) - 3 items (ordered)
Component (ORDER 2) - 6 items (unordered)
Component (ORDER 3) - 3 items (ordered)

This will allow the query string to be ordered on a per-component basis, and will not reorder itself.

PR #5589, opened 2023-02-28
Add locale prefix

joshhanley:
This PR fixes an issue raised in discussion #5526, relating to Livewire's persistent middleware route matching not being able to match routes with localised paths.

To see more context on this fix, see PR #5577, which is a previous attempt to fix this issue.

This PR adds a localisation prefix to Livewire's endpoint to ensure that any apps that make use of localisation can load the correct local, therefore load correctly translated route paths.

I had to ensure that if an initial request wasn't using a locale, that Livewire's endpoint also didn't use a locale, as some localisation packages have an option to "hide" default locales. https://github.com/mcamara/laravel-localization#redirect-middleware

I've managed to add some tests to ensure that Livewire's endpoint works with and without the locale prefix.

I also have a demo repo, that makes use of [mcamara/laravel-localization](https://github.com/mcamara/laravel-localization), which contains a bunch of dusk tests so we can ensure everything does work as expected. It makes use of both default locales (with and without locale prefix) and configured locales, all with translated route paths. Demo repo can be found here [joshhanley/livewire-localisation-demo](https://github.com/joshhanley/livewire-localisation-demo).

Hope this helps!


calebporzio:
Looks great, thanks Josh!

PR #5592, opened 2023-03-01
fix HydrationMiddleware\HydratePublicProperties: Connection is used f…

psprokofiev:
Connection is used forced from the model. 
`$serialized['connection']` is inherited incorrectly when the included widget uses model with the different connection.

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -7,6 +7,7 @@ use Illuminate\Queue\SerializesAndRestoresModelIdentifiers;
 use Illuminate\Contracts\Queue\QueueableCollection;
 use Illuminate\Contracts\Database\ModelIdentifier;
 use Illuminate\Support\Carbon as IlluminateCarbon;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Contracts\Queue\QueueableEntity;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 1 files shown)


joshhanley:
@psprokofiev thanks for the PR! Could you please add some tests for this and explain in more detail (with code examples) what issue this solves? As I don't follow exactly what this solves. Thanks!

joshhanley:
@psprokofiev Livewire uses Laravel's queue serialisation under the hood. So it will use whichever connection is currently on the model instance, not the model class.

If your model has a different connection than the default on the class, then that's why it will be showing up differently.

You can check this by doing a `dump()` of your model and seeing what connection it is using.

It seems like your model is being loaded on the `production` connection, so you will need to debug why that is happening.

Hope this helps!

psprokofiev:
Bytheway, do you find it secure to pass the connection name via xhr request?
Why we can't take the connection name of the model instance at backend?

joshhanley:
@psprokofiev I personally don't see an issue with it, but other people may disagree.

The reason we can't just use the connection name from the class on the back end is, models can be loaded on the fly using a different database connection (typically used in a multi-database multi-tenant app, like `User::connection('some-other-database')->first();`. So if we don't have that connection information and Livewire tries to restore the model from the default database listed on the `User` class, then it will error as it won't be found.

psprokofiev:
@joshhanley hi there, have a look at the repository
https://github.com/psprokofiev/lw-test
If you run it and press the button you will catch the exception.

To fix it just uncomment this line
https://github.com/psprokofiev/lw-test/blob/main/app/Models/ModelA.php#L12

There is a bug with default connection for models.
If a model without connection attribute has relative relationships with another model which has connection attribute, we get an error.

So the best way to avoid this exception without filling all the project models is not to define the connection from frontend, I've told about that above )

PR #5601, opened 2023-03-03
prevent error when assets are out of date

tanthammar:
Fix: When published Livewire assets are out of date, the warn message causes an error instead of displaying the message.


joshhanley:
@tanthammar thanks! It seems php is evaluating the `\n` hence the error. Can I suggest just escaping it `\\n` instead of changing it to a `.`?

PR #5610, opened 2023-03-10
Adding assertion to test that an event was emitted to self

titantwentyone:
As per title and discussion #5455. Simple assertion to `assertEmittedSelf($value, ...$params)`.

As a related issue/request/offer to PR, I would like to add `assertNotEmittedSelf` as well as other "negatives" for "to" and "up". We have `assertNotEmitted` so the others would be handy also. There may be some instances where these may be useful.


joshhanley:
@titantwentyone thanks for the PR! Looks good to me.

PR #5662, opened 2023-04-20
maintain cookies on subsequent requests

AidasK:
Full description here: https://github.com/livewire/livewire/discussions/5660


edalzell:
Love this one!

PR #5678, opened 2023-05-03
Moved file cleanup into a separate job.

Boorinio:
**The problem**
We are using livewire in a high traffic site with GCS as a temp folder for uploads.  Currently the clearing of files older than 24 hours is being done in sync which causes file uploads to timeout, even by just going through the tmp folder.

**Solution**
Added a unique job that gets dispatched ~~after 10 minutes (Just to group cleanups)~~*  for systems that don't use S3 **or** GCS. I also added a command to dispatch that job manually if the gcs users choose to do so (or any other filesystem for that matter).

I hope this helps, thanks for the hard work, hoping for a quick response 😅

*(Forgot laravel 7 doesn't provide unique jobs)

Thanks for contributing! 🙌

### DIFF
--- a/src/Commands/CleanupCommand.php
+++ b/src/Commands/CleanupCommand.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Livewire\Commands;
+
+use Illuminate\Console\Command;
+use Livewire\FileUploadConfiguration;
+use Livewire\Jobs\CleanupUploadedFilesJob;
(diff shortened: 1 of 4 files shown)


Boorinio:
@joshhanley @calebporzio  is there a chance of this getting merged anytime soon? Otherwise we are going to have to fork the repo 😓

PR #5682, opened 2023-05-05
Ability to encrypt temp files in S3

cby016:

Thanks for contributing! 🙌

### DIFF
--- a/src/GenerateSignedUploadUrl.php
+++ b/src/GenerateSignedUploadUrl.php
@@ -2,6 +2,7 @@
 
 namespace Livewire;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\URL;
 use League\Flysystem\Cached\CachedAdapter;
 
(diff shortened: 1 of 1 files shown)


cby016:
Hi @calebporzio  I tried to submit a PR that will auto encrypt temp files in S3 if specified in the filesystem.php config file.  Looks like the deployment has failed. I tried to link my GitHub account to Vercel to see what's going on but I get a 404.  I feel like this is a useful feature and would save people from banging their head against the wall when their S3 buckets are setup to only accept encrypted objects.

Boorinio:
> Hi @calebporzio  I tried to submit a PR that will auto encrypt temp files in S3 if specified in the filesystem.php config file.  Looks like the deployment has failed. I tried to link my GitHub account to Vercel to see what's going on but I get a 404.  I feel like this is a useful feature and would save people from banging their head against the wall when their S3 buckets are setup to only accept encrypted objects.  

I think it also fails on master, so it's probably not caused by your changes

PR #5742, opened 2023-06-22
[POC] Encrypt dataMeta model

francoism90:

https://github.com/livewire/livewire/discussions/2627

Yes

No

Not yet, I have some questions + it's a POC.

I really dislike Livewire exposing database connection details. From what I understand  `HydratePublicProperties.php` is responsible for creating the meta bag (including it's hydration and dehydration).

So why not encrypt the models data by default, and return a decrypted form when hydrating the model? :)
I simply added the Crypt classes as POC, and only for one model, but I was just wondering if this could lead to problems or any other issues.

From what I've tested, it works perfectly fine, but I'm not a Livewire expert.

If you could please take a look, and give a bit more inside information, I would be really happy.

Thanks!

### DIFF
--- a/src/HydrationMiddleware/HydratePublicProperties.php
+++ b/src/HydrationMiddleware/HydratePublicProperties.php
@@ -20,6 +20,7 @@ use DateTimeImmutable;
 use Carbon\Carbon;
 use DateTime;
 use DateTimeInterface;
+use Illuminate\Support\Facades\Crypt;
 use stdClass;
 use Normalizer;
 
@@ -147,6 +148,8 @@ class HydratePublicProperties implements HydrationMiddleware
 
     protected static function hydrateModel($serialized, $property, $request, $instance)
     {
+        $serialized = Crypt::decrypt($serialized);
+        
         if (isset($serialized['id'])) {
             $model = (new static)->getRestoredPropertyValue(
                 new ModelIdentifier($serialized['class'], $serialized['id'], $serialized['relations'], $serialized['connection'])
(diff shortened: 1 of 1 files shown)


francoism90:
Screenshot from 2023-06-22 09-44-57

No worries, local test data.

francoism90:
I have created a package https://github.com/foxws/livewire-encrypt

Use at your own risk.

RenoLooijmans:
Leaving a message here so I can track this pull. I am interested in using model binding, but not when it exposes the model to the end user. I don't feel comfortable with that.

Would love other's opinions about this pull as well.

ju5t:
This may or may not change in v3 (I honestly wouldn't know). Model binding is probably discouraged in v3.

francoism90:
@ju5t Could you give more details? How would it work without Model binding?

ju5t:
It doesn't really fit this PR (which I like!), but there's a discussion about removing model binding here: https://github.com/livewire/livewire/discussions/5586.

Model binding is insecure in Livewire v2. Encryption doesn't solve that. I don't know how v3 will look like so I can't tell you what would be the best practice there. Here's what is probable: v3 will likely include an attribute or `@locked` as a doc block to designate properties that cannot be changed. I would use that to store the id of my model. All other properties (just the ones I need) would be in a simple array.

francoism90:
@ju5t Yeah, I think the `@locked` docblock will help solve this issue. The problem is that someone can swap the model ID using JS, it doesn't really matter if it is a model binding or array, both can be changed (if I'm not mistaken).

If you like the PR, I could expand this to `models` as well, I did it here: https://github.com/foxws/livewire-encrypt/blob/main/src/Middlewares/HydratePublicProperties.php

So far it seems to work fine, and I'm glad it doesn't expose the database record anymore. :)

PR #5777, opened 2023-07-20
[v3] Add failing test for navigate-once

PhiloNL:
I wasn't 100% sure if this is a bug or expected behavior, but if you add `navigate-once` it will only execute when you visit the page directly or do a hard refresh. This corresponds with the docs "Livewire will only run it on the initial page visit" but not sure if with an initial page visit, a direct visit is meant or whenever you visit a page for the first time (going from page 1 to page 2 and visiting page 2 for the first time). If it's the latter, it's not working as expected. https://github.com/livewire/next/pull/103

For example:

```blade
<a href="page-1" wire:navigate>Page 1</a>
<a href="page-2" wire:navigate>Page 2</a>

<!-- Page 1 -->
No scripts

<!-- Page 2 -->
<script data-navigate-once>
    console.log('Runs only on page one')
</script>
```

If you visit **page-1** and click on **page-2** the script tag is never run.
If you do a hard reload on page-2 or visit page-2 directly from the browser the script will run once.


coleshirley:
To add some of my findings to the original issue it seems that the issue is in `js/features/supportNavigate.js`
```js
import { on, trigger } from "@/events"

let isNavigating = false

window.addEventListener('alpine:navigated', () => {
    isNavigating = true

    // Forward a "livewire" version of the Alpine event...
    window.dispatchEvent(new CustomEvent('livewire:navigated', { bubbles: true }))
})

export function shouldRedirectUsingNavigateOr(effects, url, or) {
    let forceNavigate = effects.redirectUsingNavigate

    if (forceNavigate || isNavigating) {
        Alpine.navigate(url)
    } else {
        or()
    }
}
```
the `alpine:navigated` event is dispatched once on page load and `isNavigating` is set to true and then never is set back to false. Thus all redirects from livewire are passed to `Alpine.navigate()`

calebporzio:
Mmmm you're right, you would assume it will run "once" per session even if it appears late

calebporzio:
Fixed! Thanks @PhiloNL

PR #5778, opened 2023-07-20
[v3] Add failing test for resetting properties on form object

PhiloNL:
When you try to reset a value on a form object like `$this->reset('form.title', 'form.content');` it will thrown an `PropertyNotFoundException: Property [$form.title] not found on component` https://github.com/livewire/next/pull/107


devajmeireles:
@PhiloNL, are you able to reset properties in V3 beta 2 seeing the client side changes visually?

PhiloNL:
> @PhiloNL, are you able to reset properties in V3 beta 2 seeing the client side changes visually?

Yes, but not when using form objects.

calebporzio:
Fixed

PR #5780, opened 2023-07-20
[v3] Add component lazy loading to full page components

PhiloNL:
As discussed, this PR contains a few different approaches how to control lazy loading on full page components.

You can enable or disable it via routes

```php
Route::get('/users', \App\Livewire\Users::class)->lazyLoad();
```

You can enable lazy loading by including the `use WithLazyLoading;` trait:

```php
<?php

namespace App\Livewire;

use Livewire\Component;
use Livewire\WithLazyLoading;

class Revenue extends Component
{
    use WithLazyLoading;

    // ...
}
```

You can also enable lazy loading by using the `Lazy` attribute:

```php
<?php

namespace App\Livewire;

use Livewire\Attributes\Lazy;
use Livewire\Component;

#[Lazy]
class Revenue extends Component
{
    // ...
}
```

Users can opt out of the default lazy load behavior as follows:

```php
// Via route
Route::get('/users', \App\Livewire\Users::class)->lazyLoad(enabled: false);

// Via attribute
<livewire:revenue :lazy="false" />
```

I've added a default placeholder that can be overwritten via either the config or by publishing the view:

```php
'lazy_loading_placeholder' => 'livewire::lazy-load-placeholder',
```

```bash
php artisan livewire:publish --lazy-loading
```

Happy to make any changes based on what you think is the best way forward  😄 

Screen Capture on 2023-07-16 at 22-21-46


devajmeireles:
@PhiloNL ,

Amazingggggg, thank you for that. I've some thoughts about:

1. If `->lazyLoad();` will enable lazy loading on the full page component, why do we need to provide a parameter to disable it? Wouldn't disabling just avoid using `->lazyLoad();` in the route?
2. Following the new style of attributes from Livewire V3 I disagree with the idea of providing a trait. The `#[Lazy]` would be enough and very elegant approach, but I think we need to express the full name: `#[LazyLoad]` instead of `#[Lazy]`

calebporzio:
I'm focusing on bug fixes at the moment, but I think this is a good idea.

I also think you're floating jellyfish placeholder is awesome and hillarious

I agree, we only need an attribute so there's not too many ways to do the same thing

My input is that `#[Lazy]` is shorter but better personally (than LazyLoad)

Also `Route::get()->lazy()` is better than `Route::get()->lazyLoad()` imo

Like I said, not ready to merge but I like it and will return to it after fixing more bugs

calebporzio:
Thanks everyone!

PR #5781, opened 2023-07-20
[v3] Fix upgrade command bug that would prevent it from moving to the next step 

PhiloNL:
Found a small little bug that would stop the command after the introduction if it didn't detect the old namespace.


babacarcissedia:
@PhiloNL  would one additional test help catch that bug ?

PhiloNL:
> @PhiloNL would one additional test help catch that bug ?

The upgrade command doesn't have any tests at this time. I'm still in the progress of adding a few more changes and automation. Given the nature of the command, I'm not sure if it's worth adding tests; maybe when v3 approaches stable.

joshhanley:
@PhiloNL thanks!

PR #5787, opened 2023-07-20
[v3] Add failing redirect browser test

PhiloNL:
I've added a failing test for: https://github.com/livewire/livewire/discussions/5786


coleshirley:
To add some of my findings to the original issue it seems that the issue is in `js/features/supportNavigate.js`
```js
import { on, trigger } from "@/events"

let isNavigating = false

window.addEventListener('alpine:navigated', () => {
    isNavigating = true

    // Forward a "livewire" version of the Alpine event...
    window.dispatchEvent(new CustomEvent('livewire:navigated', { bubbles: true }))
})

export function shouldRedirectUsingNavigateOr(effects, url, or) {
    let forceNavigate = effects.redirectUsingNavigate

    if (forceNavigate || isNavigating) {
        Alpine.navigate(url)
    } else {
        or()
    }
}
```
the `alpine:navigated` event is dispatched once on page load and `isNavigating` is set to true and then never is set back to false. Thus all redirects from livewire are passed to `Alpine.navigate()`

coleshirley:
Further findings:
the line in `dist/livewire.esm.js` and `dist/livewire.js` that dispatches the `alpine:navigated` event is
```js
setTimeout(() => {
   fireEventForOtherLibariesToHookInto();
});
```
which does not appear to be in the livewirejs or alpinejs source anywhere. So it may just be that a recompile of the js assets will fix the issue. 

setting isNavigating back to false somewhere may be necessary as well though

killjin:
image
---
PHP:
```php
$this->redirect($url);
```
This default use Alpine.navigate?

joshhanley:
@PhiloNL thanks! We also had another existing test that was failing due to this bug. See PR #5859 for details on why it is happening.

joshhanley:
@PhiloNL FYI test was failing because you need to add `->waitForLivewire()` before any server call to ensure it waits for the Livewire request to finish before progressing with the test 🙂

joshhanley:
Thanks!

PhiloNL:
> @PhiloNL FYI test was failing because you need to add `->waitForLivewire()` before any server call to ensure it waits for the Livewire request to finish before progressing with the test 🙂

Thanks! Good to know 😄

joshhanley:
No worries! 🙂

PR #5788, opened 2023-07-20
[v3] Exclude files from packagist

nuernbergerA:
Simulate what will end up on packagist
```shell
git archive --worktree-attributes -o latest.zip HEAD
```

Currently its:
image

with this PR:
image


babacarcissedia:
this is pretty cool

joshhanley:
@nuernbergerA looks good to me. Will leave this for Caleb to have a look at.

joshhanley:
I believe this should fix #5808 too

nuernbergerA:
due too the new pattern that test life next to the code in `src` we need to double check if all required files are included

joshhanley:
@nuernbergerA yep will do that now before I merge

joshhanley:
Thanks!

PR #5790, opened 2023-07-20
[v3] Fix tests

nuernbergerA:
Fix failing test due to L9/10 validation changes
Fixed also the turolinks test
and a bug with the legacy support for `queryString`


joshhanley:
Had a look at the last failing test `LegacyTests\Browser\Redirects\Test::it_correctly_shows_flash_messages_before_and_after_direct` test and the reason it is failing is because it's not doing a full page refresh.

Navigate is taking control and redirecting it internally for some reason. The `alpine:navigated` event is firing, so `isNavigating` is being set to `true` in the below snippet, hence navigate is handling the redirect. Not sure yet why that event is being fired.

https://github.com/livewire/livewire/blob/master/js/features/supportNavigate.js#L5-L6

joshhanley:
Looking at the stack trace, it seems there is a `setTimeout(() => fireEventForOtherLibariesToHookInto())` that is being run when the navigate plugin is being initialised. But I can't find that code anywhere.

image

nuernbergerA:
@joshhanley i assume the dist js files are build from some local source, because history & navigate are not published on npm

image

joshhanley:
Thanks!

PR #5823, opened 2023-07-21
[v3] Upgrade command improvements

PhiloNL:
When choosing to migrate to the new namespace:
- It will update your `vite.config.js` if it contains `App/Http/Livewire`
- It will now update any namespace references, so your routes, tests, etc will be updated as well.


joshhanley:
Thanks!

PR #5858, opened 2023-07-21
[V3] Add assertions to assert that actions and properties are bound or not.

julesgraus:

Yes. Please have a look at it. https://github.com/livewire/livewire/discussions/5599

Yes, first for v2 a couple of months ago. But that was closed because v3 was upcoming. Since there isn't a 3.x branch yet, I am forced to create this pull request for the mater branch.

No

Yes

I've noticed that even tough I test as much as I possible can using Livewire::test, the component still not always works like expected when using it in the browser. Mainly because I did forget to add wire:model statements or use non existing public properties or actions in them.

This pull request adds some assertions to assert if ```actions``` and / or ```properties``` are bound or not.

This is how you would assert that a property is bound:

```
Livewire::test(CreatePost::class)
    ->set('title', 'foo')
    ->assertPropertyBound('title')
```

And this is how you would assert that an action is bound:
```
Livewire::test(CreatePost::class)
    ->assertActionBound('title')
```

Please have a look at the included tests in ```src/Features/SupportTesting/Tests/TestableLivewireCanAssertViewBindingsTest.php``` for more of these methods.

### DIFF
--- a/src/Features/SupportTesting/MakesAssertionsOnView.php
+++ b/src/Features/SupportTesting/MakesAssertionsOnView.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Livewire\Features\SupportTesting;
+
+use Illuminate\View\View;
+use PHPUnit\Framework\Assert as PHPUnit;
+
+trait MakesAssertionsOnView
+{
+    /**
+     * @param string $propertyName The name of the property on the Livewire component that is expected to be bound in the view/
+     * @param int|null $times If null, then the property is checked to be bound at least once. Else, exactly that amount.
+     * @return Testable|MakesAssertionsOnView
+     */
+    public function assertPropertyBound(string $propertyName, int $times = null): self
+    {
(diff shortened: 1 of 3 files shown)


joshhanley:
@julesgraus thanks for the PR! I've trigged the tests to run. Just be aware, at the moment we're focusing on fixing bugs, so a new feature like this may have to wait a bit. But we will get to it as soon as we can 🙂

julesgraus:
Thank you for everything.

calebporzio:
Thank you for the contribution @julesgraus and like Josh said, we're more focused on bug fixing at the moment.

I do see how this can be useful, but I'm not sure about the name, and implementation as it might not cover all cases.

Because I don't see this as an easy merge, I'm going to close this and instead recommend that you add a macro for your own applications using something like this in your AppServiceProvider:

```php
Livewire\Features\SupportTesting\Testable::macro('assertX', function (...) {
    //
});
```

This way, you can use this within your apps just the same as if we provided it out of the box.

All the best!

PR #5865, opened 2023-07-22
remove word be from upgrade guide documentation page

MuhammadSadeeq:
The text on upgrade guide is "After [installing Livewire version 3](https://livewire.laravel.com/docs/upgrading#update-livewire-to-version-3), run the following command and you will **be** receive prompts to upgrade each breaking change automatically:". Here the word "be" should be removed. Or it will also be correct if we change it to "you will be receiving prompts...".


joshhanley:
Thanks!

PR #5869, opened 2023-07-22
docs: remove unnecessary .prevent modifier

markjaquith:
`wire:submit` does `.prevent` by default now, so we can omit it from this place in the docs.


joshhanley:
Thanks!

PR #5872, opened 2023-07-22
[v3] Fix Dusk test usage in user land

nuernbergerA:
`DuskTestable` is broken in user land

moved the `Mixin` to Livewire `namespace` so its available
its now in the sam directory as `DuskTestable` where its actually used, this also fits more the new spirit to couple code

----
Fixes https://github.com/livewire/livewire/discussions/5808


joshhanley:
Thanks!

PR #5874, opened 2023-07-22
[v3] Initial implementation loading indicator config + failing test

PhiloNL:
The `livewire.navigate.show_progress_bar` config doesn't do anything at the moment.
I've added the config to the script config. I've also added a failing test.
The reason it's failing is because of the config that determines if it should be shown is hardcoded in the `navigate` plugin, here to be exact: https://github.com/alpinejs/alpine/blob/main/packages/navigate/src/index.js#L13

This should do the trick, but I'm not sure if you want Livewire-related code leaking into Alpine:

```
let showProgressBar = !!document.querySelector('[data-progress-bar]') ||
  (window.livewireScriptConfig && window.livewireScriptConfig.progressBar !== undefined
    ? window.livewireScriptConfig.progressBar
    : true);
```

Maybe it's a bit cleaner if it can be set in the lifecycle hook when registering the navigate plugin:
https://github.com/livewire/livewire/blob/master/js/lifecycle.js#L25


calebporzio:
Done. Thanks!

PR #5876, opened 2023-07-22
[v3] Repeatable `#[Rule]`

nuernbergerA:
maybe `array_merge_recursive` isn't the best solution but for now it works.

```php
// Style A
#[Rule(['required', 'min:2', 'max:3'])]
public $foo = '';

// Style B
#[Rule('required')]
#[Rule('min:2')]
#[Rule('max:3')]
public $foo = '';

// Style C
#[
    Rule('required'),
    Rule('min:2'),
    Rule('max:3'),
]
public $foo = '';
```

what styles should be documented?

---
Fixes https://github.com/livewire/livewire/discussions/5873

### DIFF
--- a/src/Features/SupportValidation/HandlesValidation.php
+++ b/src/Features/SupportValidation/HandlesValidation.php
@@ -23,17 +23,17 @@ trait HandlesValidation
 
     public function addRulesFromOutside($rules)
     {
-        $this->rulesFromOutside = array_merge($this->rulesFromOutside, $rules);
+        $this->rulesFromOutside = array_merge_recursive($this->rulesFromOutside, $rules);
     }
 
     public function addMessagesFromOutside($messages)
     {
-        $this->messagesFromOutside = array_merge($this->messagesFromOutside, $messages);
+        $this->messagesFromOutside = array_merge_recursive($this->messagesFromOutside, $messages);
     }
 
     public function addValidationAttributesFromOutside($validationAttributes)
     {
-        $this->validationAttributesFromOutside = array_merge($this->validationAttributesFromOutside, $validationAttributes);
+        $this->validationAttributesFromOutside = array_merge_recursive($this->validationAttributesFromOutside, $validationAttributes);
     }
 
     public function getErrorBag()
--- a/src/Features/SupportValidation/Rule.php
+++ b/src/Features/SupportValidation/Rule.php
@@ -2,11 +2,12 @@
 
 namespace Livewire\Features\SupportValidation;
 
+use Attribute;
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
 
 use function Livewire\wrap;
 
-#[\Attribute]
+#[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_ALL)]
 class Rule extends LivewireAttribute
 {
     // @todo: support custom messages...
--- a/src/Features/SupportValidation/UnitTest.php
+++ b/src/Features/SupportValidation/UnitTest.php
@@ -136,7 +136,7 @@ class UnitTest extends \Tests\TestCase
     }
 
     /** @test */
-    public function rule_attributes_can_contain_multiple_rules()
+    public function rule_attributes_can_contain_rules_for_multiple_properties()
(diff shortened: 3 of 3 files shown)


vlados:
I think Style A is most common used and for me make most sense

devajmeireles:
> maybe `array_merge_recursive` isn't the best solution but for now it works.
> 
> ```
> // Style A
> #[Rule(['required', 'min:2', 'max:3'])]
> public $foo = '';
> 
> // Style B
> #[Rule('required')]
> #[Rule('min:2')]
> #[Rule('max:3')]
> public $foo = '';
> 
> // Style C
> #[
>     Rule('required'),
>     Rule('min:2'),
>     Rule('max:3'),
> ]
> public $foo = '';
> ```
> 
> what styles should be documented?
> 
> Fixes #5873

As I understand it, the problem happens when we need to add the validation message to more than one rule at the same time, so how will `style A` avoid the problem? I think you forgot to demonstrate correctly, no?

Check the example again:

```php
#[Rule('required', message: 'Please enter a title.')]
#[Rule('min:5', message: 'Your title is too short.')]
public $title = '';
```

👆 Extracted from [here](https://github.com/livewire/livewire/discussions/5873)

The proposal of the the `style A` is provide us something like that?

```php
#[Rule(
    ['required', message: 'Please enter a title.'], 
    ['min:5', message: 'Your title is too short.')
]
public $title = '';
```

If the answer is no, then the `style B` is the ideal.

joshhanley:
@nuernbergerA should `#[On()]` also be repeatable do you think? See #5912.

calebporzio:
Legend, thanks Adrian

nuernbergerA:
@RikvdHeijden pls open a discussion and post a simple code example to reproduce the bug you facing - or even better submit a PR with a failing test

RikvdHeijden:
@nuernbergerA 

thank you for the quick action. For the problem's I've described in my edit it seems they only happen when using a form object. Below test should recreate the issues I encountered.

```php
<?php

namespace Livewire\Features\SupportValidation;

use Livewire\Form;
use Livewire\Livewire;
use Tests\TestComponent;

class TestForm extends Form
{
    #[Rule('required')]
    #[Rule('min:2')]
    #[Rule('max:3')]
    public $foo = '';

    #[
        Rule('sometimes'),
        Rule('max:1')
    ]
    public $bar = '';
}

class UnitTest extends \Tests\TestCase
{
    /** @test */
    public function base_rule_attributes_merge_error()
    {

            public TestForm $form;
        })
            ->set('form.foo', '')
            ->assertHasErrors(['form.foo' => 'required'])
            ->set('form.foo', '1')
            ->assertHasErrors(['form.foo' => 'min'])
            ->set('form.foo', '12345')
            ->assertHasErrors(['form.foo' => 'max'])
            ->set('form.foo', 'ok')
            ->assertHasNoErrors()
            ->set('form.bar', '12')
            ->assertHasErrors(['form.bar' => 'max'])
            ->set('form.bar', '1')
            ->assertHasNoErrors()
        ;
    }
}
```

citricguy:
I just updated to beta.7 and still having issues with multiple rule calls:

This is based on:
```
    #[Rule('required', message: 'Required!')]
    #[Rule('min:3', message: 'Too short!')]
    public string $phone = '';
```

When an empty string is supplied (failing 'required) I get:
`htmlspecialchars(): Argument #1 ($string) must be of type string, array given`

When a 2 character string is supplied (failing 'min:3') I get:
`Undefined array key "string"`

Here is my submit function that calls the validation:

```
    public function submitForm()
    {
        $this->authorize('create', Company::class);
        $this->validate();
        $this->locationsFound = $this->findPlacesByPhoneNumber($this->phone);
        $this->step = 2;
    }
```

I've been staring at this for half an hour. I'm hoping this is just user-error.

I get exactly the same errors when using the [multi-rule code from the docs](#[Rule('required', message: 'Please enter a title.')]
#[Rule('min:5', message: 'Your title is too short.')]).

```
#[Rule('required', message: 'Please enter a title.')]
#[Rule('min:5', message: 'Your title is too short.')]
```

PR #5881, opened 2023-07-22
[V3] Updating the Upgrade Guide: Reinforcing Instructions to the Breeze & JetStream

devajmeireles:
# The problem

Two days ago I started experimenting with Livewire V3 and I found something that for me was a bug: I couldn't reset the property on the client side, **visually**. For example, resetting the value of `$title` back to empty within the input after it was used to create something like a todo.

# The report

👉  https://github.com/livewire/livewire/discussions/5797

**For these cases of test I was using Laravel Breeze**. After a while I discovered that carelessly I had not removed the Alpine loaded via `app.js` that is **delivered by default from Laravel Breeze and Laravel JetStream.**

# The solution

I decided to write one more point in the specific upgrade guide of the Livewire V3 for anyone trying to use Livewire V3 with these solutions, Breeze or JetStream.

## Definitive Solutions, in Breeze or JetStream

The Laravel team will adapt the Breeze and JetStream to the Livewire V3 when it officially released.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -260,7 +260,7 @@ Alpine.start()
 You can remove them entirely because Livewire includes Alpine and many popular Alpine plugins by default.
 
 > [!warning] "Livewire V3 (beta) with Laravel Breeze or Laravel Jetstream"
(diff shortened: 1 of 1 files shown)


calebporzio:
Sounds good, thank you for the note!

PR #5889, opened 2023-07-22
[V3] Add morph failing test for nested children with empty divs

joshhanley:
Currently, if there are some empty divs in a parent component, when a network request happens on the parent component, it breaks the child component. See error below

image

This PR adds a failing test, that only seems to fail intermittently with the browser test, but if you put a `->tinker()` in and interact with the refresh button a couple of times, you will see the above error.

It seems like this is a morph issue.

Related discussions #5860, #5935, #5960


devajmeireles:
> Currently, if there are some empty divs in a parent component, when a network request happens on the parent component, it breaks the child component. See error below
> 
> image
> This PR adds a failing test, that only seems to fail intermittently with the browser test, but if you put a `->tinker()` in and interact with the refresh button a couple of times, you will see the above error.
> 
> It seems like this is a morph issue.
> 
> Related discussion #5860

Awesome! One of my friend report me this issue, @luanfreitasdev

joshhanley:
@edalzell FYI this is still failing after the latest morph changes. So will have to dig further to find out what is going on.

joshhanley:
@edalzell FYI this is still failing after the latest morph changes. So will have to dig further to find out what is going on.

joshhanley:
So digging further into this issue, the problem related to morph's forward lookup https://github.com/alpinejs/alpine/blob/69c640f9231933d77b304c3f0bbdcc16524cbdc5/packages/morph/src/morph.js#L221-L235

Currently, what is happening is, as there are two empty `<div>`s in the test, morph is iterating through the "to" elements and finding the later empty div, and moving it to next to the currently morphing `<div>`.

This seems to then throw out the look up for the child component, so its element is currently being rebuilt - rather than moved - so all Livewire state is being wiped.

Hence then the snapshot error is occurring as Livewire is trying to initialise that component when there isn't any data to initialise it with.

Not sure exactly how to fix it though.

calebporzio:
Digging into this now

calebporzio:
Found the issue - the lookahead was being triggered against every element in the list, not just elements that are different.

I added a check so the lookahead is only triggered if the elements being compared are different in some way.

I think there are probably deeper issues with the lookahead algorithm, but we'll take these one at a time.

Thanks for the research and the failing test!

edalzell:
https://media.giphy.com/media/jDF4mCLDsBe1i/giphy.gif

luanfreitasdev:
wow that fixed the problem 🫶

PR #5890, opened 2023-07-23
[V3] Add Method: `resetSafe`

devajmeireles:
# The Problem

Most of the time when we're working with Livewire, we need to get the value of the properties to perform an operation and then it's important to reset the properties so that they go back to their original states 👇 

```php
public $title = '';

// ...

public function create()
{
    $todo = new Todo();

    // binding $title with front-end and edit it...

    $todo->title = $this->title;

    $todo->save();

    $this->reset();
}
```

OK! That works, but having to use the property and then reset it can be a bit exhausting, no? 🤔 So this PR aims to "solve" this problem by adding the `resetSafe` method 👇 

# Solution

The `resetSafe` method allows us to reset the properties, BUT before the reset is actually performed we can capture the current values of the properties:

### Single Property

```php
public $title = '';

// ...

public function create()
{
    // binding $title with front-end and edit it...

    $todo = new Todo();

    $todo->title = $this->resetSafe('todo')

    $todo->save();

    // $this->reset(); // 👈 This is not necessary here since `resetSafe` already does it.
}
```

### Multiple Properties

```php
public $todo = '';

public $description = '';

// ...

public function create()
{
    // binding $title and $description with front-end and edit it...

    [$title, $description] = $this->resetSafe('todo', 'description');

    $todo = new Todo();

    $todo->title = $title;

    $todo->description = $description;

    $todo->save();

    // $this->reset(); // 👈 This is not necessary here since `resetSafe` already does it.
}
```

---

Hope this help,

Thanks. ❤️

### DIFF
--- a/docs/properties.md
+++ b/docs/properties.md
@@ -167,7 +167,7 @@ class ManageTodos extends Component
 
     public function addTodo()
     {
-        [$title] = $this->resetSafe('todo'); // [tl! highlight]
+        $title = $this->resetSafe('todo'); // [tl! highlight]
         
         $this->todos[] = $title;
     }
--- a/docs/properties.md
+++ b/docs/properties.md
@@ -128,7 +128,7 @@ namespace App\Livewire;
 
 use Livewire\Component;
 
-class ManageTodos extends Controller
+class ManageTodos extends Component
 {
     public $todos = [];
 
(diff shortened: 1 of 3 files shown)


docs/properties.md
```diff
@@ -128,7 +128,7 @@ namespace App\Livewire;
 
 use Livewire\Component;
 
-class ManageTodos extends Controller
+class ManageTodos extends Component
```
devajmeireles, line 131: This was wrong. I fixed at the same PR.

joshhanley: @devajmeireles please split unrelated changes, like this one, into a new PR. Thank!

devajmeireles: > @devajmeireles please split unrelated changes, like this one, into a new PR. That way if Caleb doesn't merge this feature, then that docs typo will still be fixed. Thanks!

Make sense. Thank you! I'll update the PR.

calebporzio:
Yes, I totally agree it's annoying.

I originally had `->reset()` return the old value in V3 so that you could use it like you're suggesting.

I tweeted about it and many thought it was confusing and a bad idea so I reverted it.

So I validate your point and would like a solution, but I'm not sure a new method name like `resetSafe` is it.

I'm open to other API's but going to close this one for now especially as we're more focused on just making V3 stable before adding new features. Thanks!

devajmeireles:
> Yes, I totally agree it's annoying.
> 
> I originally had `->reset()` return the old value in V3 so that you could use it like you're suggesting.
> 
> I tweeted about it and many thought it was confusing and a bad idea so I reverted it.
> 
> So I validate your point and would like a solution, but I'm not sure a new method name like `resetSafe` is it.
> 
> I'm open to other API's but going to close this one for now especially as we're more focused on just making V3 stable before adding new features. Thanks!

Make sense, mainly the focus on making V3 stable.

PR #5893, opened 2023-07-23
[v3] Fix @entangle upgrade regex

PhiloNL:
This fixes: https://github.com/livewire/livewire/discussions/5868


joshhanley:
Thanks!

PR #5895, opened 2023-07-23
Add failing test when using `@persist` with iframes

PhiloNL:
Added failing test for https://github.com/livewire/livewire/discussions/5847

### DIFF
--- a/src/Features/SupportNavigate/BrowserTest.php
+++ b/src/Features/SupportNavigate/BrowserTest.php
@@ -18,6 +18,7 @@ class BrowserTest extends \Tests\BrowserTestCase
             Livewire::component('first-page', FirstPage::class);
             Livewire::component('first-page-with-link-outside', FirstPageWithLinkOutside::class);
             Livewire::component('second-page', SecondPage::class);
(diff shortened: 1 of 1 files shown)


joshhanley:
@PhiloNL wow nice job! 😁

calebporzio:
This is an interesting one. Apparently not a simple fix.

Turbo drive has dealt directly with this problem as well:
https://github.com/hotwired/turbo/issues/50
https://github.com/hotwired/turbo/issues/283

I'm not sure they've found a solution other than weird workarounds.

I'm hopeful we can figure something out, but I'm going to keep working through more pressing bugs before returning to this.

I'll leave this open for now because it's a valid problem.

Thanks for the test Philo!

levu42:
@calebporzio just saw this and was curious, so I googled a bit around and apparently a working solution seems to be to use shadow DOM & slots: https://bugzilla.mozilla.org/show_bug.cgi?id=254144#c106 https://github.com/whatwg/html/issues/5484#issuecomment-620481794 - didn't test it or anything though, but just to have it noted down here, in case this comes up again in the future

PR #5898, opened 2023-07-23
[v3] imageSize method on TemporaryUploadedFile to support dimensions validation rule

gdebrauwer:
Same as #5896 but for v3.


calebporzio:
Thanks!

PR #5900, opened 2023-07-23
[v3] Fixes stubs namespace generation

rahulhaque:
Fixes the issue with wrong namespace generation for forms and attributes, as they were hard coded. For example, `artisan livewire:form Auth/LoginForm` and `artisan livewire:attribute Auth/Content` will ignore the `Auth` part completely. This fixes the issue.


joshhanley:
Thanks!

PR #5903, opened 2023-07-23
[v3] Add beta warning to quickstart

PhiloNL:
The beta warning is now only shown on the upgrade guide. This PR will add the warning to the default documentation page (quickstart) as well to manage expectations.


nuernbergerA:
Maybe also mention that the focus is not to PR new features. It's about finding bugs and testing livewire as it is

PR #5904, opened 2023-07-23
[v3] Add dynamic validation attributes in Form Object

viliusvsx:
Added PR for dynamic validation attributes usage from Form Object 
Discussion: [#5853](https://github.com/livewire/livewire/discussions/5853)

Currently, dynamic attributes are available from component itself but not from form object.
This PR allow use dynamic attributes in method and do things, that in rule attribute is impossible


viliusvsx:
If in Livewire component, there is Form Object named "form", then validationAttribute method will be overwritten by Rule Attribute class. This feature with dynamic validationAttribute working if form name is not "form" but something like "postForm" or anything. Dynamic messages working

viliusvsx:
But what about attribute names that need to be translated using this __()

joshhanley:
@viliusvsx yeah you're correct, translation strings can't be passed into attributes, so I'm going to reopen this.

calebporzio:
Ok, I see.

You made the right choice being consistent with Livewire component classes for: 'validationAttributes' and 'messages' as method names

I always disliked 'validationAttributes' instead of 'attributes'. We did it so that people weren't confused with Livewire properties or any other "attributey" type thing.

Should we start clean with `attributes()` in Form objects instead of `validationAttributes()` or is that too much confusion. Should be add both? For bc?

Yeah, how do we feel about adding both.

We can merge this after we decide.

Thanks!

nuernbergerA:
@calebporzio now is the time to make BC :muscle:

viliusvsx:
I think method name could be attributes() instead of validationAttributes(). Since this will be added in Form Objects, short method name speak for itself

calebporzio:
Yeah, let's do `attributes()` - mind updating the PR?

viliusvsx:
Yeah, in a minute

viliusvsx:
I also not quite sure, but maybe these methods in form object `addValidationRulesToComponent()`, `addAttributesToComponent()`, `addMessagesToComponent()` make `protected` instead of `public` ? I am thinking that in form object users could never use these directly

devajmeireles:
Will `attributes` also accept custom validation class?

joshhanley:
@devajmeireles what do you mean by that?

joshhanley:
@viliusvsx just to give you an update, we were looking at changing it to `$attributes/attributes()` as well for component validation (not just FormObject validation) but there is possible confusion/ conflict if someone passes `$attributes` into their component like a blade component. 

So we're thinking it might be better to just stick to `$validationAttributes/validationAttributes()`.

@nuernbergerA and @PhiloNL what are your thoughts?

nuernbergerA:
@joshhanley naming is hard - i vote for consistency

PR #5910, opened 2023-07-23
[v3] Add failing test for route caching

nuernbergerA:
Failing test for `php artisan route:cache`

----
Fixes https://github.com/livewire/livewire/discussions/5825

### DIFF
--- a/src/Tests/LivewireRoutCacheTest.php
+++ b/src/Tests/LivewireRoutCacheTest.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Livewire\Tests;
+
+use Error;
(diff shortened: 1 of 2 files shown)


zepfietje:
Thanks for looking into this, @nuernbergerA!
I ran into this when deploying an application yesterday.

nuernbergerA:
@zepfietje Problem is that the tests do not fail in ci but fail locally.

Underlying problem:
When the routes are cached the closure will be serialized but somehow every is bound to the closure causing a memory overflow.

calebporzio:
We can close this as it's been fixed - thanks!

PR #5913, opened 2023-07-23
[v3] Create failing test for missing `fromChildren` property in On attribute

nexxai:
Refers to #5886


nexxai:
Ok, I don't know why these are passing.  They failed on my machine.  (I feel like I'm in opposite land today)

joshhanley:
@nexxai thanks for the PR! Hmm can you try making it a browser test instead and see if that works? Also can you put it `src/Features/SupportEvents/BrowserTest.php`? Thanks!

joshhanley:
@nexxai actually you could leave it as a unit test and call `->dispatch()` in the test, which then should trigger the error. I would still move the test but if it's a unit test, move it to `src/Features/SupportEvents/UnitTest.php`.

nexxai:
Give me a minute to fix the unintentional autoformatting

nexxai:
Ok that's better.  Thanks @joshhanley for your guidance.

joshhanley:
@nexxai no worries! Thanks for fixing it 🙂

calebporzio:
Sorry about that - I documented it but actually decided to remove it as a feature in V3. It adds a non-trivial amount of complexity and now that there is `$parent`, there's less of a direct need for it.

If the community agrees that this was a mistake we can consider adding it back.

For now I just removed the docs. Thank you!

riasvdv:
@calebporzio 

We actually have a need for this, I have 2 components on a page that each have a child component:

- ListUploadsComponent (images)
    - UploadComponent
- ListUploadsComponent (downloads)
    - UploadComponent 
    
When a file gets uploaded, I do some processing on the file inside `UploadComponent`, after that I dispatch a "fileAdded" event, right now both `ListUploadsComponent`s handle the event, I need a way to only handle the event if it comes from a child component.

We could probably work around this by prefixing events, but would love it if this functionality worked.

nexxai:
@riasvdv has almost my exact same use case. My difference is that I have multiple of the same component on a single page that are deeply nested (3 levels) so I can use the $parent on the deepest level but I would use that reference to call a method on the middle level, which would then dispatch an event to the top level.

Since you can't use $parent outside of Blade, I need a way to scope the level 2 -> level 1 call to only its own children.

PR #5916, opened 2023-07-24
[V3][Docs] Change Extends from Controller to Component

devajmeireles:
This PR aims to change the "extends" of one of the examples from Controller, that's wrong for Livewire components, to Component.


joshhanley:
Thanks!

PR #5920, opened 2023-07-24
[v3] Add failing tests for updating typed properties

joshhanley:
Copy from https://github.com/livewire/next/pull/80

One of the issues developers regularly run into is when setting a public property of type ?int and binding an input to it, typing a number into the input and then clearing it causes a "Cannot assign string to property of type ?int" error.

image

I thought the synths would fix this, but it turns out it hasn't.

For now, this PR just adds a failing test, but I'll submit a fix if you agree that this should work and not throw an error.

Thoughts?

I've just used a text input for the test but the same issue happens if you use a number input <input type="number" />.

I think we need some solution for converting empty strings to null if the type supports null like ?int.

I also did some further testing and the same issue happens with Enums (which have a synth).

Here is the enum
```php
enum Suit: string
{
    case Hearts = 'H';

    case Diamonds = 'D';

    case Clubs = 'C';

    case Spades = 'S';
}
```

And the component
```php
class Test extends Component
{
    public Suit $selected;

    public function render()
    {
        return <<<'HTML'
        <div>
            <select wire:model.live="selected">
                @foreach(App\Enums\Suit::cases() as $suit)
                    <option value="{{ $suit->value }}">{{ $suit }}</option>
                @endforeach
            </select>
        </div>
        HTML;
    }
}
```

Triggers this error
image


PhiloNL:
In v2, I had the same issue, and I ended up writing a "resolver" that would use reflection to initiate the object before passing it to the component e.g., this is how I fixed this for enums:

```php
class EnumPropertyResolver extends Resolver
{
    public function handle(): array
    {
        if (PHP_VERSION_ID < 80100) {
            return $this->properties;
        }

        $reflection = new ReflectionClass($this->componentClass);
        $resolveEnumProperties = collect($reflection->getProperties())
            ->filter(function (ReflectionProperty $property) {
                return isset($this->properties[$property->getName()]);
            })
            ->reject(function (ReflectionProperty $property) {
                return $property->getType() instanceof ReflectionUnionType;
            })
            ->filter(function (ReflectionProperty $property) {
                return $property->getType()?->getName() && enum_exists($property->getType()->getName());
            })->mapWithKeys(function (ReflectionProperty $property) {
                return [$property->getName() => $property->getType()->getName()::tryFrom($this->properties[$property->getName()])];
            });

        return collect($this->properties)->merge($resolveEnumProperties)->toArray();
    }
}
```

Maybe there's a way to hook a synth earlier in the process so strings are converted into objects before it tries to assign them to the property.

joshhanley:
@PhiloNL yeah synths already have this support built in, but currently they're being bypassed here 

https://github.com/livewire/livewire/blob/b57896e44397d7c5b1c42db64dba4c8382959357/src/Mechanisms/HandleComponents/HandleComponents.php#L302-L305

PR #5931, opened 2023-07-24
[V3] Failing test: wire:loading with JS object

danharrin:
Imagine you have a button:

```html
<button wire:click="foo(@js(['bar' => 'baz']))"></button>
```

Both of these indicators will show, even though only the first should:

```html
<span wire:target="foo(@js(['bar' => 'baz']))" wire:loading>Loading</span>
<span wire:target="foo(@js(['baz' => 'bar']))" wire:loading>Loading</span>
```

I think that the JS inside the `wire:target` needs to be evaluated in a similar way to `wire:loading` to ensure that the params are matched exactly.


calebporzio:
@danharrin - have you looked into the cause of this problem at all?

zepfietje:
@danharrin, just reminding you to reply to Caleb so we can maybe have a fix in place before releasing Filament v3 Tuesday. 🙂

calebporzio:
Lol, fixed it

zepfietje:
Thanks a lot, @calebporzio!

danharrin:
Thanks so much Caleb! Super busy with everything else, sorry for not getting back in time

PR #5933, opened 2023-07-24
change public to private in the sample code for making dangerous methods private

mokhosh:
Edit: Initially I thought this code is a mistake by the author of the docs.
But then I realised it's an intentional demonstration of a possible mistake by developers. So I changed my PR from a change in the code, to a "warning" comment at the top of the code.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -858,7 +858,7 @@ class BrowsePosts extends Component
         $this->delete($id); // [tl! highlight]
     }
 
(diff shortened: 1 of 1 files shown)


mokhosh:
😅 My bad, I must have red the rest of the docs before jumping here. The next paragraph explains that this is wrong and what we should do instead.

Changed the visibility back to `public` and added one of those "warning" comments at the top, like you had in other secions.

PR #5939, opened 2023-07-24
[V3] Document the deprecation of @this.on()

markjaquith:
See discussion at https://github.com/livewire/livewire/discussions/5892

`@this.on('event-name', callbackFn)` could be used in Livewire 2 inside Blade (frequently in an `x-init`) to tell Livewire to run a JavaScript function upon a component action being triggered.

In Livewire 3, this does not work, and instead Livewire tries to call a non-existent `on()` method on the component, which fails.

Since Alpine is now included, this syntax works:

`x-on:event-name="callbackCode"`

This PR documents that change in the upgrade docs.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -606,6 +606,10 @@ Livewire.hook('commit', ({ component, commit, respond, succeed, fail }) => { // 
 
 You may consult the new [JavaScript hook documentation](/docs/javascript) for a more thorough understanding of the new hook system.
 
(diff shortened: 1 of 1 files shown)


joshhanley:
@markjaquith thanks! I will get Caleb to have a look. Also using `wire:event-name="..."` should also work now too I believe.

calebporzio:
Yes, both will work. Thanks Mark!

PR #5943, opened 2023-07-25
[V3] add failing test for empty query param

ariaieboy:
I explained about this here https://github.com/livewire/livewire/discussions/5932

### DIFF
--- a/src/Features/SupportQueryString/BrowserTest.php
+++ b/src/Features/SupportQueryString/BrowserTest.php
@@ -17,14 +17,33 @@ class BrowserTest extends BrowserTestCase
     /** @test */
     public function it_will_remove_empty_input()
     {
-        Livewire::withQueryParams(['page' => ''])->visit(TestComponent::class)
+        Livewire::withQueryParams(['page' => ''])->visit(TestComponentWithOutHistory::class)
+            ->assertQueryStringMissing('page');
+        Livewire::withQueryParams(['page' => ''])->visit(TestComponentWithHistory::class)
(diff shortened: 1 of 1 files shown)


joshhanley:
@ariaieboy thanks!

calebporzio:
I think there are two things going on here:

A) A bug that converts empty string values to null in the query string (and actual causes other problems when reloading the page)

B) The fact that an entry with "keep: false" don't get removed automatically if it's present in the query string.

A is a bug and I've fixed it

B is actually maybe intended behavior. "keep: false" will make sure a default value won't get added to query string, but if the user adds it feels fine to leave it and maybe heavy handed to auto-remove it.

Therefore I'm closing this PR. 

Thank you for the help with the failing PR - very helpful to debug this.

ariaieboy:
> I think there are two things going on here:
> 
> A) A bug that converts empty string values to null in the query string (and actual causes other problems when reloading the page)
> 
> B) The fact that an entry with "keep: false" don't get removed automatically if it's present in the query string.
> 
> A is a bug and I've fixed it
> 
> B is actually maybe intended behavior. "keep: false" will make sure a default value won't get added to query string, but if the user adds it feels fine to leave it and maybe heavy handed to auto-remove it.
> 
> Therefore I'm closing this PR.
> 
> Thank you for the help with the failing PR - very helpful to debug this.

Thanks for your response. In Livewire v2 if you define the `except`, it will remove the default value query param.
Since we no longer have an `except` API, I think we should use the `keep` option or introduce a new API.

PR #5947, opened 2023-07-25
[V3] Add failing test for testing a component without a render method

joshhanley:
Currently, if you try to test a component that doesn't have a render method, a `ViewException` is thrown because it is looking for a view with the hashed name of the component class instead of the kebab-cased name of the component. See discussion #5944.

image

This is happening because of the code below, but I'm not exactly sure what the code is meant to achieve, otherwise I would have fixed it.

https://github.com/livewire/livewire/blob/b57896e44397d7c5b1c42db64dba4c8382959357/src/Mechanisms/ComponentRegistry.php#L145-L150

If you can provide guidance on what the hash check is meant to be for, I can look at fixing it.

Hope this helps!


calebporzio:
This was a surprisingly deep one. But got it fixed and tested.

PR #5948, opened 2023-07-25
[v3] Add nullable type to Post model in forms PostFrom docs section

krekas:
Updating docs as agreed in this discussion https://github.com/livewire/livewire/discussions/5888#discussioncomment-6536012


joshhanley:
@krekas thanks! I removed `= null` as it's not needed 🙂

PR #5950, opened 2023-07-25
[V3] Add failing test for testing a component with custom directive

nleibbrandt:

i found two discussions under Alpine Repo, on of them https://github.com/alpinejs/alpine/discussions/3660

Yes

No

Yes, failing test

The main problem is that it is currently not possible to use alpine components with livewire

Thanks for contributing! 🙌


joshhanley:
@nleibbrandt thanks!

joshhanley:
@nleibbrandt thanks for the PR. Please don't add new tests to `legacy_tests` as we're trying to move everything out of there. If you're unsure where to put them, put them somewhere under the `tests` directory. 🙂

nleibbrandt:
@joshhanley aah now i see it. Thank you for support!

joshhanley:
@nleibbrandt no worries! 🙂

calebporzio:
Thanks for the failing test! It's fixed now.

For the zero people who care:

What was happening before:

- Alpine state is initialized on the page
- A Livewire request is made
- Before morphdom patches the old tree with the new tree...
- Alpine initializes itself on the new tree. This process is called "cloning" a tree
- It's called "cloning" because it takes the live Alpine x-data state from the old tree, grafts it on the new tree, then initializes with that state
- During "clone" Alpine skips registering x-datas by default

The problem is that the x-data state provided here and live on the page is NOT on the root element (common thing to do) and also it's not even in the DOM - it's bound using Alpine.bind(). So when Alpine clones the tree, (remember it skips x-data directives during clone) there is no state provided to the new Alpine component at all and it throws this error.

The fix is to CONDITIONALLY initialize x-data when cloning a tree: https://github.com/alpinejs/alpine/commit/7043509b3df8d0eb7305353105b674de64165b0d

Again, no one cares but I'm throwing this here in case anyone including myself finds it useful at some point

nleibbrandt:
@calebporzio thanks, that was last missing piece for [TALL kit components](https://github.com/tall-kit/components)!
You are the best Caleb!

PR #5956, opened 2023-07-25
[V3] Add an upgrade step that upgrades $this->forgetComputed calls to php's unset

julesgraus:

Yes. The livewire:upgrade command forgets to upgrade $this->forgetComputed calls. This PR fixes that. See #5905 

Yes.

No. But I will create another PR that upgrades computed properties to the new syntax.

No, none of the upgrade commands do this yet. If you give me directions on how you guys would like it, I will make it.

Livewire components sometimes contain calls to ```$this->forgetComputed()```. This functionality is replaced by ```unset``` calls.

Thanks for contributing! 🙌


src/Features/SupportConsoleCommands/Commands/Upgrade/UpdateForgetComputedCalls.php
```diff
@@ -0,0 +1,24 @@
...
+    {
+        $this->interactiveReplacement(
+            console: $console,
+            title: 'The forgetComputed component method is replaced by php\'s unset function',
+            before: '$this->forgetComputed()',
+            after: 'unset()',
+            pattern: '/\$this->forgetComputed\([\'"](.+)[\'"]\)/',
+            replacement: 'unset($this->$1)',
+            directories: ['app'],
+            subjectName: 'calls'
```
PhiloNL, line 19: This doesn't exist in this method.

src/Features/SupportConsoleCommands/Commands/Upgrade/UpdateForgetComputedCalls.php
```diff
@@ -0,0 +1,24 @@
...
+class UpdateForgetComputedCalls extends UpgradeStep
+{
+    public function handle(UpgradeCommand $console, \Closure $next)
+    {
+        $this->interactiveReplacement(
+            console: $console,
+            title: 'The forgetComputed component method is replaced by php\'s unset function',
+            before: '$this->forgetComputed()',
+            after: 'unset()',
+            pattern: '/\$this->forgetComputed\([\'"](.+)[\'"]\)/',
```
PhiloNL, line 16: This regular expression will only match quotes.
Someone might have various different approaches: 
```php
$someVariable = 'foo';
$this->forgetComputed($someVariable)
$this->forgetComputed($this->getSomePropertyName())
$this->forgetComputed('bar');
```

This regex will ensure all these cases are captured:
```suggestion
            pattern: '/\$this->forgetComputed\(((?:[^)(]|\((?:[^)(]|\((?:[^)(]|\([^)(]*\))*\))*\))*)\)/',
```

PhiloNL (changes requested):
Thanks for the PR! 👍 
I've added two comments.

src/Features/SupportConsoleCommands/Commands/Upgrade/UpdateForgetComputedCalls.php
julesgraus: Seems better indeed. Now i am curious about why you need to match that much of ( ) chars? The next example matches too (which isn\t a problem). But one more set of () doesn't. Why that exact amount?
$this->forgetComputed(((($someVariable))))

Now i am thinking about ```$this->forgetComputed('bar');```. It does upgrade it to unset('bar'). Is that specific case handled by these calls? I'm not sure?

```\Livewire\Features\SupportComputed\Computed::handleMagicUnset```
and then a bit further:
```\Livewire\Features\SupportComputed\Computed::evaluateComputed```

File: src/Features/SupportConsoleCommands/Commands/Upgrade/UpdateForgetComputedCalls.php (hunk 1 above)
Comment (julesgraus) [reply src/Features/SupportConsoleCommands/Commands/Upgrade/UpdateForgetComputedCalls.php:19]: Thanks. Now it does. I've added this to improve the message for the user. It now states this:
```This means all $this->forgetComputed() calls must be changed to unset().``` instead of 
```This means all $this->forgetComputed() directives must be changed to unset().```

**Comment by joshhanley** [comment]
@julesgraus don't worry about squashing commits, we do that when we merge it into the main branch 🙂

**Comment by PhiloNL** [comment]
Taking a look at this tomorrow :)

File: src/Features/SupportConsoleCommands/Commands/Upgrade/UpdateForgetComputedCalls.php (hunk 2 above)
Comment (PhiloNL) [reply src/Features/SupportConsoleCommands/Commands/Upgrade/UpdateForgetComputedCalls.php]: I had some issues with writing the regex that would cause catastrophic backtracking errors when dealing with parentheses. This regex might be a bit much, but it should cover all weird cases when there are a ton of parentheses. But feel free to PR the regex if you have an optimized version :)

Comment (PhiloNL) [reply src/Features/SupportConsoleCommands/Commands/Upgrade/UpdateForgetComputedCalls.php]: Looking at the original function, there's still some work to do.

```php
public function forgetComputed($key = null)
    {
        if (is_null($key)) {
           $this->computedPropertyCache = [];
           return;
        }

        $keys = is_array($key) ? $key : func_get_args();

        collect($keys)->each(function ($i) {
            if (isset($this->computedPropertyCache[$i])) {
                unset($this->computedPropertyCache[$i]);
            }
        });
    }
```

So people would use it as follows:

```php
$this->forgetComputed('one', 'two', 'three');
unset($this->one, $this->two, $this->three);

$this->forgetComputed('one');
unset($this->one);

$this->forgetComputed(['one', 'two', 'three']);
unset($this->one, $this->two, $this->three);

// The following would be difficult to automate and out of scope.
$this->forgetComputed($this->returnSomeName());
```

PhiloNL: Ok I've pushed some changes to match and replace the following:

```php
$this->forgetComputed('bar');
$this->forgetComputed('bar', 'hello');
$this->forgetComputed(['bar', 'hello']);
```

Other scenarios are out of scope, don't you think? Or do you want to give those a try 🙈

julesgraus: No I think this is fine. I think that the majority of users will only pass one value at a time to forget computed. I think you need to use something like  https://github.com/nikic/PHP-Parser to sanely refactor other cases. Anyways. Thank you for this stuff. I hope I can contribute some valuable stuff to livewire soon. Even bugfixes.

baoc75:
I think we missed an edge case where people use `$this->forgetComputed()` to refresh computer properties since v2. So we should give warning to people about this.

PR #5965, opened 2023-07-25
[V3] Various doc fixes

PerryvanderMeer:


devajmeireles:
Awesome, @PerryvanderMeer ! ❤️

docs/computed-properties.md
```diff
@@ -1,6 +1,6 @@
 Computed properties are a way to create "derived" properties in Livewire. Like accessors on an Eloquent model, computed properties allow you to access values and cache them for future access during the request.
 
-Computed properties are particularly useful in combination with a component's public properties.
```
joshhanley, line 3: I would revert this as it reads better to me with the "a".

docs/computed-properties.md
```diff
@@ -72,7 +72,7 @@ Consider the following problematic scenario:
 
 To clear, or "bust", the stored cache, you can use PHP's `unset()` function.
 
-Below is an example of an action called `createPost()` that, by creating a new post in the application, makes the `posts()` computed stale—meaning the `posts` computed property needs to be re-computed to include the newly added post:
```
joshhanley, line 75: I would revert this, as they're called computed properties, so makes more sense to me to leave "computed" and "property" together.

docs/computed-properties.md
```diff
@@ -109,7 +109,7 @@ In the above component, the computed property is cached before a new post is cre
 
 Sometimes you would like to cache the value of a computed property for the lifespan of a Livewire component, rather than it being cleared after every request. In these cases, you can use [Laravel's caching utilities](https://laravel.com/docs/cache#retrieve-store).
 
-Below is an example of a `user()` computed property, where instead of executing the Eloquent query directly, we wrap the query in `Cache::remember()` to ensure that any future requests retrieve it from Laravel's cache instead of re-executing the query:
```
joshhanley, line 112: Same as above.

docs/dirty.md
```diff
@@ -13,7 +13,7 @@ Livewire allows you to easily toggle visual elements on the page using the `wire
 
 By adding `wire:dirty` to an element, you are instructing Livewire to only show the element when the client-side state diverges from the server-side state.
 
-To demonstrate, here is an example of an `UpdatePost` form containing an "Unsaved changes..." visual indication that signals to the user that the form contains input that has not been saved:
```
joshhanley, line 16: I would revert this as it reads better this way to me.

docs/dirty.md
```diff
@@ -27,9 +27,9 @@ To demonstrate, here is an example of an `UpdatePost` form containing an "Unsave
 </form>
 ```
 
-Because `wire:dirty` has been added to the "Unsaved changes..." message, the message will be hidden by default. Livewire will automatically display the message only when the user starts modifying the form inputs.
```
joshhanley, line 30: I think the work "only" is required here.

docs/offline.md
```diff
@@ -12,7 +12,7 @@ By adding `wire:offline` to an element inside a Livewire component, it will be h
 
 ## Toggling classes
 
-Adding the `class` modifier allows you to add a class to an element when the user loses their connection. The class will be removed again once the user is back online:
```
joshhanley, line 15: I would revert this or change it to "...removed again, once the user...".

docs/pagination.md
```diff
@@ -51,11 +51,11 @@ For more information on pagination using Laravel, check out [Laravel's comprehen
 
 ## Resetting the page
 
-When sorting or filtering results, it is common to want to reset the page number back to `1`.
```
joshhanley, line 54: I'm 50/50 on this change, I think "to want to" reads better, but not overly against your change either.

docs/pagination.md
```diff
@@ -219,10 +219,10 @@ class ShowPosts extends Component
 
 The previous hooks only apply to the default paginator. If you are using a named paginator, you must define the methods using the paginator's name.
 
-For example, below is an example of what a hook for a paginator named `invoices-page` would look like:
```
joshhanley, line 222: I would revert this, as I think the plural is better here, as the paginator is for a collection of invoices, not a single invoice.

docs/pagination.md
```diff
@@ -219,10 +219,10 @@ class ShowPosts extends Component
...
+For example, below is an example of what a hook for a paginator named `invoice-page` would look like:
 
 ```php
-public function updatingInvoicesPage($page)
```
joshhanley, line 225: Same as above.

joshhanley (changes requested):
@PerryvanderMeer thanks for the great and thorough PR! I have requested some minor changes 🙂

docs/computed-properties.md
PerryvanderMeer: Seems a little bit off to me to write about plurals with an 'a' in between

docs/pagination.md
PerryvanderMeer: There was an inconsitency with the code above. I've changed all to invoices-page.

PerryvanderMeer:
@joshhanley Think this is it for now 👍

joshhanley:
@PerryvanderMeer ah sorry I didn't realise you weren't finished with the PR. In future, can you please mark PRs draft if they aren't ready to review, as we don't want to review PRs multiple times (especially big ones like that). I will have to go through your new changes later when I get a chance 🙂

PerryvanderMeer:
@joshhanley My bad, thought I was ready, but missed a few files. 🙂

docs/components.md
```diff
@@ -175,13 +175,13 @@ class CreatePost extends Component
     public function render()
     {
         return view('livewire.create-post')->with([
-	        'author' => auth()->user()->name,
+	        'author' => Auth::user()->name,
```
nuernbergerA, line 178:
```suggestion
	        'author' => auth()->user()->name,
```
this was fine.
With`Auth::user()` you also need to import the class

PerryvanderMeer: `Auth::user()` is used everywhere else, so I wanted to be consistent with that.

docs/computed-properties.md
joshhanley: "componet's" is possessive, if it was "components" then yeah I'd agree 🙂

docs/pagination.md
joshhanley: Need to change this back to "to want to".

docs/properties.md
```diff
@@ -365,7 +365,7 @@ In order to update the property without sending a network request to the server,
 
 While Livewire properties are a powerful feature, there are a few security considerations that you should be aware of before using them.
 
-In short, always treat public properties as user input—as if they were request input from a traditional endpoint. In light of this, it's essential to validate and authorize properties before persisting them to a database—just like you would do when working with request input in a controller.
```
joshhanley, line 368: Need to revert this as "request input" is the correct wording.

docs/properties.md
```diff
@@ -527,7 +527,7 @@ class UpdatePost extends Component
 
 ### Properties expose system information to the browser
 
-Another essential thing to remember is that Livewire properties are serialized or "dehydrated" before they are sent to the browser. This means that their values are converted to a format that can be sent over the wire and understood by JavaScript. This format can expose information about your application to the browser, including the names and class names of your properties.
```
joshhanley, line 530: Swap this back.

joshhanley (changes requested):
@PerryvanderMeer thanks again for this! A couple of small changes, which I will make and then merge 🙂

joshhanley (approved):

PR #5967, opened 2023-07-25
[V3] Add `reset` method to Form Object

devajmeireles:
# Problem

```php
{
    public $title = '';

    public $description = '';

    public function save(): void
    {
        $this->validate();

        Todo::create($this->all());
    }
}
```

And then...

```php
{

    public CreateTodoForm $form;

    // ...

    public function add(): void
    {
        $this->form->save();

    }
}
```

# Solution

```php
{
    public $title = '';

    public $description = '';

    public function save(): void
    {
        $this->validate();

        Todo::create($this->all());

        $this->reset(); // all...

    }
}
```

Even knowing that the focus now is the release of the V3, this PR will be of great help to anyone who wants to use Form Objects with a more concentrated logic, like above examples.

---

Hope this help,

Thanks. 🚀

### DIFF
--- a/src/Features/SupportFormObjects/Form.php
+++ b/src/Features/SupportFormObjects/Form.php
@@ -49,6 +49,24 @@ class Form implements Arrayable
         return $this->component->validate($filteredRules)[$this->propertyName];
     }
 
+    public function reset(...$properties): void
+    {
+        $properties = array_filter($properties, fn ($property) => array_key_exists($property, $this->all()));
+
+        $reset = [];
+
+        $empty = empty($properties);
+        $properties = $properties ?: $this->all();
+
+        foreach ($properties as $key => $value) {
+            if (str($value)->startsWith($this->propertyName . '.')) continue;
+
+            $reset[] = $this->propertyName .'.'. ($empty ? $key : $value);
+        }
+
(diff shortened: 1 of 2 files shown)


joshhanley:
@devajmeireles thanks for the PR! Yep I agree having reset on the FormObject would be helpful.

I don't know whether it's better to proxy the reset call onto the component like you have done or just have a reset implementation on the FormObject itself?

I personally think the full reset implementation would be better that way the whole component and FormObject don't both need to be instantiated to reset a property. Let's wait to see what Caleb thinks.

devajmeireles:
> @devajmeireles thanks for the PR! Yep I agree having reset on the FormObject would be helpful.
> 
> I don't know whether it's better to proxy the reset call onto the component like you have done or just have a reset implementation on the FormObject itself?
> 
> I personally think the full reset implementation would be better that way the whole component and FormObject don't both need to be instantiated to reset a property. Let's wait to see what Caleb thinks.

Makes sense. The problem is that with this we would have two isolated codes for maintenance. Let me know about news. I can continue work on that.

calebporzio:
Yeah I agree, it's probably better to just have the reset on the form object itself? Maybe we extract into a trait that both share?

PR #5969, opened 2023-07-25
[v3] Add 'upload' as reserved keywords to the documentation

PhiloNL:
Follow up on: https://github.com/livewire/livewire/pull/5852#issuecomment-1650212964

### DIFF
--- a/docs/uploads.md
+++ b/docs/uploads.md
@@ -37,6 +37,9 @@ class UploadPhoto extends Component
 </form>
 ```
 
+> [!warning] The term "upload" is reserved
+> The term "upload" is reserved in Livewire. You cannot use it as a method or property name.
+
 From the developer's perspective, handling file inputs is no different than handling any other input type: Add `wire:model` to the `<input>` tag and everything else is taken care of for you.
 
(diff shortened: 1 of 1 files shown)


devajmeireles:
Perhaps this should be updated?

CleanShot 2023-07-25 at 5 43 55

joshhanley:
@devajmeireles those reserved words are more for component names, which the make command already uses anyway, so I don't think a warning is necessary for them (plus they are standard PHP reserved words).

Where as this PR is for a word that Livewire itself is reserving, so a method name shouldn't match it. If there were other words that Livewire specifically reserves, then yeah I agree they should be added 🙂

calebporzio:
Thanks @PhiloNL and @joshhanley for the extra context!

akunbeben:
I also have a component property named `$attributes`, which is reserved by the `Livewire\Features\SupportAttributes\HandlesAttributes` trait. Is it necessary to add a warning for this? Or it's just my specific case? What do you guys think?

image

joshhanley:
@akunbeben yeah it should probably be listed on the validation page.

PR #5971, opened 2023-07-25
[V3] Fix upgrade command to upgrade computed properties too

julesgraus:
…to methods which names are just the part that was between 'get' and 'property'. Adds the Computed attribute above the upgraded methods, and adds an import for the Computed property if it was not already imported.

Yes. see #5905

Yes

No.

No, because none of the already present upgrade steps have.

It adds an upgrade step to the ```php artisan livewire:upgrade``` command which upgrades methods whom names start with 'get' and end with 'Property' to methods which names are just the part that was between 'get' and 'property'. The upgraded methods are also annotated with the 'Computed' attribute. And the computed attribute will also get imported if the user did not already do so.

Thanks for contributing! 🙌


calebporzio:
This seems related to #6133 

I'm going to close these two right now because they've been left unattended for a little while. When @PhiloNL gets time, I imagine he will review them and can re-open and merge them at that time.

Thanks for the contribution!

PR #5972, opened 2023-07-25
[V3] failing test: Conditional markers containing @ with property-based co…

danharrin:
I did try hard to fix the regex but I just couldn't get it to play ball. I have narrowed it down quite far:

`>` from `$foo->bar` seems to incorrectly match the end of that tag

and then `@` inside that condition is not being matched anywhere?

The result here is that the first `@endif` has a marker, even though it's inside the html tag


joshhanley:
@danharrin thanks!

calebporzio:
Fixed - thanks Dan

PR #5975, opened 2023-07-26
Update Pagination Docs to Include Helpful Example

Infamoustrey:
Add's example to the docs for inline component pagination. I enjoy working exclusively with inline examples, and thought this example might save some others confusion. If there is a more proper way to do this please let me know!

### DIFF
--- a/docs/pagination.md
+++ b/docs/pagination.md
@@ -47,6 +47,43 @@ class ShowPosts extends Component
 
 As you can see, in addition to limiting the number of posts shown via the `Post::paginate()` method, we will also use `$posts->links()` to render page navigation links.
 
+Alternately if you prefer to use inline components you may simply add the render logic directly in the render method.
+
+```php
+<?php
+
+namespace App\Livewire;
+
+use Illuminate\Support\Facades\Blade;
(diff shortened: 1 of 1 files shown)


calebporzio:
Yeah, for inline component's I'd recommend using a computed property for Posts and not doing that Blade::render thing - thanks though!

Infamoustrey:
Okay rereading the computed properties section of the docs makes that obvious in retrospect, however it was not when reading the pagination. Is below something you'd be interested in including in the docs? Let me know and I'll update that.

For inline components you'll need a [computed property](https://livewire.laravel.com/docs/computed-properties#using-inline-templates) to provide the pagination object to the view.

```php
<?php

namespace App\Livewire;

use Livewire\WithPagination;
use Livewire\Component;
use App\Models\Post;

{
    use WithPagination;

    #[Computed()]
    public function posts()
    { 
        return Post::paginate(10);
    }

    public function render()
    { 
        return <<<'HTML'
            <div>
                <div>
                        <!-- ... -->
                    @endforeach
                </div>

                {{ $this->posts->links() }}
            </div>
        HTML; 
    }
}
```

PR #5994, opened 2023-07-26
[v3] Add support for third-party upgrade commands

PhiloNL:
I've been working on making other packages compatible with v3, like the [modal component](https://github.com/wire-elements/modal). I thought it might be cool to allow other third parties to hook into the upgrade command instead of each having to implement their own logic.

This PR will allow this as follows:

```php
UpgradeCommand::addThirdPartyUpgradeStep(\App\TestCommand::class);
```
At the same time, I'm adding closure support for more advanced replacements.
For example:

```php
// Basic replacement
$this->interactiveReplacement(
    console: $console,
    title: 'assertEmitted is now assertDispatched.',
    before: 'assertEmitted',
    after: 'assertDispatched',
    pattern: '/assertEmitted\((.*)\)/',
    replacement: 'assertDispatched($1)',
    directories: 'tests'
);

// Advanced replacement
$this->interactiveReplacement(
    console: $console,
    title: 'assertEmitted is now assertDispatched.',
    before: 'assertEmitted',
    after: 'assertDispatched',
    pattern: '/assertEmitted\((.*)\)/',
    replacement: function($matches) {
        $value = $matches[1];
        return "assertEmitted('$value')";
    },
    directories: 'tests'
);
```

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/Upgrade/ThirdPartyUpgradeNotice.php
+++ b/src/Features/SupportConsoleCommands/Commands/Upgrade/ThirdPartyUpgradeNotice.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Livewire\Features\SupportConsoleCommands\Commands\Upgrade;
+
+use Illuminate\Console\Command;
+
+class ThirdPartyUpgradeNotice extends UpgradeStep
+{
+    public function handle(Command $console, \Closure $next)
+    {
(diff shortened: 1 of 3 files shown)


calebporzio:
I'm down with that

PhiloNL:
> I'm down with that

Wohaa, that was quick! 2 min to get a PR merged 🚀 that's a new record haha
Thanks! 🙌

PR #5996, opened 2023-07-26
[v3] Add —run-only option to upgrade command

PhiloNL:
Added an option to run only a specific upgrade command, which is helpful if you don't want to run the entire upgrade command again and go through each step. It uses a kebab case to filter.

```shell
php artisan livewire:upgrade --run-only upgrade-introduction

# Will run only the following step:
# Livewire\Features\SupportConsoleCommands\Commands\Upgrade\UpgradeIntroduction
```

I've also moved down the third-party steps to be ran after the official steps are done.


PhiloNL:
I also added a bit of cleanup to make all the commands look better.

Before:
image

After:
image

PR #6001, opened 2023-07-26
Update volt.md

yoeriboven:
Link currently redirects to `https://livewire.laravel.com/docs/docs/actions#skipping-re-renders`. (`docs` twice)


calebporzio:
Thanks!

PR #6003, opened 2023-07-26
[v3] Fix clear is not defined (esm)

PhiloNL:
Trying to write a failing test, when I use wire:poll, it seems to cause:

livewire.esm.js:9527 Uncaught ReferenceError: clear is not defined
    at start (livewire.esm.js:9527:13)
    at livewire.esm.js:9511:3
    at livewire.esm.js:8657:7
    at Array.forEach (<anonymous>)
    at livewire.esm.js:8656:67
    at Array.forEach (<anonymous>)
    at initDirectives (livewire.esm.js:8655:30)
    at livewire.esm.js:8761:7
    at livewire.esm.js:2401:60
    at livewire.esm.js:1999:44

I didn't spot any wire:poll tests (besides legacy tests), so I considered adding a new one. 
Attempting to get the test failing with the error above.

**Update**
The issue only occurs when using the esm version. `let` was missing; I've added the fix to this PR as well.


joshhanley:
@PhiloNL if there are any existing tests in `legacy_tests`, they should already be running in CI. But feel free to move them and update then to the new `Livewire:visit()` style rather than having to write a whole new test.

Moving everything out of `legacy_tests` is a chore we have to do at some point, so if we can manage to do it gradually, then even better 😁

PhiloNL:
@joshhanley ✅ I've removed the new test, so I'm not doing work twice 😄

calebporzio:
Ah shoot, thanks!

PR #6005, opened 2023-07-26
[V3] Add: `reset` Method to Form Object

devajmeireles:
# Problem

```php
{
    public $title = '';

    public $description = '';

    public function save(): void
    {
        $this->validate();

        Todo::create($this->all());
    }
}
```

And then...

```php
{

    public CreateTodoForm $form;

    // ...

    public function add(): void
    {
        $this->form->save();

    }
}
```

# Solution

```php
{
    public $title = '';

    public $description = '';

    public function save(): void
    {
        $this->validate();

        Todo::create($this->all());

        $this->reset(); // all...

    }
}
```

# New Attribute: `Reset`

This PR implements a new attribute dedicated to Form Objects: `#[Reset]`. With this attribute we have the possibility of not allowing the reset of a property, when there is a direct attempt:

```php
use Livewire\Attributes\Form\Reset; // 👈

{
    #[Reset] // 👈
    public $title = '';

    #[Reset(false)] // 👈
    public $description = '';

    public function save(): void
    {
        $this->validate();

        Todo::create($this->all());

        $this->reset(); // ✅ Only the title will be reset...
        
        // $this->reset('description'); // ❌ Throws ResetPropertyNotAllowed exception...
    }
}
```

# Notes

1. The usage of the attribute `#[Reset]` defined as `false` provide us the same behavior as `resetExcept`
2. This PR is a continuation of: https://github.com/livewire/livewire/pull/5967

---

Hope this help,

Thanks. 🚀

### DIFF
--- a/src/Concerns/InteractWithPropertyReset.php
+++ b/src/Concerns/InteractWithPropertyReset.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Livewire\Concerns;
+
+use Livewire\Features\SupportFormObjects\Reset;
+use Livewire\Features\SupportFormObjects\Form;
+
+trait InteractWithPropertyReset
+{
+    public function reset(...$properties)
+    {
+        if ($this instanceof Form) {
+            $properties = Reset::getResettableProperties($this, $properties);
+        } else {
+            if (count($properties) && is_array($properties[0])) {
+                $properties = $properties[0];
+            }
+
+            if (empty($properties)) {
+                $properties = array_keys($this->all());
+            }
+        }
+
+        foreach ($properties as $property) {
+            $freshInstance = $this instanceof Form
+                ? new static($this->getComponent(), $this->getPropertyName())
+                : new static;
+
+            data_set($this, $property, data_get($freshInstance, $property));
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


src/Concerns/Tests/ResetPropertiesTest.php
devajmeireles, line 1: It was a refactoring as part of a readability improvement.

src/Concerns/InteractWithPropertyReset.php
devajmeireles, line 1: It was introduced according to [Caleb comment](https://github.com/livewire/livewire/pull/5967#issuecomment-1650847837)

joshhanley:
@devajmeireles thanks for the PR!

Please remove the `#[reset]` attribute. At this stage we want it to work the way it did before. If you think we should have the reset attribute, that's a new feature that should be in it's own PR.

Also, please keep it as `reset()` instead of `resetProperty()`.

Finally, with the test refactors, generally that should also go into a separate PR, so can you change that too please.

If you get that fixed up on this PR, I will be able to merge it! 

We really appreciate your contributions and this PR is great for everything it does, we just prefer to keep PRs focused on the one change, so in the future try to keep that in mind as it makes it easier for us to review and merge PRs 🙂. Plus PRs that introduce extra things or different changes required Caleb to review, where as he gave the ok for the reset to be moved into a trait, so if this PR is exactly just that, then I am allowed to merge it without needing to run by him again 😁

Thanks!

PR #6007, opened 2023-07-27
[v3] Fix route caching

gehrisandro:
This PR fixes Laravel route caching. This problem has been mentioned in https://github.com/livewire/livewire/discussions/5910 and https://github.com/livewire/livewire/pull/6000

Test implementation taken from here and sightly modified: https://github.com/livewire/livewire/pull/5910

To be honest I am not sure why the routes were created with this `$handle` closure. But in my case it works perfectly fine without.


gehrisandro:
Now I saw, why it has to be a closure 😓 

I will try again.

PR #6010, opened 2023-07-27
[v3] Fix stream variable and ob_flush issue

phuclh:
The 3rd args of the stream method should be $replace, not $append because of its behavior.

Only need ob_flush() if an output buffer is active, so always check it before running ob_flush() to avoid this error:

`ob_flush(): Failed to flush buffer. No buffer to flush`


calebporzio:
Thank you!

PR #6011, opened 2023-07-27
[v2] Fix extension guessing

nuernbergerA:
`guessExtension` detects excel files as `.bin`
`clientExtension` correctly detects `xlsx`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,7 +73,7 @@ jobs:
           RUNNING_IN_CI: true
 
       - name: Setup dusk/chrome
-        run: vendor/bin/dusk-updater update 109 --no-interaction
+        run: vendor/bin/dusk-updater detect --no-interaction
 
       - name: Run Browser tests
         run: vendor/bin/phpunit --testsuite Browser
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks Adrian!

dalbo:
This change has broken things for us with csv files. My computer submits a mime type of `application/vnd.ms-excel` for a csv file and so after this change it is given a .xls extension. While my computer is likely wrong to submit that mime type, I have no control over how my client's computers will be configured. The server-side mime guesser correctly identifies it as csv.

If the server-side component is incorrectly identifying xlsx as Adrian says, then that is what should be addressed, rather than blindly trusting the client. The client value could be used as a fallback in case a valid guess cannot be made on the server. I can see there have been a number of discussions over the years relating to Livewire's mimetype/file extension support, so I recognise it is not a simple thing to solve (and keep everyone happy!)

nuernbergerA:
@dalbo i see your problem.
i think the best way would be to just use `getClientOriginalExtension` and fallback to `guessExtension`. i will open a PR for that

PR #6022, opened 2023-07-27
[v3] Add failing morphdom test for actions in conditional

Jubeki:

See Bug Report https://github.com/livewire/livewire/discussions/5974

Yes

No

It is a failing test

It seems like the alpine/livewire handler is not correctly overwritten with morphdom.

Component for testing:
```php
<?php

namespace App\Livewire;

use Livewire\Component;

class Test extends Component
{
    public bool $enabled = true;

    public function render()
    {
        return <<<HTML
            <div>

                @if(\$enabled)
                    <button wire:click="disable">Disable</button>
                @else
                    <button wire:click="enable">Enable</button>
                @endif

            </div>
        HTML;
    }

    public function enable()
    {
        $this->enabled = true;
    }

    public function disable()
    {
        $this->enabled = false;
    }
}
```

Thanks for contributing! 🙌


calebporzio:
Thanks for this. It was caused because although Alpine picks up attribute changes in the mutation observer, Livewire wasn't hooking into them to cleanup the old event listener and register a new one.

This has been done now. Thanks!

PR #6025, opened 2023-07-27
[v3] Fix Github Actions Deprecations

Jubeki:
Uses newer Versions of `actions/checkout` and `actions/cache` and remove the warnings which occur because of deprecations.

See also https://github.com/actions/cache/blob/main/examples.md#php---composer for the changes made to `::set-output` which has been deprecated.


calebporzio:
Thanks!

PR #5909, opened 2023-07-23
[V3] Add documentation for testing the parameters of dispatched events

julesgraus:
Review the contribution guide first at: https://laravel-livewire.com/docs/2.x/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?
Yes, the discussion is [over here](https://github.com/livewire/livewire/discussions/5908).

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)
Yes. It updates the documentation.

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
I don't think it does

4️⃣ Does it include tests? (Required)
No, because it's documentation only. I've tested the examples in a local project to make sure it matches the documentation.

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.
The current V3 documentation tells you about ```assertDispatched```, but not about the fact that you can use it to assert that an event was dispatched with certain parameters.

Thanks for contributing! 🙌

PR #6044, opened 2023-07-27
Removed a reference to a parameter that is not used in the example.

julesgraus:

Yes. Because the documentation isn't correct. I did not create a discussion because I think it is such small change that it would be overkill

Yes

No

No. Not applicable.

The documentation tells you that the example contains a message and style parameter. But it only contains a message parameter. I just fixed the text to match the example.

Thanks for contributing! 🙌

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -524,7 +524,7 @@ class PostCountBadgeTest extends TestCase
 }
 ```
 
-Sometimes it may come in handy to assert that an event was dispatched with one or more parameters. Let's have a look at a component called `ShowPosts` that dispatches an event called `banner-message` with parameters `message` and `style`:
+Sometimes it may come in handy to assert that an event was dispatched with one or more parameters. Let's have a look at a component called `ShowPosts` that dispatches an event called `banner-message` with a parameter called `message`:
 
 ```php
 <?php


joshhanley:
@julesgraus thanks for the PR! I think it would be better to add the style parameter, which demonstrates the testing of multiple parameters, instead of removing it from the text. Maybe set it to `style: 'success'`? Thoughts?

julesgraus:
@joshhanley Agree. I first created a pull request that did exactly that that was merged in yesterday. 

See [commit](https://github.com/livewire/livewire/pull/5909/commits/43607d9f01f2325a7d3856d1ad7ca60d1137600b) of my previous [PR](https://github.com/livewire/livewire/pull/5909). 

But Caleb removed that in commit https://github.com/livewire/livewire/commit/f0272e2e65d9a6946b258ac0f006764fdc49b57b

joshhanley:
@julesgraus I suspect it's because you changed the example to be too specific and potentially too complicated (using an enum). But maybe just leave it the way it is then, I will get Caleb to have a look and he can decide which he'd prefer 🙂

julesgraus:
Yeah I can see that perspective too. And I can agree with that perspective too. Non the less, we need to fix it one direction or the other.

calebporzio:
Ooops thanks! Exactly what @joshhanley said

PR #6048, opened 2023-07-27
[V3] Improved error message for setting a nested property to an unsupported value

julesgraus:

Yes. See https://github.com/livewire/livewire/discussions/6047

Yes.

No

Yes. I've modified an existing test.

Please see the detailed information over here: https://github.com/livewire/livewire/discussions/6047

Thanks for contributing! 🙌

### DIFF
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Mechanisms\HandleComponents;
 
+use Exception;
 use function Livewire\{ store, trigger, wrap };
 use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
 use Livewire\Exceptions\MethodNotFoundException;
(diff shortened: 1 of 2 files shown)


julesgraus:
Notice: I've intentionally made the tests fail

joshhanley:
@julesgraus thanks for the PR! Not sure why you've set the tests to fail, can you explain that? We prefer PRs tests to pass before we merge 😅

julesgraus:
In version 2 you could do this:

```Livewire::test(YourComponent::class)->set('emails.0', 'test@example.org')```

This exact same code, fails in version 3. But the tests of version 3 that should cover this succeeds / passed, while it should not. So, I've modified the test, to match the situation I've encountered in my app. The situation that made me create this issue and pr.

I suspect we need to create 2 PR's. One that simply improves the error message like I did in this pull request. And one that resolves the actual issue. But I don't know IF it is an actual issue, or that I need to upgrade my code. In the latter part, we may need to address this issue when running ```livewire:upgrade``` I guess.

joshhanley:
@julesgraus ah yep, understood. Can you change the existing test to back the way it was, and create a new test that fails covering the break, so we can investigate it? We prefer not to change existing tests where possible. Thanks!

julesgraus:
Yes. Will do. Got to get some sleep right now since it is midnight already. Will do it when I get back from work tomorrow

joshhanley:
@julesgraus no worries! Thanks, much appreciated 🙂 have marked this as draft for now

calebporzio:
Thank you for this, however, I'm going to close it in the interest of keeping this part of the codebase a little cleaner (rather than passing that extra parameter all around) - again, thank you

PR #6050, opened 2023-07-27
[V3] Add: `reset` Method to Form Object

devajmeireles:
# Problem

```php
{
    public $title = '';

    public $description = '';

    public function save(): void
    {
        $this->validate();

        Todo::create($this->all());
    }
}
```

And then...

```php
{

    public CreateTodoForm $form;

    // ...

    public function add(): void
    {
        $this->form->save();

    }
}
```

# Solution

```php
{
    public $title = '';

    public $description = '';

    public function save(): void
    {
        $this->validate();

        Todo::create($this->all());

        $this->reset(); // all...

    }
}
```

---

Hope this help,

Thanks. 🚀


devajmeireles:
Hey, @joshhanley .

Thanks for your feedback. I closed PR https://github.com/livewire/livewire/pull/6005 because it looks like there was some bug with GitHub. I sent the commit and the PR didn't receive it.

Your feedback was valuable. As these are my first contributions to Livewire, I understand your work format. **Thank you very much for your feedback, it makes a difference.**

About your mentions, totally agree. I did as you asked, there's only one issue I wanted to flag. [Caleb's idea](https://github.com/livewire/livewire/pull/5967#issuecomment-1650847837) was to abstract the reset into a trait to reuse in two places, so I dubbed `reset` to `resetProperty` only inside InteractsWithProperties trait because otherwise there would be a name conflict.

Please feel free to review the PR from now on and make sure this is fine with you now. Let me know if you want more adaptations.

calebporzio:
Thank you so much for the well formed PR! (docs, tests, code... what more can a maintainer ask for!)

PR #6053, opened 2023-07-28
[V3] Add --pest command to Artisan commands

skegel13:
This PR adds a `--pest` flag to the Artisan commands when creating a Livewire component to also create a Pest PHP test.

```shell
artisan make:livewire create-post --pest
```

It also handles copying and deleting the Pest files when the respective commands are run.

A new `livewire.pest.stub` was created for the Pest tests and it also gets published when publishing the stubs.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -9,6 +9,8 @@ artisan make:livewire create-post --test
 
 In addition to generating the component files themselves, the above command will generate the following test file `tests/Feature/Livewire/CreatePostTest.php`:
 
+If you would like to create a [Pest PHP](https://pestphp.com/) test, you may provide the `--pest` option to the make:livewire command:
+
 ```php
 <?php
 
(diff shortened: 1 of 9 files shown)


joshhanley:
@skegel13 thanks for the PR!

I don't think the other commands need the `--pest` flag, do you?
To me, if another command like copy already has a `--test` flag, then it should copy the test regardless of if it is a PHPUnit or Pest test.

So my thoughts would be that this is only needed on the make command. If you agree, can you update the PR?

Thanks!

skegel13:
@joshhanley I can make the update if you think that makes sense. I just figured if someone created a component with a pest test, they might use the other commands with the pest command too. Personally, I've never used to other commands so I'm fine with the change though.

joshhanley:
@skegel13 yeah if you can, that'd be great! Keep it simple for now, and if there are issues, we can always add the pest flag to the other commands later 🙂

skegel13:
Changes have been made, let me know if there's anything else.

calebporzio:
Thank you @skegel13 !!

PR #6059, opened 2023-07-28
Use array for only method

christophrumpel:
This PR fixes a code example from the docs where the array syntax is missing for the `only` method on the form page.


calebporzio:
Thanks @christophrumpel!!

PR #6063, opened 2023-07-28
Add missing closing pho tag in Volt documentation

einenlum:


joshhanley:
@Einenlum thanks! Not sure why you closed, but I reopened and also added the opening PHP tags 🙂

PR #6072, opened 2023-07-28
[V3] Failing test for setting a nested value in an uninitialized array.

julesgraus:

Yes. Please see: https://github.com/livewire/livewire/discussions/6047

Yes.

No

Yes

See https://github.com/livewire/livewire/discussions/6047 for the details.

Thanks for contributing! 🙌


calebporzio:
Yeah I see this... in this case it does feel slightly weird to me to set a deep property of a property that has never been initialized as an array.

If you don't want to initialize it as an array I recommend adding an `array` type to the property and I tested that that would fix it:

```php
public array $yourProperty;
```

julesgraus:
Ok. Since this behavour did work in v2 and not in v3. It might be worth noting this in the upgrade guide right?

joshhanley:
@julesgraus yep! Any chance you could submit a PR with a note in the upgrade guide? Thanks!

PR #6076, opened 2023-07-28
[V3] Docs: fix `composer require` commands

PerryvanderMeer:
It seems that Windows machines have difficulties with the current `composer require` commands, because they contain a `^`.
With the current commands, it causes errors while installing, like:

afbeelding

Could be resolved when the version parameter is [enclosed by double quotes](https://stackoverflow.com/questions/72877159/using-composer-require-without-quotes/72904848#72904848):

afbeelding

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -3,7 +3,7 @@ Livewire is a Laravel package, so you will need to have a Laravel application up
 To install Livewire, open your terminal and navigate to your Laravel application directory, then run the following command:
 
 ```shell
-composer require livewire/livewire:^3.0@beta
+composer require livewire/livewire "^3.0@beta"
 ```
 
 That's it — really. If you want more customization options, keep reading. Otherwise, you can jump right into using Livewire.
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -15,7 +15,7 @@ Before we start, make sure you have the following installed:
 From the root directory of your Laravel app, run the following [Composer](https://getcomposer.org/) command:
 
 ```shell
-composer require livewire/livewire:^3.0@beta
+composer require livewire/livewire "^3.0@beta"
 ```
 
 ## Create a Livewire component
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -26,13 +26,13 @@ Livewire now requires that your application is running on PHP version 8.1 or gre
 Run the following composer command to upgrade your application's Livewire dependency from version 2 to 3:
 
 ```shell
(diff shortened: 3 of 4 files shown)


joshhanley:
@PerryvanderMeer thanks! You could also just wrap the whole livewire string in quotes `composer require "livewire/livewire:^3.0-beta"`

PerryvanderMeer:
@joshhanley Definitely! Matter of taste. It's up to you! 😃

joshhanley:
@PerryvanderMeer yep! If it's ok (and it actually works - have you got windows to test on?) I'd say change it, as I try to match the style Caleb originally has as much as possible.

PerryvanderMeer:
@joshhanley Hmm nope, for some reason that isn't doing the job.

afbeelding

joshhanley:
@PerryvanderMeer no worries! Thanks for testing, will merge as it is then 🙂

PerryvanderMeer:
@joshhanley Hmm seems to be a bigger problem in PowerShell (see: https://getcomposer.org/doc/articles/versions.md#caret-version-range-)
The above command has some issues in PS as well.

The proposed solution would be to use: `composer require livewire/livewire:^^^^3.0@beta`
Looks a bit silly, but works in both command line as PowerShell.

joshhanley:
@PerryvanderMeer ah yep, haha yeah doesn't look great. Let's just stick with the quotes you've added, which I'm guessing works find in command prompt yeah? Does using quotes work in powershell?

PerryvanderMeer:
@joshhanley Tested some things out. PowerShell is the bottleneck in this. Only the last one works in both PowerShell as CMD.

PowerShell:
🔴 `composer require livewire/livewire:^3.0@beta`
🔴 `composer require "livewire/livewire:^3.0@beta"`  
🔴 `composer require livewire/livewire "^3.0@beta"` 
🟢 `composer require livewire/livewire:^^^^3.0@beta`  

CMD:
🔴 `composer require livewire/livewire:^3.0@beta`
🟢 `composer require "livewire/livewire:^3.0@beta"`  
🟢 `composer require livewire/livewire "^3.0@beta"` 
🟢 `composer require livewire/livewire:^^^^3.0@beta`

joshhanley:
@PerryvanderMeer ok thanks for testing, let's just leave it the way you have it in the PR for now. It should get most people running if they are using command prompt (better than before).

Plus `composer require livewire/livewire:^^^^3.0@beta` doesn't work on mac 😂

image

PR #6078, opened 2023-07-28
[v3] Fix layout path in docs

austenc:
Congrats on v3 Launch! 🎉

Noticed that the layout paths in the docs are not what Livewire v3 expects as the default full page component layout path.

Cheers!


joshhanley:
@austenc thanks!

PR #6083, opened 2023-07-29
[v2] Fix for imageSize method on TemporaryUploadedFile to support dimensions validation rule

jordanhavard:
This PR is an extension of #5896 by @gdebrauwer which fixes the `imageSize` method name to be `dimensions` which was amended by Taylor here 👉  [https://github.com/laravel/framework/pull/46912](https://github.com/laravel/framework/pull/46912) before being merged into L10

### DIFF
--- a/src/TemporaryUploadedFile.php
+++ b/src/TemporaryUploadedFile.php
@@ -74,7 +74,7 @@ class TemporaryUploadedFile extends UploadedFile
         return $this->extractOriginalNameFromFilePath($this->path);
     }
 
-    public function imageSize(): ?array
+    public function dimensions(): ?array
     {
         stream_copy_to_stream($this->storage->readStream($this->path), $tmpFile = tmpfile());
 
--- a/tests/Unit/ValidationTest.php
+++ b/tests/Unit/ValidationTest.php
@@ -4,15 +4,10 @@ namespace Tests\Unit;
 
 use Illuminate\Contracts\Validation\Rule;
 use Illuminate\Contracts\Validation\Validator as ValidatorContract;
-use Illuminate\Http\UploadedFile;
-use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\ViewErrorBag;
 use Livewire\Component;
-use Livewire\Features\SupportFileUploads;
 use Livewire\Livewire;
-use Livewire\TemporaryUploadedFile;
-use Livewire\WithFileUploads;
 
 use function PHPUnit\Framework\assertTrue;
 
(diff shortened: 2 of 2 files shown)


jordanhavard:
Closes #6085

joshhanley:
@jordanhavard thanks for this! We currently support Laravel 7 to 10 with V2, do you know whether this change will work on anything other than Laravel 10?

jordanhavard:
> @jordanhavard thanks for this! We currently support Laravel 7 to 10 with V2, do you know whether this change will work on anything other than Laravel 10?

Hey 👋 

Actually I'm not 100% however would be happy to dig into this to find out more. 

I'm using L10 with DO spaces and was running into this issue where dimensions were failing validation and saw the parent PR was merged as a proposed fix.

However after updating to the version you tagged Josh, it was still failing and upon looking into the Laravel framework PR, I saw Taylor changed the base method name before merging which caused this parents PR to be redundant.

(Taking a guess here) but would assume dimension validation would fail as the base functional (the one this & the parent PR are overriding) doesn't exist in the Laravel framework < 10

(Worth noting that this functionality was broken anyway before these PRs). 

I'm under the impression that for this to work in previous versions of Laravel, we'd need to make PRs to those versions of Laravel to create a dimension size method so it can be overridden by Livewire. 

Hopefully this helps although let me know if I can help with anything 🙏

Appreciate the work everyone does here 🫶

joshhanley:
@jordanhavard ah no worries, if it was only included in Laravel 10, then it's fine to leave the way it is. Laravel won't accept PRs to older versions.

Any chance you could add a test for this, so we can ensure it works this time?

Thanks!

jordanhavard:
@joshhanley I've added a test for image dimension validation; let me know if you'd like any changes 🙏 
Once happy, I can make the changes in the V3 branch too

joshhanley:
@jordanhavard great! Thanks for that. Looks good. I've trigged the tests to run, so if it all passes, then I'd say copy across to the V3 PR 🙂

joshhanley:
@jordanhavard I just ran the test locally, without your fix in this PR, and the test still passed. Can you confirm if the same happens for you?

jordanhavard:
> @jordanhavard I just ran the test locally, without your fix in this PR, and the test still passed. Can you confirm if the same happens for you?

It would still work as the underlying method would be called in the Laravel framework - as this PR fixes a typo. The parent PR allows the base method to be overridden, which streams the image to a temp file for validation.

joshhanley:
@jordanhavard ah ok thanks, sorry I misunderstood - maybe the test wasn't needed then 😅

jordanhavard:
> @jordanhavard ah ok thanks, sorry I misunderstood - maybe the test wasn't needed then 😅

Aboslutely no need to appologise 🙏  I appreciate your input on the matter 

With the above in mind, Is this something that is able to be merged in and pushed out with the next release? At the moment, I'm extending (pretty much) every file under the sun to get image dimension validation to work at the moment 😅

joshhanley:
@jordanhavard thanks! I decided to remove the test as it doesn't actually help for the original use case (streamed image size). Yep, it will be in the next release 🙂

PR #6084, opened 2023-07-29
[v3] Fix for imageSize method on TemporaryUploadedFile to support dimensions validation rule

jordanhavard:
same as #6083


joshhanley:
@jordanhavard thanks!

PR #6087, opened 2023-07-29
[V3] Add failing test for component with wire:model and custom Alpine directive

nleibbrandt:
This is follow up for https://github.com/livewire/livewire/pull/5950

Yes

No

Failing test

For some reason the custom Alpine directive gets called twice after refreshing and cloning livewire component.
The first init call seems to do everything right, but second resets the state.
If you add wire:ignore to Alpine directive, things work as expected: 

Bildschirmfoto 2023-07-29 um 11 45 05

Thanks for Livewire! 🙌


calebporzio:
Wow, that was a big one. Thanks for the report @nleibbrandt !

PR #6093, opened 2023-07-29
Docs Fix - Code example bug about property visibility

christophrumpel:
This PR fixes a code example about form objects. I'm pretty sure this model property needs to be `protected`, because when `public` it will be used for the `$this->all()` method which tries to fill a column on the model that does not exist.

CleanShot 2023-07-29 at 17 26 19@2x

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -288,7 +288,7 @@ use App\Models\Post;
 
 class PostForm extends Form
 {
-    public ?Post $post;
+    protected ?Post $post;
 
     #[Rule('required|min:5')]
     public $title = '';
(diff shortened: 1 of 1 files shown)


calebporzio:
Good catch Christoph! It actually has to be public to persist between requests. Instead `$this->only()` should be used (just added it)

christophrumpel:
Ah got it 👍 thanks

christophrumpel:
@calebporzio I believe the `only` method is not available on the form object.

joshhanley:
@christophrumpel thanks! Hmm, I thought it was available on the component too? If not, it possibly should be 😆

PR #6102, opened 2023-07-30
[v3] Add failing test for input reset

krekas:

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportFormObjects/BrowserTest.php
+++ b/src/Features/SupportFormObjects/BrowserTest.php
@@ -12,6 +12,8 @@ class BrowserTest extends \Tests\BrowserTestCase
     {
         Livewire::visit(Page::class)
             ->assertSee('Page')
+            ->value('input[id="title"]', 'This is a title')
+            ->assertInputValue('#title', 'This is a title')
             ->assertValue('input[id="title"]', 'This is a title')
 
             ->waitForLivewire()->click('@save')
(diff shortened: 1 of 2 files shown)


devajmeireles:
As per my review, the test `src/Features/SupportFormObjects/BrowserTest.php` doesn't make sense.

- You are trying to reset a property that initially contains the value as `This is a title`
- After resetting, you expect the value to be empty, but the correct thing is that the value is `This is a title`, [since the property reset consist on reset the property value to the initial state.](https://livewire.laravel.com/docs/forms#resetting-form-fields)

krekas:
@devajmeireles yes good point. Updated test

devajmeireles:
I found the problem. This PR https://github.com/livewire/livewire/pull/6118 will solve this. As per my review I think you need to abstract your PR for a new one only related with lifecycle hooks to the properties.

joshhanley:
@krekas thanks for the PR!

I'm not sure I follow exactly, one test uses `->reset()` and the other test doesn't, but this is meant to be testing `->reset()` right? Also he first test, in `SupportFormObjects\BrowserTest`, doesn't actually have a FormObject in it so either you need to add one or that's the wrong place for the test.

The first test is failing because you're using `->input()` which Livewire doesn't support. Try changing that to `->type('@title', 'This is a title')` and it will probably pass.

Can you update and let me know when done?

Also can you provide me a component class and view with instructions that I can use locally to test your issue?

Thanks!

PR #6104, opened 2023-07-30
[v3] assign classname from laravel morph map

aminetiyal:

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first? ⚠️ 
It did not open any discussion but it is supposed to be according to [docs: properties expose system information to the browser](https://livewire.laravel.com/docs/properties#properties-expose-system-information-to-the-browser)
also there are some old discussions about it on V2, #4074 #3812 

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed) ✅ 

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out. ❌ 

4️⃣ Does it include tests? (Required) ✅ 

Thanks for contributing! 🙌

## Context
According to [docs: properties expose system information to the browser](https://livewire.laravel.com/docs/properties#properties-expose-system-information-to-the-browser) class names exposed to JS through snapshot `component.snapshot.data.model[1].class` property can be overridden by Laravel default `Relation::morphMap()` which is not working with Livewire V3,

Meanwhile, it is still possible to do so if the `legacy_model_binding` is turned on in the config which will change the behavior of livewire toward model binding according to the [docs: eloquent model binding](https://livewire.laravel.com/docs/upgrading#eloquent-model-binding)

### DIFF
--- a/src/Features/SupportModels/ModelSynth.php
+++ b/src/Features/SupportModels/ModelSynth.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features\SupportModels;
 
+use Illuminate\Database\Eloquent\Relations\Relation;
 use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
 use Illuminate\Queue\SerializesAndRestoresModelIdentifiers;
 use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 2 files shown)


joshhanley:
@aminetiyal thanks for the PR! I agreeed this should be supported.

Can you explain why the existing tests needed refactoring/ why you couldn't use the Sushi model for your tests too?

Thanks!

aminetiyal:
> @aminetiyal thanks for the PR! I agree this should be supported.
> 
> Can you explain why the existing tests needed refactoring/ why you couldn't use the Sushi model for your tests too?
> 
> Thanks!

@joshhanley The thought won't work with Sushi, forgot that is still a model instance :D rolled back with Sushi

joshhanley:
@aminetiyal thanks! 🙂

joshhanley:
@rompetomp That's how I'd expect it to work, a morph map would be required if using enforce. How would you expect it to work?

rompetomp:
enforceMorphMap is only for morphOne or morphMany relations. In the example I sent, Team is a direct (belongsToMany) relation to User.

joshhanley:
@rompetomp right, I think I understand.

The reason this is implemented is because people didn't want the full class names of their models to be sent to the front end.

So it makes use of Laravels morph map (if available) to provide a string instead. 

But if your using enforce then you need to make sure you have a morph map for any model classes that may be sent to the front end, such as your team model.

I would just be added them to your morph map, even though it's not necessarily used in a morph relationship.

Hope this helps!

PR #6110, opened 2023-07-30
[V3] Remove excess #[On('echo:orders,OrderShipped')] from events docs

machacekmartin:
This might be a tiny detail, but since `notifyShipped` is registered in `getListeners` method, it should not be registered again in `#[On]` attribute.


joshhanley:
@machacekmartin thanks!

PR #6118, opened 2023-07-31
[v3] Fix reset all

gehrisandro:
In v2 when calling `$this->reset();` without any parameters, all properties were reset, but in v3 none of the properties are reset.

I guess this is a bug, because there is already a test file with a failing test: [src/Concerns/Tests/ResetPropertiesTest.php](https://github.com/livewire/livewire/blob/e3a1a5a70aa037276b98e8d69f19ff194240cb89/src/Concerns/Tests/ResetPropertiesTest.php#L25)
But this test file is not executed when running the PHPUnit test suite nor is it run in the testing pipeline.

This PR brings this behaviour back but does not update the PHPUnit configuration.

### DIFF
--- a/src/Concerns/InteractsWithProperties.php
+++ b/src/Concerns/InteractsWithProperties.php
@@ -44,6 +44,11 @@ trait InteractsWithProperties
             ? $properties[0]
             : $properties;
 
+        // Reset all
+        if (empty($properties)) {
+            $properties = array_keys($this->all());
+        }
+
         $freshInstance = new static;
 
         foreach ($properties as $property) {


devajmeireles:
> In v2 when calling `$this->reset();` without any parameters, all properties were reset, but in v3 none of the properties are reset.
> 
> I guess this is a bug, because there is already a test file with a failing test: [src/Concerns/Tests/ResetPropertiesTest.php](https://github.com/livewire/livewire/blob/e3a1a5a70aa037276b98e8d69f19ff194240cb89/src/Concerns/Tests/ResetPropertiesTest.php#L25) But this test file is not executed when running the PHPUnit test suite nor is it run in the testing pipeline.
> 
> This PR brings this behaviour back but does not update the PHPUnit configuration.

I recently worked on resetting properties in this PR: https://github.com/livewire/livewire/pull/6050, and you can see that the tests passed. Maybe someone modified something? I don't know. There's a commit from Caleb in my PR, maybe it had something to do with that. 😅

Thanks for your PR.

joshhanley:
@gehrisandro thanks! Yeah, all unit test files should now have the `UnitTest.php` suffix but some were missing that. I fixed that in PR #6134, which you can see has a failing test for reset properties that passes now it has been merged into this PR.

PR #6127, opened 2023-07-31
[v3] Fix multiple #[On] an #[Rule]

nuernbergerA:


joshhanley:
@nuernbergerA thanks for this! Seems strange it doesn't inherit it 😕 does that mean it isn't needed on the parent attribute classes then?

nuernbergerA:
I will check if [`class_alias()`](https://www.php.net/manual/en/function.class-alias.php) work's for attributes. If so I will open a PR

nuernbergerA:
@joshhanley see https://github.com/livewire/livewire/pull/6148

PR #6135, opened 2023-08-01
[V3] Failing test for layout stacks

joshhanley:
Pushing to layout stacks worked in V2 but not in V3, this PR adds a failing test for this scenario.

See #5993, #6130, #6131, #6173.


nuernbergerA:
@joshhanley could you also add one for `@pushOnce`?

joshhanley:
@nuernbergerA yeah, although I don't think it needs it, as the underlying code that will enable the stacks to run will cover pushOnce too 🙂

calebporzio:
Fixed!

PR #6141, opened 2023-08-01
[v3] withDefault model not in DB falling test

fripig:
turn on legacy_model_binding then withDefault model got fail 

See #5979


calebporzio:
Thank you for submitting this - however, I'm going to close this. Please re-submit a PR with a distinct test case rather adding code to existing classes already under test. Make sense? Like ideally a new test method called: "can_use_with_default_on_model_relations" or something like that.

Thank you for your efforts

PR #6148, opened 2023-08-01
[v3] `#[Attribute]` aliases

nuernbergerA:
Unfortunately php attributes cant inherit attribute declarations from parent class.

we can simple use `class_alias` to register all attributes under the `\Livewire\Attibutes\` namespace.

if you don't like the decoupling from defining an attribute and registering the alias in the ServiceProvider, we can easily move the `class_alias` inside the corresponding attribute class itself.

bonus with this pattern we can also register `aliases` e.g. TestCase


calebporzio:
Going to close this in favor of: #6389 

Thanks @nuernbergerA!

PR #6154, opened 2023-08-02
[v3] Small fix on understanding-nesting.md

SRWieZ:
While reading the documentation, I encountered this little error.

Neat documentation otherwise!


calebporzio:
Ooops good call - thanks!

PR #6155, opened 2023-08-02
[v3] Added additional information on #[Modelable] attribute 

SRWieZ:
When reading I asked myself if there was any way to have multiple `#[Modelable]` attributes.
I checked `SupportWireModelingNestedComponents` and find out the answer, so it seemed a good idea to share it with others.


calebporzio:
Thanks for the extra info. Maybe we'll have named modelables some time - maybe version 4 of Livewire, but for now this is helpful context for folks

PR #6160, opened 2023-08-02
[v3] Add except method to form

donkooijman:
When using the same Form class for both storing and updating, one might want to get all properties except for the 'Model' property. Currently, this requires utilizing the `->only()` method and individually listing all the other properties.

I propose to add the `->except()` method to make this easier. 

Let me know what you think, happy to make any changes.


calebporzio:
Thanks!

PR #6166, opened 2023-08-03
Fail delete child test

lonnylot:
This _should_ be a failing test for https://github.com/livewire/livewire/discussions/5935

If you try to delete a child from an action within a child you receive:

Screenshot 2023-08-02 at 10 11 33 PM

I say _should_ because the @alpinejs/history package is [still](https://github.com/livewire/livewire/pull/5790#issuecomment-1645020052) not published so I cannot run the tests locally


joshhanley:
@lonnylot Thanks!

You don't need to install npm packages to run tests. At the moment we actually don't recommend it as we're having to manually link alpine packages.

Have you tried to run the test locally?

lonnylot:
@joshhanley ah ok. Tried running, failed, then tried following readme. I'll get a proper failed test tomorrow. 🙏

joshhanley:
@lonnylot all good thanks, Caleb has already started to look at it and has updated your test for you 🙂

lonnylot:
> @lonnylot all good thanks, Caleb has already started to look at it and has updated your test for you 🙂

Great thanks guys. Let me know if I can do anything else to help

calebporzio:
I think this has been fixed by my other recent morph changes. Passing now.

PR #6170, opened 2023-08-03
[v3] Fix typo in docs for creating form

donkooijman:
I think this is a typo in the docs. The command is used to create a PostForm class.


calebporzio:
Ah thanks!

PR #6171, opened 2023-08-03
[V3] Support Defer Loading

nicolagianelli:
Refers to #6153 

Yes, defer-loading

No

Yes

Livewire with the lazy param can load components when they are in the viewport but for us and other people can be useful to start loading a component even when outside of the viewport, for example in a scrollable page, so when the user scrolls down they can find the components already loaded.
I named this param "defer" following the HTML defer attribute used for scripts.


DanielSpravtsev:
I agree that Livewire need defer loading besides lazy loading, but now it's unclear what each option making because of similar naming. Maybe add something referencing to viewport or intersect (like in Alpine)? 

What do you think @calebporzio @joshhanley?

joshhanley:
@nicolagianelli I think it would be best to wait to see Caleb's thoughts on it before making changes, but my opinion is that this should be the default for `lazy` and for intersect using `lazy.intersect` or something like that. Then it is just one feature class with a toggle instead of two classes (defer and lazy) that are basically the same.

nicolagianelli:
> @nicolagianelli I think it would be best to wait to see Caleb's thoughts on it before making changes, but my opinion is that this should be the default for `lazy` and for intersect using `lazy.intersect` or something like that. Then it is just one feature class with a toggle instead of two classes (defer and lazy) that are basically the same.



> 

Yes.. using only a lazy param is more simple and better for the structure so for me it's good.
I would use lazy.viewport or lazy.intersect and lazy.load or lazy.init for the names, and the lazy without any modifier would be the viewport one like now because other js frameworks like nuxt or with images the lazy param is usually used to load the element when it is in the viewport.
Anyway I don't mind too much and I can make these changes to use these names and only 1 class when Caleb decides what naming convention he prefers 😁

calebporzio:
Ok, thank you for this PR. I agree that it should be a feature.

Here's my ideal syntax to opt-into non-viewport-based loading:

```html
<livewire:counter lazy="on-load" />
```

Here are my notes for what to do to this PR before I merge it:

* Instead of a seperate "SupportDeferredLoading" folder, just add this feature and it's test to the "SupportLazyLoading" feature folder.
* Also we probably only need a single test case for this syntax
* Update the docs portion

If these things are done, I'd be happy to merge this. Thanks @nicolagianelli!

nicolagianelli:
@calebporzio 
I did the requested changes:
- changed defer to lazy="on-load"
- only a SupportLazyLoading folder
- 2 new tests (1 to try the lazy="on-load" and 1 to try the lazy without "on-load" outside the viewport)
- updated the docs

edalzell:
Looks like there are conflicts @nicolagianelli, can you please resolve so that it's easier for Caleb to merge?

calebporzio:
Thanks @nicolagianelli !!

PR #6177, opened 2023-08-03
[V3] Add tests for modelable nested component

quentinhnrt:
Modelable nested components throw an error when bound to an array / form object

This PR adds tests for modelable nested components bound to an array or a form object

=> Related discussion : https://github.com/livewire/livewire/discussions/6106


julesgraus:
I believe this is a duplicate I did post last week: 
https://github.com/livewire/livewire/pull/6072

I already created a failing test

joshhanley:
@quentinhnrt thanks for the PR!

@julesgraus nah this is a different issue, this one has to do with using `modelable` between parent and child components 🙂

calebporzio:
Ok, I see the problem. Here's a little writeup describing it:

The problem:

```php
// Parent
public $foo = '';

// Child
#[Modelable]
public $bar;
```


This works by binding the two properties on the front-end:
```html
<div>
    <!-- Child component -->
    <div wire:id="..." x-model="$wire.$parent.foo" x-modelable="$wire.bar">
    </div>
</div>
```

When an update is made to either "foo" or "bar" in JS, then one component commits it's changes, the other component is bundled in the request and they both have those updates in the payload:

```js
{ // Parent
    updates: { foo: 'changed' },
}

{ // Child
    updates: { bar: 'changed' },
}
```

This works great. Here's the problem:

If the parent changes `$foo` during it's lifecycle, that change will be overwritten by the child update. For example:

```php
class Parent extends Component {
    public $foo = '';

    public function save() {
        $this->foo = 'yo';
    }
}
```

After this parent is finished, the child get's rendered. `$bar` is set to "yo" at first because it's passed in from the parent, then the `updates` are processed from the payload, setting it back to `changed`.

What should be done here?

A)
* If, after an update, detect if the parent property is "dirty" (has been changed by something other than an "update").
* Then, remove the `updates` entry from the child's payload

B) 
* If, after an update, detect if the parent property is "dirty" (has been changed by something other than an "update").
* Reset the property after the "updated" hook has been called back to the dirty value...

C)
* Simply check that the parent has an "updates" for the new value, and unconditionally remove the "updates" entry from the child?

D)
* Check that the parent is part of the request at all and remove any "updates" for the modelable field from the child.

Problems:

* A & B imply that parent/child requests will always be bundled together. That isn't the case. The child should be bound to the parent in terms of request bundling, but not the other way around: the child can make a request on it's own...
* C & D are good except that the child might have lifecycle hooks that it should run when the modelable field is updated? and removing those updates would be an issue

We could try a more complex approach:
* Before the child is about to update... if the parent has an "update" hook, set the child updated payload bit to the newly passed in value from the parent...? idk

PR #6179, opened 2023-08-03
Update testing.md to change method for testing URL Params

favouraka:
Change `withURLparams()` to `withQueryParams()` in some sections of the testing page in the documentation.


calebporzio:
Ooops good call, instead I ended up adding "withUrlParams()" to core. Thanks!

PR #6180, opened 2023-08-03
[v3] add failing test for nested reactive property

gehrisandro:
This PR adds a failing test for multi level nesting of reactive properties.

The weird thing here is, that it updates on the first increment, but not on the second call. But only if I add this `sleep` call. Otherwise the test would pass.

And if we are going to use blade files instead of an inline component, it already fails on the first increment. (I have not added a test for this)

So it looks like multi level nesting is in some way implemented, but does not properly work in all cases and / or has some timing issue.

This is related to https://github.com/livewire/livewire/discussions/6125


joshhanley:
@gehrisandro if this failing test PR is ready, can you mark it as ready so we can investigate? Thanks!

calebporzio:
All fixed now! thanks for the test

PR #6188, opened 2023-08-04
Add failing test for `null` values in arrays in query string

archilex:
Currently when using arrays in the query string, when setting one of the values in the array, all the other values get populated with `null`. This is related to #5932, but is still unsolved when dealing with arrays in the query string. This is also being discussed as an issue in Filament: https://github.com/filamentphp/filament/issues/7129

Thanks for contributing! 🙌


archilex:
I found a work around using assertScript() to match the query string and make this fail in a more eloquent way. Now the test fails with: 

```bash
--- Expected
+++ Actual
@@ @@
-'tableFilters%5Bfilter_1%5D%5Bvalue%5D=test'
+'tableFilters%5Bfilter_1%5D%5Bvalue%5D=test&tableFilters%5Bfilter_2%5D%5Bvalue%5D=null&tableFilters%5Bfilter_3%5D%5Bvalue%5D=null'
```

joshhanley:
@archilex thanks for the PR! Not sure I follow what your second comment means, do you have a fix?

archilex:
> @archilex thanks for the PR! Not sure I follow what your second comment means, do you have a fix?

If you look at the original commit (and the comment there), I used `->assertQueryStringMissing('tableFilters')` just so the test would fail. I did this because there isn't a way to test arrays in query strings using `->assertQueryStringMissing()`, but I wanted the test to fail so you could see the problem. The problem with that is that the test would never pass either. 

However, with the latest commit I switched to `assertScript()` so now the test fails as it should, but once you implement a fix, it will also pass.

Hope that makes more sense.

joshhanley:
@archilex ah right! Makes sense. We don't look at the commits, just files changed 🙂 thanks!

archilex:
@joshhanley Yeah my bad...I should have put those original comments in PR here.

calebporzio:
Hmmm I see why you wouldn't want this - what do you think @joshhanley - is this how query strings should behave?

Maybe null is represented by the removal of the query string entry, but an empty string is represented as an empty query string entry... hmmm

calebporzio:
The tough part is that this is an array of filters - and that difference matters here. If those entries weren't in the query string, how would we know to not "remove" them when the page refreshes if only some of the values are in the query string.

This is a tough one.

What would one do in V2 in this scenario, would they have the same problem?

archilex:
@calebporzio v2 handled this as expected in that only the keys with values were populated in the query string. I modified this test for v2 just to be sure and the test does successfully pass. This is the output in v2: `http://127.0.0.1:8001/livewire-dusk/Tests%5CBrowser%5CQueryString%5CComponent?tableFilters[filter_1][value]=test`

archilex:
And regarding this:

> The tough part is that this is an array of filters - and that difference matters here. If those entries weren't in the query string, how would we know to not "remove" them when the page refreshes if only some of the values are in the query string.


I'm not sure if I'm completely following, but in the case of Filament and LW v2 this was all handled as expected in that if the filter didn't exist in the query string when the page was loaded, that filter wasn't applied to the query.

calebporzio:
Yeah, `filter: true` could work as an API to "opt-in" to this behavior.

The struggle with this is the implementation.

In V2, the entire query string feature is almost entirely contained in the backend. Meaning PHP is in charge of constructing query strings and URLs and such and the JS portion just changes the URL based on what PHP tells it to.

We've agreed that doing that in PHP is a bad idea and was the cause of lots of issues in Livewire 2, so we moved all the query-string related behavior to the front-end.

That said, I don't think this behavior is impossible to build, but it would take a rework of the JavaScript portion.

Currently, we use pretty simple JS logic to construct a query string based on a piece of data and visa versa.

We would need to update that logic to instead be intelligent about how it constructs and references the query string against data. Like we would need to recursively check arrays and objects to see if each key's value has differed from the original value and only act on those differing portions.

It might not be as complex as I'm imagining, but just putting it out there for anyone that wants to try implementing it.

I don't think I'm going to prioritize doing this work myself for the 3.0 tag as it doesn't break behavior. (it just looks awful lol)

danharrin:
Subscribing so I remember to revert filamentphp/filament#7999 when this is fixed

calebporzio:
K, this has been fixed - Livewire will now remove nulls in array from the query string by default.

danharrin:
@archilex I think nulls are only handled here, but the PR you had made previously also handles booleans. Do you believe the fix still needs to be in place for those?

archilex:
@danharrin I just tested this and the [fix](https://github.com/filamentphp/filament/pull/7999) we implemented in Filament still needs to be applied to handle booleans. I'll try to submit a separate failing test to livewire.

PR #6190, opened 2023-08-04
[v3] Fix ArraySynth hydrate loop __rm__

donkooijman:
I'm not sure why value is set to \_\_rm\_\_. But I think this fixes the issue for trying to loop string \_\_rm\_\_ mentioned in https://github.com/livewire/livewire/discussions/6146


joshhanley:
@donkooijman Thanks for the PR! Could you please add a test for this? Thanks!

donkooijman:
@joshhanley I've added a browser test which mimics the behavior of my component. Without the change I've implemented, this component encounters an error, causing the test to fail.

I'm not entirely certain about the best location for it within the project structure or the most appropriate name. I'm open to your suggestions and will make any necessary adjustments based on your feedback.

nuernbergerA:
ok this helped me a lot to figure out whats the real problem is.

joshhanley:
@donkooijman thanks! Test locations look ok to me, Caleb can always change if he'd prefer somewhere else 🙂

nuernbergerA:
i think this will represent the bug better, not only the `string was given` error will be triggered but also if the value is actually removed
```php
public function test_update_from_js_with_rm(): void
{
    app('livewire')->component('example', Example::class);

    $raw = str(app('livewire')->mount(Example::class))->match('/wire\:snapshot="([^"]*)"/')->toString();
    $initialSnapshot = json_decode(htmlspecialchars_decode($raw), true);

    [$snapshot, $effects] = app(HandleComponents::class)->update($initialSnapshot, ['subject' => '__rm__'], []);
    [$component, $context] = app(HandleComponents::class)->fromSnapshot($snapshot);

    $this->assertEquals([], $component->subject);
}

{
    public array $subject = [['id' => '123']];

    public function render()
    {
        return <<<'HTML'
            <div></div>
        HTML;
    }
}
```

nuernbergerA:
@andybeesee agree that'S why my test is on a higher level.
 your solution seems to do the trick. :+1:

andybeesee:
Should we submit a new PR with the alternative solution? or wait to see if Caleb has any input?

nuernbergerA:
lets wait for caleb

calebporzio:
Ok here's what's happening:

When an item is removed from an array in Livewire's JS, it is sent as an "update" to the value of "__rm__".

You'll have to just take my word for it that this is a good pattern as opposed to alternative systems for removing individual items from an array.

The problem comes in with Livewire's system for updating values.

If you have a complex value like "Stringable" in your Livewire component, it will get dehydrated into a simple string in the frontend.

In a more naive system, if you updated the value of the stringable in JavaScript to a new string, Livewire would change the Stringable to a plain string in the backend.

Instead, Livewire fetches the "synthesizer" of the current value before updating it with the new value. It then hydrates the value about to be updated using that synthesizer. This way a stringable will stay a stringable but have a new value.

Does that make sense? Anyhow, I tried a few alternatives and I found the current solution to be the best. THANKS!

PR #6198, opened 2023-08-04
[V3] Added information about $dispatchTo

TheTomRoelofs:
I was missing information about directly emitting events to a component from a blade template.
It seems you can use `$dispatchTo` to do this. I've added some information about this to the docs.


joshhanley:
@TheTomRoelofs thanks!

lokutus24:
Livewire.dispatchTo is not working from Javascript!
Any update in this case?

PR #6199, opened 2023-08-04
[v3] Trait Lifecycle Hooks

TWithers:
Trait hooks are `hookTraitName()` and not `traitNameHook()`

### DIFF
--- a/docs/lifecycle-hooks.md
+++ b/docs/lifecycle-hooks.md
@@ -311,42 +311,42 @@ trait HasPostForm
 
     public $content = '';
 
-    public function hasPostFormMount()
+    public function mountHasPostForm()
     {
         // ...
     }
 
-    public function hasPostFormHydrate()
+    public function hydrateHasPostForm()
     {
         // ...
     }
 
-    public function hasPostFormBoot()
+    public function bootHasPostForm()
     {
         // ...
     }
 
-    public function hasPostFormUpdating()
+    public function updatingHasPostForm()
     {
         // ...
     }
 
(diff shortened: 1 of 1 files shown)


docs/lifecycle-hooks.md
```diff
@@ -311,42 +311,42 @@ trait HasPostForm
...
     }
 
-    public function hasPostFormRendered()
+    public function renderedHasPostForm()
     {
         // ...
     }
 
-    public function hasPostFormDehydrated()
+    public function dehydratedHasPostForm()
```
PerryvanderMeer, line 349: Should be `dehydrateHasPostForm`, as the hook method is `dehydrate()`.
See https://github.com/livewire/livewire/blob/ad33d14f9ca37a74f270b2495ec8f99aa3e29ba0/src/Features/SupportLifecycleHooks/SupportLifecycleHooks.php#L112C37-L112C37

PerryvanderMeer, line 349: Should be `dehydrateHasPostForm`, as the hook method is `dehydrate()`.

See https://github.com/livewire/livewire/blob/main/src/Features/SupportLifecycleHooks/SupportLifecycleHooks.php#L112C37-L112C37

PerryvanderMeer (changes requested):

docs/lifecycle-hooks.md
joshhanley: @PerryvanderMeer thanks, I have updated.

joshhanley:
@TWithers thanks!

PR #6201, opened 2023-08-04
[V3] Passing test for layout stacks #6135

bpudenz:
Adding back in layout stacks for V3. Added a test scenario for once.


joshhanley:
@bpudenz thanks for the PR! It looks like you took the code from V2, which actually was an experimental feature for dynamic stacks which we never enabled. You can see the blade directives are actually disabled in the V2 service provider.

https://github.com/livewire/livewire/blob/96a249f5ab51d8377817d802f91d1e440869c1d6/src/LivewireServiceProvider.php#L310-L317

So I don't think this PR fixes the issue of stacks running in V3 (without the custom blade directives).
I'll get Caleb to have a look and see what he thinks but he will probably close this for now as dynamic stacks are complex (hence why we disabled them in V2).

cpbret:
Yeah, I realized some of it was commented out after getting half way there, but tests were starting to pass so I kept going. 

The large site I'm upgrading won't go live for another 2 months, but the only real issue I've found so far is the lack of blade directives working. I am a little stumped at how these weren't in use in V2, but the directives worked fine in the app. @hasSection, @stack, @push, @once are relatively important in most Laravel apps, and if the choice is to hold off support for a while, it'd be very helpful to know.

calebporzio:
Ok all sections, pushes, appends, and slots should all be forwarded to the layout now just like V3

calebporzio:
FYI: I removed the V2 code you brought over. The reason it worked in V2 without that code:

When a full-page livewire component was rendered in v2 it was rendered in the same blade view as the layout itself.

In v3 I separated these two process which creates a disconnect between defining pushes and using them

bpudenz:
That all makes sense. Really appreciate the quick turn around here. Thanks Caleb

bpudenz:
It's working well. You can use dev-main as 3.0.0-beta.7 until a tag is created

PR #6204, opened 2023-08-05
[v3] Add example for testing a page contains Volt component

thinkverse:
This PR adds a small section to the Volt documentation on how to test if a page contains a Volt component. The example might be a bit long character-wise given you need to pass in the path to the component. 🤷‍♂️

AFAIK this is the only way as of now to test for this, so thought it'd be good to have an example of it available.

### DIFF
--- a/docs/volt.md
+++ b/docs/volt.md
@@ -682,3 +682,25 @@ If your Volt component is nested, you may use "dot" notation to specify the comp
 ```php
 Volt::test('users.stats')
 ```
+
+#### Testing a page contains a Volt component
+
+The simplest way of asserting a given endpoint contains and renders a Volt component is by using `assertSeeVolt`:
+
+```php
+<?php
+
+namespace Tests\Feature\Volt;
+
+use Tests\TestCase;
+
+class CreatePostTest extends TestCase
+{
+    /** @test */
+    public function component_exists_on_the_page()
+    {
+        $this->get('/posts/create')
+            ->assertSeeVolt('create-post');
+    }
+}
+```


docs/volt.md
```diff
@@ -682,3 +682,20 @@ If your Volt component is nested, you may use "dot" notation to specify the comp
...
+Where the first argument is the name passed into the `@volt` Blade directive, and the second is the path to where 
+the Volt component is located at:
+
+```php
+use Livewire\Volt\FragmentAlias;
+
+$this->get('/posts/create')
+    ->assertSeeLivewire(
+        FragmentAlias::encode('posts.create', resource_path('views/livewire/posts/create.blade.php'))
+    );
```
thinkverse, line 700: Or would it be better to suggest using `FragmentMap::get`? That also works but the tests do run a bit slower due to the look-up.

```suggestion
use Livewire\Volt\FragmentMap;

$this->get('/posts/create')
    ->assertSeeLivewire(
        FragmentMap::get('posts.create')
    );
```

thinkverse: I've decided against suggesting this, upon further testing it sometimes doesn't find the component, could possibly be because a component isn't in the look-up. I don't know the inner workings of Volt so I'm basing this on a rudimentary look of the source.

joshhanley:
@thinkverse thanks for the PR! We'll get one of the Laravel team to have a look to see what they think 🙂

thinkverse:
@nunomaduro added this, so this PR is no longer needed. 👍

PR #6210, opened 2023-08-05
[V3] Correct Installation command in installation and quickstart

ripdog:
Hi all,

I went to install livewire for the first time, and the installation command errored out with the following error:

```
    - Root composer.json requires livewire/livewire 3.0@beta, found livewire/livewire[dev-throw-error-when-testing-non-livewire-class, ..., dev-bad-hotfix-test, v0.0.1, ..., v0.7.4, v1.0.0, ..., 1.x-dev, v2.0.0, ..., 2.x-dev, v3.0.0-beta.1, ..., v3.0.0-beta.7] but it does not match the constraint.
```

This PR provides a working install command.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -3,7 +3,7 @@ Livewire is a Laravel package, so you will need to have a Laravel application up
 To install Livewire, open your terminal and navigate to your Laravel application directory, then run the following command:
 
 ```shell
-composer require livewire/livewire "^3.0@beta"
+composer require livewire/livewire "^3.0.*@beta"
 ```
 
 That's it — really. If you want more customization options, keep reading. Otherwise, you can jump right into using Livewire.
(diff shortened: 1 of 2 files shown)


PerryvanderMeer:
Please note that the caret version range isn't the same as the asterisk version range (see [here](https://getcomposer.org/doc/articles/versions.md#caret-version-range-)).

`^1.2.3` is equivalent to `>=1.2.3 <2.0.0`
`1.4.*` is equivalent to `>=1.4.0.0-dev <1.5.0.0-dev`

Also, the compability of a command depends on the operating system and software used, see https://github.com/livewire/livewire/pull/6076#issuecomment-1656336235.

joshhanley:
@ripdog thanks for the PR! I'm going to close this for now as I believe this is an issue unique to using powershell on windows (see comment https://github.com/livewire/livewire/pull/6076#issuecomment-1656336235 as linked above by @PerryvanderMeer).

@PerryvanderMeer I've messaged Caleb about this, if he is happy to add windows specific details in, I think we put a section on the installation page for powershell install, maybe we can put a warning on the quick start page for windows users, linking to the install page.

joshhanley:
Have submitted a PR with a warning for PowerShell users https://github.com/livewire/livewire/pull/6232

PR #6233, opened 2023-08-07
[V3] Move auto-injected livewire styles to end of head tag

joshhanley:
There is an issue #6040 with `wire:loading` elements not disappearing after loading is finished if the element has another class that sets the display value. This is because the auto-injected styles are injected at the beginning of the `<head>` tag instead of before the closing `</head>` tag. So Livewire's styles are not being given priority over any other styles.

This PR moves the auto-injected styles to the end of the head tag.

Having Livewire styles after all other styles was actually something I used to recommend to people in V2.

Hope this helps!

### DIFF
--- a/src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
+++ b/src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
@@ -48,7 +48,7 @@ class SupportAutoInjectedAssets extends ComponentHook
 
         $html = str($html);
 
-        if ($html->test('/<\s*head(?:\s|\s[^>])*>/i') && $html->test('/<\s*\/\s*body\s*>/i')) {
+        if ($html->test('/<\s*\/\s*head\s*>/i') && $html->test('/<\s*\/\s*body\s*>/i')) {
             return $html
                 ->replaceMatches('/(<\s*\/\s*head\s*>)/i', $livewireStyles.'$1')
                 ->replaceMatches('/(<\s*\/\s*body\s*>)/i', $livewireScripts.'$1')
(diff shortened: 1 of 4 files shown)


nuernbergerA:
Wouldn't it be better to wrap each livewire style in a `:where()`?

joshhanley:
@nuernbergerA Can you give me an example? 😅

nuernbergerA:
```html
<!-- Livewire Styles -->
<style {$nonce}>
    :where([wire\:loading]),
    :where([wire\:loading\.delay]),
    :where([wire\:loading\.inline-block]),
    :where([wire\:loading\.inline]),
    :where([wire\:loading\.block]),
    :where([wire\:loading\.flex]),
    :where([wire\:loading\.table]),
    :where([wire\:loading\.grid]),
    :where([wire\:loading\.inline-flex]) {
        display: none;
    }

    :where([wire\:loading\.delay\.shortest]),
    :where([wire\:loading\.delay\.shorter]),
    :where([wire\:loading\.delay\.short]),
    :where([wire\:loading\.delay\.long]),
    :where([wire\:loading\.delay\.longer]),
    :where([wire\:loading\.delay\.longest]) {
        display:none;
    }

    :where([wire\:offline]) {
        display: none;
    }

    :where([wire\:dirty]:not(textarea):not(input):not(select)) {
        display: none;
    }

    :where([x-cloak]) {
        display: none;
    }
</style>
```

nuernbergerA:
@joshhanley  forget it it will not work, because every other declaration will overwrite it

src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
```diff
@@ -50,7 +50,7 @@ static function injectAssets($html)
 
         if ($html->test('/<\s*head(?:\s|\s[^>])*>/i') && $html->test('/<\s*\/\s*body\s*>/i')) {
```
nuernbergerA, line 51:
```suggestion
        if ($html->test('/(<\s*\/\s*head\s*>)/i') && $html->test('/<\s*\/\s*body\s*>/i')) {
```

calebporzio:
Good call, thanks @joshhanley and @nuernbergerA !

PR #6234, opened 2023-08-07
Add warning to docs regarding auto-inject of Alpine.js

machacekmartin:
I _**think**_ the docs should _warn_ the users that since livewire's assets and alpine will be auto-injected only to sites that detect livewire, it will also NOT include alpine.js where there is no livewire.
 
Meaning that when you want to use pure alpine.js on a page that does not contain any livewire component, you will not have it at your disposal and will need to import it manually. What do you think?

I put the warning into the installation step, but it could potentionally be in the upgrade guide.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -39,6 +39,9 @@ If you want more control over this behavior, you can manually include the assets
 
 By including these assets manually on a page, Livewire knows to not inject the assets automatically.
 
+> [!warning] Alpine.js will only work where Livewire is detected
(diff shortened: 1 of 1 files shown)


nuernbergerA:
i think a good change could be to make `inject_assets` a bit more intelligent:
```php
// new will inject on all pages
'inject_assets' => 'always',
//current true
'inject_assets' => 'auto', // or 'detect'
// current false
'inject_assets' => 'off', // or false
```
with this alpine would be available on all pages

machacekmartin:
Oh yeah, that would make a lot of sense imho!

calebporzio:
Thanks!

joshhanley:
@nuernbergerA I'd say do it 😁 I would happily use `always`

PR #6237, opened 2023-08-07
Remove confusing npm install instructions (for now)

markjaquith:
I know the README is a WIP. This just updates it enough so that people who want to contribute to v3 aren't immediately led astray. These are the changes that would have helped me as I worked on a Dusk test for an issue I was tracking.

---

Adds a note to the README.md file that building assets is not usually necessary for development, and adds a note about running an individual PHPUnit test.


calebporzio:
Definitely an improvement. Thanks Mark!

PR #6256, opened 2023-08-07
[3.x] Add support for Laravel Echo

cmizzi:
Laravel Echo exposes a socket ID. This socket ID is mandatory when using `broadcast(...)->toOthers()`. During my test, the mandatory `X-Socket-Id` is not injected automatically by `fetch`, so we need to add it when available.

Temporarily, you can make it work by adding the following code to your JS code:

```js
    Livewire.hook('request', ({ options }) => {
        options.headers["X-Socket-Id"] = Echo?.socketId();
    })
})
```

### DIFF
--- a/js/request.js
+++ b/js/request.js
@@ -46,6 +46,7 @@ async function sendRequestToServer() {
             headers: {
                 'Content-type': 'application/json',
                 'X-Livewire': '',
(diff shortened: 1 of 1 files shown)


calebporzio:
Thank you for this! I forgot to add this back into Livewire.

PR #6262, opened 2023-08-08
[V3] Failing test for navigate back button not working when using scripts on non-livewire page

joshhanley:
This PR adds a failing test for a scenario where a page without a Livewire component has been navigated to, and clicking the back button doesn't work. See #5805 for video demonstration and example repo.

The catch with this one is, the initial page - that has the Livewire component - does not have `@vite` in the layout. Where as the second page - that does not have a Livewire component - does have `@vite in the layout.

The thing that `@vite` adds is `data-navigate-track` to the script tag it injects. As such, when navigate transitions the page, it unloads Livewire's scripts, due to that tracked script. As such the back button stops working.

This test simulates what is going on in the above test repo. If you remove `data-navigate-track` from `<script src="/test-navigate-asset.js" data-navigate-track></script>` on the `/page-without-livewire-component` route, you will see the test passes.

Hope this helps!


calebporzio:
Fixed! Thanks @joshhanley

PR #6264, opened 2023-08-08
[V3] Add failing test for child events not being listened to correctly after navigate

joshhanley:
This PR adds a failing test for the issue reported in #6099.

When there are two components on the page where on emits an event to the other, after the page has been navigated to using `wire:navigate` the parent component is no longer receiving the events.

See below error of what is being thrown to the console.

image

Hope this helps!


calebporzio:
Fixed! Thanks for the test

PR #6266, opened 2023-08-08
[V3] Failing test for navigate triggering on right click

joshhanley:
This PR adds a failing test for issue #6077 where navigate it being triggered on right/ middle clicks, instead of just left clicks.

Hope this helps!


calebporzio:
All fixed - thanks Josh!

PR #6272, opened 2023-08-08
Failing test for can't set model as property if it hasn't been persisted yet

riasvdv:
This adds a failing test for #6120 - "Can't set model as property if it hasn't been persisted yet" when component deletes model

I've used actual eloquent with sqlite here with inline table creation as you can't delete models with Sushi as far as I know.

### DIFF
--- a/src/Features/SupportNestingComponents/BrowserTest.php
+++ b/src/Features/SupportNestingComponents/BrowserTest.php
@@ -2,6 +2,8 @@
 
 namespace Livewire\Features\SupportNestingComponents;
 
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\Schema;
 use Livewire\Attributes\Layout;
 use Livewire\Component;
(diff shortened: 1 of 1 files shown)


riasvdv:
This had some conflicts so I've re-applied the failing test to the latest main branch

masterix21:
Works If you remove lines 20-22 from https://github.com/livewire/livewire/blob/266c356a11d584dc95961a7cf0737e33f7b80a06/src/Features/SupportModels/ModelSynth.php#L20-L22

Anyway, I don't know if it could be the solution.

3alampro:
i'm using the same controller to edit and create new model by using optional route model binding. but with this Exception i can't use it like v2

joshhanley:
@riasvdv thanks for the PR! I'm going to close this in favour of PR #7003 which includes a fix. I have merged in PR #6858 which was similar to this one but uses Sushi.

PR #6273, opened 2023-08-08
[V3] Make sure `orchestra/testbench-dusk` is installed

jerodev:
When running tests on CI without having `orchestra/testbench-dusk` installed, the CI will crash because the `\Orchestra\Testbench\Dusk\Options` is not found.

This PR adds a check if the classes of this package exist before using them.


calebporzio:
Thanks!

PR #6286, opened 2023-08-09
[V3] Livewire page component layout view not found if a component contains nested component

riasvdv:
This adds a failing test for #6281

The exception only happens when you do not have the default livewire layout blade file present and the parent component uses a custom layout.


calebporzio:
All fixed. Thanks!

PR #6289, opened 2023-08-09
failing test: Removing entangled component

danharrin:

This should fix many issues that Filament users are currently facing.

Yes

No

Yes

This was the easiest way of reproducing the issue in a test. Essentially, once an entangled property is removed from the DOM by Livewire, I don't think it's being properly destroyed. For example, watchers continue to be run. In my test case, something is happening so that removing items from a "repeater" does not actually remove them, and they end up reappearing.


Focusnik:
I'm using 3.0.0-beta10 version and Filament Builder sill broken. Same error - Undefined array key "type".

Focusnik:
@joshhanley as i can see @riasvdv have already completed test and fix here - https://github.com/riasvdv/livewire/commit/7c8312792f1c5a20272f1893f29aaae66a2c8c93

Focusnik:
My bad, i've checked the code and we really need another test case for Builder component example

danharrin:
I've submitted a new failing test PR #6521

PR #6296, opened 2023-08-10
[v3] Add failing test for rendering and updating components with a root <tr> element

markjaquith:
see #6247

If your Livewire v3 component has a root element that is a `<tr>` element,  when the component updates, its re-render will "unwrap" instead of re-rendering the `<tr>` element.

In my app where I saw this issue, the `<tr>` rendered correctly, but on update it would "unwrap".

### DIFF
--- a/src/Tests/UpdatingTableRowsTest.php
+++ b/src/Tests/UpdatingTableRowsTest.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Livewire\Tests;
+
+use Livewire\Component;
+use Livewire\Livewire;
+
+class UpdatingTableRowsTest extends \Tests\BrowserTestCase
+{
+    /** @test */
+    public function component_renders_table_rows_and_updates_properly()
+    {
+        Livewire::visit(new class extends Component {
+            public int $counter = 0;
+
+            public function increment()
+            {
+                $this->counter++;
+            }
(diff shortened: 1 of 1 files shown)


markjaquith:
@calebporzio I see what's going on here.

In `morph()`, you're creating a `<div>` wrapper and injecting the component HTML into it:

https://github.com/livewire/livewire/blob/310b44e4d7a4d6ba068564130e80a19f7f21a2f2/js/morph.js#L6-L8

But [the HTML5 spec says](https://html.spec.whatwg.org/#the-tr-element) you can't have a `<tr>` element as a direct child of a `<div>`. Look what happens if you try:

CleanShot 2023-08-10 at 01 22 40@2x

The HTML spec is a maze of these sorts of restrictions.

Idea: `el` has a parent, and presumably the person who nested this component inside it did it correctly... So if `el` is a `<tr>` the parent will be a `<table>`, `<tbody>`, `<thead>` or `<tfoot>`.

So why not just see what type of element `el`'s parent is, and use that type for the temporary wrapper?

i.e.

```js
let wrapper = document.createElement(el.parentElement.tagName.toLowerCase());
```

joshhanley:
Closing this in favour of PR #6299 which includes this test and a fix.

PR #6298, opened 2023-08-10
[V2] Fix extension

nuernbergerA:
See https://github.com/livewire/livewire/pull/6011#issuecomment-1672525845

i think the best solution is to just use the extension from the client filename and in case there is non we detect base on the mimetype.


dalbo:
Thanks Adrian. Using the client's original extension does seem more sensible than guessing it based on the mime-type provided. 

I guess it depends on the desired outcome, but for me I'm not sure that trusting the client at all is the best way to do this. But I'm going to try and move our code away from relying on file extensions at all.

PR #6299, opened 2023-08-10
[v3] Use the actual wrapper element type when wrapping HTML during morph

markjaquith:
See bug report in #6247 

And explanation of fix: https://github.com/livewire/livewire/pull/6296#issuecomment-1672637691

- Add failing test for rendering and updating components with a root <tr> element
- Implement table row test as a child component
- Make the test fail instead of error
- Use the actual wrapper element type when wrapping HTML during morph

closes #6296
fixes #6247


N1ebieski:
Generally it works, but if for example some filter changes number of elements from 25 to 10, I get the error: "Uncaught TypeError: e.parentElement is null"

markjaquith:
Hmm. Maybe there's some way to hold on to the data of what the parent element type was?

N1ebieski:
Maybe just add an additional condition:

```js
// js/morph.js
var import_alpinejs9 = __toESM(require_module_cjs());
function morph2(component, el, html) {
    if (!el.parentElement) return;

    let wrapper = document.createElement(
        el.parentElement.tagName.toLowerCase()
    );
    wrapper.innerHTML = html;
```

in my case it works...

joshhanley:
@markjaquith thanks for the PR! In future, if you've already submitted a failing test PR and end up working on a fix, please just push it to the same PR and update the title and description. Thanks!

PR #6308, opened 2023-08-10
[v3] Use Laravel Console components in commands

datlechin:
Use Laravel Console components in commands

Before:
image
After:
image

___
Before:
image
After:
image


calebporzio:
Thank you for the contribution, however I think we'll stick with what currently exists.

At some point we will probably end up rewriting all this using Laravel's new "Prompts"

PR #6312, opened 2023-08-11
[V3] Fix typo in emitUp step of upgrade tool

liamja:
The typo inferred that Livewire 3 does NOT use browser events, when in fact it does.

`s/not/now`


PhiloNL:
Thanks!

PR #6315, opened 2023-08-11
[v3] Show livewire version from about command

thethunderturner:
I added these lines of code to quickly check the livewire version from laravel's `php artisan about` command. I feel like this could be a good enhancement.


calebporzio:
Nice! thanks

PR #6317, opened 2023-08-11
Allow customizable middlewares

francoism90:

https://github.com/livewire/livewire/discussions/6316

Yes

No

It should use the default Livewire tests

It's not possible to customize the middlewares: order of loading, location, etc.

I also think a config file would be useful, as this allows the user to overrule the behavior.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -143,4 +143,25 @@ return [
     */
 
     'pagination_theme' => 'tailwind',
+
+    /*
+    |---------------------------------------------------------------------------
+    | Persistent Middlewares
+    |---------------------------------------------------------------------------
+    |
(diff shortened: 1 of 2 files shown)


francoism90:
Hmm, it seems you can actually do this:
```php
use Livewire\Livewire;

Livewire::setPersistentMiddleware([
    // custom providers here
]);
```
Filament seems to add them instead.

I don't know what is better, as a config file would mean caching.

bpudenz:
@francoism90 are you able to get the persistentMiddleware working via Livewire v3 or Filament? I see the middleware existing in livewirePersistentMiddleware when dumping the Panel config, but it isn't applied at all.

francoism90:
@bpudenz It is working for me, could you post example code?

PR #6321, opened 2023-08-11
[v3] Add failing test for automatic route property binding

PhiloNL:
In the [docs](https://livewire.laravel.com/docs/components#using-route-model-binding) the following is mentioned:

> Livewire knows to use "route model binding" because the Post type-hint is prepended to the $post parameter in mount().
> Like before, you can reduce boilerplate by omitting the mount() method

When you remove the mount method, the binding will no longer work. If you add the `mount` method without any properties it will work again.

```php
// Works
class Component {
    public $post;

    public function mount($post) 
    {
        $this->post = $post;
    }
}

// Works
class Component {
    public $post;

    public function mount() 
    {
    }
}

// Fails
class Component {
    public $post;
}
```

It seems to be related to https://github.com/livewire/livewire/blob/11c75143c19d4f4559d0cf67e27f458cbd0c154e/src/Drawer/ImplicitRouteBinding.php#L84-L88

This will only resolve properties if they have a type-hint:

```php
// Works
class Component {
    public Post $post;
}

// Fails
class Component {
    public $post;
}
```

Shouldn't the route property binding always bind regardless of whether the property is type-hinted? I believe this is how it worked in v2.

### DIFF
--- a/src/Drawer/ImplicitRouteBinding.php
+++ b/src/Drawer/ImplicitRouteBinding.php
@@ -65,13 +65,19 @@ class ImplicitRouteBinding
         return $this->getPublicPropertyTypes($component)
             ->intersectByKeys($route->parametersWithoutNulls())
             ->map(function ($className, $propName) use ($route) {
-                $resolved = $this->resolveParameter($route, $propName, $className);
+                // If typed public property, resolve the class
+                if($className) {
+                    $resolved = $this->resolveParameter($route, $propName, $className);
 
-                // We'll also pass the resolved model back to the route
-                // so that it can be used for any depending bindings
-                $route->setParameter($propName, $resolved);
+                    // We'll also pass the resolved model back to the route
+                    // so that it can be used for any depending on bindings
+                    $route->setParameter($propName, $resolved);
 
-                return $resolved;
+                    return $resolved;
+                }
+
+                // Otherwise, just return the route parameter
+                return $route->parameter($propName);
             });
     }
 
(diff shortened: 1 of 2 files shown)


calebporzio:
I tested in V2 and the behavior seems to be the same in both V2 and V3. 

After looking at the V2 documentation I also don't find mention of being gable to set non-typed properties from route params.

Question is: do we add this? does it introduce a vulnerability?

If you think the answer is yes we should add it and no it doesn't introduce a vulnerability, then I'd be down for a fix PR (it should be a pretty simple change to the code you linked)

PhiloNL:
Thanks for checking!

I personally would prefer if it works regardless if it's type hinted.

It already works kinda by simply adding the mount method (even if you leave everything empty).

I'm not 100% sure if there are any security implications but I think they are quite limited as your route variable name must match a public property name, given it already is a public component property it could be manipulated anyhow. 

I'm curious if you think type hinted properties would be safer versus non typed-hinted properties. Maybe I'm missing something.

calebporzio:
Nah, just wanted to sanity check. Because this is already the behavior in V2 I'm not going to prioritize implementing. However, if you want to add the implementation to this PR, I would merge it (especially if we can do that before v3 is tagged)

Thanks Philo

PhiloNL:
Perfect! :) Added the implementation https://github.com/livewire/livewire/pull/6321/commits/39185ec14d189b2dda0d4e402c4bd2b49fb7cb8a

calebporzio:
Perfect, thank you!!

PR #6323, opened 2023-08-11
[v3] Rename duplicate Suit enum in test

markjaquith:
This prevented PHPUnit from running.

CleanShot 2023-08-11 at 14 10 00@2x

There were two enums named Suit in the same namespace.

Instead of referencing the one from the other test, I just renamed the Unit test one to UnitSuit. I feel like tests shouldn't depend on assets from other tests.


joshhanley:
Thanks!

PR #6324, opened 2023-08-11
[V3] Fix Laravel Octane asset loading

francoism90:

[Yes](https://github.com/livewire/livewire/discussions/6322)

Yes

No

No, I don't think you can add tests for Octane?

I'm using Laravel Octane, see https://laravel.com/docs/10.x/octane#managing-memory-leaks

The problem is that `app(FrontendAssets::class)->hasRenderedScripts` will always be `true`, even when you (hard) reload the page, as this value is being cached.

I think the only workaround is to remove this line, or use `static::$forceAssetInjection` something like this:
```php
if (! static::$forceAssetInjection && app(FrontendAssets::class)->hasRenderedScripts) return;
```

Thanks

### DIFF
--- a/src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
+++ b/src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
@@ -26,7 +26,6 @@ class SupportAutoInjectedAssets extends ComponentHook
             if (! str($handled->response->headers->get('content-type'))->contains('text/html')) return;
             if (! method_exists($handled->response, 'status') || $handled->response->status() !== 200) return;
             if ((! static::$hasRenderedAComponentThisRequest) && (! static::$forceAssetInjection)) return;
-            if (app(FrontendAssets::class)->hasRenderedScripts) return;
 
             $html = $handled->response->getContent();
 


francoism90:
It seems one can use something like this:
```php
 isset($_SERVER['LARAVEL_OCTANE'])
```
I don't know if Livewire should be used with Octane, but I would like to know your thoughts about this. I'm a bit worried about the aggressive caching when using Octane, and using something dynamic as Livewire.

calebporzio:
I think Livewire is actually better with Octane - we just need to make sure we clean everything up between requests properly.

We have a hook we use internally that we can unset variables like this inside of.

Does this line of code in core cover this already though? https://github.com/livewire/livewire/blob/c341e462b54756042f20f9e1632231f7de3d2580/src/Mechanisms/FrontendAssets/FrontendAssets.php#L37

francoism90:
Sorry, I just wanted to ask if it was Octane compatible, not blaming Livewire or something.

It didn't for me, as it would still return the value true.

calebporzio:
Ok, thanks.

Can you investigate this problem further, ideally writing a test to cover this case?

joshhanley:
Hey @francoism90 just wondering if you'd had a chance to explore this further?

francoism90:
@joshhanley I'lll try later this week/weekend.

joshhanley:
@francoism90 thanks!

skeemer:
While working on #6509 I ran into problems with the static changes crossing over between tests. Is there some hook we could grab when a request starts to run 'flush-state'?

calebporzio:
Closing due to inactivity

PR #6327, opened 2023-08-12
[v3] Add failing test for lazy nested components calling boot method twice

archilex:
Issue: Nested components that use `lazy` call the boot method twice when initially loaded. In scenarios where the boot method is used to render additional content, this content is then duplicated on the page. 

Removing `lazy` prevents this from happening. 

Expected: The boot method would only be called once.

### DIFF
--- a/src/Features/SupportLazyLoading/SupportLazyLoading.php
+++ b/src/Features/SupportLazyLoading/SupportLazyLoading.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features\SupportLazyLoading;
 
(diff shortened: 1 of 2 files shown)


danharrin:
Subscribing myself here so I remember to revert #7689 when this is merged

joshhanley:
@archilex thanks for the PR! Yeah makes sense to me, if mount is deferred, then the initial boot should be too.

calebporzio:
Yeah I think it was supposed to be, but there was a bug. Fixed now!

PR #6335, opened 2023-08-12
[V3] Use route-key

francoism90:

https://github.com/livewire/livewire/discussions/6334

The main issue is exposing database details, e.g. of user-ids:
- 1
- 2
- 3
- ..

Someone could guess the number of active users (or any other models).
It's pretty easy, just do `window.livewire.all()` in a console window.

Also it can be confusing, as I'm talking to the API over prefixed ID's, instead of actual model-ids.

Yes

No (not yet).

It doesn't change collection model id's yet.

No, because I would like to gain some feedback first.

```json
{
    "data": {
        "book": [
            null,
            {
                "class": "book",
                "key": "book-gxMRBP0zzh", // <-- this value
                "s": "mdl"
            }
        ]
    },
    "memo": {
        "id": "Um0csVv3RF021PpgJVjM",
        "name": "books.book-show-controller",
        "path": "books\\/book-gxMRBP0zzh",
        "method": "GET",
        "children": [],
        "lazyLoaded": false,
        "errors": [],
        "locale": "en"
    },
    "checksum": ".."
}
```
As you can see, the `key` is now the route-key, instead of the current model id.

It would be great if class could be hidden as well, but it seems to be used in two locations/keys.

Thanks!

### DIFF
--- a/src/Features/SupportModels/ModelSynth.php
+++ b/src/Features/SupportModels/ModelSynth.php
@@ -24,11 +24,11 @@ class ModelSynth extends Synth {
         // If no alias is found, this just returns the class name
         $alias = $target->getMorphClass();
 
-        $serializedModel = (array) $this->getSerializedPropertyValue($target);
+        $key = $target->getRouteKey();
 
         return [
             null,
-            ['class' => $alias, 'key' => $serializedModel['id']],
+            ['class' => $alias, 'key' => $key],
         ];
     }
 
@@ -43,7 +43,11 @@ class ModelSynth extends Synth {
             $class = $aliasClass;
         }
 
-        $model = (new $class)->newQueryForRestoration($key)->useWritePdo()->firstOrFail();
+        $model = (new $class)
+            ->newQueryWithoutScopes()
+            ->where((new $class)->getRouteKeyName(), $key)
+            ->useWritePdo()
+            ->firstOrFail();
 
         return $model;
     }


calebporzio:
I commented in the discussion. What do you think about an `#[Encrypted]` attribute instead?

francoism90:
Sounds interesting. :)
I did it before: https://github.com/foxws/livewire-encrypt

As long as it can encrypt/decrypt on (de)hydrate, it should work fine.

francoism90:
@calebporzio It would be cool if you could introduce the Encryption attribute. :)

However, after reading the docs, it seems something like this is possible:
```php
<?php

namespace Support\Livewire;

use Livewire\Features\SupportModels\ModelSynth as Synth;
use Illuminate\Database\Eloquent\Relations\Relation;

class ModelSynth extends Synth
{
    public function dehydrate($target) {
        if (! $target->exists) {
            throw new \Exception('Can\'t set model as property if it hasn\'t been persisted yet');
        }

        // If no alias is found, this just returns the class name
        $alias = $target->getMorphClass();

        $serializedModel = (array) $this->getSerializedPropertyValue($target);

        return [
            null,
            ['class' => $alias, 'key' => $target->getRouteKey()],
        ];
    }

    public function hydrate($data, $meta) {
        $key = $meta['key'];
        $class = $meta['class'];

        // If no alias found, this returns `null`
        $aliasClass = Relation::getMorphedModel($class);

        if (! is_null($aliasClass)) {
            $class = $aliasClass;
        }

        $model = (new $class)
            ->newQueryForRestoration($key)
            ->where((new $class)->getRouteKeyName(), $key)
            ->useWritePdo()
            ->firstOrFail();

        return $model;
    }
}
```
See https://livewire.laravel.com/docs/synthesizers#registering-a-custom-synthesizer for details.

It seems to be able to introduce encryption using this as well, but I still would prefer a proper solution.

calebporzio:
Closing this in favor of: #6598

PR #6337, opened 2023-08-13
[V3] Failing test for Rule attributes

edalzell:

Yes, discussion: https://github.com/livewire/livewire/discussions/6157

Yes

No

Yes

Issue here: https://github.com/livewire/livewire/discussions/6157

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportValidation/BrowserTest.php
+++ b/src/Features/SupportValidation/BrowserTest.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Livewire\Features\SupportValidation;
+
+use Livewire\Attributes\Rule;
+use Livewire\Component;
+use Livewire\Livewire;
+
+/** @group morphing */
+class BrowserTest extends \Tests\BrowserTestCase
+{
+    /** @test */
+    public function validation_attribute_rules_triggered_on_save()
+    {
+        Livewire::visit(new class extends Component
+        {
+            #[Rule('required', 'min:3')]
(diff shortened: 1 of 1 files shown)


joshhanley:
Spoke with Erin about this and the existing test (below), and either the functionality is wrong and needs to be fixed (this failing test PR); or the docs are incorrect (which based on the test below, seems possible) and they need changing.

https://github.com/livewire/livewire/blob/11c75143c19d4f4559d0cf67e27f458cbd0c154e/src/Features/SupportValidation/UnitTest.php#L17-L40

Basically the summary is, that the docs suggest that if you use live validation with rule attributes, then you don't need to call `$this->validate();` in your save method. But the unit test above actually calls `$this->validate();` in the save method, even though validation is also run on property update.

See docs here https://livewire.laravel.com/docs/validation#rule-attributes

image

calebporzio:
Yeah, that documentation is misleading. I re-worded it to be this now:

image

calebporzio:
I ran the test locally and it failed, but because of other reasons. 

```php
#[Rule('required', 'min:3')] // INVALID
#[Rule('required|min:3')] // VALID
#[Rule(['required', 'min:3'])] // Valid
```

(I also added a ->waitForLivewire())

So in summary, I'm closing this because the test actually passes and behaves how you expect. The documentation was actually the misleading part here and that has been fixed and published.

Thanks!

PR #6344, opened 2023-08-13
[V3] Make rule aliases translatable

elcapo:
### Discussed in https://github.com/livewire/livewire/discussions/6186

<div type='discussions-op-text'>

<sup>Originally posted by **elcapo** August  3, 2023</sup>
### Livewire version

^3.0@beta

### Which PHP version are you using?

PHP 8.2

### Repository URL

_No response_

### Steps To Reproduce

1. Define a form property with an aliased rule

~~~php
    #[Rule('required|min:3|max:32', as: 'first name')]
    public ?string $firstName;
~~~

2. Set the application locale to a non English language

~~~php
// config/app.php
    'locale' => 'es',
~~~

3. Install some validation translations in your lang/es folder

4. Expose your property in a Livewire component

5. Forcing a validation with an invalid value (for instance, with an empty value)

At this point, you'll see your error message translated with the alias as it was provided in the rule (in English). In other words, you won't have any chance to add a translation for **first name**. You'll get:

> El campo first name es obligatorio.

Where the message would be translated but the alias wouldn't. Also, no call to the translation engine is made as an attempt to find a translation.

### Additional information

Although I'm not experienced enough to send a PR to Livewire repository, I think this patch for the [src/Features/SupportValidation/Rule.php](https://github.com/livewire/livewire/blob/9431cab3c3d3fa0b48f070481336364c39e667c4/src/Features/SupportValidation/Rule.php#L50) would be really close to solve the issue:

~~~php
        if ($this->as) {
            if (is_array($this->as)) {
                $this->component->addValidationAttributesFromOutside(trans($this->as));
            } else {
                $this->component->addValidationAttributesFromOutside([$this->getName() => trans($this->as)]);
            }
        }
~~~

Note that the change consists in adding a call to the **trans** helper before storing the alias.

- [X] I have provided easy and step-by-step instructions to reproduce the bug.
- [X] I have provided code samples as text and NOT images.
- [X] I understand my bug report will be removed if I haven't met the criteria above.</div>


calebporzio:
Thanks for the contribution, my only hesitation is: are there cases where people WOULDN'T want the name to be translatable?

Like maybe instead, we could make it opt-in via:

```php
// Via a parameter
#[Rule('required', as: 'date of birth', localize: true)]
public $dob = '';

// Via a special "_as"
#[Rule('required', _as: 'date of birth')]

// Via a special helper
#[Rule('required', as: '_(date of birth)')]
```

Idk, probably better to just do it by default. Thoughts @elcapo and @joshhanley ?

joshhanley:
@calebporzio maybe localise parameter is true by default and they can set to false on the chance they don't want something translated?

calebporzio:
Makes good sense - thank you!

PR #6353, opened 2023-08-14
[v3] fix: edited toggleBooleanStateDirective

sakanjo:

[Discussion #6352](https://github.com/livewire/livewire/discussions/6352)

yes

no

no

Thanks for contributing! 🙌


joshhanley:
@sakanjo thanks for the PR! Can you please add a test? Thanks!

sakanjo:
@joshhanley just a quick note before merging, since i wasn't able build the javascript files due to "@alpinejs/history" library because i think it's private, could do that please because the **esm** and the **minified** files **aren't updated**.

danharrin:
@sakanjo I added some compilation instructions in #6363

sakanjo:
@danharrin I couldn't build files

image

i added `external: ['alpinejs', '@alpinejs/morph', '@alpinejs/history', '@alpinejs/navigate'],` to `scripts/build.js` but another error comes in

image

sakanjo:
@joshhanley This PR just needs to build js files and it's ready to go!

calebporzio:
Thanks for the contribution @sakanjo.

I'm a little unclear on a few things. In general, please write out a description of the problem and the proposed solution in the PR description (rather than linking to a discussion).

Given the following HTML, this is what I would EXPECT livewire to do:

```html
<button wire:loading.attr="disabled" disabled>Submit</button>
```

### Expected behavior:
* The browser receives the `<button>` tag with `disabled` on it. It initially renders the button as disabled
* Livewire's JavaScript kicks in and removes "disabled" because of `wire:loading.attr`
* When the button is pressed, `disabled` is re-added while the request is out
* When the request comes back, `disabled` is removed from `<button>`

This is similar to how, say, Alpine would behave in this case: `<div x-bind:disabled="isLoading" disabled>` (I haven't verified that's the behavior but I'm pretty sure)

### Actual behavior:
* `disabled` is rendered before Livewire's JS is initialized
* Livewire removes `disabled`
* When a request is sent, `disabled` is added
* DIFFERENCE: When a response is received, `disabled` is NOT removed

The reason this happens is because Livewire handles loading states BEFORE morphing the HTML on the page. Therefore, `disabled` is removed and then immediately re-added by morph because it exists again in the HTML from the server.

### Behavior of this PR:
* `disabled` is rendered by the browser before Livewire's JS is initialized
* `disabled` is kept on the element for the initial render because of the change in this PR
* `disabled` is added when a request is sent (even though it's already there)
* `disabled` is removed by Livewire's JS, then re-added by Livewire's morph

So effectively, with this PR, the button in the example HTML is ALWAYS `disabled`. Which I don't think is the expected behavior, because otherwise, why would you have put `wire:loading.attr="disabled"` on it in the first place?

## My questions:
* What behavior do you expect?
* If it's the behavior added by this PR, please give more reasoning why it should be that way?

Also, if the change made in the PR WAS the right way to go, the implementation isn't comprehensive enough. It isn't checking that `.attr` is used and that it's a boolean attribute, so the current proposed code (`, el[directive.expression]`) will cause weird things in things like the following case:

```js
<button wire:loading.class="disabled" disabled>
```

With the current implementation, `el[directive.expression]` would resolve to "true" because you are trying to add a disabled class.

Thanks for being willing to contribute. I hope to hear more from you about this problem.

danharrin:
Imagine you have a `<button>` blade component, which has a feature where it gets disabled with `wire:loading`. Or maybe an `<input>` Blade component which gets disabled while `wire:loading`.

And now you pass a `disabled` attribute to that button or input Blade component, which forwards that attributes to the `<button>` or `<input>` itself. Now, it looks like this:

```html
<button wire:loading.class="disabled" disabled>
```

So it should always stay disabled, regardless of the Livewire loading state, since that was passed to the Blade component.

Hence why it is correct that morph re-adds the attribute after a request, as it was there in the first place.

I think this is a pretty common use case when building Blade components, and this is how it worked in v2 - morph overwrote the loading state.

danharrin:
Caleb, can you think of any use cases where morph should not overwrite the loading state that has now finished?

danharrin:
Oh - I have another use case off the top of my head. An enabled button which, after you click it, gets disabled by the action that was taken - "one time".

A button that is enabled, then `wire:loading` makes it disabled during a Livewire request, and the response of that Livewire request makes it disabled. It wasn't disabled before, but it is now, and that has been ignored since the `wire:loading` overwrote it.

sakanjo:
@calebporzio Thanks for reaching out, the examples given by @danharrin are what i would consider as practical and insightful demonstrations of the concept.

Now regarding to `el[directive.expression]` i think it's a bad idea since you don't know what could be passed in, i think the appropriate solution is to remove the line since `wire:loading` means only run when **loading** and by doing so that initial call is unnecessary.

calebporzio:
Great points. I agree with both of Dan's examples.

And good recommendation with removing the first line - I checked and all the other loading state tests still pass, so went with that.

danharrin:
Thanks Caleb!

sakanjo:
Thanks @calebporzio!

PR #6360, opened 2023-08-15
[v3] Documentation typos

klaasgeldof:
I read the entire v3 docs and noted some typos and small improvements.

Some `wire:key`s were missing as well. If you instruct the user to always include a key, I think you should do it yourselves in the docs as well 😉


calebporzio:
These are great- thank you!

PR #6362, opened 2023-08-15
fix: Call to undefined method ReflectionUnionType::getName() on multiple-types property

sakanjo:

#6361

yes

no

absolutely

Thanks for contributing! 🙌

### DIFF
--- a/src/Mechanisms/HandleComponents/BrowserTest.php
+++ b/src/Mechanisms/HandleComponents/BrowserTest.php
@@ -92,6 +92,32 @@ class BrowserTest extends \Tests\BrowserTestCase
     }
 
     /** @test */
+    public function it_uses_the_synthesizers_for_multiple_types_property_updates()
+    {
+        Livewire::visit(new class extends \Livewire\Component {
+            public string|int $localValue = 15;
+
+            public function render()
+            {
+                return <<<'HTML'
+                <div>
+                    <input type="text" dusk="localInput" wire:model.live.debounce.100ms="localValue">
+
+                    <span dusk="localValue">{{ $localValue }}</span>
+                </div>
+                HTML;
+            }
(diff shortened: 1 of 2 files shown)


calebporzio:
Thank you!

sakanjo:
@calebporzio That part of code still needs to be updated since if you do `string|int` it and assign a string value it will coerce it to int `0`,

We should add something like `BuiltinSynthesizers` for built-in types.

nuernbergerA:
I don't understand this. Why should a value be an int or string or any other mixed type?

sakanjo:
https://github.com/livewire/livewire/assets/121197517/1eb14a7d-8d0f-4a2e-8b3c-e4678557853f

```php
<?php

namespace App\Livewire;

use Livewire\Component;

class CreateInt extends Component
{
    public string|int $value = 10;

    public function save(): void
    {
        back()->with('message', 'The value was saved!');
    }

    public function render()
    {
        return view('livewire.test');
    }
}
```

```html
<form wire:submit.prevent="save">
    @if (session()->has('message'))
        <div class="bg-green-200 text-green-800 px-3 py-2 rounded mb-6">
            {{ session()->get('message') }}
        </div>
    @endif

    <div>
        <label for="value">Value</label>
        <input wire:model="value" id="value" class="border" />
        @error('value')
            <span class="text-red-500">{{ $message }}</span>
        @enderror
    </div>

    <div class="mt-10">
        <button
            class="text-sm font-bold rounded bg-blue-500 disabled:bg-red-500 px-4 py-2 text-white"
            type="submit"
        >
            Save
        </button>
    </div>
</form>
```

@nuernbergerA we should expect different scenarios not only string|int but null|int which if you pass null it will be coerced to 0 and so on...

PR #6345, opened 2023-08-14
[V3] Can reset unset properties

edalzell:
Review the contribution guide first at: https://laravel-livewire.com/docs/2.x/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

I ran into an issue trying to reset a property with no default value.

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)

Yes

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No

4️⃣ Does it include tests? (Required)

Yes

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

You can now call `reset` when there are typed properties that have no default values

Thanks for contributing! 🙌

PR #6375, opened 2023-08-16
More robustly handle resetting properties that are unset by default

markjaquith:
#6345 introduced a test failure.

```
❯ ./vendor/bin/phpunit --filter can_reset_form_object_property
PHPUnit 9.6.10 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

Time: 00:00.093, Memory: 48.50 MB

There was 1 failure:

1) Livewire\Features\SupportFormObjects\UnitTest::can_reset_form_object_property
--- Expected
+++ Actual
@@ @@
-''
+'Some Title'

/Users/mark/git/livewire/src/Features/SupportTesting/MakesAssertions.php:93
```

This happened because it's doing `isset($freshInstance->{$property})` and `unset($this->{$property})` but `$property` could be using dot notation, like... `form.title` in the test.

https://github.com/livewire/livewire/blob/8f3511589cf0148943bab54334f5934ce565883b/src/Concerns/InteractsWithProperties.php#L56-L60

This PR uses `data_forget()` and makes the checking for unset-by-default properties more robust. It also beefs up `getPublicProperties()` to not blindly call `->getType()` on values that can sometimes be null.

### DIFF
--- a/src/Concerns/InteractsWithProperties.php
+++ b/src/Concerns/InteractsWithProperties.php
@@ -52,14 +52,20 @@ trait InteractsWithProperties
         $freshInstance = new static;
 
         foreach ($properties as $property) {
-            // Handle uninitialized properties differently to avoid an error being thrown.
-            if (! isset($freshInstance->{$property})) {
-                unset($this->{$property});
+            $defaultValue = data_get($freshInstance, $property);
 
-                continue;
+            // Handle unsetting properties that are unset by default.
+            if (!$defaultValue) {
+                // Generate a unique handle so we can distinguish between falsey and unset properties.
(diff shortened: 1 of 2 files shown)


markjaquith:
>[!NOTE]
> The 8.2 failure just seems like an unstable test. I get random errors on the browser tests... so when that happens I just re-run that individual test and it usually passes.

edalzell:
Oh sorry about that, nice find Mark

calebporzio:
Thank you for this @markjaquith 👌  (I should have checked that that PR introduced a breaking test)

PR #6380, opened 2023-08-16
flaky-test: Use waitForText() and waitUntilMissingText() instead of manual pause()

markjaquith:
This test was failing ~3% of the time. It was using a manual `pause(100)` which was sometimes not long enough. By changing to `waitForText()` and `waitUntilMissingText()`, we can let Dusk take care of re-polling the DOM.

As a bonus, it might end up being faster, because if ~3% of the time 100ms isn't long enough, then ~97% of the time, 100ms is more than is needed!

### DIFF
--- a/src/Features/SupportJsEvaluation/BrowserTest.php
+++ b/src/Features/SupportJsEvaluation/BrowserTest.php
@@ -31,12 +31,13 @@ class BrowserTest extends \Tests\BrowserTestCase
                 </div>
                 HTML; }
         })
+        ->waitUntilMissingText('Toggle Me!')
         ->assertDontSee('Toggle Me!')
         ->click('@toggle')
-        ->pause(100)
+        ->waitForText('Toggle Me!')
         ->assertSee('Toggle Me!')
         ->click('@toggle')
-        ->pause(100)
+        ->waitUntilMissingText('Toggle Me!')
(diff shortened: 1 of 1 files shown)


joshhanley:
@markjaquith do you mean `waitForText()` and `waitUntilMissingText()`? https://laravel.com/docs/10.x/dusk#waiting-for-text

markjaquith:
I did... had two windows open and had `--filter` set to test a different test. Doh. Kicked it to draft. About to fix.

markjaquith:
@joshhanley There we go.

As you can tell, I'm new to Dusk... I was very pleased that I had guessed the method names correctly. But no. I was `--filter`ing the wrong test. 😬

edalzell:
> @joshhanley There we go.
> 
> As you can tell, I'm new to Dusk... I was very pleased that I had guessed the method names correctly. But no. I was `--filter`ing the wrong test. 😬

FYI, if you use VS Code, there is an extension (Better PHPUnit) that can run one test at a time, it's what I do all the time, and why my previous PR had a test failure in it...forgot to run the whole test suite.

markjaquith:
> if you use VS Code, there is an extension (Better PHPUnit) that can run one test at a time

Sure do. Thanks for the rec!

calebporzio:
Golden, thanks @markjaquith!

PR #6381, opened 2023-08-16
Add failing test for unreliable `wire:navigate` back navigation

markjaquith:
The "back" button is unreliable, after clicking a `wire:navigate` link.

I discovered this while debugging a flaky test. The flakiness seems to come from Livewire itself, not from the test.

This is a test that visits a page, clicks a `wire:navigate` link, presses back, and verifies that the original page contents are displayed. It does this multiple times. The failure is non-deterministic for me, so I'm looping the test ~~20~~ 40 times.

I experimented with injecting `->pause(1000)` between every step of the process, but that doesn't fix it... so it isn't an issue with how rapidly the navigation is happening.

Curiously, the test seems to fail about 1% of the time on the first iteration, but 50% of the time on the second iteration.

When it fails, the second page is still being displayed. Note the mismatch:

CleanShot 2023-08-16 at 18 52 59@2x


calebporzio:
Ugh thanks for this but as it's not easy to reproduce, I think I'm going to close until we can surface and reproduce the issue more reliably. Thanks for killing it on all these Mark - I just don't want this left hanging forever so I'm going to close

PR #6389, opened 2023-08-17
Rename base attribute classes

calebporzio:
Before this PR, every Livewire attribute (Rule, Url, etc.) had two files of the exact same name:

```php
Livewire\Attributes\Rule; // Vanity namespace for userland consumption...
Livewire\Features\SupportValidation\Rule; // Deeper core source code namespace...
```

After this PR, the deeper ones are renamed to `Base[Attribute Name]` as to not confuse users about which ones to use:

```php
Livewire\Attributes\Rule;
Livewire\Features\SupportValidation\BaseRule;
```

### DIFF
--- a/src/Features/SupportEvents/UnitTest.php
+++ b/src/Features/SupportEvents/UnitTest.php
@@ -13,7 +13,7 @@ class UnitTest extends \Tests\TestCase
         $component = Livewire::test(new class extends Component {
             public $foo = 'bar';
 
-            #[On('bar')]
+            #[BaseOn('bar')]
             public function onBar($param)
             {
                 $this->foo = $param;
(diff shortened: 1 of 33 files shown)


src/Features/SupportEvents/UnitTest.php
```diff
@@ -13,7 +13,7 @@ public function receive_event_with_attribute()
         $component = Livewire::test(new class extends Component {
             public $foo = 'bar';
 
-            #[On('bar')]
+            #[BaseOn('bar')]
```
PerryvanderMeer, line 16: Wouldn't it be better to use the `src/Attributes/On` for all those tests? Since users also will use those?

PR #6393, opened 2023-08-17
Add exception if parameters is not array or traversable

danidoble:
Just for know whats wrong when send data like
```js
Livewire.dispatch('something-here', 1); // user id "1"
```

otherwise, developers never know what's wrong on his code, In docs i can see sending parameters in V3 is with javascript object

```js
// In Livewire V2, works
Livewire.emit('something-here', 1); // user id "1"
```

But in upgrade to V3, is needed change too from code above to

```js
Livewire.dispatch('something-here', {id: 1}); // user id "1"
```

Thanks for contributing! 🙌


calebporzio:
Thanks for this, however, going to close this for now as I'm not sure this is the right place to warn about this.

PR #6394, opened 2023-08-17
[V3] Add dynamic validation attributes in Form Object

viliusvsx:
Added PR which allow use dynamic validation attributes inside Form Object. Currently, this behavior is working in component itself but not from Form Object. 
This PR allow use dynamic validation attributes and messages in method and do things, that in rule attribute is impossible, like use translatable names or messages, use rules based by class.

I chose to use validationAttributes() method name instead of attributes(), because it is better to stick with consistency 

P.S. Previous PR [#5904](https://github.com/livewire/livewire/pull/5904), but I had to close this, because something happened with commits


calebporzio:
Thank you!

PR #6396, opened 2023-08-17
flaky-test: can_transition_blade_conditional_dom_segments

markjaquith:
E2E test for `can_transition_blade_conditional_dom_segments` was skipped as flaky.

I fixed it by increasing the transition time and choosing tests that were a little more timing-lenient, while still confirming that the transitions are happening.


calebporzio:
Thanks a ton for these @markjaquith!!

PR #6401, opened 2023-08-18
[V3] Fix rendering/rendered hooks

joshhanley:
As per the docs, there should be rendering and rendered hooks that can be run in the component. But currently they're only firing for traits https://livewire.laravel.com/docs/lifecycle-hooks#render.

This PR adds that functionality.

Hope this helps!


calebporzio:
Thanks @joshhanley !!!

PR #6406, opened 2023-08-18
Explain class-based Volt pagination

iambateman:
It wasn't clear how to paginate a Collection in class-based Volt, since `render()` is unavailable.

This just notes to use a computed property for that.


taylorotwell:
I don't understand? You should just follow the class based pagination documentation on the Pagination section of the traditional Livewire docs.

nunomaduro:
We've updated the docs.

PR #6407, opened 2023-08-18
flaky-test: can_pass_reactive_props_to_lazyilly_loaded_component

markjaquith:
This test was marked as flaky.

One issue it had was that it was using `#[Prop(reactive: true)]` instead of `#[Reactive]` (this was causing the test to always fail).

The other issue was that it was immediately calling `->assertSee()` without waiting for the DOM to settle. It doesn't seem like `->waitForLivewire()->click('@button')` will completely wait for the DOM to settle.

>[!NOTE]
> A recurring issue with flaky tests is immediate `assertSee()` for something that may take a little while to actually be reflected in the DOM. I suggest using methods like `->waitForText()` and `waitFor()` before such assertions to give the DOM time to settle.

### DIFF
--- a/src/Features/SupportLazyLoading/BrowserTest.php
+++ b/src/Features/SupportLazyLoading/BrowserTest.php
@@ -5,6 +5,7 @@ namespace Livewire\Features\SupportLazyLoading;
 use Tests\BrowserTestCase;
 use Livewire\Livewire;
 use Livewire\Component;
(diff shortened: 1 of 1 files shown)


calebporzio:
Yeah good call - in a perfect world we have some kind of programmatic way to ensure the DOM has settled, maybe that's even a simple "setTimeout" to wait 1 tick inside ->waitForLivewire()

PR #6408, opened 2023-08-18
Remove test skips for old versions of Laravel

markjaquith:
Livewire now requires Laravel 10, so these are not necessary.


calebporzio:
Dope, thank you!

PR #6411, opened 2023-08-18
Remove unused tests

markjaquith:
These tests were both skipped and commented out with a note that they are better done as Dusk E2E tests. If there is no desire to resurrect them as unit tests, removing them reduces the "S" skipped tests noise when running the test suite.


calebporzio:
I'm down with this

PR #6413, opened 2023-08-18
[v3] Adds failing test for Laravel Echo with payload

yoeriboven:
Discussion: https://github.com/livewire/livewire/discussions/6378

The listener doesn't accept the payload sent with the event. It throws the error as seen in the screenshot.

Unwrapping the `$event` parameter using `...$event` works but it returns the values without the keys so you wouldn't have certainty you are using the right value.

261357165-fcdb4ef5-28e9-4c04-90e1-69a0fa4f6814


src/Features/SupportEvents/EchoBrowserTest.php
```diff
@@ -43,4 +43,37 @@ function render()
...
+    {
+        Route::get('/dusk/fake-echo', function () {
+            return Utils::pretendResponseIsFile(__DIR__.'/fake-echo.js');
+        });
+
+        Livewire::visit(new class extends Component {
+            public $orderId = 0;
+
+            #[On('echo:orders,OrderShipped')]
+            function foo($event) {
```
yoeriboven, line 58: Here the error is thrown.

calebporzio:
Thanks! Fixed and documented!

PR #6429, opened 2023-08-20
Add failing lifecycle test when using synths

PhiloNL:
This might be an edge case and very specific to my use case, but in v2, the boot method of a component is triggered before calling the `fromLivewire` methods on any `Wirable` objects.

In v2 I have a collection that looks something like this (simplified):
```php
class MyCustomCollection extends Collection implements Wireable
{
    //...

    public static function fromLivewire($values)
    {
        $values = collect($values)
            ->map(function ($value) {
                return collect(MyServiceProvider::$foobar)
                    ->firstWhere('type', '=', $value['type'])
                    ->fill(...$value);
            })->toArray();

        return new static($values);
    }
}

class MyComponent extends Component {
    public function boot()
    {
        // Calls a bunch of dynamic callbacks, which boils down 
        // to building various objects and storing them on a static property; simple example:
        MyServiceProvider::$foobar = [new Something(['type' => 'test']);
    }
}
```

The `fromLivewire` attempts to get a static property `MyServiceProvider::$foobar`, which contains the array of resolved objects. This static value is set every time using the boot method. Now in v3, `MyServiceProvider::$foobar` is `null` when you try to access this inside the `fromLivewire` method. 

The boot method, in v3, seems to be called **after** calling the `fromLivewire method, while in v2, it was done before.

I wasn't sure where to place the test, so these still need to be put in the correct place if this is a correct failing test.

### DIFF
--- a/src/BootTest/BrowserTest.php
+++ b/src/BootTest/BrowserTest.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Livewire\BootTest;
+
+use Livewire\Component;
+use Livewire\Livewire;
+use Livewire\Wireable;
+use Tests\BrowserTestCase;
+
+class BrowserTest extends BrowserTestCase
+{
+    /** @test */
+    public function test_component_booth_method_is_called_before_running_synths()
+    {
+        Livewire::visit(new class extends Component {
+
+            public SomeObject $someObject;
+
+            public function boot()
+            {
+                SomeOtherObjectWithStaticProperty::$someValue = 'valueSetByComponentBoot';
+            }
+
+            public function mount()
+            {
+                $this->someObject = new SomeObject('test');
+            }
+
+            function render()
+            {
+                return <<<'HTML'
+                <div>
+                    <div dusk="value">{{ $someObject->bar }}</div>
+                    <button dusk="button" wire:click="$refresh">Click</button>
+                </div>
+                HTML;
+            }
+        })
+            ->waitForLivewire()
+            ->click('@button')
+            ->waitForLivewire()
+            ->assertSeeIn('@value', 'valueSetByComponentBoot')
+            ->pause(5000)
+            ;
+    }
+}
+
+class SomeObject implements Wireable {
+    public function __construct(public $foo, public $bar = null)
+    {
+    }
+
+    public function toLivewire()
+    {
+        return [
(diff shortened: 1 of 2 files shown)


PhiloNL:
Just a quick update on this. I was able to get this to work by creating a `__construct` method inside my component and moving the code there, but I'm not sure if that's ideal? 🤔  I still think `fromLivewire` should be called after the `boot` method, given this is part of the `hydrate` process, and according to the lifecycle, `boot` has priority over `hydrate`.

joshhanley:
@PhiloNL yeah boot should run first. And yeah using a constructor isn't a great idea, it's what boot is for 😆 In regards to where to put it, check to see if there are any existing lifecycle tests.

calebporzio:
Hmmm purely from an idealistic standpoint, I think v3 has it right:

The component should be "re-hydrated" before "boot" is run. "boot" is meant to be "mount" or "hydrate" but running every time. That's why it was added in V2: so that you don't have to put the same code inside mount and hydrate.

Using the constructor in this case might not be a bad idea. It solves the problem and makes sense for something like that.

Going to close this, but keep my ears open to more reasons I'm wrong on this. Thank you Philo

joshhanley:
@calebporzio FYI in V2, boot was first always, then mount/hydrate, then booted. So if that has changed, it's probably something we need to list in the upgrade guide.

calebporzio:
Here's the relevant code from "SupportLifecycleHooks.php":
image

So to your point: code inside "boot" STILL runs before mount/hydrate.

However, at this point the component is already hydrated where in V2 it wasn't fully hydrated yet.

In V3 you're not really meant to access a component's state and such between when it's constructed and it's properties are hydrated.

I think it makes for a more solid system and experience, but I get there's discrepancies like this one.

Have we heard of others getting hung up on this besides @PhiloNL?

PhiloNL:
I haven't seen any other people report this specifically besides some similar issues where values are `null`. Looking at the lifecycle in relation to the synths, I still think `fromLivewire` should be called after the `boot` method, given this is part of the hydrate process, and according to the lifecycle, boot has priority over hydrate.

But I can imagine this might be a discrepancy, in that case, I'm fine using the constructor if that works and won't cause any problems.

PR #6449, opened 2023-08-22
flaky-test: can_transition_blade_conditional_dom_segments (again)

markjaquith:
I made it less-flaky the first time, but I still saw it fail in CI [here](https://github.com/livewire/livewire/actions/runs/5927960803/job/16072569596).

This time it should really be good. I increased the transition to 2000ms and used `waitUntil()` with JavaScript expressions to let Dusk manage the timing of the tests, instead of trying to manually `pause()` for the right amount of time. It could still theoretically fail if somehow it takes 2 seconds to execute the DOM check, but it seems unlikely. I ran this hundreds of times on my machine with no issue.


calebporzio:
Amazing - thank you Mark

PR #6450, opened 2023-08-22
Add falling test for default model relations 

fripig:
fix for #6141 

turn on legacy_model_binding then withDefault model got fail


joshhanley:
@fripig can you confirm, did `withDefault()` work with models in V2?

fripig:
I'm sure it can work in v2.
I use it on the feature I use most often,
the editing content function of CMS.
So, this issue was discovered during the upgrade process.

calebporzio:
Hmmm I think I know what's going on here:

Livewire 3 doesn't support storing un-persisted models as properties (yet at least). Therefore, "withDefault()" would return a new model that hasn't been persisted and Livewire wouldn't know how to handle it.

I'm going to close this as there is another PR open about this: #6272

PR #6453, opened 2023-08-22
Fix wrong `is` attribute on dynamic components

flexchar:
It appears that docs are wrong on dynamic components and `component` must be used instead of `is` to specify component name. I bumped into this myself.

Relevant discussion: https://laracasts.com/discuss/channels/livewire/livewire-dynamic-component?page=1&replyId=902066

Cheers :)


joshhanley:
@flexchar in V2 we supported both, so I believe there is an issue with the implementation in V3, as both should work 🙂

flexchar:
So, guys, this means:
1. I have to update docs to have both examples.
2. The issue has to be opened because it didn't work with `:is` attribute.

Agreed?

calebporzio:
Good points. I added support for `:is` and also added a new syntax (with documentation) for:

```html
<livewire:is component="some-component" />
```

Which I think I like the best actually.

Thanks!

PR #6466, opened 2023-08-23
[V3] Failing test: Slots with the same name are bubbling up

riasvdv:
It's common to have a layout with a `$title` slot, and a component in your app which also accepts a `$title` slot.

However Livewire currently causes the component's slot to bubble up to the layout.

Given a layout with:

```blade
<html>
<body>
    {{ $title ?? 'No title' }}
</body>
</html>
```

and a component:

```blade
<div>
    {{ $title ?? 'Default component title' }}
</div>
```

With a view:

```
<x-component-with-slot>
    <x-slot name="title">
        The component title
    </x-slot>
</x-component-with-slot>
```

We should see "The component title", and "No title", but the view renders with 2x "The component title"

I've added a failing test below trying to reuse a few of the existing slot views


freekmurze:
This issue is a showstopper for the releases of Media Library Pro and Laravel Comments. As soon as this is fixed we'll release major new versions for these packages.

edalzell:
I gotta ask, why the `[-1]`?

nuernbergerA:
@edalzell https://github.com/laravel/framework/blob/909ea24597b86cfc7a49ce41074280d0347bf6a3/src/Illuminate/View/Concerns/ManagesComponents.php#L207

PR #6470, opened 2023-08-23
[V3] update the event name in navigate document

yilanboy:
Navigating event name seems not correct in navigate document page.

```js
// event name should be 'livewire:navigating'
document.addEventListener('livewire:navigating', () => {
    // Mutate the HTML before the page is navigated away...
})
```


calebporzio:
Ooops, thanks!

PR #6471, opened 2023-08-23
[V3] Add `assertReturned` method to test return value of a method

gdebrauwer:
Currently there is no easy way to assert the return value of method:

```php
{
    public function someMethod()
    {
        return 'foo'; // How to test that 'foo' is returned when this method is called?
    }
}
```
This PR adds an   `assertReturned` method to make that possible.
```php
   ->call('someMethod')
   ->assertReturned('foo'); //  OR  ->assertReturned(fn ($value) => $value === 'foo');
```


calebporzio:
Thanks!

PR #6475, opened 2023-08-23
[v3] Fix/Cleanup dynamic component

nuernbergerA:
* replaced the dead `app('livewire')->getClass()`
* optimized for loop (count in condition runs on each step in the loop so we count before)
* fixed the check if there is a component named "is"or "dynamic-component" if there are no arguments provided @calebporzio wouldn't it be time to just make check in the component registry and throw there an error that its not allowed to use those component names?

### DIFF
--- a/src/Mechanisms/CompileLivewireTags.php
+++ b/src/Mechanisms/CompileLivewireTags.php
@@ -78,8 +78,7 @@ class CompileLivewireTags extends ComponentTagCompiler
                 // Does not need quotes as resolved with quotes already.
                 $component = $attributes['component'] ?? $attributes['is'];
 
-                unset($attributes['component']);
-                unset($attributes['is']);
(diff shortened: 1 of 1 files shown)


joshhanley:
@nuernbergerA thanks for the PR! Caleb has just fixed it in https://github.com/livewire/livewire/commit/7b3e98567033208175be84856d72b237f9c5821f so going to close this.

nuernbergerA:
@joshhanley it fixes the part where it detects if it's a real livewire component.

joshhanley:
@nuernbergerA That can be removed, now thinking about it. I only put it there to stop it being a breaking change for v2 as some people already had components called dynamic-component.

nuernbergerA:
Ok I will change that and reopen.

nuernbergerA:
@joshhanley see #6484

PR #6476, opened 2023-08-23
[V3] Cleanup version checks

nuernbergerA:
* removed php 7.4 checks as 8.1 is now the minimum version
* removed test checks for pre laravel 10 versions as its now the minimum


joshhanley:
Thanks!

PR #6478, opened 2023-08-23
[V3] Return null on IntSynth when null is passed in

ejonasson:
The `IntSynth` in Livewire V3 adds the ability to convert an empty string that's passed into the synth to `null`. This is great on the frontend (since a form input will generally return `''` instead of `null`, but there are still scenarios where `null` might be returned. For example, in a unit test (as demonstrated in the test on this PR), or when updated though Javascipt.

This PR just makes sure we pass though that `null` value in those cases, as opposed to passing it to the `int` cast, which converts it to `0`. 

Did not create a discussion, but an explanation is included above.

Yes

The only changes are to the `IntSynth` class, and adding a test for that Synth.

Yes

Included above :)

### DIFF
--- a/src/Mechanisms/HandleComponents/Synthesizers/IntSynth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/IntSynth.php
@@ -15,7 +15,7 @@ class IntSynth extends Synth {
     }
 
     static function hydrateFromType($type, $value) {
-        if ($value === '') return null;
+        if ($value === '' || $value === null) return null;
 
         return (int) $value;
     }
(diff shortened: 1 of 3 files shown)


joshhanley:
@ejonasson thanks for the PR! I just pulled it down, and commented out your fix and the test still passes. Can you confirm the same happens for you?

joshhanley:
@ejonasson actually don't worry, I managed to fix it 🙂

ejonasson:
@joshhanley Oops, sorry about that. Thank you for fixing!

calebporzio:
Thank you!

PR #6483, opened 2023-08-24
Rewrite morph marker injection system

calebporzio:
The previous system was too complex and slow.

This is a rewrite to make the regex much simpler, much faster, and much more comprehensive.

Now, markers will be injected around all Blade control directives (@ if, foreach, forelse, error, guest, admin, etc..) including custom registered `Blade::if()` directives in people's apps.

### DIFF
--- a/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php
+++ b/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php
@@ -2,12 +2,12 @@
 
 namespace Livewire\Features\SupportMorphAwareIfStatement;
 
-use Illuminate\Support\Facades\Blade;
+use Livewire\Livewire;
 use Livewire\ComponentHook;
+use Illuminate\Support\Facades\Blade;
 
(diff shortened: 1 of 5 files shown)


saschagysel:
@calebporzio Since 3.0 final (in Beta it was not a problem) I get a JS error when I create a JSON / array data structure directly in a JS script in a livewire blade file. Here is an example: 

```
<script>
let players = [
   @foreach ($players as $player)
        {
             label: "{{ $player->name }}",
             value: "{{ $player->id }}"
        },
    @endforeach
];
</script>
```

The output of this is:

```
let players = [
     <!-- __BLOCK__ -->
    {
          label: "Player X",
          value: "21"
    },
    { 
         label: "Player Y",
         value: "27"
      }
<!-- __ENDBLOCK__ -->
]
```

This makes the JS array invalid because of the morph markers. As a fix it would be great if we could disable the markers on a component base. I know there is a global setting for disabling morph markers but I think it would be better to only disable them where they interfere.

nuernbergerA:
@saschagysel any changes you can submit a PR with this as a failing test?

PR #6484, opened 2023-08-24
[v3] Cleanup dynamic component

nuernbergerA:
* optimized for loop (count in condition runs on each step in the loop so we count before)
* removed check for livewire components named 'dynamic-component' or 'is'


calebporzio:
Why this? "removed check for livewire components named 'dynamic-component' or 'is'" because it's uneede?

calebporzio:
Josh just filled me in - thanks @nuernbergerA

PR #6485, opened 2023-08-24
[v3] Remove unused precompiler

nuernbergerA:
`$precompile` is unsued because the facade is called directly
https://github.com/livewire/livewire/blob/c534fca8e219d807b2a3228a0fe75b69182df4f9/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php#L57


calebporzio:
ooops, good catch, thanks!

PR #6487, opened 2023-08-24
[V3] Give wireable synth higher priority over other synths

joshhanley:
In V2 wireables had priority over all over types. This PR make anything with the wireable trait use wireable synth over the defined type, see discussion #6346.

Hope this helps!


calebporzio:
Thanks Josh

PR #6491, opened 2023-08-24
Lazy refactor to support lazy on load with attributes also, and fix route macro

nicolagianelli:
I merged my branch with #5780 to add lazy on load for attributes.
I also changed how the route macro works because the test was working but when I tested it locally it wasn't working.
Yes, I updated the merged branch
No
Yes

I already updated the docs


joshhanley:
@nicolagianelli thanks for the PR! I don't exactly follow what this PR is for. Any chance you could explain it in more detail? Also if this PR fixes multiple issues, can you split them into seperate PRs to make it easier to review and merge? Thanks!

nicolagianelli:
This PR is mainly to add lazy on-load support for #[Lazy] attributes, if someone wants to set a component to lazy load "on load" instead of when in viewport.
Now If I set #[Lazy] on the component it will load when in the viewport and I can't set to the "on-load" behaviour instead.

The fix, was because when I tried to use the lazy() route macro to load a full page component it wasn't working even if the test wasn't failing so I changed the implementation.

calebporzio:
Thank you for the PR. Going to hold off on this feature until there is more demand. THANKS!

JacoTome:
I'm arriving late to the discussion, is this pull request still closed? It would be helpful for me atm.

PR #6494, opened 2023-08-24
[V3] Toggleable query string

nicolagianelli:
In V2 was possibile to override the getQueryString() function and with that we were able to pass to our components a prop to disable the use of query string (pagination, etc) so in a dashboard page we could reuse our components multiple times without them influencing the query string (otherwise when the dashboard was reloaded with page refresh, all the components used the same page)
Yes
No
Yes

We used a custom prop with the getQueryString() override in V2 to disable the use of query string on our components when needed in some pages like dashboards to avoid them influencing the query string and go all to the same page after page refresh.
Now I tried that and it isn't working anymore in V3, so I created a custom param named "query-string" and a custom Attribute to define default use of query string in components (default true) so when someone needs to use a component in a dashboard page or similar we can pass the :query-string="false" to disable it for Url Attributes and Pagination.
We can also define #[QueryString(false)] in the component class and use :query-string="true" only when we need it.

I already updated the docs, also, if you don't like this implementation for me it's ok to change it but we need this or something similar to replicate what we did in V2.

Thank you!

### DIFF
--- a/docs/pagination.md
+++ b/docs/pagination.md
@@ -375,3 +375,9 @@ As you can see, you can use Livewire's page navigation helpers like `$this->next
 </div>
 ```
 
+### Disable query string pagination
+
+You can disable query string functionality for pagination and variables with Url attribute.
+<br>
+<a href='https://livewire.laravel.com/docs/url#disable-query-string'>Check how from URL Query Parameters</a>
+
--- a/docs/url.md
+++ b/docs/url.md
@@ -146,7 +146,7 @@ To force Livewire to use `history.pushState` when updating the URL, you can prov
 
 ```php
 use Livewire\Component;
-use Livewire\With\Url;
+use Livewire\Attributes\Url;
 
 class ShowUsers extends Component
 {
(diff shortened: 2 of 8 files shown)


joshhanley:
@nicolagianelli thanks for the PR! I'll be honest, I don't exactly follow what you're trying to achieve here. If it's not too much trouble, would you be able to provide a demo V2 repo that shows how you were using it before, so we can have a look? Thanks!

calebporzio:
Thank you for your contribution. I understand what you're trying to achieve here, but I'm not sure it's the right API or implementation so I'm going to close this.

I think the right API for this would be something like:

```
{
    public function render()
    {
            'users' => User::paginate(10)->withoutUpdatingUrl();
        );
    }
}
```

You can add this method by macroing to the "Collection" class as the "lengthawarepaginator" is returned from `paginate()` and forwards calls to the underlying collection.

Then, possibly, you can set some sort of option to implement the feature, although I haven't looked into the specifics.

Also, it looks like many tests were added that didn't need to be, so please try to make the feature as minimal and simple as possible so it's easier to merge. Thanks!

nicolagianelli:
@calebporzio 
I understand what you're saying, but an implementation over the lengthawarepaginator, instead of one more generic, it won't work for #[Url] attributes... and, probably, when someone wants to disable the url changes for the paginator is because all the url changes aren't wanted for that component in a specific context (not in all contexts).
The implementation I made worked for all the changes on query string (by paginator or/and #[Url] attributes)

PR #6497, opened 2023-08-24
[V3] Refactored js/directives.js

sakanjo:
This PR refactors a small block of code that was rewritten in the same way as [js/directives.js](https://github.com/livewire/livewire/blob/main/js/directives.js#L8) on line 8.


joshhanley (approved):
@sakanjo thanks for the PR! @calebporzio looks good to me, assets will need to be merged from main.

calebporzio:
Thanks!

PR #6499, opened 2023-08-24
[V3] Improve phpunit instructions in CONTRIBUTING.md

skeemer:
The instructions for running specific tests didn't work.


joshhanley:
@skeemer thanks for the PR! I just tested and all we need to do is remove the `--filter` so the command is `./vendor/bin/phpunit path/to/your/test.php`. That keeps it simple. If people need to do more, they can look up the PHP Unit documentation 🙂

calebporzio:
You're right Josh. Thanks!

PR #6501, opened 2023-08-24
[V3] Fix a minor bug by adding a null check for S3 filesystem disk root to `FileUploadConfiguration`

Rezrazi:

https://github.com/livewire/livewire/discussions/5715

The original function made an assumption that the root would be never null if available in the config, but Laravel's Filesystem implementation allows for an empty root, and transforms it to an empty string later on.

This should enable having a nullable root and assign it properly to an empty string.

The use case for this, is that I have the same application deployed on multiple environments, and the root is variable between a string and a null value

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportFileUploads/FileUploadConfiguration.php
+++ b/src/Features/SupportFileUploads/FileUploadConfiguration.php
@@ -64,15 +64,11 @@ class FileUploadConfiguration
 
     protected static function s3Root()
     {
-        if(!static::isUsingS3()){
-            return '';
-        }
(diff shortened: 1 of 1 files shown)


Rezrazi:
@joshhanley Hi, any updates to this?

I would appreciate if someone can take a look as this has already been a bug in V2 https://github.com/livewire/livewire/pull/5716

Thank you 🙏

joshhanley:
@Rezrazi thanks for the PR! This is something Caleb will need to review 🙂

calebporzio:
Thanks!

PR #6504, opened 2023-08-24
[V3] Events inline scripts - docs

nicolagianelli:
The events via inline scripts with @this.on and @this.dispatch/@this.dispatchSelf are still usable and not documented so I updated the docs.


calebporzio:
Good context. Thank you!

PR #6509, opened 2023-08-24
[V3] Make \Livewire\Livewire::forceAssetInjection() override config livewire.asset_injection = false

skeemer:
This PR improves the tests for SupportAutoInjectedAssets and fixes the problem where \Livewire\Livewire::forceAssetInjection() would be overridden by the config property 'asset_injection' being set to false.

Discussion: #6355


calebporzio:
Perfect, thanks!

PR #6512, opened 2023-08-25
[V2] Add failing tests for `TestableLivewire`

austincarpenter:
Firstly not sure if v2 is being maintained due to v3 being released only some hours ago – great timing on my part.

Found a memory leak when running multiple Livewire tests in succession. This could be somehow by design but the static `$instancesById` array property on `TestableLivewire` is never reset back to empty, meaning that it continues to grow throughout the test suite.

Pretty sure it's needed. Haven't made a discussion.

Yes

No

Yes

As above.

Thanks for contributing! 🙌

### DIFF
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -259,4 +259,9 @@ class TestableLivewire
         
         return $this;
     }
+
+    public function __destruct()
+    {
+        unset(self::$instancesById[$this->id()]);
+    }
 }
--- a/tests/Unit/TestableLivewireClearsInstancesTest.php
+++ b/tests/Unit/TestableLivewireClearsInstancesTest.php
@@ -1,0 +1,26 @@
+<?php
+
+use Livewire\Component as BaseComponent;
+use Livewire\Livewire;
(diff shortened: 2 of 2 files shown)


austincarpenter:
Adding the following to the end of `TestableLivewire` partially solves the problem with the repo's test suite, cutting about 316 left over components down to in my case, 19 by the time the failing test I've added gets executed. I believe this is probably due to a race condition of not clearing the previous instances from memory in time before the test begins, but I could be wrong.

```php
protected function __destruct()
{
    unset(self::$instancesById[$this->id()]);
}
```

Either way this reduced the memory usage on the test suite from 32MB to 24MB, a 25% reduction. This would be far higher in a real world project.

Will leave the (still) failing test in for now as there is likely a better alternative to the above.

austincarpenter:
For now, I'm adding the following inside my `Pest.php` file.
```php
->afterEach(function () {
    (new ReflectionClass(TestableLivewire::class))->setStaticPropertyValue('instancesById', []);
})
```

calebporzio:
Good addition - thanks!

austincarpenter:
No worries! 😀

PR #6513, opened 2023-08-25
[V2] Add config option to disable s3 direct download links

apreiml:
The S3 direct download feature is always enabled, if the disk driver is s3. This PR introduces an option that allows to disable direct download, so that down and uploads always go through the livewire controllers.

This feature allows to use s3 or an s3 compatible system like minio, without exposing it to the user. It's useful when for example the minio storage is in the internal network behind a firewall.

If this is accepted, I can also port this to livewire 3.x

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -106,6 +106,7 @@ return [
             'jpg', 'jpeg', 'mpga', 'webp', 'wma',
         ],
         'max_upload_time' => 5, // Max duration (in minutes) before an upload gets invalidated.
+        's3_direct_url' => true,
     ],
 
     /*
--- a/src/Controllers/CanPretendToBeAFile.php
+++ b/src/Controllers/CanPretendToBeAFile.php
@@ -4,10 +4,9 @@ namespace Livewire\Controllers;
 
 trait CanPretendToBeAFile
 {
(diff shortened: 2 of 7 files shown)


apreiml:
I've started a discussion here https://github.com/livewire/livewire/discussions/6778

calebporzio:
Going to hold off on this feature because of the lack of demand. If more people come through with desire for this feature we can consider it later on. Thanks!

apreiml:
@calebporzio Fair enough.

What about only the commit https://github.com/livewire/livewire/pull/6513/commits/e35a638552db42cd1e6d935439f669eead8d0555 which allows file previews to work with any other filesystem that is not local or s3?

lmo-wt:
We are interested with this! Our s3 bucket is private and not exposed. Yet, we want to store temporary files on s3 because working with local temporary files actually breaks when having mutiple instances of our app, as subsequent request may hit on an instance that do not have the locally temporary file.

apreiml:
@lmo-wt with https://github.com/livewire/livewire/pull/7889 being merged you can do this by wrapping the s3 store in a normal FilesystemAdapter. See https://github.com/livewire/livewire/discussions/7466#discussioncomment-8020637

PR #6519, opened 2023-08-25
[V3] Don't trigger navigate on CMD + click

riasvdv:
A CMD+Click (or CTRL+click on windows) is the same as a middle click, we don't want to trigger navigate in this case but let the browser open a new window/tab

I've added the other modifiers as a check as well, referencing what InertiaJS defines for their Link element as "interceptable": https://github.com/inertiajs/inertia/blob/master/packages/core/src/shouldIntercept.ts


riasvdv:
I just ran `npm install` and `npm run build` but it seems like a lot more changed than expected in the compiled assets. 

@joshhanley Any idea what I did wrong?

calebporzio:
Thanks a bunch!

PR #6521, opened 2023-08-25
[V3] Fix: Remove $entangled components from DOM without side effects

danharrin:
Copy of #6289 using `$entangle()` instead of `entangle()`


danharrin:
Alright I think I fixed it, having trouble building the JS though

joshhanley:
@danharrin seems like the package.lock was still referencing an old version of Alpine and plugins. So I have updated them to v3.13.0 and rebuilt the assets, which seems to have fixed the issue.

joshhanley:
Failing test is unrelated. Am fixing on main and will merge in once sorted.

PR #6309, opened 2023-08-10
[V3] failing test: removing entangled watcher

danharrin:
Review the contribution guide first at: https://laravel-livewire.com/docs/2.x/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

Filament issues

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)

Yes

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No

4️⃣ Does it include tests? (Required)

Yes

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

Entangled watchers are not being destroyed when they are removed from the DOM, and get run even after all the Alpine properties in the component object are returning null.

Thanks for contributing! 🙌

PR #6528, opened 2023-08-25
Add failing test for nested entangled items duplicating requests

archilex:
When there are `entangled` items in an array, removing one or more of them causes duplicate requests to be made. If there are entangled and non-entangled items in the array, then removing _any_ of the items will also cause duplicate requests to be made.

Expected behavior: only a single request should be made.

See https://github.com/filamentphp/filament/issues/8077 for an example of this happening in Filament.

Could this also be related to #6309?


archilex:
Updated to remove unneeded nested component I forgot to remove when developing the test.

satoved:
Encountered this issue when updating to V3 too.

joshhanley:
@archilex thanks for the PR! Looks like after merging in the latest main branch changes (after I fixed the failing tests there), that this failing test is now passing 😁 I'll leave it up to Caleb as to whether he wants to merge this test in or not.

joshhanley:
@archilex thanks! Caleb was happy to include the test 🙂

PR #6531, opened 2023-08-26
[V3] Fix into specific validation rule example on docs

vverardO:
fixing the giving example for asserting if a component has a specific rule error.


joshhanley:
@vverardO thanks for the PR! I think the original syntax is correct. You can also see it here on the validation page https://livewire.laravel.com/docs/validation#testing-validation

image

PR #6532, opened 2023-08-26
[V3] Feature Enhancement - Pass component parameters as an argument to placeholder() method

derekcavaliero:

**See discussion:** 
https://github.com/livewire/livewire/discussions/6495#discussioncomment-6827650

`features/add-component-attributes-to-placeholder`

Nope

**Added new test:**
```
phpunit --filter can_access_component_parameters_in_placeholder_view ./src/Features/SupportLazyLoading/BrowserTest.php
```

Having access to the components attribute values inside the placeholder method/view can be helpful when you need to output more than one skeleton loader inside a grid layout and/or adjust layout classes that wrap around each skeleton loader.

Let's assume I have a component that fetches data from a slow API and uses the `lazy` attribute:

```blade
<livewire:article-list 
    lazy
    per-page="4" 
    header-element="h3"
    wrapper-class="g-col-12 g-col-sm-6 g-col-xxl-3"
/>
```

And I want to use the `per-page` and `wrapper-class` attribute values to change the output of the view returned inside my components `placeholder()` method:

```php
// App\Livewire\ArticleList.php
public function placeholder()
{
    return view('livewire.placeholders.article-list');
}
```

There is no current way to access the livewire component parameters inside the `placeholder()` method to do this kind of thing. 

This PR will pass an `$params` argument to the `placeholder()` method which can then be either passed directly to the view or can be accessed and used as needed.

```php
// App\Livewire\ArticleList.php
public function placeholder(array $params = [])
{
    return view('livewire.placeholders.article-list', $params);
}
```

```blade
{{-- resources/views/livewire/placeholders/article-list.blade.php --}}
<div class="grid g-col-12">
  @for ($i = 0; $i < $perPage; $i++)
  <div class="{{ $wrapperClass }}">
    <x-cards.article skeleton />
  </div>
  @endfor
</div>
```

This offers a lot of flexibility for complex layouts that require more than an single placeholder element, and can allow placeholder views to dynamically change based on the main component parameters being passed.

### DIFF
--- a/docs/lazy.md
+++ b/docs/lazy.md
@@ -86,6 +86,19 @@ class Revenue extends Component
 
 Because the above component specifies a "placeholder" by returning HTML from a `placeholder()` method, the user will see an SVG loading spinner on the page until the component is fully loaded.
 
+### Rendering a placeholder via a view
+
+For more complex loaders (such as skeletons) you can return a `view` from the `placeholder()` similar to `render()`. 
+
+```php
+public function placeholder(array $attributes = [])
+{
+    return view('livewire.placeholders.skeleton', $attributes);
+}
+```
+
+Any attributes from the component being lazy loaded will be available as an `attributes` arugment passed to the `placeholder()` method.
+
 ## Lazy loading outside of the viewport
 
 By default, Lazy-loaded components aren't full loaded until they enter the browser's viewport, for example when a user scrolls to one.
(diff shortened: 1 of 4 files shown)


derekcavaliero:
I did notice that some unrelated existing tests were failing for me when running all of them - not sure what to do about those (perhaps they're known failures).

joshhanley:
@derekcavaliero I think it would be better to name them params (like in your discussion) rather than attributes, as to not confuse them with Blade attributes (which Livewire components currently don't support).

derekcavaliero:
Agreed I'll make that change.

calebporzio:
This is great, thank you!

PR #6534, opened 2023-08-26
[V3] Updated code in a quotation block installation.md

nasirkhan:
This is a proposed fix to the issue mentioned in https://github.com/livewire/livewire/discussions/6526

In the quotation block the code is not visible on the docs site. Converting ```   to `<code>` might fix this.


joshhanley:
@nasirkhan thanks for the PR! Turns out it was because blade was trying to parse the tag, so adding `@verbatim` fixes it.

PR #6535, opened 2023-08-26
[V3] Fix navigate test to detect OS and determine which key to press

joshhanley:
Navigate test for cmd + click is failing in CI due to it being a linux environment. Have added a check to determine what OS is being used and to use the appropriate key. Have tested it locally on a mac and it passes.


calebporzio:
Dope, thank you @joshhanley !

PR #6536, opened 2023-08-26
[V3] Support @empty directive

klaasgeldof:
This PR adds back support for the `@empty` directive.

We only insert the morph markers if `@empty` is followed by an expression (= optional whitespace followed by opening parenthesis).


klaasgeldof:
Some test failed but only on PHP 8.2. I think it doesn't have anything to do with what I changed...

calebporzio:
Thanks! Some changes have been made to the regex - if you can merge main into this I'll merge it

PR #6549, opened 2023-08-27
Use the `route()` helper to get the update URI

stancl:
> 1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first? **Yes**
> 
> 2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed) **Yes**
> 
> 3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out. **No**
> 
> 4️⃣ Does it include tests? (Required) **No** — changed code doesn't seem to be tested. If it were I'd like to add a regression test for the bug being fixed here.

See https://github.com/livewire/livewire/discussions/6119. This fixes a bug where update URIs with extra parameters were unusable.

TLDR of the linked discussion:
- Livewire docs mention that the customizable update URI may be used for cases like multi-tenancy, however in those cases adding an extra parameter may be desired: `/{tenant}/livewire/update`
- `$updateRoute->uri` (current code) doesn't take into account parameters
- Calling `route()` on the route provides a way for route parameters to be handled (e.g. `URL::defaults()` or url generator overrides)


stancl:
@joshhanley Any clue why the tests are failing? The change works just fine in actual Laravel apps, so it might be related to something test-specific.

calebporzio:
Hey @stancl - thanks for the PR and report!

I went to add a test for you and couldn't really write a passing one because I'm not sure I totally understand the scenario.

Can you weigh in by either modifying my test definition to pass or better explaining where I'm going wrong. Thanks!

stancl:
Writing a detailed answer to fully explain how this would be used in a real app, but feel free to skip to the last two paragraphs. The answer ended up way longer than I intended 😬

> I'm not sure I totally understand the scenario

A simple example of how this can be used with multi-tenancy is that you'd have a middleware that detects the tenant from the `{tenant}` parameter (and then drops it so that it doesn't get passed to the controller, but that detail can be ignored here). Once the tenant is detected, it will run something like `URL::defaults(['tenant' => $currentTenantId])` so that all generated routes will have the `{tenant}` parameter filled.

So in the case of Livewire, you would:
- visit `/{tenant}/foo` which contains a `Foo` Livewire component. The tenant would be detected as id `2` when visiting `/2/foo`, and that'd be applied as the default value for generated routes. So on that page, if you had a route to `/{tenant}/bar`, it'd be `/2/bar`
- the same thing would apply to the Livewire update URI. The route (in Laravel) would be `/{tenant}/livewire/update`, and since the tenant was identified as `2`, it will be `/2/livewire/update` on this page
- when Livewire makes subsequent requests, they will always go to `/2/livewire/update`, letting the app read the `{tenant}` parameter and run any needed multi-tenancy logic

Regarding failing tests, what do you mean exactly? The tests seem to pass here.

That said, I *think* `can_use_a_custom_update_route_with_a_uri_segment` should be failing, since the value for `{custom}` is never defined/provided. My guess would be that the component mocking API doesn't use the update routes, and instead executes updates on the component directly (since all the logic takes place on the backend, so there's little need to emulate requests). But that's something you'd be more familiar with.

If you wanted to fully test this in a way that'd emulate how this works in a real multi-tenant app, you'd need to:
- make changes via the update routes
- add a middleware that reads the custom value from the `{custom}` parameter and sets `URL::defaults()`
- add a listener for dropping forgetting the parameter from the route once it's processed by the middleware (in controllers this is an issue since it'd pass the parameter to every route action as the first parameter, but in Livewire it might not matter, not sure)

I think an easier approach to testing this would be simply checking the update URI string that will ultimately be used on the frontend. So you'd just set an update route with a `{custom}` parameter, set a value for that parameter using `URL::defaults(['custom' => 'foo'])`, and then assert that the frontend runtime was instructed to make requests to `/foo/livewire/update`.

It's the user who's setting the custom route, so whether this actually works (aka has all the needed middleware setup etc) is in a sense their responsibility. Livewire should just make sure that it will generate correct update URIs when:
- a route with a parameter is used (`/{custom}/livewire/update`)
- the route is provided with a value (e.g. using `URL::defaults()`)

stancl:
Sure that's fine, appreciate you being invested in making multi-tenancy work with Livewire.

Thanks for writing the test template, I made some changes and wrote assertions in the test. It now checks that the component works, that it's using the right route, and that the tenant is always identifiable (present in route parameters).

There's a few hacky things I had to add, namely the `URL::defaults(['tenant' => 'custom-tenant'])` and `app('router')->getRoutes()->refreshNameLookups()` inside `tweakApplicationHook()`, but it seems that this is the most straightforward way to be able to test this. Other approaches would be significantly more complex I think.

> in a way where it doesn't pass without your addition

I now tested the latest version of the PR with `getUpdateUri()` reverted to its original version and the tests do fail, so I can confirm that the tests work as intended, and require the changes made here to pass.

PR #6553, opened 2023-08-27
Fix SupportFileDownloads - Multibyte character support

nxsdev:
Yes
https://github.com/livewire/livewire/discussions/4946 
This is not a discussion I created, but another person put it up.
Yes
No
Yes

```php
if (preg_match('/filename\*=utf-8\'\'(.+)$/i', $header, $matches)) {
    return rawurldecode($matches[1]);
}
```

I added the above code to `livewire/src/Features/SupportFileDownloads.php` above. Currently, there was a problem that file names containing Japanese characters would end up as `______.csv` when trying to download them. We believe this needs to be fixed, especially since Livewire is a very popular library in Japan.
We have improved the problem by adding a decoding process in the case of filename*, and confirmed that it works with UnitTest.

I have tested this modification on Laravel in the following environment
`PHP 8.2.9`
`Laravel 10.22.0`

https://github.com/livewire/livewire/discussions/4946
I tested with `Ελληνικά.xlsx` and Japanese `ダウンロード.csv` in the same situation as

Before the modification, as shown below, `________.xlsx` and `______.csv` with multibyte characters did not download with the correct file name.

[Image from Gyazo](https://gyazo.com/5ff0ee890766bbedcfa6cfe199fe2119)

With this correction, we can confirm that we were able to download the file with the correct file name.
Below is a screenshot of the corrected file.

[Image from Gyazo](https://gyazo.com/fc4f4c53f4ce9bf6e620890c442ad896)

### DIFF
--- a/src/Features/SupportFileDownloads/SupportFileDownloads.php
+++ b/src/Features/SupportFileDownloads/SupportFileDownloads.php
@@ -70,7 +70,6 @@ class SupportFileDownloads extends ComponentHook
          *
          * Content-Disposition: attachment; filename=filename.jpg
          * Content-Disposition: attachment; filename="test file.jpg"
-         * Content-Disposition: attachment; filename*=utf-8''%E3%83%80%E3%82%A6%E3%83%B3%E3%83%AD%E3%83%BC%E3%83%89%E7%B5%90%E6%9E%9C.csv
          */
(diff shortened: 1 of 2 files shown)


noxis-dev:
It appears that you need to get approval from the maintainer. Can anyone help me get approval?

calebporzio:
Thanks!

PR #6556, opened 2023-08-28
[V3] Fixes Lazy Load By Default Exception - Undefined array key "lazy"

Cbrad24:
This PR fixes Lazy Loading throwing an exception under a certain use case.

According to the [Docs](https://livewire.laravel.com/docs/lazy#lazy-load-by-default) you can use Lazy Loading by default, and it implies you can omit the `lazy` parameter from your Blade View after defining the `#[Lazy]` attribute.

Unfortunately doing so will result in an error throwing `Undefined array key "lazy"` when it does the check to see if the lazy load should be an `intersect` or `init` event.

This simply adds a check to make sure the array key is indeed set and the value is `on-load` before using `x-init`.

The current **work-around** is to do the following redundant declaration:
```php
#[Lazy]
class Revenue extends Component
{
    // ...
}
```
```blade
<livewire:revenue lazy />
```

### DIFF
--- a/src/Features/SupportLazyLoading/SupportLazyLoading.php
+++ b/src/Features/SupportLazyLoading/SupportLazyLoading.php
@@ -102,7 +102,7 @@ class SupportLazyLoading extends ComponentHook
             ->placeholder();
 
         $html = Utils::insertAttributesIntoHtmlRoot($placeholder, [
-            ($params['lazy'] === 'on-load' ? 'x-init' : 'x-intersect') => '$wire.__lazyLoad(\''.$encoded.'\')',
+            ((isset($params['lazy']) and $params['lazy'] === 'on-load') ? 'x-init' : 'x-intersect') => '$wire.__lazyLoad(\''.$encoded.'\')',
         ]);
 
         return $html;


nicolagianelli:
The refactor of lazy in #6491 should already fix this, and also add support for on-load with lazy attribute.
Just in case try it to be sure and let me know.

Cbrad24:
> The refactor of lazy in #6491 should already fix this, and also add support for on-load with lazy attribute. Just in case try it to be sure and let me know.

Haven't had a chance to test yet but judging by your code yes this would fix the issue and also bring in what feels like the Lazy attributes are missing support for. Will try test tomorrow.

If #6491 is accepted then this can probably be closed, otherwise note this does fix an exception.

iAdil:
> The refactor of lazy in #6491 should already fix this, and also add support for on-load with lazy attribute. Just in case try it to be sure and let me know.

Yes, this fix resolves the error

src/Features/SupportLazyLoading/SupportLazyLoading.php
```diff
@@ -102,7 +102,7 @@ public function generatePlaceholderHtml($params)
             ->placeholder();
 
         $html = Utils::insertAttributesIntoHtmlRoot($placeholder, [
-            ($params['lazy'] === 'on-load' ? 'x-init' : 'x-intersect') => '$wire.__lazyLoad(\''.$encoded.'\')',
+            ((isset($params['lazy']) and $params['lazy'] === 'on-load') ? 'x-init' : 'x-intersect') => '$wire.__lazyLoad(\''.$encoded.'\')',
```
nuernbergerA, line 105:
```suggestion
            ($params['lazy'] ?? '' === 'on-load' ? 'x-init' : 'x-intersect') => '$wire.__lazyLoad(\''.$encoded.'\')',
```

calebporzio:
Thank you!

PR #6559, opened 2023-08-28
Fix wireables not being able to be updated

levu42:
This is a test for what's described in #5989 - basically updating Wireables where only partial data changed doesn't currently work as of 3.0.1 and it seems like this wasn't covered by tests before.

I created three tests: a wireable unit test & two browser tests.

- The wireable unit test first shows that updating the whole object works - good.
- The wireable unit test then shows that updating a single property inside a wireable doesn't work - as expected.
- The first browser test then shows, that even though I try to update the whole object on client side, the JS apparently does diff to minimize the payload and it triggers on the server the same error as if I try to only update some property inside the wireable (as seen in the unit test)
- My real world application then is the second browser test: the possibility to partially update a wireable via inputs.

I didn't know if I should've created a new Browser tests file inside the `SupportsWireables` directory, I hope it's fine that I appended it to the `HandleComponents` browser tests.

PS: In my actual real world scenario, the wireable has a corresponding alpine component with multiple inputs, using `x-modelable` & `@entangle`, but I think there's no need in making the test case that complex, this here seems to be enough to reproduce the bug I'm seeing.


levu42:
Hi! @calebporzio Is there any chance you could check this out and figure out, where it's broken? I can help fixing it, but I don't know how it's supposed to work now with Wireables in v3, so I can't really figure out, what exactly is broken.

joshhanley:
@levu42 thanks for the PR! Yeah, please move the browser test to a new browser tests file in `SupportWireables` directory 🙂

levu42:
@joshhanley moved the test into `SupportsWireables` and rebased onto main

calebporzio:
Thank you for reporting this. Unfortunately Livewire 3 doesn't intend to support updating wireables because it could be a security hole where people don't expect that their wireable object properties are mutable.

The answer is to create a synthesizer for the object you want to support: https://livewire.laravel.com/docs/synthesizers

calebporzio:
Re-opening this because of insight from @joshhanley

joshhanley:
I've fixed this issue and it's ready to merge 🙂

levu42:
@calebporzio Thanks for reconsidering this! And @joshhanley thank you for fixing it - amazing to see how small a fix this was in the end :)

PR #6562, opened 2023-08-28
[v3] add failing test for morph markers added within an html tag if a condition contains a "<"

gehrisandro:
Starting from v3.0.0 Livewireinjects morph markers within HTML tags where it should not.

This is because of a change on the regex between v3.0.0-beta.11 and v3.0.0: https://github.com/livewire/livewire/commit/17898dced0e11bbbea775d9a2634731d2c839686

Failing example:
```blade
<div @if (0 < 1) bar="bob" @endif></div>
```
=> Livewire injects an unwanted morph marker because of the `<` in the `if` condition:
```blade
<div <!-- __BLOCK__ --><?php if(0 < 1): ?> bar="bob" <?php endif; ?>></div>
```

But it works fine if only a `>` is used or a `>` is used before the `<`:
```blade
<div @if (1 > 0) bar="bob" @endif></div>
<div @if (1 > 0 && 0 < 1) bar="bob" @endif></div>
```

=> Everything is fine:
```blade
<div <?php if(1 > 0): ?> bar="bob" <?php endif; ?>></div>
<div <?php if(1 > 0 && 0 < 1): ?> bar="bob" <?php endif; ?>></div>
```


levu42:
@gehrisandro since you have this test case `@if (1 > 0 && 0 < 1)`, just to make it complete, do you also want to add the other way round? `@if (0 < 1 && 1 > 0)`. Not sure if that's something that fails currently or not, but it might be good to have a test case coverage for it, since the order of the `<`s and `>`s seems to matter here.

calebporzio:
Thank you for this report, I fixed this already on the main branch so I'm closing this. Thank you for the test cases!

whousehomestay:
This syntax is still showing this bug:
<button class="bs-btn bs-btn-warning" onClick="showConfirm(null)" @if (count($cart) <= 0) disabled @endif >Hapus Semua Item</button>

gehrisandro:
Hi @whousehomestay 

As mentioned [here](https://github.com/livewire/livewire/pull/6616#issuecomment-1710021333) the problem still persists as the fix got reverted before being tagged.

PR #6564, opened 2023-08-28
Failing test for `$dispatch` being called twice instead of once

nicolagianelli:
This is a failing test that covers the $dispatch from JS being called twice instead of once


calebporzio:
Great failing test PR- all fixed, thanks!

js/features/supportEvents.js
```diff
@@ -12,6 +12,7 @@ function registerListeners(component, listeners) {
     listeners.forEach(name => {
         // Register a global listener...
         let handler = (e) => {
+           console.log('this one')
```
levu42, line 15: @calebporzio I think you merged a console.log here :)

PhiloNL: Good catch! I've removed it https://github.com/livewire/livewire/pull/6811

PR #6565, opened 2023-08-28
Failing test for Form Object Validation uninitialized properties

nuernbergerA:
for my buddy @austenc 

given the following code
```php
class ClientForm extends Form
{
    public ?Client $client = null; // this is set in the component mount

    #[Rule('required')]
    public $name;

    #[Rule(['required', 'email:rfc,filter'])]
    public $email;

    public function rules()
    {
        dd($this->client); // will be null

        return [
            'email' => collect([])->when($this->client,
                fn ($rules) => $rules->push(ValidationRule::unique('clients', 'email')->ignore($this->client->id)),
                fn ($rules) => $rules->push(ValidationRule::unique('clients', 'email'))
            )->toArray(),
        ];
    }
}
```

### DIFF
--- a/src/Features/SupportFormObjects/UnitTest.php
+++ b/src/Features/SupportFormObjects/UnitTest.php
@@ -6,6 +6,7 @@ use Livewire\Attributes\Rule;
 use Livewire\Component;
 use Livewire\Form;
 use Livewire\Livewire;
+use PHPUnit\Framework\Assert;
 
 class UnitTest extends \Tests\TestCase
 {
@@ -205,6 +206,32 @@ class UnitTest extends \Tests\TestCase
     }
 
     /** @test */
+    function all_properties_are_available_in_rules_method()
+    {
+        Livewire::test(new class extends Component {
+            public PostFormWithRulesStub $form;
+
+            public function mount()
+            {
+                $this->form->setPost(42);
+            }
+
+            function save() {
+                $this->form->validate();
+            }
+
+            function render() {
+                return '<div></div>';
+            }
+        })
+        ->assertSet('form.post', 42)
+        ->call('save')
+        ->assertSet('form.post', 42)
+        ->assertHasErrors()
+        ;
+    }
+
+    /** @test */
     function can_get_properties_except()
     {
         $component = new class extends Component {};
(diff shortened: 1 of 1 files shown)


austenc:
Thanks, @nuernbergerA!

Encountered this when trying to use a validation rule on an edit form like this:
```php
Rule::unique('clients', 'email')->ignore($this->client->id)
```

The `ignore` portion requires a primary key to be passed in, but it seems the form's `rules()` method is called before it is fully hydrated. 

The component that uses the form looks something like this:
```php
class ClientEdit extends Component
{
    public ClientForm $form;

    public function mount(Client $client)
    {
        $this->form->setClient($client);
    }

    public function update()
    {
        $this->form->validate();
        $this->form->update();
    }
}
```

calebporzio:
Thanks for the failing test Adrian!

Here's the problem:
* Form object validation rules are gathered when the object is instantiated
* THEN the component's mount method is called

We need to just make sure that validation rules aren't being gathered before the `mount()` method.

There are two solution paths here:
Solution path A) (Hooks): Find or create a hook in the codebase to gather form object validation rules
Solution path B) (JIT): Only provide validation rules when they are needed by a user via `$this->validate()`;

Hooks pros:
* Most of the existing code stays the same as it is right now, we just find a new hook
Hooks cons:
* We risk a user calling `$this->validate()` BEFORE all the rules are gathered.

I think the clear answer here is the JIT approach, we just need to implement it in a way that doesn't feel horrible.

fitimvata:
@calebporzio I have done the implementation after your stream.

sohailahmad07:
Hi I have the kind same issue but it is a bit different i am using a custom rule and when I pass the data to that rule it gets an empty string which is the initial value my code

FormObject
```
public UploadedFile|string|null $body = null;
```


```
class ValidMediaFile implements ValidationRule
{
    
    public function __construct(protected $type)
    {
    }

    public function validate(string $attribute, mixed $value, Closure $fail): void
    {
       dd($this->type) //  output is ""
     // check if the upload media is the same as the select type
    }
  }
```

calebporzio:
This has been fixed in #6731

PR #6566, opened 2023-08-28
[V3] Docs - Fixes forms code example

Maarten-Dekker:
This pr fixes the code sample and aligns it with the expectations.

The docs describe that a user who tries to submit the form without filling in any of the fields will see the validation messages. This is not true, as these messages will only appear if an user adjusts a field. A `$this->validate();` line will fix this. Also adds the missing brackets in `$this->only();`

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -82,8 +82,10 @@ class CreatePost extends Component
 
     public function save()
     {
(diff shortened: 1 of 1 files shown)


calebporzio:
Thank you!

PR #6577, opened 2023-08-29
[v3] add failing test for pagination on nested alpine tabs

jack-interconnectit:
Since v3, if a page loads a set of alpine navigation tabs and within those tabs there is sub navigation, pagination breaks with the following console error. 
```
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'alpine')
    at push (livewire.js?id=af0b760a:6048:53).....
```
If the page is refreshed while on the broken sub tab, then everything loads correctly and the pagination works. 

This did work on v2.

Original discussion is here https://github.com/livewire/livewire/discussions/6520


calebporzio:
Good catch - fixed!

jack-interconnectit:
Excellent thanks!

PR #6579, opened 2023-08-29
[v3] fix wrong string concatenation in exception message

gehrisandro:
This PR fixes an error in the exception message creation if a component can not be found.

(The reason why I do run into this error is not clear for me yet, but trying to reproduce it in a clean way and submit a different PR)


calebporzio:
Ooof thanks

PR #6581, opened 2023-08-29
[V3] add failing test for redirect with navigate via $parent

gehrisandro:
This PR adds a failing test for redirects with `navigate = true` triggered from a child component via `$parent`.

How to reproduce:
1. You need two full page components each one having a method to redirect to the other page: 
```php
public function visit(): void
{
    $this->redirect('/second', true); // and '/first'
}
```
2. Create another component used as a child on the first page, with a button calling the visit method of the parent:
```blade
<button wire:click="$parent.visit">Go to page 2</button>
```
3. After redirecting from the first to the second page and back again, clicking on the button once more will result in an error:

Screenshot 2023-08-29 at 16 44 16

There is another bug preventing us from having a proper error message here. Fixing this bug will lead to a better error message. See my [PR here](https://github.com/livewire/livewire/pull/6579).

Screenshot 2023-08-29 at 16 45 06

I was not able to fix the underlying bug but it looks like the snapshot of the parent component still contains the IDs of the child components from before the redirect. Which are used here: https://github.com/livewire/livewire/blob/9c3d71d0d3ef9068feebb03c832028d1e93e439c/js/component.js#L97C20-L97C36

When executing the redirect directly from the parent component with `wire:click="visit"` everything works fine.

Hope my explanations and the failing test helps to find the bug 😅


gehrisandro:
Hey @calebporzio 

Thank you for fixing quickly! 🚀 

And sorry for the next [failed test](https://github.com/livewire/livewire/pull/6616) 😅

jmas:
@gehrisandro  @calebporzio Thank you, guys! It took me forever to find the reason of this bug! And I can throw out my code that "fix" it:

```js
let lastPopStateTime = 0;

addEventListener("unhandledrejection", () => {
  if (isJustBackNavigated()) {
    location.reload();
  }
});
```

PR #6592, opened 2023-08-30
[V3] Add failing test for morphing from placeholder to component of different element type

PhiloNL:
If your placeholder is a different element type compared to what the component returns it looks like everything is working but it will break the front end, any javascript will no longer work and the console will throw an error about missing snapshot data which seems to be removed while morphing from a `<div>` to a `<form>` element, for example.

### DIFF
--- a/src/Features/SupportLazyLoading/BrowserTest.php
+++ b/src/Features/SupportLazyLoading/BrowserTest.php
@@ -183,6 +183,37 @@ class BrowserTest extends BrowserTestCase
     }
 
     /** @test */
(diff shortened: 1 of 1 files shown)


calebporzio:
Hmmm tricky, I'm tempted to forbid this and instead put a warning in the docs.... I kind of violates how morph works and would probably need a special case added just for this. what do you think?

PhiloNL:
> Hmmm tricky, I'm tempted to forbid this and instead put a warning in the docs.... I kind of violates how morph works and would probably need a special case added just for this. what do you think?

That would be fine with me :) Added a PR to the docs https://github.com/livewire/livewire/pull/6626

PR #6593, opened 2023-08-30
[V3] Failing test for Events (x-on)

nicolagianelli:
These failing tests are covering events handled by x-on from generic/self and child event that don't seem to work

### DIFF
--- a/src/Features/SupportEvents/BrowserTest.php
+++ b/src/Features/SupportEvents/BrowserTest.php
@@ -9,20 +9,244 @@ use Livewire\Livewire;
 class BrowserTest extends BrowserTestCase
 {
     /** @test */
-    public function can_listen_for_component_event_with_this_on_in_javascript()
+    public function can_listen_for_component_event_with_x_on()
     {
         Livewire::visit(new class extends Component {
-            function foo() {
-                $this->dispatch('foo');
+            function dispatchTestEvent() {
+                $this->dispatch('test-event');
             }
 
             function render()
             {
                 return <<<'HTML'
                 <div>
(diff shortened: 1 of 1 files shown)


nuernbergerA:
i think this is more a alpine "problem".

i assume events are fired at root level and then due to the nature of events they will bubble only up and not down.

lets take your test code
```html
<div>
    <button wire:click="dispatchTestEvent" dusk="button">Dispatch "foo"</button>
    <span x-on:test-event="$el.textContent = 'bar'" dusk="target" wire:ignore></span>
</div>
```
this is what livewire will do
```html
<div> <!--event will bubble up and could be captured here -->
    <button x-on:click="$dispatch('test-event')">Dispatch "foo"</button>
    <span x-on:test-event="$el.textContent = 'bar'" wire:ignore></span> <!--this is on the same level and will not be informed about the event -->
</div>
```
how you can handle it
```html
<div> <!--event will bubble up till the root in dom = window -->
    <button x-on:click="$dispatch('test-event')">Dispatch "foo"</button>
    <span x-on:test-event.window="$el.textContent = 'bar'" wire:ignore></span> <!--event will be captured -->
</div>
```

calebporzio:
This should probably be better documented, but you can use `$wire.on()` to listen for an event within a component (not on the root) as well.

I agree that this isn't totally ideal, but I don't think normal Alpine listeners will ever be able to listen for Livewire events bubbling "down" if you know what I mean. In other words, there might be a better solution, but I don't think it's bubbling "down" (one reason is there's no real way to do that in the browser without hacking a lot)

So, for now I'll say use `$wire.on` or `@this.on` if you want to listen for a Livewire event within itself. 

Thanks!

PR #6595, opened 2023-08-30
[V3] Fix example in docs 📚

michael-rubel:
Fixes the example in `events.md` docs.

Newcomers will avoid this error trying to reproduce the example:

```
Call to undefined method App\Events\OrderShipped::dispatch()
```


calebporzio:
Thank you!

PR #6598, opened 2023-08-30
Add `#[Encrypted]` properties attribute

calebporzio:


francoism90:
@calebporzio Let me know if you want me to test. :)

calebporzio:
Will do @francoism90 ! Thanks.

I think I'm going to put together an implementation but not sure I'm going to merge it right away. Want to make sure there aren't cases I'm missing.

francoism90:
@calebporzio I have taken a look at the implementation, looks good so far. :)

Just a question, is it possible to add multiple conditions, like this:
```php
#[Encrypted]
#[Locked]
public Post $post;
```
Or does encrypted imply locked as well?

Thanks.

calebporzio:
Good question, encrypted doesn't necessarily imply "locked", you would need both to lock it (and it should work with locked)

docs/encrypted.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+When this component is _dehydrated_ (serialized) and sent to the browser, the `$post` property will be sent as the following array of data:
+
+```json
+{
+    "type": "model",
+    "class": "App\Models\Post",
+    "key": 1,
+    "relationships": [].
```
dxnter, line 22:
```suggestion
    "relationships": [],
````

calebporzio:
I'm going to close this until theres more demand

patriktoth67:
May I ask what the situation is with this feature?

francoism90:
I tried to implement myself, but it seems to need overrule Livewire logic:
https://github.com/livewire/livewire/blob/495aac722c08b620d7e4a48b3c4e76ecad31a4b2/src/Mechanisms/HandleComponents/HandleComponents.php

@calebporzio Would it be possible to implement this, without modifying the `HandleComponents.php` or any Livewire logic directly? :)

PR #6602, opened 2023-08-31
[V3] Remove confusion from session flash docs

skeemer:
I think the `$this->` was a typo as the `$this->redirect*()` methods are void and there isn't currently any way to support chaining on them. There's also a `return` that doesn't make sense for a void function. It could be built, but this seems to be creating quite a bit of confusion in the meantime.

Also, is there any reason the examples have a `return` before any of the `$this->redirect*()` functions? Should I remove those? I think that's part of what's causing confusion as well.

### DIFF
--- a/docs/redirecting.md
+++ b/docs/redirecting.md
@@ -81,7 +81,9 @@ class UpdatePost extends Component
     {
         // ...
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Yeah, you're right. I've tweaked the docs a bit to address the things you mentioned. Thank you!

PR #6610, opened 2023-08-31
Allow only/except to be called without an array

dwightwatson:
I did not - but I felt it was worthy for consideration so that `only` and `except` work in a similar manner to those in Laravel.

Sure did.

No - all related.

I updated the relevant (and added new tests) for the form objects.

When using `only`/`except` on the HTTP request in standalone Laravel you have the option to pass a single key, an array, or multiple keys as arguments and it works regardless.

```php
Request::only('foo');
Request::only('foo', 'bar');
Request::only(['foo', 'bar']);
```

However, the `only` method in Livewire currently only supports an array as an argument, and `except` supports a single argument or an array. [You can view Laravel's `InteractsWithInput` to see the inspiration for this change](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Http/Concerns/InteractsWithInput.php#L424-L464).

With this change, the following should all work identically.

```php
$this->only('foo');
$this->only('foo', 'bar');
$this->only(['foo', 'bar']);

$this->except('foo');
$this->except('foo', 'bar');
$this->except(['foo', 'bar']);
```

### DIFF
--- a/src/Concerns/InteractsWithProperties.php
+++ b/src/Concerns/InteractsWithProperties.php
@@ -81,7 +81,7 @@ trait InteractsWithProperties
     {
         $results = [];
 
-        foreach ($properties as $property) {
+        foreach (is_array($properties) ? $properties : func_get_args() as $property) {
             $results[$property] = $this->hasProperty($property) ? $this->getPropertyValue($property) : null;
         }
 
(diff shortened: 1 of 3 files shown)


calebporzio:
I dig it, thank you!

PR #6612, opened 2023-08-31
[V3] Highlight full methods in synthesizers docs

chillbram:
Just a small visual fix, as the first method signature and last closing brace are not highlighted


calebporzio:
Thanks!!

PR #6613, opened 2023-08-31
[v3] $wire.toggle: allow deferring 👌

levu42:
Currently, `$wire.set` has a 3rd argument `live = true` that allows deferring the update of a property until the next network request. The `$wire.toggle` shortcut doesn't allow deferring so far. I fixed that with this PR.

Given that Caleb highlighted the reduction of unnecessary network requests as one of the big benefits of v3 vs v2, I assume that this is a wanted feature :)

Before:
```html
<!-- sends network request - ugh! -->
<button type="button" @click="$wire.toggle('foo')">...</button>

<!-- reinventing the wheel - ugh! -->
<button type="button" @click="$wire.set('foo', ! $wire.get('foo'), false)">...</button> 
```

After:
```html
<!-- No network request 😍 -->
<button type="button" @click="$wire.toggle('foo', false)">...</button>
```

### DIFF
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -195,10 +195,10 @@ let $wire = {
 
     // Set a property on the component by name...
     // Usage: $wire.$set('count', 5)
-    $set(name, value) { ... },
+    $set(name, value, live = true) { ... },
 
     // Toggle the value of a boolean property...
-    $toggle(name) { ... },
+    $toggle(name, live = true) { ... },
 
     // Call the method
     // Usage: $wire.$call('increment')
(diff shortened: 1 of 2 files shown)


nuernbergerA:
I like that.

Can you add a test?

levu42:
@nuernbergerA I was torn between "It doesn't need a test, because toggle is just a shortcut of set('foo', ! get('foo')) and it's a one-liner" and "I can't seem to find a way to assert in dusk that no network request has been made". Do you have any idea how I can test that?

calebporzio:
Sure, I'm in. Thanks!

nuernbergerA:
The master has spoken

PR #6614, opened 2023-08-31
Shorten form object validation rules

nicolagianelli:
#6280 
Yes
No
Yes

When validating a Form Object you have to use the full key for the rule like "form.other_field" inside a required_if or other rules with other fields dependency and if you use the full key it works only when the name of variable used for the form matches the one used in the rule and this isn't a good approach.
With these changes you can continue to use scoped keys inside the Form Object like "other_field" for required_if and other rules that have field dependency.


calebporzio:
Ah I see. I validate that this is a necessary bug fix, however, I'm unhappy with the implementation in the amount of complexity it adds.

Are there any other ways to accomplish this?

nicolagianelli:
I think this is the most direct approach to the problem..
An other approach could be to write field dependent rules like this: required_with:{field},{field2} or required_if:{field},value so with a Regex I can replace the fields by replacing and appending the correct form/variable name to the fields.
I don't think there are other approaches to this problem..

joshhanley:
@nicolagianelli thanks for the PR! We've had a closer look at this and agree this is probably the best way to approach it for now (without completely changing how FormObject validation works).

I've added some extras to the tests to ensure we have covered all our bases (which are now failing). So there is a section that needs fixing regarding the shortening of the field names in the validation attributes, so when the message is output, it looks correct. I re-added a bit of code that you removed and updated it to use the new `$propertyName` variable. But it then caused other tests to fail, so comment it out for now and left a note in the comment.

I've also added comments for context to make it clearer what is going on in each section and did a re-arrange of the logic in the Rule Attribute.

There is also a conflict with the main branch, which we will need to address.

Thanks again for your effort on this issue! 🙂

nicolagianelli:
@joshhanley
- I merged the branch with the main.
- Fixed the tests because camel case variables should be showed as "camel case" like laravel default validation does.
- Now rules in forms have dependant fields also shortned in validation messages.
- Fixed a bug where the BaseRule attribute was using $this->getName() in update triggering all the form validation instead of only the changed field.
- I had to change the visiblity of getValidationAttributes() in HandlesValidation from protected to public because I need to check, in BaseRule, if there is already an attribute to not override it with the default one.

joshhanley:
@nicolagianelli thanks for the PR! Going to close this as Caleb restructured how validation runs for FormObjects so that each FormObject runs it's own validation. See PR #7266 for more details.

PR #6616, opened 2023-08-31
Fix morph markers on long content without html tags

gehrisandro:
This only affects the current "main" branch, because issue has been introduced [on a change not released yet](https://github.com/livewire/livewire/commit/6c865772d85990f3debaa8cd86a320551b14e229)!

I do not fully understand the reason behind this error but it looks for me like a weird combination between the updated regex with the newly introduced `negative lookahead` and a limitation on PHP preg functions.

Given a matching directive like `@if` followed by long content (~> 6150 characters) not containing a `<` will lead to an unexpected behaviour in the `preg_replace_callback` statement: It returns an empty string.
Which ultimately results in this exception:
```
Livewire\Exceptions\RootTagMissingFromViewException: Livewire encountered a missing root tag when trying to render a component. 
 When rendering a Blade view, make sure it contains a root HTML tag.
 ```

The tests added may look a bit silly, but here is what happens:
```php
'@if(true) ' . str_repeat('a', 6126) . ' @endif', // Here the preg replace works as expected

'@if(true) ' . str_repeat('a', 6127) . ' @endif', // Here it returns am empty string

'@if(true) ' . str_repeat('a', 6126) . ' @endif <div></div>', // same here, because the < from the div is too far away from the @if

'@if(true) <div></div> ' . str_repeat('a', 6127) . ' @endif', // But this one does work again because of the < on the div
```

And here is a little snippet with a simplified regex (only `@if`):

```php
$pattern = '/@if(?![a-zA-Z])(?!([^<]|(?!<[a-zA-Z\/])<)*(?<![?=-])>)/';

preg_replace($pattern, 'REPLACE', '@if(' . str_repeat('a', 6141)); // returns null !!!

preg_replace($pattern, 'REPLACE', '@if(' . str_repeat('a', 6140)); // returns the expected string
```

Hope this helps to find the root cause of the issue 😅

### DIFF
--- a/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php
+++ b/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php
@@ -54,7 +54,7 @@ class SupportMorphAwareIfStatement extends ComponentHook
                         [^<]                      # All non "<" characters
                         |                         # OR
                         (?!<[a-zA-Z\/])<          # A "<" character without a forward slash or letter after it
-                    )*                            # As many characters as it can until ">" is reached
+                    ){0,1000}                     # As many characters as it can until ">" is reached (limited to 1000 characters for performance reasons)
                     (?<![?=-])                    # Ignore "?>", "->", and "=>"
                     >                             # A ">" character
                 )
--- a/src/Features/SupportMorphAwareIfStatement/UnitTest.php
+++ b/src/Features/SupportMorphAwareIfStatement/UnitTest.php
@@ -347,6 +347,22 @@ class UnitTest extends \Tests\TestCase
                 </div>
                 HTML
             ],
+            25 => [
+                1,
+                '@if(true) ' . str_repeat('a', 6126) . ' @endif',
+            ],
+            26 => [
+                1,
+                '@if(true) ' . str_repeat('a', 6127) . ' @endif',
+            ],
+            27 => [
+                1,
+                '@if(true) ' . str_repeat('a', 6126) . ' @endif <div></div>',
+            ],
+            28 => [
+                1,
+                '@if(true) <div></div> ' . str_repeat('a', 6127) . ' @endif',
+            ],
         ];
     }
 


gehrisandro:
Just realised there is a `preg_last_error_msg()` method.

It returns `"JIT stack limit exhausted"`. Looks like the regex is too complex to resolve.

Given this information I was able to find a possible fix or at least a workaround by limiting the look ahead to 1000 characters.

joshhanley:
@gehrisandro just to confirm, this is no longer just a failing test and actually has a fix, correct? If so, I have updated the title 🙂

gehrisandro:
@joshhanley Yes that's true.

But the solution is more a workaround than a clean fix because the 1000 characters limit is a "random" value working for my use case.

joshhanley:
@gehrisandro ah ok, thanks. Let's wait to see what Caleb thinks.

gehrisandro:
@calebporzio Before the latest release (v3.0.2) you have [reverted](https://github.com/livewire/livewire/commit/0130160033048f62bc7bbcaca271861e4d96592e) the changes relevant for this PR.

But this reintroduces the original problem using [`<` within an HTML tag](https://github.com/livewire/livewire/pull/6562).

How do you want to proceed here? Should I open another PR to add the failing tests again?

gehrisandro:
@calebporzio I tried to figure out a solution for this. But I have to give up, because I can't imagine a 100% reliable way to detect, if a Blade directive is used inside or outside a tag.

But maybe you can come up with a solution for this?

If not, I think it would be better to remove the feature entirely and encourage the users to wrap conditions within a html element (like in v2). If they do rely on the automatically injected markers and further down the road they hit a limitation of the regex, they will be forced to set `inject_morph_markers` to `false` and by doing this they will possibly run into problems with components built before.

Sorry for being pessimistic here, but I would be really impressed and surprised if someone comes up with a proper solution 😄

calebporzio:
Hey @gehrisandro, yeah you're 100% right that it's nearly impossible.

I've sort of "hung my hat" on finding a way to make this feature work.

I'm still exploring solutions and happy to hear others, but I'm irrationally opposed to going back to the V2 ways because it hurts newcomers more than this solution does (I think).

Whatever happens, I know there are better solutions than the one we currently have. Thank you for attempting them.

(I reverted the ">" solution because the regex wasn't performant enough and was pretty memory hungry and would therefore timeout easily)

I think the two paths forward are:
A) Hooking into Laravel's Blade tokenizer somehow and detecting/injecting at that point
B) parsing and regexing in "stages" to make it more memory effecient

I'm going to close this for now as I reverted the commit it depends on, but I'm very interested in further contributions in this arena. Thanks!

gehrisandro:
Hi @calebporzio 

I've already tried to use / modify the Blade tokenizer to solve the problem but I wasn't successful.

Another idea I had was to use `strip_tags` to determine which markers are inside or outside a tag. But this does not work within a compiled blade file because of all the PHP stuff in there.

But what would be possible is to use `strip_tags` after the component has fully rendered. So I've created a little POC to test this out: https://github.com/livewire/livewire/commit/98fdda22beb157076c5003139dfc739149b5e582

What it actually does:
1. Insert morph markers with a count to make them unique
2. Let the component render
3. Use `strip_tags` to determine which morph markers are used in a proper place
4. Strip all morph markers NOT used in the proper place.
5. Rollback the remaining morph markers to their correct value.

Here my thoughts on this:
- Determining if a morph marker is needed or not does not seem to be possible for me before rendering the full html.
- The new regexes are run after each render and not in the blade compile stage which is maybe a performance issue. I didn't do any performance checks yet.

What are your thoughts on this idea?

PR #6625, opened 2023-09-01
Add `nonce` to `nprogress` stylesheet

joostdebruijn:
As described in discussion #6588: when using Livewire v3 in an application with a nonce-based CSP policy, a CSP-violation is triggered. The Bar-plugin injects the nprogress css stylesheet without a nonce.

If there is a scriptConfig, it gets the nonce from the script config. Otherwise, it gets the nonce from the already injected Livewire style (if it has a nonce).


enkhjargal (approved):

joostdebruijn:
@joshhanley Is there anything I can do in order to get this one merged?

joshhanley:
@joostdebruijn Caleb and I had a quick look at this but I can't remember why we held off on this one. We will have to have another look when we get the chance. Conceptually it seems fine.

joshhanley:
@joostdebruijn thanks for the PR!

PR #6626, opened 2023-09-01
Add lazy load warning to use same root element type

PhiloNL:
Follow up on https://github.com/livewire/livewire/pull/6592


joshhanley:
@PhiloNL I think there might be a bug, as I've seen some reports that root elements currently can't be anything other than a div. If that's still true and related then I'd imagine fixing it would allow the placeholder element to be different that the root element maybe?

calebporzio:
Perfect, thanks @PhiloNL

PR #6629, opened 2023-09-01
Only allow response-content-disposition to include utf-8 characters

jimmitjoo:
This PR is assuming we use the S3 as filesystem.

Since this is not throwing an error from S3 when trying to create the signed url, I am not sure how to provide a test for it? Please guide me.

If we upload a file with a name that contains for example åäö or other non utf-8 characters, we will face an error when trying to fetch the temporary url. The error S3 gives us looks like this:

```
<Error>
  <Code>InvalidArgument</Code>
  <Message>Header value cannot be represented using ISO-8859-1.</Message>
  <ArgumentName>response-content-disposition</ArgumentName>
  <ArgumentValue>filename="Skärmavbild 2023-08-17 kl. 15.10.49.png"</ArgumentValue>
  <RequestId>HEHY0ZACMTRHP2KW</RequestId>
  <HostId>d5yi+s8NycsiInzS4BH5FWdYfGsaxElTflenY/lAigtqVL/0hrAojVQWGZX1Qv0fPhuLNaq6cQc=</HostId>
</Error>
```

I have only changed code in one file. Instead of providing the original filename as "ResponseContentDisposition filename", I am making sure the filename does not contain anything except utf-8 characters as that is what S3 will support.

`$filename = preg_replace('/[^(\x20-\x7F)]*/', '', $this->getClientOriginalName());`

With this fix the temporary file works fine.

I experienced this issue when taking a screenshot on a Swedish MacBook, where the screenshot files are named "Skärmavbild [datetime].png". When removing the character ä from the string it worked fine.


joshhanley:
@jimmitjoo thanks for the PR! Just to confirm, this will then store the filename with the characters missing, correct?

jimmitjoo:
@joshhanley that is correct. 

That would be a repo with just a basic Livewire upload form. Because everything will work excellent as long as your local file does not contain any special characters like åäö.

The more I think about it, I think my preg_replace should be replaced by a utf8_encode of the original filename.

joshhanley:
@jimmitjoo ok thanks. So if you upload a file now (after your changes), say called `Skärmavbild.png`, will it be stored as the UTF-8 encoded file name? If so, is that still problematic as it's not the same name the user uploaded it as?

Was doing a quick bit of reading and found this article, which may allow us to support maintaining the original filename (if it is currently being changed as per my question above).
https://engineering.resolvergroup.com/2022/02/aws-s3-utf-8-content-disposition/

joshhanley:
@jimmitjoo any thoughts on my questions above?

calebporzio:
Closing because of lack of response

jimmitjoo:
I can now confirm that this is a S3 bug. When trying to upload files to Laravel Vapor including "åäö" characters, it fails in the same way.

To respond to you @joshhanley, it will have a different name. However it should not be any problems opening the file and such. The link you provided seems to do the trick to keep the original filename. I may have time to fix it during christmas.

PR #6641, opened 2023-09-03
Fix `livewire:navigated` not fired on first page load

iksaku:

No

Yes

No

Yes

It is stated in the docs that relying on the `DOMContentLoaded` for your scripts is incompatible when using `wire:navigate` feature, instead, it is recommended to rely on `livewire:navigated` (alias for `alpine:navigated`) for scripts that require interacting with the page once the page is fully loaded.

This behavior was broken on commit 310c92c9e67134a06a86fc086838423ee290de47.

Thanks for contributing! 🙌


levu42:
@iksaku are you potentially confusing `livewire:navigated` and `alpine:navigated` in the title of the PR? In #6490 I wrote that `alpine:navigated` is fired on first load, but `livewire:navigated` is not. Your tests are also testing `livewire:navigated`.

iksaku:
@levu42 You're right, I confused the event names, let me fix that. Thanks!

joshhanley:
@iksaku thanks for the PR! It was removed as it was causing navigate to be used for all page changes regardless of if they were using `wire:navigate` or not. Not sure if you have managed to fix that issue, but there is a related failing test, if you could have a look at that? Thanks

image

levu42:
@joshhanley while `livewire:navigated` isn't fired on the first page load, the documentation should at least be updated, it currently says:

>When using wire:navigate, DOMContentLoaded is only fired on the first page visit, not subsequent visits.
> 
> To run code on every page visit, swap every instance of DOMContentLoaded with livewire:navigated:
> 
> [...]
> 
> Now, any code placed inside this listener will be run on the initial page visit, and also after Livewire has finished navigating to subsequent pages.

joshhanley:
@levu42 actually, the docs are correct. As mentioned above, there was a big bug caused by the way it was implemented. It was temporarily removed to stop navigate running for every link in every app. But it would be better if a solution can be found to get it all to run happily as documented 🙂

iksaku:
Closing in favor of #6699

PR #6647, opened 2023-09-03
Fix example for assertHasErrors in docs

koenwoortman:
From testing this out, it looks like the example should be like this


calebporzio:
Thank you!

PR #6654, opened 2023-09-04
[V3] Multiple messages/As translation fix on rule attribute

nicolagianelli:
Hello,
if you pass an array of messages or As to a Rule attribute with translate enabled it won't work, this PR fixes the problem.


joshhanley:
@nicolagianelli Thanks for the PR! Can you please add some tests? Thanks!

calebporzio:
Thanks!

PR #5239, opened 2022-09-08
Fix for unstable temporaryUrl() URLs

levu42:
Review the contribution guide first at: https://laravel-livewire.com/docs/2.x/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

Yes, #5036 

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)

Yes

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No

4️⃣ Does it include tests? (Required)

Yes

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

See #5036 :)

Thanks for contributing! 🙌

PR #6669, opened 2023-09-05
[V3] Make temporary URLs stable over some time to prevent browser cache problems

levu42:
This is a reintroduction of my #5239 PR that was rejected for v2 because v3 was in development. The original discussion is in #5036 with a bigger real-world example.

The problem is that every time component state changes and the server re-renders the component, temporary URLs are currently changing and being re-generated, since they depend on the exact timestamp. This means, that if I have e.g. an input with `wire:model.live`, and a temporary preview URL for an uploaded photo, the photo gets re-downloaded every time I'm typing in the input field. This is obviously not a good thing in any way, so this PR makes the temporary URLs stable for one hour at a time.

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -86,7 +86,7 @@ class TemporaryUploadedFile extends UploadedFile
         if ((FileUploadConfiguration::isUsingS3() or FileUploadConfiguration::isUsingGCS()) && ! app()->runningUnitTests()) {
             return $this->storage->temporaryUrl(
                 $this->path,
-                now()->addDay(),
+                now()->addDay()->endOfHour(),
                 ['ResponseContentDisposition' => 'filename="' . $this->getClientOriginalName() . '"']
             );
         }
@@ -97,7 +97,7 @@ class TemporaryUploadedFile extends UploadedFile
         }
 
         return URL::temporarySignedRoute(
-            'livewire.preview-file', now()->addMinutes(30), ['filename' => $this->getFilename()]
+            'livewire.preview-file', now()->addMinutes(30)->endOfHour(), ['filename' => $this->getFilename()]
         );
     }
 
--- a/src/Features/SupportFileUploads/UnitTest.php
+++ b/src/Features/SupportFileUploads/UnitTest.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features\SupportFileUploads;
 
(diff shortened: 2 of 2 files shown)


src/Features/SupportFileUploads/TemporaryUploadedFile.php
```diff
@@ -86,7 +86,7 @@ public function temporaryUrl()
         if ((FileUploadConfiguration::isUsingS3() or FileUploadConfiguration::isUsingGCS()) && ! app()->runningUnitTests()) {
             return $this->storage->temporaryUrl(
                 $this->path,
-                now()->addDay(),
+                now()->addDay()->endOfHour(),
                 ['ResponseContentDisposition' => 'filename="' . $this->getClientOriginalName() . '"']
```
edalzell, line 90: I don't understand the need for this, it's already adding a day, why do you need a few more minutes?

src/Features/SupportFileUploads/TemporaryUploadedFile.php
```diff
@@ -97,7 +97,7 @@ public function temporaryUrl()
         }
 
         return URL::temporarySignedRoute(
-            'livewire.preview-file', now()->addMinutes(30), ['filename' => $this->getFilename()]
+            'livewire.preview-file', now()->addMinutes(30)->endOfHour(), ['filename' => $this->getFilename()]
         );
```
edalzell, line 101: why not `->addHour()`?

src/Features/SupportFileUploads/TemporaryUploadedFile.php
levu42: @edalzell `now()->addDay()` generates a different timestamp now than it does in a minute. With `->endOfHour()`, the timestamp generated doesn't change between a request now and a subsequent request in a few minutes => the URL stays the same => the browser doesn't re-download the file. That's the whole purpose of this change, to keep the URLs stable with different request times :)

edalzell: sure but what happens if this time period spans a day? like from 11:59pm to 12:01am?

levu42: @edalzell Then it's a new URL, but the point of this PR is to make this happen only once an hour, instead of at every new request. In #5036 I described back in the day the real-world example where this fixed all our problems (multi-file upload, preview of already uploaded files)

calebporzio:
Makes sense. thank you!

PR #6696, opened 2023-09-07
Fix wire transition elements hidden on page load

joshhanley:
Currently, if a list of elements has `wire:transition` on them (so if any new elements are added, they transition in) they're not being displayed on page load.

This is because visibility is set to false by default.

So instead I've added a check to see if the element has been added via morph and if so, then set to false initially, so it can be transitioned in, otherwise just display it.

Hope this helps!


RyleaStark:
I can confirm this PR fixes the behavior I reported in the Discord; Thanks again for the quick fix! 👍🏻

skeemer:
Is there a way to make this optional or is there a better way to do transitions on navigate? I thought it was intended because it makes it easy to do a nice quick fade-in of the new page. It's nice to have especially between similar pages as a visual cue that you did navigate.

joshhanley:
@skeemer this problem and fix is just for normal page loads, not using `wire:navigate` 🙂

skeemer:
Sorry, I guess I crossed wires following the discussion.

joshhanley:
@skeemer no worries! 🙂 hmm not sure. Will test in the morning. That seems like it could work though

PR #6699, opened 2023-09-07
Add support to disable navigate on redirects

gehrisandro:
This PR adds support to explicitly disable `navigate` on redirects.

Currently after a redirect with `navigate` has been performed every subsequent redirect will use `navigate` too and there is no way to turn this off.
I would have assumed that passing `false` as the `$navigate` parameter would prevent Livewire from using `navigate`, but this isn't the case.
```php
$this->redirect('/page-2', navigate: false);
```

In my use case this is a problem because sometimes we have to ensure a redirect is performed without `navigate`. Otherwise we are running into an issue when recreating the session (CSRF token mismatch). But for sure there are more reasons why one would like to turn off this behaviour.

This PR changes the behaviour of the `$navigate` parameter on the `redirect()` method slightly:

- Passing `true` is unchanged
- Not passing the parameter is unchanged
- Passing `false` (which was the default before) does now explicit disable the usage of `navigate`.
- Passing `null` uses `navigate` if it has been used before.

This PR is a breaking change if the following conditions apply:
- The `redirect()` method was explicitly called with `$navigate` set to `false`. In this case before `navigate` was used if it was used before and now it turns `navigate` off for this redirect.

In my opinion we can do this breaking change because most of the users will not have passed `false` explicitly because this is was the default value. And if one has passed it explicitly he probably did it to ensure it is not using `navigate` which was not the case before this PR.

I hope my reasoning is understandable. Otherwise I am happy to answer your questions.


joshhanley:
@gehrisandro thanks for the PR! Wouldn't it be better to reset the state back to what it is on the first page like setting `isNavigating` back to false (or something like that, I haven't dug in so not exactly sure which to change)? That way each page is handled consistently regardless if it has been browsed to directly or been navigated to?

gehrisandro:
Hi @joshhanley 

I am not sure what your intention is behind you suggestion. Sorry 🙃 
But I think it does not matter what we are going to do with `isNavigating` because after a "normal" redirect with `navigate: false` the state will be reseted anyway.

joshhanley:
@gehrisandro so what I was thinking, is a redirect (on the first page) is being handles as a normal redirect, but on the second page (if it has been navigated to) then the next redirect is also triggering the navigate code to run. There must be a flag or check still left over from the navigation to the second page, that is causing navigate to run again when it shouldn't. Does that make sense?

So if my theory is correct, we just need to find out what is causing navigate to keep being active and turn that off once navigation has finished. I hope - I am only just guessing here as I haven't dug thoroughly into the navigate code.

gehrisandro:
@joshhanley I think it was the intention by @calebporzio that navigate is triggered on every subsequent redirect after one was performed with `navigate`. See this test: [can_redirect_without_reloading_from_a_page_that_was_loaded_by_wire_navigate](can_redirect_without_reloading_from_a_page_that_was_loaded_by_wire_navigate)

So my PR is not a `fix` in that sense. Instead it is a way to disable that behaviour if required.

joshhanley:
@gehrisandro hmm ok both links in that test have `wire:navigate` on it though. I'll get him to clarify, as I personally don't see the benefit of navigate still being active. It should only happen for links (or redirects) that explicitly say to use navigate.

Thanks for finding that for me 🙂

gehrisandro:
@joshhanley No, the link from the second page back to the first does not have "wire:navigate":
```blade
<button type="button" wire:click="redirectToPageOne" dusk="redirect.to.first">Redirect to first page</button>
```

It calls `redirectToPageOne` which does not set `navigate` on the redirect:
```php
public function redirectToPageOne()
{
    return redirect('/first');
}
```

joshhanley:
@gehrisandro ah yep, sorry, didn't realise it was the one calling the method. Either way, I think that behaviour is wrong, so I will get Caleb to clarify what the intentions were there, to determine if your fix is best or whether there is an issue with the implementation 🙂

gehrisandro:
@joshhanley Thank you very much 👍

calebporzio:
Ok yeah, the original intention was this (as @gehrisandro guessed):

- When redirecting
- Use navigate if the page visited used navigate
- Or use if "navigate: true"

I see that this PR addresses the problem of passing `navigate: false`. Which I affirm is better than what is currently there.

The question now is (that @joshhanley proposed):

Do we merge this PR and move on? Or change the default behavior to ONLY use navigate when `navigate: true` is passed in?

I can see it both ways. Would love to hear both your thoughts or any other thoughts from the community.

the current behavior does feel a little odd.

Ok, as I'm writing this, I'm making up my mind.

Let's switch it to: Only redirect using navigate if "navigate: true" is passed.

If one of you could update this PR with that behavior and tweak the documentation to reflect that, I'll merge this and we'll be all set.

Thanks!!

gehrisandro:
Hi @calebporzio 

Thank you for your reply.

I totally agree and I can update the PR.

gehrisandro:
PR is updated.

I didn't make any changes on the documentation because they already look fine to me:
1. The behaviour of keep using `navigate` was not documented before
2. In my opinion the new behaviour does not need any documentation as it is what one would expect by default.

P.S: The failing test is not related to my changes.

joshhanley:
Thanks @calebporzio! @gehrisandro thanks for the changes 🙂 I've re-run the failing tests. I think this change will mean we can possibly fix the problem that PR #6641 is attempting to fix.

gehrisandro:
@joshhanley I agree fixing https://github.com/livewire/livewire/pull/6641 is quite easy now.

So I've added a new [commit](https://github.com/livewire/livewire/pull/6699/commits/a04353e9ebfe83947d7248f33ad64e238ecfabd1) here which fixes the issue and adds two new tests to ensure the event is fired regardless if it is the initial page load or the site has been loaded through `wire:navigate`.

If you think it would be better to have a separate PR, just let me know. But in my opinion the two problems are closely related.

calebporzio:
Hmm @gehrisandro - now that you've added that other fix it's slightly harder for me to "instamerge" lol.

Can you explain the problem that the second one is solving? It's hard for me to keep that context in my head

gehrisandro:
@calebporzio This resolves the problem, that the `livewire:navigated` event is not fired on the first page load. Here are the relevant Docs, which tells to use this event instead of `DOMContentLoaded`: https://livewire.laravel.com/docs/navigate#dont-rely-on-domcontentloaded

I have fixed it here because the code which causes this issue has no purpose anymore when we are removing the behaviour of automatically using `wire:navigate` for subsequent requests.

iksaku:
This PR indeed fully fixes both what I was trying to fix with #6641 and also resolves @joshhanley mention of the failing test.

I've tried the lastest changes (which are way more simple than the initial commits) and works perfectly 👍 

Good work @gehrisandro

helturkey:
@calebporzio @joshhanley hope you merge this important PR before the next release, thanks in advance.

robsontenorio:
@gehrisandro Could please update docs explain its usage ?

joshhanley:
@robsontenorio this PR got changed, it's now just a fix for what is already documented.

Wulfheart:
@joshhanley wouldn’t it avoid more papercuts if it would just automatically disable wire:navigate if the redirect is to an external URL?

joshhanley:
@Wulfheart I'm not sure I follow. This PR doesn't address external URLs so I don't understand. Can you explain what you mean?

PR #6707, opened 2023-09-07
[v3] Add fill() method to Form object

jeffchown:
This PR adds the `fill()` method to the new v3 `Form` object. (see https://github.com/livewire/livewire/discussions/6688)

As already available in a `Component`, this makes it easier to pre-populate a `Form` object's public properties with data from a given `Model` or array.  This helps make the initialization of large forms much easier.

Tests are included for both Model and array usage.

As this is my first contribution to any open source repo, I'm open to any suggestions.

Thanks!

PS I had to add a `.vscode/settings.json` file to disable all of my formatting on save settings as they were making many formatting changes in the files I saved (which, ironically, I setup using @calebporzio 's VSCode eBook ;))


calebporzio:
Wow, great contribution for the first time! Thanks a bunch!

I removed that settings file (funny that it came from me lol) to keep changes isolated to this feature and not pollute the repo

I also renamed the "Post" class to something more bespoke so that it doesn't pollute the namespace when running many tests at once. (weird edge case problem I watch out for)

Other than that, great implementation and test. Thanks!

jeffchown:
You're welcome, @calebporzio !  And Thanks for the feedback!

Thanks re: settings file, makes sense (didn't mean to include it in the end) - I did have to laugh re: my vscode making so many formatting changes re: your recommendations ;)

Makes total sense re: `Post` change to avoid any collisions and maintain consistency re: testing resource naming.  I'll make sure to follow that pattern going forward 👍

Thanks for the great work on Livewire - loving v3 🎉

PR #6728, opened 2023-09-08
Added validation call to form object docs

AdrianBav:
Added 1 line to the **Forms** page of the v3 docs.

In the **Extracting a form object** section I believe the call to validation `$this->validate();` is missing from the first code snippet.
Without that line, I experienced a situation where the form objects validation rules were not being run.

This PR adds that 1 line.


calebporzio:
This is great. Thank you!

PR #6729, opened 2023-09-08
Fix `reset()` to ensure a property initially set to `null` can be reset

Simoneu01:
Failing test for #6540

When the `reset()` method is called on a property with default `null` will destroy the property from the component.
`Livewire\Exceptions\PropertyNotFoundException : Property [$nullProp] not found on component:`

With Livewire V2 instead when the `reset()` method was called it would reset the property to null

### DIFF
--- a/src/Concerns/InteractsWithProperties.php
+++ b/src/Concerns/InteractsWithProperties.php
@@ -52,11 +52,27 @@ trait InteractsWithProperties
         $freshInstance = new static;
 
         foreach ($properties as $property) {
-            $isInitialized = (new \ReflectionProperty($freshInstance, $property))->isInitialized($freshInstance);
+            $property = str($property);
+
+            // Check if the property contains a dot which means it is actually on a nested object like a FormObject
+            if (str($property)->contains('.')) {
+                $propertyName = $property->afterLast('.');
+                $objectName = $property->beforeLast('.');
+
+                $object = data_get($freshInstance, $objectName, null);
+
+                if (is_object($object)) {
+                    $isInitialized = (new \ReflectionProperty($object, (string) $propertyName))->isInitialized($object);
+                } else {
+                    $isInitialized = false;
+                }
+            } else {
+                $isInitialized = (new \ReflectionProperty($freshInstance, (string) $property))->isInitialized($freshInstance);
+            }
 
             // Handle resetting properties that are not initialized by default.
             if (! $isInitialized) {
-                data_forget($this, $property);
+                data_forget($this, (string) $property);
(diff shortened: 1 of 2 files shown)


edalzell:
Likely caused by https://github.com/livewire/livewire/pull/6345

joshhanley:
@Simoneu01 thanks for the PR!

I've fixed the issue and made sure that the FormObject's reset test also passes (as per PR #6375 thanks to @markjaquith).

Hope this helps!

src/Concerns/InteractsWithProperties.php
```diff
@@ -52,17 +52,31 @@ public function reset(...$properties)
...
 
         foreach ($properties as $property) {
-            $defaultValue = data_get($freshInstance, $property);
-            $unset = '__unset__';
-            $unsetByDefault = !$defaultValue && $unset === data_get($freshInstance, $property, $unset);
+            $property = str($property);
+
+            // Check if the property contains a dot which means it is actually on a nested object like a FormObject
+            if (str($property)->contains('.')) {
+                $propertyName = $property->afterLast('.');
```
levu42, line 59: @joshhanley shouldn't it be `afterFirst` & `beforeFirst`? When I unset `a.b.c`, then it's `$component->a` that should be unset, not `$component->{'a.b'}`, no?

joshhanley: @levu42 Hmm not sure I exactly follow the abc example, but if you have a form object that has a $foo property that should be unset, then first you need to get $component->form and then unset $foo on the form object.

levu42: @joshhanley When a form has an array, let's say `$component->postForm` has a `user` property as an array with `user.name`

```
class PostForm {
    public array $user = [
        'name' => 'Foo',
        'email' => null,
    ];
}   
```

And then in the component you do `$this->reset('postForm.user.email')`, then the split should be `postForm` & `user.email`. With your code of `afterLast` & `beforeLast`, it would be split into `postForm.user` & `email`, no?

I know it's a super narrow edge case, and not sure, if it's possible at all, but I thought about it reading the code and wanted to just make sure there's not a new bug introduced quietly with this fix that then takes ages to debug in the future :)

joshhanley: Yep that's how I intended it to be split, because I need to see if email is already unset or not on portForm.user. Hence I use after last as the end property I need is email. The data_get handles the postForm.user part. Does that makes sense?

calebporzio:
Lovely, thanks!

PR #6731, opened 2023-09-08
Fix Form Object Validation uninitialised properties.

fitimvata:
Implementation for https://github.com/livewire/livewire/pull/6565


calebporzio:
This is amazing! Thank you both @nuernbergerA and @fitimvata !

PR #6736, opened 2023-09-09
Fix pagination getPage after reload always equals 1 even if /page=x

helturkey:
# fix pagination $this->getPage after reload always equals 1 even if ?page=x 

if you used $this->getPage() before ->paginate() as for example cache identifier Cache::get('posts:' . $this->getPage()) it always gives you 1 after reload even if ?page=x . because of:
``` code
public function getPage($pageName = 'page'){
        return $this->paginators[$pageName] ?? 1;
    }
```

to override this we should use Paginator::resolveCurrentPage($pageName):

```code
public function getPage($pageName = 'page'){
        return $this->paginators[$pageName] ?? Paginator::resolveCurrentPage($pageName);
    }
```

to get the exact page number after reload.


joshhanley:
@helturkey thanks for the PR! Could you please add a test? Thanks!

helturkey:
@joshhanley test added, Thanks

joshhanley:
@helturkey thanks!

PR #6737, opened 2023-09-09
Fix capitalization of `app` directory and remove unnecessary hyphen

peterhauke:
On Windows and in the Laravel 10 docs, the `app` directory has a lowercase `a`.

"real time" is not being used as a compound adjective here, so the hyphen is unnecessary.


joshhanley:
Thanks!

PR #6756, opened 2023-09-11
Fix `$persist` not working anymore with `@entangle`

PhiloNL:
Add failing test for https://github.com/livewire/livewire/discussions/6271.

When using Alpine's `$persist` helper the value is reset when reloading the browser.
```
<div x-data="{ test: $persist(@entangle('test')) }">
    <input x-model="test" />
</div>
```


joshhanley:
@PhiloNL thanks for the failing test. When comparing V2 entangle with V3 entangle, I found that this is the order things get processed in:
- V2: Entangle -> Persist
- V3: Persist -> Entangle.

In V2 all entangles are processed first, and then persisted values are then processed. If the persisted value is different to Livewire's value, due to the now active entangle, then Alpine will update Livewire's value to the persisted value (in turn triggering a network request).

It turns out the V3 implementation, which processes the entangle, was wrapped in `queueMicrotask`. Because of that persist was being processed first, setting the Alpine property to the persisted value, but then the queueMicrotask runs, procsses the entangle, and sets the Alpine property value to Livewire's value.
 
So I have removed the `queueMicrotask` and everything is working as it did in V2, except the extra network request doesn't fire now as updates are deferred by default - not sure what the implications are of that, should be fine?

calebporzio:
If everything else works, I think we're good here. Thanks Josh!

PR #6757, opened 2023-09-11
Full page components that are lazy loaded don't evaluate javascript

PhiloNL:
Failing test for https://github.com/livewire/livewire/discussions/6749.
`<script>` tags don't seem to be evaluated when using Lazy loading. It does seem to evaluate Alpine init handlers though.

```blade
<!-- Won't trigger -->
<script>console.log('foo');</script>

<!-- Will trigger -->
<div x-data x-init="console.log('foo');"></div>
```


francoism90:
@PhiloNL Is this related?
```
Alpine Expression Error: 'open' called on an object that does not implement interface Window.

Expression: "open"
```
This happens when using `wire:navigation` and using Alpine (e.g. `x-init x-data="{ open: false }"`.

Is wire:navigation related to lazy? If not, I think a new ticket needs to be created, sorry. :/

PhiloNL:
> @PhiloNL Is this related?
> 
> ```
> 
> Alpine Expression Error: 'open' called on an object that does not implement interface Window.
> 
> 
> 
> Expression: "open"
> 
> ```
> 
> This happens when using `wire:navigation` and using Alpine (e.g. `x-init x-data="{ open: false }"`.
> 
> 
> 
> Is wire:navigation related to lazy? If not, I think a new ticket needs to be created, sorry. :/

I think this is a different problem.

joshhanley:
@PhiloNL at the moment, Livewire can't dynamically load scripts because it has no way to update the scripts stack in the layout once the page has been loaded. We attempted to add a feature called dynamic stacks to V2 but weren't happy with the implementation.

I believe it's something we should try to get working for V3 as I am constantly running into issues with conditionally added blade components that contain scripts. But it's not a small thing. So let's see what Caleb thinks 🙂

PhiloNL:
@joshhanley @calebporzio what do you guys think of some sort of dependency injection system like the good old WordPress days e.g. [wp_register_script](https://developer.wordpress.org/reference/functions/wp_register_script/).

So instead of having to deal with Blade, we add some sort of API to the component:

```php
#[Script(path: 'js/foobar.js', defer: false)]
class Example extends Component {}

#[Script(url: 'https://some-cdn/script.js')]
class Example extends Component {}
```

I suspect that this is much easier to track and load the scripts if missing.

isaacdew2:
Is it possible that the solution lies in the JS when the DOM is being morphed? Adding this JS solves the issue for me but is admittedly a little hacky. Maybe something similar can be done if the stack blade directive were to add an element that JS can append to based on an ID?
```js
Livewire.hook('element.init', ({ component, el }) => {
  if (el.tagName === 'SCRIPT') {
    let script = document.createElement('script');
    script.innerText = el.innerText;
    if (el.hasAttributes()) {
      for (const attribute of el.attributes) {
        script.setAttribute(attribute.name, attribute.value);
      }
    }

    let body = document.querySelector('body');
    body.appendChild(script);
    el.remove();
  }
});
```

alexjustesen:
Following along, just ran into this issue

jscott-rawnet:
> ```js
> Livewire.hook('element.init', ({ component, el }) => {
>   if (el.tagName === 'SCRIPT') {
>     let script = document.createElement('script');
>     script.innerText = el.innerText;
>     if (el.hasAttributes()) {
>       for (const attribute of el.attributes) {
>         script.setAttribute(attribute.name, attribute.value);
>       }
>     }
> 
>     let body = document.querySelector('body');
>     body.appendChild(script);
>     el.remove();
>   }
> });
> ```

Can confirm, this appears to be a workaround that functions, with one small change:

```
Livewire.hook('element.init', ({ component, el }) => {
    if (el.tagName === 'SCRIPT') {
        let script = document.createElement('script');
        script.innerHTML = el.innerHTML;
        if (el.hasAttributes()) {
            for (const attribute of el.attributes) {
                script.setAttribute(attribute.name, attribute.value);
            }
        }
        let body = document.querySelector('body');
        body.appendChild(script);
        el.remove();
    }
});

```

( `script.innerText = el. innerText`; becomes `script.innerHTML = el.innerHTML`; )

innerText seems to include <br> tags etc within the js and breaks it.

isaacdew2:
@jscott-rawnet Good notes!

I eventually ran into the same issue with non-lazy components and solved it by adding `&& Livewire.initialRenderIsFinished` to the conditional.

**Caleb is working on a really cool and much better solution to this so I recommend waiting for that if you can.** In the meantime though, this seems to be an okay solution if you absolutely need it.

Another issue I ran into was the scripts getting duplicated every time the lazy component was loaded again. I solved this by adding IDs to the script tags and checking if they had been appended to the body yet.
```js
Livewire.hook('element.init', ({ component, el }) => {
  if (el.tagName === 'SCRIPT' && Livewire.initialRenderIsFinished) {    
    let alreadyExists = [...document.querySelectorAll('body > script')]
      .some(node => el.id && el.src && node.id === el.id);
    if (alreadyExists) {
      return;
    }
    let script = document.createElement('script');
    script.innerHTML = el.innerHTML;
    if (el.hasAttributes()) {
      for (const attribute of el.attributes) {
        script.setAttribute(attribute.name, attribute.value);
      }
    }
    
    let body = document.querySelector('body');
    body.appendChild(script);
    el.remove();
  }
});
```

potsky:
Same problem without the lazy mode but just by loading dynamically livewire components, the `<script>` tag is not parsed by the browser.

The code snippet just above works fine but just keep in mind that if you have a `@if` statement in your script tag like this:

```html
            @if (!empty($reloadEventName))
                if (...) {
                    // do something
                }
            @endif
```

it will be injected in HTML like this:

```html
            <!--[if BLOCK]><![endif]--> if (...) {
                    // do something
                }
```

And the browser will crash because it will not parse the first line and the last `}` will not have the opening one from `if (...) {`

GC-Mark:
As per @potsky comments, we are seeing the same issue without the lazy load too.

We were relyin on functionality like this...

```html
@if($successfulAuth)
    <script>
        doSomething()
    <script>
@endif
```

`doSomething()` is no longer triggered  in `v3`.

potsky:
@GC-Mark I have updated my previous answer! There is a solution in fact!

GC-Mark:
Thanks! I'll test it out

GC-Mark:
No, that doesn't work for us. even without the `@if` statement, the `<script>` tag never gets parsed on render.

potsky:
> No, that doesn't work for us. even without the `@if` statement, the `<script>` tag never gets parsed on render.

No indeed, you need to paste the code above (https://github.com/livewire/livewire/pull/6757#issuecomment-1804304260) in your `app.js`. My comment above the `@if` statement is an other problem related to script injection!

davitrix:
tengo este mismo problema, estoy intentando ejecutar un script dentro de un componente livewire que se carga de forma diferida pero no pasa absolutamente nada

` <div class="col-12 col-md-6">
                <div class="bg-white shadow rounded py-3 h-100 d-flex align-items-center">
                    <livewire:graficas.grafica-comparativa-ingresos-anio-presente-pasado lazy />
                </div>
            </div>`

`<div class="w-100">

    <div class="d-flex justify-content-between align-items-center mx-3 mb-3">
        <span class="fw-bold">Ventas mensuales {{ $chart['anios'][0] }} vs {{ $chart['anios'][1] }}</span>
    </div>

    <div id="comparativaVentasAniosPasadoPresente"></div>

    <script>
        document.addEventListener('livewire:init', function() {


            alert("aqui se deberia ejecutar mi script pero no lo hace, por lo tanto este alert nunca aparecera");


        });
    </script>



</div>
`

isaacdew2:
Caleb just released a better way to do this - https://livewire.laravel.com/docs/javascript#executing-scripts

Now that there's an official solution in place, I recommend removing the JS above and switching to that by updating to the latest Livewire release and using the `@script` blade directive. From the docs:
```blade
<div>
    ...
</div>
 
@script
<script>
    // This Javascript will get executed every time this component is loaded onto the page...
</script>
@endscript
```

Issue #2686, opened 2021-04-02
Pagination not populating querystring


ctschetter:
### Exact steps to reproduce
The route in web.php:
```
Route::get('/reviews/{slug}', ShowProfile::class);
```

ShowProfile:
```<?php

namespace App\Http\Livewire;

use App\Models\Profile;
use App\Models\ProfileReview;
use Livewire\Component;
use Livewire\WithPagination;

class ShowProfile extends Component
{
    use WithPagination;
    public $profile;


public function mount($slug)
    {
        $this->profile = Profile::where('name_slug', $slug)->orderBy('id', 'desc')->first();
    }

public function render()
    {
        $reviews = ProfileReview::where('profile_id', $this->profile->id)->orderBy('created_at', 'desc')->paginate(10);
        return view('livewire.show-profile', [
            'reviews' => $reviews
        ])->layout('layouts.search');
    }
}
```

And finally in show-profile.blade.php:
```
<div class="pt-6">
    {{ $reviews->links('vendor.livewire.tailwind') }}
</div>
```

The links appear fine, they page back and forth as you would expect, and manually changing the URL to include ?page=2 also works. The only problem is the URL does not change or ever show the page query.

**Context**
- Livewire version: 2.4.1
- Laravel version: 8.35.1
- Alpine version: 2.8.1
- Browser: Chrome

ctschetter:
I was able to get the query string to update properly by REMOVING 
```
use WithPagination;
```
and changing
```
{{ $reviews->links('vendor.livewire.tailwind') }}
```
to
```
{{ $reviews->links() }}
```
But the pagination templates are stock frrom the livewire:publish - I have not made any changes. Tried all of the above in a fresh project and the results are the same.

joshhanley:
@ctschetter I have just run a test on Livewire 2.4.1 locally and pagination works fine for me and the query string updates.

Are you able to supply an example repo demonstrating the issue?

Also you don't need to specify `vendor.livewire.tailwind` in your links method to use the published pagination views, as it will use them automatically.

But you do need to use `use WithPagination;`, otherwise, as you said, you end up with full page loads.

Here is the component I'm using to test , its a full page component

web.php
```php
Route::get('/', Main::class);
```

main.php
```php
<?php

namespace App\Http\Livewire;

use App\Models\User;
use Livewire\Component;
use Livewire\WithPagination;

class Main extends Component
{
    use WithPagination;

    public function render()
    {
[...]

PR #6760, opened 2023-09-11
Fix pagination query string not set when using lazy components

PhiloNL:
Failing test for https://github.com/livewire/livewire/discussions/6527.
If you enable lazy loading on a component the `page` query parameter will no longer be set.


joshhanley:
@PhiloNL thanks for the failing test!

I've been working on a fix for this issue. There are some failing tests, one which is related and two that are not (teleport back button stuff - these are also failing on PR #6759 after I merged in master).

Normally (without lazy), a component's root element will have an attribute `wire:effects` which contains the logic for setting up pagination using the query string on the front end.

But the issue with the lazy component is the component gets initialised on the frontend, and the (empty) `wire:effects` get processed. Then the component actually mounts and initialises the pagination on the back end. But as the query string feature hasn't been told to track the paginators property, the query string then never gets updated.

So this fix has required a few changes:

One is `SupportPagination` should add an effect to the first json payload (aka the first time the lazy component is actually mounted).

I then had to update the queryString implementation on the front end to listen to effects changes on subsequent requests and not just `component.init` which it was before. That way the effect that comes through from mounting the lazy component actually can be processed.

So the above is all working happily.

The last piece, which I haven't managed to get running yet (which you can see in the pagination failing test) is loading the initial pagination pages from the queryString. Currently, when the lazy component mounts and pagination is initialised, the request URL and queryString is actually that of Livewire's update route, as such it doesn't have any pagination page details in the queryString at all.

So not sure how to solve that just yet, but open to ideas.

calebporzio:
Good points @joshhanley - this is a deep one.

I think I need to explore a different, more stable implementation for lazy components, before we add lots of little conditionals for lazy components.

#6992

mozex:
Any update on this?

calebporzio:
fixed in #2686

HuseyinYeldan:
Still pagination with lazy loading doesn't work. Any fix?

PR #6761, opened 2023-09-11
Attribute validation rules precedence

PhiloNL:
Add failing test for https://github.com/livewire/livewire/discussions/6734.

When using repeated validation attributes, validation seems to be done in the wrong order:

```php
#[Rule('required', message: 'Please provide a post title')]
#[Rule('min:3', message: 'This title is too short')]
public $title;
```

This will return `This title is too short` while `required` should take precedence. 
If you swap the attribute order, it will return the correct error message:

```php
#[Rule('min:3', message: 'This title is too short')]
#[Rule('required', message: 'Please provide a post title')]
public $title;
```


calebporzio:
Thanks @PhiloNL!

PR #6762, opened 2023-09-11
Modelable input with .live modifier doesn't work

PhiloNL:
Add failing test for https://github.com/livewire/livewire/discussions/6275.
You can't seem to use the `.live` modifier on modelable inputs.
I'm also curious how to approach this from a precedence perspective, what if you add `.live` from the parent and the child doesn't have a `.live` modifier? Will this then automatically make this a `live` property? Or is the child component required to have a `.live` modifier 🤔 which kind of defeats the purpose of adding `.live` from the parent. 

I think in the ideal world you can add a `.live` modifier from the parent and this will make the `modelable` property live if it isn't already. Or in other words, it inherits all parent modifiers like `blur`, `live.debounce` etc.

```blade
<!-- Parent -->
<livewire:foo wire:model.live.debounce="bar">

<!-- Foo Component -->
<input wire:model="value">

<!-- Should dynamically become based on the parent calling the component -->
<input wire:model.live.debounce="value">
```


joshhanley:
@PhiloNL this should have been fixed in PR #7173

PhiloNL:
@joshhanley perfect, closing this 😄

PR #6763, opened 2023-09-11
Fix morph aware statements case sensitivity

gehrisandro:
This PR makes the insertion of morph markers case insensitive.

At the moment morph markers are only inserted if the blade directive is written in all lowercase but Laravels Blade directives are case insenstive. `@IF(true)` for example is valid, but Livewire does not insert the morph marker.


calebporzio:
Great, thanks!

PR #6770, opened 2023-09-11
Command enhancement for livewire:make using Laravel Prompts

PhiloNL:
The first step towards implementing Laravel Prompts. If you run `php artisan livewire:make` without any arguments it will prompt for the various options.

image

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
+++ b/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
@@ -154,7 +154,7 @@ class MakeCommand extends FileManipulationCommand implements PromptsForMissingIn
 
         if(
             $testSuite = select(
-                label: 'Do you want to create an test?',
+                label: 'Do you want to create a test file?',
                 options: [
                     false => 'No',
                     'phpunit' => 'PHPUnit',
--- a/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
+++ b/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
@@ -2,9 +2,14 @@
 
 namespace Livewire\Features\SupportConsoleCommands\Commands;
 
+use Illuminate\Contracts\Console\PromptsForMissingInput;
 use Illuminate\Support\Facades\File;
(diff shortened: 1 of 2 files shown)


src/Features/SupportConsoleCommands/Commands/MakeCommand.php
```diff
@@ -131,6 +136,41 @@ public function isReservedClassName($name)
...
+                default: false
+            )
+        )
+        {
+            $input->setOption('inline', true);
+        }
+
+        if(
+            $testSuite = select(
+                label: 'Do you want to create an test?',
```
chillbram, line 157: Typo: "an" should be "a". Maybe it's even better to extend it to say `Do you want to create a test file?`, to clarify its functionality a little bit more

PhiloNL: Thanks! Updated :)

utsavsomaiya:
I think this code shouldn't be maintainable because there is a multiple scenario are covered!!

1. For first livewire installed user to display the livewire logo and open the github repo...
2. Other we need to use the stub files and validate it and then we create it!!

---
So can we create as laravel does? e.g. : https://github.com/laravel/installer/blob/master/src/NewCommand.php

PhiloNL:
> I think this code shouldn't be maintainable because there is multiple scenario are covered!!
> 
> 
> 
> 1. For first livewire installed user to display the livewire logo and open the github repo...
> 
> 2. Other we need to use the stub files and validate it and then we create it!!
> 
> 
> 
> ---
> 
> So can we create as laravel does? e.g. : https://github.com/laravel/installer/blob/master/src/NewCommand.php
> 
> 

Sorry I don't follow. This PR is just an enhancement that will prompt for input if you don't provide any arguments instead of throwing an error.

calebporzio:
Awesome, thanks Philo!

calebporzio:
Hey @PhiloNL,I had to revert this, this is what showed up for me locally:

I tried tweaking it and couldn't get it. If you get the time, can you fix and re-PR - thanks!

image

PR #6774, opened 2023-09-11
Add `shouldBoot` hook for conditional component features

joshhanley:
This PR adds a `shouldBoot` hook for conditional component features, such as `SupportPagination`. See PR #6708 for details.

I've implemented it so that any existing features, that currently don't have a `shouldBoot` method, still boot anyway.

I looked at calling it like the other hooks using `proxyCallToHooks` and having a `callShouldBoot` method on the base `ComponentHook` class, but it looked like it wouldn't work due to needing the return value, so just called it manually if it existed.

Hope this helps!

### DIFF
--- a/src/ComponentHookRegistry.php
+++ b/src/ComponentHookRegistry.php
@@ -37,20 +37,18 @@ class ComponentHookRegistry
 
         foreach (static::$componentHooks as $hook) {
             on('mount', function ($component, $params, $key, $parent) use ($hook) {
-                $hook = static::initializeHook($hook, $component);
-                
-                // If hook hasn't been initialized, then don't continue.
-                if (!$hook) return;
+                if (! $hook = static::initializeHook($hook, $component)) {
+                    return;
+                }
 
                 $hook->callBoot();
                 $hook->callMount($params, $parent);
             });
 
             on('hydrate', function ($component, $memo) use ($hook) {
-                $hook = static::initializeHook($hook, $component);
-                
-                // If hook hasn't been initialized, then don't continue.
-                if (!$hook) return;
+                if (! $hook = static::initializeHook($hook, $component)) {
+                    return;
+                }
 
                 $hook->callBoot();
                 $hook->callHydrate($memo);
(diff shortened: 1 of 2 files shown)


calebporzio:
Sounds good, thanks Josh. I changed it to a "skip" method instead of "shouldBoot" heads up

devajmeireles:
> This PR adds a `shouldBoot` hook for conditional component features, such as `SupportPagination`. See PR #6708 for details.
> 
> I've implemented it so that any existing features, that currently don't have a `shouldBoot` method, still boot anyway.
> 
> I looked at calling it like the other hooks using `proxyCallToHooks` and having a `callShouldBoot` method on the base `ComponentHook` class, but it looked like it wouldn't work due to needing the return value, so just called it manually if it existed.
> 
> Hope this helps!

It should be documented?

joshhanley:
@devajmeireles nah, it's for internal features only, not components themselves.

lrljoe:
@joshhanley - I use the ComponentHookRegistry to auto-inject some additional JS/CSS for a package 
(I.e. a chunk of Alpine.data() items and custom functions that are used in various blades throughout the package) 

Not sure, but it may be useful to document for others, unless I've gone nuts in my approach!

joshhanley:
@lrljoe at this stage, I would say use at your own risk. I'd have to check with Caleb about it, but for now I would say treat it as if it's not considered part of the supported public api so could be subject to breaking changes. But, I agree. Once we're on top of things, I think it could be a useful extension point for people writing their own features. At that stage (and if Caleb agrees) then yeah we will document it 🙂

lrljoe:
Oh absolutely at-risk, and definitely steering people towards traditional approaches, but some people like autoinject.

 It's definitely handy, and once I got past it re-registering hooks, it works flawlessly.

  If you guys need any examples of random implementations using it, then give me a shout.

lrljoe:
Oh absolutely at-risk, and definitely steering people towards traditional approaches, but some people like autoinject.

 It's definitely handy, and once I got past it re-registering hooks, it works flawlessly.

  If you guys need any examples of random implementations using it, then give me a shout.

PR #6785, opened 2023-09-12
Child elements don't render after failed validation

jerodev:
After validating a component using `$this->validate()`, child components of that component are no longer rendered.

The browsers' developer tools indicate that the snapshot of the component is no longer available so the child elements container is empty. Below is a screenshot from the application where I encountered the issue.

image

This PR adds a failing test for this issue as I'm unable to find the cause.

### DIFF
--- a/src/Features/SupportNestingComponents/BrowserTest.php
+++ b/src/Features/SupportNestingComponents/BrowserTest.php
@@ -9,6 +9,58 @@ use Livewire\Livewire;
 class BrowserTest extends \Tests\BrowserTestCase
 {
     /** @test */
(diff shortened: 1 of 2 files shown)


joshhanley:
@jerodev thanks for the PR! I checked out your test, and the reason the test is failing is Livewire components are isolated once they are loaded. So the parent component won't re-render the child component view - hence why the unit test is failing.

I've converted your test to a browser test and it seems to be passing and it's not getting the snapshot missing error in the browser console.

I'm going to close this for now, but if you manage to replicate the issue in the browser test, let me know and I can reopen it or feel free to open a new PR. Thanks!

PR #6811, opened 2023-09-14
Remove console.log

PhiloNL:
Remove `console.log`


PhiloNL:
> @PhiloNL this broke tests from 417592d08812623dd5408597c697200a3fbf2f96

The test will pass, but it pulled down the latest tag of Alpine that don't yet include the main branch changes. Once the main branch changes are back in it will pass.

PR #6819, opened 2023-09-14
Add wire poll isolate

calebporzio:
`<div wire:poll.isolate>`

This PR is working but there are 2 problem scenarios if a batched request fires simultaneously with an isolated one, and the batched request takes a long time (long enough for the isolate request to attempt a poll again)

Here are the scenarios 

We have 3 components polling:
* one is isolated (A) the rest are not (B, C)
* Component's A & (B/C) all send at first
* If the (B/C) request takes too long, component A's polling is held up

Alternatively:
* We have 3 components polling
* one is isolated (C) the rest are not (A, B)
* Component's (A/B) & C all send at first
* If the (A/B) request takes too long, component C spawns an entirely new request for A/B while the other one is held up


woodspire:
Here is my opinion, do whatever you want with it.

Instead of doing isolation and group requests differently, why not simply support groups requests.

But default, 1 group is created and all request go in that group.

If some component (or only one) need to be in a specific group: wire.poll.group.#name where #name is the group name.
With that, setting PER group can be applied (I don't know where right now) like the interval to poll, etc...

wire.poll.isolate simply create a unique group and add it to the group list. Bam !


so 3 wire.poll.isolate component will result in 4 groups:

- default one
- 3 unique groups for each isolated components

Then Livewire "engine" simply handle "groups" when doing the requests.

markjaquith:
Where was the test code you were using when developing this on stream? Could you dump that here?

calebporzio:
Here @markjaquith:

This goes in the main page view:
```html
<livewire:test-polling />
    <livewire:test-polling />
    <livewire:test-polling />
    <livewire:test-polling />
    <livewire:test-polling />
    <livewire:test-polling-isolated />
```

```php
class TestPolling extends Component
{
    function hydrate()
    {
        // sleep(2);
    }

    public function render()
    {
        return <<<'HTML'
        <div wire:poll>
            {{ now() }}
        </div>
        HTML;
    }
}
```

```php
class TestPollingIsolated extends Component
{
    function hydrate()
    {
        sleep(1);
    }

    public function render()
    {
        return <<<'HTML'
        <div wire:poll.isolate>
            {{ now() }}
        </div>
        HTML;
    }
}

```

eilerth:
I hate to be "that person", but I need to ask- any idea when this might be able to get merged?

It has become a blocker (without really messing up the user experience) for us to upgrade to livewire v3 (which itself is a blocker for other stuff)

ErsanJSano:
Checking if any timing is available for this PR. It is a blocker so hopping to get some info to better decide next steps. Thanks much and sorry for trouble.

markjaquith:
@eilerth @ErsanJSano This isn't ready to merge. I was hoping to have time to dig into the issue, but haven't had a chance. ☹️

eilerth:
@markjaquith I took a closer look at this PR and found it wouldn't actually solve the problem I'm faced with. This is very focused on `wire:poll`, but most of the problems I'm faced with are not even using `wire:poll`

This is very similar, though, and I found some of the logic helpful in creating a PR of my own to solve the problem I'm faced with: https://github.com/livewire/livewire/pull/7102

PR #6821, opened 2023-09-14
Allow configurable computed prop cache keys

jasonlbeggs:
This allows users to set a custom cache key for computed props so they can be manually cleared easier if necessary.

```php
#[Computed(cache: true, cacheKey: 'featured-article')]
public featuredArticle()
{
    // ...
}
```


calebporzio:
@jasonlbeggs - this is a dandy PR you got here. Mind adding docs for it?

jasonlbeggs:
Documented 👍

calebporzio:
tytytytyty

PR #6822, opened 2023-09-15
Add support for adding cookies to tests

joshhanley:
Laravel has `withCookie()` and `withCookies()` methods for injecting cookies to a request for testing purposes. But currently there is no way to use cookies with a Livewire test.

This PR adds Livewire support for `withCookie()` and `withCookies()` with the same API as Laravel https://laravel.com/docs/10.x/http-tests#cookies

They can be used like:

```php
Livewire::withCookie('name', 'Taylor')
    ->test(MyComponent::class);
```

```php
Livewire::withCookies(['colour' => 'blue, 'name' => 'Taylor'])
    ->test(MyComponent::class);
```

I have also added docs to `testing.md`.

Hope this helps!

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -319,6 +319,57 @@ class SearchPostsTest extends TestCase
 }
 ```
 
+### Setting cookies
+
+If your Livewire component depends on cookies, you can use the `withCookie()` or `withCookies()` methods to set the cookies manually for your test.
+
+Below is a basic `Cart` component that loads a discount token from a cookie on mount:
+
+```php
+<?php
+
+namespace App\Livewire;
+
+use Livewire\Component;
+use Livewire\With\Url;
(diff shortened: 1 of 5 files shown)


satoved:
This is awesome!

calebporzio:
Perfect, thanks Josh!

joshhanley:
No worries! 🙂

riomigal:
Hi. Would it be possible to merge this as well in version 2.12? We are still using laravel 9 and can't upgrade at the moment.

joshhanley:
@riomigal no, sorry. V2 development has finished. So only bug fixes will likely be backported to V2.

fikri-kompanion:
This works with cookie if you are in mount or render method. What about cookie inside action? it will return null

PR #6824, opened 2023-09-15
Revert new code causing navigate to remove injected assets such as nprogress styles

joshhanley:
Since commit 5c28851 any injected assets, such as nprogress styles, are being removed on page change.

See discussion #6823.

This PR adds a failing test covering this issue.

This is now happening because of the below addition.

https://github.com/livewire/livewire/blob/23d035b6353c6dfe7ea9c8d288295dbbb46ef84a/js/plugins/navigate/page.js#L102-L109 

What I'm not sure, is how should we handle this?

In theory, we could change Livewire to make use of the `wire:navigated` event and re-inject the nprogress styles on the new page.

Do other libraries also need to implement this? Not sure how to progress.


levu42:
Since v3.0.4 I can't use `wire:navigate` anymore, because it removes all stylesheets (in head) on navigate, e.g. the output of `@vite(['resources/js/app.js'])`.

joshhanley:
Short term, I have removed some the new code that was just added in 5c28851 that cleans up the old styles, so that this issue is no longer happening.

This is just a quick fix and will need to be looked into further.

joshhanley:
@levu42 this fix has been tagged and released in v3.0.5.

levu42:
> @levu42 this fix has been tagged and released in v3.0.5.

@joshhanley Thanks, Josh!

PR #6830, opened 2023-09-15
Add failing test for checkboxes not using `value` attribute correctly

Simoneu01:
In V3 when I'm trying to update a value of an array with a custom index it will set the value only to true/false.

In V2 everything is working as expected, this is the code that I've used in the V2 repo to test it:

Component:
```php
<?php
namespace Tests\Browser\DataBinding\InputCheckboxRadio;
use Livewire\Component as BaseComponent;
class CheckboxesWithString extends BaseComponent
{
    public $filters = [];
    public function render()
    {
        return <<< 'HTML'
<div>
    @foreach(['a', 'b', 'c'] as $key)
        <input dusk="checkbox.int.{{ $key }}" value="{{ $loop->index + 1}}" type="checkbox" wire:model="filters.{{ $key }}" />
    @endforeach
    @foreach(['d', 'e', 'f'] as $key)
        <input dusk="checkbox.string.{{ $key }}" value="value_{{ $key }}" type="checkbox" wire:model="filters.{{ $key }}" />
    @endforeach
    <button wire:click="$set('filters', [])" dusk="filters.reset">Reset</button>
    <div dusk="output">@json($filters)</div>
</div>
HTML;
    }
}
```

Test in tests/Browser/DataBinding/InputCheckboxRadio/Test.php
```php
   function test_can_set_array_value()
    {
        $this->browse(function (Browser $browser) {
            Livewire::visit($browser, CheckboxesWithString::class)
                ->check('@checkbox.int.a')
                ->check('@checkbox.int.b')
                ->check('@checkbox.int.c')
                ->waitForLivewireToLoad()
                ->assertSeeIn('@output', '{"a":"1","b":"2","c":"3"}')
                ->click('@filters.reset')
                ->waitForLivewireToLoad()
                ->assertSeeIn('@output', '[]')
                ->check('@checkbox.string.d')
                ->check('@checkbox.string.e')
                ->check('@checkbox.string.f')
                ->waitForLivewireToLoad()
                ->assertSeeIn('@output', '{"d":"value_d","e":"value_e","f":"value_f"}');
        });
    }
```

Result:
Screenshot 2023-09-15 alle 15 31 39


calebporzio:
I see that this is different than in V2.

However, Livewire V3 now uses Alpine's x-model under the hood for wire:model.

This means we would have to change x-model in Alpine.

I'm open to that, but in general I like to keep parity with VueJS for these kinds of behaviors.

I checked and VueJS works like Alpine currently works:
CleanShot 2024-01-20 at 16 06 23@2x

Therefore, I'm closing this and probably not going to change it. Thanks!

ITSYV:
@calebporzio Can you clarify? We're having the same issue with dynamically created properties (ex. wire:model="ticket.ID" which doesn't use the value but returns true/false. 

The Livewire docs state that you can use the value field for checkboxes ( https://livewire.laravel.com/docs/wire-model#multiple-checkboxes ), but does this only work for properties which are not dynamically created (ticket.*? Or do I misunderstand you?

gregorysouzasilva:
Hi @calebporzio. I understand you want to keep alpine with the same behavior however this makes very hard to handle checkboxes. For models we can cast to convert to boolean but If I have a json field, I would need to loop them to change all the values to boolean before rendering. Wouldn't be better to handle on livewire if it's a checkbox check the value parameter and check if it's true or not? Otherwise it will mislead the person to thing that should work setting a value for checkbox. Tks.

PR #6831, opened 2023-09-15
Fix infinite loading when `delay` modifier is applied to `wire:loading`

michael-rubel:
It seems `wire:loading` wents mad when the `delay` modifier is used.


joshhanley:
@michael-rubel thanks for the PR!

I pulled down your PR but wasn't able to replicate the issue with your tests as they would pass locally without your fix.
I'm going to close this for now, but if you're still experiencing the issue, can you share a small example repo that demonstrates the issue so we can investigate?

patrickweh:
hi @joshhanley,

i have the same issue, it ssems like it happens when a snapshot without a method call or update is performed.

when i delete an item of my userFilters array this is called, the spinner shows up and is removed:
image

after this call this one is called, the spinner shows up and stays where it is even after the request is finished.

image

when i remove the delay from the spinner everything works fine.

PR #6833, opened 2023-09-15
Add failing test for a bug with removal of nested entangled properties

satoved:
I'm submitting a failing test for a scenario with nested entangled properties involved.

This same scenario would work fine on V2.

**Scenario**
1) There is a nested array of data
```
            public array $items = [
                [
                    'id' => 1,
                    'value' => 5_000,
                ],
                [
                    'id' => 2,
                    'value' => 10_000,
                ],
                [
                    'id' => 3,
                    'value' => 15_000,
                ]
            ];
```
2) I'm removing values one by one from last to first using array key
```
            public function remove($key)
            {
                // Removing an item from array
                unset($this->items[$key]);

                // Resetting the array keys
                $this->items = array_values($this->items);
            }
```

**Problem**
I expect to have an empty array after removing all items from it.

✅ It works fine with normal _wire:model_
❌ It fails with _@entangle_, though I expect the behavior to be the same as _wire:model_

**What happens**
From my shallow understanding, after I'm doing the third removal, for some reason, it sets the entangled value to null.

I expect to get $items to be `[]`

But what I get is:
```
[
    [
        'value' => null
    ]
]
```

It ruins the structure, because the 'id' key is missing, and I get 500 because I try to access `$item['id']` in the template.

image

Might be related to: #6528 #6521 #6830

Thanks.

P.S. The passing test is added to better illustrate the issue, but it shouldn't be merged to real tests.


joshhanley:
@satoved thanks for the failing test!

I think I have worked out what is going on. When we remove elements as in this test, the entangle effect isn't being cleaned up. So it's still trying to process empty data, causing an update to be added to the request where there should only be the "remove" method call.

image


When using `@entangle()` or `Livewire.find().$entangle()`, they both use Livewire's internal `$wire` object. But the problem with that is that there is no effect `cleanup` function passed into the `generateEntangleFunction()` as such an empty closure get's generated.

https://github.com/livewire/livewire/blob/b25402aff4911dbccae5583308e94af4c9681ebb/js/%24wire.js#L118-L120

https://github.com/livewire/livewire/blob/b25402aff4911dbccae5583308e94af4c9681ebb/js/features/supportEntangle.js#L5-L6

But if we instead use Alpine's magic `$wire.$entangle()`, then everything works correctly as the cleanup closure is passed into the `generateEntangleFunction()` correctly 

https://github.com/livewire/livewire/blob/b25402aff4911dbccae5583308e94af4c9681ebb/js/%24wire.js#L85-L87

So the simple work around for now is to change from using `@entangle()` to using `$wire.$entangle()`. If we update the test as an example, we would change from

```blade
<input x-data="{ value: @entangle('items.'.$itemKey.'.value') }"
```

 to
 ```blade
 <input x-data="{ value: $wire.$entangle('items.{{ $itemKey }}.value') }"
 ```
 
 We will keep working on the problem to see if we can solve it, but the above work around should work happily.
 
 Hope this helps!

satoved:
@joshhanley thanks a lot, that helped!

ju5t:
@joshhanley it is a bit more involved for us as we have this issue inside blade-components. In a test-repo I've replaced this:

```blade
x-data="{ value: @entangle($attributes->wire('model')) }"
```

To this:

```blade
x-data="{ value: $wire.$entangle('{{ $attributes['wire:model'] }}') }"
```

But we need live updates (we use this for calculations) and I haven't been able to set that up yet. I've added the .live-modifier but that doesn't seem to work.

joshhanley:
@ju5t I'm not sure I follow, is using `$wire.$entangle` not working for you? They should work exactly the same.

satoved:
@ju5t it's not .live in this case, it's a "true" as second argument.

See https://livewire.laravel.com/docs/javascript#the-wire-object

image

Try this: 
```
x-data="{ value: $wire.$entangle('{{ $attributes['wire:model'] }}', true) }"
```

ju5t:
@joshhanley it does solve the issue of not being able to remove components from the dom, but it comes with some side-effects. For example:

```
<x-input.number wire:model.live.debounce="items.{{ $key }}.quantity"
```

The key in the `$attributes` array becomes `wire:model.live.debounce`. We may use different modifiers elsewhere. This component is used all over the place. And it's not the only component where we use `@entangle` too. 

I think we will delay the upgrade to v3 for now and hope the fix to get `@entangle` back to its old behaviour is not too difficult. 

@satoved that works. Thanks.

satoved:
@joshhanley are there any reason why blade @entangle can't actually be compiled to AlpineJS magic `$wire.$entangle('property')`?

AlpineJS is part of the core now, so can this be a solution?

ju5t:
@joshhanley @satoved speaking for our app, this would mean we require modifier support too. We use `blur` for example and this currently isn't supported with `$wire.$entangle`. That would make our life a little easier.

I was thinking, perhaps the upgrade guide should mention that `@entangle` is not fully supported in v3 at the moment.

joshhanley:
@ju5t Ah yeah I understand your issue. You shouldn't be using `{{ $attributes['wire:model'] }}` as you can't determine what modifiers might be on it. Instead you should be using `{{ $attributes->wire('model') }}` which will handle all the possible modifiers for you. It returns an object with more details of what has been passed in.

So like `$wire.$entangle('{{ $attributes->wire('model')->value }}')`.

Regarding blur, to me it doesn’t make sense to use blur on entangle, as blur is triggered by an input event. So that should be handled by the x-model on the input. I’m not sure how we would map from wire:model.blur to the x-model.blur as Livewire doesn’t know where the x-model is inside your alpine component. Inside you should probably detect that yourself and add blur if found on the wire:model. 

Re adding a warning, we’d rather just fix the issue. Just haven’t had time to dig into getting it sorted. 

@satoved yeah I suggested that to Caleb. The problem is in html it would need to output `$wire.$entangle` but inside a Alpine data script, it needs to be `this.$wire.$entangle`.

ju5t:
@joshhanley thanks! That works for us.

joshhanley:
I did a test by changing the underlying `@entangle` implementation to use `$wire.entangle()` and while it works fine for most cases, as suspected it does break the following scenario:

```js
Alpine.data('test', () => ({
    count: @entangle('count'), // `$wire` is invalid in this context, and even trying `this.$wire` is also invalid
    increment() {
        this.count++
    },
    decrement() {
        this.count--
    },
    reset() {
        this.count = 0
    },
}));
```

So this solution isn't going to work.

ju5t:
@joshhanley thanks for your efforts. I'm terrible at JS, so I don't know if this is possible, but could you make `$wire` available globally in these cases as well? Omitting `this`?

joshhanley:
@ju5t no worries!

Sadly not, it's the main issue with `@entangle()` expanding to `window.Livewire.find().entangle()`, where it needs a reference to the element that Alpine is initialised on, which using the Livewire global doesn't have.

`$wire` on it's own, is an Alpine magic that is bound to the element that has the `x-data` on it, so it all works happily and clean up runs without issue.

calebporzio:
Going to close this as there's no clear solution path for V3.

The answer to this problem for now is: Use `$wire.entangle()` instead of `@entangle` so that they get cleaned up when elements are removed.

Thanks!

PR #6843, opened 2023-09-17
Add option to modify full page component response

PhiloNL:
In Laravel you can modify response headers by using the `response` helper:

```php
return response()
            ->view('hello', $data, 200)
            ->header('Content-Type', $type);
```

This will not work when you use a full-page component as Livewire expects a view to be returned and not a response object.
This PR adds the option to hook into the response when rendering full-page components.

```php
class Dashboard extends Component
{
    public function render()
    {
        return view('livewire.dashboard')->response(function(Response $response) {
            $response->header('x-hello-world', true);
        });
    }
}
```

The class name `LayoutConfig` feels a bit off now because a response is not really related to the layout. Maybe rename `LayoutConfig` to `PageComponentConfig`?

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -634,3 +634,27 @@ class ShowPost extends Component
 ```
 
 The `$post` property will automatically be assigned to the model bound via the route's `{post}` parameter.
+
+### Modifying the response
+
+In some scenarios you might want to modify the response and set a custom response header. You can hook into the response object by calling the `response()` method on the view and use a closure to modify the response object:
+
+```php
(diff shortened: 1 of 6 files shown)


calebporzio:
This makes sense to me, thanks @PhiloNL - can you add documentation for this as well?

PhiloNL:
@calebporzio great! Added docs and renamed LayoutConfig to PageComponentConfig.

calebporzio:
Thanks Philo!

PR #6858, opened 2023-09-19
404 after deleting item via event

PhiloNL:
Add failing test for https://github.com/livewire/livewire/discussions/6792 and I think this might could be the same issue as https://github.com/livewire/livewire/discussions/6120 but a different error.


joshhanley:
@PhiloNL closing in favour of PR #7003. I have merged your PR into that one 🙂

PR #6872, opened 2023-09-20
Support Laravel-style CSRF token definition

zepfietje:

Required to support Laravel's way of defining the CSRF token (https://github.com/laravel/jetstream/blob/2f170dada485713c64853872256f8e2f7405e674/stubs/livewire/resources/views/layouts/app.blade.php#L6). When regenerating sessions after login, one can dispatch a Livewire event and update the token using JS. This worked in v2, but in v3 I've had to add `data-csrf="{{ csrf_token() }}"`. Since Laravel Jetstream uses a `<meta>` tag instead, I've added support for it. I've kept the check for `[data-csrf]` too, but I'm not sure what standard it is and if you'd like to keep it?

Yup.

Nope.

I believe there are no existing tests for this util?

See description at 1️⃣.

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -168,6 +168,11 @@ export function isSynthetic(subject) {
 export function getCsrfToken() {
     // Purposely not caching. Fetching it fresh every time ensures we're
     // not depending on a stale session's CSRF token...
+
+    if (document.querySelector('meta[name="csrf-token"]')) {
+        return document.querySelector('meta[name="csrf-token"]').getAttribute('content')
+    }
+
     if (document.querySelector('[data-csrf]')) {
         return document.querySelector('[data-csrf]').getAttribute('data-csrf')
     }


joshhanley:
@zepfietje thanks for the PR! `[data-csrf]` is an attribute on Livewire's script tags, added by Livewire with the CSRF token details. Can you explain in a bit more detail (possibly with code or repo) what wasn't working before before this PR?

zepfietje:
Ah, thanks for clarifying, @joshhanley!

So, I have a login form in a modal, which regenerates the session after successful sign in. Subsequent requests would throw a `419 Page Expired` exception, as the CSRF token has changed because of the new session. To work around this, I dispatch a `sessionRegenerated` event in Livewire which passes along the `csrf_token()` value. Then in JS, I update the CSRF token defined in the `<meta>` tag:
```js
Livewire.on('sessionRegenerated', ({ csrfToken }) => {
    document.head
        .querySelector('meta[name="csrf-token"]')
        .setAttribute('content', csrfToken);
});
```

This is how it used to work in Livewire v2. However, with v3 this stopped working. I'm not sure how v2 handled this exactly, but it seems like it used the CSRF token defined in the `<meta>` tag. I tried setting `window.livewireScriptConfig['csrf']` directly, but `window.livewireScriptConfig` was `undefined` in the way I tried.

I've temporarily fixed this issue by updating above JS snippet to:
```diff
Livewire.on('sessionRegenerated', ({ csrfToken }) => {
    document.head
        .querySelector('meta[name="csrf-token"]')
-       .setAttribute('content', csrfToken);
+       .setAttribute('data-csrf', csrfToken);
});
```

I know this is hacky (especially after reading your explanation 😂), and maybe I should update the `data-csrf` attribute on Livewire's script tag instead?

joshhanley:
@zepfietje yep, that makes sense! Thanks for the detail. 🙂

We recently added the code that makes sure it pulls the CSRF token from the page each time (hence the comment above your change) for the scenario of a regenerated session, but it looks like it doesn't deal with the backend side updating the layout with the latest CSRF token (after session regeneration), hence the 419.

Yeah I probably would just target the livewire scripts tag directly and update the `data-csrf` attribute on that. I think leave this open for now, as Caleb can have a look when he is back and see if he wants to approach it a different way, but auto-updating the CSRF token from the backend for you.

zepfietje:
Thanks, @joshhanley, I'll go with updating the `data-csrf` attribute on the script tag for now.

Do projects actually still need a CSRF token in a meta tag on the page?

zepfietje:
Yeah probably needed for some cases without Livewire, though probably not when all forms and such are Livewire. 😄

joshhanley:
@zepfietje yep exactly! 😁 Just out of curiosity, how come you don't do a redirect after login? As generally that will be fine as the new page will already have the CSRF token.

zepfietje:
Yeah, if only that would suffice haha!
The login/registration flow is kinda unique, since it pops up in a modal after following a creation flow (even when you have no account yet).

Easiest to show by just linking you there 👉 https://picci.app/en/forms/create.
It needs to save the new user (or get the authenticated user) first, then connect it to the form you're creating, and save the form. It also supports existing accounts without password, which will receive an email login link, which can be opened in a new tab. The main tab will then poll and see if the user is logged in, after which it saves the form. Complex? I know. 🫠😂

joshhanley:
@zepfietje ah nice! Makes sense, thanks for sharing 😁

calebporzio:
Yeah, this makes total sense. I recently removed the caching of this value, so after this addition there shouldn't be any more csrf issues like we've been having. Now anytime Livewire needs a csrf token it gets it fresh from either the meta tag or the script tag. Meta should take priority like you've done here. Thanks @zepfietje!

zepfietje:
Cool, thanks @calebporzio!

Note that users still have to update the meta tag using JS like I posted above.
Would there be some sort of way for Livewire to detect a change in CSRF token and automatically update the token? 🤔

PR #6883, opened 2023-09-21
Tweak forms.md

austincarpenter:

#6882

https://github.com/austincarpenter/livewire/tree/add-wire-key-in-forms-docs

No

No, as docs-only change.

#6882 

> With Livewire 2, when a dependent select's options were changed to no longer include the currently selected option (in terms of its value), it would default back to having selected a blank option.

> With Livewire 3, once the options are changed, the select will show the nth option available (where n is the selected option), despite its value being different.

Have since updated the discussion to recommend adding `wire:key`.

Thanks for contributing! 🙌

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -776,7 +776,7 @@ Often, you may build your dropdown options dynamically using Blade:
 ```blade
 <select wire:model="state">
     @foreach (\App\Models\State::all() as $state)
-        <option value="{{ $state->id }}">{{ $state->label }}</option>
+        <option value="{{ $state->id }}" wire:key="value="{{ $state->id }}">{{ $state->label }}</option>
     @endforeach
 </select>
 ```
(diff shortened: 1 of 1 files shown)


docs/forms.md
```diff
@@ -776,7 +776,7 @@ Often, you may build your dropdown options dynamically using Blade:
 ```blade
 <select wire:model="state">
     @foreach (\App\Models\State::all() as $state)
-        <option value="{{ $state->id }}">{{ $state->label }}</option>
+        <option value="{{ $state->id }}" wire:key="value="{{ $state->id }}">{{ $state->label }}</option>
```
PerryvanderMeer, line 779: This won't do the job..

PerryvanderMeer (changes requested):

docs/forms.md
austincarpenter: You certainly can't!

PR #6889, opened 2023-09-21
Add failing test for dependent select options

austincarpenter:

https://github.com/livewire/livewire/discussions/6882
https://github.com/filamentphp/filament/pull/8633

Yes

No

Yes

In Livewire 2, `select`s whose options are dependent on another property and enumerated through using `@foreach` work in a certain way. The added test uses the data below, where the array keys are options for the parent select and respective values are the options for the child (or dependent).
```php
'foo' => [
    'bar',
],
'baz' => [
    'qux',
],
```

Steps and expected output:
1. Select `foo` on parent
2. Expect to see `bar` as child option, select `bar` on child
3. Select `baz` on parent
4. Expect nothing to be selected on child
5. Select `foo` again on parent
6. Expect to see `bar` selected on child

In Livewire 3, step 4 sees `qux` selected on the child, as it is option #0. Adding `wire:key` as an attribute to the child's `option`s fixes step 4, however step 6 still sees nothing selected on the child, instead of `bar`.


austincarpenter:
This dependent style of select behaviour is extremely common and `wire:key` is not currently referenced in the docs (see https://github.com/austincarpenter/livewire/tree/add-wire-key-in-forms-docs, https://github.com/livewire/livewire/pull/6883). 

To reduce friction, this should perhaps be handled automatically (as would seem the case for v2).

Also is mentioned above, the explicit addition of `wire:key` does not fix step 6.

austincarpenter:
Test passing on CI/CD but not locally. Will need to investigate.

calebporzio:
Ok, so turns out adding `wire:key` to the second dependent select DOES fix both #4 and #6 of the steps you've described. 

It actually kind of makes sense that you would need to add wire:key, because the options are changing dynamically and you actually want completely separate select menus.

Because of this, I added documentation for dependent selects and changed the test to have wire:key and pass now.

Thanks!

austincarpenter:
Can't remember exactly, but perhaps in testing I had added `wire:key` to the options themselves, as recommended here https://laravel-livewire.com/docs/2.x/troubleshooting#dom-diffing-cures. Cheers all the same Caleb!

PR #6891, opened 2023-09-21
Fix asset publishing

soundslocke:
This was discussed over in https://github.com/livewire/livewire/discussions/6328 and according to @joshhanley's [comment](https://github.com/livewire/livewire/discussions/6328#discussioncomment-7063624) it seems that breaking the `livewire:publish --assets` command was an oversight in v3 and PRs are welcome to fix it.

It looks to me like the fix is to add the relevant `publishes()` call in `LivewireServiceProvider.php` to go along with the similar call for setting up publishing of the config file. So, that's what I've done here.

I also added a relevant section to documentation in `installation.md` where this command had been documented for v1 and v2.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -18,6 +18,26 @@ php artisan livewire:publish --config
 
 This will create a new `livewire.php` file in your Laravel application's `config` directory.
 
+## Publishing the frontend assets
+
+If you prefer the JavaScript assets to be served by your web server not through Laravel, use the `livewire:publish` command:
+
+```shell
+php artisan livewire:publish --assets
+```
+
+To keep the assets up-to-date and avoid issues in future updates, we **highly recommend** adding the command to the `post-update-cmd` scripts in your `composer.json` file:
+
+```json
+{
+    "scripts": {
+        "post-update-cmd": [
+            "@php artisan livewire:publish --assets"
+        ]
+    }
+}
+```
+
 ## Manually including Livewire's frontend assets
 
 By default, Livewire injects the JavaScript and CSS assets it needs into each page that includes a Livewire component.
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -10,6 +10,7 @@ class LivewireServiceProvider extends \Illuminate\Support\ServiceProvider
     {
         $this->registerLivewireSingleton();
         $this->registerConfig();
+        $this->registerAssets();
         $this->bootEventBus();
         $this->registerMechanisms();
     }
@@ -36,6 +37,13 @@ class LivewireServiceProvider extends \Illuminate\Support\ServiceProvider
         $this->publishes([$config => base_path('config/livewire.php')], ['livewire', 'livewire:config']);
 
         $this->mergeConfigFrom($config, 'livewire');
+    }
+
+    protected function registerAssets()
+    {
+        $assets = __DIR__.'/../dist';
+
+        $this->publishes([$assets => public_path('vendor/livewire')], ['livewire', 'livewire:assets']);
     }
 
     protected function bootEventBus()


SmmiZ:
Could you please tell us - when this changes will be applied?
It's annoying to manual update assets.

robsontenorio:
I understand this PR, but personally I prefer config Nginx to serve /livewire.js as static file instead manual update assets in each version. 

A recent note about it was added to docs #7083

soundslocke:
The installation documentation update (new recommendation?!) within PR #7083 does not lead to nginx serving the path as a static file. It only _looks_ like a static file.

It's no different than before: still running that route through PHP, which is not ideal or best practice in production. This invokes FastCGI, the PHP runtime, Composer autoloading, Laravel kernel and facilities, and subsequently the Livewire facilities, to simply to serve static JavaScript.

robsontenorio:
I don’t think it is a recommendation. Nothing has changed , juste a note how to proper solve nginx error.

christoferd:
I got the same error while not using nginx, only apache on windows.
`Uncaught TypeError: Cannot read properties of undefined (reading 'uri')` at livewire.esm.js:7265:124

Latest version of livewire at 4-Dec-2023

gbrits:
At the moment with the latest Livewire & WireUI I'm unable to control when AlpineJS is loaded and therefore I need this... so I've just required this PR on my project and published the assets and its just given me the livewire.js file, minified mind you. Is my only option to fish through the minified JS file & remove what I believe to be all traces of AlpineJS & then I... somehow using the config/livewire `asset_url` direct it to this new javascript file?

soundslocke:
@gbrits To me that sounds like you may want to dig into [manually bundling Livewire and Alpine](https://livewire.laravel.com/docs/installation#manually-bundling-livewire-and-alpine) in a custom app, in whatever way(s) desired. Then you could control exactly how you wish that custom app's assets to be loaded.

joshhanley:
@soundslocke thanks for the PR! I pulled this down to test it, but Livewire wasn't using the published version of `livewire.js` as such this PR doesn't work currently.

I'm going to close this for now, but please reopen a new PR if you manage to get it all working. Thanks!

soundslocke:
>Livewire wasn't using the published version of `livewire.js` as such this PR doesn't work currently.

@joshhanley I'm not sure what you mean by this. This PR provides asset publishing and the implication is the same as how it worked in prior versions of Livewire.

The docs describe [customization of the asset URL](https://livewire.laravel.com/docs/installation#customizing-the-asset-url), which is how you would use the published assets rather than the dynamic Livewire asset route. This is seemingly true whether you manually copy the assets yourself, copy the assets via Composer hook, or copy the assets via Artisan publish command, which this PR puts back into place akin to the prior two versions of Livewire.

Is there something else expected here? As far as our use case (to have "real" static assets for production) goes, this is seemingly all that is desired.

joshhanley:
@soundslocke try pulling this PR into a test app, publish the assets, make a change in the published assets and then load the app in a browser. You will find that it’s not using the static published assets like it should, and instead is still pulling from Livewire’s dist directory.

To me, livewire.js should use the public published version if it’s available, but it’s not currently. That’s how it worked in v2. You shouldn’t need to use a custom asset route if publishing the assets is set up correctly.

soundslocke:
I have tested this, which is how I crafted the admittedly minor change here. I'm guessing a difference in paths and/or nginx configuration (which I'd assume will often be bespoke when wanting to publish assets for production like this) may be the hold up?

@joshhanley I do really appreciate you checking this out. It seems you're describing behavior beyond what I'm aware of and perhaps beyond the need we seem to have, at least. [The prior docs appear to describe](https://laravel-livewire.com/docs/2.x/installation#configuring-the-asset-base-url) what we previously had been doing and what we continue to do, which is to publish the assets and use a new/different asset location. That first example in the old docs:

>You publish the Livewire assets and are now serving them from a sub-folder like "assets".

We configured the asset URL when publishing assets via the Artisan command in v2 as well, otherwise we were unable to use the static assets. I don't think we'll ever be changing the provided assets as you mention. We'd likely go with manual bundling if we ever had to adjust something in the core?

For others who might desire "real" static production assets here's what we are doing in `composer.json` (note your paths may be different):

```
...

    "scripts": {
        "livewire-publish": [
            "cp -r vendor/livewire/livewire/dist/* ../public/vendor/livewire/"
        ],
        "post-autoload-dump": [
            "@php -r \"if (!getenv('CI')) { shell_exec('composer run livewire-publish'); }\""
        ],

...
```

And then this, within `AppServiceProvider::boot()` (again, your paths may be different):

```php
if (!app()->runningInConsole()) {
		return Route::get('/vendor/livewire/livewire.js', $handle);
	});
}
```

joshhanley:
@soundslocke have submitted a new PR #7973 with the changes I was talking about.

The first part of publish the files to the public directory you had correct.

But the second part in your example above, where you are using a custom `setScriptRoute()`, Livewire should automatically do that for you. Which is what I have done the new PR and how it worked in V2 🙂

soundslocke:
Thank you for working on it @joshhanley, I hope it's officially supported again soon.

PR #6897, opened 2023-09-22
Adding temporarySafeUrl() and hasTemporaryUrl() for safe usage of temporaryUrl()

michaelkonstantinou:

_Yes, there have been several discussion already:_

Explicit discussions: 

- https://github.com/livewire/livewire/discussions/5056

Other relevant discussions

- https://github.com/livewire/livewire/discussions/6578
- https://github.com/livewire/livewire/discussions/5618

_Yes_

_No, only the two new functions_

_Yes: Unit tests + Browser tests_

Thanks for contributing! 🙌

### Problem and motivation

When using image uploads, livewire provides the ability of previewing the image. As stated in the documentation, this is the _"correct"_ way of doing it
```php
@if ($photo) 
        <img src="{{ $photo->temporaryUrl() }}">
 @endif
```

In the relevant discussions, you will find similar discussions on how other developers tried to avoid issues with temporaryUrl. However, this approach is **dangerous**

1. temporaryUrl() throws an exception. Since this function is mostly used in the front-end, it is very dangerous as it might throw an exception to the user.
2. Does not provide a way to catch the exception when in the front-end. One could either use a bunch of if-statements (ugly code) or add further if-statements in the backend to catch the error
3. The example on the documentation and in the relevant discussions is not safe. The example in the documentation can still fail if the `$photo` is a `TemporaryUploadFile` which does not support the temporaryUrl. 
4. In my case, the function threw an exception when the file was already uploaded. Although a bunch of if-statements can fix the issue, I believe **this PR will provide a cleaner way of handling such issues.**
5. This issue becomes worse when other image libraries are being used.

### Solution

The solution in this PR, **does not affect any pre-existing code**. It only adds two new functions. The PR consists of the following

1. temporarySafeUrl() function

- This catches the exception before thrown. If the temporaryUrl fails, the method will return null (additional configuration can be added to have a fallback string, but not included at this PR. If you agree, we can add it in another PR )

2. hasTemporaryUrl() function

- As already asked in a discussion (view above), some developers (including myself) would benefit from a hasTemporaryUrl function instead of adding a bunch of ugly if-statements to make sure that the dangerous code will not be executed. I kept it simple, to check only if temporarySafeUrl returns a valid string. 

### New possibilities after merging this PR

To begin with, the example in the documentation and anything similar, can be changed to the following:
```php
@if ($photo?->hasTemporaryUrl()) 
        <img src="{{ $photo->temporaryUrl() }}">
@endif
```

At the same time, one can use the following
```php
 <img src="{{ $photo?->temporarySafeUrl() }}">
```

### DIFF
--- a/src/Features/SupportFileUploads/BrowserTest.php
+++ b/src/Features/SupportFileUploads/BrowserTest.php
@@ -61,4 +61,90 @@ class BrowserTest extends \Tests\BrowserTestCase
         })
         ;
     }
+
+    /** @test */
+    public function can_check_temporary_url_before_rendering()
+    {
+        Storage::persistentFake('tmp-for-tests');
+
+        Livewire::visit(new class extends Component {
+            use WithFileUploads;
+
+            public $photo;
+
+            function mount()
+            {
+                Storage::disk('tmp-for-tests')->deleteDirectory('photos');
+            }
+
+            function save()
+            {
+                $this->photo->storeAs('photos', 'photo.png');
+
+                // Break temporaryUrl
+                $this->photo = new TemporaryUploadedFile('', '');
+            }
+
+            function render() { return <<<'HTML'
+            <div>
+                <input type="file" wire:model="photo" dusk="upload">
+
+                <div wire:loading wire:target="photo">uploading...</div>
+
+                <button wire:click="$refresh">refresh</button>
+
+                <div>
+                    @if ($photo?->hasTemporaryUrl())
+                        <img src="{{ $photo->temporarySafeUrl() }}" dusk="preview">
+                    @else
+                        <p dusk="nopreview">Preview not available</p>
+                    @endif
+                </div>
+
+                <button wire:click="save" dusk="save">Save</button>
+            </div>
+            HTML; }
+        })
+            ->assertMissing('@preview')
+            ->attach('@upload', __DIR__ . '/browser_test_image.png')
+            ->pause(250)
+            ->assertVisible('@preview')
+            ->tap(function () {
+                Storage::disk('tmp-for-tests')->assertMissing('photos/photo.png');
+            })
+            ->waitForLivewire()
+            ->click('@save')
+            ->tap(function () {
+                Storage::disk('tmp-for-tests')->assertExists('photos/photo.png');
+            })
+            ->assertVisible('@nopreview')
+            ->assertMissing('@preview')
+        ;
+    }
+
+    /** @test */
+    public function can_safely_call_temporary_url()
+    {
+        Storage::persistentFake('tmp-for-tests');
+
+        Livewire::visit(new class extends Component {
+            use WithFileUploads;
+
+            public $photo;
+
+            function mount()
+            {
+                // Break temporaryUrl
+                $this->photo = new TemporaryUploadedFile('', '');
+            }
+
+            function render() { return <<<'HTML'
+            <div>
+                <p>Link: {{ $photo->temporarySafeUrl() ?? 'Not found' }}</p>
+            </div>
+            HTML; }
+        })
+            ->assertSee('Link: Not found');
+        ;
+    }
 }
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -7,6 +7,7 @@ use Illuminate\Http\UploadedFile;
 use Illuminate\Support\Facades\URL;
 use Illuminate\Support\Facades\Storage;
 use League\MimeTypeDetection\FinfoMimeTypeDetector;
+use Throwable;
 
 class TemporaryUploadedFile extends UploadedFile
 {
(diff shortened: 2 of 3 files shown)


levu42:
@Mikhail-Konstantinou I used temporary URLs for upload preview many times so far without any problems, never any exceptions thrown - do you have an example of how it can fail, e.g. a failing test where an exception is thrown?

Mikhail-Konstantinou:
Hi @levu42 

That's actually the problem with temporaryUrl. You might be using it for a whole year (e.g. me), and still not fail but it will never be safe to call temporaryUrl on the front-end as it might throw an exception. I added also some discussions of other developers who had to find a way to bypass the temporaryUrl function. 

Regarding your question about a failure. Personally, to me it failed when I used Livewire 3 forms alongside spatie media library. At some point I tried to add a form where I could upload multiple images and use spatie's media library conversions. While the media library was creating additional images for me, I ended up with the following situation in the front-end

- A valid TemporaryUploadFile
- No temporaryUrl however

Obviously, you can add multiple if-statements to verify that the code will not be executed, queue or non-queue the library's operations... but it is still ugly and not 100% safe. I believe with this function you can only benefit from as it doesn't change the current temporaryUrl function. It only adds a safer way of calling it.

Regarding failing tests, I added a few tests of a valid TemporaryUploadFile that will still throw an exception of the temporaryUrl function.

levu42:
@Mikhail-Konstantinou Your tests include a `new TemporaryUploadedFile('', '')` which doesn't look like code you would expect to work. How do you get to this broken state where `temporaryUrl` doesn't work, when you don't try to manually create a broken `TemporaryUploadedFile` object?

Mikhail-Konstantinou:
Sorry, I am having a hard time to reproduce the exact error as it was a runtime one, but I can reproduce (or mention) other similar cases. Basically, everything comes to the `TemporaryUploadedFile::$path` and the `TemporaryUploadedFile::isPreviewable` function

First of all, the "inconsistency" in the storage's temporaryUrl and livewire's way of checking whether it can call it or not

**FilesystemAdapter::providesTemporaryUrls**
```php
   /**
   * Determine if temporary URLs can be generated.
   *
   * @return bool
   */
  public function providesTemporaryUrls()
  {
      return method_exists($this->adapter, 'getTemporaryUrl') || isset($this->temporaryUrlCallback);
  }
```

**TemporaryUploadedFile::temporaryUrl**
```php
if ((FileUploadConfiguration::isUsingS3() or FileUploadConfiguration::isUsingGCS()) && ! app()->runningUnitTests()) {
            // returns something that does not break
        }

        if (method_exists($this->storage->getAdapter(), 'getTemporaryUrl') || ! $this->isPreviewable()) {
            // This will throw an error because it's not used with S3. **But haven't we checked this above?**
            return $this->storage->temporaryUrl($this->path, now()->addDay());
        }
```

- FilesystemAdapter has the function `providesTemporaryUrls` to check whether a temporaryUrl can be generated
- TemporaryUploadedFile calls the function temporaryUrl in method TemporaryUploadedFile::temporaryUrl() after it checks whether the driver is S3 or GCS, and the following returns true: `method_exists($this->storage->getAdapter(), 'getTemporaryUrl') || ! $this->isPreviewable()`
- Which means that while the function FilesystemAdapter::providesTemporaryUrls might return false.... TemporaryUploadedFile still calls the function temporaryUrl

So in practice, this means the following:
- If you do not use S3 or GCS and you have a **valid** **path**, `TemporaryUploadedFile::temporaryUrl` will use `URL::temporarySignedRoute` to create a `temporaryUrl`
- If for any reason the path is changed and is **no longer previewable**... then the **same** `TemporaryUploadedFile` object will use `FilesystemAdapter::temporaryUrl` to create a `temporaryUrl`... which will crash

Usually this happens when a file is uploaded. Once the file is uploaded the path is changed. The solution I do and I see other people doing, is to immediately make the variable `$photo` null. Which still, doesn't mean that this is ideal but it works. In one of the examples I mentioned above the function `is_object` is used.

But this is not the only way to get an exception. If for any reason the path is not recognised as previewable, maybe because a third-party image library crashed, or changed something in the path, then you will end up with a TemporaryUploadedFile that throws an exception. 

TLDR; there is a difference between `FilesystemAdapter::providesTemporaryUrls` and how `TemporaryUploadedFile::temporaryUrl` calls the code `FilesystemAdapter::temporaryUrl` which leaves a small gap to throw an exception. There could be different solutions to this issue, but I added a solution that does not interfere with the current codebase. If you would like, I could make some changes in the TemporaryUploadedFile to return something if `FilesystemAdapter::providesTemporaryUrls` yields false. However, I don't think that `TemporaryUploadedFile::temporaryUrl` is "safe" to call since the exception is not caught

joshhanley:
@Mikhail-Konstantinou thanks for the PR! We're going to hold off on this for now as we think there could be a better solution for this. You could add this method to your own components as `TemporaryUploadedFile` is macroable. Thanks for being willing to contribute!

PR #6925, opened 2023-09-25
A locked property prevents updates to similar named properties

PhiloNL:
Failing test for: https://github.com/livewire/livewire/discussions/6917

```php
new class extends Component {
    #[BaseLocked]
    public $count = 1;

    public $count2 = 1;

    public function render() {
        return '<div></div>';
    }
}
```

When you try to set `$count2` it will fail and throw an exception:

```
Exception: Cannot update locked property: [count]
```

Might be a PHP bug?

### DIFF
--- a/src/Features/SupportLockedProperties/UnitTest.php
+++ b/src/Features/SupportLockedProperties/UnitTest.php
@@ -48,4 +48,21 @@ class UnitTest extends \Tests\TestCase
         ->assertSet('foo.count', 1)
         ->set('foo.count', 2);
     }
+
+    /** @test */
+    function can_update_locked_property_with_similar_name()
+    {
+        Livewire::test(new class extends Component {
+            #[BaseLocked]
+            public $count = 1;
+
+            public $count2 = 1;
+
+            public function render() {
+                return '<div></div>';
+            }
+        })
+        ->assertSet('count2', 1)
+        ->set('count2', 2);
+    }
 }


lrljoe:
Not a PHP issue

Lives in Features/SupportAttributes/SupportAttributes.php

The update function looks for a "StartsWith":
```
            ->filter(fn ($attr) => str($fullPath)->startsWith($attr->getName()))
```

Current Code
```
    function update($propertyName, $fullPath, $newValue)
    {
        $callbacks = $this->component
            ->getAttributes()
            ->whereInstanceOf(LivewireAttribute::class)
            ->filter(fn ($attr) => $attr->getLevel() === AttributeLevel::PROPERTY)
            // Call "update" on the root property attribute even if it's a deep update...
            ->filter(fn ($attr) => str($fullPath)->startsWith($attr->getName()))
            ->map(function ($attribute) use ($fullPath, $newValue) {
                if (method_exists($attribute, 'update')) {
                    return $attribute->update($fullPath, $newValue);
                }
            });

        return function (...$params) use ($callbacks) {
            foreach ($callbacks as $callback) {
                if (is_callable($callback)) $callback(...$params);
            }
        };
    }

```

Could replace with the "is" function, but not sure if this method is reused elsewhere (thus why startsWith() is currently being used)

```
    function update($propertyName, $fullPath, $newValue)
    {
        $callbacks = $this->component
            ->getAttributes()
            ->whereInstanceOf(LivewireAttribute::class)
            ->filter(fn ($attr) => $attr->getLevel() === AttributeLevel::PROPERTY)
            // Call "update" on the root property attribute even if it's a deep update...
            ->filter(fn ($attr) => str($fullPath)->is($attr->getName()))
            ->map(function ($attribute) use ($fullPath, $newValue) {
                if (method_exists($attribute, 'update')) {
                    return $attribute->update($fullPath, $newValue);
                }
            });

        return function (...$params) use ($callbacks) {
            foreach ($callbacks as $callback) {
                if (is_callable($callback)) $callback(...$params);
            }
        };
    }
```

PhiloNL:
Good catch! Thanks for adding the fix! 😄

lrljoe:
No worries!  Good catch for the bug, and thanks for all the hard work on the WE components. love and highly recommend the modal

PhiloNL:
> No worries! Good catch for the bug, and thanks for all the hard work on the WE components. love and highly recommend the modal

Thanks! Appreciate the kind words 🙏🏼

calebporzio:
Thanks for this, the fix has been merged

PR #6926, opened 2023-09-26
Fix: A locked property prevents updates to properties that start with the same string

lrljoe:

Failing tests are available in PR from another contributor:
[https://github.com/livewire/livewire/pull/6925](https://github.com/livewire/livewire/pull/6925)

Thanks for contributing! 🙌

---

This fix relates to the failing test available here:
[https://github.com/livewire/livewire/pull/6925](https://github.com/livewire/livewire/pull/6925)

Which relates to the discussion available here:
[https://github.com/livewire/livewire/discussions/6917](https://github.com/livewire/livewire/discussions/6917)

---

Presently, the behaviour of Attribute matching, matches based on "startsWith", however - this will match to other variables available on the component.

Example of issue:
```
#[Locked]
public $query = 'abcd';

public $query123 = 'abcdefg';
```
In this scenario, both $query123 and $query will be locked.  Any error returned will relate only to $query.

The present work-around for this, is to ensure entirely distinct names for each variable on a component, e.g.

```
#[Locked]
public $query = 'abcd';

public $notQuery = 'abcdefg';
```
This results in the expected behaviour of $query being locked, and $notQuery not being locked in the frontend.

The PR seeks to address this by using an "is" approach rather than startsWith to match against variables.

**Minimal testing has been performed!**

### DIFF
--- a/src/Features/SupportAttributes/SupportAttributes.php
+++ b/src/Features/SupportAttributes/SupportAttributes.php
@@ -51,7 +51,7 @@ class SupportAttributes extends ComponentHook
             ->whereInstanceOf(LivewireAttribute::class)
             ->filter(fn ($attr) => $attr->getLevel() === AttributeLevel::PROPERTY)
             // Call "update" on the root property attribute even if it's a deep update...
-            ->filter(fn ($attr) => str($fullPath)->startsWith($attr->getName()))
+            ->filter(fn ($attr) => str($fullPath)->is($attr->getName()))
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for this @PhiloNL and @lrljoe!!!

lrljoe:
@calebporzio - no problem at all, still getting used to the coding style by you guys, but can see what you tweaked, and understand.
@joshhanley has been a great help with pointers etc.

PR #6936, opened 2023-09-28
Update nesting.md

bradenkeith:
Needle and the haystack were reversed; this resolves that.


calebporzio:
Lol 🤦 , thanks Braden! (Hi! btw)

PR #6937, opened 2023-09-28
Update nesting.md

bradenkeith:
To actually change the steps in a dynamic component, we need a key on the component.


joshhanley:
@bradenkeith thanks for the PR! We're wondering if maybe this is something that Livewire should handle internally, so you don't need to add a key manually? Do you think you could PR that change or at least a failing test PR so we can have a look at it? Thanks!

PR #6943, opened 2023-09-29
Add scroll-to-top behavior for pagination links

calebporzio:
Currently Livewire doesn't scroll to the top after each pagination link visit.

This PR adds "scroll-to-top" behavior as the new default and also provides an API to opt-out or pass a CSS selector instead.

Reference the docs contained in this PR for reference.


patriktoth67:
I'd like to add 2 ideas to this:

- It doesn't work with custom pagination views
- The behavior could be changeable, to make it smooth scrolling instead of auto

stefaanbeernaert:
should this not be default false then ? Since in livewire 2 this is not the normal behavior?

because now i have to put to all my links 
 {{ $blogs->links(data: ['scrollTo' => false]) }}

PR #6974, opened 2023-10-02
Use Internet Explorer conditional comment syntax to preserve morph markers on Cloudflare-proxied sites

markjaquith:
I saw [this tweet](https://x.com/jasonlbeggs/status/1708938021255405711?s=20) from @jasonlbeggs about how Cloudflare's "Auto Minify" option strips HTML comments and thus remove the morph markers that Livewire 3 inserts, breaking that feature.

CleanShot 2023-10-02 at 18 38 54@2x

I know a way to fix that!

## It's story time

In 1999, Microsoft released Internet Explorer 5. At this time, browser engines were really flaky and didn't have good backwards or forward compatibility. For instance, things that worked in Internet Explorer 5 would break horribly in Internet Explorer 3. Microsoft wanted a way for people to target markup for certain versions of their browser engine. So they introduced a feature called [Conditional Comments](https://en.wikipedia.org/wiki/Conditional_comment)

This feature allowed developers to write HTML comments with a special syntax that Internet Explorer could interpret. This allowed for targeting markup to certain browser engines or versions. 

This would only load the `ie8.js` script on versions of Internet explorer less than 9, for example:

```html
<!--[if lt IE 9]>
<script src="./ie8.js"></script>
<![endif]-->
```

Engines other than Internet Explorer just see an HTML comment and ignore it.

The fact that these comments produce actual changes in website behavior (when using Internet Explorer) meant that when Cloudflare wrote their "Auto Minify" feature, they knew they had to preserve these special comments.

## The fix

Thus, if we use this syntax, Cloudflare will not remove the morph markers:

`<!--[if gte livewire 3]> __BLOCK__ <![endif]-->`

`<!--[if gte livewire 3]> __ENDBLOCK__ <![endif]-->`

This PR implements that syntax. The fact that it says `livewire` there instead of using `ie lt 3` or some other condition that will always evaluate as false was just for fun, and because it indicates that these comments are here for Livewire, not Internet Explorer.


joshhanley:
@markjaquith I thought we already had this in V2. Turns out it never got merged as the PR was doing to much. PR for reference https://github.com/livewire/livewire/pull/4153

So I think it should be good 😁

The reason tests are failing will be due to an out of date version of Alpine. Looks like Caleb just tagged a new release with all the teleport changes. So if you npm update and rebuild, it should all work.

RoccoHoward:
> I thought we already had this in V2. Turns out it never got merged as the PR was doing to much. PR for reference #4153

Could we backport this change V2?

levu42:
I agree with @danharrin, it adds unnecessarily quite a bit of overhead to the returned HTML, when the problem is not inside Livewire, but in 3rd party tools, where you even have an option to turn off this behavior.

aarondfrancis:
> It would increase the size of the DOM quite dramatically

@danharrin Would it? "Quite dramatically" seems to overstate it a bit. The comments are already there, it's just adding a few more characters. `livewire` could be shortened to `lw` but I don't think it matters at all!

PhiloNL:
Maybe only do this when running in Cloudflare? I think Cloudflare always adds an CF header when they proxy a request.

danharrin:
@aarondfrancis In applications with complex interfaces, large numbers of conditions are absolutely inevitable:

Screenshot 2023-10-03 at 16 59 54

So 1174 * 2 (__BLOCK__ and __ENDBLOCK__) * 29 (extra characters) = 68092 extra characters

Which is roughly 68kb, right?

Remember that these comments are often spat out when there is absolutely no content within them - the content is conditional, and currently hidden.

ericlbarnes:
If you use HTML minification, aren't you saving more KBs than these extra conditions would add? So, without the checks, you are forced to send more characters through the HTML than you are saving.

danharrin:
@ericlbarnes That could be the case sometimes, but many people aren't using Cloudflare, and even more aren't using minification. Philo's idea sounds good, but it does end up even impacting those using Cloudflare without minification.

I definitely have experimented with minification before, but it caused quite a few other issues elsewhere and I had to disable it again.

IMO this should be opt-in, probably a config option to remap the content of the conditional comments.

nuernbergerA:
i think if your webserver uses brotli or gzip (which all should do) it shouldn't effect the real payload  because its a repeated string and compression loves that

@danharrin can you test your big dom example with this https://www.multiutil.com/text-to-brotli-compress/

markjaquith:
> @aarondfrancis In applications with complex interfaces, large numbers of conditions are absolutely inevitable:
> 
> Screenshot 2023-10-03 at 16 59 54
> So 1174 * 2 (`__BLOCK__` and `__ENDBLOCK__`) * 29 (extra characters) = 68092 extra characters
> 
> Which is roughly 68kb, right?
> 
> Remember that these comments are often spat out when there is absolutely no HTML content within them - the content can be conditional, and currently hidden.

@danharrin I loaded a page up with 1174 pairs of my proposed marker (plus `<!doctype html>`) The total size of the page was 783 bytes.

With the current markers, the total size of the page was 695 bytes.

**An additional 88 bytes for 1174 marker pairs**. I knew compression on repeated text was efficient but I didn't know it was THIS good! 😁

If visual clutter in the source is a concern, we could also do:

`<!--[if BLOCK]><![endif]-->`
`<!--[if ENDBLOCK]><![endif]-->`

That brings the size of the above page down to 757 bytes.

markjaquith:
> Also I noticed you committed the dist js changes but not the source? Can you ensure that is added? Thanks!

@joshhanley Ah, it looks like that JS code [is coming from Alpine](https://github.com/alpinejs/alpine/blob/590250dbc4a88a449e84000da923d0e89019ba3c/packages/morph/src/morph.js#L158-L160), so there would have to be an equivalent change over there. Happy to help with that if this change (or some version of it) gets the green light.

calebporzio:
@markjaquith - holy crap, you're a legend!

Thanks for the input everyone.

What about this:
- By default Livewire sends normal comments over the wire
- If Livewire detects a cloudflare header present, it sends: `<!--[if __BLOCK__]><![endif]-->` (this way we don't have to make changes on the Alpine end, only the Livewire end)

What do we think?

markjaquith:
@calebporzio Unfortunately I think you will still have to make changes on the Alpine end... it's looking for an exact match to ` __BLOCK__ ` (even with surrounding spaces).

https://github.com/alpinejs/alpine/blob/590250dbc4a88a449e84000da923d0e89019ba3c/packages/morph/src/morph.js#L158-L160

But I don't think this is a big deal... I doubt anyone else but Livewire 3 is using these morph markers. Changing them at this point seems fine.

As for making their output conditional on detecting Cloudflare... to what benefit? There are likely other HTML-minification pipelines that remove comments (but know to leave Internet Explorer conditionals alone). And there might be a situation where the resultant HTML is going out to Cloudflare, but is being double-proxied, so you won't be able to dynamically detect it in the PHP server if that header doesn't survive. Think like double-nginx situations or nginx/Varnish/nginx/PHP layer cakes.

These aren't significantly longer than what you have now (and basically not at all, with compression), and should skate through proxies without a problem.

`<!--[if BLOCK]><![endif]-->`
`<!--[if ENDBLOCK]><![endif]-->`

There's no extra Cloudflare detection code to maintain or run... it will likely work with non-Cloudflare proxies, it's just a slightly different placeholder that is much more likely to survive through to the browser.

joshhanley:
@markjaquith just to confirm, how did you make the javascript changes? Did you manually overwrite the dist files? If so, the next time we build those changes will be wiped out. So we will need to PR the changes to Alpine first and then build Livewire's assets from there. Let me know if you need a hand with that 🙂

Yeah I like the shorter changes you just made, they're basically the same length. I think that's better than trying to detect cloudflare (and any other pipeline).

markjaquith:
> And yeah in terms of green light, yeah we're happy with what this PR does.

Sweet. I'll get a PR in to Alpine.

And yeah, I was just editing the `dist` files for now because it seemed like the easiest way to showcase the change and have the Livewire unit tests pass and let other people try it out. Yes it felt gross. The DX for handling the coupling between Alpine and Livewire when making PRs is not great. Not sure how much better it could get though, given how tightly they're coupled.

joshhanley:
@markjaquith awesome, thanks!

Yep agreed, does make it harder. It's why history and navigate Alpine packages were folded into the Livewire codebase for now, while we keep working on them. Not sure how to make upstream changes easier.

calebporzio:
Agree, by all accounts. Merged the Alpine changes and re-built.

Thanks a bunch again for this clever little fix - not sure if I ever would have thought of this

PR #6976, opened 2023-10-03
Fix typo

woodspire:
The current code *might* loop because the external OpenAI API is called ChatBot and the Livewire class is also called that.

The external API "ask" function is called statically, so I am not sure if it would loop or simply throw an exception.

With this change, there is no confusion.

And OpenAI is referenced in this page so using a class with this name will simply further detail that it's an external system.


calebporzio:
You're right. Thanks!

PR #6979, opened 2023-10-03
Update wire-transition.md

valpuia:
Following https://github.com/livewire/livewire/discussions/6978

I am not sure the sentence are correct or not but I think it's better to display this way instead of current (see below attachment)
Screenshot from 2023-10-03 12-30-17


joshhanley:
@valpuia thanks for the PR! This has been fixed in PR #6972 so going to close this. Thanks for being willing to contribute! 🙂

PR #6981, opened 2023-10-03
Fix documentation about attributes method

beholdr:
Method attributes() should be called validationAttributes()


joshhanley:
@beholdr thanks!

PR #6990, opened 2023-10-04

apreiml:
Currently preview file download relies on the file being stored on a local disk.

This commit introduces a new function that fetches the file and metadata directly from storage to enable non-local filesystems.


joshhanley:
@apreiml thanks for the PR! Any chance you could add a test for this? Or if that is too difficult, then could you create an example repo that demonstrates how this is being used? Thanks!

apreiml:
@joshhanley I've added a test and did a rebase on top of main.

calebporzio:
Thanks!

PR #7002, opened 2023-10-05
Fix missing semicolon inside `messages()` method in `validation.md`

becker:
Just adding the missing semicolon to the return array in the `messages()` method example, in case someone is copying and pasting like me. :sweat_smile:


joshhanley:
@becker thanks!

PR #7003, opened 2023-10-05
Fix eloquent models and collections

joshhanley:
This PR fixes issues with Eloquent Models and Collections when are assigned to public properties and then are deleted, throwing 404s and causing n+1 problems. See #6120, #6297, #6302, #6457, #6511, #6792, #6964.

As per a discussion with Caleb, we have decided to:
- Allow empty models to be assigned (gets rid of `Can't set model as property if it hasn't been persisted yet` error)
- If a model has been assigned to a property and then it has been deleted, then we just assign an empty model of the same class to the property (this stops a 404 issue) - we decided this was better than setting to `null`
- Add an Eloquent Collection synth that ensures Eloquent Collections are correctly rehydrated using a `whereIn()` query (this also stops a different 404 issue and also ensures that restoring collections don't cause an n+1 problem) 

I've merged the test from PR #6858 into this PR, so will close that one and #6272.

Hope this helps!

Edit: I have also added a warning box to the properties page regarding custom queries and Eloquent Collection and Model properties, under where Eloquent Collections and Models are listed for the first time, and added a link to the section at the bottom of the page which explains in more detail.

**I wonder if we should also mention in that warning box that any data stored on the Model/Collection that hasn't been persisted won't be preserved between requests?**

image

### DIFF
--- a/src/Features/SupportModels/EloquentCollectionSynth.php
+++ b/src/Features/SupportModels/EloquentCollectionSynth.php
@@ -64,6 +64,9 @@ class EloquentCollectionSynth extends Synth {
             return new $class();
         }
 
+        // We are using Laravel's method here for restoring the collection, which ensures
+        // that all models in the collection are restored in one query, preventing n+1
+        // issues and also only restores models that exist.
         $collection = (new $modelClass)->newQueryForRestoration($keys)->useWritePdo()->get();
 
         return $collection;
--- a/src/Features/SupportModels/EloquentCollectionSynth.php
+++ b/src/Features/SupportModels/EloquentCollectionSynth.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Livewire\Features\SupportModels;
+
+use Illuminate\Database\Eloquent\Collection as EloquentCollection;
+use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Queue\SerializesAndRestoresModelIdentifiers;
+use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
+
+class EloquentCollectionSynth extends Synth {
+    use SerializesAndRestoresModelIdentifiers;
+
+    public static $key = 'elcln';
+
+    static function match($target)
+    {
+        return $target instanceof EloquentCollection;
+    }
+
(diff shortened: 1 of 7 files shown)


calebporzio:
Legend @joshhanley 👊

src/Features/SupportModels/EloquentCollectionSynth.php
```diff
@@ -0,0 +1,86 @@
...
+
+    static function match($target)
+    {
+        return $target instanceof EloquentCollection;
+    }
+
+    function dehydrate(EloquentCollection $target, $dehydrateChild)
+    {
+        $class = $target::class;
+        $modelClass = $target->getQueueableClass();
```
QuentinGab, line 23: `getQueueableClass` return null if the collection is empty. 
Then it throws error when calling`$modelAlias = (new $modelClass)->getMorphClass();`.

I'm not comfortable myself with Synths, maybe @joshhanley you have an idea on how to fix this ?

QuentinGab: @calebporzio I always wonder if you would prefer to discuss newly introduced bug in the MR or in a new Discussion ?

QuentinGab: I'm working on a fix myself

QuentinGab: I submitted a PR #7017, tests are passing on my machine.

joshhanley: For anyone else seeing this, fix has been released in v3.0.7.

MarcoCazzaro:
Suggestion: put a giant banner in the [Upgrade Guide](https://livewire.laravel.com/docs/upgrading), please. I think it's a massive change, that usually run completely undetected. E.g. if like me the order in the database of my models is the same of their "display_order" field which I use in my orderBy method, everything _looks_ OK, but as soon as you change the order in the fields everything messes up and you have absolutely no clue.

joshhanley:
@MarcoCazzaro we've just fixed that issue in PR #7359. So should be available in the next release.

PR #7008, opened 2023-10-06
Simple togglable query string

nicolagianelli:
#6494 
Yes
No
Yes

As we discussed in the PR #6494 the implementation was changing too many files so I reimplemented it similar to V2 where you can disable query string by returning NULL, 0 or false from queryString() method.
Now it does change only 3 files and few rows, I think this is the best way to let people disabling query string if needed.
The docs are also updated with this info.
With this change, you could use it like said in this PR: #6799 (by returning null/false or 0 instead of an empty array)


edo2313:
Any updates on this? It's not really a new feature, more like a fix to make it work similarly to v2... @joshhanley

calebporzio:
Yeah thanks for this, I feel like the solution is to provide better control on the PHP for query string values. Like a way to register a sort of middleware or something to control serialization and deserialization, but I can't think of a good one off hand.

I hesitate to merge this as it doesn't feel very ergonomic to me and doesn't use #[Url] but instead `->queryString()` returning null.

So I know none of that is very helpful but I'm just giving context into why I'm not just merging this and moving on.

Thanks for your efforts. Going to leave this open for now

calebporzio:
I ended up going with a trait called `WithoutUrlPagination` to solve the exact problem you're facing. Thanks!

edo2313:
> I ended up going with a trait called `WithoutUrlPagination` to solve the exact problem you're facing. Thanks!

While the trait is a great addition, it's not really what was proposed here as it's not dynamically configurable... The PR was about having a way to toggle the querystring "at runtime", by passing some kind of variable to the component (like in V2).

PR #7013, opened 2023-10-07
Add `->extends()` section to `components.md`

amshehzad:
In Livewire v2 there was a section on using traditional Blade layout files using `@extends` which is missing in v3 docs.
This PR added that section to the docs.


calebporzio:
Perfect, thank you for adding this!

PR #7015, opened 2023-10-07
Fix comma characters in query strings

nullthoughts:
Needed, as Livewire V3 is currently encoding commas in the query string and the expected behavior is for comma characters to not be encoded.

For reference, please see the Vue Router [discussion on this behavior](https://github.com/vuejs/vue-router/pull/1047#issuecomment-274623717) and [their solution](https://github.com/vuejs/vue-router/commit/3f91065e5165d0457ee6295980086b86e107d034).

Yes

No

Yes

This is needed as a bug fix, but also to support array compacting in the query string as documented in multiple discussions and PRs over the years. @calebporzio has previously recommended utilizing hooks and separate properties to achieve this (to avoid adding a custom API): https://github.com/livewire/livewire/pull/3679#issuecomment-989218652

The change simply adds one line of code (`.replaceAll('%2C', ',')`) to js/plugins/history/index.js, and an accompanying test.

Currently, Livewire result's in a query string of`?condition=new%2Cused`, when `?condition=new,used` is expected for the following:
```php
#[Url]
public $condition;

public function render()
{
    $this->condition = 'new,used';
}
...
```


joshhanley:
@nullthoughts thanks for the PR! Assets are currently being built from Alpine's main branch, so thats probably why there were test failures. I have rebuilt the assets and rerun the tests.

calebporzio:
Perfect, thank you!

PR #7042, opened 2023-10-10
fix: Stream cuts chunked JSON objects in half

danharrin:

No

Yes

No

No streaming tests at all yet, I don't have the time to dedicate to creating those yet

Sometimes, `reader.read()` returns an incomplete JSON object - half of it is stuck in the next chunk of the response. Incomplete JSON objects cannot be identified as stream chunks, and therefore do not get extracted, and end up in the `finalResponse`. When they remain in the `finalResponse`, they corrupt the JSON structure of the real final response when it gets parsed. These need to be removed from the final response at all costs otherwise the streamed request cannot complete.

This PR changes the logic, so instead of discarding incomplete JSON chunks as "finalResponse", it appends them onto the start of the next response. This allows Livewire to reattempt parsing them as a complete chunk, meaning that they don't end up in the final response when it is parsed.

Thanks for contributing! 🙌

### DIFF
--- a/dist/livewire.esm.js
+++ b/dist/livewire.esm.js
@@ -8344,18 +8344,18 @@ on("request", ({ respond }) => {
 });
 async function interceptStreamAndReturnFinalResponse(response, callback) {
   let reader = response.body.getReader();
-  let finalResponse = "";
+  let remainingResponse = "";
   while (true) {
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks Dan!

PR #7057, opened 2023-10-11
Add failing test when updating form data containing a locked property (regression introduced in v3.0.6).

ildyria:
I'm submitting a failing test for a scenario with form properties involved.

Assuming a component with a form.
We may want to lock one of the property of the form. Unfortunately this locks all the properties of said form.
**This is a regression introduced by version 3.0.6.**

```php
class SomeForm extends Form {
	#[Locked]
	public ?string $id = null;
	public string $foo = '';

	public function init(?string $id) {
		$this->id = $id;
	}
}

class Component extends BaseComponent
{
	public SomeForm $form;

	public function mount() {
		$this->form->init('id');
	}

    public function render()
    {
        return <<< 'HTML'
<div>

	<input type='text' wire:model='form.foo' />
</div>
HTML;
	}
}
```

Changing the value of `foo` will fail with the following error: `Exception: Cannot update locked property: [form.id]`


joshhanley:
@ildyria thanks for the PR! In future, please don't open two PRs. If you work out a fix, just push changes to the original failing test PR and rename it 🙂 Going to close this in favour of PR #7060.

PR #7060, opened 2023-10-12
Fixes regression on locked properties in forms introduced by v3.0.6

ildyria:

Fixes a fix #6926

:heavy_check_mark: 

:x: Does not contains other changes.

Failing tests are available from my other PR: #7057

Thanks for contributing! 🙌

---

The fix presented in #6926 did not work for locked properties in components with a form.
We may want to lock one of the property of the form. Unfortunately this locks all the properties of said form.
**This is a regression introduced by version 3.0.6.**

```php
class SomeForm extends Form {
	#[Locked]
	public ?string $id = null;
	public string $foo = '';

	public function init(?string $id) {
		$this->id = $id;
	}
}

class Component extends BaseComponent
{
	public SomeForm $form;

	public function mount() {
		$this->form->init('id');
	}

    public function render()
    {
        return <<< 'HTML'
<div>

	<input type='text' wire:model='form.foo' />
</div>
HTML;
	}
}
```

This is because in a locked `form.id` property, when updating `form.foo` the code would check `'form' == 'form'`.
This refactored the logic back to `startWith` but adding the nested property character delimiter `.` or compare with equality.

#### Old logic:
```php
            ->filter(function ($attr) use ($fullPath) {
                $attributeRoot = (string) str($attr->getName())->before('.');
                $updatePathRoot = (string) str($fullPath)->before('.');

                return $attributeRoot === $updatePathRoot;
            })
```

#### New logic:
```php
            ->filter(fn ($attr) => str($fullPath)->startsWith($attr->getName() . '.') || $fullPath === $attr->getName())
```


lrljoe:
@ildyria - believe it or not, I actually used the logic below in my original PR:
```
->filter(fn ($attr) => ((str($fullPath)->startsWith($attr->getName() . ".")) || str($fullPath)->is($attr->getName())))
```
Which roughly matches yours.  I'm not sure whether my approach of
```
str($fullPath)->is($attr->getName)
```
is cleaner than your approach of
```
$fullPath === $attr->getName()
```
or whether yours is the better of the two!

The reason I used str($fullPath)->is($attr->getName() , was to ensure that the $fullPath is actually a string prior to comparison, which may avoid mismatches, or anything odd being introduced.

It was @calebporzio who tweaked my implementation, so we can all blame him 😆

lrljoe:
@lyriad - may be worthwhile merging in the failing test into this PR to be honest, if they take the fix, then they'll be glad not to have to merge the failing test in too.

swissinside:
I hope that this PR will be merged soon. I cannot update from v3.0.5 because of this bug.

lrljoe:
Noting that on this PR, I did some light testing:

1) Locked Form (Form $form)
2) Unlocked Form (Form $form) with:
  a) Locked Fields 
  b) Unlocked Fields
3) Locked String
4) Unlocked String
5) Locked Array
6) Unlocked Array

All seemed to work okay.

calebporzio:
Thank you for this!

PR #7075, opened 2023-10-15
Utilize PHPUnit\Framework\TestCase class to broaden compatibility with other packages

ph7jack:
Enhanced Compatibility with Other Packages Utilizing `PHPUnit\Framework\TestCase` Class

Yes, enhancing compatibility is always beneficial as it promotes interoperability with a wider array of packages. 
No discussion was created.

**2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)**

Yes, a separate branch named `add-testcase-compatibility` has been created for this purpose.

No, this PR solely focuses on enhancing compatibility by utilizing the `PHPUnit\Framework\TestCase` class.

No, existing tests continue to pass with these changes, serving primarily to extend compatibility with other packages/projects, ensuring seamless integration as they evolve.

Testing Livewire components or integrations within packages can sometimes pose challenges. This PR aims to streamline this process, enabling developers to leverage the Livewire testing infrastructure for a broader range of testing scenarios. This change is fairly straightforward yet significantly impactful.

```php
Livewire::visit(new class() extends Component
{
    public $color = '#000';

    {
        return <<<'BLADE'
            <div>
                <x-color-picker label="Color Picker" name="color" wire:model="color" />
            </div>
        BLADE;
    }
})
    ->waitForAlpineJs()
    ->assertInputValue('color', '#000');
```

By adhering to PHPUnit's native TestCase class, we foster a more standardized approach, which in turn facilitates integration with other packages. This is particularly useful for projects that rely on a variety of packages and aim for a cohesive working environment.

Thanks for this awesome package! 🙌

### DIFF
--- a/src/Features/SupportTesting/DuskTestable.php
+++ b/src/Features/SupportTesting/DuskTestable.php
@@ -120,7 +120,7 @@ class DuskTestable
 
             $components = null;
 
-            try { (new $testClass)->$method(); } catch (\Exception $e) {
+            try { (new $testClass($method))->$method(); } catch (\Exception $e) {
                 if (! $e->isDuskShortcircuit) throw $e;
                 $components = $e->components;
             }
--- a/src/Features/SupportTesting/DuskTestable.php
+++ b/src/Features/SupportTesting/DuskTestable.php
@@ -4,9 +4,9 @@ namespace Livewire\Features\SupportTesting;
 
 use Illuminate\Support\Facades\Route;
 use Laravel\Dusk\Browser;
+use PHPUnit\Framework\TestCase;
 use function Livewire\{ invade, on };
 use Illuminate\Support\Arr;
-use Tests\TestCase;
 
 class DuskTestable
 {
--- a/src/Features/SupportTesting/DuskTestable.php
+++ b/src/Features/SupportTesting/DuskTestable.php
@@ -120,7 +120,13 @@ class DuskTestable
 
             $components = null;
 
-            try { (new $testClass($method))->$method(); } catch (\Exception $e) {
+            try { 
+                if (\Orchestra\Testbench\phpunit_version_compare('10.0', '>=')) {
+                    (new $testClass($method))->$method(); 
+                } else {
+                    (new $testClass())->$method(); 
+                }
+            } catch (\Exception $e) {
                 if (! $e->isDuskShortcircuit) throw $e;
                 $components = $e->components;
             }


PH7-Jack:
I believe also, it would be interesting to expose the livewire test case. https://github.com/livewire/livewire/blob/07de2790f0aaf47c53f3a8cf942139a54b4609d8/tests/BrowserTestCase.php

calebporzio:
I'm sorry, I don't quite understand what's preventing people from using these testing utilities in their current package's test suite?

PH7-Jack:
> I'm sorry, I don't quite understand what's preventing people from using these testing utilities in their current package's test suite?

Hi, it's because The `Livewire::visit()`, is not working with PHPUnit 10, and inside packages.
For my package [wireui](https://github.com/wireui/wireui), the usage of `Livewire::visit()` will be excellent. Tests will be easier and simpler to create.
However, the lines 99 and 123 block it. 

Line 99: only accept your test case class 
Line 123: PHPUnit 10 needs the method name parameter in the class constructor.

lrljoe:
I'd second this being a pain presently, very keen to use some of the native testing methods for packages, but the current approach stops me in my tracks when it comes to Dusk/Livewire::visit.

Might be that I'm missing something tho!

src/Features/SupportTesting/DuskTestable.php
```diff
@@ -120,7 +120,7 @@ static function loadTestComponents()
 
             $components = null;
 
-            try { (new $testClass)->$method(); } catch (\Exception $e) {
+            try { (new $testClass($method))->$method(); } catch (\Exception $e) {
```
crynobone, line 123:
```suggestion
            try { 
                if (\Orchestra\Testbench\phpunit_version_compare('10.0', '>=') {
                    (new $testClass($method))->$method(); 
                } else {
                    (new $testClass())->$method(); 
                }
            } catch (\Exception $e) {
```

Support both PHPUnit 9 and PHPUnit 10.

crynobone:
Looks good, the failing tests can be fixed by adding a `pause()`:

```diff
        ->assertDontSee('Toggle Me!')
        ->waitForLivewire()->click('@toggle')
+       ->pause(500)
        ->assertSee('Toggle Me!')
        ;
```

crynobone (approved):

devajmeireles:
> Looks good, the failing tests can be fixed by adding a `pause()`:
> 
> ```diff
>         ->assertDontSee('Toggle Me!')
>         ->waitForLivewire()->click('@toggle')
> +       ->pause(500)
>         ->assertSee('Toggle Me!')
>         ;
> ```
> 
> I believe it depends on Livewire event cycle where there's a small wait between click and DOM updates.

I think `waitForText` is most appropriate in this case.

crynobone:
Wait for text would be good but it doesn't add to the assertion count.

PH7-Jack:
I believe this test should be fixed in another PR to livewire, since it's not part of this feature.

calebporzio:
Gotcha, sounds good!

PR #7083, opened 2023-10-16
Docs: note about nginx 404 `livewire.js`

robsontenorio:
This is the beginner's very first issue when using Nginx.

Feel free to tweak the text. This note was taken from  https://benjamincrozat.com/livewire-js-404-not-found

image

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -8,6 +8,17 @@ composer require livewire/livewire
 
 That's it — really. If you want more customization options, keep reading. Otherwise, you can jump right into using Livewire.
 
+## Note about 404 `/livewire.js`
+
+Once this file is served by Livewire. Nginx can’t find it, so it responds with a 404 error. Add this config, which will prevent Nginx from interfering with this specific location.
+
+```
+location = /livewire/livewire.js {
+    expires off;
+    try_files $uri $uri/ /index.php?$query_string;
+}
+```
+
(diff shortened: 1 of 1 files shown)


calebporzio:
Good call - thanks for this PR! I reworked the wording a little, but the content is still there for people. Thanks!

johnbacon:
Alternatively, we're curious if this PR is something being considered: https://github.com/livewire/livewire/pull/6891

As discussed in https://github.com/livewire/livewire/discussions/6575#discussioncomment-6853744, it's generally not considered best practice to serve static files in this way. I believe that PR's solution should supersede an Nginx workaround.

clementmas:
That nginx config recently stopped working for me until I changed the filename to `location = /livewire/livewire.min.js`. So I guess the docs are wrong now.

It's clearly not a good practice to serve a JS file through PHP. There's probably a better solution but I haven't dug into the subject yet.

PhiloNL:
> That nginx config recently stopped working for me until I changed the filename to `location = /livewire/livewire.min.js`. So I guess the docs are wrong now.
> 
> It's clearly not a good practice to serve a JS file through PHP. There's probably a better solution but I haven't dug into the subject yet.

Currently, Livewire changes the route depending on `APP_DEBUG` if the value is true the file is `/livewire/livewire.js` otherwise it's `/livewire/livewire.min.js`

PR #7085, opened 2023-10-17
Enhancing custom namespace, layout, and view path for the HMVC concept.

hexters:
Thank you for creating this incredibly amazing package to simplify and modernize the application development process. However, I have a small suggestion as outlined below.

Sometimes, we want to separate Livewire files outside the `app` folder, for example, in a `Modules` folder, especially for larger projects to maintain proper file organization. This can be achieved by modifying the configuration in the Middleware of each module route, as shown below:

```php
. . .

namespace Modules\Admin\Http\Middleware;

. . .

Config::set('livewire.view_path', base_path('Modules/Admin/views/livewire'));
Config::set('livewire.class_namespace', "Modules\\Admin\\Livewire");
Config::set('livewire.layout', "admin::components.layouts.app");

return $next($request);

```

However, there is a minor issue when performing a submit event. The configuration set up previously cannot be received by the `livewire.update` route because it is located outside the middleware that has already been created. Therefore, I attempted to make a slight change by adding three variables to the `wire:snapshot`:

- `view_path`
- `class_namespace`
- `layout`

These three variables will then be set in the `Livewire\Mechanisms\HandleComponents\HandleComponents` file so that the response from the `livewire.update` route has the same configuration as the Middleware or any other technique.

I hope these minor changes can make Livewire more flexible for various real-world application development scenarios.

### Discussions
Related issues can also be found here: 
- [#1813 - Use Livewire with nwidart/modules](https://github.com/livewire/livewire/discussions/1813)
- [#3873 - Support for modularity in Livewire](https://github.com/livewire/livewire/discussions/3873)

### Thank you.


calebporzio:
Thanks for the submission, however, I don't think we're going to go with this for now. You can configure Livewire's update endpoint to use specific middleware: https://livewire.laravel.com/docs/installation#configuring-livewires-update-endpoint

That might help you with applying this config to update requests.

Thanks!

PR #7099, opened 2023-10-18
Fixes Livewire.dispatchTo

vladrusu:
This pull request fixes #6222, so that now Livewire.dispatchTo and $wire.dispatchTo accept the same arguments.

According to [documentation](https://livewire.laravel.com/docs/javascript#interacting-with-events) you can do something:
```Livewire.dispatchTo('dashboard', 'post-created', { postId: 2 });```

... except that **you can't.**

Why? That's because currently Livewire.dispatchTo requires 4 arguments, the first being **component**, [which is not used in the code](https://github.com/livewire/livewire/blob/5ab209d65f5409fac781dda1d28f030706b1619f/js/features/supportEvents.js#L67).

So, your code has to be something like:
```Livewire.dispatchTo(null, 'dashboard', 'post-created', { postId: 2 });```
which is weird and counter-intutive.

But this works:

```$wire.dispatchTo('dashboard', 'post-created', { postId: 2 });```

So, this pull request makes that Livewire.dispatchTo and $wire.dispatchTo work the same.

### DIFF
--- a/js/$wire.js
+++ b/js/$wire.js
@@ -155,7 +155,7 @@ wireProperty('$on', (component) => (...params) => listen(component, ...params))
 
 wireProperty('$dispatch', (component) => (...params) => dispatch(component, ...params))
 wireProperty('$dispatchSelf', (component) => (...params) => dispatchSelf(component, ...params))
-wireProperty('$dispatchTo', (component) => (...params) => dispatchTo(component, ...params))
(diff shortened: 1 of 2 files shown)


SanderMuller:
Ran into this issue, took me a while to find out why `dispatchTo` wasn't working. LGTM!

claushellsing:
Thanks , I ran into this issue aswell

vladrusu:
Great, thanks!

PR #7102, opened 2023-10-18
add isolate parameter

eilerth:

This is needed for my company to be able to upgrade our site to livewire v3, and use packages that require livewire v3 (like filament). Since livewire v3 now bundles requests, the user experience becomes very synchronous. We have many places where we're using multiple livewire components on the page, and sometimes some of them take time to load content, so making one interaction rely on another interaction to complete ends up significantly hurting the user experience.

I did create a discussion post here: https://github.com/livewire/livewire/discussions/6875
Looking now, I see there are also a few other posts that highlight the issue, or that I think this would solve:
- https://github.com/livewire/livewire/discussions/7081
- https://github.com/livewire/livewire/discussions/6147
- https://github.com/livewire/livewire/discussions/3614
- https://github.com/livewire/livewire/discussions/6143

Yes

No.
However, I'll note that "npm run build" on the raw forked project would build js files that would fail existing tests. So I ran "npm update" and then "npm run build" and then those existing tests ran just fine. I decided to keep package-lock.json in the PR for that reason.

Yes, it does include a thorough test.
However! It does not pass. I couldn't find specific documentation to confirm this, but I think the testing framework itself does not allow this to be properly tested. I suspect the testing framework cannot allow multiple php processes to be running concurrently, which makes this impossible to really test since the nature of this requires multiple php processes to be responding to a javascript request while another is still being processed. 
I included the test anyway, but very curious for thoughts on how to proceed, or perhaps I'm just doing something wrong.
I've tested this by substituting this forked package into real projects (just in my local environment) and it has worked great.

Consider you have a livewire component that, once interacted with, may take several seconds to get a response back from the server.
Any other components on that page will need to wait until that component's request gets a response back from the server. Interactions while waiting for that end up looking like buggy behavior or just feels awkward/slow.
With that, the user experience becomes a very synchronous one, which is a big change from livewire v2's asynchronous behavior.
This change allows you to include an "isolate" parameter on a livewire component to then allow it to skip the request bundling (similar to how "lazy" can be added).
Example:
`<livewire:my-component isolate />`

Note: some of the logic I actually borrowed from this other PR that attempts to implement similar logic, but only for calling `wire:poll`: https://github.com/livewire/livewire/pull/6819

Thanks for contributing! 🙌
Happy to! Thanks for the detailed guidance on contributing. This is the first time I've contributed to a project like this.


nanodreams:
Hi @eilerth 

Much appreciated your contribution and I confirm I face the same situation. While we might have multiple components or just a single component, if you trigger few actions, they will need to be "sequential" as you indicated in your other posts.

Not sure if might be a new PR or could be included in this one, but in your definition, you are including the "isolate" at the level of livewire "html element" like `<livewire:my-component isolate />`. In my use case, I also use "event/dispatch" to trigger actions, let me share some use case:

1. You trigger an action that might be having multiple effects, so to be able to "react" the front, I process the first request and then trigger an event at the PHP level before returning to the browser, that way .. the UI get refreshed, then it will fire the "indicated" event. 
2. In the meantime, you might want to interact with the UI triggering additinoal details, like in your case with a livewire html element "isolate".

So not sure if we might be talking by a "different" level of "isolation".

From what I understand, the idea behind the `<livewire:my-component isolate />` is that it will execute every action within this component as "isolated" so like in V2, all request will be independent (totally support the use case). But I also can see a more narrow down isolation, where you can control each event/action.

* Scenario A (yours): Configure a component to act like V2 (totally isolated)
`<livewire:my-component isolate />` 

* Scenario B: Where you can control each action/event how to be triggered.
`<button wire:click.isolate="myAction" />`
`<button @click="$dipatch('myAction').isolate()"/>`
`$this->dispatch('myAction')->self()->isolated();`

Not sure if this is going too far, but this way seems to be "very much" having the flexibility to decide what and when do you want the isolation to happen.

PS: Again .. much appreciated your contribution and also I'm here to help ;-)

eilerth:
@nanodreams, I see how there could be multiple levels of isolation. That's a great point. I like the idea of having more granular control over what requests get isolated. I can see how you might not necessarily want to isolate everything within a component, but still allow certain actions within the component to be isolated. 

[This other PR (#6819)](https://github.com/livewire/livewire/pull/6819) was taking things in a similar direction to what you're describing. I also thought a comment in there was nifty, which suggested giving control over grouping requests too.

I'm currently thinking the PR as it is now solves the problem for several discussion posts, and the ability to have more granular control could be done in a follow-up PR.

vladrusu:
Alternate name: parallel (loading).
Isolate sounds weird.

Another good idea, i think, or workaround for the "sync" nature of Livewire 3 whould be a "priority" option.
For example:

```
<livewire:my-component /> (or  priority="0") -> loaded first
<livewire:my-component priority="2" /> -> loaded third
<livewire:my-component priority="1" /> -> loaded second
```

This way, we could load first some components that are loaded/fast or required on the page. And we could de-prioritize slow components.

What do you think?

eilerth:
@vladrusu, thanks for the thoughts! 

> Alternate name: parallel (loading). Isolate sounds weird, even if I get the idea of isolating requests.

"Isolate" feels more natural/accurate to me, since at a component level you're telling it to "isolate" subsequent requests from the default bundling of requests with other livewire components. 
"Parallel" seems to me like it could be misleading. For example, during the initial page request, all livewire components will be processed synchronously on the server side. It isn't until interactions happen in the browser that things can start to run asynchronously/in parallel. Though, after writing this, I suppose the same could be said about "isolate" during the initial request. Personally, "isolate" still seems more natural to me, but I could see either was working.

> Another good idea, i think, or workaround for the "sync" nature of Livewire 3 whould be a "priority" option. For example:
> 
> ```
> <livewire:my-component /> (or  priority="0") -> loaded first
> ...
> <livewire:my-slow-component priority="2" /> -> loaded third
> ...
> <livewire:my-component priority="1" /> -> loaded second
> ```
> 
> This way, we could load first some components that are loaded/fast or required on the page. And we could de-prioritize slow components.
> 
> What do you think?

I like the idea, but I'm not sure how practically it'd play out. A couple thoughts:

- The initial page request would load all the components synchronously, and deliver them all to the browser at the same time, so I'm not seeing much for use cases where it'd matter what order they're processed. Perhaps if it implicitly applied "lazy" to components with priority > 0, but I personally feel like that is getting a little to "magical" to have that going on behind the scenes.
- Once the page is loaded, the ability to isolate requests seems to solve the problem with prioritizing requests since they can all be handled asynchronously. 

So I think the "priority" is a nifty idea, but I suppose I'm not seeing the specific use cases where it'd be helpful, particularly if this "isolate" feature makes it in. Once combined with "lazy" to delay loading it until the browser can trigger everything asynchronously, I think prioritizing components becomes mostly irrelevant.

vladrusu:
If your "isolate" feature is approved (hope so), there is little sense for the "priority" idea.
But if it doesn't, I think it's an interesting idea.

Really, I am one of the "victims" of the synchronous nature of Livewire V3.
Why? Because in my admin I have a header of slow loading LW components (stats) and beneath, a component (button) which basically does nothing until all the stat components above are loaded.

For me, both your solution and my alternative solution would do the trick.

nxgcx:
Hope to see this merged in soon as it is very useful for our project.

alxndrmlr:
> Hi @eilerth
> 
> Much appreciated your contribution and I confirm I face the same situation. While we might have multiple components or just a single component, if you trigger few actions, they will need to be "sequential" as you indicated in your other posts.
> 
> Not sure if might be a new PR or could be included in this one, but in your definition, you are including the "isolate" at the level of livewire "html element" like `<livewire:my-component isolate />`. In my use case, I also use "event/dispatch" to trigger actions, let me share some use case:
> 
> 1. You trigger an action that might be having multiple effects, so to be able to "react" the front, I process the first request and then trigger an event at the PHP level before returning to the browser, that way .. the UI get refreshed, then it will fire the "indicated" event.
> 2. In the meantime, you might want to interact with the UI triggering additinoal details, like in your case with a livewire html element "isolate".
> 
> So not sure if we might be talking by a "different" level of "isolation".
> 
> From what I understand, the idea behind the `<livewire:my-component isolate />` is that it will execute every action within this component as "isolated" so like in V2, all request will be independent (totally support the use case). But I also can see a more narrow down isolation, where you can control each event/action.
> 
> * Scenario A (yours): Configure a component to act like V2 (totally isolated)
>   `<livewire:my-component isolate />`
> * Scenario B: Where you can control each action/event how to be triggered.
>   `<button wire:click.isolate="myAction" />`
>   `<button @click="$dipatch('myAction').isolate()"/>`
>   `$this->dispatch('myAction')->self()->isolated();`
> 
> Not sure if this is going too far, but this way seems to be "very much" having the flexibility to decide what and when do you want the isolation to happen.
> 
> PS: Again .. much appreciated your contribution and also I'm here to help ;-)

I like the control idea, I wanted to mention an idea to have this defined at the listener level as well re: my comment on the discussion https://github.com/livewire/livewire/discussions/6143#discussioncomment-7447762

Something like (using the `isolate` terminology that folks seem to be coalescing around...

```<?php

namespace App\Livewire\Directory;

use App\Services\DirectoryService;
use Illuminate\Support\Facades\DB;
use Livewire\Attributes\On;
use Livewire\Component;

class DirectoryMapComponent extends Component
{
    public $shouldSuggestFilterMapBounds = false;

    #[On('directory.filtersReset', isolate: true)]
    public function handleFiltersReset()
    {
        $this->computeFilterOrganizationCounts();
    }

    #[On('directory.filtersUpdated', isolate:true)]
    public function handleFilterUpdated($filters)
    {
        $this->shouldSuggestFilterMapBounds = $filters['limitToMap'];
        $this->computeFilterOrganizationCounts($filters);
    }

// ... rest of the file
}
```

calebporzio:
I like this, thanks for the PR! Here are my thoughts.

Ideal world is two points in the API to declare isolation:

A) A nested component: `<livewire:child isolate/>`
B) A component itself (attribute): `#[Isolate] class SomeComponent extends Component`

(these two are taken care of by this PR, I removed the left over route macro stuff)

Here are reasons I'm not ready to merge this:

A) The test had a `return;` at the top to skip it, I ran the test and it failed. I think there needs to be a cleaner passing test (multiple ideally).

B) Currently, if a component is isolated, it's next polling request is hung up until other components are finished with their request. Here's the scenario:

* We have 3 components polling
* one is isolated (A) the rest are not (B, C)
* Component's A & (B/C) all send at first
* If the (B/C) request takes too long, component A's polling is held up

Here's example code:
```html
<livewire:polling-child isolate />
<livewire:polling-child sleep="4" />
<livewire:polling-child sleep="4" />
```

C) If a component is isolated, it triggers new requests for un-isolated components even if they're not finished with their previous request:

We have 3 components polling:
* One is isolated (C) the rest are not (A, B)
* Component's (A/B) & C all send at first
* Component (C) properly polls, but triggers new requests for (A/B) even though the old requests are not finished. Causing race-conditiony type stuff.

```html
<livewire:polling-child sleep="4" />
<livewire:polling-child sleep="4" />
<livewire:polling-child isolate />
```

D) If a parent and child component are bound together with a (`#[Reactive]`) property between them, and the child is isolated, then when the parent attempts to bundle the components together for a request, the isolation will break that and ruin the reactivity of the property.

(I haven't tested this personally but I assume this is what would happen)

It would be great if someone wrote failing tests for these scenarios so we can fix them. I know it's not the easiest thing to test but it kinda needs to be for such a deep core change like this

Thanks everyone!

vladrusu:
@calebporzio Had any time to look at my "priority" idea ? (from [above](https://github.com/livewire/livewire/pull/7102#issuecomment-1773733921))

alxndrmlr:
I started doing some source diving to understand / write tests tonight. The one thing I'm trying to find and understand is the queuing for requests.

That is to say, is it possible for us to remove the blocking (or conditionally do that) wherein subsequent requests don't need to finish resolving before the current one has...

Can anyone chime in on why that was introduced re: problems it solves/alleviates?

alxndrmlr:
Just poking around now if I do this

```js
export async function waitUntilTheCurrentRequestIsFinished(callback) {
    return new Promise(resolve => {
        resolve(callback())

        // if (sendingRequest) {
        //     afterSendStack.push(() => resolve(callback()))
        // } else {
        //     resolve(callback())
        // }
    })
}
```
In request.js just let the requests fly, I'm only seeing a couple failing browser tests `avigate_scrolls_to_top_and_back_preserves_scroll` and `navigate_back_works_from_page_without_a_livewire_component_that_has_a_script_with_data_navigate_track` 

I'm having some issues running the LegacyBrowser tests locally (before I made any changes) so I still need to look in there.

alxndrmlr:
Ignore my comments I was diving on the main branch and not this PR branch I see that @eilerth has done this work to ensure its its not getting queued.

calebporzio:
Thanks Alex! Yeah we need a queue so that single component only has a single request out at a time. Because component's use "snapshots" to capture there server-side state at a given time and we can't create divergent snapshots. livewire network requests for a component are kinda like git commits in that way - they are snapshots in time.

alxndrmlr:
Thanks @calebporzio!

Okay so doing some more digging it seems that we'd need to change the logic for `sendingRequest` and `waitUntilTheCurrentRequestIsFinished` such that `isolated` components are not held up by other requests in-flight re: this scenario

> B) Currently, if a component is isolated, it's next polling request is hung up until other components are finished with their request. Here's the scenario:
> 
> * We have 3 components polling
> * one is isolated (A) the rest are not (B, C)
> * Component's A & (B/C) all send at first
> * If the (B/C) request takes too long, component A's polling is held up
> 
> Here's example code:
> 
> ```
> <livewire:polling-child isolate />
> <livewire:polling-child sleep="4" />
> <livewire:polling-child sleep="4" />
> ```

Since we need to respect the sequencing of requests with regards to snapshots, we will need to have `isolated` components `waitUntilTheCurrentRequestIsFinished` for themselves only. So there would now need to be state tracking for `sendingRequest` in;

- a global context (for all non-isolated components)
- a per-isolated component

I have a feeling that [Binding to child data using wire:model](https://livewire.laravel.com/docs/nesting#binding-to-child-data-using-wiremodel) will not be workable with isolated components with this change... I still need to source dive on that and play around with it (thinking out loud here so others can chime in with suggestions that are following along).

My hunch on this is re: BaseModelable:update(...)

https://github.com/livewire/livewire/blob/3c30eabe7e878d5aa8b9ebe883bdd0936c54866a/src/Features/SupportWireModelingNestedComponents/BaseModelable.php#L34-L45

If the parent component is `isolated` then the child component will not get bundled up with the request and thus the 

> The request WILL NOT contain both parent and child component updates.

So we probably have to do more thinking about how to handle the usage of both `isolated` and `modelable`... Perhaps its something like this;

**If you are `isolated` we will not bundle you with other requests, UNLESS you have `modelable` children components**. In which case, we will still bundle your children components with you...

^ That seems like a more complex change and I thought I'd raise this for discussion first before doing another session of digging/thinking about this (I've only been able to find an hour or so here or there so I'm trying to peel back a layer, document, and then have something to come back)

alxndrmlr:
> **If you are `isolated` we will not bundle you with other requests, UNLESS you have `modelable` children components**. In which case, we will still bundle your children components with you...

Or... if you're isolated, modelable is not supported... This is probably a saner approach initially since isolated is an opt-in feature... The only catch will be making that apparent to a user if they attempt to use both to avoid them getting stuck...

I've solved B) with a simple modification to `waitUntilTheCurrentRequestIsFinished` that just doesn't wait if the requests is isolated, but not I need to make the overall thing more robust re: `isolated` components need to wait inline behind themselves.

eilerth:
Thanks for all the discussion! 
Much of it I'll need to take more time to absorb before I can respond intelligently, but I wanted to address the testing part.

@calebporzio, the testing is something I got hung up on and my theory is that the testing framework doesn't allow us to write a proper test. Hopefully I'm wrong..
I didn't prove this out (not sure how), but I think each test it is only allowing one php process to be running at a time. If that is true, then it is forcing synchronous behavior, and the nature of this change requires us to test an asynchronous behavior.

The test I wrote was following this logic:
- When the page first loads, it should have a "Fast" component loaded already, but a "Slow" component is still being rendered.
- While it is still waiting for the "Slow" component to render, it sends another request for the "Fast" component, and it should see the change to the "Fast" component even before the "Slow" component has rendered.

The first parts of the test work fine (page loads showing "Fast" component, but "Slow" component not yet rendered):
```
            ->assertSee('Fast Toggle')
            ->assertDontSee('Slow Toggle')
```

However, after it interacts with the "Fast" component, it seems that request doesn't actually get processed until after the "Slow" component is finally processed. So my theory is that only one php process can be running at a time, and although the browser in the test sends a request to php asynchronously via javascript, it doesn't get processed server-side until after the Slow component's php process completes.
```
            ->click('@fastToggle')
            ->waitForText('Fast Checked')
            ->assertDontSee('Slow Toggle')
```


I'm one of those that loves the idea of writing tests, but hasn't been disciplined enough to regularly do it, so I have limited experience writing tests and I was hoping someone more experienced could provide some help/insight.

epessine:
I'm currently working on updating a big Livewire 2 project to v3, and we ended up halting the update because of this issue. The thing is, since one of the core ideas behind Livewire is that "every component is an island", I don't see how this isolation behavior should be inserted on a separate attribute. Probably makes much more sense to make `lazy` components "isolated" by default, otherwise the whole purpose of lazy loading is defeated.

On project I'm working on, for example, we have pages that have _multiple_ "lazy loaded" components, each of them making expensive requests. With the request bundling on v3, all these processes are made together, even causing the request to timeout, because on v2 all of these requests were separate. I can't figure out a situation were a lazy loaded component `wouldn't` be isolated, since we usually lazy load components because of expensive processes/queries, and bundling them up could easily extrapolate processing time unexpectedly and make the page less "reactive".

These are my thoughts, and thanks very much to @eilerth and @calebporzio for trying to resolve this issue!

ercsctt:
I agree with @epessine. Alongside my original discussion #6147 I prototyped an "always isolate" pattern for lazy components. I actually struggled to find any real-world use cases where lazy requests should be bundled.

Regardless of the current issues with reliably testing this functionality, I strongly believe it would be better to isolate lazy components by _default_ with the ability to bundle optionally (if that's actually necessary).

joshhanley:
Lazy components are now isolated on the initial request by default in version v3.3.0, see PR #7487. So if you upgrade this should now work.

We may still make an isolate parameter for every component's request (polling or lazy, etc), but that is up to Caleb to decide whether he wants that first. This PR will probably need to be reworked to suit the new request pooling introduced in PR #7456.

calebporzio:
This has been added and is done now. Closing. Thanks!

eilerth:
Wonderful news! I'll be testing this out tomorrow. Very happy to have this resolved :) 
Thank you!!!

PR #7108, opened 2023-10-19
Blade Components with @entangle are not removed from the dom

ju5t:
When upgrading to v3 we cannot remove Blade-components from the dom if they have an entangled value in them. This only happens if there are several of them. Components without `@entangle` do not have this problem.

This PR is a WIP as it throws some unrelated errors at the moment:

```bash
Livewire\Features\SupportEntangle\BrowserTest::can_remove_entangled_blade_components_from_dom_without_side_effects
Facebook\WebDriver\Exception\UnexpectedJavascriptException: javascript error: Cannot read properties of undefined (reading 'hook')
```

Test command:

```bash
phpunit --testsuite Browser --filter can_remove_entangled_blade_components_from_dom_without_side_effects
```

The discussion can be found here.
https://github.com/livewire/livewire/discussions/7095

### DIFF
--- a/src/Features/SupportEntangle/BrowserTest.php
+++ b/src/Features/SupportEntangle/BrowserTest.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features\SupportEntangle;
 
+use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


satoved:
See https://github.com/livewire/livewire/pull/6833

PR #7118, opened 2023-10-20
[v3] Fix wire:loading.delay corner case

austenc:
Encountered a corner case with wire:loading.delay where sometimes the loader shows when it shouldn't, then doesn't hide.

This is related to the following discussion:
https://github.com/livewire/livewire/discussions/6338

**The scenario:**
- First request takes **more** time than the wire:loading.delay, so the loader shows (then hides)
- Second request takes **less** time than the wire:loading.delay, the loader shows when it shouldn't and never hides

This is happening in a real-world scenario when a filter input queries the database (first request), but then clearing the input (backspacing or whatever) only updates Livewire state (no queries). In that case, the first request is slow, the second one is much faster.

~~_This is only a failing test for now, but I'll update this PR with a fix if I figure that out_~~

Edit: included a fix. `started` wasn't getting set back to false when the loading of the first request is finished. Tests are failing because I didn't include built assets


calebporzio:
Wooohoo, thanks Austen - a failing test AND a fix. It must be my birthday. rock on mate

PR #7141, opened 2023-10-24
Allow tags for computed properties

comes:
This PR introduces support for tags in computed and cached properties. Tags provide a way to manage and invalidate cached data associated with specific entities within your application.

If the current cache driver supports tags, you then can do something like.

```php
#[Computed(cache: true, tags: ['posts'])]
public function latestPosts(): Collection
{
	// Caching because this is global
	return Post::latest()->take(10)->get();
}

#[Computed(persist: true, tags: ['posts'])]
public function ownPosts(): Collection
{
	// Persisting because this is scoped to the current instance of the Livewire component
	return Auth::user()->posts()->latest()->take(10)->get();
}
```

When a Post-model is updated/deleted/restored then we clear the caching tag 'posts':
`Cache::tags('posts')->flush();`
After the next refresh, we get a fresh copy, instead of the outdated cached one.

Please see the modified UnitTest for detailed test cacse

### DIFF
--- a/docs/computed-properties.md
+++ b/docs/computed-properties.md
@@ -196,10 +196,31 @@ public function posts()
 }
 ```
 
+Instead of the `key` parameter, you can also use 'tags'.
+
+
+```php
+use Livewire\Attributes\Computed;
+use App\Models\Post;
+
+#[Computed(cache: true, tags: ['foo', 'bar'])]
+public function posts()
+{
+    return Post::all();
+}
+```
+
+
 Then somewhere else in your app, you may clear the cache for that property:
 
 ```php
 Cache::forget('homepage-posts');
+```
+
+or clear the entire tag:
+
+```php
+Cache::tags('foo')->flush();
 ```
 
 ## When to use computed properties?
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -19,6 +19,7 @@ class BaseComputed extends Attribute
         public $seconds = 3600, // 1 hour...
         public $cache = false,
         public $key = null,
+        public $tags = null,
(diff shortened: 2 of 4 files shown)


comes:
I've updated the docs as well.

calebporzio:
Love it! Thanks!

francoism90:
Laravel/upstream removed instructions about caching tags, as being too complicated: https://github.com/laravel/docs/commit/63dde39470588ab1ad39cea9a592f0838ca5ad47

It would be great if you would mention which cache drivers are support, I think it's only redis. Also pruning stale cache tags, is highly recommended.

Maybe the removed instructions should be added to the Livewire docs? Or upstream should revert the change.

comes:
> Laravel/upstream removed instructions about caching tags, as being too complicated: [laravel/docs@63dde39](https://github.com/laravel/docs/commit/63dde39470588ab1ad39cea9a592f0838ca5ad47)
> 
> It would be great if you would mention which cache drivers are supported, I think it's only redis. Also pruning stale cache tags, is highly recommended.
> 
> Maybe the removed instructions should be added to the Livewire docs? Or upstream should revert the change.

In that case, @calebporzio , does Livewire support 3rd party Attributes? Am I able to create my own `Computed` Attribute to achieve this via a package?

francoism90:
@comes It's not the same, but since Laravel uses Cache-keys, you could do `Cache::forget('key');`.

But it's not the same. Tags offer a lot of flexibility, like being able to clear caches for multiple items at once. I wish Laravel would revert the removal of this, as it's used by a lot of packages already.

joshhanley:
@comes yep, there is a `php artisan livewire:attribute {name}` command that should work for making your own 🙂

PR #7142, opened 2023-10-24
Rely on providesTemporaryUrls for preview files

apreiml:
The `providesTemporaryUrls` function tells whether the filesystem supports such urls. It's properly implemented in the S3 and GCS adapters. It also correctly recognizes customized temporary url functions added like described in the larave doc:

https://laravel.com/docs/10.x/filesystem#temporary-urls

By directly using providesTemporaryUrls, there's no need for a filesystem whitelist here.

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -83,6 +83,10 @@ class TemporaryUploadedFile extends UploadedFile
 
     public function temporaryUrl()
     {
+        if (!$this->isPreviewable()) {
+            return null;
+        }
+
         $expiration = now()->addMinutes(30)->endOfHour();
 
         $temporaryUrlOptions = [];
@@ -99,7 +103,7 @@ class TemporaryUploadedFile extends UploadedFile
         );
     }
 
-    public function isPreviewable()
+    public function isPreviewable(): bool
     {
         $supportedPreviewTypes = config('livewire.temporary_file_upload.preview_mimes', [
             'png', 'gif', 'bmp', 'svg', 'wav', 'mp4',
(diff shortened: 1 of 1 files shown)


apreiml:
Reference to that function: https://github.com/laravel/framework/blob/61ea0bc68c4f293194e3132b0bae8af01f41aec7/src/Illuminate/Filesystem/FilesystemAdapter.php#L728

apreiml:
The test failure does not seem to have anything to do with this patch. Was this a temporary failure?

joshhanley:
@apreiml yep failure is run related. If you see most test suites running and one fails, it's likely just a flaky test.

apreiml:
@joshhanley I've reworked the patch, because the method was rather confusing. 

First of all now there's a check, if the file is even previewable. This should match the documentation that temporaryUrls are only available for previewable files. Is `null` here the proper return type? Should this be an exception or an empty string?

Also now the expiration time is the same for all results. 30 minutes is okay, I guess or was there a specific reason for having the url valid for a whole day?

levu42:
@apreiml The reason for having the `->endOfHour()` (https://github.com/livewire/livewire/pull/6669) was to keep preview URLs stable over some time. Back then, when I did that PR, I just fixed that bug, without touching the expiration time (30min vs whole day).

When thinking about it now: Neither a whole day nor 30min are the perfect fit. The application that I was working with had session durations of well over 30min, but never a whole day, rather in the range of 3-4 hours. So maybe something like `->addHours(4)->endOfHour()` could make sense as a general expiration duration.

I'm curious to hear, if @calebporzio had some deeper thoughts of why there was 1 day for S3 storages and 30min for local storages.

calebporzio:
Thank you for this, however going to close this because it's hard for me to grasp the benefit easily and I don't want to break anything.

PR #7145, opened 2023-10-24
Failing test wire:navigate with redirect

saschalalala:
### Quick summary first

Clicking a `wire:navigate` link with target /foo updates the url bar with /foo even if /foo redirects to /bar. I don't think that this is intended behaviour but I suspect that this is how the js component of this feature works right now.

I'm not sure but I think that the current behaviour is a bug. I briefly discussed it in the discord but did not create a discussion yet. 

Yes.

Nope.

It only includes tests right now, because, tbh, I don't know where and how to fix this.

I did that above and added comments to the code. For the record: I would be willing to implement a fix if someone would push me into the right direction. Right now I think about something like 

"If you need this, add `<body wire:navigation-target-url="{{ currentRoute() }}"` to your base template and add 

```javascript
 // something something window.location without reloading the page
})
```

 to your template" because I can't think about a better solution right now.


calebporzio:
All fixed now.

PR #7153, opened 2023-10-25
Add example for overriding methods using the functional API

RVxLab:
I wasn't able to find this in the documentation for the functional API. 

Using an anonymous class it's pretty straightforward to override a method from a trait, but using the functional API it's not immediately obvious.

This PR adds a subsection in the documentation about how to override methods using the functional API for public methods and for protected methods using the `protect` function.


nunomaduro:
I don't think we plan to document this at the moment. But going to let Taylor to decide.

PR #7162, opened 2023-10-26
Add missing new keyword in docs

josebailo:
In the Volt package section, in an example creating a class-based component there is a missed `new` keyword.


calebporzio:
Thanks!

PR #7165, opened 2023-10-26
Update modelable to support numeric array keys

joshhanley:
Currently, if you have a numeric array in a parent component and try to use modelable to bind it to a child component `<livewire:child wire:model="items.0.name" />`, it throws this error:

image

This is because javascript doesn't support using dot notation to access array indexes. So `items.0.name` won't work, and needs to be changed to `items[0].name`.

This PR fixes this by changing any key indexes to array notation just before we output `x-model` and `x-modelable` on the root element.

Hope this helps!


DanielCoulbourne:
LGTM

joshhanley:
Closing in favour of PR #7173 which also fixes this issue.

PR #7166, opened 2023-10-26
Add `.live` support to modeable

joshhanley:
This PR adds support for `.live` to `#[Modelable]`, allowing you to do `<livewire:child wire:model.live="foo" />` and the parent will update reactively when the child updates.

The way it does this is to add a `$live` property to the `$wire` object. So any property that is prefix with `$live` will commit changes immediately. For example `$wire.$live.name = 'test'` will update the name property and request a commit to the server.

This is something that could also be useful when using Alpine/ javsacript, as instead of doing `$wire.$set('name', 'test', true)` to live update a Livewire property, we can instead just write `$wire.$live.name = 'test'`.

Now, when using `#[Modelable]`, if it detects `wire:model.live="foo"` on the child component directive, it will output `x-model="$wire.$parent.$live.foo"`.

~~Definitely open to ideas on how to improve the implementation of this PR, but I thought I'd push it up as it currently works.~~

Edit: Changed `$live` to be a proxy and it works much better and looks cleaner.


joshhanley:
Changed to draft as there are failing tests

joshhanley:
Closing in favour of PR #7173.

PR #7172, opened 2023-10-27
Fix: Add `! important` modifier to `x-cloak` style

ralphjsmit:
I noticed that an Alpine `x-cloak` attribute was not being used – there was still the tiny flicker on page load. I disabled JavaScript and I noticed that it was being overwritten by a normal `flex`.

Flex

The CSS-file came after the `@livewireStyles` and moving it before the `@livewireStyles` fixed the issue. However, it was quite confusing to me and I believe that `x-cloak` should have the highest priority always that you can sanely give to it. In particular since just a very simple (accidental) thing like the order of a CSS-file/directives already can override it.

Thanks!

### DIFF
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -108,7 +108,7 @@ class FrontendAssets
             }
 
             [x-cloak] {
-                display: none;
+                display: none !important;
             }
         </style>
         HTML;


vladrusu:
I second this. I already use ```display: none !important``` in my code (I overwrited the [x-cloak] style).

mi-dave:
This caught me out because the Alpine docs [explicitly recommend](https://alpinejs.dev/directives/cloak) using `!important`, so I assumed Livewire did too.

calebporzio:
Perfect yep, makes total sense. thanks!

PR #7180, opened 2023-10-28
Fix bug with legacy model binding where uninitialised attributes couldn't be bound to

jasonlbeggs:
This PR adds a failing test for a bug where you can't bind to model attributes that haven't been initialized yet.

Feel free to refactor the test if you want!


calebporzio:
Thanks fellas!

PR #7076, opened 2023-10-16
Utilize Laravel's Container for Dependency Resolution in LivewireServiceProvider

Omranic:
This pull request updates the LivewireServiceProvider class to utilize Laravel's container for resolving dependencies and instantiating objects, improving modularity and flexibility for app developers. With this change, developers can easily override core Livewire functionalities in their service providers as needed, such as Mechanisms and Features.

These changes have been made with the aim of maintaining backward compatibility and ensuring that there are no side effects or incompatibilities.

-----

Code snippet of the improvement and reasons why it's useful:

```php
namespace App;

use Illuminate\Support\ServiceProvider;

class AppServiceProvider extends ServiceProvider
{
    public function register()
    {
        $this->app->singleton(\Livewire\Mechanisms\HandleRequests\HandleRequests::class, \App\HandleRequests::class);
    }
}
```

Example of a `HandleRequests` Mechanism override, changing the mechanism boot logic. This is just an example, but there's many other use cases.
```php
namespace App;

use Livewire\Mechanisms\HandleRequests\HandleRequests as BaseHandleRequests;

class HandleRequests extends BaseHandleRequests
{
    function boot()
    {
        // ...
    }
}

```

Thanks for your consideration! 🙌

PR #7181, opened 2023-10-29
Utilize Laravel's Container for Dependency Resolution in LivewireServiceProvider

Omranic:
Livewire Reference: #7076
Jetstream Reference: https://github.com/laravel/jetstream/pull/1390
The required update in `laravel/jetstream` has been merged into core, and tagged in [v4.0.4](https://github.com/laravel/jetstream/releases/tag/v4.0.4)

-----

This pull request updates the LivewireServiceProvider class to utilize Laravel's container for resolving dependencies and instantiating objects, improving modularity and flexibility for app developers. With this change, developers can easily override core Livewire functionalities in their service providers as needed, such as Mechanisms and Features.

These changes have been made with the aim of maintaining backward compatibility and ensuring that there are no side effects or incompatibilities.

-----

Code snippet of the improvement and reasons why it's useful:

```php
namespace App;

use Illuminate\Support\ServiceProvider;

{
    public function register()
    {
        $this->app->singleton(\Livewire\Mechanisms\HandleRequests\HandleRequests::class, \App\HandleRequests::class);
    }
}
```

Example of a `HandleRequests` Mechanism override, changing the mechanism boot logic. This is just an example, but there's many other use cases.
```php
namespace App;

use Livewire\Mechanisms\HandleRequests\HandleRequests as BaseHandleRequests;

class HandleRequests extends BaseHandleRequests
{
    function boot()
    {
        // ...
    }
}

```

Thanks for your consideration! 🙌

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -40,7 +40,7 @@ class LivewireServiceProvider extends \Illuminate\Support\ServiceProvider
 
     protected function bootEventBus()
     {
-        (new \Livewire\EventBus)->boot();
+        app(\Livewire\EventBus::class)->boot();
     }
 
     protected function getMechanisms()
@@ -61,7 +61,7 @@ class LivewireServiceProvider extends \Illuminate\Support\ServiceProvider
     protected function registerMechanisms()
     {
         foreach ($this->getMechanisms() as $mechanism) {
-            (new $mechanism)->register($this);
+            app($mechanism)->register($this);
         }
     }
 
@@ -74,7 +74,7 @@ class LivewireServiceProvider extends \Illuminate\Support\ServiceProvider
         }
 
         foreach ($this->getMechanisms() as $mechanism) {
-            (new $mechanism)->boot($this);
+            app($mechanism)->boot($this);
         }
     }
 


calebporzio:
Thank you for seeing this feature through all the way. super valuable!

calebdw:
This PR introduced a memory leak---my test suite can't even make it through 1/10th of the tests without running out of memory:

https://github.com/livewire/livewire/discussions/7344

craigpotter:
> This PR introduced a memory leak---my test suite can't even make it through 1/10th of the tests without running out of memory: #7344

Yes we are having the same issue

alejandrotrevi:
Same here, having the same issue, can't update to 3.2.0 because of this 🤔

Omranic:
@calebdw @craigpotter @alejandrotrevi @WayneHarris Thanks for reporting this issue.

Would someone help me with a step-by-step guide on how to reproduce this locally? Thanks

craigpotter:
> @calebdw @craigpotter @alejandrotrevi @WayneHarris Thanks for reporting this issue.
> 
> Would someone help me with a step-by-step guide on how to reproduce this locally? Thanks

I believe there is a demo repo discussed here https://github.com/livewire/livewire/discussions/7344#discussioncomment-7659901

WayneHarris:
> @calebdw @craigpotter @alejandrotrevi @WayneHarris Thanks for reporting this issue.
> 
> Would someone help me with a step-by-step guide on how to reproduce this locally? Thanks

@Omranic Did the demo of the issue help? https://github.com/livewire/livewire/discussions/7344#discussioncomment-7659901

WayneHarris:
@calebporzio @Omranic I've tested the following suggestion with a new $skipWhenRegistered property, and it stops the skip when callbacks registering multiple times.

https://github.com/livewire/livewire/discussions/7344#discussioncomment-7805496

Do you think the approach would be suitable? Is it worth raising a PR to livewire for further consideration?

calebdw:
Looks to be related: https://github.com/laravel/framework/pull/49334

TijmenWierenga:
I ran into this issue as well yesterday and did some digging.

Anyone looking for a **temporary quick fix** can do the following:

1. Create a file that extends `Livewire\Mechanisms\HandleRequests\HandleRequests` in your project and overwrite the two following functions:
```php
class HandleRequests extends \Livewire\Mechanisms\HandleRequests\HandleRequests
{
    public function getUpdateUri(): string
    {
        return (string) str(
            route('livewire.update', [], false)
        )->start('/');
    }

    public function setUpdateRoute($callback): void
    {
        $route = $callback([parent::class, 'handleUpdate']);

        // Append `livewire.update` to the existing name, if any.
        if (! str($route->getName())->endsWith('livewire.update')) {
            $route->name('livewire.update');
        }
    }
}
```

2. Register the class as a singleton to replace the original mechanism in a service provider:
```php
$this->app->singleton(\Livewire\Mechanisms\HandleRequests\HandleRequests::class, HandleRequests::class);
```

-----------------------------------------------

### So what's happening?
When the update route is registered, a reference is stored in a protected variable called `$updateRoute`:

```php
// Livewire\Mechanisms\HandleRequests\HandleRequests

    {
        $route = $callback([self::class, 'handleUpdate']);

        // Append `livewire.update` to the existing name, if any.
        if (! str($route->getName())->endsWith('livewire.update')) {
            $route->name('livewire.update');
        }

        $this->updateRoute = $route;
    }
```

The value is later used to return a URL for the update route in `getUpdateUri`:

```php
function getUpdateUri()
    {
        return (string) str(
            route($this->updateRoute->getName(), [], false)
        )->start('/');
    }
```

When I dump the property value I found that the `Route` contains a reference to the `Router`. My assumption is that this prevents the `$updateRoute` property from being garbage collected.

-------------------------------------------

### Potential fix
The `$updateRoute` is only accessed once in the `getUpdateUri` method. It will only get the route name from the `Route` object. Therefore, storing the route name only would be sufficient to keep everything working:

```diff
function setUpdateRoute($callback)
    {
        $route = $callback([self::class, 'handleUpdate']);

        // Append `livewire.update` to the existing name, if any.
        if (! str($route->getName())->endsWith('livewire.update')) {
            $route->name('livewire.update');
        }

-       $this->updateRoute = $route;
+       $this->updateRoute->getName();
    }

function getUpdateUri()
    {
        return (string) str(
-            route($this->updateRoute->getName(), [], false)
+            route($this->updateRoute, [], false)
        )->start('/');
    }
```

This removes the reference to the `Route` in the class and removes the memory leak.

WayneHarris:
> Looks to be related: [laravel/framework#49334](https://github.com/laravel/framework/pull/49334)

This fix framework may help, https://github.com/laravel/framework/pull/49334

joshhanley:
Yep, upgrading Laravel to this Laravel version or later should fix it https://github.com/laravel/framework/releases/tag/v10.37.3

TijmenWierenga:
Yes, applied the patch yesterday which fixed the issue! Thanks!

PR #7184, opened 2023-10-29
update ComponentParser.php

spcbfr:
Hi everyone, this is my first ever code-related pull request to a library I use!

fix the `generateTestPathFromNamespace` function to apply base_path only after applying modifications to the namespace

when given an argument like so `'Tests\Feature\Livewire'` the function in question replace `\` with `/` and replaces the first `T` with a `t` so that the argument becomes `tests/Features/Livewire`.

the catch is that this is applied after applying the base_path to the namespace.

which means that if, for instance, your project root is `BloodyTwitter` like mine, it will set the test path `Bloodytwitter/Tests/Features/Livewire` (note the lowercase t in twitter and the uppercase Tests)

my pull request simply applies the base_path after those string manipulations are done

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/ComponentParser.php
+++ b/src/Features/SupportConsoleCommands/Commands/ComponentParser.php
@@ -218,8 +218,8 @@ class ComponentParser
 
     public static function generateTestPathFromNamespace($namespace)
     {
-        return str(base_path($namespace))
+        return base_path(str($namespace)
             ->replace('\\', '/', $namespace)
-            ->replaceFirst('T', 't');
+            ->replaceFirst('T', 't'));
     }
 }


joshhanley:
@spcbfr thanks for the PR! And congrats on the first PR, all PRs are much appreciated 😁

Can you explain why this PR is needed? What issue were you running into before this fix?

Also, in regards to your "catch", presumably, your goal is to convert the uppercase `Test` to lower case `test`, so you could first check if the string starts with "Tests", using `startsWith('Test')`, and only then convert it to lowercase "tests" using `replaceFirst('Tests', 'tests')` if it does.

If you could add a test for this too, that'd be great, thanks!

spcbfr:
@joshhanley

> Can you explain why this PR is needed? What issue were you running into before this fix?

I ran `php artisan livewire:make` inputted my component name and chose to create a test file with pest, it created my class and view normally but the Test file was created outside my project `BloodyTwitter`, instead it went ahead and made a new folder `Bloodytwitter` with a lowercase `t`.

here is a demo, showing the bug:
image

this is why this PR is needed.
> If you could add a test for this too, that'd be great, thanks!
Sure.

> using startsWith('Test'), and only then convert it to lowercase "tests" using replaceFirst('Tests', 'tests') if it does.
Okay, but the base_path still needs to be applied _after_ those string manipulations in case the project root has the word `Tests` in it.

joshhanley:
@spcbfr ah right, yep that makes sense. Thanks! Don't worry about adding a test (I didn't realise it was for the command), I don't think there are any tests are the moment for the commands.

Fix looks good as it is 🙂

mnaderian:
I wish all PRs had a nice explanation like this one.

PR #7185, opened 2023-10-30
Adding `assertSetModel()` / `assertNotSetModel()`

nexxai:
This uses Laravel's `->is(...)` method on models to do a comparison that isn't as strict as PHPUnit's default `assertSame()` method.

This is helpful for when you've set a public property as a particular model and need to ensure that the model you've set is the one you've meant to set.

If this PR is accepted, I will update the docs to reflect the new assertion types.


calebporzio:
To prevent expanding the testing API too far I'm going to close this. thanks though!

PR #7203, opened 2023-11-01
Additions related to the single-root requirement

kokoshneta:
This PR adds two sections related to the requirement that components must have a single root element:

- a warning on the _Quickstart_ page about the existence of the requirement (probably accidentally left out in the v3 docs, cf. #7187)
- a section on the _Components_ page (under _Full-page components_, after _Setting the page title_) to clarify that, despite the requirement, `<x-slot>`s for the containing layout file should go _outside_ the root element

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -561,6 +561,38 @@ public function render()
 }
 ```
 
+### Setting additional layout file slots
+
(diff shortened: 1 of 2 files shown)


PhiloNL:
Perfect, thanks @kokoshneta!

PR #7214, opened 2023-11-03
Fix bug when the first item in a collection is a new model

jplhomer:
This fixes a bug related to legacy binding of properties to Eloquent collections with models.

When adding a new model (not yet persisted) to a property which is an _empty_ Eloquent DB Collection, this error is observed:

```
InvalidArgumentException: Database connection [1] not configured.
```

Since the new model has a database connection of `null`, and it does not match the model's default connection name, we hit this line of code:

https://github.com/livewire/livewire/blob/main/src/Features/SupportLegacyModels/EloquentCollectionSynth.php#L29

The value of `$connection` is accidentally set to `true` rather than a string of the real connection due to missing parentheses.

```diff
-        if ($modelClass && $connection = $this->getConnection($target) !== $modelClass::make()->getConnectionName()) {
+        if ($modelClass && ($connection = $this->getConnection($target)) !== $modelClass::make()->getConnectionName()) {
```

The reason this was not caught before was because we use Sushi in the unit test, and Sushi always returns the model class for the connection name regardless of whether it's a "new" model.

I've updated the model in this test file to use testbench as the DB connection. Maybe there's a more elegant way to do it while keeping Sushi — let me know!

### DIFF
--- a/src/Features/SupportLegacyModels/EloquentCollectionSynth.php
+++ b/src/Features/SupportLegacyModels/EloquentCollectionSynth.php
@@ -26,7 +26,7 @@ class EloquentCollectionSynth extends Synth
         $meta['class'] = $class;
         $meta['modelClass'] = $modelClass;
 
-        if ($modelClass && $connection = $this->getConnection($target) !== $modelClass::make()->getConnectionName()) {
+        if ($modelClass && ($connection = $this->getConnection($target)) !== $modelClass::make()->getConnectionName()) {
             $meta['connection'] = $connection;
         }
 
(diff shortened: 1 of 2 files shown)


jplhomer:
@joshhanley :wave: following up from Discord. Here's some code which reproduces the issue (works in v2, fails in v3) https://github.com/jplhomer/livewire-repro-eloquent-new-model

jplhomer:
Ha, looks like @jasonlbeggs applied the same fix here: https://github.com/livewire/livewire/pull/7659/files

PR #7217, opened 2023-11-04
Support optional middleware.

selcukcukur:
The middleware does not always have to be web, its name can be changed. A new middleware option has been added to the configuration options to use it optionally.

An exception will occur if you delete or rename the default 'web' middleware group for any reason. To prevent this, this pull request makes the middleware option optional in the configuration file.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -30,16 +30,16 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Livewire Guard
+    | Livewire Route Middleware
     |--------------------------------------------------------------------------
     |
-    | Here you may specify the authentication guard Livewire will use while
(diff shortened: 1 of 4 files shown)


calebporzio:
Because you can update this middleware for most endpoints using "Livewire::setUpdateRoute(...)" (https://livewire.laravel.com/docs/installation#configuring-livewires-update-endpoint)

I'm going to close this. I understand the other still use "web" by default, but if we find that more people are running into this we can re-consider. thanks!

wojciechczerniak:
@calebporzio where do you track this? I would like to add myself to the list. My very first experience with new Livewire: digging into source code to find out that Laravel configurable options are hardcoded

PR #7222, opened 2023-11-05
Prefix data-uri to data-livewire_uri to avoid conflicts 

csnuknet:
Prefixs data-uri to data-livewire_uri to avoid conflicts with other uses of data-uri in the dom. 

Discussion raised https://github.com/livewire/livewire/discussions/7221

Yes.

Single change.

Existing tests would cover this use case.

Prevents conflict with other uses of the same data-uri attributes which can cause unintended side effects for livewire.


calebporzio:
All fixed -thanks

PR #7223, opened 2023-11-05
[3.x] Update testing docs for forbidden vs unauthorized

peterfox:
The docs are a bit unclear on authorization and authentication. It only mentions the Authentication assertion but alludes to it being for Authorization. The changes made are there to clear up possible confusion developers might have between the two when reading the livewire docs.

Hope this is helpful!


calebporzio:
Lovely, thanks!

PR #7236, opened 2023-11-07
Fix nested component listeners not working with multi-word events

joshhanley:
This PR fixes an issue with the nested component listeners, where multi-word events weren't working properly.

With the following event listener:
```blade
<livewire:child @my-event="$refresh" />
```

The `my-event` was being attached to the root element as `x-on:myevent`.
This PR changes it so it gets attached as `x-on:my-event`.

Hope this helps!


calebporzio:
Absolute dopeness. thanks @joshhanley

PR #7239, opened 2023-11-07
Bugfix/handle  rm   in complex synths

samlev:
When attempting to remove something that can't be explicitly typed (e.g. an untyped property or an array value), the frontend sends back the string `__rm__`. This is handled in `ArraySynth` by skipping hydration for non-array values, but other synths like `CarbonSynth`, `EnumSynth`, and `StdClassSynth` will attempt (and fail) to hydrate the string.

There have been a couple of discussions mentioning this ([App\Enums\TopicEnum::from(): Argument #1 ($value) must be of type int, string given](https://github.com/livewire/livewire/discussions/6801#top); [[V3] Difference in update hook when binding to array compared to V2](https://github.com/livewire/livewire/discussions/6665#top)), and I've built custom hydration steps to account for it, but it's a frustration.

Toy example:
```php
enum FooEnum: string { case foo = 'f'; case bar = 'b'; }

class Example extends Component {
    public $selected = [ FooEnum::foo, FooEnum::bar ];

    #[\Livewire\Attributes\Computed]
    public function cases() { return UnitSuit::cases(); }

    public function render() {
        return <<<'HTML'
            @foreach($this->cases() as $case)
                <input type="checkbox" wire:model.live="selected" value="{{ $case->value }}"> {{ $case->name }}
            @endforeach
            HTML;
    }
}
```

In this example, `$selected` will initially be dehydrated as: 
```json
{
  "selected" : [
    [
      ["f", {"s" : "enm", "class": "FooEnum"}],
      ["b", {"s" : "enm", "class": "FooEnum"}],
    ],
    {"s" : "arr"}
  ]
}
```
If you uncheck one of the checkboxes, it will send back something like:

```json
{
  "updates": {
    "selected.1": "__rm__"
  }
}
```

With the current code, it will attempt to hydrate `__rm__` with the previous enum synth, but it will fail with the message: `ValueError: "__rm__" is not a valid backing value for enum FooEnum`.

This update skips hydration if it detects the `__rm__` string, and properly unsets the property. It happens outside of the Synths, so it should work relatively consistently, no matter what synth is used.

### DIFF
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -211,6 +211,11 @@ class HandleComponents
 
         [$value, $meta] = $tuple;
 
+        // Nested properties get set as `__rm__` when they are removed. We don't want to hydrate these.
+        if ($this->isRemoval($value) && str($path)->contains('.')) {
+            return $value;
+        }
+
         $synth = $this->propertySynth($meta['s'], $context, $path);
 
         return $synth->hydrate($value, $meta, function ($name, $child) use ($context, $path) {
@@ -311,7 +316,7 @@ class HandleComponents
         // If this isn't a "deep" set, set it directly, otherwise we have to
         // recursively get up and set down the value through the synths...
         if (empty($segments)) {
-            if ($value !== '__rm__') $this->setComponentPropertyAwareOfTypes($component, $property, $value);
+            if (! $this->isRemoval($value)) $this->setComponentPropertyAwareOfTypes($component, $property, $value);
         } else {
             $propertyValue = $component->$property;
 
@@ -326,7 +331,6 @@ class HandleComponents
     protected function hydrateForUpdate($raw, $path, $value, $context)
     {
         $meta = $this->getMetaForPath($raw, $path);
-        $component = $context->component;
 
         // If we have meta data already for this property, let's use that to get a synth...
         if ($meta) {
@@ -399,7 +403,7 @@ class HandleComponents
             $toSet = $this->recursivelySetValue($baseProperty, $propertyTarget, $leafValue, $segments, $index + 1, $context);
         }
 
-        $method = ($leafValue === '__rm__' && $isLastSegment) ? 'unset' : 'set';
+        $method = ($this->isRemoval($leafValue) && $isLastSegment) ? 'unset' : 'set';
 
         $pathThusFar = collect([$baseProperty, ...$segments])->slice(0, $index + 1)->join('.');
         $fullPath = collect([$baseProperty, ...$segments])->join('.');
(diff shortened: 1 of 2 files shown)


samlev:
Just pushed an update to the PR because I re-checked and realised that `__rm__` will never be set for top-level properties of a component. It only gets set for nested properties, so this has been updated to only target sub properties (i.e. the path contains a `.`).

Test failures appear to be flaky browser tests.

src/Drawer/BaseUtils.php
```diff
@@ -20,6 +20,10 @@ static function isAPrimitive($target) {
             is_null($target);
     }
 
+    static function isRemoval($value) {
```
calebporzio, line 23: Can this live as a method in "handlecomponents" so that it doesn't bloat the utils? utils are kinda meant for more general utilities used in multiple places.

src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -211,6 +211,11 @@ protected function hydrate($valueOrTuple, $context, $path)
 
         [$value, $meta] = $tuple;
 
+        // Nested properties get set as `__rm__` when they are removed. We don't want to dehydrate these.
+        if (Utils::isRemoval($value) && str($path)->contains('.')) {
```
calebporzio, line 215: You mean "hydrate these"?

src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -412,7 +416,8 @@ protected function recursivelySetValue($baseProperty, $target, $leafValue, $segm
     protected function setComponentPropertyAwareOfTypes($component, $property, $value)
     {
         try {
-           $component->$property = $value;
+            // If the property is getting removed, unset it (making it null)
+            $component->$property = $value;
```
calebporzio, line 420: Sorry, is the comment applicable here? Seems like we're "setting" the property on line 420, but the comment says "unset it"

Maybe you're just saying that an __rm__'d value is being unset by it being set as null earlier in the codepath?

maybe a more descriptive comment would be helpful?

calebporzio (changes requested):
Thanks for the PR! This makes sense to not hydrate on removal, just a few small notes for you when you get a chance

src/Mechanisms/HandleComponents/HandleComponents.php
samlev: Oh, that's a remnant comment from an earlier revision. I'll clean it up.

src/Mechanisms/HandleComponents/HandleComponents.php
samlev: Yep. I'll clean that up.

src/Drawer/BaseUtils.php
samlev: Sure thing. I'll move it.

calebporzio:
Thanks @samlev!

PR #7247, opened 2023-11-08
Support static partials

calebporzio:


SRWieZ:
This look cool! I can already see what problems it will solve. I can also see people saying there components it's not working, it will need a good documentation 😄

calebporzio:
Yeah definitely will be a "sharp knife" that will need good docs

PR #7248, opened 2023-11-08
Fix Pest Scaffolding Behavior in `php artisan make:livewire` Command and Prioritize `--pest` Flag

manelgavalda:
Currently, when the `php artisan make:livewire` command is executed without any specific parameters, and you select the Pest option in the terminal, it generates the test using the PHPUnit scaffold. 

This PR addresses this behavior and ensures that Pest scaffolding is used as expected when Pest is chosen.

In addition, this PR enhances the command's behavior by giving precedence to the `--pest` flag over the `--test` flag. This means that even if both flags are provided, Pest scaffolding will be utilized, providing a more consistent and predictable experience.

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
+++ b/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
@@ -41,7 +41,7 @@ class MakeCommand extends FileManipulationCommand implements PromptsForMissingIn
         $force = $this->option('force');
         $inline = $this->option('inline');
         $test = $this->option('test') || $this->option('pest');
-        $testType = $this->option('test') ? 'phpunit' : 'pest';
+        $testType = $this->option('pest') ? 'pest' : 'phpunit';
 
         $showWelcomeMessage = $this->isFirstTimeMakingAComponent();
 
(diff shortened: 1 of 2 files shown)


src/Features/SupportConsoleCommands/Commands/MakeCommand.php
```diff
@@ -108,7 +108,7 @@ protected function createView($force = false, $inline = false)
         return $viewPath;
     }
 
-    protected function createTest($force = false, $testType = 'phpunit')
+    protected function createTest($force = false, $testType)
```
manelgavalda, line 111: it is a ternary operator so no need to have a default value

src/Features/SupportConsoleCommands/Tests/MakeCommandUnitTest.php
```diff
@@ -39,7 +39,7 @@ public function component_test_is_created_by_make_command_with_test_option()
     /** @test */
     public function component_pest_test_is_created_by_make_command_with_pest_option()
     {
-        Artisan::call('make:livewire', ['name' => 'foo', '--pest' => true]);
+        Artisan::call('make:livewire', ['name' => 'foo', '--test' => true, '--pest' => true]);
```
manelgavalda, line 42: Before if both flags were present it created the test with the PHPUnit scaffold, and now if the `--pest` flag is present it will create the test with the current scaffold even if the `--test` flag is not there

calebporzio:
Great, thanks!

PR #7250, opened 2023-11-08
Create PHPStan extension for computed properties

calebdw:
Hello!

This PR creates a [PHPStan](https://phpstan.org/) extension for Livewire---it currently only supports computed properties but other functionality can be added as necessary. The extension supports [auto-installation](https://phpstan.org/user-guide/extension-library#installing-extensions) but can be manually included as well, depending on the user's configuration.

Some might question if this should be a separate package, however, including the PHPStan extension for a library's special functionality provides a much better developer experience. For example, consider the Carbon library which provides a [PHPStan extension for their macros](https://github.com/briannesbitt/Carbon/blob/master/extension.neon) out-of-the-box without having to install a separate phpstan package.

Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
     "orchestra/testbench": "^8.0",
     "orchestra/testbench-dusk": "^8.0",
     "calebporzio/sushi": "^2.1",
-    "laravel/prompts": "^0.1.6"
+    "laravel/prompts": "^0.1.6",
+    "phpstan/phpstan": "^1.10"
   },
   "autoload": {
     "files": [
@@ -49,6 +50,11 @@
       "aliases": {
         "Livewire": "Livewire\\Livewire"
       }
+    },
+    "phpstan": {
+      "includes": [
+        "extension.neon"
+      ]
     }
   },
   "minimum-stability": "dev",
--- a/extension.neon
+++ b/extension.neon
@@ -1,0 +1,4 @@
+services:
+    -
+        class: Livewire\PHPStan\Properties\ComputedPropertyExtension
+        tags: [phpstan.broker.propertiesClassReflectionExtension]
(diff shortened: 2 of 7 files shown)


levu42:
@CalebDW This only works for computed properties with the `#[Computed]` attribute, doesn't it? It doesn't cover the `getFooProperty()` old syntax - isn't that still supported in v3?

Other than that, it looks really cool!

CalebDW:
@levu42, yes only the `#[Computed]` attribute syntax is supported---while v3 might support the older syntax there's no mention of it in the v3 docs so I would consider it deprecated.

nuernbergerA:
Is there any reason why this can't be a package?

levu42:
Since Livewire is now officially a Laravel project, could this be integrated into Larastan?

CalebDW:
@nuernbergerA, no this could be a package. The benefit, however, for Livewire including support for its magic out of the box is that everything Just Works™ without the user having to search out and install a separate package

@levu42 not sure, that would be up to the maintainers (@nunomaduro, @canvural)---previously they had stated that Livewire was not supported. However, I also don't think that it should be on Larastan to support every Laravel package

lrljoe:
> @nuernbergerA, no this could be a package. The benefit, however, for Livewire including support for its magic out of the box is that everything Just Works™ without the user having to search out and install a separate package
> 
> @levu42 not sure, that would be up to the maintainers (@nunomaduro, @canvural)---previously they had stated that Livewire was not supported. However, I also don't think that it should be on Larastan to support every Laravel package

Not an official opinion in any way, shape or form, but to my mind, a plugin for PHPStan/LaraStan instead might be cleaner?  I think introducing more maintenance for the core team, is just going to cause them pain in the long run!

joshhanley:
@calebdw thanks for the PR! I think this would be best as an external package for now. So going to close this. Thanks for being willing to contribute 🙂

calebdw:
Hi folks! 

I finally got around to publishing this as a package, so just wanted to follow up in case anyone was interested!

https://github.com/calebdw/larastan-livewire

@joshhanley, do you think it would be possible to link to this so others can easily find it?

PR #7257, opened 2023-11-09
Add failing test for wire:transition

SRWieZ:
TLDR: When using wire:transition; **transitions-out** does not work or it works but transition-out siblings too. transition-in works great.

**Context:** 
I'm building an app that compose a script dynamically with a form next to it. The goal was to make transitions when lines are added (or removed). So I cut each lines on an array and wrapper them with `<span wire:key wire:transition></span>`  so the user can see differences.

**What I discovered:** 
First, don't give too much importance at the name of my tests, I'm not good at naming things. sorry
I made a lot of tests before finding what's may cause the issues but I made some progress.

All 3 tests are supposed to show this code:
```
code
    indented code
/code
```

Here are the differences between them and their effect
1. `indented code` is removed from the array. **Effect:** `indented code` is removed ✅ but did not transition out ❌
2. `/code` is removed from the array.  **Effect:** `/code` goes 1 character to the right ❌ then transition out ✅ with all the 3 lines ❌
3. `/code` is removed from the array. **Effect:** `/code` is removed ✅ but did not transition out ❌

1 and 2 are strictly the same component, but for the 3rd one I lied a bit, it actually show `code    indented code/code` at the initial request.

Here is all the things I have discovered
- Wrapping each line with `@if(true) @endif` doesn't change the behaviour
- Removing `@foreach` and echoing a string imploded with \n instead does not change the behaviours
- In the 2nd test, all the code is transitioned out but in my app with more lines, that "self-dissapear and transitioning out sibling" behaviour is way more "random" than that. Ex: removing line 3 and 8 from the array, transition out line 1, 3 and 8. By random I mean "don't know why these line and not the others", it's actually the same behaviour each FIRST time I remove the same lines.
- Something I can't reproduce in test, but as an example some other weird thing can happen: 
    - Take an 8 lines code, slightly change the content of 1 and 2, remove 3, 4, 5 and 8, keep 6 and 7.
    - **Effect:** 3,4,5 removed without transition ❌, 8 transitioning out ✅, 1 is not morphed correctly ❌.
- Another weird example:
   - Remove a line, transition out ✅ but with 2 other lines with him ❌. Then put it back, all 3 are back. Then remove the same line. Removes it without transitioning ❌ and without the 2 others ✅.
- transition:in works every time. So out is the problem. 
- this transition bug can mess up with other morph on other part of the Livewire component (not removing element that are not int he dom anymore, essentially)

**How to fix?** 
Remove all the wire:transition and everything works again! Kidding, I really don't know.

That's my first failing test and I'm not really sure how to write them.
I almost bailed out of this PR because I wasn't sure if I could explain it. Hope it's understandable!


SRWieZ:
Hi! I made a little wirebox to show you the bug.
https://wirebox.app/b/4l5dx

A simple todo example, removing the last element causes the transition bug.

joshhanley:
@SRWieZ not sure what is going on, but the failing tests aren't the ones you submitted. So I have merged the main branch in to see if that makes a difference.

joshhanley:
@SRWieZ looks like the correct tests are failing now 🙂

SRWieZ:
Honestly forgot about this PR. What a long description 😅 

Thanks @joshhanley for merging to main. Seems that's transition bug is still there.

@calebporzio I know transitions are hard and I consider this PR a low priority one. I appreciate the work you put in Livewire, don't let this one take your focus off other thing 😄

Have a nice holidays! To both of you

calebporzio:
Ok, I did some digging on this and it's actually not a simple problem to fix by any means.

The problem is:

* You have a list of elements that are keyed with `wire:key`
* One of the them is removed
* When "morph" "removes" that element, it keeps it in a "holdover" array because it has a key and might be used later on
* Because it's never actually "removed", it never has a chance to transition

The way way keyed items works in Livewire right now doesn't allow for dynamic list transitions like this.

I added a "limitations" section to the wire:transition documentation to note this for people that they can't use it with wire:key in lists.

I think we'll keep wire:transition scoped to single elements inside Blade conditionals. Thank you for the report.

PR #7260, opened 2023-11-09
Use #[Validate] instead of #[Rule]

calebporzio:
There are two problems that led to this decision.

**Problem #1:**
The `#[Rule]` attribute class (`Livewire\Attributes\Rule`) conflicts with Laravel validation rule objects (`Illuminate\Validation\Rule`). This leads to people needing to alias one of them when importing which is confusing.

**Problem #2:**
#[Rule] attributes are restrictive for adding validation because they don't support run-time PHP code like Laravel rule objects.

Therefore, most times people are using `rules()` methods instead to declare their validation rules.

However, there are still times where it's nice to use the real-time validation capability of `#[Rule]` objects. You can use a bare `#[Rule]` attribute for this behavior (after a change in this PR), however, it reads oddly.

**Solution:**
Change `#[Rule]` to `#[Validate]` and make `#[Validate]` support not passing in rules and only using it to declare a field as real-time validating like so:

```php
#[Validate]
public $username = '';

public function rules()
{
    return ['username' => Rule::unique('users')->ignore($this->post)];
}
```

**Backwards compatibility:**
We will retain `#[Rule]` and `#[BaseRule]` in the codebase to prevent breaking apps until the next major release.

However, I updated all documentation to move to `#[Validate]` instead of `#[Rule]` and added notes about using Rule objects in Livewire to better educate people.

I also added a docs note about `#[Rule]` so that it still shows up in algolia search and SEO for people with old codebases or knowledge searching for documentation.

In the future we should probably re-work the docs a bit more to put less emphasis on `#[Validate]` in favor of using form objects and the `rules()` method.

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -57,9 +57,9 @@ After the new post is created in the database, we redirect the user to the `Show
 
 To avoid storing incomplete or dangerous user input, most forms need some sort of input validation.
 
-Livewire makes validating your forms as simple as adding `#[Rule]` attributes above the properties you want to be validated.
+Livewire makes validating your forms as simple as adding `#[Validate]` attributes above the properties you want to be validated.
 
-Once a property has a `#[Rule]` attribute attached to it, the validation rule will be applied to the property's value any time it's updated server-side.
+Once a property has a `#[Validate]` attribute attached to it, the validation rule will be applied to the property's value any time it's updated server-side.
 
 Let's add some basic validation rules to the `$title` and `$content` properties in our `CreatePost` component:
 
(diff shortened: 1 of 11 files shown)


lrljoe:
Once merged in, I'll review my tweak for
validateOnly()
validate() 

And see if still relevant.

I have been tinkering with validateOnly(field) to look at validateOnly(field|array) in my limited downtime, as I think being able to pass in an array there can be handy.
Also considered
validateSome(array) which frankly may be more friendly all round. Thoughts @calebporzio ?

calebporzio:
Can you provide a link to the tweak you're talking about?

CyberPunkCodes:
@calebporzio is it possible to allow just `#[Rule]` ? Using it like that throws errors. So does `#[Rule()]`. We have to use `#[Rule('')]` or `#[Rule([])]` to not throw an error.

Use case is when we are using forms and use both the attribute and rules method combined to get live validation on more complex rules. Having to pass `('')` or `([])` visually looks tacky and would be cleaner if we could just use `#[Rule]`.

chillbram:
@CyberPunkCodes As `#[Rule]` is now deprecated, you should just use `#[Validate]`. The only reason Rule isn't gone yet is to prevent breaking changes, but Rule probably won't be getting any changes.

CyberPunkCodes:
@chillbram I am aware. I was asking about a bare `#[Validate]` attribute, without anything extra, not even the parentheses `()`. Before, you had to include the parentheses and either an empty string or array, like `#[Rule('')]`. I have tested and `#[Validate]` without anything extra is now supported and documented nicely.

Very nice update @calebporzio 👍 So much cleaner!

PR #7263, opened 2023-11-10
Fix: form submit accessibility

SRWieZ:
TLDR: Submitting a form now set a `data-submitting` attribute on the `form` instead of `disabled` on buttons.

In the same week I saw your tweet and an interesting article.
- Your tweet about modal accessibility https://twitter.com/calebporzio/status/1722671639622504649
- And a article on the Stack Overflow newsletter about accessibility on forms https://gomakethings.com/dont-disable-buttons/

So I figured that you will be interested on a fix 🙂 You may want to refactor this PR a little bit to contain the feature in `supportDisablingFormsDuringRequest.js`. Don't know the best way to do this, I let you decide.

Don't know if it's considered a breaking change since it is a fix. But this will change the look of a form submitting since classes like `disabled:opacity-50` won't work anymore. I didn't removed the readonly attributes on input, only the disabled.


SRWieZ:
Fair enough 🙂 I didn't take the time to test it with voice over. Have a good day!

PR #6715, opened 2023-09-08
Use session errors as default error bag

yungifez:
possibly fixes

#6586
#6546

Review the contribution guide first at: https://laravel-livewire.com/docs/2.x/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?
Yes 

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)
Yes

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
No 

4️⃣ Does it include tests? (Required)
No tests 

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.
see #6546

Thanks for contributing! 🙌

PR #7265, opened 2023-11-10
Fix error forwarding / sharing

austenc:
**Scenario:**
You are on a page with a full-page Livewire component.
That component submits a form **without** Livewire to a POST endpoint. 
Validation fails and Laravel redirects back (with errors) to the full-page Livewire component. 

**In v2 the errors would get shared from the shared view error bag correctly, but in v3 they do not.** 

This scenario seems a bit strange I know, but occurred for us with a couple routes that needed to use Form Requests. It seems that others have also encountered this in https://github.com/livewire/livewire/pull/6715. I opened a discussion about it here, too: https://github.com/livewire/livewire/discussions/7189

Laravel handles forwarding errors from the session to the view with the `ShareErrorsFromSession` middleware. This PR initializes the component's error bag by looking at the shared error bag first, just like we did in v2.

Thanks for checking it out! ✌️


wojciechczerniak:
Been there, this only works half way. `HandlesValidation` works on a single `MessageBag`. This will fail if your `FormRequest`s use named `ErrorBag` instead default. To make Livewire compatible with Laravel you have to look into rendering hook that assigns the `ViewErrorBag`, as `ShareErrorFromSession` does.

https://github.com/livewire/livewire/blob/ddae28540f94a82d1ec271eee4820fabc4a1da7e/src/Features/SupportValidation/SupportValidation.php#L21-L23

First step would be to keep all the `MessageBag`s by getting whole `ViewErrorBag` and shadow only the `default` one with component errors.

```diff
- $errors = (new ViewErrorBag)->put('default', $this->component->getErrorBag()); 
+ $errors = (session()->get('errors') ?: new ViewErrorBag)->put('default', $this->component->getErrorBag()); 
  
$revert = Utils::shareWithViews('errors', $errors); 
```

Next is to use Laravel `default` bag if component `getErrorBag()` returns an empty one.

austenc:
@wojciechczerniak The goal with this PR is to fix a corner case where v3 isn't behaving as v2 did.

Supporting named error bags is out of scope for what I intended to do here. Seems like that should be its own (additional) pull request.

calebporzio:
Good call, thanks Austen!

PR #7266, opened 2023-11-10
Isolate form object validation from root livewire component

calebporzio:


richardkeep:
PR Looking solid! 👍

I love all the endless refactoring we did! 💪

calebporzio:
Only final concern with this implementation:

When `$this->validate()` is run from the root component, currently in this PR, it first runs all the child form object's validations.

What's the best behavior here?

Right now, if a form object validation fails it will stop continuing validation.

Is that the best?

If you have two form objects and the first fails, it would prevent the second from running so the user would only see validation errors from the first one?

we could easily merge all the error bags and throw them together? but that feels a little icky.

also "bail" wouldn't work as expected.

I think I like what we currently have because of simplicity. Maybe the only change I would make is to run form object validations AFTER the root component has run it's own validation.

Yeah?

nuernbergerA:
To avoid a BC we should merge errors.

It would also be bad to have an error fill the field and then get another error from the component (i think this will confuse ppl)

calebporzio:
Yeah, good point. This is a tough one though.

Pro of running all validators and combining the error bags:
* All errors would show up on the page when validation fails (even from multiple form objects and base component errors)

Cons:
* We would also have to combine "validator" instances somehow so that the validation error that gets thrown has access to a single validator with a single error bag. I'm not really sure how this would even work? like the validator would only have the data of one of the validators? but the messages of all of them?
* "bail" in one validator wouldn't bail the rest. Probably not a big deal either way.

I suppose I'm in favor of this behavior but I can't visualize how to solve for the first "con".

Any ideas?

PR #7267, opened 2023-11-10
Fix Documentation Examples for the `assertViewHas` method

manelgavalda:
This PR addresses an issue with the `assertViewHas` method in the documentation. The current examples demonstrate the use of expectations that leads to a `Failed asserting that null is true` error. The root cause is that the `assertViewHas` closure expects to return a boolean.

This PR updates the examples in the documentation to align with the usage outlined in the Laravel documentation [assetViewHas method](https://laravel.com/docs/10.x/http-tests#assert-view-has).

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -120,7 +120,7 @@ class ShowPostsTest extends TestCase
 
         Livewire::test(ShowPosts::class)
             ->assertViewHas('posts', function ($posts) {
-                $this->assertEquals(2, count($posts));
+                return count($posts) == 2;
             });
     }
 }
@@ -169,7 +169,7 @@ class ShowPostsTest extends TestCase
         Livewire::actingAs($user)
             ->test(ShowPosts::class)
             ->assertViewHas('posts', function ($posts) {
-                $this->assertEquals(3, count($posts));
+                return count($posts) == 3;
             });
     }
 }


manelgavalda:
Yes, this could be done, but I think it is an unnecessary step to do it inside the method. You can just take the property from the `Testable` and do it:
```php
    $testable = Livewire::test(ShowPosts::class)
        ->assertviewhas('posts', function ($posts) {
            return count($posts) == 2;
    });
    
   //
   
    $this->assertEquals(2, count($testable->posts));

```

This process is very similar to extracting a property from the `response` (in this case `testable`), as described in the Laravel documentation [assertViewHas](https://laravel.com/docs/10.x/http-tests#assert-view-has). Perhaps it's also a good idea to consider adding some documentation about this in the Livewire docs so it is aligned with the Laravel docs.

> In addition, view data may be accessed as array variables on the response, allowing you to conveniently inspect it:

```php
$this->assertEquals('Taylor', $response['name']);
```

Also, I think placing multiple assertions like this looks very clean:

```php
$testable = Livewire::test(ShowPosts::class)
    ->assertViewHas('posts', function ($posts) {
        return count($posts) == 2
            && $posts->first()->title == 'First Post'
            && $posts->first()->description == 'Post description';
    });
});
```

robsontenorio:
With this approach you write standard asserts like usual.

```php
->assertViewHas('posts', function ($posts) {
       $this->assertXXXX(...);
       $this->assertXXXX(...);
       $this->assertXXXX(...);
       
       // This is the new thing.
       return true;
    });
```

manelgavalda:
Yes, you can do it, but the `assertViewHas` method is not doing anything here, it is just returning true. So it would be the same as doing this

```php
    $testable = ...->assertViewHas('posts', function ($posts) {
       return true;
    });
    
    $this->assertXXXX(.., $testable->posts);
    $this->assertXXXX(.., $testable->posts);
    $this->assertXXXX(.., $testable->posts);
```

robsontenorio:
If assert fails, the test fails. Btw, it just a option.

manelgavalda:
Yeah, initially, I was also returning `true` to make it pass, so I believe it could be a good readability option to use it as a wrapper for other assertions. 👍 

It would be valuable to hear what others think about this change

manelgavalda:
An advantage of using a boolean instead of expectations is that you can leverage arrow functions:

```php
Livewire::test(ShowPosts::class)
    ->assertViewHas('posts', fn ($posts) => 
        count($posts) == 2
            && $posts->first()->title == 'First Post'
            && $posts->first()->description == 'Post description'
    );
});
```

joshhanley:
@manelgavalda thanks! Yep I agree, it should match Laravel's documentation.

PR #7270, opened 2023-11-12
[WIP] Add "Describable Enums" feature

calebporzio:


woodspire:
Question: Is it possible to support multiple languages with the enum description ?

calebporzio:
Mmk, I'm going to hold off on this for now. All context is in this twitter thread: https://twitter.com/calebporzio/status/1723704970732568968

Maybe I'll revisit this later. Going to leave the PR in WIP in case anyone else has input.

levu42:
So basically the `->description()` method in a trait just gives the value of the `#[Description('...')]` attribute on that case? That feels to me like it's a nice-to-have feature but it's Livewire independent.

I'm using backed enums a lot in my LW3 application and I just always do a method

```
public function description(): string
{
  return match($this) {
    self::Foo => 'Foo',
    self::Bar => 'Bar',
    self::FooBar => 'Foo & Bar',
  }
}
```

Given that people already ask about extensions like translations - is this really something that Livewire should solve in the core? Isn't it better just a package, given that it is totally standalone? Or is there any connection to core code that I don't see right now?

calebporzio:
Killing this one, no plans for it anymore

PR #7277, opened 2023-11-14
Add separate CannotUpdateLockedPropertyException

chillbram:

No

Yes

No

Yes

This PR creates a new Exception class used by the `#[Locked]` attribute. The functionality is unchanged, but this allows people to more easily scope to this exception. An example where this is useful is in writing regression tests, ensuring that properties thought locked stay locked. Rather than hoping that the general `\Exception` class is referring to the correct error or testing to the exception message which might be changed in future, being able to write `$this->expectException(CannotUpdateLockedPropertyException::class)` is much easier and more legible.

### DIFF
--- a/src/Features/SupportLockedProperties/CannotUpdateLockedPropertyException.php
+++ b/src/Features/SupportLockedProperties/CannotUpdateLockedPropertyException.php
@@ -4,7 +4,7 @@ namespace Livewire\Features\SupportLockedProperties;
 
 class CannotUpdateLockedPropertyException extends \Exception
 {
-    public function __construct($property)
+    public function __construct(public $property)
     {
         parent::__construct(
             'Cannot update locked property: ['.$property.']'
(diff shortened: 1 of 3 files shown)


src/Features/SupportLockedProperties/CannotUpdateLockedPropertyException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Livewire\Features\SupportLockedProperties;
+
+class CannotUpdateLockedPropertyException extends \Exception
+{
+    public function __construct($property)
```
levu42, line 7: is it possible to do `__construct(public $property)` to store the property in a way that it can be used without parsing the exception message afterwards?

chillbram: Done!

calebporzio:
Good call - thanks!

PR #7291, opened 2023-11-15
Support dynamic SEO data for full page components.

selcukcukur:
### About
This pull request makes it possible to set dynamic data that is truly search engine friendly. In doing so, it is free of incompatibilities with previous versions and supports different types of usage with the macro included in the attribute or view file.

It is a very good helpful pull request for full page components.

### Example

When a use is made as in the example below, data such as title and description will be automatically defined in meta tags such as Twitter and Opengraph. If you wish, it is also possible to set these values optionally.

```php
<?php

namespace App\Livewire;

use Livewire\Attributes\Seo;
use Livewire\Component;

class CreatePost extends Component
{
	// ...

	#[Seo('Page Title', 'Page description', 'keywords', 'author')] // [tl! highlight]
	public function render()
	{
	    return view('livewire.create-post');
	}
}
```

Once the title, description and image are set, all the remaining data will be assigned automatically and will be available for use as in the example below.

```blade
<head>
    <title>{{ $title ?? 'Page Title' }}</title>

    <!-- META -->
    <meta name="title" content="{{ $title ?? 'Page Title' }}" />
    <meta name="description" content="{{ $description ?? 'Page description' }}" />
    <meta name="keywords" content="{{ $keywords ?? 'page,keyword' }}" />
    <meta name="robots" content="{{ $robots ?? 'noindex, nofollow' }}" />
    <meta name="author" content="{{ $author ?? 'author' }}" />
    <!-- META -->

    <!-- OPENGRAPH -->
    <meta property="og:type" content="{{ $ogType }}" />
    <meta property="og:site_name" content="{{ $ogSitename }}" />
    <meta property="og:title" content="{{ $ogTitle }}" />
    <meta property="og:description" content="{{ $ogDescription }}" />
    <meta property="og:image" content="{{ $ogImage ?? '' }}" />
    <meta property="og:locale" content="{{ $ogLocale }}" />
    <meta property="og:url" content="{{ $ogUrl }}" />
    <!-- OPENGRAPH -->

    <!-- TWITTER -->
    <meta name="twitter:card" content="{{ $twitterCard }}" />
    <meta name="twitter:site" content="{{ $twitterSite }}" />
    <meta name="twitter:title" content="{{ $twitterTitle }}" />
    <meta name="twitter:description" content="{{ $twitterDescription }}" />
    <meta name="twitter:image" content="{{ $twitterImage }}" />
    <meta name="twitter:author" content="{{ $twitterAuthor }}" />
    <meta name="twitter:creator" content="{{ $twitterCreator }}" />
    <meta name="twitter:url" content="{{ $twitterUrl }}" />
    <!-- TWITTER -->

	// ...
</head>
```

**Notice :** This pull request does not contain any incompatibilities with previous versions. Apart from this, if there are any improvements that can be made, of course I can listen to these suggestions and make updates.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -498,33 +498,64 @@ public function render()
 }
 ```
 
-### Setting the page title
+### Setting the page SEO
 
-Assigning unique page titles to each page in your application is helpful for both users and search engines.
+Defining specific meta tags for each page in your app is useful for search engines.
 
-To set a custom page title for a full-page component, first, make sure your layout file includes a dynamic title:
+To set custom metadata for a full page component, first make sure your layout file contains the necessary keys for a dynamic data:
 
 ```blade
 <head>
     <title>{{ $title ?? 'Page Title' }}</title>
+
+    <!-- META -->
+    <meta name="title" content="{{ $title ?? 'Page Title' }}" />
+    <meta name="description" content="{{ $description ?? 'Page description' }}" />
+    <meta name="keywords" content="{{ $keywords ?? 'page,keyword' }}" />
+    <meta name="robots" content="{{ $robots ?? 'noindex, nofollow' }}" />
+    <meta name="author" content="{{ $author ?? 'author' }}" />
+    <!-- META -->
+
+    <!-- OPENGRAPH -->
+    <meta property="og:type" content="{{ $ogType }}" />
+    <meta property="og:site_name" content="{{ $ogSitename }}" />
+    <meta property="og:title" content="{{ $ogTitle }}" />
+    <meta property="og:description" content="{{ $ogDescription }}" />
+    <meta property="og:image" content="{{ $ogImage ?? '' }}" />
+    <meta property="og:locale" content="{{ $ogLocale }}" />
+    <meta property="og:url" content="{{ $ogUrl }}" />
+    <!-- OPENGRAPH -->
+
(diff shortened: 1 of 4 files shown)


levu42:
I do not think this should be part of livewire core, that's perfect for a plugin.

vlados:
I agree that this is something that is nicety, not so important to be in the core
You can always use https://github.com/artesaos/seotools

lrljoe:
Adding my 2p/2c in.

I am really not convinced that adding in something that extensive is really a core thing? 

For starters, you already have clashes with the existing Title attribute that people are using in production environments.

Then based on the codebase, you're suggesting that Livewire needs to natively/actively support and update core code whenever Twitter / OpenGraph updates their approach, which introduces headaches all round.

It feels like there's a breaking change just waiting to happen here.  Definitely think that a package/plugin approach for this is smoother, and definitely check out @vlados comment, as that seems to do what you're suggesting and more.

The more "bloat" introduced, the more painful it becomes to:
- Support
- Update
- Maintain
- Load

PhiloNL:
I agree with the sentiment that this should probably be a third-party package. I don't think the Livewire core should get involved with how people should approach SEO. But let's see what @calebporzio & @joshhanley think 😄

PR #7292, opened 2023-11-16
Fix: Resolve JavaScript Asset Retrieval in Livewire

Fredwared:
Resolved an issue in Livewire where the designated route path was not correctly returning the primary JavaScript asset as a file. This problem was particularly evident in local Windows environments using OpenServer (Apache, Nginx). The issue, often manifesting as a 404 error, occurred when routes ended with a dot (.). The fix implemented now allows modification without overriding and binding a new service provider.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -156,4 +156,15 @@ return [
     */
 
     'pagination_theme' => 'tailwind',
+
+    /*
+    |---------------------------------------------------------------------------
+    | Asset url
+    |---------------------------------------------------------------------------
+    |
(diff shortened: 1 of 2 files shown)


calebporzio:
The "asset_url" has been removed in favor of configuring at runtime: https://livewire.laravel.com/docs/installation#customizing-the-asset-url

marwandhiaurrahman:
i have add
```
    return Route::get('/custom/livewire/livewire.js', $handle);
});
```
but still not working, the asset still in default livewire

PR #7303, opened 2023-11-17
Missing documentation on rendering components

MrWeb:
Coming from V2 one knows how to render components, but in V3 I cannot find anywhere how to render a single component in a blade file.


joshhanley:
@MrWeb thanks for the PR! The docs contain how to render a component here https://livewire.laravel.com/docs/components#rendering-components

PR #7304, opened 2023-11-17
Add layout method

lucascardial:
**About**

This pull request makes it possible to obtain the layout from a `layout()` method. This is especially useful when you have different views of the application: admin, financial, buyer... And so instead of defining `#[Layout()]`  in each component class, it becomes possible to define base classes that provide the same layout.

**Example**
```php
<?php
use Livewire\Component;

abstract class AdminComponentBase extends Component
{
    public function layout()
    {
        return new BaseLayout('layouts.admin-layout');
    }
}

class MonthlyBilling extends AdminComponentBase
{
    public function render()
    {
        return view('livewire.admin.monthly-billing');
    }
}


class ActivePayers extends AdminComponentBase
{
    public function render()
    {
        return view('livewire.admin.active-payers');
    }
}
```
This way, all the components in the admin context inherit from the AdminComponentBase class, avoiding code duplication.

Same here:

```php
<?php
use Livewire\Component;

abstract class BuyerComponentBase extends Component
{
    public function layout()
    {
        return new BaseLayout('layouts.buyer-layout');
    }
}

class ShopList extends BuyerComponentBase
{
    public function render()
    {
        return view('livewire.buyer.shop-list');
    }
}

class ShopCart extends BuyerComponentBase
{
    public function render()
    {
        return view('livewire.buyer.shop-cart');
    }
}
```

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -76,7 +76,7 @@ abstract class Component
         store($this)->set('skipHydrate', true);
     }
 
-    function getLayout(){
+    function layout(){
         return null;
     }
 
--- a/src/Component.php
+++ b/src/Component.php
@@ -76,6 +76,10 @@ abstract class Component
         store($this)->set('skipHydrate', true);
     }
 
+    function getLayout(){
+        return null;
+    }
+
     function __isset($property)
     {
         try {
--- a/src/Features/SupportPageComponents/SupportPageComponents.php
+++ b/src/Features/SupportPageComponents/SupportPageComponents.php
@@ -90,7 +90,7 @@ class SupportPageComponents extends ComponentHook
         // Only run this handler once for the parent-most component. Otherwise child components
         // will run this handler too and override the configured layout...
         $handler = once(function ($target, $view, $data) use (&$layoutConfig, &$slots) {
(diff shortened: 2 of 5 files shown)


PerryvanderMeer:
I would prefer `layout()`, just as you can use `$rules` and `rules()`

lucascardial:
> I would prefer `layout()`, just as you can use `$rules` and `rules()`

I agree

lrljoe:
Suggest exercising caution here.  Livewire uses a Macro to make "layout()" available on the view() method, so it may create some confusion here if there's two methods of the same name, that do roughly similar things.

joshhanley:
@lucascardial wouldn't it be better to add support for `#[Layout()]` in a parent class? Not sure how easy/hard that will be, but that would be better in my opinion that adding a new method called `layout()`.

calebporzio:
I think for now let's just make #[Layout] work in a parent class and start there. That solves most of the problem I think

calebporzio:
I made layout work on parents here: https://github.com/livewire/livewire/pull/7753

Closing this PR

PR #7313, opened 2023-11-19
fix docs v3 in components.md

imranwijaya:
in the [components docs](https://livewire.laravel.com/docs/components), the quote block is not rendered properly, as it should render ```Using @script and @assets inside Blade components```

image

Changing `@script` to `@verbatim@script@endverbatim` might fix this (similar to #6534)

This is my first pull request, so I apologize if I'm doing it incorrectly.


calebporzio:
Thanks!

PR #7327, opened 2023-11-20
Data store has() with non-existent array key

danharrin:

Just a bug fix

Yes

No

There are no tests for the Data Store yet.

When the `$iKey` key does not exist in the array, `Undefined array key` is thrown. I think this is because `!! $this->lookup[$instance][$key][$iKey]` is evaluated first, before `?? false`, so the null catch does not happen early enough.

Thanks for contributing! 🙌

### DIFF
--- a/src/Mechanisms/DataStore.php
+++ b/src/Mechanisms/DataStore.php
@@ -42,7 +42,7 @@ class DataStore
         }
 
         if ($iKey !== null) {
-            return !! $this->lookup[$instance][$key][$iKey] ?? false;
+            return !! ($this->lookup[$instance][$key][$iKey] ?? false);
         }
 
         return true;


src/Mechanisms/DataStore.php
```diff
@@ -42,7 +42,7 @@ function has($instance, $key, $iKey = null) {
         }
 
         if ($iKey !== null) {
-            return !! $this->lookup[$instance][$key][$iKey] ?? false;
+            return !! ($this->lookup[$instance][$key][$iKey] ?? false);
```
chillbram, line 45: Unrelated, but it would be nice to use explicit casting here. So instead of `!!` using `(bool)` to return a boolean

danharrin: I agree, but I don't want to go changing code style unless I'm asked to

chillbram: Understandable. As this type of casting isn't used elsewhere in Livewire PHP code I thought it uncontroversial enough to bring up :)

danharrin: Let's see what Caleb wants :)

calebporzio:
Thanks Dan! And agreed re: keeping the double bang

gregorysouzasilva:
Tks for this fix, this was delaying my upgrade. I was trying to fix it just on my side but it wasn't working.

PR #7337, opened 2023-11-21
Add `assertNoFileDownloaded()` test method

SjorsO:
This PR adds an `assertNothingDownloaded` test method to complement the existing [`assertFileDownloaded`](https://livewire.laravel.com/docs/downloads#testing-file-downloads) method.

My use-case is that one of my components has a method like this:

```php
public function zipDownload()
{
    if (Auth::user()->is_premium) {
        return response()->download($this->model->zip_file_path);
    }
}
```

I can use `assertFileDownloaded` to test that premium users are able to download the file, but there currently isn't a way to check that non-premium users aren't able to download the file.


joshhanley:
@SjorsO thanks for the PR! We've renamed it to `assertNoFileDownloaded()` and added some docs 🙂

PR #7359, opened 2023-11-22
Fix Hydrating Eloquent Collection With Duplicate ID's

ChristopherCarranza:

Yes, bugfix noted in [this discussion](https://github.com/livewire/livewire/discussions/7329)

Yes

No

Yes

### Steps To Reproduce Original Issue

1. Create a component with a public eloquent Collection property
2. Add the same Model (with the same ID) to the collection more than once
3. On initial load of component, the collection has multiple items. 
4. Add anything to the compnent that causes a network round trip ( a button or wired event), and click or trigger the action.
5. When the component rehydrates itself the Collection will only contain a single instance

Example

```php
<?php

namespace App\Livewire;

use App\User;
use Livewire\Component;

class CollectionsTest extends Component
{
    public Collection $users;

    public function mount()
    {
        $user = User::find(1);

        $this->users = new Collection([$user, $user]);
    }

   public function doNothing()
   {
   }
    
    public function render()
    {
        return <<<'HTML'
        <div>
            @foreach($users as $user)
                <div>{{ $user->name }}</div>
            @endforeach
            <a href="#" wire:click.prevent="doNothing">Do Nothing</a>
        </div>
        HTML;
    }
}
```

In Livewire V2 this was solved with the following code in the `trait SerializesAndRestoresModelIdentifiers`. The last operation ensures all ID's are restored even if they are duplicate ID's.

```php

protected function restoreCollection($value)
    {
        if (! $value->class || count($value->id) === 0) {
            return ! is_null($value->collectionClass ?? null)
                ? new $value->collectionClass
                : new EloquentCollection;
        }

        $collection = $this->getQueryForModelRestoration(
            (new $value->class)->setConnection($value->connection), $value->id
        )->useWritePdo()->get();

        if (is_a($value->class, Pivot::class, true) ||
            in_array(AsPivot::class, class_uses($value->class))) {
            return $collection;
        }

        $collection = $collection->keyBy->getKey();

        $collectionClass = get_class($collection);

        return new $collectionClass(
            collect($value->id)->map(function ($id) use ($collection) {
                return $collection[$id] ?? null;
            })->filter()
        );
    }
```

This PR reverts back to using laravel's way of hydrating a collection that was used in Livewire V2


nuernbergerA:
Exposing the connection to the meta payload might not be a good introduction (security wise).

ChristopherCarranza:
> Exposing the connection to the meta payload might not be a good introduction (security wise).

I'm not sure what affect that will have on hydrating the collection? I suppose maybe not much since it wasn't providing that before 🤷🏻‍♂️ but it can be removed if there is a security concern.

joshhanley:
@ChristopherCarranza thanks for the PR! I removed the connection details as we'd prefer not to send them where possible. I also added a test collection order which this PR also manages to fix.

turbo124:
@joshhanley removing the connection here breaks anyone using multiple database connections? we are seeing this as a regression now?

joshhanley:
@turbo124 connection was added in this PR, so I haven't removed anything from Livewire's codebase, just modified this PR 🙂

PR #7365, opened 2023-11-23
@assets directive does not load before @script when assets are remote

PhiloNL:
If you take the example of the documentation and use the assets and script directive, it will throw an error:

```
Uncaught ReferenceError: Pikaday is not defined
```

```
<div>
    <input type="text" data-picker>
</div>
 
@assets
<script src="https://cdn.jsdelivr.net/npm/pikaday/pikaday.js" defer></script>
<link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/pikaday/css/pikaday.css">
@endassets
 
@script
<script>
    new Pikaday({ field: $wire.$el.querySelector('[data-picker]') });
</script>
@endscript
```

The test suite contains a test that loads a script locally, so it seems to be isolated to remote assets.

Wirebox example: https://wirebox.app/b/x2vey


calebporzio:
Phew, that was a gnarly one. Thanks for the failing test @PhiloNL!

PR #7388, opened 2023-11-27
Change 'onUpdate' in the Validate attribute through a 'config' file

luanfreitasdev:

https://github.com/livewire/livewire/discussions/7380
https://github.com/livewire/livewire/discussions/7376
https://github.com/livewire/livewire/discussions/5782

No
No

The default behavior when using the Validate attribute appears to be different than when calling a $this->validate method on the component itself, causing data to be lost and validating a single field.

A common example is when we fill out all the fields on a form and click save, errors will appear, then the user selects using "select" (wire: model.blur) and all errors will disappear, leaving only "select" " ". However, when clicking save the errors reappear for the user.

This pull request changes the behavior to not allow the update to happen unless the user manually indicates it.

If approved, the documentation must be changed.

Thanks for contributing! 🙌


luanfreitasdev:
Keeping onUpdate false directly could break the current behavior; we don't want that.
I updated it to read via the configuration file, which maintains the current behavior, allowing you to change it globally. However, I have some doubts about this:

* We can read the config in each attribute and wouldn't that be a good practice?
* is `validate_realtime` a suggestive name for this case?

calebporzio:
Yeah I see what you're talking about. I agree the error bag clearing behavior as it currently stands is awkward, but I don't think it's the right approach to add a config. Maybe we neeed a different PR that changes that behavior directly as I don't think it's expected behavior. Thanks

PR #7390, opened 2023-11-28
Fetch post before authorization in Security docs example

isaacdew:
Just a small fix to the security docs. In an authorization example, the `$post` variable was referenced in the authorization call but defined after.


calebporzio:
Ooops, thank you!

PR #7396, opened 2023-11-28
Add locked documentation to security.md

chillbram:
Added short `#[Locked]` explanation to the security docs, as one of the means to protect from tampering public properties. I have linked to the Locked docs and included a cautionary note with regards to changing the property through PHP code

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -169,6 +169,37 @@ class ShowPost extends Component
 
 This component is now secured because there is no way for a malicious user to change the `$post` property to a different Eloquent model.
 
+### Locking the property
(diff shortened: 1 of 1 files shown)


docs/security.md
```diff
@@ -169,6 +169,37 @@ class ShowPost extends Component
...
+    public $postId;
+
+    public function mount($postId)
+    {
+        $this->postId = $postId;
+    }
+
+    public function delete()
+    {
+        $post = Post::find($this->postId);
```
PerryvanderMeer, line 196: Maybe it's better to add authorization here, or highlight that this method is insecure?

calebporzio:
Great context, thank you!

PR #7403, opened 2023-11-29
Fix route uri generation

nmalinoski:
Adjust generation of URIs for the livewire.js and update routes to include directories in the URL path above the app root.
This allows Livewire 3.x (and packages that depend on it, like Filament) to work with Laravel apps with endpoints in subdirectories.
https://github.com/livewire/livewire/discussions/7285


calebporzio:
I'm in favor of this if someone tests that crap out of it an an actual app and makes sure it doesn't cause all sorts of other issues.

calebporzio:
Yeah, I'm thinking about this further and just feel like it will cause issues for people in other scenarios when we start using the URL helper to wrap the output. Differently configed environments could get wonky.

If there's more community demand on a PR like this I think I would consider it, but unfortunately it's not something I can merge without lots of support for it and assurance it won't break other people's apps. Thanks

PR #7406, opened 2023-11-29
Add the ability to unregister a global Livewire listener

calebporzio:
This PR adds:

```js
let cleanup = Livewire.on('some-event', () => { ... })

// Calling cleanup() will remove the event listener
cleanup()
```


calebporzio:
Thanks Philo!

PR #7409, opened 2023-11-30
Fix undefined error when payload does not contain assets

PhiloNL:
I noticed I was getting a `TypeError: Cannot convert undefined or null to object`, I'm not sure if `payload.intercept` should always contain the `asset` array, if it does, then this probably isn't the correct fix.


calebporzio:
Yeah, might as well just to be safe. Thanks!

PR #7421, opened 2023-12-01
Add Enum binding support

geisi:
## Description
This pull request significantly enhances Livewire by adding support for binding Enums in both Full Page and regular Livewire components. It extends this functionality to include route binding and binding through component mounts and actions.

## Problem
Currently, while Laravel supports enum route binding, Livewire does not extend this support to its components, leading to a `BindingResolutionException` when an Enum is bound to a Livewire route or component. This limitation affects both Full Page and standard Livewire components.

## Code Examples

1. **Enum Declaration in `app/Enums/Color.php`:**
    ```PHP
    enum Color: string {
        case BLUE = 'blue';
        // ... other cases ...
    }
    ```

2. **Route Definition in `web.php`:**
    ```PHP
    Route::get('/create/{color}', ColorForm::class);
    ```

3. **Livewire Component in `Livewire/ColorForm`:**
```PHP
namespace App\Livewire;

use Livewire\Component;
use Enums\Color;

class ColorForm extends Component
{
   public Color $color;

   public function changeColor(Color $color) 
   {
      $this->color = $color;   
   }
}
```

## Proposed Solution
This PR introduces modifications to enable seamless Enum binding within Livewire components.

## Request for Feedback
I am keen to hear your thoughts on this implementation, particularly regarding any potential for simplification or further optimization.

Thank you very much!

### DIFF
--- a/src/Drawer/ImplicitRouteBinding.php
+++ b/src/Drawer/ImplicitRouteBinding.php
@@ -95,18 +95,8 @@ class ImplicitRouteBinding
             return $parameterValue;
         }
 
-        if ($parameterValue instanceof BackedEnum) {
-            return $parameterValue;
-        }
-
-        if ((new ReflectionClass($parameterClassName))->isEnum()) {
-            $enum = $parameterClassName::tryFrom($parameterValue);
-
-            if (is_null($enum)) {
-                throw new BackedEnumCaseNotFoundException($parameterClassName, $parameterValue);
-            }
-
-            return $enum;
+        if($enumValue = $this->resolveEnumParameter($parameterValue, $parameterClassName)) {
+            return $enumValue;
         }
 
         $instance = $this->container->make($parameterClassName);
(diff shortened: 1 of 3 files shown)


joshhanley:
@geisi thanks for the PR! I wonder if, to make it a little cleaner, we could implement it in a similar way to how Laravel have implemented it https://github.com/laravel/framework/blob/b0ed9e1a601d8cbf122c815198d86b333a881f11/src/Illuminate/Routing/ImplicitRouteBinding.php#L28

What do you think?

geisi:
@joshhanley, thank you for your input. To clarify, are you suggesting improvements for the routing implementation alone, or does this also extend to the action call aspect of the code?

joshhanley:
@geisi I'm not sure I follow the difference. I just like how there is a named method in the Laravel implementation that encapsulates and describes the backed enum behaviour. Was just a suggestion/ inspiration, you'd need to decide if it fits in or not, as you know your changes better than me (I've only had a cursory look) 🙂

joshhanley:
@geisi I'm not sure I follow the difference. I just like how there is a named method in the Laravel implementation that encapsulates and describes the backed enum behaviour. Was just a suggestion/ inspiration, you'd need to decide if it fits in or not, as you know your changes better than me (I've only had a cursory look) 🙂

geisi:
@joshhanley, I took another look at the part about route resolving. I found that Livewire uses a different way to manage route bindings compared to the main framework. Because of this difference, we can't use the logic from the main framework here. But, I have made some small changes to make the code easier to read.

calebporzio:
Thanks!

PR #7423, opened 2023-12-02
Add failure test when using wire:model together with x-mask

luanfreitasdev:

https://github.com/livewire/livewire/discussions/7100

Yes

No

Yes

Livewire binds the value before the mask is applied when a mask is added to an input with a wire:model.
You can see the issue by accessing: https://wirebox.app/b/4j29y.
We can set a timeout for the dataset to make sure that the value is only returned after the mask has been applied.

Thanks for contributing! 🙌


calebporzio:
ok I see the problem here, I think the solution is to change x-mask to override input events so that they are only fired after x-mask has already changed the input. Right? or do you think that would cause unwanted side effects in some way - it seems the cleanest and most robust to me though

clinvest:
Hi, same problem here, using the mask feature through filamentphp: https://github.com/filamentphp/filament/issues/10694

kismonguz:
> Hi, same problem here, using the mask feature through filamentphp: [filamentphp/filament#10694](https://github.com/filamentphp/filament/issues/10694)

If you using money formatter, it will also mess up the thousand separator.

calebporzio:
Thanks for the report and failing test - I was able to identify and fix it on the Alpine side.

Here's the alpine PR: https://github.com/alpinejs/alpine/pull/3988

This fix will be included in the next release.

PR #7424, opened 2023-12-02
Add 'except' support for queryString hooks

colinc:
#7255 added support for the "except" parameter in the query string feature, however, it only works for the URL attribute. This adds support for the "except" parameter in the queryString method and trait hooks:

```
use Livewire\Component;
 
{
    // ...
 
    protected function queryString()
    {
        return [
            'search' => [
                'as' => 'q',
                'except' => 'something',
            ],
        ];
    }
}
```

```
trait WithSorting
{
    // ...
 
    protected function queryStringWithSorting()
    {
        return [
            'sortBy' => ['as' => 'sort'],
            'sortDirection' => ['as' => 'direction', 'except' => 'asc'],
        ];
    }
}
```


joshhanley:
@colinc thanks for the PR! Can you please add some tests? Thanks!

joshhanley:
@luanfreitasdev thanks! @colinc I have uncommented the test, but it seems there are issues. I suspect it's not related to your PR but something to do with the except implementation from PR #7255.

joshhanley:
@colinc thanks for the PR! Not sure why the original tests weren't working, so I created a new one and it works happily. So just going to run with that for now. When I merged main in, I noticed the attribute except test had been removed, which I suspect was an error so I have bought that back.

PR #7426, opened 2023-12-02
Allow computed snake case within blade view

luanfreitasdev:

https://github.com/livewire/livewire/discussions/7352
https://github.com/livewire/livewire/discussions/7186

Yes

No

Yes

When it is necessary to utilize a computed property without the "Computed" attribute, we can access a method in the view using Snake Case, as demonstrated in version 2 of the software. However, in version 3, this behavior is not consistently maintained when the Computed Attribute is employed.

### It works currently

> component

```php
public function getFooBarProperty()
{
  return 'foo'
}
```

> blade

```blade
<div>
     {{ $this->foo_bar }} // foo
</div>
```

---

### Does not currently work

> component

```php
#[Computed]
public function fooBar()
{
  return 'foo'
}
```

> blade

```blade
//  Using a computed property with an attribute I got this error:
//  Property [$foo_bar] not found on the component:
<div>
     {{ $this->foo_bar }}
</div>
```

---

* This pull request enables us to employ Snake Case in the same manner as the conventional method when 
utilizing the investment attribute of the get.Property magic method.

* In addition, a corresponding test has been included for validation purposes.

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -42,7 +42,7 @@ class BaseComputed extends Attribute
     protected function handleMagicGet($target, $property, $returnValue)
     {
         if ($target !== $this->component) return;
-        if ($property !== $this->getName()) return;
+        if ($this->generatePropertyName($property) !== $this->getName()) return;
 
         if ($this->persist) {
             $returnValue($this->handlePersistedGet());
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for the PR!

shawnveltman:
This (or at least I think it's this) breaks instances for projects using snake_case for method names in the component.  3.2.6 allowed me to have a property:

```php
#[Computed]
    public function some_long_property_name_thats_easier_to_read_as_snake_case(): string
    {
        return "Yeah!!!";
    }
```

3.3.0 returns a 500 error with that same property now.

PR #7431, opened 2023-12-03
Update upgrading.md to include additional new config key

mferrara:
The upgrade process mentions 4 of the 5 keys added to the livewire.php configuration file, I've added the final key to reduce any confusion for those updating.

The `pagination_theme` key was missing from the list of new keys.


PhiloNL:
Thanks @mferrara!

PR #7433, opened 2023-12-04
Add redirectIntended

binaryfire:
Adds a `redirectIntended` method to allow redirecting to the previous page.

See https://github.com/livewire/livewire/discussions/7273

### DIFF
--- a/docs/redirecting.md
+++ b/docs/redirecting.md
@@ -65,6 +65,14 @@ In case you need to pass parameters to the route you may use the second argument
     $this->redirectRoute('profile', ['id' => 1]);
 ```
 
+## Redirect to Intended
+
+In case you want to redirect the user back to the previous page they were on you can use `redirectIntended`. It accepts a default URL as its first argument which is used as a fallback if no previous page can be determined:
+
+```php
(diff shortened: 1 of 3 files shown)


joshhanley:
@binaryfire sweet thanks! I’ve triggered the 8.1 tests to run again 🙂

calebporzio:
Thanks for the contribution, however I'm not sure about it seeing as Laravel uses: `redirect()->intended('dashboard');` instead of `->redirectIntended()`.

You can add this method to `src/Features/SupportRedirects/Redirector.php` instead to achieve a similar API. Going to hold off on merging until it matches Laravel a bit better. Thanks!

binaryfire:
Hi @calebporzio. I'm not that will solve this issue: https://github.com/livewire/livewire/discussions/7273 ? The goal with this PR was to be able to redirect to the intended url and also specify `navigate: true` / `navigate: false` .

It's not possible to do `$this->redirect(navigate: true)->intended('https://defaulturl.com')`, which is why I added a separate method.

calebporzio:
I see, thanks!

PR #7444, opened 2023-12-04
Added `.boolean` modifier to wire-model.md

edwinvdpol:
I was watching a new Livewire screencast and I noticed that the `.boolean` modifier was used, but was undocumented.


PhiloNL:
Thanks @edwinvdpol!

PR #7457, opened 2023-12-05
Extract out Mechanism API

inxilpro:
Currently, each mechanism handles registering itself as a singleton slightly differently. This PR just cleans that up by moving the shared code into a single `Mechanism` class and then deleting ~30 lines of duplicate code.

The only other change is to introduce a `LivewireTagPrecompiler` class, which should help isolate any future issues with changes to the upstream `ComponentTagCompiler` class in Laravel. This improves the mechanism API and, hopefully, minimizes future conflicts.


calebporzio:
This is better. Thanks @inxilpro !!

PR #7458, opened 2023-12-05
Docs note aboute redering components in subdirectories

Ligniperdus:
The syntax for rendering components or including nested components in subdirectories is not clear. I have therefore adopted and adapted the Blade documentation from Laravel on this topic.


PhiloNL:
Thanks @Ligniperdus!

PR #6984, opened 2023-10-03
Failing test for add and select `option` in `select`

PerryvanderMeer:
Review the contribution guide first at: https://laravel-livewire.com/docs/2.x/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?
#6505

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)
`add-option-and-select`

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
✅ 

4️⃣ Does it include tests? (Required)
✅

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.
A combination of adding items to `$customers` and setting `$customer` will cause a failure in updating the front-end variable (the `<select>` will be set to the first item instead of the newly added item).

It works without adding the item to `$customers` and setting `$customer` to `3` though.

Thanks for contributing! 🙌

PR #7500, opened 2023-12-09
Add failing test for select options not updating like they did in V2

joshhanley:
Currently, if the options inside a select input change between requests, if the value that the select input is set to isn't on the page, the selected option gets reset and nothing is displayed.

In V2, Livewire would morph the options first then it would deal with the databinding updates.

But in V3, this operation runs the opposite way, as Alpine's `forceModelUpdate()` is triggered inside an effect which updates the selected value, and then morph is being run.

So we need to find a way to change the order of operations in V3 to ensure the DOM is changed first, then we try and update the select input's selectedIndex, as trying to set the selected index before the option exists results in a no-op and the selectedIndex get's reset to -1.

Morph is an effect and effects run after Livewire merges the new snapshot.

## Debugging

Code sections to note:
Alpine:
`packages/alpinejs/src/directives/x-model.js:108`
`packages/alpinejs/src/utils/bind.js:122`

Livewire:
`js/commit.js:132`
`js/morph.js:29`

We can see in this screenshot that in V3 the `forceModelUpdate` happens before `morph-updating`, causing the issue.
Pasted image 20231202162040

This screenshot demonstrates what happens if we set a selected option to an option that doesn't exist (get's set to `-1`)
Pasted image 20231202162012

## V2 Code Path

Code:
`js/component/index.js:287`
`js/component/index.js:358`
`js/dom/dom.js:153`
`js/dom/dom.js:191` - updateSelect (same as Alpine `bind.js:122`)

Pasted image 20231202162646

Pasted image 20231202162706

Pasted image 20231202162722

### DIFF
--- a/src/Features/SupportDataBinding/BrowserTest.php
+++ b/src/Features/SupportDataBinding/BrowserTest.php
@@ -35,4 +35,96 @@ class BrowserTest extends BrowserTestCase
             ->assertDontSee('Unsaved changes...')
         ;
     }
+
+    /** @test */
+    public function can_display_correct_option_in_a_select_input_when_options_have_changed()
+    {
+        Livewire::visit(new class extends Component {
+            public $country = 'US';
+            public $state = 'TX';
+
+            public $countries = [
+                'US' => [
+                    'code' => 'US',
+                    'name' => 'United States',
+                    'states' => [
+                        'CA' => 'California',
+                        'TX' => 'Texas',
+                        'NY' => 'New York',
+                    ],
+                ],
+                'AU' => [
+                    'code' => 'AU',
+                    'name' => 'Australia',
+                    'states' => [
+                        'QLD' => 'Queensland',
+                        'SA' => 'South Australia',
+                        'NSW' => 'New South Wales',
+                    ],
+                ],
+            ];
+
+            public function getStateOptionsProperty()
+            {
+                if ($this->country === null|| $this->country == '') {
+                    return null;
+                }
+
+                return $this->countries[$this->country]['states'];
+            }
+
+            public function change()
+            {
+                $this->country = 'AU';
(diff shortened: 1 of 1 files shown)


PerryvanderMeer:
This one might be related to #6984

joshhanley:
@PerryvanderMeer yep looks like it!

calebporzio:
Thanks for the failing test @joshhanley, definitely a deep on as I think many other parts of the system rely on/benefit from the snapshots being merged before the morph happens.

I understand this wouldn't fix it for every case, but in this case, couldn't/shouldn't you add a "wire:key" to the dependent select so that it is torn down and re-created when the entirely of it's options change? Then the new binding should work right?

PerryvanderMeer:
@calebporzio Hmm, tried that trick for the case in #6984 and 
```php
<select wire:model.live="customer" wire:key="{{ str()->random(50) }}">
```
 seems to make the magic work..

joshhanley:
@calebporzio thanks, yeah I should have mentioned adding a `wire:key` does solve it. It was different in V2 though as it just worked without the `wire:key`. Probably not worth the effort to fix in V3 due to the major structural change to get it working. Would be nice though, not having to use the `wire:key` outside of a loop 😁

@PerryvanderMeer the issue with your key, is that it will change on every request, so it will remove and re-add the element everytime. So if the element is focused or anything, it will probably loose it.

Instead I would try to derive the key from the data that may change, like

```blade
<select wire:model.live="customer" wire:key="{{ count($customers) }}">
```

So that way it is only replaced when the select options change.

calebporzio:
Good tip @joshhanley - I added docs for dependent selects and I think we can move past this for now. Let's re-examine other solutions if the docs telling people to add wire:key to dependent selects don't do it.

PR #7506, opened 2023-12-11
Support the new Alpine magic $model

calebporzio:


js/lifecycle.js
```diff
@@ -70,6 +70,21 @@ export function start() {
                     } })
                 })
             }
+        }, el => {
```
nuernbergerA, line 73: dear future caleb: make sure to remember to bail out like we did above :rofl:

PR #7507, opened 2023-12-11
Fix lazy isolation

joshhanley:
This PR fixes an issue where lazy components were no longer batched on subsequent requests and adds a test for it.

It also fixes an issue with polling, where all polling components commits were being isolated by default now. I have added a polling test to cover this case.


jessarcher:
I've tested this locally with Pulse, and it works great!

The initial page load sends off an update for each card individually, so a slow card doesn't impact the initial loading of other cards. Once loaded, they all poll together on the same request so there aren't 8 separate requests every 5 seconds :grin:

Thanks @calebporzio and @joshhanley! This is a really neat feature for improving the perceived speed of an application with a complex set of independent components.

calebporzio:
Thanks Josh!

PR #7508, opened 2023-12-11
Not defined functions when using @scripts tag

Sparclex:
These are failing tests for the following scenario:

An error occurs in the js console when trying to call a function defined in the `@scripts` tag using for example alpine's `@click`. Furthermore the function is somehow run on initialization without clicking the button.

Minimal Reproduction: https://wirebox.app/b/4n5l3
Discussion: https://github.com/livewire/livewire/discussions/7348


calebporzio:
Thanks!

PR #7510, opened 2023-12-11
Update bound value from lifecycle hook

danharrin:

filamentphp/filament#9743

Yes

No

yes

When there is a pending change in the frontend that is yet to be hydrated, it can overwrite more recent changes made in lifecycle hooks.

I expect lifecycle hooks like `updating*` to run after every property from the frontend has been properly hydrated.

In this case, making the second select `.live` fixes the problem, as the value in the backend is already up to date by the time the lifecycle hook runs.

Also, if you swap the order of the `$foo` and `$bar` property definitions, it works as expected. Probably because `$bar` is then hydrated first.

Thanks for contributing! 🙌


calebporzio:
Hey Dan, I think this makes sense. I changed it so that all properties are updated before ANY of the updated hooks run.

danharrin:
Thanks!

PR #7519, opened 2023-12-12
Fix incorrect filepath in quickstart.md

YanDatsiuk:
Location (filepath) of layout file in documentation (page: docs/quickstart ) was inaccurate.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -102,7 +102,7 @@ Now, our _counter_ component is assigned to the `/counter` route, so that when a
 
 ## Create a template layout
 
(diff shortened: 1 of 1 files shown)


PhiloNL:
Hi @YanDatsiuk,

`resources/views/components/layouts/app.blade.php` is the correct path. I think the path you are referring to was used in [Livewire v2](https://laravel-livewire.com/docs/2.x/rendering-components#page-components).

https://github.com/livewire/livewire/blob/main/config/livewire.php#L31-L41
https://github.com/livewire/livewire/blob/main/src/Features/SupportConsoleCommands/Commands/LayoutCommand.php#L18

Closing this PR but if you think this is incorrect, please let me know.

PR #7520, opened 2023-12-12
Add `assertHasError`

jasonmccreary:
I wanted to test the following line of code which adds a custom error message, but couldn't.

```php
$this->addError('repository', 'Please enter the repository in `vendor/package` format.');
```

With this PR, I can:

```php
Livewire::test(MyComponentClass::class)
    ->submit()
    ->assertHasError('repository', 'Please enter the repository in `vendor/package` format.');
```

This assertion mimics `assertSessionHas` in Laravel - you may assert the key exists, it matches a specific value, or even pass a closure for more advanced checks.

There may be opportunity to update the existing `assertHasErrors` method. But changing its signature potentially felt like a breaking change. Furthermore, I personally feel the parameter overloading gets nuts. I'd rather have single purpose, straightforward methods.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -415,7 +415,7 @@ You can also pass parameters to actions by passing additional parameters into th
 
 ### Validation
 
-To test that a validation error has been thrown, you can use Livewire's `assertHasErrors()` method:
+To test that a validation error has been thrown, you can use Livewire's `assertHasError()` or `assertHasErrors()` methods:
 
 ```php
 <?php
@@ -434,7 +434,7 @@ class CreatePostTest extends TestCase
         Livewire::test(CreatePost::class)
             ->set('title', '')
             ->call('save')
-            ->assertHasErrors('title');
+            ->assertHasError('title');
     }
 }
 ```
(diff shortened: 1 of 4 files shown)


calebporzio:
Nice, @jasonmccreary; what do you think about my changes? I made `assertHasError()` and `assertHasErrors()` use the same underlying logic so that the experience of using them both matches.

The only "breaking" change is now asserting against messages instead of just rules.

I don't really consider this a breaking change, though, because I can't really imagine a scenario where you are asserting against a rule with the exact same name as another unrelated message.

I also added some docs and re-worked the existing ones to use `assertHasError()` as the default because I think it's more intuitive.

jasonmccreary:
@calebporzio, thanks. Your call, but I probably wouldn't add mine anymore. Just allow a message to be passed in the existing `assertHasErrors`. I don't want to create any confusion in the DX.

As a newcomer to Livewire, I was surprised that `assertHasErrors` asserted the validation rule name. I might have expected that to be `assertValidationRuleFailed` or something. But again, since it already exists in its current capacity, I'd simply allow it to check a message too. I'd be rare someone did something like `$this-addError('field', 'required');` Even if they did, it would pass. So meh.

Again, your call. Just my thoughts.

calebporzio:
Yeah, good call - ok, I'll refactor to make the API slimmer and remove `assertHasError`.

By default we assert rules instead of messages because it's far more useful in a lot of cases, especially with localization and such.

But yeah, it's not symmetrical with Laravel assertions, which is what we usually aim for.

The good news is, if we ever want to become symmetrical, it wouldn't be a breaking change because we don't even have those methods available right now.

jasonmccreary:
Right on.

Thanks again! Liking Livewire. Will probably do more PRs around the testing layer as I like that stuff. ✌️

calebporzio:
Sweet, I welcome it @jasonmccreary !

PR #7528, opened 2023-12-13
Handle event dispatching after script execution

calebporzio:
Currently, because the events feature and event utilities are in the same file, the event dispatching hook is registered too early in the lifecycle.

This PR breaks them apart so that earlier parts of the lifecycle can still dispatch and listen for events, but the actual effect hook to dispatch events from a Livewire component on initialization is deferred.

This enables the following scenario where you might dispatch an event from `render()` and listen for it inside a script:

```php
public function render()
{
    $this->dispatch('some-event')->self();
}
```

```html
@script
<script>
    $wire.on('some-event', ({ data }) => {
        //
    }) 
</script>
@endscript
```


SwitchCase-dev:
Hello @calebporzio 

I have a small problem with this PR.

Scenario: I have a lot of forms in my back-end panel where I track changes from different angles with an event. So when a user changes the page, an alert is triggert to inform the user the changes aren't saved yet.
When the form is submitted, the same event is dispatched to set the dirty state of the page to false, so the alert does not popup (which is correct because the user is saving the form).

It all wen't well until Livewire 3.1.1, until this PR, because the event is triggert at the end of the lifecycle. Is there any way to force events to be executed earlier?

`//-- Disable browser dirty state alert while submitting
  $this->dispatch('isDirty', state: false);`

`window.addEventListener('isDirty', event => {
...
//-- Init dirty state browser alert
window.onbeforeunload = function() {
          return true;
};
...
}`

jj15asmr:
Hey, @calebporzio,

This PR also seems to have broken my code when updating to v3.3.1 and above. In my case, I am dispatching an event from my component and then listening for it using Alpine to add an Animate.css class to a persisting element like so:

```html
<div
    x-data="{ elementClass: '' }"
    x-bind:class="elementClass"
    x-on:my-event.window="elementClass = 'animate__animated animate__fadeIn'"
>
    <!-- ... -->
</div>
```

It'll work fine the first time the event's dispatched but not after that point. Alpine does continue to pick up the event though as proved with a `console.log()`.

Is it clear to you why this is happening and possibly a way to fix it?

Thank you.

PR #7531, opened 2023-12-13
Fix loading state on incorrect targets

solarsilk:
See https://github.com/livewire/livewire/discussions/7485

### Steps To Reproduce

When targeting 2 similarly-named nested properties, and one is in the loading state, the other is also put in the loading state.

Eg. Below, I'm using an array property (`myModel`) with 2 nested properties with similar names. When `myModel.prop2` is loading, `myModel.prop` is also put in the loading state.

```
<input type="text"	
wire:model.live="myModel.prop" 
wire:target="myModel.prop"
wire:loading.class="bg-red-50"
>

<input type="text"	
wire:model.live="myModel.prop2" 
wire:target="myModel.prop2"
wire:loading.class="bg-red-50"
>
```

I believe this is caused by [this line](https://github.com/livewire/livewire/blob/ecded08cdc4b36bbb4b26bcc7f7a171ea2e4368c/js/directives/wire-loading.js#L121C47-L121C47) which is `property.startsWith(target)`  Using `startsWith` will incorrectly find targets whose names start with the intended target name.

See https://wirebox.app/b/4mp2r for an example

### DIFF
--- a/js/directives/wire-loading.js
+++ b/js/directives/wire-loading.js
@@ -118,7 +118,7 @@ function containsTargets(payload, targets) {
         }
 
         let hasMatchingUpdate = Object.keys(updates).some(property => {
-            return property.startsWith(target)
+            return property.matches(target)
         })
 
         if (hasMatchingUpdate) return true


joshhanley:
@solarsilk thanks for the PR! Can you please add a test for this? Thanks!

solarsilk:
@joshhanley I'm not very well versed in writing tests, but I'm trying hard. Maybe you can give me a tip to help? Here's what I've done:
- forked livewire/livewire main branch to https://github.com/solarsilk/livewire
- cloned repo locally
- created `fix-target-bug` [branch](https://github.com/solarsilk/livewire/tree/fix-target-bug)
- composer install
- typed `php artisan dusk` and i get: `Could not open input file: artisan`
- to do: create test for this issue in `src/Features/SupportWireLoading/BrowserTest.php`

Am I on the right path? I see that there is no artisan in the root of the project. Not sure what to do from here. Any tips are appreciated! Thanks.

joshhanley:
@solarsilk no worries!

So with the package, you need to run `vendor/bin/phpunit` which will run the dusk test for you, as there is no artisan.

If you want to just run the one file/ test in the file, you just use phpunit as per normal specifying the file or using `--filter` to specify the test name.

Hope this helps!

solarsilk:
Ah, amazing, thank you. I think the link to the Contribution Guide was broken last time I checked. That will be super helpful. I'll continue and hopefully have a PR soon. Thanks again!

solarsilk:
@joshhanley  FYI I read that I need to create a separate branch for my fix (which I didn't do in this PR), so I closed this one and made a [new PR](https://github.com/livewire/livewire/pull/7548). I learned a ton through doing this so thanks for your help!

joshhanley:
@solarsilk yeah I fixed it just before I messaged you about it 😆

Yep makes sense, thanks for doing that. Awesome! No worries 🙂

PR #7535, opened 2023-12-14
Alpine re-rendering removed elements

danharrin:

Reported in Filament quite a lot in different cases. Kinda related to #6309 but not exactly the same.

Yes

No

Yes

Most of the issues we face with Livewire 3 relate to requests that remove Alpine components from the DOM. After the response arrives, the Alpine component appears to re-render before being removed from the DOM.

This causes issues in quite a few different cases. In this failing test, we are relying on an entangled property that gets removed by the same request that hides the Alpine component. Since this Livewire property does not exist anymore, the entangled property cannot be used by any part of the Alpine component, and is `undefined`. In this case, the entangled property is being used in an `x-if`, and is causing a console error, as it is `undefined` and we are attempting to accessing a property on it. The easiest way for me to write this test was to assert that there were no console errors, but I can tweak it if you suggest a better approach.

The expected behaviour IMO is that if the Alpine component is not present in the Livewire response, it should be destroyed without a chance to re-render.

Flaws in this PR:
- This test is probably in completely the wrong file, so sorry about that. But I couldn't seem to find a better place. If you suggest a better file, I can move it.
- The `sleep()` is nasty, but required otherwise the element is still visible. `waitForLivewire()` doesn't seem to wait for all Alpine side effects? I'm sure this can be refactored somehow, but I can't figure it out.

Thanks for contributing! 🙌


calebporzio:
Hey Dan, here's what's happening:

* You are entangling an Alpine property to a deeply nested Livewire property
* That deeply nested Livewire property is removed on an update (that also removes the component itself)
* When the request comes back, before morph happens, Livewire's `$wire` state is updated to the latest state
* When this happens it triggers any reactivity related to that property (still before morphing)
* Your component (that is about to be removed) reacts to the change, and throws an error because nested data it depends on is gone
* THEN morphdom happens and removes the component

So the userland options here are:
* Don't depend on nested data
  * Like if `bar` wasn't a nested array and it was just a boolean everything works fine
* Add an additional check to ensure the nested data you're depending on is present: `x-if="state && state.isVisible"`

Both of those fix the test

As far as fixing this scenario in the core, there's really only one solution I see:
* Make morph happen before Livewire internally merges new data changes

I pushed a commit to see if the tests pass and they all pass with this change

The fix i made isn't sustainable, it's just a hack to test if it breaks other things and apparently it doesn't.

However, I think we gain a lot by doing all the data merging first because at the time morph runs, it's dealing with the most up-to-date state and avoiding lots of duplicate work and potentially other side effects. It's more efficient on the JS side.

I also means when morphdom is doing it's "cloning" phase to initialize the new tree, it's working with up-to-date state instead of stale state.

Which logically makes sense. Like you want that in most cases, but there are times like this example where you don't want it.

I think this is one of those weird timing issues, where we probably fixed a lot of issues by doing this in V3, but introduced a new set of issues.

The question is, which issues do we prefer? (and of course are we breaking anything if we change this)

Do you have any opinions on this or experience with it, @joshhanley ?

danharrin:
Hey @calebporzio! Thanks for diving into this. I think it would help me if you summarised a few of problems that are solved (aside from performance) by updating the state before morphing?

I do depend on nested entangling very heavily, and it's one of the most powerful features of Livewire that I'm using when building Filament. So it's not really an option for me to stop.

As for doing the additional checks, that is what I do right now to avoid console errors. However, the experience doesn't scream out to me as the "expected" behaviour in this situation, and does feel like a papercut honestly. But knowing more about bugs that would be introduced by morphing before updating the state would help me figure out if it's a worthwhile tradeoff in my opinion 😊

luanfreitasdev:
What I'm observing is that `state.isVisible` is removed from the DOM when clicking "hide", and alpine will not be able to understand and construct the object because it no longer exists. When applying null safe with `state?.isVisible`, JavaScript will recognize that it has been removed.

danharrin:
I know there is a workaround, but as I said before, this is definitely a papercut, since it's not expected that a removed element is rendered with the updated Livewire component state before it is removed. And it worked absolutely fine in v2 without any workarounds.

calebporzio:
Dan and I went pretty deep on this on a pair.

We realized what needs to happen is data synching has to happen AFTER morphing. to prevent these problems.

Forgetting all the potentially negative consequences of that in V3, we couldn't even get them to run in reverse order because of the way it's architected using microtask queueing for the morphing call.

Anyhow, I think I'm going to table this further.

I think this might have to be something to explore for the next major version as it's a pretty fundamental part of Livewire right now.

Please re-open this PR or submit a new one if there are any developments, but for now I'm going to close.

Thanks everyone.

PR #7536, opened 2023-12-14
Remove `rel` attribute from button

utsavsomaiya:
image

Button does not support rel attribute. So I removed it


SagarNaliyapara (approved):

utsavsomaiya:
image

But @calebporzio as I have seen this anchor tag. And anchor tag support the `rel` attribute. But button tag can not..

PR #7541, opened 2023-12-14
Support mount() hooks for form objects

calebporzio:


calebporzio:
Closing this because I thought I needed it and I don't currently. It's obv useful in some ways but there are questions:

* Does it run BEFORE or AFTER component `mount()`. I think the answer is BEFORE. But not totally sure.
* What if you pass as form object as a reactive property to a nested component, with this implementation the mount() will be called for the originating form object and the passed through one. I think we would want only the first `->mount()` to be called first if it's a prop passed in right? right.

DanielSpravtsev:
This is interesting feature
> * Does it run BEFORE or AFTER component `mount()`. I think the answer is BEFORE. But not totally sure.

I think before, because you can interact with forms, for example `$form->setUser()` inside component `mount()`, so at this point form mount should be already finished


Anyway form `mount()` method can be handled inside component mount just by using `$form->mount() `

joshhanley:
@DanielSpravtsev yeah I think if people want a form mount method, they can easily create one and call it from the component mount like you suggested.

PR #7547, opened 2023-12-15
Remove `rel` attribute from button

utsavsomaiya:
This is reopened as I have seen that @calebporzio missed something.

Sorry for interrupting.

Here is the reference : https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel


joshhanley:
@utsavsomaiya please edit this PR description to include your findings, so Caleb doesn't need to go digging 🙂

joshhanley:
This PR is reopened PR #7536.

lrljoe:
Just a note that rel attribute is only used when href is present, and using on "button" elements isn't supported, so makes sense to remove it.

utsavsomaiya:
> Just a note that rel attribute is only used when href is present, and using on "button" elements isn't supported, so makes sense to remove it.

Yupp!! Thanks

calebporzio:
Sounds good, thanks for pushing this back up!

PR #7548, opened 2023-12-15
Fix `wire:loading` target bug

solarsilk:
See https://github.com/livewire/livewire/discussions/7485

### Steps To Reproduce

When targeting 2 similarly-named nested properties, and one is in the loading state, the other is also put in the loading state.

Eg. Below, I'm using an array property (`myModel`) with 2 nested properties with similar names. When `myModel.prop2` is loading, `myModel.prop` is also put in the loading state.

```
<input type="text"	
wire:model.live="myModel.prop" 
wire:target="myModel.prop"
wire:loading.class="bg-red-50"
>

<input type="text"	
wire:model.live="myModel.prop2" 
wire:target="myModel.prop2"
wire:loading.class="bg-red-50"
>
```

I believe this is caused by [this line](https://github.com/livewire/livewire/blob/ecded08cdc4b36bbb4b26bcc7f7a171ea2e4368c/js/directives/wire-loading.js#L121C47-L121C47) which is `property.startsWith(target)`  Using `startsWith` will incorrectly find targets whose names start with the intended target name.

See https://wirebox.app/b/4mp2r for an example


joshhanley:
@solarsilk thanks! I've triggered the tests to run 🙂

solarsilk:
@joshhanley  oops, didn't realize I needed to commit the compiled JS. I added a new commit with those files. The tests should pass now.

joshhanley:
@solarsilk no worries! Seems like there might be an issue with the implementation as the other loading tests are now failing 😕 I haven't had a chance yet to look at what might be causing that.

solarsilk:
@joshhanley  yeah I looked at the legacy tests but wasn't sure why they are failing. if I can help in any way, let me know.

calebporzio:
Ok yep, yeah I don't think the implementation is right as it is, because we still need a "startsWith" so that people can use target on a root object and still trigger loading for nested changes.

The solution here is to just make startsWith() smarter to look for dots and only do like a "startsWithDotAware()" or some better named method than that.

Thanks!

joshhanley:
@solarsilk thanks for the PR! I fixed up the issue causing the other tests to fail.

solarsilk:
Thank you @joshhanley!

PR #7549, opened 2023-12-15
Enable live updates for the search example in `volt.md`

osbre:
I found it confusing that nothing was updating or changing as I typed into the search input. Don't know how it was intended to work without `.live`.


joshhanley:
@osbre thanks!

PR #7550, opened 2023-12-16
Fix ReflectionException when getting property on stdClass object

petergroeneweg:
Using a class property ($obj) of type stdClass in a livewire component and binding to the object property ($obj->property) results in a ReflectionException originating from the class BaseUtils and the method getProperty. This method calls ReflectionClass($target) with target being $obj. This PR uses ReflectionObject($target) instead.
```
class Test extends Component
{
    public $obj;

    public function mount()
    {
        $this->obj = (object)[];
    }

    public function render()
    {
        return <<<'HTML'
            <div>
                <input type="text" wire:model.live="obj.property">
                <span>{{ $obj?->property ?? '' }}</span>
            </div>
        HTML;
    }
}
```

### DIFF
--- a/src/Drawer/BaseUtils.php
+++ b/src/Drawer/BaseUtils.php
@@ -2,8 +2,6 @@
 
 namespace Livewire\Drawer;
 
-use ReflectionClass;
-
 class BaseUtils
 {
     static function isSyntheticTuple($payload) {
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks! This can be a unit test right? Let's make it a unit test instead of a browser test so it doesn't slow down the test suite

joshhanley:
@petergroeneweg thanks for the PR! I have converted the test to a unit test.

PR #7551, opened 2023-12-16
Fix full page lazy components and using a custom layout using the layout attribute

gazzoy:
See #6907 

Yes

Yes

No

Yes

Currently Livewire doesn't support using custom layout for lazy-loaded component. This fix aims to fix this.


src/Features/SupportPageComponents/HandlesPageComponents.php
```diff
@@ -10,14 +12,18 @@ function __invoke()
...
+        $layout = null;
 
         $layoutConfig = SupportPageComponents::interceptTheRenderOfTheComponentAndRetreiveTheLayoutConfiguration(function () use (&$html) {
             $params = SupportPageComponents::gatherMountMethodParamsFromRouteParameters($this);
 
             $html = app('livewire')->mount($this::class, $params);
         });
 
-        $layoutConfig = $layoutConfig ?: new PageComponentConfig;
+        $reflectionClass = new \ReflectionClass($this);
```
DanielSpravtsev, line 23: We need to Reflect class only when there are no `$layoutConfig` defined

src/Features/SupportPageComponents/HandlesPageComponents.php
```diff
@@ -10,14 +12,18 @@ function __invoke()
...
+        $layout = collect($reflectionClass->getAttributes(Layout::class))->first()?->getArguments()[0];
```
DanielSpravtsev, line 24: `collect` function takes extra resources to extract array and create it's own, so I think it is better to just use array index [0] without adding `collect()` just for ->first() methods, that equivalent to array[0]

src/Features/SupportPageComponents/HandlesPageComponents.php
gazzoy: Thank you for your feedbacks. I updated on bc930f96deaa3a8fbf860575ef860c4be383f171 to do  so.

src/Features/SupportPageComponents/HandlesPageComponents.php
gazzoy: Undersootd! Fixed on bc930f96deaa3a8fbf860575ef860c4be383f171 as well.

gazzoy: I implemented as array with `isset`, but not too sure we need to call it. Please let me know if we don't need to check array indices with `isset`.

DanielSpravtsev: We could just use ->getAttributes()[0] ?? null 
Without any additional if statements :)

gazzoy: Got it! I fixed it on ce10eac16d0683eafac50c48225b651006a08a7e.

I mimicked below code:
https://github.com/livewire/livewire/blob/main/src/Features/SupportLazyLoading/SupportLazyLoading.php#L35

Additionaly, put null check for `$layoutAttribute` before calling `->getArguments()[0]` as we would get null if we don't have any `Layout` attribute.

Please let me know if there's anything i can improve.

DanielSpravtsev: Looks good. Did you manually test it? In my case I see empty div after page load and nothing changes 😩

gazzoy: Thank you for your review. Hmm, I manually test it and for me it seems works.

gazzoy: @DanielSpravtsev is there anything I can help with?

DanielSpravtsev: @gazzoy Nope. I think we can call @joshhanley for test and approval if everything works as expected.

lrljoe:
Phew, someone else got there before I had to go through numerous forks and code dumps to find where I'd got to.

However - to my mind, the issue with lazily loading a layout, is if there's intermediary determination as to whether the user **should** be able to access the area/section/layout.

E.g. user is on "Generic Homepage" using the "layout.app".

The user clicks on the "Admin Home", which navigates to "Admin Homepage" which uses "layout.admin" (for example).

This new layout is loaded, perhaps with a new navigation set etc, and with the relevant "admin loading placeholder" etc.  However the layout.admin layout may also disclose links/data that the user shouldn't be seeing before authorization.

I'd need to really throw something more at it to clear it up in my mind, or have someone else unpick my insanity!

I personally love the idea of lazily loading in layouts, but I equally want to validate everything/everyone before I start giving them access to anything, security through obscurity being a definite fad, not exposing any endpoints being another matter.

gazzoy:
@lrljoe Thanks for your input! Correct me if I'm wrong, but I belive your concern could be happen on non-lazily loading component as theoritically its a common issue for changing layouts between pages, isn't it?

Let's say we have a login page renders w/ `layouts.app`, and then we'd redirect to dashboad page renders w/ `layouts.admin` after login. Both of pages are implemented as non-lazily loading componnent. Even in that case, I belive your concerns could be occured, so it's not a matter of lazily loading or none-lazily loading.

What do you think?

gazzoy:
@DanielSpravtsev FYI, I've got rebased to main since there was a conflict on `src/Features/SupportLazyLoading/BrowserTest.php`. Please let me know if there's anything I can do.

lrljoe:
So my experience testing this was:

**Step 1** - First Page
Navigate to a page (which is using the default layout)
Check it displays the correct layout
Check it displays the correct placeholder text
After the sleep() period , it displays the normal content ( a wire:navigate link to a second page)

**Step 2** - Second Page (using different layout)
Click on the ink from First Page, the page is set to lazy in the routes file.
The placeholders use the "alt" layout to render"
The nested component loads, using the alt layout.

I did get a few oddities going on, such as placeholders appearing on the wrong screen, or not disappearing cleanly, o probably does need a bit more testing.

gazzoy:
@lrljoe Thanks for testing this. I added a new test on 3d5c3fff946f278bd01039950ce85ecf5e151b26 to test the case you pointed out, but unfortunately I couldn't reproduce the issue. Am I missing something?

calebporzio:
I tweaked the implementation and tests but got the feature working. Thanks everyone!

PR #7557, opened 2023-12-17
Wrote documentation about update hooks for arrays

ramonrietdijk:
I've written some more documentation about the update lifecycle hooks regarding arrays. There are some notes about this in the documentation of the [previous](https://laravel-livewire.com/docs/2.x/lifecycle-hooks#class-hooks) version of Livewire.

I could not find it anywhere in the new version and have added a new section for this.


PhiloNL:
Thanks @ramonrietdijk!

PR #7558, opened 2023-12-17
Name change reset function

fouteox:
In the documentation it is given as an example the addition of a reset() function, but if we take the example as is, we get an error because we are trying to redefine the function present in Component.

To avoid any confusion, I suggest renaming the function in the example to resetData which no longer poses any problems.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -517,7 +517,7 @@ class SearchPosts extends Component
 <div>
     <input wire:model.live="query">
 
-    <button wire:click="resetData">Reset Search</button> <!-- [tl! highlight] -->
+    <button wire:click="resetQuery">Reset Search</button> <!-- [tl! highlight] -->
 
     @foreach ($posts as $post)
         <!-- ... -->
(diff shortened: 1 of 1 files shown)


docs/actions.md
```diff
@@ -497,7 +497,7 @@ class SearchPosts extends Component
     public $query = '';
 
     #[Js] // [tl! highlight:6]
-    public function reset()
+    public function resetData()
```
joshhanley, line 500:
```suggestion
    public function resetQuery()
```

docs/actions.md
```diff
@@ -517,7 +517,7 @@ class SearchPosts extends Component
 <div>
     <input wire:model.live="query">
 
-    <button wire:click="reset">Reset Search</button> <!-- [tl! highlight] -->
+    <button wire:click="resetData">Reset Search</button> <!-- [tl! highlight] -->
```
joshhanley, line 520:
```suggestion
    <button wire:click="resetQuery">Reset Search</button> <!-- [tl! highlight] -->
```

joshhanley (approved):
Thanks, makes sense. I've changed it to resetQuery as it matches the context better than resetData

PR #7561, opened 2023-12-17
Add option to add custom headers to Livewire update request

PhiloNL:
In v2 (https://github.com/livewire/livewire/blob/2.x/js/index.js#L69-L71) it was possible to add custom headers to the update request sent by Livewire. This PR adds back this functionality.

- [x] Add test
- [x] Add documentation

```js
Livewire.addHeaders({
    Authorization: 'Bearer Hello'
});
```

**Update**
When using `wire:navigate` these headers are not applied. Given this is powered by a separate Alpine plugin I've added the same function for the `navigate` plugin. 

```js
Alpine.navigate.addHeaders({ 
    Authorization: 'Bearer foo'
});
```


calebporzio:
Thanks Philo!

Here are my initial reactions to this:
* Can you validate the need for this? Like what is the practical reason you want this added?
* Disregarding what V2 did, I think I would rather see this configurable server-side. As a dev, I'd rather configure this in my service provider like so: `Livewire::addHeaders(...)`
* Is the `wire:navigate` header feature needed? In general, navigate requests are just page fetches, not "livewire" updates. If there's a valid need for them, then I think a better API might be a php one like: `Livewire::addNavigateHeaders(...)`. Kind of a mouthful, but I imagine that is the lesser-needed feature anyways.

Thanks for your work on this, lmk what you think

PhiloNL:
Thanks for the feedback! 😄 

**Can you validate the need for this? Like what is the practical reason you want this added?**
Tldr version: Shopify apps are required to authenticate every request and sessions are not supported. So you get a token once and have to use it for every request.
--
I have a bunch of Livewire v2 apps that are Shopify applications, I know there are more people who use Livewire to build Shopify apps. The way Shopify works is by embedding the app using an iframe. Given sessions no longer work they use these "session tokens" which you have to send along with every request. They kind of push you towards using SPA's and build using React or Vue because that means you get the token and send it along with your API requests.

**Disregarding what V2 did, I think I would rather see this configurable server-side. As a dev, I'd rather configure this in my service provider like so: Livewire::addHeaders(...)**
Maybe allow both?

**Is the wire:navigate header feature needed? In general, navigate requests are just page fetches, not "livewire" updates. If there's a valid need for them, then I think a better API might be a php one like: Livewire::addNavigateHeaders(...). Kind of a mouthful, but I imagine that is the lesser-needed feature anyways.**
The same answer applies here as mentioned above. Shopify requires a SPA and wire:navigate is perfect but I need the option to hook into the request and add the headers.

So if not specifically adding something like `addHeaders`, having an option / callback that allows me to hook into the request and modify the `fetch` params and add headers would be great as well.

calebporzio:
Ok, then for now, let's hold off on this and you can use the following:

image

If you hook into that event like so:

```js
Livewire.hook('request', ({ options }) => {
    options.headers['Your-Header'] = 'Header Value'
})
```

That should work because the options object is passed by reference

I went ahead and added a hook for the navigate side too: (it's on the main branch, and will be tagged in the next release - you can test it and let me know how it works for you)

```js
Livewire.hook('navigate.request', ({ options }) => {
    options.headers['Your-Header'] = 'Header Value'
})
```

calebporzio:
Closing this for now. thanks

PR #7565, opened 2023-12-18
Compile assets from source during tests

PhiloNL:
This will always ensure the Livewire assets are built from the source during tests. The goal is to prevent failing tests that are caused by users not compiling and adding the distribution assets to their PR. This is also meant to take us one step closer to not having `dist` files in the repository itself and only having the distribution files be part of a release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,7 @@
 name: Test
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - "**"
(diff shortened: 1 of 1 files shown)


calebporzio:
Nice, I think I'm in favor of this for PRs but not for the main. Like with this PR unless I miss-understand the following could happen:

- all tests pass on main because assets are built in the CI pipeline
- we tag a release
- we introduce a bug because we didn't commit the re-build assets and someone messed up by commiting not up-to-date assets

Right?

edalzell:
Don't know if this is relevant but I'm a fan of https://github.com/pixelfear/composer-dist-plugin, which downloads built assets so you don't have to commit them. It's used by Statamic and other packages (like all of mine).

calebporzio:
Going to close this for now because of my concerns - thanks for the contribution and for weighing in everyone. Will probably revisit this for the next version of Livewire

PR #7573, opened 2023-12-19
Fix `assertHasErrors`

nuernbergerA:
#7520 has broken validation for class base validation rules

this PR fixes it


calebporzio:
Thanks Adrian!

PR #7576, opened 2023-12-19
Real-time events using Mercure

benbjurstrom:
This pull request adds support for [Mercure](https://mercure.rocks/) real-time events to Livewire. Mercure is a modern alternative to WebSockets that's a little simpler to setup in my opinion. It also ships included with [FrankenPHP](https://frankenphp.dev/) which has been getting some buzz recently.

The integration is fairly lightweight and closely follows the Echo implementation. Once Mercure's been configured as a Laravel Broadcast channel your Livewire component can listen for Mercure events by adding them to getListeners. Here's an example component that calls the `addMessage` method whenever it receives a `public.newMessage` event:

```php
<?php

namespace App\Livewire;

use Illuminate\Support\Carbon;
use Livewire\Component;

class Messages extends Component
{
    public array $messages = [];

    public function getListeners()
    {
        return [
            "mercure:public.newMessage" => 'addMessage',
        ];
    }

    public function addMessage(array $event)
    {
        array_push($this->messages, [
            'time' => Carbon::now()->format('g:i:sa'),
            'message' => $event['data']['payload'],
        ]);
    }

    public function render()
    {
        return view('livewire.messages');
    }
}
```

I also setup a demo repo showing how to configure Mercure as a broadcast channel. You can find the source code for the demo here: https://github.com/benbjurstrom/livewire-mercure-demo.

https://github.com/benbjurstrom/livewire-mercure-demo/assets/12499093/62f8372f-d79c-4e5a-ba46-aacb9792b428

### DIFF
--- a/js/features/index.js
+++ b/js/features/index.js
@@ -12,4 +12,5 @@ import './supportRedirects';
 import './supportNavigate';
 import './supportMorphDom';
 import './supportEntangle';
+import './supportMercure';
 import './supportEvents';
--- a/js/features/supportMercure.js
+++ b/js/features/supportMercure.js
@@ -1,0 +1,53 @@
+import { on } from '@/events'
+import { dispatchSelf } from './supportEvents'
+
+let mercureUrl = null
+
+function getMercureUrl() {
+    if (!mercureUrl && document.querySelector('[data-mercure-url]')) {
+        mercureUrl = document
+            .querySelector('[data-mercure-url]')
+            .getAttribute('data-mercure-url')
+    }
+    return mercureUrl
+}
+
+on('effects', (component, effects) => {
+    const listeners = effects.listeners || []
+
+    listeners.forEach((event) => {
+        if (event.startsWith('mercure')) {
+            const mercureUrl = getMercureUrl()
+            if (!mercureUrl) {
+                console.warn(
+                    "Warning: 'data-mercure-url' attribute not found. Ensure it is present in the HTML.",
+                )
+                return
+            }
+
+            if (typeof EventSource === 'undefined') {
+                console.warn(
+                    'Warning: EventSource API (for Mercure) not available. Check browser compatibility.',
+                )
+                return
+            }
+
+            const eventParts = event.split('mercure:')
+            const [s1, topic] = eventParts
+            const url = `${mercureUrl}?topic=${topic}`
+
+            const mercureEventSource = new EventSource(url, {
+                withCredentials: true,
+            })
+
+            mercureEventSource.onmessage = (e) => {
+                const data = JSON.parse(e.data)
+                dispatchSelf(component, event, [data])
+            }
+
+            mercureEventSource.onerror = (e) => {
+                // console.log(e)
+            }
+        }
+    })
+})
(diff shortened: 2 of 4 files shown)


benbjurstrom:
The tests are failing because I didn't include the `dist/livewire.js` file. When I ran `npm run build` in my development environment there appeared to be a lot of unrelated changes. The tests should pass once an updated copy of the dist folder is included.

joshhanley:
@benbjurstrom this is probably better suited as a userland package to start with, but wait to see what Caleb thinks 🙂 regarding the build, yeah you will need to npm link to a local copy of Alpine and any Alpine packages Livewire uses to build the assets, as we've fixed stuff there but a new release hasn't yet been tagged for Alpine.

benbjurstrom:
@joshhanley that makes sense, at least until Mercure gains more popularity within the community. Any tips for hooking into the listeners from a third party script?

Was trying to replicate what I did in [js/features/supportMercure.js](https://github.com/livewire/livewire/pull/7576/files#diff-c98580036417f84c2ef1feb92ea04c717bdcbc96672240d411804f4eec213e69) by following some of the examples in the [javascript docs page](https://livewire.laravel.com/docs/javascript#using-javascript-in-livewire-components) but haven't made any progress. For example, I never get any console output when trying things like this:

```js
    window.Livewire.on('effects', (component, effects) => {
        const listeners = effects.listeners || []

        listeners.forEach((event) => {
            console.log(event)
        })
    })
})
```

joshhanley:
@benbjurstrom hmm I would expect that to work the same way. Have you checked that the `livewire:init` is being picked up correctly and the closure is actually firing?

benbjurstrom:
I think it's initialized correctly. Using the code below I get "foo", but I never get "bar." Unless I manually dispatch an event called "effects", but that obviously would't include the listeners.

```js
document.addEventListener('livewire:init', (event) => {
    console.log('foo')

    window.Livewire.on('effects', () => {
        console.log('bar')
    })
})
```

calebporzio:
Hey @benbjurstrom, really nice stuff! I've had my eye on symfony mercure for a while and think it could be a great pairing with Livewire.

I agree with Josh here though that I think this should be a userland package for now until there's wider adoption and a clearer path to usage.

As for why your snippet's not working, it's be cause you need to be using `Livewire.hook()` instead of `Livewire.on()` see: https://livewire.laravel.com/docs/javascript#javascript-hooks

I'm so stoked that octane supports frankenphp now, which in turn, supports mercure. Getting close to a solid real-time solution in laravel apps!

calebporzio:
Going to close this, but I hope to see more of what comes of this!

benbjurstrom:
Thanks @calebporzio! Was able to get it working as a 3rd party script. Appreciate all the work you put into Livewire.

@edalzell you should be able to add this to your app without a package pretty easily. I updated the demo repo linked above with an example implementation where the necessary JS is loaded from [/resources/js/supportMercure.js](https://github.com/benbjurstrom/livewire-mercure-demo/blob/main/resources/js/supportMercure.js)

PR #7579, opened 2023-12-20
Fix `class "string" not found` bug

joshhanley:
This PR fixes an issue #7574 introduced in v3.3.1 where it was trying to hydrate all strings from the query string as backed enums.

image


edwinvdpol:
Thanks!

PR #7628, opened 2023-12-29
Add dark mode for default tailwind pagination links

sabinchacko03:
Currently Livewire Pagination Link doesn't include css class for dark mode. I have added it for the default tailwind links.


joshhanley:
Looks like Laravel's default pagination has darkmode support, so I believe we should also https://github.com/laravel/framework/blob/10.x/src/Illuminate/Pagination/resources/views/tailwind.blade.php

calebporzio:
I think this should be added because Laravel added it to theirs. However, Laravel's is more comprehensive.

I'm gong to close this - if you have a chance please re-PR with all dark mode changes from Laravel's views. Thanks!

sabinchacko03:
I've added it to Laravel.

PR #7629, opened 2023-12-29
Fixed doulbe 'has' during upgrade

Bird87ZA:
I just fixed a grammatical error on the update script. Don't think tests etc are needed. :)


PhiloNL:
Thanks!

PR #7633, opened 2023-12-29
fix spelling of "Libraries"

browner12:
Review the contribution guide first at: https://livewire.laravel.com/docs/contribution-guide

PR #7639, opened 2023-12-29
fix spelling of "prefetched"

browner12:


PhiloNL:
https://github.com/livewire/livewire/pull/7633#issuecomment-1873049048

PR #7642, opened 2023-12-29
fix `livewire.js` reference

browner12:
the `livewire.js` file lives at `/livewire/livewire.js`, as referenced previously.


PhiloNL:
https://github.com/livewire/livewire/pull/7633#issuecomment-1873049048

PR #7644, opened 2023-12-30
Fix rules from component are ignored when a form object without rules is used

AntonioPrimera:
Fixed: when rules are provided in component class, and a form object is used, with no rules available, the component rules are forgotten - they should be merged (array_merge expects multiple arrays to be merged).

It looks like a simple bug array_merge() with a single array given as an argument does not do anything, it just returns the provided array.

yes

no - just a single line was changed

No unit tests were included - the fix is very basic. Did manual tests in the project where I discovered the bug, and everything works like before. Also, when running unit tests locally, several errors occur after 567 successful tests and for the time being it seems like a waste of time to find the issue at the moment, which is probably some Dusk / Chrome / OSX setup issue.

I am implementing a complex project using livewire 3 and I just ran a composer update, so the livewire package was updated to v3.3.3. After the update, I got an error that no $rules / rules() were provided for a form. My form was working correctly before, and there is a public rules() method on the Form component. The Component uses a FormObject, which has no $rules or rules() property / method. When debugging, I noticed the following bug in the "HandlesValidation.php" trait:

foreach ($this->getFormObjects() as $form) {
    $allRules = array_merge($form->getRules());
}

This overrides the $allRules for every form object, with the rules extracted from the form object. Instead, it should state:

foreach ($this->getFormObjects() as $form) {
    $allRules = array_merge($allRules, $form->getRules());
}

After updating it locally, my forms run like before, so you can say that I manually tested the change.

Thanks for contributing! 🙌


AntonioPrimera:
It seems that the problem is still not fixed. There are a few other places in the same trait, where an array_merge() with a single array is used. Even after fixing all of them, I still get the same error. Seems like the providedOrGlobalRulesMessagesAndAttributes(...) method is run twice when submitting the form. First time it has all the provided rules in $allRules, the second time $allRules is empty.

I'm still looking into it, but being new to the Livewire internals it takes some time.

luanfreitasdev:
I reproduced this issue and it seems to have fixed it by just adding this additional validation. From the looks of it, whenever you have a form inside a component, the rules will be called twice, so array_merge won't work because it doesn't maintain state.

`Livewire\Features\SupportValidation\HandlesValidation::261
`

**before**:
```php
if ($this->isRootComponent() && $isUsingGlobalRules) {
// ...
```

**after**:
```php
if ($this->isRootComponent() && $isUsingGlobalRules && filled($this->rulesFromOutside)) {
// ...
```
Summary, simply check if $this->rules From Outside exists before calling the form class's validate

calebporzio:
Ok, I did some digging on this. I tweaked the implementation and added tests. Will be in the next release. Thanks!

PR #7647, opened 2023-12-31
Improvement: Support for Wireable casting on Model binding

LuanHimmlisch:
Currently, if you want to work with a custom casting with model binding, you can't. Even if the value is being cast to a class that  has the  `Wireable` interface implemented.

This is because the `unwrapDataForValidation` function hasn't taken into consideration (neither on V2 nor V3) the properties of a model. Only casting the first depth of the values.

```php
protected function unwrapDataForValidation($data)
{
    return collect($data)->map(function ($value) {
        if ($value instanceof Wireable) return $value->toLivewire();
        else if ($value instanceof Arrayable) return $value->toArray();

        return $value;
    })->all();
}
```

## The problem

Having a custom class, with Wireable implemented, as it's described on docs:

```php
class CommaList implements Wireable
{
    // ...
    {
        return new static($value);
    }

    public function toLivewire(): string
    {
        return $this->toString();
    }
}
```

And a casting for that class:

```php
class AsCommaList implements Castable
```

Being implemented on a Model:

```php
protected $casts = [
    'tags' => AsCommaList::class
];
```

Does throw a validation exception:

```php
public $rules = [
    'post.tags' => ['required', 'string', 'max:255'],
];
```

image

This is because the Wireable function is never called.

image

## The solution

This PR extracts the casting to its own recursive function, to cast any attribute from a Model.

```php
protected function castDataForValidation($value)
{
    // @todo: this logic should be contained within "SupportWireables"...
    if ($value instanceof Wireable) {
        return $value->toLivewire();
    }

    if ($value instanceof Model) {
        return array_map(
            fn ($attribute) => $this->castDataForValidation($attribute),
            $value->toArray()
        );
    }

    if ($value instanceof Arrayable) {
        return $value->toArray();
    };

    return $value;
}

protected function unwrapDataForValidation($data)
{
    return collect($data)
        ->map(fn ($v) => $this->castDataForValidation($v))
        ->all();
}
```
The corrected result:

image

### DIFF
--- a/src/Features/SupportValidation/HandlesValidation.php
+++ b/src/Features/SupportValidation/HandlesValidation.php
@@ -496,15 +496,32 @@ trait HandlesValidation
         return Utils::getPublicPropertiesDefinedOnSubclass($this);
     }
 
+    protected function castDataForValidation($value)
+    {
+        // @todo: this logic should be contained within "SupportWireables"...
+        if ($value instanceof Wireable) {
+            return $value->toLivewire();
+        }
+
+        if ($value instanceof Model) {
+            return collect($value->toArray())
+                ->map(
+                    fn ($attribute) => $this->castDataForValidation($attribute)
+                )->toArray();
+        }
+
+        if ($value instanceof Arrayable) {
+            return $value->toArray();
+        };
+
+        return $value;
+    }
+
     protected function unwrapDataForValidation($data)
     {
-        return collect($data)->map(function ($value) {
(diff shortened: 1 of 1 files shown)


joshhanley:
@LuanHimmlisch thanks for the PR! Please add tests for this. Also can you confirm, how are you using model binding in V3 as we disabled it? Have you re-enabled it? Thanks!

PR #7649, opened 2023-12-31
fix computed attribute - snake case in method name

luanfreitasdev:

N
Y
N
Y

Fix: https://github.com/livewire/livewire/pull/7426#issuecomment-1873020792

The problem was generated when searching for the value of the method that was overwritten. Now, the value is taken based on the original name of the loaded method.

```php
#[Computed]
public function some_long_property_name_thats_easier_to_read_as_snake_case(): string
{
     return "Yeah!!!";
}
```

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -133,7 +133,7 @@ class BaseComputed extends Attribute
 
     protected function evaluateComputed()
     {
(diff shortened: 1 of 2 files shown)


shawnveltman:
Wow that was quick - you are an absolute legend!!  Thank you so much for doing this.

calebporzio:
Thanks!

PR #7641, opened 2023-12-29
Fix Testing - SubsequentRender response - effects payload does not work correctly

luanfreitasdev:
Review the contribution guide first at: https://livewire.laravel.com/docs/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)
Y
3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
N
4️⃣ Does it include tests? (Required)
Y
5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

This PR improves the way test response is handled. Instead of getting the components via the json() method, it is captured by the "original" property.

This is because `json()` doesn't handle some returns well.
I also removed a line that would cause the test to error.

I submitted a failed test to secure PR

---

Here is a comparison:

### response->json()
```php
array:2 [ // ...livewire/src/Features/SupportTesting/SubsequentRender.php:53
  "components" => array:1 [
    0 => array:2 [
      "snapshot" => "{"data":[],"memo":{"id":"ffsGZ5S8JXeFduvL68gU","name":3207924939,"path":"livewire-unit-test-endpoint\/QmwXjJiNtLJl1ENjBWlK","method":"GET","children":[],"scripts":[],"assets":[],"errors":[],"locale":"en"},"checksum":"9c1601139aa55f54c79f1542e5d1c2261f48c9328a45bacce47f99459a9c4e4c"}"
      "effects" => array:3 [
        "returns" => array:1 [
          0 => null
        ]
        "html" => "<example wire:id="ffsGZ5S8JXeFduvL68gU">Hello!</example>"
        "dispatches" => array:1 [
          0 => array:2 [
            "name" => "my-event"
            "params" => array:1 [
              "payload" => array:3 [
                "value_1" => 1
                "value_2" => 3.45
                "value_3" => 40
              ]
            ]
          ]
        ]
      ]
    ]
  ]
  "assets" => []
]

```

### response->original
```php
array:2 [ // .../livewire/src/Features/SupportTesting/SubsequentRender.php:53
  "components" => array:1 [
    0 => array:2 [
[...]

PR #7651, opened 2024-01-01
Pass the same parameter format in assertDispatch to avoid divergent comparisons

luanfreitasdev:

https://github.com/livewire/livewire/pull/7641

Y
N
Y

When checking whether an event has been triggered, sometimes we have no way of knowing what format it is or what data will be lost due to the `$response->json()` method that converts the payload to javascript format.

This PR follows this same idea.

**Today it looks like this:**
* User dispatches an event with a carbon date format

```php
$this->dispatch('my-event', payload: [
      'datetime' => Carbon::create(2023, 12, 1, 10, 10, 58)
]);
```

* The test does not validate whether the "Carbon" date the user passed is the same as `2023-12-01T10:10:58.000000Z `compared in the test.

image

---

**Ideal**:

* User dispatches an event with a carbon date format
* The test can check if a Carbon instance is the same as the one sent back by the response. (javascript). Then these two will pass:

```php
->assertDispatched('my-event', payload: [
      'datetime' => Carbon::create(2023, 12, 1, 10, 10, 58)
 ])
 ->assertDispatched('my-event', payload: [
      'datetime' => '2023-12-01T10:10:58.000000Z'
 ])
```

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportEvents/TestsEvents.php
+++ b/src/Features/SupportEvents/TestsEvents.php
@@ -48,7 +48,6 @@ trait TestsEvents
     protected function testDispatched($value, $params)
     {
         $assertionSuffix = '.';
-        $params = json_decode(json_encode($params), true);
 
         if (empty($params)) {
             $test = collect(data_get($this->effects, 'dispatches'))->contains('name', '=', $value);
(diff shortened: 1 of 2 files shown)


calebporzio:
I think this is a breaking change, so holding off. Thanks

PR #7660, opened 2024-01-02
[3.x] Reverts PHPUnit attributes

nunomaduro:
This pull request reverts PHPUnit attributes introduced earlier today.


crynobone:
Do note that using PHPUnit annotation will start showing deprecation notices in PHPUnit 11.

PR #7662, opened 2024-01-02
Give computed properties precedence over standard properties in the view

devajmeireles:

N
Y
N
Y

When we have two properties with the same name in the same component, but one of them is computed, and we are accessing the property using `$this` through Blade, Livewire is not prioritizing the computed property over the common property, when in my opinion it should prioritize the computed property due to the fact of using `$this` to access the property like mentioned in the docs: https://livewire.laravel.com/docs/computed-properties


calebporzio:
I think this is a breaking change so I'm going to close. thanks

PR #7667, opened 2024-01-03
Change `#[On]` attribute to support being defined at the component level

gdebrauwer:
When I upgraded a Livewire project from v2 to v3, I wanted to remove all the `$listeners` and use the new php attribute syntax.

In my project, most of my listeners just trigger a refresh of a component when an event is fired from a (non-child) component (e.g. [wire-elements/pro](https://wire-elements.dev/pro) modal or slide-over). Those kind of listeners can currently not be converted to php attributes, because you can not add them to a method (or you would have to create a dummy empty method which seems a bit weird).

```php
// Before

{
    protected $listeners = [
        'foo' => '$refresh',
        'bar' => '$refresh',
    ];
}
```

This PR fixes that by allowing the `#[On('eventName')]` php attribute on a Livewire class. That feels a lot more like Livewire v3 and also reduces the amount of lines in your Livewire class 😄 

```php
// After

#[On(['foo', 'bar'])]
{
}
```


devajmeireles:
Why not only `#[Refresh]`?

gdebrauwer:
'refresh on' seems a bit clearer in my opinion

levu42:
this looks awesome! When I migrated a bigger project from v2 to v3, I created those empty dummy methods, which indeed look weird.

joshhanley:
@gdebrauwer makes sense to me that we should have a way to call `$refresh` from an attribute listener.

Could we just change how `#[On()]` is handled so if it's placed on the class definition it just refreshes the component?

Like

```php
#[On('foo')]
{
}
```

levu42:
@joshhanley that indeed looks like the cleanest solution, I like that one more than `#[RefreshOn]`

gdebrauwer:
@joshhanley Good idea! I will change it
(btw, it was awesome meeting you at LaraconAU 😄)

joshhanley:
@gdebrauwer great thanks! Was great to meet you too 😁

calebporzio:
Looks good to me, thanks!

PR #7670, opened 2024-01-04
Add warning about Laravel Mix

browner12:
it appears that Laravel Mix will not generate working code if you want to manually bundle Livewire and AlpineJS, so we'll add a note to warn users about this.

I banged my head for 2 days trying to figure this out, so hopefully this helps other people in the same boat.


calebporzio:
Thanks for this!

PR #7671, opened 2024-01-04
Failing test: reordering the keys of an object with entangle

binaryfire:
When reordering the keys of an object with $entangle, or setting the entire object, Livewire doesn't actually overwrite the old object, so the keys remain in the same order.

The problem is reproduced here: https://wirebox.app/b/y8232

This bug is breaking the sortable functionality of the Filament Key-Value field. See: https://github.com/filamentphp/filament/issues/9030

Dusk doesn't work on my system because of my custom headless Chrome setup, which means I couldn't test this PR locally. So a maintainer might need to be double-check / tweak it.


calebporzio:
Ok, I dug into this, here's what's happening:

* Livewire V2 has a very primitive data tracking system. It's worse in many ways, but better at preserving data update integrity.
* In V3, when `$wire.$set` is used, it actually manipulates the component's in-memory JS data.
* When a request is triggered, before it goes to the server a "diff" is taken of the last known server state and the current JS state
* That diff is sent to the server to make the updates.

The reason Livewire isn't honoring the re-order of keys is:
* When the diff is taken, it doesn't pay attention to key order, it just looks for added and removed keys and makes a diff entry for those.

I fixed this problem in this PR by introducing a system so that when you call `$wire.$set`, the data you pass in is both added to the JS in-memory object AND queued as a full update for the server.

This could be the ticket, but it means the Alpine side isn't fixed, I commented out those tests for now.

This is something we can keep an eye on in the future and decide if we want to handle it in a deeper way.

The deeper solution is to change how V3's data diffing algorithm works and make it key-order aware. This will likely result in slower diffs and bigger payloads. But clearly the benefit is better key order preservation.

Hopefully this fix gets you where you need to go and we can explore further when those problems arise.

Thanks!

binaryfire:
@calebporzio Thank you! And thanks for the explanation🙂 

> The deeper solution is to change how V3's data diffing algorithm works and make it key-order aware. This will likely result in slower diffs and bigger payloads. But clearly the benefit is better key order preservation.

Performance and payload size are critical IMHO, so if this solves things without needing to slow anything down then awesome.

VitorTS:
this commit feature has a bug, see https://github.com/livewire/livewire/discussions/7872

PR #7673, opened 2024-01-04
Allow dispatches to many components

luanfreitasdev:

N
Y
N
Y

Allows you to add other components that will receive the event. This is useful when we want to trigger the same event for several components passing an array as a parameter

**before**:
```php
$this->dispatch('$refresh')->to(QuickView::class);
$this->dispatch('$refresh')->to(Index::class);
$this->dispatch('$refresh')->to(Edit::class);
```

**after**:
```php
 $this->dispatch('$refresh')->to(QuickView::class, Index::class, Edit::class);
```

* The change does not change the current behavior of single components and also works with `assertDispatchTo`

Thanks for contributing! 🙌


calebporzio:
I think the amount of code changes doesn't justify the value. I'm personally fine with the duplicated lines. It keeps the API streamlined and also makes chaining not confusing for multiple targets. Thanks anyways

PR #7669, opened 2024-01-03
Improve console error message

luanfreitasdev:
Review the contribution guide first at: https://livewire.laravel.com/docs/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?
N
2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed)
Y
3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
N
4️⃣ Does it include tests? (Required)
N
5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

---

Improves the display of error messages to show the component name and adds the element for the user to click and go to the element.

This makes it easier for the developer to understand which component or element is having a problem among many.

image

Thanks for contributing! 🙌

PR #7675, opened 2024-01-04
Improve console error message

luanfreitasdev:

N
Y
Y
N

Closed: https://github.com/livewire/livewire/pull/7669

Note: I added an "el" to findComponent to try to find who the component's father is. This comes from return `childIds.map(id => findComponent(id, this.el))`

image

Thanks for contributing! 🙌


luanfreitasdev:
In another test on a larger project, this would be very useful to show which component a problem occurred.

**before:**
image
**after:**
image

luanfreitasdev:
Update:

In my last commit, remove the throw in the initComponent (store.js) check and in the new Component class, returning the instance itself.

```php
//components.js
if (el.__livewire) return el.__livewire;

//store.js
if (components[component.id]) return component
```

luanfreitasdev:
I added the ability to identify the component's name with the issue instead of the ID. To do this, I needed to add the component name to memo.children.

In this case, within 'estimates.show.specifications.production.edit' there may be several child components, if one gives an error, showing only the ID would not help

image

calebporzio:
I see the usefullness here, however there are fundamental changes mixed in with this PR: for example, instead of throwing an error when a component is not found you're just not throwing the exception now and returning null.

I think there might be a PR here but would have to be cleaner and more minimal than this.

If you want, reference how I throw exceptions in Alpine, similar to this. Thanks

luanfreitasdev:
This happened because I saw the need to change the PR behavior in the same scope. In other words, “improve developer usability”

Alpine removes the component, but it doesn't happen in the child snapshot. However, I think it should also be removed to avoid the error on the console, which did not happen in version 2.

Are there any changes to removing the child ID when Alpine removes it from the livewire lifecycle?

tvarwig:
@calebporzio and @PhiloNL  can we discuss this again? Due to issues i am having with livewire this would be a life saver.  [discussions/9267](https://github.com/livewire/livewire/discussions/9267)

PR #7677, opened 2024-01-05
Fix memory leak when adding persistent middleware

timacdonald:
## Why

There is currently a memory leak that occurs in testsuites due to the static nature of the `Livewire\Mechanisms\PersistentMiddleware\PersistentMiddleware::$persistentMiddleware` array.

```php
Livewire::addPersistentMiddleware('MyMiddleware');
Livewire::addPersistentMiddleware('MyMiddleware');
Livewire::addPersistentMiddleware('MyMiddleware');
Livewire::addPersistentMiddleware('MyMiddleware');
Livewire::addPersistentMiddleware('MyMiddleware');

Livewire::getPersistentMiddleware();

// [
//     '\Laravel\Sanctum\Http\Middleware\EnsureFrontendRequestsAreStateful::class',
//     '\Laravel\Jetstream\Http\Middleware\AuthenticateSession::class',
//     '\Illuminate\Auth\Middleware\AuthenticateWithBasicAuth::class',
//     '\Illuminate\Routing\Middleware\SubstituteBindings::class',
//     '\App\Http\Middleware\RedirectIfAuthenticated::class',
//     '\Illuminate\Auth\Middleware\Authenticate::class',
//     '\Illuminate\Auth\Middleware\Authorize::class',
//     '\App\Http\Middleware\Authenticate::class',
//     'MyMiddleware',
//     'MyMiddleware',
//     'MyMiddleware',
//     'MyMiddleware',
//     'MyMiddleware',
// ];
```

This is problematic in a testsuite where static variables are not cleared between runs. If an app was to call `Livewire::addPersistentMiddleware('MyMiddleware')` once in their service provider, as documented, the on every test a new entry would be added to the array.

If a testsuite had 1,000 tests there would be 1,000 `'MyMiddleware'` entries in this array for the last test.

In `laravel/pulse` we add two persistent middleware in the service provider which means this imaginary testsuite would now have 2,000 `'MyMiddleware'` entries. It also means that all non-livewire applications that install Pulse can be impacted by this memory leak.

## The fix

Middleware are now ensured to be unique while being added to Livewires stack. We use Laravel's internal `Router::uniqueMiddleware` to ensure the order matches how Laravel would handle things in the router.

## Notes

There is still a potential weirdness for long lived environments, such as Octane, testsuites, etc., where middleware added dynamically outside of a service provider would not be forgotten for the next request. I'm not sure if this is really a concern as the documentation shows `Livewire::addPersistentMiddleware` being used in a service provider.

It was on my mind so I wanted to share it but I don't know that it needs to be resolved as I can't think of a way that it would be problematic in a real-world scenario.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,8 +11,9 @@
   "require": {
     "php": "^8.1",
     "symfony/http-kernel": "^6.2|^7.0",
-    "illuminate/support": "^10.0|^11.0",
     "illuminate/database": "^10.0|^11.0",
+    "illuminate/routing": "^10.0|^11.0",
+    "illuminate/support": "^10.0|^11.0",
     "illuminate/validation": "^10.0|^11.0",
     "league/mime-type-detection": "^1.9"
   },
--- a/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
+++ b/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Mechanisms\PersistentMiddleware;
 
+use Illuminate\Routing\Router;
 use Livewire\Mechanisms\Mechanism;
 use function Livewire\on;
 use Illuminate\Support\Str;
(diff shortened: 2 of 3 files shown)


calebporzio:
Good catch Tim, thanks! (statics strike again lolol)

timacdonald:
Thanks, Caleb (and yea, man, they'll get ya!)

PR #7686, opened 2024-01-05
Add `#[View(...)]` attribute

gdebrauwer:
When upgrading a project from Livewire v2 to v3, I got a lot of errors on components that did not define a render-method. After some digging I found that in v3 the view filename is determined based on the component alias instead of the component class namespace (which was the case in v2). That meant I had to add a render-method to all those components to point livewire to the correct view. 

```php
// Before

{
   public function render() : View
   {
      return view('foo');
   }
}
```

That is a lot of lines of codes for such a basic thing, so why not provide a php attribute to do this? That's a lot less lines of codes 😄 

```php
// After

#[View('foo')]
{
}
```

### DIFF
--- a/src/Attributes/View.php
+++ b/src/Attributes/View.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Attribute;
+use Livewire\Mechanisms\HandleComponents\BaseView;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+class View extends BaseView
+{
+    //
+}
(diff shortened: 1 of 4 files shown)


devajmeireles:
I don't know what Caleb thinks, but I wonder if we're adding too much. Using this PR as an example, `render` is so... enough! And as we know, we can omit `render`, based on the component naming convention.

gdebrauwer:
The Layout attribute exists already, so I think it makes sense to also have a View attribute.

calebporzio:
Mmm yeah I seee the usefulness, but I think I'm going to say no FOR NOW. This isn't me saying this is a bad feature that shouldn't be added at some point. Consider it on my radar now - thanks for being willing to contribute!

PR #7688, opened 2024-01-06
Add Wireable support for #[Url] properties

mattfawcett:
Support using more complex `Wireable` objects when using `#[Url]` properties.

I haven't created a discussion for this. It does seem to make sense to me though that if a more complex object can be used for serialization across the wire using the Wireable interface, and using the `Livewire\Attributes\Url` attribute causes such properties to be included in the url, then these objects should be built properly when a request comes in with them in the querystring. Currently if you try this, the application will error.

2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed): yes

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.: no

4️⃣ Does it include tests? (Required): yes

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportAttributes/Attribute.php
+++ b/src/Features/SupportAttributes/Attribute.php
@@ -69,6 +69,10 @@ abstract class Attribute
             $value = $enum::from($value);
         }
 
+        if ($wireable = $this->tryingToSetToWireable($value)) {
+            $value = $wireable::fromLivewire($value);
+        }
+
         data_set($this->component, $this->levelName, $value);
     }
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for the pull request. Couple thoughts:

* I added enum support as a one-off addition in the code to prevent extra complexity.
* I realized at some point we may want a better solution that doesn't feel like a bunch of conditionals and somehow uses the synthesizers system instead

That to say, I don't think I'm going to merge this as-is. 

Do you or anyone else have insight or ideas about how we can support Wireables and othere types in URL easily without adding a bunch of extra code to this same place?

What are the potentially desired types? datatimes? stringables? etc.? I think we should either add support for all or none honestly.

Thanks for the contribution, let's keep talking about it!

mattfawcett:
Hi, thanks for the feedback. I am pretty new to using Livewire and wasn't really aware of the concept of synthesizers until now, Wireables had been meeting my requirements. I agree that it probably makes sense to support them as well.

I don't really have enough experience with the framework yet to be able to have much of an insight into how best to handle that at the moment. Feel free to close this as you see fit.

Thanks.

calebporzio:
Yeah, I think I'm going to close this for now. Let's wait and gauge demand for this feature, then re-open it or create a new PR when that time comes.

PR #7689, opened 2024-01-06
Fix: dispatch 'upload:removed' after deleted file in array property

luanfreitasdev:
When `$removeUpload` is executed, an event is only fired when the file has been found. The same should happen for multiple uploads

The ideal would be to fire the 'upload:removed' event only when at least one file is found

### DIFF
--- a/src/Features/SupportFileUploads/UnitTest.php
+++ b/src/Features/SupportFileUploads/UnitTest.php
@@ -102,6 +102,23 @@ class UnitTest extends \Tests\TestCase
     }
 
     /** @test */
+    public function cant_remove_a_file_from_an_array_of_files_property_with_mismatched_filename_provided()
+    {
+        $file1 = UploadedFile::fake()->image('avatar1.jpg');
+        $file2 = UploadedFile::fake()->image('avatar2.jpg');
+
+        $component = Livewire::test(FileUploadComponent::class)
+            ->set('photos', [$file1, $file2]);
+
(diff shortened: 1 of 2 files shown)


calebporzio:
Can you add a deeper explanation for this? I don't know the potential impacts without hearing more about how the current behavior compares to this. Thanks!

luanfreitasdev:
Sure!. A simple example would be to add an invalid name when calling the _removeUpload method when working with multiple files, then callback will be called anyway without checking. This already works with a simple upload

luanfreitasdev:
Bug:

- I select 2 files
- I try to remove one, but the name is invalid
- The function executes the call back even when doing nothing

Bug fix, it would look like this:

- I select 2 files
- I try to remove one, but the name is invalid
- The function is not executed because there is no valid file

https://github.com/livewire/livewire/assets/33601626/b64656b1-0df6-4629-8840-0ade8fcbe6e1

calebporzio:
Hey @luanfreitasdev - I'm not sure I consider this a bug. With the fix you've contributed, it would break behavior because it would dispatch a  "uploadd:removed" event multiple times instead of once, which it currently does. This would cause other front-end issues.

I'd say, just make sure the names you're passing into "removeUpload()" are valid.

You can re-submit with a better implementation and better explanation about why I'm wrong lol. Thanks!

luanfreitasdev:
Hello @calebporzio In my implementation, the dispatch is made only once after the file is deleted, the same as in the single upload implementation.

PR #7690, opened 2024-01-06
Add dependency injection to computed properties

luanfreitasdev:

https://github.com/livewire/livewire/discussions/6967

Y
N
Y

Fix the use of computed property with dependency injection when using Computed Attribute

```php
#[Computed]
public function fooBar(FooDependency $foo)
{
    return $foo->baz;
}
```

### DIFF
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -2,12 +2,12 @@
 
 namespace Livewire\Features\SupportComputed;
 
-use function Livewire\invade;
 use function Livewire\on;
 use function Livewire\off;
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Makes sense, thanks!

calebporzio:
Hey @luanfreitasdev - I had to revert this because it broke using protected methods as computed properties. If you still want the fix, please account for that, then re-PR

luanfreitasdev:
Hello @calebporzio. Using a computed property similar to version 2, it is not possible to access it if it is protected. And no document talks about this. Reverting this PR confuses concepts or should we allow access to it too?

image

PR #7693, opened 2024-01-07
Fix error in sample code div → form

mboynes:
A titled, there is a minor error in the docs where a `div` was used that should be a `form`.


PhiloNL:
Thanks!

PR #7699, opened 2024-01-08
Update components.md

ijpatricio:
Hey all,

PS: I don't know how to create this on a new branch, I had to select `main`. Sorry!

Credits to @nuernbergerA saved me on this one!

I started recently using `@livewire` blade directive for better refactoring, and I was hours and hours trying to debug a weird mess.

Turns out, when we use `@livewire`, the "key" passed as 3rd argument must be a `key($foo)`.

I was passing `$foo` only. Couldn't find in the docs.

I put it here, but let me know if you want it somewhere else.

Thanks!

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -207,6 +207,16 @@ For example, if you are looping through an array of posts, you may set the `wire
             <!-- ... -->
         </div>
     @endforeach
+
+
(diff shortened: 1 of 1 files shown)


PhiloNL:
Thanks! Added the blade component syntax as well.

ijpatricio:
Nice, thanks! Nice to find you here, didn't know!

PR #7716, opened 2024-01-11
Move middleware definitions to static controller methods

stancl:
This PR updates middleware definitions in `FileUploadController` and `FilePreviewController` to be done using static methods.

The reason for this is to support Livewire v3 file uploads with multi-tenancy (specifically the Livewire integration we have in the upcoming version 4 of Tenancy for Laravel).

It's a bit hard to explain why this matters without getting into very low level details of multi-tenancy, but in short, if Laravel can get the middleware list for a controller *statically*, it will avoid instantiating it until later in the request lifecycle. We have updated Tenancy v4 to also read middleware (which can be used for tenant identification) in this way.

Therefore, it's preferable to define middleware using the static `middleware()` method defined by the (relatively new) [`HasMiddleware` interface](https://github.com/laravel/framework/pull/44516), instead of using the (old) `getMiddleware()` instance method.

The changes in this PR:
1. Removed `middleware()` calls on the `Route` definitions in `SupportFileUploads` (I have moved the `web` group to the middleware definitions used in the controllers, though this change can be reverted if you don't like it)
2. Replaced `hasMiddleware()` with `static middleware()` in `FileUploadController`
3. Made `FileUploadController` middleware configurable using a static property (I didn't want to introduce a new config key, but if you prefer that I can make the change)

Note on 1: only reason I did that is so that the `$middleware` property in `FilePreviewController` isn't empty. It makes more sense to me to put all the middleware in one place, rather than applying `web` on the `Route::...` definition, and applying additional middleware in the controller. But I'm happy to revert that if you don't like it, it's not required for the tenancy support here.

---

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first? Related to multi-tenancy support, previous PR https://github.com/livewire/livewire/pull/6549

2️⃣ Did you create a branch for your fix/feature? **Yes**

3️⃣ Does it contain multiple, unrelated changes? **No**

4️⃣ Does it include tests? **No**

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportFileUploads/FilePreviewController.php
+++ b/src/Features/SupportFileUploads/FilePreviewController.php
@@ -3,6 +3,7 @@
 namespace Livewire\Features\SupportFileUploads;
 
 use Illuminate\Routing\Controllers\HasMiddleware;
+use Illuminate\Routing\Controllers\Middleware;
 use Livewire\Drawer\Utils;
 
 class FilePreviewController implements HasMiddleware
@@ -11,7 +12,7 @@ class FilePreviewController implements HasMiddleware
 
     public static function middleware()
     {
-        return static::$middleware;
+        return array_map(fn ($middleware) => new Middleware($middleware), static::$middleware);
     }
 
     public function handle($filename)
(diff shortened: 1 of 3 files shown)


stancl:
CI failure appears to be some sort of a race condition unrelated to the changes here.

calebporzio:
Makes sense to me @stancl - thanks for the contribution!

PR #7723, opened 2024-01-11
Update `dump` and `dd` testing helpers to support writing the dumped HTML to disk

emmadesilva:
## Abstract

Updates the Livewire testing helpers to support writing the HTML to disk using an optional parameter. This makes it easier to debug things as you can open them up in a browser.

### Example

The following would save the dumped state to `example.html` in the base path.

```php
<?php

use App\Livewire\MyComponent;
use function Pest\Livewire\livewire;

test('Livewire component', function () {
    livewire(MyComponent::class)
        ->dd('example');
});
```

### Tests

I could not find any tests for these helpers, so I didn't include any. If you still want me to add tests, let me know.

### DIFF
--- a/src/Features/SupportTesting/Testable.php
+++ b/src/Features/SupportTesting/Testable.php
@@ -241,15 +241,23 @@ class Testable
         return $this->lastState->getComponent();
     }
 
-    function dump()
+    function dump(?string $toFile = null)
     {
         dump($this->lastState->getHtml());
+
+        if ($toFile) {
+            $this->writeDumpedFile($toFile);
+        }
 
         return $this;
     }
 
-    function dd()
+    function dd(?string $toFile = null)
     {
+        if ($toFile) {
+            $this->writeDumpedFile($toFile);
+        }
+
         dd($this->lastState->getHtml());
     }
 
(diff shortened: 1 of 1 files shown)


calebporzio:
This is interesting. I think I'm going to close this as there's no precedence in Laravel as far as I can tell. Please re-submit the PR if I'm wrong, but I don't want to introduce a surprising API to our users. Thanks.

caendesilva:
> This is interesting. I think I'm going to close this as there's no precedence in Laravel as far as I can tell. Please re-submit the PR if I'm wrong, but I don't want to introduce a surprising API to our users. Thanks.

Hey @calebporzio! There are actually quite a few places in Laravel where something like this is used. Here's some just to name a few from the docs:
- https://laravel.com/docs/10.x/collections#method-dd
- https://laravel.com/docs/10.x/queries#debugging
- https://laravel.com/docs/10.x/http-client#dumping-requests
- https://laravel.com/docs/10.x/notifications#inspecting-slack-blocks

Looking at the source code there are even more undocumented ones, like `dump` and `dd` functions in the `Stringable` class. There's also a Laravel trait called `Debugging` with similar helpers. So I think the precedence is quite established in my opinion.

Also to note, I've been using the added helpers in this PR and have found it to be incredibly useful in quickly creating Livewire tests.

nuernbergerA:
@caendesilva it's not the fact to have dd or dump (which we already have) it's about the method signature.

Not that the underlying class is [macroable](https://laravel.com/docs/10.x/http-client#macros) so you can add this function easily yourself even as a package.

Cheers

PR #7735, opened 2024-01-15
Fix livewire:move command

SRWieZ:
Enhanced livewire:move command

- Fix Test renaming by updating the contents 
- Added prompt with a search function for better DX


calebporzio:
Thanks for the addition. A few notes for you for next time:

* I stripped out the prompts addition stuff because even though it's better, it makes the move command inconsistent with the other commands. I have plans to convert everything to prompts for the next version of Livewire. As a general note: it's more helpful when PR's are scoped tightly and don't couple multiple changes together.
* The code styling for things like "if" statements was inconsistent with the rest of the codebase - it's helpful when it's consistent
* There was some sloppiness like unused "use" statements that were added. Sometimes just checking through your own PR changes before submitting can be helpful to catch things like this.

Overall thank you for the contribution!

SRWieZ:
Thanks for your feedback!

PR #7744, opened 2024-01-18
Pass `hashName` thru for testing

jasonmccreary:
I was unable to test simply storing an uploaded file exactly as demonstrated in the [Laravel docs](https://laravel.com/docs/10.x/http-tests#testing-file-uploads).

This is because Livewire "reuploads" the uploaded file. This doesn't really matter in live mode, but in test mode it leads to unpredictable behavior as the uploaded file is not what you set in your test. More specifically, its `hashName` does not match. So methods like `store` end up using a different filename than expected.

More experienced Livewire developers may be aware of workarounds, but as a noob to Livewire this burnt my afternoon. So feel free to close this if I missed something. Otherwise…

This PR contains a proposed solution for ensuring Livewire preserves the `hashName` of the `UploadedFile::fake()`, thus leading to easier, more predictable testing which aligns with the examples outlined in the Livewire/Laravel docs. This solution  is similar to how the file size is preserved.

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -156,6 +156,15 @@ class TemporaryUploadedFile extends UploadedFile
         return $hash.$meta.$extension;
     }
 
+    public function hashName($path = null)
+    {
+        if (app()->runningUnitTests() && str($this->getfilename())->contains('-hash=')) {
+            return str($this->getFilename())->between('-hash=', '-')->value();
+        }
+
+        return parent::hashName($path);
+    }
+
     public function extractOriginalNameFromFilePath($path)
     {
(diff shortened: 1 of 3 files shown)


jasonmccreary:
I'm going to assume since Laravel 11 isn't stable, these failures are not caused by my changes. 😅

calebporzio:
This is great @jasonmccreary, just like "-size". Thanks for taking the time to fix this for others 👍 

(And yeah Laravel 11 broke some tests recently - I'm taking it off the CI build for now)

jasonmccreary:
@calebporzio, yeah, I wonder if there's an opportunity to refactor this in "test mode". It seems there is a lot of passing down the _size_, _original filename_, and now _hashName_.

Since `TemporaryUploadedFile` is a child of `UploadedFile`, if a test uses an `UploadedFile` that it could simply be downcast or delegated, instead of recreated. Again, I don't know the importance of simulating the JS uploading events in test mode. But that's where things were getting lost in translation.

PR #7748, opened 2024-01-18
Add #[Session] attribute

calebporzio:
## Basic usage

Here's a `ShowPosts` component that allows users to filter visible posts by a string stored in a `$search` property:

```php
<?php

use Livewire\Component;
use App\Models\Post;

{
    #[Session]
    public $search;

    protected function posts()
    {
        return $this->search === ''
            ? Post::all();
            ? Post::where('title', 'like', '%'.$this->search.'%');
    }

    public function render()
    {
        return view('livewire.show-posts', [
            'posts' => $this->posts(),
        ]);
    }
}
```

Because the `#[Session]` attribute has been added to the `$search` property, after a user enters a search value, they can refresh the page and the search value will be persisted. Every time `$search` is updated, its new value will be stored in the user's session and used across page loads.


luanfreitasdev:
What do you think of `#[Persist]`?

devajmeireles:
> What do you think of `#[Persist]`?

Sounds very appropriate for this case.

calebporzio:
Yeah, I think there's another world where there's `#[Persist]` with different strategies: "url", "session", "cache". 

But because "Computed(persist: true)" is a thing, I don't want to confuse definitions. I think I'll stick with #[Session] for now. Thanks

ju5t:
@calebporzio this is exactly what we need. 

At what point can we expect performance issues? Worst scenario we would store ~30-40 different keys per user with this, in total probably 50 keys. Would this be an issue?

The ~30-40 keys would be short-lived. We can ditch them after a few minutes. Would a cached attribute make sense for this? We could use `session()->getId()` to build the key. If you're up for it, happy to PR!

PR #7749, opened 2024-01-18
Do not prepend slash for script asset url with a protocol

rynomars:
The static method `js` in the `FrontendAssets` class prepends a slash on $url string. This presents a problem when a user wants to serve their js asset on a server different than what the app resides on. e.g someone wanting to host their livewire.js file on an asset cdn.

This PR uses the laravel `Str` class to determine if the $url string is a valid url.

When $url is not a url, we then prepend the slash.

This PR address submitted bug #7747 

***

Thanks for contributing! 🙌


calebporzio:
Thanks!

PR #7750, opened 2024-01-18
Safari DOM walking issues

nuernbergerA:
Given a very big and deep DOM with around ~700k nodes rendered via a controller with some super tiny livewire components close to root.

```html
<html>
<body x-data>
<form>
    <div>
        <livewire:some-component/>
    </div>
    <div>
        <main x-data> <!-- loop -->
            <div>
                <main><!-- loop -->
                    <div x-data>
                        <main>
                            <div>
                                <div><!-- loop -->
                                    <div>
                                        <div><!-- loop -->
                                            <label>
                                                <input type="text" />
                                            </label>
                                        </div>
                                    </div>
                                </div>
                            </div>
                        </main>
                    </div>
                </main>
            </div>
        </main>
    </div>
</form>
</body>
</html>
```

With the current implementation this page renders in under 1 second in Chrome but takes up to 90 seconds on Safari (the new IE...).

After some profiling and testing it was clearly this code fault:
```js
Alpine.interceptInit
//...
let component = closestComponent(el, false);
```

If we stop seeking a livewire component if we dont need one we can speed things up and reach same results on both browsers.


nuernbergerA:
maybe we need to check also for attribute names starting with `x-` to support `_x_teleportBack` or if needed also calls to `$parent` within `x-data` or `x-init`

calebporzio:
Thanks @nuernbergerA !!

PR #7751, opened 2024-01-19
Fix Livewire component key generation across load balanced servers

calebporzio:
Because Livewire child component keys are random strings and are generated at Laravel compile-time, load-balanced environments experience bugs because the Livewire component keys are different across servers.

This PR makes Livewire component "key" generation deterministic across load balanced environments by using the current Blade view and an incrementing counter as the key instead of a random string.

### DIFF
--- a/src/Mechanisms/RenderComponent.php
+++ b/src/Mechanisms/RenderComponent.php
@@ -13,9 +13,7 @@ class RenderComponent extends Mechanism
 
     public static function livewire($expression)
     {
-        $key = app(\Livewire\Mechanisms\ExtendBlade\DeterministicBladeKeys::class)->generate();
-
-        $key = "'{$key}'";
+        $key = null;
 
         $pattern = "/,\s*?key\(([\s\S]*)\)/"; // everything between ",key(" and ")"
 
@@ -23,6 +21,11 @@ class RenderComponent extends Mechanism
             $key = trim($match[1]) ?: $key;
             return "";
         }, $expression);
+
+        if (! $key) {
+            $key = app(\Livewire\Mechanisms\ExtendBlade\DeterministicBladeKeys::class)->generate();
+            $key = "'{$key}'";
+        }
 
         return <<<EOT
 <?php
--- a/src/Mechanisms/RenderComponent.php
+++ b/src/Mechanisms/RenderComponent.php
@@ -3,7 +3,7 @@
 namespace Livewire\Mechanisms;
 
 use Illuminate\Support\Facades\Blade;
-use Illuminate\Support\Str;
+use Livewire\Livewire;
 
 class RenderComponent extends Mechanism
 {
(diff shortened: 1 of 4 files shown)


nuernbergerA:
@calebporzio performance wise the new implementation seems more "heavy" then a simple `Str::random`

wouldn't it be better to replace and fetch a possible key upfront and only generate when non was provided?

src/Mechanisms/RenderComponent.php
```diff
@@ -3,7 +3,7 @@
 namespace Livewire\Mechanisms;
 
 use Illuminate\Support\Facades\Blade;
-use Illuminate\Support\Str;
+use Livewire\Livewire;
```
nuernbergerA, line 6:
```suggestion
```

unused import

src/Mechanisms/RenderComponent.php
```diff
@@ -14,26 +14,17 @@ function boot()
...
 
-        $pattern = "/,\s*?key\(([\s\S]*)\)/"; //everything between ",key(" and ")"
+        $key = "'{$key}'";
+
+        $pattern = "/,\s*?key\(([\s\S]*)\)/"; // everything between ",key(" and ")"
 
         $expression = preg_replace_callback($pattern, function ($match) use (&$key) {
             $key = trim($match[1]) ?: $key;
             return "";
         }, $expression);
```
nuernbergerA, line 26:
```suggestion
        $key = null;
        $pattern = "/,\s*?key\(([\s\S]*)\)/"; // everything between ",key(" and ")"

        $expression = preg_replace_callback($pattern, function ($match) use (&$key) {
            $key = trim($match[1]);
            return "";
        }, $expression);
        
        if (! $key) {
                $key = app(\Livewire\Mechanisms\ExtendBlade\DeterministicBladeKeys::class)->generate();
                $key = "'{$key}'";
        }
```

maybe something like this

nuernbergerA (changes requested):

dwightwatson:
This PR may have been the cause of the issues discussed here: https://github.com/livewire/livewire/discussions/7934

Can see above that `enflow/livewire-twig` had to override this change to get around it.

I've only just run into it after upgrading to Laravel 11 but looks like people are seeing it on Laravel 10 as well. Generally isolated to Linux rather than Mac which.

MickProPay:
I dont think this solved the problem guys. We still having issues and unless our forge blancers is set to IP hash livewire is a problem.

ettiennelouw:
We are also still having the same issue and need our load balancer on IPHash, some of the filament panel components are also having issues.

peterjbassi:
Also experiencing this issue with Livewire 3.6.3 in GCP Cloud Run behind a load balancer, namely with Filament components reaching out to `/livewire/update`. Enabling sticky sessions in the load balancer or reducing to only 1 container instance both work around the problem, but neither are a solution for us given the nature of our application.

joshhanley:
@MickProPay @ettiennelouw and @peterjbassi thanks for reporting! When you're all saying your having issues, can you please be specific about exactly what the issues you are seeing, with error messages and screenshots and a description of what is happening, so we can try and replicate it so we can investigate it? Thanks!

peterjbassi:
@joshhanley 
```
laravel/framework 11.44.7
filament/filament 3.3.13
livewire/livewire 3.6.3
```

I first encountered this issue when I deployed a fully developed Filament application to GCP Cloud Run (serverless) with 3 container instances. A resource index (table) page with a couple of chart widgets would break (displaying a basic Laravel 500 error modal) after the initial payload had rendered. Chrome devtools showed that some of the `/livewire/update` requests for the chart widgets returning a `500`. The number of requests that failed varied with page reload. I had never experienced that problem in the prior two months of development on my local machine with a single Docker container running the app.

As part of my investigation, I created a fresh Laravel project, installed Filament, and put 10 simple bar chart widgets on the default dashboard. The same issue presented itself when I deployed the new project to Cloud Run in a different GCP project. 

Through testing/tweaking I've encountered a variety of symptoms, including:
- `500` responses with no exception/stacktrace (`APP_ENV=development, APP_DEBUG=true`)
- Partial `200` responses missing various random pieces of the chart widget mark-up
- `development.ERROR: Livewire encountered a missing root tag when trying to render a component.`

The following configurations have "fixed" the issue:
- A revision with a single container instance
- A revision with multiple container instances and Cloud Run's Session Affinity feature turned on
- A revision with multiple container instances and a GCS bucket mounted at `./storage/framework/views` so that all container instances were sharing a common view cache

It was that last working configuration that really pushed me to thinking this was a Livewire load-balancer-related Laravel view/blade cache problem.

The network tab ends up looking like this, with the count and position of failed requests different on every page load:
image

That's probably the best I can provide right now. I'm not familiar with Livewire's internals. 

Happy to answer any questions.

peterjbassi:
Created a small app to demonstrate: https://github.com/peterjbassi/filamentapp

joshhanley:
@peterjbassi thanks!

joshhanley:
Going to keep this discussion going in #9323

PR #7754, opened 2024-01-19
feature: add an X-Livewire-Navigate header to navigate requests

ryangjchandler:
This adds an `X-Livewire-Navigate` header to all requests made when fetching HTML for a `wire:navigate` request – it's in a similar vein to the `X-Livewire` header that gets added to internal Livewire calls.


calebporzio:
Dig it, thanks Ryan!

PR #7755, opened 2024-01-19
Remove Component not found message because alpine already removed it from livewire chidrens

luanfreitasdev:
…ewire before

This PR removes the Component not found message for the following reasons:

The Message is only displayed when Livewire tries to restructure the children again in the browser, however, Alpine removed one of these components, causing this message.

```js
Alpine.onAttributeRemoved(el, 'wire:id', () => {
     destroyComponent(component.id)
})
```

---

If the component already exists in the element, it returns itself:

```js
// if (el.__livewire) return el.__livewire;

// if (components[component.id]) return component
```

The PR also treats the component as empty, but it is no longer used. Maybe this approach can be improved and I would like to know if what I did improves or if we can try something else.

Thanks for contributing! 🙌


calebporzio:
Sorry @luanfreitasdev  - we will need a more thorough explanation and tests

luanfreitasdev:
Can you tell me which test case this can prove the concept? I can try to develop one, but I haven't been able to think of a reverse test yet, that is, one that proves that something doesn't happen due to the structure

PR #7757, opened 2024-01-19
[3.x] Add `livewire:upload-cleanup` Command

devajmeireles:

N
Y
N
Not necessary.

Because clearing temporary files is only performed when a new upload is made, **temporary upload files will never be deleted if a new upload is not made.** With this pull request, we introduce the command `livewire:upload-cleanup` which allows us, for example, to create a daily cronjob for cleaning files, consuming the same logic as the `WithFileUploads` trait


calebporzio:
Ah, this makes sense. I'll say for now, people can achieve this in their own apps however they want (probably by creating a simple artisan command with the file upload trait like you have here).

If there's more demand for it, we'll add it to the core.

Thanks!

PR #7758, opened 2024-01-19
Fix modification of original response value on asset injection

rynomars:
When using Livewire v3 unit tests fail when using Laravel's assertViewIs or assertViewHas on a standard Laravel View that implements a Livewire component. Note this only happens when auto injecting the livewire js asset.

These failures occur do to Livewire calling setContent on the Response object when injecting the script tag for livewire.js. Livewire passes a string of the html to setContent after injection. The Laravel assertions mentioned above have a check to make sure the response uses a view. This check is done on the "original" property of the response. The setContent method populates this property and livewire has changed it to be a string. That is when the PHP Unit failure of "The response is not a view" is set.

This PR fixes the submitted bug of #7756 
 
---

Thanks for contributing! 🙌


calebporzio:
Thank you!

PR #7769, opened 2024-01-21
Add assertions to test the contents of file downloads

colinmackinlay:
This PR adds two methods to test thae contents of file downloads: `assertFileDownloadedContains()` and `assertFileDownloadedNotContains()`.

@joshhanley you high level reviewed this in #5192 but closed it down while V3 was being finalised. I've updated to be V3 compatible.

### DIFF
--- a/docs/downloads.md
+++ b/docs/downloads.md
@@ -106,3 +106,35 @@ public function does_not_download_invoice_if_unauthorised()
         ->assertNoFileDownloaded();
 }
 ```
+
+You can test to ensure a downloaded file contains a string using the `->assertFileDownloadedContains()` method:
+
+```php
+use App\Models\Invoice;
+
+/** @test */
+public function download_invoice_contains_text()
+{
+    $invoice = Invoice::factory();
+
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks for contributing!

In order to keep Livewire as focused and maintainable as possible, I'm going to close this pull request.

This isn't "no", it's just "no for now".

If you disagree, or there's further demand from the community, please re-open this PR and we'll reconsider!

nuernbergerA:
@colinmackinlay as an alternative you could PR this to https://github.com/christophrumpel/missing-livewire-assertions

colinmackinlay:
> @colinmackinlay as an alternative you could PR this to https://github.com/christophrumpel/missing-livewire-assertions

Thanks @nuernbergerA , good idea. I think its pretty important that, when a file is created to download, its contents can also be tested. 

I copy these tests into my local source after every livewire upgrade!

The architecture of @christophrumpel package is quite different so it might take me a little while to work that out

nuernbergerA:
@colinmackinlay i already ported your code, i will open a PR soon

PR #7779, opened 2024-01-23
Fix querystring except after refresh

joshhanley:
This PR fixes an issue with the query string, when using except, if a different value is loaded from the query string on refresh, then the except value doesn't work and the query string key isn't removed.

I've also updated the `package.lock` reference to point to Alpine v3.13.4 so if someone runs `npm install` it's up to date.


calebporzio:
Thanks Josh!!

PR #7784, opened 2024-01-23
Add #[Cache] attribute

ju5t:
I feel bad for opening this PR after reading the 'inbox zero' email. Sorry! 

This PR is nearly identical to #7748.

There are many use-cases for cached properties. Ours isn't the most persuasive one I think :)

`#[Session]` is awesome. It's such a smart idea. I really like it. But I'm not sure yet if it works for us in the current app I'm working on, as it comes with some performance bottlenecks. And we have some things planned where it makes sense to cache larger filters of several pages (and keep them active for some time too). So here's the `#[Cache]` attribute.

It comes with a `private`-option which allows you to cache properties just for a user. I think if you set `private` it should also set a default ttl, as you don't want cached items hanging around that will never be used again. This isn't implemented yet.

```php
use Livewire\Attributes\Cache;
use Livewire\Component;
use App\Models\Post;

{
    #[Cache(private: true, ttl: 360)]
    public $search;

    protected function posts()
    {
        return $this->search === ''
            ? Post::all();
            ? Post::where('title', 'like', '%'.$this->search.'%');
    }

    public function render()
    {
        return view('livewire.show-posts', [
            'posts' => $this->posts(),
        ]);
    }
}
```

<s>Sorry, tests are limited. I can't run Dusk tests on my local machine. I have followed the contribution guide but no luck for me.</s>

Turns out it's looking for a session table. I've changed the phpunit.xml.dist on my machine to circumvent that, obviously not committing this here :)

### DIFF
--- a/src/Attributes/Cache.php
+++ b/src/Attributes/Cache.php
@@ -3,10 +3,10 @@
 namespace Livewire\Attributes;
 
 use Attribute;
-use Livewire\Features\SupportSession\BaseSession;
+use Livewire\Features\SupportCache\BaseCache;
 
 #[Attribute(Attribute::TARGET_PROPERTY)]
-class Cache extends BaseSession
+class Cache extends BaseCache
 {
     //
 }
--- a/src/Attributes/Cache.php
+++ b/src/Attributes/Cache.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Attribute;
+use Livewire\Features\SupportSession\BaseSession;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+class Cache extends BaseSession
+{
+    //
+}
--- a/src/Features/SupportCache/BaseCache.php
+++ b/src/Features/SupportCache/BaseCache.php
@@ -5,6 +5,7 @@ namespace Livewire\Features\SupportSession;
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
 use Attribute;
 use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\Session;
 
 #[Attribute(Attribute::TARGET_PROPERTY)]
 class BaseCache extends LivewireAttribute
@@ -12,15 +13,16 @@ class BaseCache extends LivewireAttribute
     function __construct(
         protected $key = null,
         protected $ttl = null,
+        protected $private = false,
     ) {}
 
     public function mount($params)
     {
         if (! $this->exists()) return;
 
-        $fromSession = $this->read();
+        $fromCache = $this->read();
 
-        $this->setValue($fromSession);
+        $this->setValue($fromCache);
     }
 
     public function dehydrate($context)
@@ -45,6 +47,8 @@ class BaseCache extends LivewireAttribute
 
     protected function key()
     {
-        return $this->key ?: (string) 'lw' . crc32($this->component->getName() . $this->getName());
+        $key = $this->key ?: (string) 'lw' . crc32($this->component->getName() . $this->getName());
+
+        return $this->private ? Session::getId() . '-' . $key : $key;
     }
 }
--- a/src/Features/SupportCache/BaseCache.php
+++ b/src/Features/SupportCache/BaseCache.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Livewire\Features\SupportSession;
+namespace Livewire\Features\SupportCache;
 
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
 use Attribute;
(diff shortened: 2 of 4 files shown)


levu42:
@ju5t assuming this PR was merged, I would be quite confused what the difference is between `#[Cached]` and `#[Session]`. If the only thing is, that `#[Session]` has performance bottlenecks, wouldn't it be better to try to improve the code there to make it more performant? How would you describe the difference between the two of them in the docs?

And what would be the use case for non-private cached properties, that are shared between all components of all users? Would this be some data-heavy calculation that's not user-dependent? Could that not be solved by already existing cached computed properties?

ju5t:
> If the only thing is, that `#[Session]` has performance bottlenecks, wouldn't it be better to try to improve the code there to make it more performant? 

I don't believe this is possible. It's how Laravel handles session data, so I'm not sure how Livewire can improve on it. 

> Would this be some data-heavy calculation that's not user-dependent? Could that not be solved by already existing cached computed properties?

Computed exist for the lifetime of a component. Session for as long as the session is valid. Cache for as long as the ttl doesn't expire, which could be indefinitely. There's no denying they have a lot similarities, but I think each has its own use.

levu42:
@ju5t what do you mean with "Computed wouldn't handle dynamic situations however where `#[Cache]` can"? Do you have an example of that?

ju5t:
> what do you mean with "Computed wouldn't handle dynamic situations however where #[Cache] can"? Do you have an example of that?

```php
class Computed extends Component
{
    public $search = '';

    #[\Livewire\Attributes\Computed(persist: true)]
    public function searched()
    {
        return User::query()->where('name', "%{$this->search}%");
    }
}
```

This is a simple example. `searched()` would not update if you change `$search`. It would only change if you `unset()` it.

With `#[Cache]` this would look like this:

```php
#[Cache(private: true, ttl: 600)]
public $search = '';

#[Cache(private: true, ttl: 600)]
public $results = null;

public function searched()
{
    $this->results = User::query()->where('name', "%{$this->search}%");
}
```

In this scenario both `$search` and `$results` would be cached. A user can navigate away from this component and come back later. They can adjust their filters whenever they want and it would all be cached. This is not something I would like to do in a session, as our datasets grow over time.

We have more complicated use-cases which are a bit more difficult to explain here.

calebporzio:
Interesting. Here are a few thoughts:

A) `#[Cache]` is a natural extension of the `#[Cache]` and `#[Url]` family of attributes
B) Persisting properties in a cache often makes more sense than the session. Or is at least better performance wise
C) A part of me wonders if we should unit "session" and "cache" into a single `#[Persist]` attribute that supports different strategies. Idk. Maybe that's something to think about for the next major version. I don't want to change direction so quickly on something so knew.
D) I wonder about what defaults would be?
  D.1) How long for ttl?
  D.2) Session is naturally scoped to individual users, would Cache be global? or would we auto-scope it to specific users? How would we scope it to a user if they're not logged in (session would work fine)
  D.3) Would people want different key strategies like: scope for this instance of the component, scope for this component across instances, scope for all components? idk...
  
Just a pile of thoughts. Interested to hear feedback.

ju5t:
Defaults are tricky--but I went for the same defaults as the facade in this PR for now (indefinite ttl). 

This doesn't feel right for user scoped data when you're using `Session::getId()` as a unique identifier. The cache becomes invalid when the session expires. You could use `auth()->id` and default to the session id for unauthenticated users. In that case, users could use the default session timeout as a ttl instead and when they authenticate it could use the defined ttl.

Some key strategies are already possible for both `#[Session]` and `#[Cache]`:

- Specifying a key would be global scope.
- Not specifying a key would be component scope.

calebporzio:
Yeah, good points, I think the only one I'm hung up on is:

- Is the cached scoped to the user automatically or not?

I think this gets confusing seeing that `#[Session]` IS scoped to user, but `#[Computed(persist: true)]` (which uses caching) is NOT.

ju5t:
`#[Computed]` could support user scoped cache by adding `private`?

The only thing that bothers me a bit is that we have some implicit scopes. 

- Specifying a key would be global scope. Combine this with `private`: a global & user scoped cache.
- Not specifying a key would be component scope. Combine this with `private`: a component & user scoped cache.

I thought about using an enum to explicitly define the scope, but that can get messy.

In the end I think it's the easiest to add `private` to computed. As computed already has implicit scopes baked in. It will either use the specified (global scope) key or it uses a generated key based on the component name (component scope).

calebporzio:
Yeah, I see your points. I think I want to let this feature simmer a little longer so I'm going to close this for now until there's more demand/clarity on the feature.

We can always add it in the future, but it will be harder to remove it or change it if we regret it. So thanks for all your effort, and hopefully this gets used some day!

ju5t:
No worries!

We will implement it in our app and see where it takes us. I think that's the best approach for now. I hope it brings some clarity and prepare a better case/implementation for it, but if not, we will know that too :)

PR #7802, opened 2024-01-25
add @entangle documentation

timkley:
Hi everyone,

maybe I'm mistaken, but apart from the mention in the upgrading guide I couldn't find any documentation about the `@entangle` function in the docs for v3.

I've added it to the Alpine section, it's basically a copy of the [v2 documentation](https://laravel-livewire.com/docs/2.x/alpine-js#sharing-state) but adjusted to the default `defer` behaviour.

### DIFF
--- a/docs/alpine.md
+++ b/docs/alpine.md
@@ -228,52 +228,63 @@ You can easily refresh a Livewire component (trigger network roundtrip to re-ren
 <button type="button" x-on:click="$wire.$refresh()">
 ```
 
-## Sharing state between Livewire and Alpine: `$wire.entangle`
+## Sharing state using `$wire.entangle`
 
-Livewire has an incredibly powerful feature called "entangle" that allows you to "entangle" a Livewire and Alpine property together. With entanglement, when one value changes, the other will also be changed.
+In most cases, `$wire` is all you need for interacting with Livewire state from Alpine. However, Livewire provides an additional `$wire.entangle()` utililty that can be used to keep values from Livewire in-sync with values in Alpine.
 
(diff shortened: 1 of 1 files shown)


levu42:
@timkley There have been a few discussions going on recently about `@entangle` having some limitations / bugs, and `$wire.entangle` being preferable in these situations: https://github.com/livewire/livewire/issues/6833#issuecomment-1902260844 - maybe it's worth adjusting the documentation text with that in mind.

timkley:
@levu42 thank you for the heads up, I've updated it and also added a warning to the bottom to make people aware of this issue.

docs/alpine.md
```diff
@@ -228,6 +228,64 @@ You can easily refresh a Livewire component (trigger network roundtrip to re-ren
...
+<div x-data="{ open: $wire.entangle('showDropdown').live }">
+    ...
+</div>
+```
+
+> [!tip] You might not need `$wire.entangle`
+> In most cases, you can achieve what you want by using `$wire` to directly access Livewire properties from Alpine rather than entangling them. Entangling two properties rather than relying on one can cause predictibility and performance issues when using deeply nested objects that change frequently. For this reason, `$wire.entangle` has been de-emphasized in Livewire's documentation starting with version 3.
+
+> [!warning] Refrain from using the @@entangle directive
+> In Livewire version 2, it was recommended to use Blade's `@@entangle` directive. That is no longer the case in v3. `$wire.entangle()` is preffered as it is a more robust utility and avoids certain [issues when removing DOM elements](https://github.com/livewire/livewire/pull/6833#issuecomment-1902260844).
```
erikscheepers, line 287: preffered should be preferred

PR #7808, opened 2024-01-25
Fix nullable/optional enum implicit binding

rynomars:
When a component has a nullable/optional mount param that is of an enum type we are getting an error when a null value is passed.

For example, if we have a component, Foo that has the mount signature:
`public function mount(?EnumToBeBound $enum = null)`

if we pass null we get the following error:
`"0" is not a valid backing value for enum`

Currently the method `getImplicitBinding` is calling `from` on the enum class. Doing so causes the error.

This PR changes the `from` to a `tryfrom` which returns null. You can see the examples of this in the test `mount_method_receives_route_and_implicit_enum_optional_binding_and_dependency_injection` that has been provided.


calebporzio:
This is great! Thank you!

PR #7809, opened 2024-01-25
fix: use key in combination with persist

ju5t:
This PR makes sure that the `key` is actually used when you use `persist`. It includes a browser test as well but to be fair it's pretty pointless. It wouldn't fail if you didn't pass a key before so it might be better to just remove that test. The unit test covers it.


calebporzio:
Thanks!

PR #7810, opened 2024-01-25
Prevent ClassMorphViolationException when enforceMorphMap is used

Naoray:
As explained in https://github.com/livewire/livewire/discussions/6862, when using `Relation::enforceMorphMap()` in your app, every time a Livewire v3 component dehydrates and the Model is not present in the morph map, it will throw a `ClassMorphViolationException`.

This PR simply wraps the `$model->getMorphClass()` call within a try/catch statement to prevent the exception from being thrown.

### DIFF
--- a/src/Features/SupportLegacyModels/EloquentModelSynth.php
+++ b/src/Features/SupportLegacyModels/EloquentModelSynth.php
@@ -3,6 +3,7 @@
 namespace Livewire\Features\SupportLegacyModels;
 
 use Illuminate\Contracts\Database\Eloquent\CastsAttributes;
+use Illuminate\Database\ClassMorphViolationException;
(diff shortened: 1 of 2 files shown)


calebporzio:
Hey, @Naoray - thanks for the contribution.

Currently this PR is breaking many tests. Can you make sure the entire test suite passes before I consider this merge? Thanks!

Naoray:
@calebporzio sure! Sorry, was in a hurry when I initially created this PR.

The errors seem to arise from the issue that when the morph map is required through `Relation::requireMorphMap()` it is required in every following tests. Not sure if there is a better way to handle this.

calebporzio:
Thanks!

PR #7812, opened 2024-01-26
files download: properly return 304 response on cache match

apreiml:
hotfix for https://github.com/livewire/livewire/discussions/7794


calebporzio:
Thanks for this fix, however I chose to revert the original PR as there is no test to verify that this fixes the problem people have reported.

Thanks again for pitching in! Maybe in the future someone will re-PR that feature with your strategy instead.

All the best!

PR #7813, opened 2024-01-26
destroy array key using unset when using dot com notation with except and as

luanfreitasdev:

N
Y
N
Y

---

When using dot notation in the query string it works as expected. However, introducing '`as`' and '`except`' leads to an error arising: "`Uncaught SyntaxError (in promise): "undefined" is not valid JSON.`" when the key is destroyed using unset. This blocks the page.

### Expected behavior:

- Use dot notation instead of array in conjunction with **as** and **except**. ❌

**A test failing with this behavior has been added:**

```php
protected function queryString() {
   return [
       'tableFilters.filter_1.value' => [
         'as' => 'filter',
         'except' => '',
      ],
  ];
}
```

### Similar behavior that works:

- Use dot notation instead of array in conjunction with and **except**. and without the '**as**'. ✅

```php
protected function queryString() {
   return [
       'tableFilters.filter_1.value' => [
         'except' => '',
      ],
  ];
}
```

- Use array instead of dot notation ✅

```php
protected function queryString() {
   return [
        'tableFilters' => [
              'filter_1' => [
                'value' => [
                    'as' => 'filter',
                    'except' => ''
                ]
        ],
   ];
}
```


calebporzio:
Thanks @luanfreitasdev

PR #7825, opened 2024-01-27
Add a session prefix for uniqueness

ju5t:
First of all: this PR is a proof of concept. I can imagine you prefer a different approach. Or perhaps do not want to address this at all, especially as this increases session counts. 

I found something while testing `#[Session]`. We have the following reusable component:

```php
class CustomerOrders extends Component
{
    #[Locked]
    public int $id;

    #[Session]
    public array $filters = [];
}
```

This component is used on the customer edit page to filter orders.

In this component we set `$id` to the customer id. But as this doesn't affect the session key, your filters will be the same for every customer. As the component is the same--it's just the `$id` that's different. It doesn't matter if you specify a key.

The end-result is:

1. You browse to customer 1 and use the filter. It gets stored in the session.
2. You go to customer 2 and the same session is used, which loads the same filters as customer 1.

So an end-user will have to reset those filters when they go the next page. They will not expect that.

My solution is a contract. It can be used as follows:

```php
class CustomerOrders extends Component implements SessionPrefix
{
    #[Locked]
    public int $id;

    #[Session]
    public array $filters = [];

    public function sessionPrefix()
    {
        return 'customer-orders-'.$this->id;
    }
}
```

This ensures we have a unique prefix per customer.

### DIFF
--- a/docs/session-properties.md
+++ b/docs/session-properties.md
@@ -66,3 +66,25 @@ class ShowPosts extends Component
 ```
 
 When Livewire stores and retrieves the value of the `$search` property, it will use the given key: "search".
+
+Additionally, if you want to generate the key dynamically from other properties in your component, you can do so using the following curly brace notation:
+
+```php
+<?php
+
+use Livewire\Attributes\Session;
+use Livewire\Component;
+use App\Models\Author;
+
+class ShowPosts extends Component
+{
+    public Author $author;
+
+    #[Session(key: 'search-{author.id}')] // [tl! highlight]
+    public $search;
+
+    // ...
+}
+```
+
+In the above example, if the `$author` model's id is "4", the session key will become: `search-4`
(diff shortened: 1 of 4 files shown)


calebporzio:
Ah, I see. I think better solution might be doing what we do in `#[On]` where we support accessing property values in a string with curly braces. Here's what would be ideal to me in your example:

```php
class CustomerOrders extends Component
{
    #[Locked]
    public int $id;

    #[Session(key: 'custom-orders.{id}')]
    public array $filters = [];
}
```

What do you think?

ju5t:
100%. I've updated the code. There is some code duplication--`replaceDynamicPlaceholders` is copied over and only its variables and exception message have been changed. I can imagine this could be extracted into either a helper method or a action-like class. I'm not sure what your preference is on this so I've done it like this for now.

calebporzio:
Perfect! I added some docs and this will be available in the next release!

PR #7828, opened 2024-01-28
Add fix for standard model synth morph map exception

joshhanley:
This PR is an extension of PR #7810 by @Naoray to ensure that the morph map exception is also handled for the standard model synth.


calebporzio:
Thanks Josh!!

francoism90:
https://laravel.com/docs/10.x/helpers#method-rescue

For these cases I like to use `rescue`, maybe I'm overseeing something here. Just found it interesting. :)

Issue #7824, opened 2024-01-27
undefined error in x-for after upgrading to v3


gofish543:
### Discussed in https://github.com/livewire/livewire/discussions/7475

<div type='discussions-op-text'>

<sup>Originally posted by **JonathanLouw** December  6, 2023</sup>
### Livewire version

3.2.6

### Which PHP version are you using?

PHP 8.1

### Repository URL

https://github.com/JonathanLouw/livewire-playground

### Steps To Reproduce

clone the provided repo and navigate to /example route or set up your own deferred loading example page:

```
class ExampleComponent extends Component
{
    public bool $load = false;

    public function loadUsers(): void
    {
        $this->load = true;
    }

    public function render()
    {
        return view('livewire.example-component', [
            'users' => $this->load
                ? User::query()->inRandomOrder()->limit(5)->get()
                : collect(),
        ]);
    }
}
```
```
<div>
    <h1>Example Component</h1>
    <button wire:click.prevent="loadUsers">Load</button>
    <div x-data="{ users: @js($users) }">
        <template x-for="user in users">
            <div x-text="user.name"></div>
        </template>
    </div>
</div>
```
```
Route::get('/example', \App\Livewire\ExampleComponent::class);
```
Then try to click load to fetch the results

### Additional information

With the provided example the behaviour I'm getting is this:

click "load" button once, nothing happens, click it a second time and I get a warning that user is undefined, for example:

```
Alpine Expression Error: user is not defined

Expression: "user.name"
```
it also output a console error pointing to this bit of code:
```
(async function anonymous(__self,scope
) {
with (scope) { __self.result = user.name }; __self.finished = true; return __self.result;
})
```
And strangely enough when I click it a third time it starts working perfectly fine and correctly fetches 5 random users on every click.

In various other places we are using alpine in our project we're getting the same alpine errors like the one above where a lot of errors are being output that things are undefined which all worked fine on v2 but not after upgrading to v3.

### Please confirm (incomplete submissions will not be addressed)

- [X] I have provided easy and step-by-step instructions to reproduce the bug.
- [X] I have provided code samples as text and NOT images.
- [X] I understand my bug report will be removed if I haven't met the criteria above.</div>

joshhanley:
This issue has been closed as we are now using discussions to track bugs and ideas, and as there is already a discussion open we will continue investigating in that discussion.
See this announcement for more details https://github.com/livewire/livewire/discussions/3147

PR #7830, opened 2024-01-28
Attempting to add failing test for xfor with deferred data loading

gofish543:
- https://github.com/livewire/livewire/discussions/7475
- https://github.com/livewire/livewire/issues/7824

I don't even know if this is a truly failing test case as I can't even get the Livewire to successfully run any of its test cases due to database issues, session issues, and CSRF issues.

Either way, the test case documents my issue and I cannot use Livewire or publish my two months of work to production until this gets resolved.

### DIFF
--- a/src/Features/SupportModels/BrowserTest.php
+++ b/src/Features/SupportModels/BrowserTest.php
@@ -125,6 +125,35 @@ class BrowserTest extends \Tests\BrowserTestCase
             ->assertSeeIn('@typedPostsInitializedIsEloquentCollection', 'true')
             ;
     }
+
+    /** @test */
+    public function x_for_breaking_on_deferred_load() {
+        Livewire::visit([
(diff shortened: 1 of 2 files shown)


gofish543:
I don't even know if I put the test case or the views in the correct spot

calebporzio:
Thanks for submitting this. I pulled it down and took a look. It actually isn't really supposed to work - using `@js` inside x-data and expecting Alpine to adapt to data changes was more of a quirk that it ever worked.

I suggest using `#[Lazy]` on your component to lazy load the component with all the posts right away - instead of differing the load manually. Thanks!

gofish543:
@calebporzio So this is actually not my use-case but an example of the problem. In my specific use-case I have a model window that is being opened and when it is opened, the data is loaded into x-data. Data isn't being merged and the component is rendered.

PR #7837, opened 2024-01-29
session flash data is missing when using redirect with nested component and Modelable

jhonoryza:
This PR is an attempt to fix: https://github.com/livewire/livewire/discussions/7797

this will skip session forget if child component has attribute modelable and $to is null

### DIFF
--- a/docs/session-properties.md
+++ b/docs/session-properties.md
@@ -67,7 +67,7 @@ class ShowPosts extends Component
 
 When Livewire stores and retrieves the value of the `$search` property, it will use the given key: "search".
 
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks for the contribution! Would you mind adding a test for this?

jhonoryza:
@calebporzio i have added a test for this, thank you

calebporzio:
Ok so after digging into this, it has nothing to do specifically with Modelable, but it's a bug that happens any time there is more than one component in a request where one of those components does a redirect.

I tweaked the implementation to only clear the flash data if NO components have redirected in a single request.

Thank you.

PR #7838, opened 2024-01-29
Add assertPropertyHasAttribute test helper

ju5t:
I would like to test if an attribute is assigned to a property. This may fit better in Christoph's [missing assertions package](https://github.com/christophrumpel/missing-livewire-assertions) but I leave that up you guys.

```php
class Filter extends Component
{
    #[Session(key: 'foo')]
    public array $filters = [];
}

Livewire::test(Filter::class)
    ->assertPropertyHasAttribute('filters', Session::class, key: 'foo');
```

I thought about supporting methods as well, but I think some of them are already covered by other tests (e.g. events). And it's also why I explicitly called it `assertPropertyHasAttribute`--to leave room for testing method attributes, and to have a similar naming convention: [`assertClassHasAttribute`](https://docs.phpunit.de/en/9.6/assertions.html#assertclasshasattribute).

### DIFF
--- a/src/Features/SupportTesting/MakesAssertions.php
+++ b/src/Features/SupportTesting/MakesAssertions.php
@@ -2,9 +2,10 @@
 
 namespace Livewire\Features\SupportTesting;
 
-use Illuminate\Testing\Constraints\SeeInOrder;
-use PHPUnit\Framework\Assert as PHPUnit;
+use ReflectionClass;
 use Illuminate\Support\Arr;
+use PHPUnit\Framework\Assert as PHPUnit;
+use Illuminate\Testing\Constraints\SeeInOrder;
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Yeah, this might belong better over there. I think I'm gonna hold off on adding this as I don't think I would really find this useful myself as it doesn't really test behavior, but more the existence of an attribute (which admittadly is userful) - so yeah, let's leave this to another package or userland code. Thanks though!

robsontenorio:
@ju5t It would be cool to add it on docs.

ju5t:
@robsontenorio this has been reverted by Caleb. I haven't found the time to PR it to Christoph's package. That's why there is no documentation for it yet :)

PR #7841, opened 2024-01-29
#[Computed] Computed property dependency injection #6967

luanfreitasdev:

Y - https://github.com/livewire/livewire/discussions/6967
Y
N
Y

Reverted in https://github.com/livewire/livewire/pull/7834

A failing test was added to check the possibility of the property computed using an attribute working the same as the dynamic computed property (v2)

2 tests were put in place so that when this is fixed, other things without tests won't break.
- Access computed property using `protected/private`

call_computed_property_attribute_non_public_method - ✅ pass `[#Compute]`
call_computed_dynamic_property_non_public_method - ❌ failed `getProperty()`

### DIFF
--- a/src/Features/SupportComputed/UnitTest.php
+++ b/src/Features/SupportComputed/UnitTest.php
@@ -387,6 +387,13 @@ class UnitTest extends TestCase
         Livewire::test(NullIssetComputedPropertyStub::class)
             ->assertSee('false');
     }
+
+    /** @test */
+    public function injected_computed_property_attribute_is_accessible_within_blade_view()
+    {
+        Livewire::test(InjectedComputedPropertyWithAttributeStub::class)
+            ->assertSee('bar');
+    }
 }
 
 class ComputedPropertyStub extends Component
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the failing test. I started working on this and it's not going to be as trivial as: `wrap(invade($thing))->method()`

There will need to be tweaks made to either of those to get this to work, or something else that's clever and I can't see without spending more time on it.

I'm going to close this. You're welcome to attempt an implementation and re-PR. Thanks!

luanfreitasdev:
I think it is infeasible to change many things in the structure just to allow access to protected assets. Wouldn't it be better to leave the default, just public?

luanfreitasdev:
Some approaches didn't work. Changing the behavior of invade or wrap could leave some things behind as it is used in many places, so I tried to create a new method, however, __call is called inside the Component which causes it to go into a loop.

nickfls:
@luanfreitasdev I know the issue is closed; yet is it still advisable to use `get{Property}Property(DependencyInjection $service)` method syntax? I do not see it covered in v3 documentation, and there isn't a deprecation notice. Is it future-forward safe?

joshhanley:
@nickfls the old syntax is not deprecated and we don't have plans to remove it. But it's not documented as we believe people should use the attribute syntax going forward. But any existing apps can still use the old method without any issues.

PR #7843, opened 2024-01-30
Undefined errors (and duplicate html elements) in x-for when data is provided by Livewire event

gdebrauwer:
Failing test for https://github.com/livewire/livewire/discussions/7815 (and possibly also https://github.com/livewire/livewire/discussions/7475)


luanfreitasdev:
Hello! Two things seem to fix the problem. Add `#[Renderless]` to the fetchItems method and add `wire:ignore` to the parent div.

I believe that making a new request is not assembling the template tag correctly.

calebporzio:
Ok, I got to the bottom of it. This commit caused the breakage: https://github.com/livewire/livewire/pull/7528

That commit, made "dispatch" happen later in the lifecycle.

Because dispatch is later now, it was being processed AFTER morphing the HTML, causing the error referenced here.

The fix in this case is actually to delay the morphing even further to ensure it happens as the LAST stage in handling a Livewire request. (as opposed to undoing the breaking commit)

The symptom of this problem was the "item is undefined" Alpine error on the x-for.

Although this was the breakage caused by this issue, it's a problem that can occur in many different scenarios, as we're seeing with other PRs like this: #7845 

The actual x-for error is: Anytime there is phantom x-for elements, when Alpine tries to re-initailize, it doesn't know those elements were created by a previous x-for, tries to evaluate them as normal elements and throws an error because it doesn't have scope.

Normally `x-for` tears down it's own created elements when the `<template>` tag is removed, however there are cases where this doesn't happen and this error happens.

Or in the case of this PR, the "clone" tree in the morph was being initialized with the new x-for elements BEFORE the live DOM tree, so those elements would get added into the DOM by morph, then live DOM would initialize a second set of x-for elements.

I know that's hard to follow, more putting those notes here for my future reference.

PR #7845, opened 2024-01-30
Undefined errors (and duplicate html elements) in x-for when using wire:navigate

gdebrauwer:
Possibly related to / same issue as https://github.com/livewire/livewire/pull/7843


calebporzio:
Ok so what's happening here is:

* Alpine "hydrates" the template x-for on page load
* wire:navigate gets clicked and the current HTML is cached
* The back button is clicked, loading the x-for template tag ALONG WITH the actual <div>s it produced
* Alpine doesn't know those divs are part of the x-for and just creates new divs, and hydrates the old ones
* When it hydrates the old ones it throws an error because it doesn't have scope

I'll have to think about this a little more, but here are few options:

* Remove the x-for elements before caching a page's HTML
* Somehow make x-for smarter when hydrating to recognize "stale" nodes

joshhanley:
@calebporzio I think I would lean towards removing `x-for` elements before caching. Not sure how hard that will be or whether you will need to make the elements as being part of an `x-for`. Disadvantage is you have to parse the DOM before storage. I think this would be cleaner when restoring though, as the html gets pulled back in and then alpine processes everything as expected.

The second option we could annotate elements when they are generated using an `x-for` with an attribute or something so when Alpine processes the `x-for` again, it could try and match restored elements with keys and either remove or reuse them. Feels like has potential for bugs, but would save re-walking the DOM.

calebporzio:
Yeah, I think you're right @joshhanley

Good news is we son't have to walk the DOM to remove the x-for's, we just need to have x-for itself hook into the "before page cache". It kind of violates the livewire/alpine barrier, but navigate was supposed to be an Alpine plugin anyways so... idk

calebporzio:
Ok, here's the solution I went with:

* Destroy the entire Alpine tree before caching the HTML. This will cleanup `x-for` and `x-if`. It will add performance overhead, but we'll cross that bridge when we come to it...
* Make sure to skip persisted and teleported elements so we don't step on rakes there

I think this is good enough for now (and maybe forever lol)

gdebrauwer:
@calebporzio the failing tests are caused by sushi's latest version 2.5.0

calebporzio:
Shoot! Thank you for the heads up!

calebporzio:
This PR contains another Alpine fix to get all tests passing: https://github.com/alpinejs/alpine/pull/4015

PR #7846, opened 2024-01-30
Add event name to exception

nullsub:
I am getting a EventHandlerDoesNotExist exception and I was wondering what event is causing it


levu42:
@nullsub I feel like this could be a custom exception class:

```php
class EventHandlerDoesNotExist extends \Exception
{
  public function __construct(public readonly string $eventName): void
  {
    parent::__construct("Handler for event $eventName does not exist");
  }
}
```

This would a) have the event name accessible for code b) make it possible to do a custom `catch` block for that exception.

calebporzio:
Thanks!. @levu42 - I'd be down for custom exception if you wanna PR that.

PR #7850, opened 2024-01-30
Alpine property undefined when reordering array returned to view

mattyforth:
This adds a failing test for #7785.

This bug seems to be related to passing data using `return view()` and reordering it, and as such I've had to include another view for the test as opposed to an inline component.

I wasn't sure where to place the test, so please let me know if anything needs tweaking and I'd be happy to sort it!

Thanks


calebporzio:
This has been fixed by one of the recent changes to Alpine's morphing core.

PR #7858, opened 2024-01-31

apreiml:
This commit introduces support for non-local storages again. This time
the cache hit path is fixed and tested which caused #7794 before.

Also the cachedFileResponse function has been changed to de-duplicate
the matchesCache check and to not return charset applied to the
Content-Type.

To support sending the header in the test, the If-Modified-Since
header is now fetched from the Laravel Request object, rather than from
$_SERVER.

Sorry for the mess that the other PR has caused.

### DIFF
--- a/src/Drawer/Utils.php
+++ b/src/Drawer/Utils.php
@@ -2,8 +2,10 @@
 
 namespace Livewire\Drawer;
 
+use Illuminate\Http\Request;
 use Livewire\Exceptions\RootTagMissingFromViewException;
 
+use Livewire\Features\SupportFileUploads\FileUploadConfiguration;
 use function Livewire\invade;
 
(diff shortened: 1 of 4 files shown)


calebporzio:
Hi there, thanks for re-submitting. Given the issues caused in the last one and my own general lack of knowledge about what this PR is doing. I'd ask that you:

- Provide a detailed explanation of what this PR enables (with code examples and such)
- Provide a detailed explanation of the code being added, what went wrong last time, and why it won't go wrong this time.

Thank you!

I'm going to close this for now to keep things clean. Just submit a PR for this branch when you've done the above and I can re-review. Thank you (sorry for being a pain in the butt, I just want to be 100% sure this won't break things again.)

PR #7863, opened 2024-01-31
Support integer backed enums in query string

satoved:
Related to https://github.com/livewire/livewire/pull/7572

Query string is only supported for string backed enums.

There are also integer backed enums, that can also be useful.

With current implementation an integer cannot be set from url to populate an Enum property, because it's skipped here https://github.com/livewire/livewire/blob/main/src/Features/SupportAttributes/Attribute.php#L77

And it gives an error of trying to assign integer to an Enum-hinted attribute.

**Solution**
Both string and integer can be used to be converted into an enum.

Tests are added.


calebporzio:
Thanks!!

PR #7867, opened 2024-02-01
Expose disk and real storage path of `TemporaryUploadedFile`

gdebrauwer:
In all our projects at work, we use `spatie/laravel-medialibrary` to save the uploaded files. To save a file that already exists on a storage disk, I need the path of the file on the storage disk (and the name of the storage disk).

Currently, there is no way to get that path. For example, the `getRealPath()` method exists, but that is not the real storage path as the `$this->storage->path()` is used. When you try to move the storage file using the path you get from the `getRealPath()` method, you will see that it does not work. The returned path is prefixed based on the config of the disk. 

This PR fixes that by introducing a `getStoragePath()` method and a `getDisk()` method.

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -82,7 +82,6 @@ class TemporaryUploadedFile extends UploadedFile
     public function getPathname(): string
     {
         return $this->getRealPath();
-
     }
 
     public function getClientOriginalName(): string
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -30,6 +30,11 @@ class TemporaryUploadedFile extends UploadedFile
         return $this->storage->path(FileUploadConfiguration::directory());
     }
 
+    public function getDisk(): string
+    {
+        return $this->disk;
+    }
+
     public function isValid(): bool
     {
         return true;
@@ -67,6 +72,11 @@ class TemporaryUploadedFile extends UploadedFile
     public function getRealPath(): string
     {
         return $this->storage->path($this->path);
+    }
+
+    public function getStoragePath(): string
+    {
+        return $this->path;
     }
 
     public function getClientOriginalName(): string
--- a/src/Features/SupportFileUploads/UnitTest.php
+++ b/src/Features/SupportFileUploads/UnitTest.php
@@ -175,6 +175,33 @@ class UnitTest extends \Tests\TestCase
     }
 
     /** @test */
+    public function can_get_disk()
+    {
+        $file = UploadedFile::fake()->image('avatar.jpg');
+
+        $photo = Livewire::test(FileUploadComponent::class)
+            ->set('photo', $file)
+            ->viewData('photo');
+
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $photo);
+        $this->assertEquals('tmp-for-tests', $photo->getDisk());
+    }
+
+    /** @test */
+    public function can_get_storage_path()
+    {
+        $file = UploadedFile::fake()->image('avatar.jpg');
+
+        $photo = Livewire::test(FileUploadComponent::class)
+            ->set('photo', $file)
+            ->viewData('photo');
+
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $photo);
+        $this->assertStringStartsNotWith('livewire-tmp', $photo->getRealPath());
+        $this->assertStringStartsWith('livewire-tmp', $photo->getStoragePath());
+    }
+
+    /** @test */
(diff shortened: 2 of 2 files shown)


calebporzio:
Thanks for this. However, I think this might be confusing.

Like I think my intuition would be that `getPath()` returns `$this->path` and `getStoragePath()` returns `$this->storage->path()`. This is the opposite.

I understand we can't change `getPath()` and this is a way around that. I just want to think more deeply before introducing ANOTHER method to get a path. Currently there are four with this:

CleanShot 2024-02-02 at 12 49 19@2x

Also as there's no precedence for this name, I hesitate.

Can you give me a detailed explanation of each type of "get path" method returns (I'm rusty on file uploads right now) so that we can talk about a suitable name? or a way around this?

Maybe we scrap this all together and you just use `invade($file)->path` to just grab the property invasively

lrljoe:
From memory!

With spatie media library, I believe that you are able to pass it the temporary uploaded file instance, and tell it to store it in it's proper location when it is adding it to the media collection/associating it with a model.

Unless they've changed it that is!  Would need to look at it properly.

I dunno, I feel like exposing the actual path of the temporary uploaded files as a public method could be a very bad thing from a security perspective, a couple of slip ups or misconfigurations on the server, and someone's managed to execute some malicious code on your box, as they have the proper path to the file.

The idea behind a temporary uploaded file is you decide if you want to keep it or not, and use the proper storage methods at that point to save it.

gdebrauwer:
@calebporzio When using the `public` disk as an example, the methods return the following results:
- `getPath()` returns `/var/www/html/storage/app/public/livewire-tmp`
- `getPathname()` returns `/tmp/phpzNluNv`
- `getRealPath()` returns `/var/www/html/storage/app/public/livewire-tmp/KjdOStivWWVNyM2v3UPA8BaqNAp29d-metabmV4dGFwcHMuanBlZw==-.jpg`

To save the file (using `spatie/laravel-medialibrary` in my case), The code checks if the file exists based on the provided path and then moves the file from its temporary location to its permanent location. Only the `getRealPath()` method returns a path with a filename, so let's use that path. The problem is when you run `Storage::exists($file->getRealPath())`, the result will be `false`: the file does not exist on that path. That is not correct, but this is caused by the fact that the path contains a prefix that is not relevant when checking the existence and moving the file on the disk (in this example `/var/www/html/storage/app/public/`).

By introducing the `getStoragePath()` method, you can get the path without that prefix:
- `getStoragePath()` returns `livewire-tmp/KjdOStivWWVNyM2v3UPA8BaqNAp29d-metabmV4dGFwcHMuanBlZw==-.jpg`

calebporzio:
Mmm ok, given @lrljoe's hesitation, the points of potential confusion I raised, and the lack of demand from others I'm going to close this.

For now, you can basically extract what you want manually by removing the everything before "livewire-tmp". Or by diffing the full path with the storage path.

Anyhow, thanks for the contribution, but closing this for now.

gdebrauwer:
```php
class SomeLivewireComponent extends Component
{
    public Event $event;

    public $image;

    public function save()
    {
        FileUploadConfiguration::storage()->setVisibility($this->image->storagePath(), 'public');

        $this->event->addMediaFromDisk($this->image->storagePath(), $this->image->storageDisk())
            ->setFileName(Str::ascii($this->image->getClientOriginalName()))
            ->toMediaCollection('image');

        $this->reset('image');
    }
}
```
And in every project I define macros:
```php
TemporaryUploadedFile::macro('storagePath', fn () => $this->path);
TemporaryUploadedFile::macro('storageDisk', fn () => $this->disk);
```

joshhanley:
@gdebrauwer so you can pass a Livewire TemporaryUploadedFile directly to media library. This is how I'm using it:

```php
class EditProductComponent extends Component
{
    public Product $product;

    public $file;

    public function save()
    {
        $this->product->addMedia($file)->toMediaCollection('product_files')

        $this->file = null;
    }
}
```

joshhanley:
@gdebrauwer can you show me how I can determine whether it has copied the file or moved it? As I want to fix that in my app too.

lrljoe:
@gdebrauwer - Please can you point me in the right direction for where media-library is downloading/streaming temporary uploaded files, as from memory at least it does try to figure out if it's a local file or remote before it tries doing any of that.
Historically at least, much of the stream/download capability was for when it's generating a temporary file for the purposes of he end-user validating that it's the correct file, not for moving it into final location.

They may well have made sweeping changes since I last looked at it granted, so interested to see where they've started doing that.

And if you're using s3 as a temporary storage, then it'll treat it as remote, which is to be expected.

gdebrauwer:
It copies the file when the file is a RemoteFile object and certain conditions are met: https://github.com/spatie/laravel-medialibrary/blob/a1ab07844e7b1be720fdb8ab9677c04a1f40eb27/src/MediaCollections/Filesystem.php#L62-L115

PR #7868, opened 2024-02-01
Handling null in querystring

gdebrauwer:
When a property `$foo`, that uses `#[Url]` attribute is set to `null`, then the querystring will be `foo=null`. On page refresh, the property `$foo` will be `'null'`. That is not correct.

In Laravel, `null` in the querystring is represented as `foo=`. Livewire should work the same, but currently, Livewire will interpret `foo=` as `''`. That means we can not easily change that behavior, as that might break existing Livewire applications.

To fix that, I'm introducing an extra `$nullable` parameter to the `#[Url]` attribute.

- If the `$nullable` parameter set to `false`, then `foo=''` will still result in `$foo = ''`
- If the `$nullable` parameter set to `true`, then `foo=''` will still result in `$foo = null`

If the`$nullable` parameter is not set (= existing Livewire applications), then I ensured that the behavior does not change or break existing Livewire applications:

- If the property is not typehinted, then `foo=''` will still result in `$foo = ''`
- If the property is typehinted and not nullable, then `foo=''` will still result in `$foo = ''`
- If the property is typehinted and nullable, then `foo=''` will result in `$foo = null`.
	- In current Livewire applications, this does not happen, but I think it is safe to assume that if you make a property nullable, you expect the value to be nullable and not an empty string.
	- This also ensures you don't have to explicitly set the `$nullable` parameter if you typehinted the property. You will only need to set the parameter if you did not typehint your property and you want null instead of empty string.
 

```php
// When `foo=` is present in querystring, then: 

#[Url]
public $foo; // will be ''


#[Url(nullable: false)]
public $foo; // will be ''


#[Url(nullable: true)]
public $foo; // will be null


#[Url]
public string $foo; // will be ''


#[Url]
public ?string $foo; // will be null
```


joshhanley:
@gdebrauwer thanks for the PR! Just out of curiosity, how does Laravel handle `null` in the querystring? Does it actually convert it to null or does it leave it as a string null?

gdebrauwer:
When `foo=null` is present in querystring of a Laravel route, the value stays a string `'null'`

calebporzio:
Good point @joshhanley - If Laravel does that, we may not want to diverge from that.

I think this PR is getting at a bigger problem: How does Livewire represent empty strings and nulls in the query string.

Please submit a proposal of how Livewire currently handles these, the problems with that approach, and how Livewire should change to rectify the issue.

Thanks

gdebrauwer:
@calebporzio I updated the PR with a proposal (see PR description) on how to fix the issue and make the behavior consistent with Laravel without breaking existing Livewire applications.

docs/url.md
```diff
@@ -94,6 +94,29 @@ class ShowUsers extends Component
 }
 ```
 
+### Nullable properties
```
calebporzio, line 97: I added some documentation (note for future PRs that docs are helpful)

docs/url.md
```diff
@@ -94,6 +94,29 @@ class ShowUsers extends Component
...
+In these cases, you can use a nullable typehint like so:
+
+```php
+use Livewire\Attributes\Url;
+use Livewire\Component;
+
+class ShowUsers extends Component
+{
+    #[Url]
+    public ?string $search; // [tl! highlight]
```
calebporzio, line 110: I'm only documented the nullable typehint for now as I think `nullable` is unergonomic and most times, the typehint is simple to add.

src/Features/SupportQueryString/BaseUrl.php
```diff
@@ -13,10 +14,13 @@ public function __construct(
         public $history = false,
         public $keep = false,
         public $except = null,
+        public $nullable = null,
     ) {}
 
     public function mount()
     {
+        $this->nullable = $this->determineNullability();
```
calebporzio, line 22: I extracted that big blog of reflection conditional into a small method (makes readability much better)

src/Features/SupportQueryString/BaseUrl.php
```diff
@@ -27,6 +31,23 @@ public function dehydrate($context)
         $this->pushQueryStringEffect($context);
     }
 
+    protected function determineNullability()
+    {
+        // It's nullable if they passed it in like: #[Url(nullable: true)]
```
calebporzio, line 36: I added some comments for future readers

src/Features/SupportQueryString/BaseUrl.php
```diff
@@ -94,7 +121,13 @@ public function urlName()
     public function getFromUrlQueryString($name, $default = null)
     {
         if (! app('livewire')->isLivewireRequest()) {
-            return request()->query($this->urlName(), $default);
+            $value = request()->query($this->urlName(), $default);
+
+            if (array_key_exists($name, request()->query()) && $value === $default) {
```
calebporzio, line 126: I don't think this logic should live here for two reasons:

1) It doesn't account for "lazy" components that don't get handled in this condition, but get handled in `getFromRefererUrlQueryString` further down.

2) Structurally, I wonder if this logic can be more closely located to the line 73 additions.

In general, changes like this that affect multiple parts of a file like this, should ideally be refactored to live together, or be well commented so people know why those conditions are added.

So at least, please add comments to describe why that logic is there.

src/Features/SupportQueryString/BrowserTest.php
```diff
@@ -555,6 +555,169 @@ public function render() { return <<<'HTML'
             ->assertQueryStringMissing('filter')
         ;
     }
+
+    /** @test */
+    public function can_handle_empty_querystring_value_as_empty_string()
```
calebporzio, line 560: I think any of these tests that only care about how values are treated on initial load can be put inside the UnitTest file so that they run much faster and can be used better to refactor the implementation.

calebporzio (changes requested):
Hey @gdebrauwer, thanks for your continued work on this.

I tweaked some things and added a review for you to make a few tweaks of your own.

Thanks for all your help on the project lately!

src/Features/SupportQueryString/BaseUrl.php
gdebrauwer: > Structurally, I wonder if this logic can be more closely located to the line 73 additions.

That is not possible because then the code would be stopped at line 59 (`if ($initialValue === 'noexist') return;`) 

I added a comment to describe why that code exists there

src/Features/SupportQueryString/BrowserTest.php
gdebrauwer: I added a couple of unit tests

gdebrauwer:
@calebporzio Glad I can help 😄

src/Features/SupportQueryString/BaseUrl.php
calebporzio: Perfect explanations. Thank you!

calebporzio (approved):

calebporzio:
Thanks for the updates @gdebrauwer - looks great.

I removed the unit tests because you could actually swap `assertSet('foo', '')` for `assertSet('foo', null)` in all the tests and they would still pass. There must be some loose equality checking under the hood there, so I axed them.

Merging this. Thank you.

joshhanley:
FYI for anyone who comes across this, you can ensure strict types by passing a third parameter `true` into the `assertSet` method, like `assertSet('foo', null, true)`

joshhanley:
Have submitted a PR for an `assertSetStrict()` assertion PR #7909.

PR #7869, opened 2024-02-01
[3.x] Introduce a working `TemporaryUploadedFile::fake()` for users

nuernbergerA:
Currently its "impossible" to test `TemporaryUploadedFile` without uploading a fiel through your component which makes testing a nightmare.

```php
// component
class UploadPhoto extends Component
{
    use WithFileUploads;
 
    public $photo;
 
    public function upload($name)
    {
        $this->photo->storeAs('/', $name, disk: 'avatars');
    }
}

// testing the component
class UploadPhotoTest extends TestCase
{
    /** @test */
    public function can_upload_photo()
    {
        Storage::fake('avatars');
 
        $file = UploadedFile::fake()->image('avatar.png');
 
        Livewire::test(UploadPhoto::class)
            ->set('photo', $file)
            ->call('upload', 'uploaded-avatar.png');
 
        Storage::disk('avatars')->assertExists('uploaded-avatar.png');
    }
}

//easy all works fine
```

```php
// component
class UploadPhoto extends Component
{
    use WithFileUploads;
 
    public $photo;
 
    public function upload($name)
    {
        resolve(SomeAction::class)->handle($this->photo, $name);
    }
}

// testing the component will still
class UploadPhotoTest extends TestCase
{
    /** @test */
    public function can_upload_photo()
    {
        Storage::fake('avatars');
 
        $file = UploadedFile::fake()->image('avatar.png');
 
        Livewire::test(UploadPhoto::class)
            ->set('photo', $file)
            ->call('upload', 'uploaded-avatar.png');
 
        Storage::disk('avatars')->assertExists('uploaded-avatar.png');
    }
}

// but now we are cooking
class SomeActionTest extends TestCase
{
    /** @test */
    public function can_handle_a_upload_photo()
    {
        Storage::fake('avatars');
 
        $file = UploadedFile::fake()->image('avatar.png');

        // will fail because $file is type-hinted as TemporaryUploadedFile
        // reason for type-hint is methods work different compared to UploadedFile
        resolve(SomeAction::class)->handle($file, 'uploaded-avatar.png');

         Storage::disk('avatars')->assertExists('uploaded-avatar.png');
    }
}
```

ideal situation for users would be to just call `TemporaryUploadedFile::fake()` and getting a functional `TemporaryUploadedFile` back as it would be "uploaded", without faking a disk and copy files around.

If you wish i could cook something.

cheers adrian

### DIFF
--- a/src/Features/SupportFileUploads/UnitTest.php
+++ b/src/Features/SupportFileUploads/UnitTest.php
@@ -724,6 +724,43 @@ class UnitTest extends \Tests\TestCase
             ->set('file', $file)
             ->assertSet('content', $file->getContent());
     }
+
+    /** @test */
+    public function user_can_create_a_functional_temporary_uploaded_file_for_testing_purpose()
+    {
+
+
+        // fake results come from \Illuminate\Http\UploadedFile
+        $fileWithContent = TemporaryUploadedFile::fake()->createWithContent('example.txt', 'some content');
+        $fileWithoutContent = TemporaryUploadedFile::fake()->create('example.mp3', 1024, 'audio/mpeg');
+        $fileWithRealImage = TemporaryUploadedFile::fake()->image('example.jpg', 20, 20);
+
+        $this->assertEquals('some content', $fileWithContent->getContent());
+        $this->assertEquals('', $fileWithoutContent->getContent());
+        $this->assertTrue(!! imagecreatefromstring($fileWithRealImage->getContent()));
+
+        // a user should not know how to fake the internal used tmp disk
+        Storage::fake(FileUploadConfiguration::disk());
+        Storage::disk(FileUploadConfiguration::disk())->put("livewire-tmp/{$fileWithContent->getFilename()}", $fileWithContent->getContent());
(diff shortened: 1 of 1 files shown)


calebporzio:
Mmmmm I see what you're saying. I think for now, we should keep it as is because:
- I haven't heard more demand for this
- You can get around this by skipping or loosening the type-hint
- I think `TemporaryUploadedFile::fake()` isn't very ergonomic as a public API

Let's let the on simmer - thanks @nuernbergerA

edalzell:
I need this as well, to test my file upload component. If this won't be implement, can you please document how to test those?

nuernbergerA:
@edalzell i use this in my projects:
```php
function livewireTmpFile(string $filename, string $content = ''): TemporaryUploadedFile
{
     Storage::fake(FileUploadConfiguration::disk());
     Storage::disk(FileUploadConfiguration::disk())->put("livewire-tmp/{$filename}", $content);

     return TemporaryUploadedFile::createFromLivewire($filename);
}
```

PR #7875, opened 2024-02-02
Fix cleanup handler for Echo presence channels

alexwass-lr:
This PR fixes the incorrect cleanup handler for Echo's presence channels which was introduced in #7833 

Full discussion here: https://github.com/livewire/livewire/discussions/7397#discussioncomment-8324865


alexwass-lr:
I just realised I haven't submitted the built files @joshhanley, but my paths seem to be slightly different so I'm not sure if I have that setup correctly?

joshhanley:
@alexwass-lr thanks! Yeah commit your assets, the path will probably be different as Caleb likely builds them using a linked version of Alpine.

alexwass-lr:
Sorry had to fix some merge conflicts with the build, hopefully that's as expected now @joshhanley

calebporzio:
Awesome, thanks!

PR #7876, opened 2024-02-02
Docs: Add example for multiple actions in wire:target

Arne1303:
Right now, there is no documentation about how to pass multiple actions into wire:target on the [wire:loading](https://livewire.laravel.com/docs/wire-loading) page in the wiki. I've added a short section and an example for it, and ordered it below the existing section [Targeting specific actions](https://livewire.laravel.com/docs/wire-loading#targeting-specific-actions).


calebporzio:
Thanks!

PR #7877, opened 2024-02-02
Fix use statements in Url docs

NiclasvanEyk:
I assume this was a copy-paste issue due to copying a `WithSomething` Trait name when writing the docs for the "Trait hooks" section. The namespace `Livewire\With` does not exist.

I also sorted the imports to match the other code samples.


calebporzio:
Ooops, thanks!

PR #7880, opened 2024-02-02
Intersect Alpine plugin does not work when livewire component is updated

gdebrauwer:
Failing test for https://github.com/livewire/livewire/discussions/4480.

https://github.com/livewire/livewire/assets/22586858/d746f786-25f9-4c34-8cc0-d91ed5716e1c


calebporzio:
Thanks for the report, looks like `x-intersect` was initializing during "clone" (morph in Livewire), which is outside the viewport which caused this problem. I made `x-intersect` not load at all in the cloned tree like it shouldn't and it fixed it. Thanks.

PR #7881, opened 2024-02-02
Removing and re-adding a livewire component causes that livewire component to not work anymore

gdebrauwer:
Failing test for https://github.com/livewire/livewire/discussions/7111.

When a nested component is removed from the DOM and after some seconds added to the DOM again, an "Uncaught Component already initialized" error is thrown and the nested component does not work anymore (e.g. "wire:click" does nothing).

I have this issue when trying to implement multidrag in my https://github.com/nextapps-be/livewire-sortablejs package.

---

When I change the following in the `js/component.js` file, then the test passes but I'm not sure if there are particular reasons why that error is thrown there / why a component should not be reinitialized:
```js
export class Component {
    constructor(el) {
		// Before: if (! el.__livewire) throw 'Component already initialized';
        if (! el.__livewire) el.__livewire = this;

        // ...
    }
}
```


lrljoe:
I don't remember off the top of my head if window.Livewire.rescan() 
is still there, but might be worth a go as a fix.

Have you tried it with and without the morph markers injected?

If you're using the index as a wire:key, then pair it with something else, e.g. the index and primary key, that way when it shuffles them, it'll treat them as new and not get confused.

Would be interested in what you're doing with looped nested components tho!

gdebrauwer:
@lrljoe the Livewire.rescan() method exists but it has not been implemented yet: https://github.com/livewire/livewire/blob/177007cadeb7143d3e85bfba521a100dbe1c618c/js/lifecycle.js#L95-L97

Setting a `wire:key` does not help. When the component is re-added to the DOM, the construct-funtion of component.js called, so it is not that Livewire does not detect it has been re-added. Livewire's code is triggered. It is just that the construct-function is immediately stopped and the "Component already initialized" error is thrown.

calebporzio:
Ok I found the problem. What was happening:

* component element is initialized on page
* the component object is attached to the element (`el.__livewire`)
* the element is removed from the page
* Livewire's mutation observer is picking that up and cleaning up the component
* However, the `el.__livewire` still exists, so when it's re-added to DOM an error is thrown

I just added cleanup for `el.__livewire` and it's fixed now. 

Thanks!

PR #7883, opened 2024-02-02
Child component is being triggered by any Parent change

dbpolito:
Looking at documentation about how Reactive works, it made me think it would be somehow more similar to how Vue works.

Currently, as soon as you have a Reative component, any updates to the parent component will trigger a render on the child component... while i would expect it only to trigger child renders when a reactive prop is changed.

Is that by design?

I think a lot of apps would benefit to avoid those unnecessary renders.

### DIFF
--- a/src/Features/SupportReactiveProps/BrowserTest.php
+++ b/src/Features/SupportReactiveProps/BrowserTest.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features\SupportReactiveProps;
 
+use Illuminate\Support\Str;
 use Livewire\Livewire;
 use Livewire\Component;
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Yeah, currently the approach is simple as you described.

I agree, it would be great if Livewire were intelligent about when to update child components.

So, the basic logic would go something like this:
* A child component is encountered in the template when rendering the parent
* Store a reference to the variables being passed in
* After the parent is rendered, the children will be rendered
* Before rendering a child, first, check to see if the properties being passed in are different than the properties on the component
* If so, re-render, otherwise skip

Here are the bits that will add complexity to this solution:
A) Knowing if a child request is "skippable" in the first-place
  * We will need a way to know if a child payload is only their because it's a dependency of a parent? OR if the child payload is there because the child itself called a method or `$refresh` or something.
  * It seems like this would be an easy problem to solve because we can just look at the payload to see if there are any "updates", but it's not that simple, because `$refresh` doesn't add any updates, just adds the payload to the request
B) Diffing new props vs old props to determine if a child should re-render
  * The simplest thing to do would be to create like a checksum of the "last" passed in props, put that in the payload so that next time, a new checksum could be created, and they could be "diffed"
    * We also would need to know that this checksum is accurate and not missing data changes when PHP refuses to serialize closures or whatever, idk.
  * Otherwise, we would need to boot up the child component, JUST to detect prop differences. Which even then may not be easy. 

So yeah, there are a few hurdles to make this happen.

I'm going to close this PR and wait for more demand. Someone is welcome to take a crack at it given some of the things I outlined and they can use this PR as a starting point and then open a PR with an implementation.

Thanks a lot!

dbpolito:
@calebporzio Thanks a lot for your explanation, that all makes sense...

If i get time i will try to crack on this issue as it feels a good addition and you seems willing to address...

For someone else who lands into this issue, the way i workaround it was not removing Reactive and using a similar hash Livewire uses internally:

```php
@php
    $hash = crc32(json_encode($props));
@endphp

@livewire(
    'component',
    $props,
    key("component-{$hash}"),
)
```

This gives you the reactivity you need while skipping when the same props are passed.

PR #7884, opened 2024-02-03
Test Methods: Add Class Related Dusk Macros

lrljoe:
### Dusk Macros
This adds three new Dusk Macros for use in testing ,specifically in relation to HTML Classes.

This allows developers to pass an array of HTML Classes, and check for presence/absence in the resultant/rendered Component against a specified element.  This effectively extends the current  two existing methods:
assertHasClass/assertClassMissing

### The Problem
While you can presently use the assertHasClass/assertClassMissing, and write out multiple assertions in your Test it's not quite as intuitive if you're trying to check multiple classes are present/absent, or an element **only** has certain classes, as this involves having to add multiple assertions into your Test.

For example - checking that a table row has "text-black", "bg-white", and "text-xl" would result in three different assertions having to be written.
However - it won't detect that "text-blue" is also present in the element's class list, nor that it has a "hidden" class.

### Fix
To try and address this - in this PR are three new Macros, which allow for the use of an array of HTML classes, which should simplify checking for multiple HTML classes against an element found using the current HTML Selector approach.

#### assertMissingAllClasses
This is used to ensure that **all** classes in the $classNames array are missing from an element, i.e. you do not want any of the HTML classes to be present on an HTML Element.  It will fail if ANY of the HTML classes defined in the array **are** present.
For example - if passed "text-blue bg-black" it will not pass an element that has either of these present in it's classlist
```php
assertMissingAllClasses($selector, array $classNames) 
```

#### assertHasAllClasses
This is used to ensure that **all** classes in the $classNames array are present in an element.  It will fail if ANY of the HTML classes defined in the array **are not** present.
For example - if passed "text-blue bg-black" it will not pass an element that has neither of these present in it's classlist
```php
assertHasAllClasses($selector, array $classNames)
```

#### assertHasOnlyClasses
This is used to ensure that **only** classes in the $classNames array are present in an element, it differs from assertHasAllClasses in that any class in the element on the rendered component **must exist** in the array provided, but not all classes in the array are required to be present.
For example passing "text-blue bg-black" into it would permit an element with "text-blue bg-black", but not one that has "bg-white" in its classlist
```php
assertHasOnlyClasses($selector, array $classNames) 
```

In addition, one Alias
#### assertMissingClass()
This has been added but is just an alias for assertClassMissing, to try and get the naming convention consistent across all of these Macros
```php
assertMissingClass($selector, $className)
```

It seems useful

Yes

No

This is a DuskMacros only

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportTesting/DuskBrowserMacros.php
+++ b/src/Features/SupportTesting/DuskBrowserMacros.php
@@ -64,9 +64,92 @@ class DuskBrowserMacros
             PHPUnit::assertContains(
                 $className,
                 explode(' ', $this->attribute($selector, 'class')),
-                "Element [{$fullSelector}] missing class [{$className}]."
+                "Element [{$fullSelector}] is missing required class [{$className}]."
             );
 
+            return $this;
+        };
+    }
+
+    public function assertClassMissing()
+    {
+        return function ($selector, $className) {
+            /** @var \Laravel\Dusk\Browser $this */
+            $fullSelector = $this->resolver->format($selector);
+
+            PHPUnit::assertNotContains(
+                $className,
+                explode(' ', $this->attribute($selector, 'class')),
+                "Element [{$fullSelector}] has class that must be missing [{$className}]."
+            );
+
+            return $this;
+        };
+    }
+
+    public function assertMissingClass()
+    {
+        // Map to existing method
+        return $this->assertClassMissing();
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for this, however I think these are better kept in userland, a blog post, or a package.

PR #7888, opened 2024-02-04
Test for Reordering of Text Input fields

Progression:
This PR is a response to the discussion #7402 and the request for a test to illustrate the following behaviour which accord with Livewire v3. (It's my first ever PR is how I got this all right!)

### Issue:

If you have two or more input controls which are created by binding them to an associate php array and you then swap the order of the items on the server, they change position in the browser. BUT if you then change the text in any input control (that has moved position) will also overwrite and change the text in any other moved input control. Effectively the input controls mirror each others content even though they all still have unique wire:keys an d wire:model's and bind back to their corresponding data items on the server. This happens whether the wire:modal or wire.modal.live attributes are used.

Examples

If you have an associate array:

```
$this->items = [
                    ['id' => 1,  'value' => 'first'],
                    ['id' => 2,  'value' => 'second'],
                ];
```

Which you bind to input controls in a blade:

```
<ul>
    @foreach($items as $index => $item)
    <li wire:key='key-{{ $item['id'] }}'>
        <input type="text" wire:model.live="items.{{ $index }}.value">
    </li>
    @endforeach
</ul>

<button type='Button' wire:click='swap'>Swap</button>
```

And then click the swap button which corresponds to the server method:

```
public function swap() {
    $this->items = [
        ['id' => 2,  'value' => 'second'],
        ['id' => 1,  'value' => 'first'],
    ];
}

```

The input controls swap position and the resulting html rendered:

```
<ul>
    <!--[if BLOCK]><![endif]-->
    <li wire:key="key-2">
        <input type="text" wire:model.live="items.0.value">
    </li>
    <li wire:key="key-1">
        <input type="text" wire:model.live="items.1.value">
    </li>
</ul>
```
 
Change the text in either input control and both will be updated with the same value, they are no longer independant.

### Test:
There is a single test can_reorder_inputs_and_bind_data_unique that currently fails after the swap action is called, the input controls are swapped, the text is typed into one and the assertInputValue is check against the other input to ensure it hasn't changed.

I have tried to work out why this is happening but can't figure it out as everything looks correct after the dom has been updated.


Progression:
If anyone has any ideas of ways to try and get to the bottom of this bug then I'm happy to try and find a fix myself but it feels like quite a low level livewire/alpine thing and I;m complete at a dead end. Im stuck upgrading to LW v3 until I can get this working again. Cheers

joshhanley:
@Progression please be patient, as you stated it's likely to be something pretty deep. As such it's not something we can just grab and go, it requires deep thinking time.

But @DanielCoulbourne ran into a similar issue the other day and we worked around it by also putting the index into the key.

So for example, in the test, if you change the list item to this `<li wire:key="key-{{ $item['id'] }}-{{ $index }}">` the test then passes.

I've fixed a few issues with the test where `waitForLivewire()` needs to be called before the action happens so it knows to wait for the response. I also changed the single quotes to double quotes on the `wire:key` (not that it should matter).

Progression:
Thanks @joshhanley, sorry I didn't want to act inpatient as I know you guys are busy with lots of things. I was  hoping I might be able to help to easy the load. But I'll try and the workaround you suggested and if you need me to get involved in anyway to get this over the line then let me know.

joshhanley:
@Progression no worries! Yeah please let me know how that goes 🙂

I've edited my comment above with some ideas of what might be causing it if you did want to have a look. I suspect the issue is in the mutation observer though, so could be interesting 😬

Progression:
@joshhanley that work round works a treat, can't believe its that simply 🤦‍♂️ and for now this is great and allows me to move forward with the upgrade to v3! And @DanielCoulbourne big thanks for discovering the work around 🙌

The mutation observer issue is probably above my paygrade but I'm gonna try and follow along with what you guys end up doing and hopefully it help me get a better understand of Livewire for the future. Hope to be able to contribute more going forwards. 

Thanks Josh and @calebporzio for all the hard grafting on Livewire, along with all the other contributors 🤘

calebporzio:
Oh sheesh. I'm still working on the `wire:navigate` issue right now so haven't transferred my focus to this. But what you're saying makes sense Josh and could very well be. This is one of those issues that could either be really easy or insanely hard lol. Thanks for all the kind words @Progression and the willingness to pitch in

calebporzio:
I did some digging. I actually think this is the expected behavior. I fixed the test to pass given a more stable setup that you can use in your own app. (I'm going to close this because I consider this not a bug, but please re-open with a new failing test if you find something that is actually a bug)

Here's what the problem was:
* You were using PHP's array index in the wire:model of the inputs.
* When you re-ordered the inputs, you reset the indexes, so both inputs had new `wire:model` values in reverse.

When you do the re-order, you need to make sure the indexes you are using are preserved in the re-order and not re-assigned like PHP does in your case.

Hope this helps. Thanks.

PR #7889, opened 2024-02-05

apreiml:
In reply to #7858:

> I'm going to close this for now to keep things clean. Just submit a PR for this branch when you've done the above and I can re-review. Thank you (sorry for being a pain in the butt, I just want to be 100% sure this won't break things again.)

Understandable and thanks for still being open to consider this. 

This commit introduces support for non-local storages. Currently if you want to use a preview storage other than S3 and a local file system one, livewire fails to preview the files. I've documented the steps to reproduce here https://github.com/livewire/livewire/discussions/7466

It happens because the FilePreviewControler uses the pretendResponseIsFileFunction https://github.com/livewire/livewire/blob/177007cadeb7143d3e85bfba521a100dbe1c618c/src/Features/SupportFileUploads/FilePreviewController.php#L22-L25 which in turn expects the `$file` parameter to be a local file path and uses `filemtime` to get `$lastModified`. 
https://github.com/livewire/livewire/blob/177007cadeb7143d3e85bfba521a100dbe1c618c/src/Drawer/Utils.php#L52-L55

The mime type is also determined in a similar way in the [FilePreviewController](https://github.com/livewire/livewire/blob/177007cadeb7143d3e85bfba521a100dbe1c618c/src/Features/SupportFileUploads/FilePreviewController.php#L9)

`$lastModified` is used to check if the browser still has the file in cash and hence return 304 or send the full file otherwise. The `$mimeType` is required to set the content type in the header.

What this PR does is to put the `pretendResponseIsFile` cache code into a separate function that takes `$lastModified` and `$mimeType` as a parameter. The new function is called `cachedFileResponse`. This way the `pretendResponseIsFile` function can get `$lastModified` using `filemtime` and the new function `pretendPreviewResponseIsPreviewFile` will get `$lastModified` and `$mimeType` directly from the storage using `FileUploadConfiguration`, rather than assuming that those are local files. The latter function is now used in the FilePreviewController.

In the last PR I've had the new function return the headers as an array. I've overlooked the return on cache hit, which returned a response and then failed. In this PR, i've changed the function to accept a callback that builds the response. This way it can either return an empty response on cache hit or call the cb if not.

There are also now testcases for both situations: Download from a non local store and download on cache hit to make sure the bug is solved.

To support sending the header in the test, the If-Modified-Since header is now fetched from the Laravel Request object, rather than from $_SERVER.


calebporzio:
Thanks

PR #7892, opened 2024-02-05
Remove unused js imports and params

gdebrauwer:
_The failing tests seem to be caused by sushi's latest version 2.5.0_


calebporzio:
Thanks!

PR #7895, opened 2024-02-05
Fix `wire:navigate` back button "null" issue

calebporzio:
Discussion referencing this issue: https://github.com/livewire/livewire/discussions/6776

How to reproduce problem:
* On a site using `wire:navigate`
* Navigate to a ton of different pages (like 25)
* Then hit the back button a bunch of times
* Somewhere around ~15-20 the page will be replaced with "null" instead of an actual page

Here's the problem:
* Before livewire navigates to a new page it caches the HTML of the current page in `sessionStorage` with a key
* When the back button is pressed, it retrieves the HTML from `sessionStorage` using the associated key
* `sessionStorage` has ~5-10mb size limit
* So when lots of pages are stored in `sessionStorage`, it starts dropping old entries to make room for new ones

Potential solutions:
A) Store html directly in `history.state` entries
  * This is the simplest option, however each history state entry has a size limit of ~600kb so I don't think this will work for big pages
B) Just cache the HTML for pages in a global variable on the page
  * I think this is what Turbo drive and other tools do
  * I don't like it because it just starts building up the memory usage of the page with no upper limit
  * It also requires some logic to delete entries once a user "forks" the history path by visiting a new page after hitting "back" a bunch of times
C) A hybrid approach that starts with history.state, but detects size limits, and uses a local cache as a fallback

I will probably go with "B" for now as it's the most bullet-proof and straightforward


sergyu91ma:
I have this problem too. I can easily reproduce this if I use spa mode on app. 
I can provide a stub project if needed.

calebporzio:
Ok, it wasn't hard to implement a global "back button cache" in javascript memory. It works quite well.

The main problem now is: that in-memory object is wiped after the browser refreshes the page, or returns to the page after back-button visiting a page on a different domain or something.

So now I'm curious how other tools like Turbo do this. Do they have some mechanism to fall-back to normal browser back-forward button type stuff? They must, because I believe their "restoration" mechanisms are in plain JS memory.

edo2313:
Shouldn't you request the page again when you go back over a defined limit?
Having an "infinite" history would be heavier and heavier on resources as you continue navigating the site (assuming you keep the same browser tab open). Right now you only need to navigate ~10 pages on a simple website like the Livewire's documentation, I assume that on a heavier website (for example, dashboards with charts that get data passed from the server to an alpine component) the slow down would be noticeable after not that many navigations...

Also for some reason, i get 2 sessionStorage entries for every wire:navigate (still on Livewire's documentation), is it normal?

lele85:
I will link here some considerations about what i think it would be a good approach in general:

https://github.com/livewire/livewire/discussions/6776#discussioncomment-8304653

I'm particularly interested in the expected behavior of `@assets` and `@scripts` and how they are supposed to work when there's a cache hit.

I created here a simple wirebox that shows the issue:

- https://wirebox.app/b/4z1j8
- https://4z1j8.wirebox.app:8443

The expectation is that once you reach page B and you hit back the alert on page A should re-fire but it does not.
Maybe i'm missing something

joshhanley:
@calebporzio worked on this today and managed to work out what the issue was that you outlined in the `@todo`.

Simply put the stored state key was different to the cached snapshot after the page refresh, hence why the last back actually just showed the same key (it was off by one).

So I went digging into how Turbo does it, and yeah it is a similar process, but they use a deterministic key (the url) for caching pages.

I changed our solution to do that and it worked great for the bug in the `@todo` but when I pushed it up, it's now causing other navigate tests to fail.

The reason for this, is one of the tests for example `events_from_child_components_still_function_after_navigation` uses the same page, for navigating forward and back, so when you navigate forward and then hit back, it should restore the page from the cache but with the old state.

But as the cache is now determined by the URL, there is no way to distinguish between the state of the page in two different history states.

So I think we still would need to add some form of index/key for any URL that is the same.

When testing Turbo it was probably fine because I don't think they retain state between pages, so if you browse to the same page, it doesn't matter.

I also discovered another small bug, where by the original history item was being overwritten by the new page history item. So in this screenshot you should see `Quickstart` and `Installation` but instead there is two installations.

Screenshot 2024-02-26 at 4 26 22 pm

Where as after the fix, this is what is looks like (as it should).

Screenshot 2024-02-26 at 4 26 42 pm

I fixed it by moving the below block of code up.

image

Ultimately this is how I think the flow should go for different pages (which currently works):

1. Page loads, so store the page in the cache
2. Navigate is triggered, so update the current page in the cache (to ensure it has correct state), change pages, store new page in the cache
3. If back is hit, get the back page from the cache
4. If page is refreshed, store the page in the cache
5. If forward is hit, as the page was refreshed, there will be no cache for this page, so request from the server and store in the cache
6. Doing another back and then forward both should now return from the cache.

But in the scenario that the page that has been navigated to is the current page (which is currently broken):

1. Page loads, so store the page in the cache
2. Navigate is triggered, so update the current page in the cache (to ensure it has correct state), change pages, store new page in the cache (but if it is the same page, store it as a separate page maybe with a URL + time as the key?)
3. If back is hit, get key of stored time from state, and look for the current URL in the cache + key, if not found, look for URL without a key.
4. If page is refreshed, store the page in the cache
5. If forward is pressed (to the same page), look for the URL + time key and if found, use that, if not use the URL without key

Hopefully that will work but the current code will need refactoring to add the time key in.

- Here is the turbo code I referenced and used https://github.com/hotwired/turbo/blob/main/src/core/drive/snapshot_cache.js

joshhanley:
@calebporzio pushed some changes and, well, I didn't expect them all to pass! 😂 I don't quite think this is ready yet as I want to do some manual testing and check the snapshots etc are being loaded at the correct times.

calebporzio:
Great, thanks for your work on this @joshhanley

joshhanley:
@calebporzio have finished this off. Ran into a couple of bugs where the snapshot wasn't loading properly. Have fixed this and added some tests around it (mimicking how the Livewire docs are structured).

You will notice I added two new test helper methods `waitForNavigateRequest` and `waitForNoNavigateRequest`, but you may wonder why I added them when there is a `waitForNavigate` method. The reason is these actually test to see if a navigate fetch request is sent to the server or not, where as the `waitForNavigate` just checks a navigation has happened regardless of whether it is running from cache or network request.

This has allowed me to add some tests to ensure that pages are correctly loading from the snapshot cache if available.

In regards to the snapshot cache, I have set the limit to be 10 pages in the cache, which is the same amount that Turbo uses. https://github.com/hotwired/turbo/blob/1cd62711847511c7d1cf3c66ea0f6fc142466247/src/core/drive/page_view.js#L10

luanfreitasdev:
Hello!. Let me leave comments on some tests I did locally. We no longer have the null problem, but when I go forward and back a few pages I can receive this error in the console. However, this only happens if you are very quick in your actions.

Another observation is that this happens in the future. In other words, they appear after I leave the page or return. It doesn't crash the livewire and Alpine in any of the cases. Looking at the code, could this Alpine error message be removed?

Thank you

image

joshhanley:
@luanfreitasdev any chance you could create a small example repo outside of your main project that demonstrates the issue so we can investigate? Thanks!

luanfreitasdev:
Hello @joshhanley . For this problem to happen, it was necessary to test it in a large project, it would be not easy to try to do something close to this. 

But, errors in the console are not a problem, because the action was so quick that there wasn't time for Alpine to process it on the same page. Maybe this console error can be ignored

joshhanley:
@luanfreitasdev yeah I can't replicate it using a local copy of the Livewire docs. If you do manage to replicate it I will be happy to have a look.

luanfreitasdev:
I can't reproduce this error in other projects 🤝

luanfreitasdev:
If you want a reference to test dev-main, go to the powergrid demo. Thank you

https://demo.livewire-powergrid.com

PR #7898, opened 2024-02-06
Create dedicated EventHandlerDoesNotExist exception

levu42:
Based on @calebporzio's [comment on a recent PR](https://github.com/livewire/livewire/pull/7846#issuecomment-1917922330) I created a dedicated `EventHandlerDoesNotExist` exception.

@calebporzio I'm up for going through the code base and creating dedicated Exception types out of all `throw new \Exception` calls, I found 22 places in the code where a basic exception is currently thrown outside of tests - is that something you'd merge?

### DIFF
--- a/src/Exceptions/EventHandlerDoesNotExist.php
+++ b/src/Exceptions/EventHandlerDoesNotExist.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Livewire\Exceptions;
+
+class EventHandlerDoesNotExist extends \Exception
+{
+    public function __construct(public readonly string $eventName)
+    {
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @levu42!

Hmmm.... I hesitate to say yes, because I'd have to agree with all 22 decisions and sub-decisions on:
- The name of the class
- The location to place them when they're not in obvious feature folders
- The value of keeping them as a simple Exception or not

So I'll say no for the giant PR lol, but happy to field these on a case by case basis, or at least a group of them that you have high confidence in the decisions you make for them?

levu42:
@calebporzio okay, sounds good, I'll check them out and see where it makes sense to create specific exception classes.

PR #7909, opened 2024-02-08
Add `assertSetStrict()` and `assertNotSetStrict()` test assertions

joshhanley:
This PR adds `assertSetStrict()` and `assertSetNotStrict()` test assertions.

Currently, if you set a property to an empty string `->set('name', '')` and then assert if it is set to null `->assertSet('name', null)`, then the test will pass.

Instead you need to pass a third parameter `true` to `assertSet` to enforce strictness `->assertSet('name', null, true)` which will then fail as expected.

But the third parameter is ambiguous as to what it is.

So instead we can use the new assertions added in this PR `->assertSetStrict('name', null)`.

I experimented with other names, like `assertStrict()` but that didn't make sense when it came to `assertNotStrict()` so decided to keep the compound names.

Hope this helps!


calebporzio:
Thanks Josh!

joshhanley:
No worries 🙂

PR #7910, opened 2024-02-08
Fix query string updates for multiple components of same type with dynamics query string

wimski:
Relates to discussion: https://github.com/livewire/livewire/discussions/7862
<details>
<summary>Original bug description from discussion</summary>

# Setup
### Component Class
```php
use Livewire\Component;

class Test extends Component
{
    public string $value = '';

    #[Locked]
    public string $queryParamName;

    /**
     * @return array<string, array<string, string>>
     */
    protected function queryString(): array
    {
        // @see: https://livewire.laravel.com/docs/url#using-the-querystring-method
        // "The query string can also be defined as a method on the component.
        // This can be useful if some properties have dynamic options."
        return [
            'value' => ['as' => $this->queryParamName],
        ];
    }
}
```
### Component Template
```blade
<div>
    <ul>
        <li>value: {{ $value }}</li>
        <li>queryParamName: {{ $queryParamName }}</li>
    </ul>
    <button type="button" wire:click="$set('value', '{{ \Illuminate\Support\Str::random() }}')">set</button>
</div>
```
### Page
```blade
<livewire:test.test queryParamName="foo" />
<livewire:test.test queryParamName="bar" />
```
# Issue
When clicking the button in the first component I would expect the URL query parameters to be `?foo={SOME_STRING}`, but it becomes `?foo={SOME_STRING}&bar={SOME_STRING}`, where both instances of `SOME_STRING` are the `$value` of the first component. Why do both query strings get set and can this be remedied in some way?
</details>

@joshhanley wrote:
> * yeah that looks like it's a bug. Any chance you could submit a failing test PR? Thanks!
> * Just try to write the dusk test best you can, and then submit the PR and mention you couldn't get Dusk running.
> * Yeah if you can mostly get a test together that you think should run, just submit it and mention you couldn't get dusk running.


wimski:
@calebporzio is there an expected timeline for this? Because I'd like to know whether this will be fixed before the deadline of my project or that I need to implement some dirty temporary fixes :grimacing:

calebporzio:
This has been fixed (you can view the code to see the solution - was pretty simple thankfully)

wimski:
Thanks! I'll be keeping my eye out for the next release then.

PR #7911, opened 2024-02-08
FilePreviewController middleware from configuration

nino-giacca:
On a multi tenant such as [archtechx/tenancy](https://github.com/archtechx/tenancy) in the views the [temporaryUrl()](https://github.com/livewire/livewire/blob/f1cd35802dd8ca47e8f187c67ce5cb74ce1a07ec/src/Features/SupportFileUploads/TemporaryUploadedFile.php#L89) method fails to correctly return the path to the tenant where the temporary file is on the livewire-tmp folder.

It is therefore necessary to pass to [FilePreviewController](https://github.com/livewire/livewire/blob/main/src/Features/SupportFileUploads/FilePreviewController.php) the middleware set in the Livewire configuration file (temporary_file_upload.middleware) as is already done on [FileUploadController](https://github.com/livewire/livewire/blob/main/src/Features/SupportFileUploads/FileUploadController.php)

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first? Related to multi-tenancy support

2️⃣ Did you create a branch for your fix/feature? Yes

3️⃣ Does it contain multiple, unrelated changes? No

4️⃣ Does it include tests? No

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportFileUploads/FilePreviewController.php
+++ b/src/Features/SupportFileUploads/FilePreviewController.php
@@ -8,11 +8,12 @@ use Livewire\Drawer\Utils;
 
 class FilePreviewController implements HasMiddleware
 {
-    public static array $middleware = ['web'];
-
     public static function middleware()
     {
-        return array_map(fn ($middleware) => new Middleware($middleware), static::$middleware);
+        return array_map(fn ($middleware) => new Middleware($middleware), array_merge(
+            ['web'],
+            (array) FileUploadConfiguration::middleware(),
+        ));
     }
 
     public function handle($filename)


stancl:
Note that this changes the approach chosen in https://github.com/livewire/livewire/pull/7716

imo the proper approach to adding middleware to `FilePreviewController` is to set a static property. The reason I didn't go for using `FileUploadConfiguration::middleware()` there is to remain consistent with the middleware sets Livewire was already using.

Which by default includes an additional throttling middleware on the uploads route, but not on the previews route. In practice, there probably isn't much difference, but that is what should decide how the middleware are defined here:
- if `FileUploadController` and `FilePreviewController` should **always** use the same middleware, merge this PR
- if there are **any** cases for them using different middleware, stick with the static property or introduce a new config key

PR #7913, opened 2024-02-08
Allows trashed binding

mauro-baptista:
Issue:

Livewire doesn't consider the `->withTrashed()` in the route definition.

For instance:

```php
Route::get('/{server}', \App\Livewire\Server\Show::class)->name('show')->withTrashed();
```

If we call `site.com/servers/1` and the server id 1 is soft-deleted it will return a 404 error, even though we set the `->withTrashed()`

Changes:

This PR intends to consider the `->withTrashed()` when calling a Livewire component directly from the route.

The intention is to, when calling the `site.com/servers/1` where the server id 1 is deleted, to receive:
 
```php
Route::get('/{server}', \App\Livewire\Server\Show::class)->name('show'); //Returns 404
Route::get('/{server}', \App\Livewire\Server\Show::class)->name('show')->withTrashed(); //Returns the page with the model loaded

```


src/Drawer/ImplicitRouteBinding.php
```diff
@@ -104,20 +104,16 @@ protected function resolveParameter($route, $parameterName, $parameterClassName)
         $parent = $route->parentOfParameter($parameterName);
 
         if ($parent instanceof UrlRoutable && array_key_exists($parameterName, $route->bindingFields())) {
```
levu42, line 106: This nested `if {} else { if {} else {} }` looks like something that could be written really neatly with `match(true)`. Or at least I'd change it from a nested if to `if {} elseif {} else {}`.

joshhanley: @levu42 It was an if elseif elseif and it was harder to read. I tried a match but due to the different conditions this seemed the simplest way to write it and still have readable code. Open to suggestions though if you can share a nice looking code snippet 🙂

levu42: @joshhanley I was thinking of this:

```php
        $model = match(true) {
            $parent instanceof UrlRoutable && array_key_exists($parameterName, $route->bindingFields()) =>
                $parent->resolveChildRouteBinding($parameterName, $parameterValue, $route->bindingFieldFor($parameterName)),

            $route->allowsTrashedBindings() =>
                $instance->resolveSoftDeletableRouteBinding($parameterValue, $route->bindingFieldFor($parameterName)),

            default => $instance->resolveRouteBinding($parameterValue, $route->bindingFieldFor($parameterName))
        };
```

But it's also not the greatest in terms of readability.

calebporzio:
Thanks!

PR #7916, opened 2024-02-08
document method from underlying assertions class

macbookandrew:
This simply documents the magic `__call()` method call to [this test method](https://github.com/laravel/framework/blob/1d0f002172d1570389e72fb474d16a6996fc943d/src/Illuminate/Testing/TestResponse.php#L573-L580) in the Laravel framework.

This tells my IDE what the return type is so I can continue chaining on other test methods.


calebporzio:
I don't think I'm going to merge this because it's only one random assertion from the underlying Laravel class.

Maybe we could use a `@mixin` flag for IDEs that would be more comprehensive? What does Laravel do in these scenarios?

I'm going to close this. If you have good answers to those questions and want to open a new PR feel free. Thanks!

PR #7921, opened 2024-02-09
Flash data is being removed on initial render/dehydrate instead of subsequent renders

TWithers:
We use flash data at the bottom of a page to display alerts/toasts, not in a component. V2 of Livewire wouldn't clear flash data on the initial dehydrate, but would wait until a subsequent dehydrate. V3 of Livewire clears flash data when the component renders. 

Here is a wirebox showcasing the issue: https://wirebox.app/b/x6r67

SupportsRedirects.php in v2 Livewire
```php
        Livewire::listen('component.dehydrate.subsequent', function ($component, $response) {
            // If there was no redirect. Clear flash session data.
            if (empty($component->redirectTo) && app()->has('session.store')) {
                session()->forget(session()->get('_flash.new'));

                return;
            }
        });
```

SupportsRedirects.php in v3 Livewire
```php
    public function dehydrate($context)
    {
        // Put the old redirector back into the container.
        app()->instance('redirect', array_pop(static::$redirectorCacheStack));

        $to = $this->storeGet('redirect');
        $usingNavigate = $this->storeGet('redirectUsingNavigate');

        if (is_subclass_of($to, Component::class)) {
            $to = url()->action($to);
        }

        if ($to && ! app(HandleRequests::class)->isLivewireRequest()) {
            abort(redirect($to));
        }

        if (! $to) {
            // If there was no redirect. Clear flash session data.
            if (app()->has('session.store')) {
                session()->forget(session()->get('_flash.new'));
            }

            return;
        }

        $context->addEffect('redirect', $to);
        $usingNavigate && $context->addEffect('redirectUsingNavigate', true);
    }
```

### DIFF
--- a/src/Features/SupportRedirects/BrowserTest.php
+++ b/src/Features/SupportRedirects/BrowserTest.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features\SupportRedirects;
 
+use Livewire\Attributes\On;
 use Tests\BrowserTestCase;
 use Livewire\Livewire;
 use Livewire\Component;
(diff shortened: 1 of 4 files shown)


calebporzio:
Can you provide a better explanation of what the current problem you are facing is?

TWithers:
Sorry, Caleb. Got a notice from Josh as well to provide info here as well last night. I just updated the description.

I originally created a few failing tests per recommendation. I hope I did those correctly. I am not sure if I should have made a browser test instead, or if unit tests were good enough.

It also looks like a recent PR was merged pretty recently to change how flash data was handled for another reason.

I attempted a fix while at the office yesterday, but it looks like it failed the browser tests. I couldn't figure out how v2 Livewire implemented the `component.dehydrate.subsequent` event to try and replicate it inside v3. I am happy to continue attempting things, but you might be able to solve it much quicker than me.

calebporzio:
Thanks! Looks like the wirebox you referenced is private - can you make it public so I can test it out?
CleanShot 2024-02-12 at 10 19 38@2x

TWithers:
It is now public.

calebporzio:
Ok, thank you. I changed the implementation (see my most recent commit to this PR) to mirror V2, however, your provided Unit tests are failing after I made that change.

Do you mind looking into this and verifying if the tests need to be changed or the implementation needs to be?

I appreciate the help. Thank you.

TWithers:
Those unit tests were updated (I hope correctly), and are checking both that flash data exists after initial test component load, and that the data doesn't exist after a call like `$refresh`. 
Finally got Dusk working properly, so I added a browser test that confirmed everything as well.

calebporzio:
Great, thank you

PR #7925, opened 2024-02-11
Docs: alternative to wire:transition for loops

SRWieZ:
With #7257 I looked for an alternative and found that AutoAnimate works great for this limitation.

Since you [told me on Twitter](https://x.com/calebporzio/status/1749604669557457334?s=46&t=Y9VGtT5ZjYQU4aUVcHgn_Q) that you would be interested in a PR proposing an alternative, here it is!

I also wrote a full article [about it on my bog](https://srwiez.com/posts/how-to-have-transition-working-with-dynamic-list-on-livewire) this morning.

### DIFF
--- a/docs/wire-transition.md
+++ b/docs/wire-transition.md
@@ -48,6 +48,16 @@ Currently, `wire:transition` is only supported on a single element inside a Blad
 
 If one of the above comment `<li>` elements were to get removed, you would expect Livewire to transition it out. However, because of hurdles with Livewire's underlying "morph" mechanism, this will not be the case. There is currently no way to transition dynamic lists in Livewire using `wire:transition`.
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Ooooh this is awesome. Auto animate is such a fantastic tool and works perfectly for this usecase.

My only hesitation merging this is: it's a third-party solution and I often don't recommend those officially.

Lemme think about this a little more.

Thanks!

SRWieZ:
I get it, which is also why I wrote a blog post about it. So you have the choice between putting it in the documentation or leaving it in the userland.

calebporzio:
Ok, after thinking about it more I think I'm going to leave it out of the docs. I tried it locally yesterday and it works perfectly with Livewire. More people should know about this. Maybe I'll reference your post in a future Livewire update email. Thanks for bringing this to my attention @SRWieZ !

PR #7926, opened 2024-02-11
Update alpine.md

justincone:
Super small change: In this context, the correct spelling should be "complement" (with an "e", not an "i"). 

Compliment refers to saying something nice, while complement refers to something that completes or fits well with something else.


PhiloNL:
Thanks!

PR #7927, opened 2024-02-11
Fix flaky navigate test

joshhanley:
The `navigate_scrolls_to_top_and_back_preserves_scroll` seems to be failing randomly on most PRs. This PR fixes that.

I've run the test suite a few times to ensure it does actually fix the issue.

image


calebporzio:
legend

PR #7928, opened 2024-02-12
Add support for minified assets

joshhanley:
This PR adds support for using minified assets when app debug is set to false (same as how V2 worked) which fixes #6838 and #7620.

Source maps were missing for the minified file, so I have fixed that. I had to update the build script to output a source map only for the minified version of the bundle. So you will see a new `livewire.min.js.map` file and an update to `livewire.min.js` to reference the source map.

I also added some tests to ensure if someone has already overridden the scripts route, that it would still function as expected (using both standard and minified assets).

I did look at potentially registering a separate route for `livewire.min.js` but found it simpler to put the conditional inside the current `setScriptRoute()` call, otherwise we would need a `setMinifiedScriptRoute()` method.

Edit: There were a few other failing tests that failed now as app debug isn’t set to `false` by default and they were looking for `livewire.js`. So updated them to look for `livewire.min.js` instead.


src/Mechanisms/FrontendAssets/FrontendAssets.php
```diff
@@ -64,12 +68,16 @@ public static function livewireStyles($expression)
...
+            config('app.debug')
+                ? __DIR__.'/../../../dist/livewire.js'
+                : __DIR__.'/../../../dist/livewire.min.js'
+        );
     }
 
     public function maps()
     {
-        return Utils::pretendResponseIsFile(__DIR__.'/../../../dist/livewire.js.map');
+        return Utils::pretendResponseIsFile(__DIR__.'/../../../dist/livewire.min.js.map');
```
levu42, line 80: @joshhanley Doesn't this again need a `config('app.debug') ? ... : ...`?

joshhanley: @levu42 There's no map for a non-minified file, so I didn't think it was needed. But I could be wrong 😬

levu42: @joshhanley ah, I see! I didn't know the code before, just looked at the diff and thought you might have overlooked something :)

sammyaxe:
Thank you for this!

PR #7935, opened 2024-02-13
Fix form errors not triggering specific rule assertion

joshhanley:
Currently if you need to assert a form object property has failed a specific validation, and you write the following test, the test will fail instead of pass as expected:

```php
    ->set('form.title', null)
    ->call('save')
    ->assertHasErrors(['form.title' => 'required']);
```

The problem is the `failedRules` in the form object validator haven't been prefixed. So currently to work around this, you need to remove the `form.` prefix from the errors assertion:

```php
    ->set('form.title', null)
    ->call('save')
    ->assertHasErrors(['title' => 'required']);
```

While this works, it's unexpected that the first test doesn't function correctly.

So this PR fixes that by prefixing the `failedRules` the same as the `messages` are. But, while `messages` is a `MessageBag`, `failedRules` is just an array, so I've added a `prefixArray()` helper method.

Now you might be wondering why testing for an error without specifying the rule works (see below)

```php
    ->set('form.title', null)
    ->call('save')
    ->assertHasErrors('form.title');
```

The reason the above works, is because `assertHasErrors()` checks both `failedRules` and `messages` when no rule is specified so it is matching on the `messages` array.

### DIFF
--- a/src/Features/SupportFormObjects/Form.php
+++ b/src/Features/SupportFormObjects/Form.php
@@ -33,6 +33,7 @@ class Form implements Arrayable
             return $this->parentValidate($rules, $messages, $attributes);
         } catch (ValidationException $e) {
             invade($e->validator)->messages = $this->prefixErrorBag(invade($e->validator)->messages);
+            invade($e->validator)->failedRules = $this->prefixArray(invade($e->validator)->failedRules);
 
             throw $e;
         }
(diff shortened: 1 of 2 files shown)


calebporzio:
Perfect. Makes total sense. Thank you @joshhanley !

PR #7939, opened 2024-02-13
Save latest Alpine (v3.13.5) in lockfile

CasEbb:
The `package-lock.json` file now contains an incompatible set of packages.

* `@alpinejs/anchor` is locked to 3.13.3
* `alpinejs` is locked to 3.13.1
* `@alpinejs/anchor` uses `Alpine.interceptClone`
* `Alpine.interceptClone` was added in `alpinejs` 3.13.3

This results in `Uncaught TypeError: Alpine3.interceptClone is not a function` when running Dusk

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,15 +7,15 @@
             "name": "laravel-livewire",
             "license": "MIT",
             "dependencies": {
-                "@alpinejs/anchor": "^3.13.3",
-                "@alpinejs/collapse": "^3.10.5",
-                "@alpinejs/focus": "^3.10.5",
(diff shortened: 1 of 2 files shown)


edalzell:
FYI, I'd recommend using something like https://github.com/pixelfear/composer-dist-plugin, it allows you to download the assets instead of putting them in the repo.

All my packages use it as does Statamic. Much nicer to not have to commit compiled code.

calebporzio:
Thanks @edalzell and good tip, we should really move to something more robust like that to avoid these problems

PR #7941, opened 2024-02-14
Fix deferred loading with alpine

JonathanLouw:
This relates to: https://github.com/livewire/livewire/discussions/7475

After upgrading an app from v2 to v3 there's a bug with [deferred loading logic](https://laravel-livewire.com/docs/2.x/defer-loading) that was previously working in v2.

Take the following component for example that previously worked:

```php
class DeferExample extends Component
{
    public bool $load = false;

    public function loadUsers(): void
    {
        $this->load = true;
    }

    public function render()
    {
        return view('livewire.defer-example', [
            'users' => $this->load
                ? User::query()->inRandomOrder()->limit(5)->get()
                : collect(),
        ]);
    }
}
```
```blade
<div>
    <button wire:click.prevent="loadUsers">Load</button>
    <div x-data="{ users: @js($users) }">
        <template x-for="user in users" :key="user.id">
            <div x-text="user.name"></div>
        </template>
    </div>
</div>

```
Now in v3 this is broken and outputs this error to the console:
```console
Alpine Expression Error: user is not defined

Expression: "user.name"
```
And this is the source code throwing the exception:
```js
(async function anonymous(__self,scope
) {
with (scope) { __self.result = user.name }; __self.finished = true; return __self.result;
})
```

As instructed by @joshhanley I've opened this PR with a failing test that replicates the above scenario in a browser test.

Main goal of this PR is to support deferred loading functionality which was possible in v2 and I think this is a genuine bug worth looking into because after initially upgrading to v3 I had similar alpine expression errors like this one happening almost everywhere. But after upgrading to the latest version (3.4.4) they all seem to be gone except for this one error which is still happening.

The two test cases added cover a pure blade implementation of deferred loading which works fine and passes as expected but the second test case added is for deferred loading of data into an alpine component which fails.

### DIFF
--- a/src/Features/SupportDeferredLoading/BrowserTest.php
+++ b/src/Features/SupportDeferredLoading/BrowserTest.php
@@ -78,13 +78,11 @@ class BrowserTest extends BrowserTestCase
                     <div>
                         <button dusk="load" wire:click="loadUsers">Load Users</button>
                         <div x-data="{ users: @js($this->users) }">
-                            <template x-if="users.length > 0">
-                                <ul>
-                                    <template x-for="user in users" :key="user.id">
-                                        <li x-text="user.name"></li>
-                                    </template>
-                                </ul>
-                            </template>
+                            <ul>
+                                <template x-for="user in users" :key="user.id">
+                                    <li x-text="user.name"></li>
+                                </template>
+                            </ul>
                         </div>
                     </div>
                 BLADE;
--- a/src/Features/SupportDeferredLoading/BrowserTest.php
+++ b/src/Features/SupportDeferredLoading/BrowserTest.php
@@ -12,7 +12,7 @@ use Illuminate\Foundation\Auth\User as AuthUser;
 class BrowserTest extends BrowserTestCase
 {
     /** @test */
-    public function can_defer_loading_a_collection_of_models_in_blade()
+    public function can_defer_loading_of_a_collection_of_models_in_blade()
     {
         Livewire::visit(new class extends Component {
             public bool $load = false;
@@ -53,7 +53,7 @@ class BrowserTest extends BrowserTestCase
     }
 
     /** @test */
-    public function can_defer_loading_a_collection_of_models_into_alpine_state()
+    public function can_defer_loading_of_a_collection_of_models_into_alpine_state()
     {
         Livewire::visit(new class extends Component {
             public bool $load = false;
--- a/src/Features/SupportDeferredLoading/BrowserTest.php
+++ b/src/Features/SupportDeferredLoading/BrowserTest.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace Livewire\Features\SupportDeferredLoading;
+
+use Sushi\Sushi;
+use Livewire\Livewire;
+use Livewire\Component;
+use Tests\BrowserTestCase;
+use Livewire\Attributes\Computed;
+use Illuminate\Foundation\Auth\User as AuthUser;
+
+class BrowserTest extends BrowserTestCase
+{
+    /** @test */
+    public function can_defer_loading_a_collection_of_models_in_blade()
+    {
+        Livewire::visit(new class extends Component {
+            public bool $load = false;
+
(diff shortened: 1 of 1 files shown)


calebporzio:
## PLEASE IGNORE FOR NOW
> (I'm just leaving these notes here for when I return to this problem)

Ok, here's what's happening:

The page loads with this HTML:

```html
<div x-data="{ users: [] }">
    <ul>
        <template x-for="user in users" :key="user.id">
            <li x-text="user.name"></li>
        </template>
    </ul>
</div>
```

Then the "load" button is pressed. Livewire sends a request and comes back with the following HTML:

```html
<div x-data="{ users: JSON.parse('[...]') }">
    <ul>
        <template x-for="user in users" :key="user.id">
            <li x-text="user.name"></li>
        </template>
    </ul>
</div>
```

Then the "morph" happens.

* It goes element by element comparing the old with the new
* As it goes it, initializes Alpine on that individual element in the "new" tree before diff/patching
* When it gets to `<div x-data`, it registers that there's data (because of "has-alpine-data")
* It transfers that data over to the new tree
* It "morphs" the current HTML with a new x-data directive
* Then keeps morphing withe the run-time data so there are no `<li>`s rendered
* After all this, mutation observer picks up the x-data change.
* It tears down the old data, and re-initializes with the new data. However, because `x-for` relies on the old `users` instance which is no longer there, the `x-for` doesn't react to the change at all.

Then the second button click happens with the second morph:

* Goes through element by element
* Encounters `<div x-data`, sees runtime data (this time the new runtime data with the users).
* Transfers that data to the new tree
* Keeps walking the DOM, encounters "x-for", it adds the "li" nodes, but doesn't initialize them because it's cloning and they will get initialized as the tree walker walks
* The "li"s never get initialized in the new tree because they are compared against "text" nodes when the patch is happening to add them in. So now the only way they'll get initialized is by the mutation observer after the morph
* The mutation observer runs and tries to initialize them
* They are initialized, but because they were added from the clone tree and not naturally in the x-for in the live tree, they don't have x-for scope to see up the tree and result in that error that get's shown.

Briefly, I should say, in Livewire V2, the fact that this worked was kind of a fluke. Basically you're expecting Alpine to detect a change to the `x-data` attribute and tear-down, then re-initialize itself. That's a problematic pattern we've moved away from. But I understand there isn't currently a simple alternative to this pattern. Something to discuss for a future API for sure.

Ok, back to what's going on with this error.

When Livewire goes to "morph" the existing HTML on the page, with the new HTML from the server, it first initializes Alpine on the "clone tree" (the new HTML received from the server) before morphing it into the existing HTML.

During this phase, it compares the new with the old and looks for any state that exists on the old (live DOM) as it crawls.

When it gets to `x-data`, it sees the existing run-time state, and transfers it to the clone tree for initialization there.

So even though the HTML has the new users object in x-data, the runtime data doesn't. So in the clone tree, the new `x-for` elements are NOT added to the HTML, and therefore not morphed onto the page.

This is why when you click the button the first time, nothing happens. (the error happens after the second click).

Ok, let's move on to the second click:

The following HTML is received from the server (the same as last time):

```html
<div x-data="{ users: JSON.parse('[...]') }">
    <ul>
        <template x-for="user in users" :key="user.id">
            <li x-text="user.name"></li>
        </template>
    </ul>
</div>
```

Now, this time here's what happens when "morphing"

joshhanley:
@JonathanLouw I've done an investigation into how this behaviour functions in V2, and I found that if I copy your test (and amend it for V2), the tests fails still.

What Caleb describes above with the list loading after the second click does with in V2 (where as it causes an error in V3).

So I just want to clarify the scenario you are using it in V2, do your users have to press the "Load" button twice for this to actually work?

Would you be able to provide a small demo of V2 code that works for you so I can investigate it, to make sure we are matching the behaviour in V3?

Thanks!

calebporzio:
Thanks for digging in on this. Because it's not really supported behavior (modifying `x-data` at runtime, I'm going to close this)

PR #7948, opened 2024-02-14
EnumSynth tryFrom

jmonday:
When hydrating within `EnumSynth` use `tryFrom()` instead of `from()`. The following describes a use case scenario using enums for select options. For example, the enum:

```php
<?php

namespace App\Enums;

enum Status: string
{
    case ENABLED = 'enabled';
    case DISABLED = 'disabled';
}
```

The Livewire form component:

```php
<?php

namespace App\Livewire\Forms;

use App\Enums\Status;
use Illuminate\Validation\Rule;
use Livewire\Form;

class AccountForm extends Form
{
    public ?Status $status = null;

    {
        return [
            'status' => ['nullable', Rule::enum(Status::class)],
        ];
    }
}
```

The Blade template:

```blade
<select wire:model="form.status">
    <option>Show All</option>

    @foreach(\App\Enums\Status::cases() as $status)
        <option value="{{ $status->value }}">{{ $status->name }}</option>
    @endforeach
</select>
```

Selecting one of the enum options works as expected. When attempting to clear the input via the empty `<option>`, we'll see the following error:

```
ValueError  "" is not a valid backing value for enum "App\Enums\Status".
```

Another way to quickly show the issue is via `tinker`:

```
$ php artisan tinker
> \App\Enums\Status::from('')

   ValueError  "" is not a valid backing value for enum "App\Enums\Status".

> \App\Enums\Status::tryFrom('')
= null
```

### DIFF
--- a/src/Mechanisms/HandleComponents/Synthesizers/EnumSynth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/EnumSynth.php
@@ -16,7 +16,7 @@ class EnumSynth extends Synth {
     static function hydrateFromType($type, $value) {
         if ($value === '') return null;
 
-        return $type::from($value);
+        return $type::tryFrom($value);
     }
 
     function dehydrate($target) {
@@ -31,6 +31,6 @@ class EnumSynth extends Synth {
 
         $class = $meta['class'];
 
-        return $class::from($value);
+        return $class::tryFrom($value);
     }
 }


levu42:
As I see the code, I think this is not the best approach to fix it: `tryFrom` fails silently to `null` instead of throwing an exception when trying to assign an invalid value. I want my code to throw exceptions when I try to assign `foo` to an enum that only knows `bar` and `baz`.

The real problem seems to be, that setting an empty string on a nullable type should convert it to null and you describe it as not working. There was something related to it (but in query strings) recently with #7868, that's not yet released.

calebporzio:
I agree @levu42 - I think this would silence valuable errors. However I understand that when deselecting something you want it to be "null"

I'm fine having special behavior for empty strings with backed enums to set the property to null.

Mind tweaking this PR to account for that differently. Thanks!

calebporzio:
Also, we will need a simple unit test to merge this, if you can add that too. Thanks!

levu42:
I think this is not actually a bug, this is just wrong HTML in the example. From MDN about the `value` attribute of the `<option>` tag:

> If this attribute is omitted, the value is taken from the text content of the option element.

So the HTML that's presented here isn't actually sending `''` to the server, but `'Show All'` which - correctly - fails. The desired behavior most likely can be achieved by adding `value=""` to the option tag. (without having it tried out myself, this is just from looking at the code)

rynomars:
I am curious how this solution is different than how a null value is handled in the IntSynth class?
https://github.com/livewire/livewire/blob/b71ba641912a26ee31a6e92df44a446c64a76dd5/src/Mechanisms/HandleComponents/Synthesizers/IntSynth.php#L18

The tryFrom method will throw a ValueError if anything other null or. valid enum value is provided.

jmonday:
> So the HTML that's presented here isn't actually sending `''` to the server, but `'Show All'` which - correctly - fails. The desired behavior most likely can be achieved by adding `value=""` to the option tag. (without having it tried out myself, this is just from looking at the code)

Correct. The code in the original description is contrived. Sorry, I had experimented with many options before opening the PR and forgot to update my example. Using `value=""` will still attempt to assign an empty string to an enum though.

levu42:
> The tryFrom method will throw a ValueError if anything other null or. valid enum value is provided.

That is wrong, from the PHP docs:

> If there is no matching case defined, it will return null.

What's actually correct as a bugfix here is the mentioned code: `if ($value === '' || $value === null) return null; ` - I just reproduced it locally and working on a unit test for it now.

jmonday:
Closing in favor of https://github.com/livewire/livewire/pull/7955

PR #7950, opened 2024-02-14
Failing Test: Morph Marker when If contains calculation

fotrino:
This bug is related to [Discussion 7819](https://github.com/livewire/livewire/discussions/7819) &  [Discussion 6663](https://github.com/livewire/livewire/discussions/6663).

Recently I run into a bug caused by Injecting Morph Markers, which causes the following error in the console.

Screenshot 2024-02-14 at 6 16 45 PM

This is the basic setup to replicate the issue:

```php
class Todo extends Component
{
    public $someProperty = 1.5;
    
    public function render() 
    {
        return '<div> @if ($someProperty > 0) <span> {{ $someProperty }} </span> @endif </div>';
    }
}
```

Instead of wrapping the `span`element with `<!--[if BLOCK]><![endif]-->` & `<!--[if ENDBLOCK]><![endif]-->`, it only renders the ENDBLOCK.
```html
<div>
    <span> 1.5 </span>
</div>
```

### DIFF
--- a/src/Features/SupportMorphAwareIfStatement/UnitTest.php
+++ b/src/Features/SupportMorphAwareIfStatement/UnitTest.php
@@ -47,6 +47,23 @@ class UnitTest extends \Tests\TestCase
     }
 
     #[Test]
+    public function handles_if_statements_with_calculation_inside()
+    {
+        Livewire::component('foo', new class extends \Livewire\Component {
+            public $someProperty = 1.5;
+
+            public function render() {
+                return '<div> @if ($someProperty > 0) <span> {{ $someProperty }} </span> @endif </div>';
+            }
+        });
+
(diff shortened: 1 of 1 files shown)


calebporzio:
Mmmm yeah, this is a tricky one, what's happening is: the parser sees `@if` before `>` and assumes it's inside an HTML tag so it skips adding it.

fotrino:
Yes, I've been trying a few different combinations but it blows up elsewhere when i run the test suite. Maybe there is a regex super genius out there that can help 😄

joshhanley:
Yep couldn't get it working just by modifying the regex. Got a different strategy working and pushed up a new PR #7967.

calebporzio:
Closing as @joshhanley fixed it in #7967

PR #7955, opened 2024-02-15
EnumSynth: Fix synthing empty string into null

levu42:
This is a fix for the bug reported first in #7948.

Currently, I can set a public property of a component to have a nullable enum as type. It works well to assign an enum case to it, however, it doesn't work, trying to set that property to an empty value:

```php
class Foo extends Component {
  public ?StatusFilterEnum $status = null;
}
```

```html
<select wire:model="filter">
  <option value="">Show all</option> <!-- selecting this breaks -->
  <option value="active">Show only active users</option>
  ...
</select>
```

This PR fixes the logic in EnumSynth to be the same as in other synths (like IntSynth) to convert also empty strings into `null`.


calebporzio:
Perfect, thanks!

PR #7958, opened 2024-02-15
Fix `CarbonSynth` to use type matching

rynomars:
The CarbonSynth does not seem to be working as expected. A string is trying to be assigned to the public property. I have created Test class CarbonSynthUnitTest to demonstrate the issue.

### DIFF
--- a/src/Mechanisms/HandleComponents/Synthesizers/CarbonSynth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/CarbonSynth.php
@@ -27,6 +27,10 @@ class CarbonSynth extends Synth {
         return false;
     }
 
+    static function matchByType($type) {
+        return is_subclass_of($type, DateTimeInterface::class);
+    }
+
     function dehydrate($target) {
         return [
             $target->format(DateTimeInterface::ATOM),
@@ -34,7 +38,15 @@ class CarbonSynth extends Synth {
         ];
     }
 
+    static function hydrateFromType($type, $value) {
+        if ($value === '' || $value === null) return null;
+
+        return new $type($value);
+    }
+
     function hydrate($value, $meta) {
+        if ($value === '' || $value === null) return null;
+
         return new static::$types[$meta['type']]($value);
     }
 }
--- a/src/Mechanisms/HandleComponents/Synthesizers/Synth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/Synth.php
@@ -44,7 +44,7 @@ abstract class Synth {
         }
 
         if ($method === 'hydrateFromType') {
-            throw new \Exception('You must define a "hydrate" method');
+            throw new \Exception('You must define a "hydrateFromType" method');
         }
 
         if ($method === 'get') {
--- a/src/Mechanisms/HandleComponents/Synthesizers/Tests/CarbonSynthUnitTest.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/Tests/CarbonSynthUnitTest.php
@@ -1,0 +1,142 @@
+<?php
+
+namespace Livewire\Mechanisms\HandleComponents\Synthesizers\Tests;
+
+use Carbon\Carbon;
+use Carbon\CarbonImmutable;
+use Livewire\Component;
+use Livewire\Livewire;
+
+class CarbonSynthUnitTest extends \Tests\TestCase
+{
+    /** @test */
+    public function public_carbon_properties_can_be_cast()
+    {
+        $testable = Livewire::test(ComponentWithPublicCarbonCaster::class)
+            ->updateProperty('date', '2024-02-14')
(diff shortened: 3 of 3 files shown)


levu42:
Yesterday I thought I understood the synth code in the background, but now this confuses me:

Apparently `Livewire::test(...)->updateProperty()` doesn't use synths: It goes `HandleComponents::update()` => `HandleComponents::updateProperties()` => `HandleComponents::updateProperty()` - but then inside `updateProperty` it doesn't use the synth, if it's not a "deep" property, it calls `Handlecomponents::setComponentPropertyAwareOfTypes` which doesn't use synths.

@calebporzio How do you want to fix this? Do you want `updateProperty` to use a synth also for non-deep updates? Or is there a logic error in other places?

calebporzio:
So first, let me say that this problem only occurs with carbon properties that are initially null. I can get the tests to pass by setting the properties to some date initially like so:

```php
class ComponentWithPublicCarbonCaster extends Component
{
    public Carbon $date;

    public function mount()
    {
        $this->date = now();
    }
```

Here's what's happening in the WORKING scenario above:
* `$this->date` is set to a carbon instance in mount
* it gets "dehydrated" into a string value with a flag that it's a carbon instance for next hydrate
* The user triggers a string update for this property
* Livewire receives the string and knows to "hydrate" it into a carbon instance because of the existing property being a carbon instance

Here's what's happening without mount in the BROKEN scenario:
* `$this->date` is null on mount
* it gets "dehydrated" into "null" with no signal that it's a carbon instance
* The user triggers a string update for this property
* Livewire receives the string and has no idea that it's supposed to be a carbon instance, so it tries to set the property to a string and gets an error.

Here are a few solution paths of the top of my head:
A) When Livewire is dehydrating and encounters a "null" or unset value, check for types on that property and dehydrate "null" along with the type so that it can be used to re-hydrate.
B) Enhance synthesizers to be "type-aware". This way, if no existing value is there to re-hydrate from, they can notice the typehint and hydrate from a string that way.

"B" is clearly the more robust and better option.

It would also solve other problems in the codebase like this: https://github.com/livewire/livewire/blob/f7c68ef294ea212a30924b457939cf026400d3f4/src/Features/SupportAttributes/Attribute.php#L68

calebporzio:
I'm picturing an addition to the synth class like "hydrateFromType" or something:
CleanShot 2024-02-16 at 10 05 47@2x

rynomars:
@calebporzio I added the following methods to the CarbonSynth class to get the tests to pass:
```
static function matchByType($type) {
        return is_subclass_of($type, DateTimeInterface::class);
}
```
```
 static function hydrateFromType($type, $value) {
        if ($value === '' || $value === null) return null;

       return new $type($value);
}
```

I also added the empty string and null check to the hydrate method:
```
function hydrate($value, $meta) {
        if ($value === '' || $value === null) return null;
```

These allow all tests to pass except `public_carbon_properties_throws_exception_when_set_to_null_cast`

This is testing that an exception is thrown when we attempt to update a public property that does not allow null.
```
class ComponentWithPublicCarbonCaster extends Component
{
    public Carbon $date;

    public function render()
    {
        return view('null-view');
    }
}
```
I was not sure if this was a case that needed to be handled.

It presents a challenge because we don't pass the allowNulls property to the getSynthesizerByType method, just the name.
` $synth = $this->getSynthesizerByType($type->getName(), $context, $path);`

rynomars:
@calebporzio Please let me know if the posted change is a valid solution and I can update my PR. If not, please feel free to provide direction and I would be glad to apply appropriate changes.

calebporzio:
Lol, I'm a dummy as soon as I posted that I started looking in the codebase and apparantly I already implemented "hydrateFromType" lol. I'll look at your updates now haha

calebporzio:
And yep! What you have is almost exactly what I have, so you can update the PR and we'll merge!

rynomars:
@calebporzio There is still a failing test, `public_carbon_properties_throws_exception_when_set_to_null_cast`
This is testing that an exception is thrown when we attempt to update a public property with null that does not allow null.
```
class ComponentWithPublicCarbonCaster extends Component
{
    public Carbon $date;

    public function render()
    {
        return view('null-view');
    }
}
```
I was not sure if this was a case that needed to be handled.

It presents a challenge because we don't pass the allowNulls property to the getSynthesizerByType method, just the name.
`$synth = $this->getSynthesizerByType($type->getName(), $context, $path);`

Thoughts?

calebporzio:
Mmm yep, tricky. I'm out of town right now, so can't really wrap my head around it but I'll say this:

If what we have now currently:
- expands support for carbon instances to support null
- doesn't break anything else

We can remove this test and merge this. Then we can work on that next if we need.

What do you think?

rynomars:
That sounds good to me! Thanks @calebporzio!

calebporzio:
Thank you!

PR #7963, opened 2024-02-16
Added check to see if a route has already been registered

chrispage1:
A simple if around the core setUpdateRoute to check if one has already been defined.

This allows a user to call Livewire::setUpdateRoute() within the RouteServiceProvider register method without causing any problems.

I've also raised this as discussion https://github.com/livewire/livewire/discussions/7962

Thanks!


calebporzio:
Sounds good, thanks

PR #7965, opened 2024-02-16
Preserve scroll on `wire:navigate`

CasEbb:
`wire:navigate` is not behaving consistently right now.

- According to the docs, the scroll position should be preserved on navigate. "_By default, Livewire will preserve the scroll position of a page when navigating back and forth between pages._" (https://livewire.laravel.com/docs/navigate#preserving-scroll-position)
- The test suite however, asserts the opposite.
  https://github.com/livewire/livewire/blob/d5ba3e9e7f8e811a8ab115dfbf30f1a82c22215a/src/Features/SupportNavigate/BrowserTest.php#L462
- The test passes, because scroll preservation on navigate is broken (assuming the docs are right).

This PR:

- Updates the test so that it tests for preserving scroll (like documented)
- Fixes the Javascript portion for `wire:navigate`. `document.body.scrollTop` is used, which is deprecated in modern browsers (its value will always be `0`). It uses `document.documentElement` instead (where `scrollTop` and `scrollLeft` will actually have a value).


PerryvanderMeer:
I assume it’s intended to only preserve the scroll position when navigating back and forth, not when actually navigating to a new page?

CasEbb:
On a side note, getting a reproducible `dist/` build is quite a challenge right now. 😅

- I was getting a lot of diffs on the comments in livewire/livewire.js ("`// ../alpine`" => "`// node_modules/alpine`")
- This is - of course - a consequence of Alpine not being locally linked. So I followed the instructions of installing Alpine locally [from the docs](https://livewire.laravel.com/docs/contribution-guide#optional-set-up-alpine-for-local-development).
- The docs mention linking `alpinejs` and `@alpinejs/morph`, but to get a reproducible build, however, you need to link `@alpinejs/collapse` `@alpinejs/focus` `@alpinejs/intersect` `@alpinejs/persist` `@alpinejs/anchor` and `@alpinejs/mask` as well.
- Livewire now builds without diffs. 🎉 Except for one:
- Each build produces a new `dist/manifest.json` given that it does not contain a checksum but [a random hash](https://github.com/livewire/livewire/blob/d5ba3e9e7f8e811a8ab115dfbf30f1a82c22215a/scripts/build.js#L22).

There's a problem now, however:

- Whenever buildling Livewire's JS assets, it uses my local checked out AlpineJS repo. 
- Which means that whatever I have checked out locally, finds its way into livewire/livewire.js.
- This time, it happens to be `main` - but it could be anything. An out of date `main`, another branch, tag, etc.
- I would even say that using `main` might be potentially dangerous. Unreleased AlpineJS code now finds its way into livewire/livewire.js (which is actually happening right now).

So a little warning is in order here to always, before every release, run a build with a version of AlpineJS as it appears in package-lock.json.

joshhanley:
@CasEbb yeah that is correct, Caleb will build assets directly from Alpine main if it has a fix on it that Livewire depends on. Alpine and Livewire won't necessarily have had a release yet, hence why relying on building from what is in `package.lock` isn't always reliable. Caleb generally ensures that the latest build is done from his machine with the latest Alpine when merging, so its not a huge issue, but its a good idea to make sure you are on Alpine main branch with latest changes pulled.

Ah yep, docs are wrong, all the packages should be listed for linking. Feel free to PR changes to it 🙂

Re the manifest, yeah if there are no changes in any of the built asset files, then you can safely skip merging the manifest changes. Its not a hash, just an indicator for browser cache invalidation, so it only looks to see if it has changed or not. It has no bearing on the actual contents of the asset files.

joshhanley:
Regarding this PR, we added scroll to the top after the docs, so it's likely that it's actually the docs that are out of date. This is because when someone navigates to a new page, they generally would expect to be at the top of the new page, not half way down it. I haven't looked at your changes in detail yet so not sure if that will be impacted or not.

CasEbb:
That explains a few things. I was led astray by the fact that in navigate/scroll.js an attempt is made to preserve the scroll position of the viewport, but because `document.body` was used it didn't work, and thus resulted inadvertently in the expected behaviour.

In that case, I would suggest implementing a 'preserve-scroll' modifier for wire:navigate, so that scroll to top is actually the default behaviour from now on and you still retain the option to preserve the viewport scroll.

philippbuerger:
This "preserve-scroll" modifier would fix so many problems for me .. is it implemented already?

PR #7967, opened 2024-02-17
Change morph markers strategy

joshhanley:
This PR is a follow up of the failing test in PR #7950.

The main issue is that if we have a directive that contains a `>` symbol, `@if ($someProperty > 0)` then the morph markers would not get injected as the regex was thinking that the directive was inside a HTML element.

I couldn't see a way to get regex to work to detect if the `>` symbol was actually a part of the directive (aka inside parenthesis). But Laravel's Blade compiler has an implementation to detect a directive and capture the directive plus it's expression (parens and anything inside of it, regardless of multiple levels of nested parens).

So I copied the implementation from the Blade compiler and modified it to suit our needs, and it works great!

I've yet to do any benchmarking though to see if there is a performance impact or not. I will wait to see what @calebporzio thinks of this PR first.

I decided to open this as a new PR incase we don't want to go with this strategy, that way we can leave the failing test PR #7950 open to investigate further.

Also there were a bunch of commented out testcases that I reenabled and now pass 😁

image


calebporzio:
Josh, this is smaaart, thank you for this.

joshhanley:
For reference, here is the method I copied from the Blade compiler (which is now called `compileDirectives()` in our implementation)

https://github.com/laravel/framework/blob/cf9513219e80608aded783e33a30f8e8d145f6c2/src/Illuminate/View/Compilers/BladeCompiler.php#L521-L567

```php
/**
 * Compile Blade statements that start with "@".
 *
 * @param  string  $template
 * @return string
 */
protected function compileStatements($template)
{
    preg_match_all('/\B@(@?\w+(?:::\w+)?)([ \t]*)(\( ( [\S\s]*? ) \))?/x', $template, $matches);

    $offset = 0;

    for ($i = 0; isset($matches[0][$i]); $i++) {
        $match = [
            $matches[0][$i],
            $matches[1][$i],
            $matches[2][$i],
            $matches[3][$i] ?: null,
            $matches[4][$i] ?: null,
        ];

        // Here we check to see if we have properly found the closing parenthesis by
        // regex pattern or not, and will recursively continue on to the next ")"
        // then check again until the tokenizer confirms we find the right one.
        while (isset($match[4]) &&
                Str::endsWith($match[0], ')') &&
                ! $this->hasEvenNumberOfParentheses($match[0])) {
            if (($after = Str::after($template, $match[0])) === $template) {
                break;
            }

            $rest = Str::before($after, ')');

            if (isset($matches[0][$i + 1]) && Str::contains($rest.')', $matches[0][$i + 1])) {
                unset($matches[0][$i + 1]);
                $i++;
            }

            $match[0] = $match[0].$rest.')';
            $match[3] = $match[3].$rest.')';
            $match[4] = $match[4].$rest;
        }

        [$template, $offset] = $this->replaceFirstStatement(
            $match[0],
            $this->compileStatement($match),
            $template,
            $offset
        );
    }

    return $template;
}
```

PR #7972, opened 2024-02-19
Fix offline breaking runtime

joshhanley:
Currently if Livewire tries to make a request while offline, the request fails and it breaks Livewire's runtime, so that nothing functions when the network comes back online. See discussion #6957.

image

The reason is we are not catching any exceptions from the `fetch()` call in the event of a failure and its breaking everything else.

In V2, we handled it by catching exceptions and running a clean up:

https://github.com/livewire/livewire/blob/c3e0031652dd7387f0a6c8cd605a4d0d253a7145/js/connection/index.js#L94-L96

https://github.com/livewire/livewire/blob/c3e0031652dd7387f0a6c8cd605a4d0d253a7145/js/connection/index.js#L14-L18

So in this PR, I've wrapped the `await fetch()` inside `request/index.js` with a try/catch block and added in the methods from the error response handler below it.

I have left commented out the `fail()` call though, as that is a request failure hook, that developers can tap into. The question is, do we want to call the failure hook? If so, what should we pass through as the status and content (as there is no response object fetch fails due to network errors)?

I did a search of the codebase to see where else we are using `fetch()` and found it in the navigate code. So I've added a `.catch()` to the promise so it also doesn't fail, but I'm not sure if there is any clean up we should be doing there.

Is there any clean up for navigate?

In my testing the navigate loading bar remained on the screen for example.

Have created this as a draft for now until we can answer the above questions.

Also there is no tests as I don't know how to put chrome into offline mode programmatically (using the `wire:offline` events won't work unless we block network requests all together if offline is true?)


samlev:
Are there methods that can be called to start/stop/pause the event loop? I'm not sure if application state is preserved in the event of an uncaught runtime error...

What would be the danger in calling `Livewire.start()` or `Alpine.start()` again if the loop dies? I assume that state might be messed up, and anything that's entangled (for example) might lose tracking on current state.

samlev:
> Also there is no tests as I don't know how to put chrome into offline mode programmatically (using the `wire:offline` events won't work unless we block network requests all together if offline is true?)

So... I haven't tested it, but it's _theoretically_ possible (for chrome, at least). It's not well documented, but `php-webdriver` has the ability to send commands to chrome dev tools, so _in theory_ you can set the network conditions with something like this:

```php
$driver->getDevTools()
    ->execute('setNetworkConditions', [
        'offline' => true,
        'latency' => 5,
        'download_throughput' = 500,
        'upload_throughput' => 1024,
    ]):
```

PhiloNL:
Maybe a different approach, how about tracking the connection state and not firing the request when the user is offline:

```
if(navigator.onLine) {
   fetch()
} else {
  // trigger fail or maybe custom offline hook for a developer to use, e.g. alert('you are offline');
}
```

calebporzio:
Good point @PhiloNL - I think that'd be a lot simpler than adding try/catch with reverting all that state.

What do you think about that Josh?

If we DO decide to add try/catch, I'd say we don't have to call the "fail()" hook with a status code. But like Philo said, ideally the request doesn't get sent at all.

calebporzio:
Yeah, assuming `navigator.online` isn't reliable in the scenario, then we should:

For Livewire requests:
* wrap in try/catch
* in catch, call `fail` with `503` status code

We should hold off on solving this for wire:navigate and put that in a separate PR when there's demand for it. That's a whole different beast with a whole different "cleanup" story and different expectations.

joshhanley:
Ok great. Will fix this up.


Yep agreed re moving navigate to its own PR. I actually don't think the runtime breaks when navigate fails as it is using the fetch promise syntax. But loading just goes on forever haha.

edalzell:
> But loading just goes on forever haha

Just had a customer hit this, the page reloaded (like it should), but never completed. Will try to get console logs (it's on mobile) to confirm but we suspect this is the issue.

joshhanley:
@calebporzio I have removed the navigate changes and updated the fail call to pass in a 503. So this PR is now ready to go 😁

edalzell:
Woo hoo, thanks Josh!

calebporzio:
Perfect, thanks @joshhanley!

mtctonyhkhk2010:
It seems this problem still exist when using wire:navigate
image

joshhanley:
@mtctonyhkhk2010 Yep, see the above comments that we still need to address if for navigate in a separate PR 🙂

PR #7973, opened 2024-02-19
Add asset publishing

joshhanley:
This PR adds the option to publish assets.

One issue I had when writing this, was with the way it worked in V2, the asset URL didn't have `livewire.js` suffixed, but in V3 the `$url` does have `livewire.js` suffixed.

So in my implementation, I couldn't make use of the existing `$url` variable, so I created another one called `$assetUrl` that checks to see if there is the legacy config option and if not, just uses the public directory path.

One thing to note though is the legacy config option `livewire.asset_url` won't work if upgraded directly from V2 (if not publishing assets or if you are) due the suffix problem mentioned above. So it would need to be changed to include the suffix, if someone is making use of it. I'm happy to update how this works if it's not ideal (see `FrontendAssets` lines 152-156 and line 188).

I've added docs for publish assets too, with a warning that it is not needed, as people in V2 would publish thinking it was required, and frequently has assets out of date.

image

I also added a tip to the manual bundling docs to rebuild after a composer update. I'm not a fan of the double up with the warning below it though. I think the Mix/Vite warning should be moved further up the "manually bundling section".

image

### DIFF
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -193,7 +193,7 @@ class FrontendAssets extends Mechanism
                 $assetWarning = <<<HTML
                 <script {$nonce}>
                     console.warn('Livewire: The published Livewire assets are out of date\\n See: https://livewire.laravel.com/docs/installation#publishing-livewires-frontend-assets')
-                </script>
+                </script>\n
                 HTML;
             }
         }
@@ -207,8 +207,7 @@ class FrontendAssets extends Mechanism
         );
 
         return <<<HTML
-        {$assetWarning}
-        <script src="{$url}" {$nonce} {$progressBar} data-csrf="{$token}" data-update-uri="{$updateUri}" {$extraAttributes}></script>
+        {$assetWarning}<script src="{$url}" {$nonce} {$progressBar} data-csrf="{$token}" data-update-uri="{$updateUri}" {$extraAttributes}></script>
         HTML;
     }
 
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -33,7 +33,7 @@ class FrontendAssets extends Mechanism
 
         app('livewire')->provide(function() {
             $this->publishes(
-                    [
+                [
                     __DIR__.'/../../../dist' => public_path('vendor/livewire'),
                 ],
                 'livewire:assets',
@@ -175,7 +175,7 @@ class FrontendAssets extends Mechanism
         $assetWarning = null;
 
         $nonce = isset($options['nonce']) ? "nonce=\"{$options['nonce']}\"" : '';
-        
+
         // Use static assets if they have been published
         if (file_exists(public_path('vendor/livewire/manifest.json'))) {
             $publishedManifest = json_decode(file_get_contents(public_path('vendor/livewire/manifest.json')), true);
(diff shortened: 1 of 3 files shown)


src/Mechanisms/FrontendAssets/FrontendAssets.php
```diff
@@ -163,7 +172,31 @@ public static function js($options)
 
         $token = app()->has('session.store') ? csrf_token() : '';
 
+        $assetWarning = null;
+
         $nonce = isset($options['nonce']) ? "nonce=\"{$options['nonce']}\"" : '';
+        
+        // Use static assets if they have been published
```
calebporzio, line 179: Can this be extracted into a method or something so it doesn't add too much confusion to this method

calebporzio:
Hey @joshhanley - can you clarify the `asset_url` config question? Is there a way to NOT cause a breaking change? If so, what do those two paths look like?

Adding this feature could already introduce confusion and bugs because there are 3 other entries in installation dealing with assets (Using `@livewireScripts`, setting a custom URL, bundling them yourself).

I trust that enough people want this that it's valuable to add, but if it's not a clear addition without breaking consequences maybe we hold off? Idk. I'm just a little fuzzy on the question.

joshhanley:
@calebporzio yep, the breaking change is already in the codebase here:

https://github.com/livewire/livewire/blob/7e7d638183b34fb61621455891869f5abfd55a82/src/Mechanisms/FrontendAssets/FrontendAssets.php#L141-L162

So if we follow the logic from the top looking at what the `$url` will be set to on each line:
1. It will get the URI from the default endpoint `/livewire/livewire.js`
2. If `config('livewire.asset_url')` is set (especially from V2), it might be set to `/mytenant/livewire/`
3. If `$options['asset_url']` is set, it might be similar to the config above
4. If `$options['url']` is set, it might be similar to the config (as it's not documented as the new option
5. We can then see there is an `rtrim($url, '/');` again this code was from V2 as the config was typically the path (without `livewire.js` on the end)
6. We then do a check to set the prefix if needed
7. Finally we append `?id{$versionHash}` to the `$url`

So if the `livewire.asset_url` config is being used in V2 and you upgrade to V3, then the URL will look something like this `/mytenant/livewire?id=af3k9r3haa2`, notice the `/livewire.js` is missing.

Here is a screenshot from the V2 docs on what we recommended people set the asset url config to

image

How that effects the asset publishing is that I cannot guarantee what state the `$url` parameter is in due to the inconsistencies above (has filename suffix or not), so found it easier to build up a new url in `$assetUrl` which points to the published asset URI, then I overwrite `$url` with that URI so that it is used correctly in the `<script>` tag output.

But as I have made use of the `livewire.asset_url` config if set, which if set as per V2 (using a path without the filename suffix) then it will be break. This is because the way I have set it, it assuming someone has changed this to match the V3 style `/mytenant/livewire/livewire.js` as required in the `$url` overrides described above.

So as I see it, adding this feature in itself isn't a breaking change.

Hopefully that makes sense. I will try and refactor that section into a method to clean it up.

calebporzio:
Thanks for Clarifying Josh!

marvinosswald:
Just for posterity, this did cause a breaking change for me upgrading for 3.4.6 to 3.4.10. But do think this very useful 👍 

to fix it i had to add this to my livewire config:
```php
'asset_url' => env('ASSET_URL', null) . "/vendor/livewire/livewire.js?id=" . json_decode(file_get_contents(public_path('vendor/livewire/manifest.json')), true)['/livewire.js']
```

@joshhanley 
I might miss something but i think the id should be appended by your new function instead of in userland in a config file.

PR #7984, opened 2024-02-20
Fix morph marker edge cases

joshhanley:
There are a couple of edge cases that have come up with the new morph marker strategy, see discussion #7981.

One is the regex wasn't matching the closing tags correctly, as it was finding `@endfor` inside of an `@endforelse`. So I have made that regex more robust.

The other was if a conditional contains a `$` or `\`, then the `preg_replace` call was trying to use them as match indicators in the replacement string. So needed to ensure they had been correctly escaped.

### DIFF
--- a/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php
+++ b/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php
@@ -142,7 +142,7 @@ class SupportMorphAwareIfStatement extends ComponentHook
 
         $foundWithSuffix = $found.$suffix;
 
-        $pattern = "/{$foundEscaped}(?!{$suffixEscaped})(?![^<]*(?<![?=-])>)/mUi";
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks Josh!

wimski:
When can we expect a release for this? I've updated to `3.4.5` this morning to fix another issue only to find that something else broke :see_no_evil:

ariaieboy:
I have same problem with `foreach` livewire throw `syntax error, unexpected token "endfor"` on every `@foreach @endforeach` block. 
@joshhanley 
If this PR fix this problem too please consider taging this so we can test it.

ekandreas:
We had to downgrade to 3.4.4 as a workaround.
syntax error, unexpected token "endfor"

joshhanley:
@wimski @ariaieboy @ekandreas this has been tagged as v3.4.6. Please let me know if there are any further issues with it 🙂

ekandreas:
This is resolved in 3.4.6, thanks!

joshhanley:
@ekandreas great! Thanks for letting us know 😁

PR #7985, opened 2024-02-20
Add Fallback Value to Utils::matchesCache()

onairmarc:
Fixing a PHP Deprecation Warning for strtotime()

Yep! fix-php-strtotime-deprecation-warning

Nope. Only a single line was changed.

No additional test was added.

The following PHP Warning was occurring in PHP 8.3: `"strtotime(): Passing null to parameter #1 ($datetime) of type string is deprecated in /workspace/vendor/livewire/livewire/src/Drawer/Utils.php on line 103"`

To fix this, I changed line 101 of src/Drawer/Utils.php from:
`$ifModifiedSince = app(Request::class)->header('if-modified-since')`
to:
$ifModifiedSince = app(Request::class)->header('if-modified-since') ?? now()->toString();

This way, there will always be a value passed to `strtotime()` and because the value will differ from `$lastModified,` it will force the return to be false. This will also prevent a fatal PHP error in future versions of PHP.

### DIFF
--- a/src/Drawer/Utils.php
+++ b/src/Drawer/Utils.php
@@ -98,7 +98,7 @@ class Utils extends BaseUtils
 
     static function matchesCache($lastModified)
     {
-        $ifModifiedSince = app(Request::class)->header('if-modified-since');
+        $ifModifiedSince = app(Request::class)->header('if-modified-since') ?? now()->toString();
 
         return $ifModifiedSince !== null && @strtotime($ifModifiedSince) === $lastModified;
     }


PerryvanderMeer:
I assume this is already fixed in https://github.com/livewire/livewire/pull/7983

onairmarc:
Whelp, I completely missed that! Looking at that PR, it doesn't look like it's guarding against passing a null value to `strtotime()` though.

PerryvanderMeer:
It does so by checking if `$ifModifiedSince` is null before passing it to `strtotime`

PR #7991, opened 2024-02-22
Add form flag to `MakeCommand`

lee-to:
It's often a two-step process to create a component and a form, but with this small improvement it will be possible to generate the form along with the component

`php artisan livewire:make MyComponent --form`


lee-to:
Fixed missing option in parent class

calebporzio:
Thank you for this contribution. I think this makes sense, however, I think I'm going to close this as we will be rebuilding the command system for v4 and can add this then. (Because even merging this now wouldn't be complete because when you run the move command you might expect it to move the form as well, idk.) Just holding off for now. Thank you though!

PR #7992, opened 2024-02-22
Fix morph marker edge case with one-liner statements

austenc:
This is related to the fixes in https://github.com/livewire/livewire/pull/7984

Actively working on a fix for this and will update the PR as soon as I learn more.

Prior to those changes you could have extra text next to a nasty one-liner if statement, but now it throws an error in the console. 
```
// Broken
<button>@if ($this->total > 0)<span>{{ $this->total }}</span> @endif Filters</button>

// Works
<button>@if ($this->total > 0)<span>{{ $this->total }}</span> @endif 
Filters</button>
```

The key is that having extra words on the same line as the `@endif` breaks now, whereas it didn't before.
image

~~WIP ✌️~~

**Edit: This now includes a fix.**

The way the directive matching regex is set up, a trailing space gets matched even for closing directives, such as `@endif `. Opening directives may or may not include the space `@if(...)` or `@if (...)`. With closing directives, we don't need to match the space, so although we tried to find a more elegant regular expression solution, the `rtrim` seemed to be the path of least resistance here. Thank you!

### DIFF
--- a/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php
+++ b/src/Features/SupportMorphAwareIfStatement/SupportMorphAwareIfStatement.php
@@ -135,6 +135,10 @@ class SupportMorphAwareIfStatement extends ComponentHook
 
     protected static function suffixClosingDirective($found, $template)
     {
+        // Opening directives can contain a space before the parens, but that causes issues with closing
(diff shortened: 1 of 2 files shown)


joshhanley (approved):
I paired with @austenc on this, so it looks good to me 😁

calebporzio:
Sick, thanks you two!

PR #8002, opened 2024-02-23
Add trigger async to global Livewire object

PhiloNL:
I want to hook into the `triggerAsync` function but at this time only `trigger` is available on the global `Livewire` object. This PR adds the `triggerAsync` function to the Livewire object.


calebporzio:
Curious what you need it for?

kaptk2:
> Curious what you need it for?

Here's a discussion about it. 

https://twitter.com/Philo01/status/1760747447004090861?t=As9iQJWIq_xvDHFv2yyzSg&s=19

PhiloNL:
I've been working on making Livewire components embeddable. The challenge I was facing was that Livewire assumes assets are rendered within the layout and therefore doesn't load them in the scenario where I render the component outside a regular application setup where you have a layout.

So instead, the API endpoint I made returns the output of `@livewire()` but also the initial assets:

```php
    return [
       'html' => Blade::render('@livewire($component, $params)', [
           'component' => $component,
           'params' => $params,
       ]),
       'assets' => SupportScriptsAndAssets::getAssets(),
   ];
```

To ensure the assets are loaded, I trigger the `payload.intercept`, similarly to how it is done within Livewire's core. But it uses the async trigger instead of the regular trigger.

```js
await Livewire.triggerAsync('payload.intercept', {assets});
Livewire.start();
```

calebporzio:
Sounds cool! Merging

PhiloNL:
Thanks! :)

PR #8006, opened 2024-02-23
Fix form errors not triggering specific rule assertion for `validateOnly()`

PerryvanderMeer:
This PR is a follow-up on #7935, but this one fixes the use of `validateOnly()`. Also adds a test for realtime validation of forms.


calebporzio:
Thank you!

PR #8007, opened 2024-02-23
Failing test for specific rule assertions in real-time form validation

PerryvanderMeer:
This PR is a failing test for a situation with specific rule assertions for real-time form validation.


joshhanley:
@PerryvanderMeer can you move this failing test to PR #8006 as I believe the fix there will fix this too, as Rule/Validate uses `validateOnly` under the hood. Also can you change the `Rule` to `Validate` in the test as that is what we recommend using now?

PerryvanderMeer:
@joshhanley moved to #8006 and indeed is working.

joshhanley:
@PerryvanderMeer sweet! Thanks for that 😁

PR #8010, opened 2024-02-23
pagination: add text option to data array to show/hide pagination text

amoktar:
This pr adds the option for showing/hiding pagination text 

I've created a discussion about it #8008 

Sorry I couldn't do the test

https://github.com/AMoktar/livewire/blob/99ee51ad25462ada2c0f6e3eb99cdb519e076265/src/Features/SupportPagination/views/tailwind.blade.php#L12

```php
// ...

if (! isset($text)) {
    $text = true;
}

// ...

         <div @class([
                'hidden sm:flex-1 sm:flex sm:items-center',
                'sm:justify-between' => $text,
                'sm:justify-center' => !$text,
                ])>
                @if ($text)
                    <div>
                        <p class="text-sm text-gray-700 leading-5">
                            <span>{!! __('Showing') !!}</span>
                            <span class="font-medium">{{ $paginator->firstItem() }}</span>
``` 

I'm not sure this is the best solution, I hope that help anyway.

Thanks guys

### DIFF
--- a/src/Features/SupportPagination/views/tailwind.blade.php
+++ b/src/Features/SupportPagination/views/tailwind.blade.php
@@ -8,6 +8,10 @@ $scrollIntoViewJsSnippet = ($scrollTo !== false)
        (\$el.closest('{$scrollTo}') || document.querySelector('{$scrollTo}')).scrollIntoView()
     JS
     : '';
+
+if (! isset($text)) {
+    $text = true;
+}
 @endphp
 
 <div>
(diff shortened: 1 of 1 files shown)


PhiloNL:
Hi @AMoktar 
Thanks for your PR. As @joshhanley suggested, the best solution is to publish the views using the `php artisan livewire:publish --pagination` command. This will publish the views in your application `views/vendor` directory so you can change them to meet your needs.

calebporzio:
Exactly what @PhiloNL said!

PR #7043, opened 2023-10-10
Add `wire:confirm`

calebporzio:
This adds a directive called `wire:confirm` (see contained documentation for details)

PR #8039, opened 2024-02-27
Failing test for wire:confirm where inputs remain disabled

jakewhiteley:
This is a failing test case for `wire:confirm` and `wire:confirm.prompt` not working as expecetd when added to a form which already has `wire:submit`.

While the logic works correctly, after confirming the dialoge sucessfully, the inputs and submit in the form remain `readonly` or `disabled`.

This issue appears in https://github.com/livewire/livewire/discussions/7176, https://github.com/livewire/livewire/discussions/7138.

This PR adds a test in the legacy_tests folder (seemed the best place for this as it is 2.* functionality). It fails on the above behaviour, but also adds tests for `wire:confirm` and `wire:confirm.prompt` in gemeral.


joshhanley:
@jakewhiteley thanks for the PR!

The `wire:confirm` feature was only recently added to V3 (PR #7043). There is already a `BrowserTest.php` file for the feature here https://github.com/livewire/livewire/blob/main/src/Features/SupportWireConfirm/BrowserTest.php

If you get a chance, can you refactor it to there, and use the new `Livewire::visit(new class extends Component {...})` syntax?

In general though `legacy_tests` folder isn't V2 features, it's just tests that we haven't moved into the correct directory yet (our goal it to remove the `legacy_tests` folder eventually 😁).

Thanks!

jakewhiteley:
@joshhanley Sure, I'll do this soon.

Out of interest, why are the newer tests not in a folder clearly marked tests? I never would have thought to look in the `src/Features` folder!

PhiloNL:
@jakewhiteley https://twitter.com/calebporzio/status/1654163650561687553 :)

edalzell:
> @jakewhiteley [twitter.com/calebporzio/status/1654163650561687553](https://twitter.com/calebporzio/status/1654163650561687553) :)

lots of us aren't on twitter anymore, stuff like that should be in the docs (yes I know I should PR it).

PhiloNL:
> > @jakewhiteley [twitter.com/calebporzio/status/1654163650561687553](https://twitter.com/calebporzio/status/1654163650561687553) :)
> 
> lots of us aren't on X anymore, stuff like that should be in the docs (yes I know I should PR it).

It's in the docs as well https://livewire.laravel.com/docs/contribution-guide#add-failing-tests

calebporzio:
Here's what's happening:

## Setup

Given the following snippet:

```html
<form wire:submit="save" wire:confirm="Are you sure?">
</form>
```

When the `<button>` is pressed the following events SHOULD happen:
* A confirmation dialog pops up in the browser asking "Are you sure?"
* You hit "cancel"
* The dialog disappears and everything goes back to the way it was.

## The Problem

However, currently, if you press "cancel", the form's submit button is permanently disabled.

### Problem explanation

The reason is:

There is a system in Livewire that looks for `wire:submit` elements and adds it's own even listener to disable the form while the network request is out to the server.

Currently this system isn't aware of `wire:confirm` at all.

Here's a deeper look at what's happening for this error to occur:

* Livewire looks for elements with `wire:submit` on them and adds a "submit" event listener to disable the form
* That system also listens for network requests to be complete to cleanup any disabled forms and set them back
* The `wire:submit/wire:confirm` button is pressed
* Livewire's event listener picks up the button press and disables the form
* `wire:confirm` intercepts the "submit" event and prompts the user for confirmation
* The user presses "cancel"
* The network request is prevented from occurring
* Because no network request is sent, the form will never be reset

## Solutions

### Solution A) Stop propagation if "Cancel" is pressed

One solution would be to ensure that the `wire:submit` listener is handled BEFORE the form disabling one. Then, from `wire:confirm` we can call `event.stopPropagation()` on the event if "cancel" is pressed.

Pro: systematic solution. could prevent other problems. minimal code changes
Con: relies on the `wire:submit` event listener being registered BEFORE the form disabling one (which may be difficult if it's not naturally that way)

### Solution B) Manually calling form cleanup code if "cancel" is pressed

This is less ideal because it mixes concerns, however, we can detect if "cancel" is pressed, and manually call the code that cleans up any outstanding disabled forms.

Pro: direct. reliable.
Con: mixes concerns. (couples two otherwise unrelated codebase files/systems)

### Solution C) Don't listen for a the "submit" event in the first place

Currently the form-disabling system listens for the "submit" event from the browser.

Instead, we could change this to listen for a Livewire update that was fired from this submit event (not trivial). This way it will only ever set forms as disabled while a Livewire network request is out, preventing this problem alltogether.

Pro: direct. reliable. potentially solves other future problems.
Con: may be difficult to impossible to link a "livewire update" with a given `wire:submit` element

---

I think my favorite is "C" so I will explore that solution. If it turns out to add complexity that's not worth it, I'll explore "A", then eventually "B" if we must.

Thanks for the report. I'll report back my findings.

calebporzio:
Ok, I went with solution "A" because "C" was too complicated and maybe not even possible to do reliably.

PR #8042, opened 2024-02-28
Add improved error logging for unclosed elements

PhiloNL:
People are experiencing the following errors (https://github.com/livewire/livewire/discussions/6993#discussioncomment-8604973) when triggering updates and components contain unclosed elements or when trailing `</div>` are present. This PR attempts to provide improved error logging to make users aware that the error is likely caused by unclosed elements.

```
TypeError: Cannot read properties of undefined (reading 'uri')
TypeError: Cannot read properties of undefined (reading 'csrf')
```

The existing `throw 'Livewire: No CSRF token detected'` is not shown because the error caused the remaining code not to be executed. The new output looks like this:

image


calebporzio:
Hmmmm I have two hesitancies about merging this:

* console warnings and error strings can't get minified so they disproportionately increase the size of the js payload
* this solution seems a little too "downstream" from the actual problem

Is there a more "upstream" solution to this problem?

Like why is not including a closing div causing the scripts to not be injected? or maybe they are, but chrome isn't rendering them because of no closing tag?

I agree this is a weird feeling that not including that closing tag would throw everything off, but that kind of thing often happens when not including closing tags right? chrome will do all sorts of weird things even without Livewire?

So yeah, not in love with this PR, hoping there's something easier we can do, otherwise, I might just leave this alone.

Thanks for caring though and helping people

PhiloNL:
Hey Caleb,

1. I've checked the size difference and it's 0.1kb more according to the built output.
2. The problem occurs on consecutive update requests, in the following example, it renders without errors, clicking the button one time works (but it will also remove the script tag), the second time it will throw an error:

```blade
<div>
    <button wire:click="$set('foo', true)">True</button>

    <div>
    @if($foo)
        <h2>Hi</h2>
            <div>
                <div>
                    <h4></h4>
                </div>
    @endif

</div>
```

So it must be related to the morphing; no idea where to start looking to fix this though. 

The error could be a temporary improvement until it has been solved upstream but to improve the overall experience for people starting with Livewire, I've seen a couple discussions where people now spend hours because the error doesn't help them in the right direction (a few examples https://github.com/livewire/livewire/discussions/5995, https://github.com/livewire/livewire/discussions/6663, https://github.com/livewire/livewire/discussions/7920, https://github.com/livewire/livewire/discussions/6603).

levu42:
It should have a good effect on the minimized JS if both error messages you're introducing, @PhiloNL, could be the same, e.g. `'Error. Check for unclosed/trailing elements.'`. To me as a user of livewire, this gives enough hints. I don't care whether it's about the CSRF token or the update URI, that's livewire internals.

Thanks for trying to keep the build file as small as possible, @calebporzio!

calebporzio:
Ok, here's what's happening:

# The Scenario

Given a LW component with the following Blade:

```html
<div>
    <button wire:click="$set('foo', true)">True</button>

    <div>
    @if($foo)
        <h2>Hi</h2>
            <div>
                <div>
                    <h4></h4>
                </div>
    @endif

</div>
```

The following HTML is sent to the browser for the initial page load:

```html
<html>
<body>
    <div wire:snapshot="..." wire:effects="..." wire:id="...">
        <button wire:click="$set('foo', true)">True</button>

        <div>
        <!--[if BLOCK]><![endif]--><!--[if ENDBLOCK]><![endif]-->

    </div>

    <!-- Livewire Scripts -->
    <script src="..." data-csrf="..." data-update-uri="..."></script>
</body>
</html>
```

> Notice the missing closing `</div>` inside the component.

# The Problem

## Problem A: Chrome tries to fix the missing `<div>`

If you load the above HTML in Chrome and inspect the rendered HTML in your devtools, you'll notice Chrome filled in the missing closing DIV, but it did it WRONG.

Here's what would show in the Chrome devtools:

```html
<html>
<body>
    <div wire:snapshot="..." wire:effects="..." wire:id="...">
        <button wire:click="$set('foo', true)">True</button>

        <div>
            <!--[if BLOCK]><![endif]--><!--[if ENDBLOCK]><![endif]-->
        </div> 

        <!-- Livewire Scripts -->
        <script src="..." data-csrf="..." data-update-uri="..."></script>
    </div>
</body>
</html>
```

As you can see, Livewire's `<script>` tag is now rendered INSIDE the Livewire component and NOT at the end of the document.

Also note `data-csrf` and `data-update-uri` are two attributes that Livewire assumes will always be on the page and are used to know the CSRF token to send to the server between network requests.

## Problem B: Livewire morph will remove `<script>`

Now that the Livewire's `<script>` tag is INSIDE the Livewire component, when a new Livewire request happens, Livewire will receive HTML from the server and "morph" the existing HTML to match it. In doing so, it will remove the `<script>` tag because it wasn't in the HTML returned from the server.

Without the `<script>` tag removed from the page, Livewire will try to look for the `data-csrf` token on the next request and throw the aforementioned error because it can't find it:

```
TypeError: Cannot read properties of undefined (reading 'uri')
TypeError: Cannot read properties of undefined (reading 'csrf')
```

# Solutions

Quickly, here's a reminder of the systems that are failing to cause this error so we can think through solution paths:
* Human error: the closing div is missing in the first place
* Chrome error: chrome is adding the missing `</div>` in the wrong place, bumping up the `<script>` element
* Livewire error: Livewire is removing the `<script>` when morphing
* Livewire error: Livewire is depending on the existence of `<script>` for knowledge like CSRF tokens.

### Solution A: Throw server-side error

We could in theory parse the HTML of each component, identify missing closing tags and thrown an error.

Con: Way too flaky and slow to ever attempt (at least until we have our own template parser with an AST)

### Solution B: Look for Livewire `<script>` tag inside Livewire component on initialization

When a Livewire component is initialized, it could do a quick tree-walk and look for any stray Livewire script tags.

Con: slow and weird

### Solution C: Find Livewire `<script>` tag and make sure it's at the root level

When Livewire initializes, we could do a `document.querySelector()` to find the script tag and make sure it's at the root level (by making sure it's parent element is the `<body>`, `<head>` or `<html>` element).

Pro: fast, simple
Con: potentially aggressive/flaky: there might be cases where users intend to nest the script tag or something.

### Solution D: Move `<script>` at morph time

We could tweak the morph algorithm to leave Livewire script tags alone or move them to the bottom of the page.

Con: complexity that is too niche for the morph algo

---

The above are different ways we can identify a misplaced `<script>`, but there's still the question of what to do when one is identified:

A) Throw an error/warning
B) Fix in-place (move `<script>` to the proper spot)

From a users perspective it would be the nicest to have it fixed automatically, however, it happens because of missing closing elements which will cause other problems.

I think throwing a console warning should be sufficient.

# The plan

After laying this all out, I think the best thing to do is add a short bit of logic to the Livewire JS to ensure the `<script>` element isn't misplaced. If it is, then we offer a simple console warning and do nothing else.

calebporzio:
I made the proposed change. I'll leave this open until I hear back on your thoughts @PhiloNL. Thanks!

PhiloNL:
Thanks for explaining what is happening under the hood, much appreciated 🙌🏻 That looks like a great solution to help point people in the right direction.

joshhanley:
@calebporzio I wonder whether it would be better to move to a multiple root element solution like we had in V2, that puts a marker at the end of the component. This might allow us to detected multiple root elements but possibly also deformed components (ie missing a closing div or something). The we can display one warning covering both cases?

calebporzio:
@joshhanley, right. that's possible. I don't see off hand right now how that could be used to detect this scenario, although I don't doubt it could. I'm going to go with this right now because it solves the immediate problem at hand well. good thoughts.

PR #8047, opened 2024-02-29
Fix TypeError / return null for deleted public legacy model

ghbob:
Discussion: https://github.com/livewire/livewire/discussions/7987

The failing testcase explains the problem quite well: if a public property legacy model is deleted and the component is refreshed a type error occurs:

`TypeError: Livewire\Features\SupportLegacyModels\EloquentModelSynth::setDataOnModel(): Argument #1 ($model) must be of type Illuminate\Database\Eloquent\Model, null given, called in .../src/Features/SupportLegacyModels/EloquentModelSynth.php on line 90`

My suggested fix in the second commit is to return it early as `null` if `loadModel` returned `null` beforehand to avoid the TypeError.

Thanks!

### DIFF
--- a/src/Features/SupportLegacyModels/EloquentModelSynth.php
+++ b/src/Features/SupportLegacyModels/EloquentModelSynth.php
@@ -70,6 +70,10 @@ class EloquentModelSynth extends Synth
             $model = $this->loadModel($meta);
         }
 
+        if($model === null) {
+            return null;
+        }
+
         if (isset($meta['relations'])) {
             foreach($meta['relations'] as $relationKey) {
                 if (! isset($data[$relationKey])) continue;
--- a/src/Features/SupportLegacyModels/Tests/PublicPropertyHydrationAndDehydrationUnitTest.php
+++ b/src/Features/SupportLegacyModels/Tests/PublicPropertyHydrationAndDehydrationUnitTest.php
@@ -897,6 +897,20 @@ class PublicPropertyHydrationAndDehydrationUnitTest extends \Tests\TestCase
         Livewire::test(ModelsComponent::class, ['model' => $model])
             ->call('$refresh');
     }
+
+    /** @test */
+    public function it_returns_null_if_the_model_was_deleted() {
+        $this->expectNotToPerformAssertions();
+
+        $model = Author::create(['id' => 1, 'title' => 'foo', 'name' => 'bar', 'email' => 'baz']);
+
+        $component = Livewire::test(ModelsComponent::class, ['model' => $model])
+            ->call('$refresh');
+
+        $model->delete();
+
+        $component->call('$refresh');
+    }
 }
 
 class PostComponent extends Component


calebporzio:
Thank you for this @ghbob, however I want to make sure `null` is the right thing to do here.

I suppose the options here are:
A) Throw a "Model has been deleted" exception or something
B) Hydrate an un-persisted version of the model with whatever data we have
C) Hydrate to `null` (or "unset" the property)
D) Leave things as-is

My first question is: what happens in V2?
(I'm assuming the same as V3, an error is thrown)

My second question is: what happens with normal models (not using the legacy binding synth)?
(I'm also assuming an error is thrown because it uses `->firstOrFail()`)

I'd love both of those assumptions tested if you have time.

If they are correct, then it would be odd to tweak the legacy synth and not the normal synth if we DO decide to do that.

However, this seems like one of those things that could easily go either way (leave as-is or make a change) and I'm inclined to leave it as-is in that case because I don't want to break people's existing functionality or expectations without thinking through all the reasoning behind this change.

So I'll leave this open for now and give you/others a chance to weigh in before making any decisions.

Thank you.

joshhanley:
@calebporzio I think the issue describe here more typically is run into when you use the model in multiple components, whenever one component deletes the model, if the other then refreshes, an error is thrown due to Livewire trying to rehydrate the deleted model and there is no path forward for an end user other than to refresh the page to clear it out.  I can also imaging this scenario being triggered by another user deleting a model that a user is currently working on, so when a component refreshes, the error is triggered. So to me that means the current behaviour (D) or option A aren't ideal.

I don't think B is a good idea there is no data other than model ID and class as we don't send any model attributes to the front end.

Option C presents the flip side to options A/D. For an end user then it is all good, they don't get blocked. But then it makes it harder for a developer to debug what is going on. Could potentially lead to other issues if a component has been written to depend on that model being set in a property, so if it becomes `null` may even lead to security concerns.

Regardless of what we decided both legacy and normal models support should both handle it the same way so it is consistent.

I think I would lean towards at least handling it better - option A - with a `ModelNotFound` exception being thrown instead of the `setDataOnModel()` error which doesn't really explain what is happening.

It's a tough problem as either eventuality is not great and one of the reasons why I'm still not sold on having models support 😆

levu42:
@joshhanley The scenarios you describe are pretty much the ones I experienced so far. I agree with your solution with one exception: When the model variable has a nullable type-hint, I would expect it to be set to `null` instead of an error being thrown. When there is a non-nullable typehint or no typehint at all, what you say makes a lot of sense.

I develop my livewire codebases with very strict typing and using PHPStan all the time, so when I set a variable to be nullable, the whole codebase will handle the `null` case and it's an expected case for me. In that case, it would be really annoying to suddenly have to deal with exceptions that are thrown because another component deleted the model, instead of it just being set to `null`.

joshhanley:
@levu42 yep I agree with that, if there is an explicit null type, then setting to null should be fine to me 🙂

calebporzio:
Alright, I'm still unsure about this because the following:

```php
$this->user = User::find(1);

$this->user->delete();

// $this->user is still a variable and won't be null in a normal environment
// Why should it be in a Livewire environment?
```

I understand it's a bit of a papercut when using Livewire and this fixes that pain point, however, I'm just not convinced it's expected behavior for a model reference to become null once it has been deleted.

I'm going to close this, but very hesitantly. Please re-open it with a new implementation that only sets it to null of it's marked with the nullable type and we can discuss further. But I want to make sure there's sufficient demand and everyone agrees it's not a breaking change or surprising behavior.

Thank you for all of your time.

Spholt:
A bit late to the party on this one but I still think this needs a resolution one way of the other.

The scenario we've been encountering occurs when we have a component that contains a snapshot of an eloquent model and provides the user with the ability to permanently delete said model. The component is mounted in either multiple windows by a single user, or multiple users have it mounted in their individual browsers. 

When one of components permanently deletes the model, all remaining instances of the component will now fail when they attempt to rehydrate. None of the available lifecycle hooks appear to be of use to catch the exception and the only recourse for the user is to refresh the page.

I like @joshhanley 's idea whereby a `ModelNotFound` exception is thrown, especially if this could then be handled with the `exception` lifecycle hook.

PR #8048, opened 2024-03-02
Fix unsetting legacy computed properties

joshhanley:
I'm in the process of upgrading a large V2 app and have refactored all of the `$this->forgetComputed()` to `unset()`.

Except this app still has all legacy computed properties `getFooProperty()` and I have found that `unset()` does nothing for legacy computed properties.

This PR adds support for unsetting legacy computed properties.


calebporzio:
Thanks Josh!

PR #8051, opened 2024-03-02
Fix form property has two same error messages

dairidong:
Use a form property in a Livewire Component, and call `$this->validate()` in the component, will see two same errors.

Example:
image

PR #8052, opened 2024-03-02
Fix validating forms via `$this->validate()`

PerryvanderMeer:
When validating a component with forms via `$this->validate()` instead of directly calling `$this->form->validate()`, it duplicates the error messages. See #8051

### DIFF
--- a/src/Features/SupportFormObjects/UnitTest.php
+++ b/src/Features/SupportFormObjects/UnitTest.php
@@ -81,6 +81,29 @@ class UnitTest extends \Tests\TestCase
     }
 
     /** @test */
+    function can_validate_a_form_with_the_general_validate_function()
+    {
+        Livewire::test(new class extends Component {
+            public PostFormValidateStub $form;
+
+            function save()
+            {
+                $this->validate();
+            }
+
+            public function render() {
+                return '<div></div>';
+            }
+        })
+            ->call('save')
+            ->tap(function ($component) {
+                $this->assertCount(1, $component->errors()->get('form.title'));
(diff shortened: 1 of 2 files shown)


joshhanley:
Had a quick look at this one and the underlying cause for this issue is that we get exceptions for each of the form objects and store them in the `$cumulativeErrors` bag.

```php
// First, run sub-validators...
foreach ($this->getFormObjects() as $form) {
    try {
        // Only run sub-validator if the sub-validator has rules...
        if (filled($form->getRules())) {
            $cumulativeData = array_merge($cumulativeData, $validateForm($form));
        }
    } catch (ValidationException $e) {
        $cumulativeErrors->merge($e->validator->errors());

        $formExceptions[] = $e;
    }
}
```

And then later if the main component validation doesn't fail, we get the first form object exception and use it's validator instance (which already has all the errors for that particular form object) and merge in the `$cumulativeErrors` bag (which contains the errors from the form already), hence why we get a double up of rules (but only for the first form with errors).

```php 
// If main validation passed, go through other sub-validation exceptions
// and throw the first one with the cumulative messages...
foreach ($formExceptions as $e) {
    $e->validator->errors()->merge($cumulativeErrors);

    throw $e;
}
```

So the fix seems ok to me. I do wonder whether there might be any side effects to using unique or not.

A longer solution would be to store the `$cumulativeErrors` and `$formExceptions` both keyed by the form property name and only merge in form errors not already present.

But I think we should just run with the solution provided here as it's much simpler.

calebporzio:
Thank you for all the helpful context @joshhanley - I'll go with this for now, sounds good. Thanks @PerryvanderMeer!

PR #8061, opened 2024-03-04
Add failing test for modifying values in the updating hook

thijsvdanker:
In Livewire V2 it was possible to mutate values in the `updatingFoo(&$value)` lifecycle hook if you accepted the value by reference.
For example, if a `$foo` property was set to `someValue`:

```php
// updatingFoo receives 'someValue'
function updatingFoo(&$value)
{
   $value = 'bar';
}
```

Would set the $value to `bar` and any subsequent hooks would receive `bar`.
The result would be that `$this->foo` was set to `bar`.

This does not work anymore in Livewire V3.
`$value` will remain `someValue` throughout the rest of the lifecycle.

This was also mentioned in this discussion: https://github.com/livewire/livewire/discussions/7870

### DIFF
--- a/src/Features/SupportLifecycleHooks/UnitTest.php
+++ b/src/Features/SupportLifecycleHooks/UnitTest.php
@@ -125,6 +125,7 @@ class UnitTest extends \Tests\TestCase
             'updatingBarBaz' => false,
             'updatedBar' => false,
             'updatedBarBaz' => false,
+            'updatingReference' => false,
         ], $component->lifecycles);
     }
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for submitting!

Unfortunately, I don't see an easy way to support his. This wasn't "official" functionality in V2, it just happened to work, so I don't feel compelled to support it in V3.

I'm open to someone submitting a PR that enables this if the change is simple enough, but if it requires a rework of that part of the system I will likely not merge.

Thanks for pitching in, sorry I can't invest my own time into this.

PR #8065, opened 2024-03-04
Add wire:target.except modifier functionality

patriktoth67:
Discussions: #3248 and #6661

Sorry in advance, the compiled assets seem to have changed some random things, but the part needed is in the js file anyways.

This PR adds the option to add the modifier .except to wire:target, making the functionality inverted to regular wire:target.

Inverted targeting with `wire:loading` allows you to display an element when a specified action is not being executed. This is particularly useful when there are more targets that should display a loading indicator than the ones which should not.

```blade
<button wire:click="process">Process</button>
<button wire:click="reset">Reset</button>

<div wire:loading wire:target.except="process">
    Waiting to process...
</div>
```

In the code above, the message "Waiting to process..." will be displayed whenever any Livewire action is triggered except for the process action. When the process action is invoked, the message will disappear, indicating that the processing action is underway.


patriktoth67:
Can anyone help me about why my compiled assets changed so many little things? I'm pretty sure that's why my tests failed, but it changed things I didn't touch. I'd like to figure it out not to run into it again if I make another PR.

calebporzio:
I haven't dug into your assets issue, but don't worry about it for now, I just rebuilt your assets so that we can see the test results.

About the PR: I think this makes total sense and will be very useful.

However, I don't love the code structure. (creating a `flags` object and passing it around). Not sure if there's any way to isolate this feature's code better to one spot, otherwise we might have to just eat the complexity.

Either way, I agree this is needed and am thankful for the PR!

calebporzio:
Oh, also we will need tests for this before merging.

It should be fairly straightforward to add them in this file: https://github.com/livewire/livewire/blob/main/src/Features/SupportWireLoading/BrowserTest.php

patriktoth67:
I re-structured it a bit, and added the test for it. Let me know if you think it needs some more attention, I'm happy to do it:)

Needs a rebuild on your end again.

joshhanley:
@patriktoth67 the assets are probably out of sync as a lot of the time there are changes on the Alpine main branch that Livewire depends upon, so you need to link Alpine and the required packages using NPM as per our contribution guide https://livewire.laravel.com/docs/contribution-guide#optional-set-up-alpine-for-local-development

patriktoth67:
Thank you @joshhanley, issue solved!

calebporzio:
Thanks for the updates @patriktoth67!

Two questions:
A) Does this allow for multiple exceptions? Like `wire:target.except="export, search"` would accept all requests EXCEPT `export` and `search`?
B) Does this work with file uploads? (I don't see code added to the file upload part of the loading js file) For example can someone do: `wire:target.except="photo"` and how the loading indicator for every request EXCEPT the photo file upload?

patriktoth67:
A) Yes it does, I'll update the browsertest to show it.
B) I didn't check it before, I'll do that and let you know!

patriktoth67:
Now it works with file uploads, and I added tests to test it with those as well.

patriktoth67:
Actually I'm having some troubles with the tests, it might be that there is an issue with the code so I'll look into it tomorrow

patriktoth67:
Okay, so I believe this shouldn't work with file uploads at all. The reason is, that if you use `wire:target.except="photo"` you expect the loading message not to appear when you upload a photo. However, when you upload the photo, it calls the `_startUpload` and `_finishUpload` (or `_uploadErrored`) function, which are not the `$photo` variable, so during the time those functions run, the loading message appears.

I'm not familiar with how file uploading works, maybe it would be as easy as to add these 3 functions to the `eventMismatch()` functions targets. Either way, I'll commit the working code for the feature for everything except file uploads, a working test for everything except file uploads, and I'll comment out a test for the feature with file upload for later.

calebporzio:
Fantastic. Thank you for being so active and responsive to feedback on this PR!

patriktoth67:
I have 2 questions. 

If  I want to continue working on this and maybe solve the file upload stuff, should I stay in this branch and continue even though you already merged it with main?

And if I want to work on some other feature before this one gets into a release, but that feature would touch the same files as this, how should I approach that? Creating a new branch from main won't contain the changes of this branch, and I'm not sure if that is an issue.

joshhanley:
@patriktoth67 I would make sure to pull the latest changes from Livewire's main branch into your fork of Livewire's main branch. And then make a new branch for the new changes/feature.

Re the second question, pulling the latest main branch changes as I mentioned above will make sure you always have the latest code, so any new development can use the features that have already been merged.

Hope that makes sense. If not, or if you have any other questions, let me know 🙂

patriktoth67:
@joshhanley Right now if I go back to the main branch of my livewire fork, then pull, I get I think the v3.4.8 without this PR's features (which have been added to v3.4.9 already). I open for example the `wire-loading.js` file and I see none of my code.

 So right now if I made a new branch from main, it wouldn't contain this wire:target.except functionality.
 
 How can I pull the latest commit? (for example yesterdays security patch #8117 isn't either on my main fork).

joshhanley:
@patriktoth67 ah yeah, so the issue will be that your local main branch is tracking your fork's main branch instead of Livewire's one.

There are two ways you can fix this:

- One is to browse to your fork on GitHub and "sync" the changes (see screenshot below)
- The other is to set up a second remote server (I call it "upstream") in your local git repo that point's to Livewire's repo, and then change the tracking on your local main branch to pull changes from upstream instead of origin (see bottom screenshots).

I use the second option personally.

### Option 1
_Use GitHub_
image

---

### Option 2

_Set up a different remote_
image

_Point new remote to Livewire repo_
image

_Change main branch tracking to track the new remote `main` branch_
image

patriktoth67:
@joshhanley Thank you for the detailed help!:)

joshhanley:
@patriktoth67 no worries! We appreciate the help, so happy to asssit where I can 🙂

PR #8068, opened 2024-03-05
Fix morph markers when using `view:cache`

27pchrisl:
The code that generates morph markers does not run when using `php artisan view:cache`. Since this is popular part of Laravel deployment pipelines it's worth mentioning in the documentation that you should remove this.

Discussion that covers this is here: https://github.com/livewire/livewire/discussions/7526


27pchrisl:
As part of solution A Livewire could insert itself into the `view:cache` process, and throw an exception that warns the developer that they've called a command incompatible with Livewire.

woodspire:
I think solution C is the best "proposed" solution. Simply add a way to define livewire view paths in the config file (instead of hardcoding views/livewire)

A better solution would be D: override the view:cache process and for each view, check if it's being used by livewire. If so, add the markers. If not, do the usual laravel process.

levu42:
Is it possible to generate the cache files in a way that they include something like

```
if ($isInLivewireContext) echo '<!--[if BLOCK]><![endif]-->';
```
?

Then the same cached view file can be both used in a non-livewire-context, without printing morph markers, and in a livewire context where it does print the morph markers.

joshhanley:
@levu42 I like that idea! I certainly think that would be possible 😁

danharrin:
If @levu42's idea is implemented, it would be useful to add a hook in so you can turn morph markers off per-Livewire-component. Maybe if you add a specific attribute to the Component class they aren't used?

27pchrisl:
Our of interest @danharrin, why wouldn't you want morph markers everywhere in a Livewire component?

levu42:
@PhiloNL the `Docs` label isn't really applicable here, is it? It's more about a bug fix, after the turn the discussion took from @calebporzio's comment on. Also, the title should probably be something like `fix morph markers with view caching`.

PhiloNL:
> @PhiloNL the `Docs` label isn't really applicable here, is it? It's more about a bug fix, after the turn the discussion took from @calebporzio's comment on. Also, the title should probably be something like `fix morph markers with view caching`.

Updated and converted PR to draft 👍

joshhanley:
I have been looking into this issue and below are my findings.

## Problem statement
There is currently no console implementation of morph markers injection so running `php artisan view:cache` at the moment renders all Livewire's views (and any blade components has may be used inside a Livewire component) without morph markers.

Ideally when a user runs `php artisan view:cache` then any views used inside Livewire should have the morph markers injected, to prevent any DOM diffing issues, but any blade views not used inside a Livewire component should not have them.

## Current browser injection process
To get the morph markers to work when a view is loaded in the browser, we have a `livewireOnlyPrecompiler()` method, inside `Mechanisms\ExtendBlade`, which registeres the morph markers blade precompiler and it only injects it into the blade compiler if a Livewire component has been rendered. It does this by listening to the `render` internal hook and if that is fired, then the `livewireifyBladeCompiler()` method is called, which actually injects the precompilers. At the end of render, Livewire also undos the precompiler registration to ensure that no other blade views are impacted.
## Possible solutions
### 1) Render all views with morph markers
**Pros**
- Easy to implement, instead of have a `livewireOnlyPrecompiler()` we would just register the morph markers precompiler with blade all the time

**Cons**
- Heavy handed, all blade views rendered will have morph markers (which get's sent to the browser) regardless of if it is a Livewire view or not.

### 2) Compile all views with a conditional to check if in a component, if so then echo the morph markers
This is the solution suggested above by @levu42

**Pros**
- We can compile all views with this check, and the marker won't actually be rendered if not in a Livewire component

**Cons**
- Potentially has a performance impact as every conditional in all blade views would have this check, but it's better then rendering the morph markers in all views.

### 3) Conditionally compile the morph marker conditional suggested in option 2
What this means is if the blade compiler runs in a console, we always add the conditional check. But if it's being compiled as part of a browser request, we could only add the morph marker conditional if the compile/render is inside a Livewire component

**Pros**
- Has some optimisation for non-livewire views when running in the browser

**Cons**
- Introduces inconsistency between how a rendered non-livewire view works locally vs when it is in production - where `view:cache` is likely to be run - which might lead to bugs. 

## Recommendation
I think option 2 would be better over option 1 as sending the morph markers to the browser with every view regardless of if it is Livewire or not isn't a great move (plus people might not like it). Where as adding a conditional to a compile view won't have the same impact. Plus PHP is good at handling conditionals.

Option 3 introduces inconsistencies which I don't think are good especially when the difference is in production.

## Implementation

I've experimented with option 2 and found that the implementation change was pretty simple. I have pushed those changes to this PR.

### Benchmarks
I have run some benchmarks to see what the potential impact is. The following tests were run on a large app with about 450 files in `resources/views`. For the render tests, one of the biggest pages in the app were chosen.

The following tests were performed with the latest version of Livewire and then again with these new changes:
- Running `php artisan view:cache`
- Rendering a large page after `php artisan view:cache` has been run
- Rendering a large page after clearing the view cache

**Running `view:cache`**

| #       | old (secs) | new (secs) |
| ------- | ---------- | ---------- |
| 1       | 8.523      | 12.772     |
| 2       | 8.232      | 12.394     |
| 3       | 8.229      | 12.682     |
| 4       | 8.878      | 12.309     |
| 5       | 8.424      | 12.430     |
| **Avg** | **8.457**  | **12.517** |

**Render with `view:cache`**

| #       | old (secs) | new (secs) |
| ------- | ---------- | ---------- |
| 1       | 1.319      | 1.321      |
| 2       | 1.305      | 1.290      |
| 3       | 1.279      | 1.302      |
| 4       | 1.301      | 1.317      |
| 5       | 1.308      | 1.288      |
| **Avg** | **1.302**  | **1.304**  |

**Render without  `view:cache`**

| #       | old (secs) | new (secs) |
| ------- | ---------- | ---------- |
| 1       | 3.130      | 3.366      |
| 2       | 3.199      | 3.438      |
| 3       | 3.163      | 3.200      |
| 4       | 3.210      | 3.144      |
| 5       | 3.163      | 3.378      |
| **Avg** | **3.173**  | **3.305**  |

### Conclusion from benchmarks

Running `php artisan view:cache` has had a significant impact taking nearly 50% longer to run but as this is typically run on deploy and not part of a users request, this should be ok.

Looking at the page load times, if `view:cache` has been run, then there is a negligible different when rendering the view with the extra conditionals in it.

If `view:cache` hasn't been run, then there is a slight increase in the rendering time but otherwise is basically the same.

## Next steps

If @calebporzio is happy with this implemention, then the following needs to happen before this PR can be merged:
- [ ] Remove docs warning
- [ ] Add tests

AurimasBal:
Good day. 
First of all thank you for creating/working on Livewire.
I would like to know the progress of this PR and is there some kind of timeline when this will be merged?

We are running into this issue currently, since `artisan view:cache` is a part of our deployment pipeline. 
We have quite complex  UI (with lots of nested components), and while everything works fine when testing on local environments (since `view:cache` is not used), everything starts to break a part on QA environments (because of missing morph markers) and we have to constantly find workarounds for these issues.

Thanks again.

src/LivewireManager.php
```diff
@@ -52,6 +52,11 @@ function precompiler($callback)
         app(ExtendBlade::class)->livewireOnlyPrecompiler($callback);
     }
 
+    function isRenderingComponent()
+    {
+        return app(ExtendBlade::class)->isRenderingLivewireComponent();
```
levu42, line 57: What's the performance of this? It looks to me like this is called hundreds / thousands of times when rendering a template, and each time calling `app(ExtendBlade::class)` - is that fast enough to be called that often? Or is there a better way to do this?

levu42: I should've read @joshhanley's comment above before looking into the code :upside_down_face: That's great, that the performance is only minimally impacted by this!

edvordo:
Any chance to get this and #8282 merged soon? 
I'd like keep `view:cache` (as part of `optimize`) running on deployment.

JackWH:
Any updates on this? I believe it's related to https://github.com/livewire/livewire/discussions/5658 (updating nested components in a loop, when conditionally rendered, breaks after the first view caching).

samharvey44:
No update on this discussion in a little while, thought I'd see if there was any chance this may be merged in the near future?
It would be great to get the performance benefits of `view:cache` in production 🙂

JackWH:
@calebporzio @PhiloNL Hi there, hope you don't mind a nudge on this important issue! 👀 Please can we get this PR merged soon? To recap, this fixes a critical issue with complex Livewire apps in production:

- Running `php artisan view:cache` pre-renders the application markup **without** including Livewire morph markers
- As this command is usually run whilst deploying to production, the lack of morph markers causes bugs and unexpected behaviour between different environments
- This issue is also causing [nested components to break](https://github.com/livewire/livewire/discussions/5658#discussioncomment-9595582) when rendered conditionally
- I've also experienced this breaking Filament modals which appear conditionally. The issues go away when view caching is disabled.
- This PR should hopefully fix a handful of non-obvious issues once merged! 🙏 
- @joshhanley has posted [a thorough explanation](https://github.com/livewire/livewire/pull/8068#issuecomment-2002717752) of the proposed solution.

neil-cldagency:
Just ran into this one as well, would be good to see a resolution as it is tough to diagnose after using seemingly benign caching commands on a remote server.

jeffchown:
Hi @joshhanley - question, while we wait for this PR to be merged, if we are trying to `view:cache`-proof our Livewire 3 code (so we can safely use `php artisan optimize` or `php artisan view:cache` in production), is it simply a matter of wrapping all Blade conditionals and loops in `<div>`s?  Or are morph markers used to 'mark' other blocks of markup related to Livewire?

thumbtech:
> Hi @joshhanley - question, while we wait for this PR to be merged, if we are trying to `view:cache`-proof our Livewire 3 code (so we can safely use `php artisan optimize` or `php artisan view:cache` in production), is it simply a matter of wrapping all Blade conditionals and loops in `<div>`s? Or are morph markers used to 'mark' other blocks of markup related to Livewire?

I would not be concerned with letting LW3 handle the caching on the fly in production. Avoid optimize and just cache config, event & route in production.

jeffchown:
@thumbtech As I have a standard configs for all my production servers running Laravel apps (most using TALL), I want to both understand the details related to this issue, and to do my best to 'immunize' my Livewire code against possible side-effects related to (even inadvertent) view caching of my projects - to help minimize config changes (e.g. Forge, Envoyer scriptsm etc.) and minimize any extensive debugging that could arise unexpectedly.

thumbtech:
@jeffchown well it is a really bad bug and can cause (in Caleb's words) "flaky" behavior. Seems like focusing on your deployment scripts is the lower-hanging fruit, but if you find a good workaround please share!

jeffchown:
@thumbtech It is definitely tricky territory and I want to make sure whatever approach I use is simple to implement/automate (for me and my team), reliable and stable with or without this PR merged.

I will share anything of note that I find 👍

joshhanley:
@jeffchown yeah wrapping the Blade directives in an element (with the directive being the only child) will help or being liberal with `wire:key`. At the moment with the morph markers, you don't need `wire:key` for elements say inside an `@if` conditional, where as if the morph markers don't work, then elements inside conditionals and next to/ after conditionals will need keys.

If you want to be really sure, you can disable morph markers in the Livewire config so it's consistent with and without view caching.

Personally I would just do as @thumbtech suggested though and just not run view cache for now.

jeffchown:
Thanks, @joshhanley !

novikor:
Have just faced this issue, luckily I spent more than a day to find the solution 😅 

Have replaced the simple `php artisan optimize` with
`php artisan optimize --except=views`
in my deployment steps. 

Whatever solution would be considered, I believe it would be nice to have it described in https://livewire.laravel.com/docs/troubleshooting

tanthammar:
> Have just faced this issue, luckily I spent more than a day to find the solution 😅
> 
> Have replaced the simple `php artisan optimize` with `php artisan optimize --except=views` in my deployment steps.
> 
> Whatever solution would be considered, I believe it would be nice to have it described in https://livewire.laravel.com/docs/troubleshooting

I just spent days figuring out why I could not deploy on Laravel Cloud. 
Got server error 500 on every other page load. Which were gone if I reloaded the page a second time... 
Or, console errors about "undefined" is not valid Json
`php artisan optimize --except=views` solved it for me to.

Agree, that this should be highlighted in the docs with capital letters...

src/LivewireManager.php
joshhanley: @levu42 decided to by pass the calls to the container, should make it a touch faster 😁

joshhanley:
Thanks everyone for your input. We're currently in the process of reviewing the performance of Livewire so hopefully this will help as view will now be able to be cached again with Livewire.

This has been merged into the `main` branch, but it might be a week or two before we tag. We'd love it if you could test it out and let us know if there are any issues or performance problems with this. Running `php artisan view:cache` will be slower than before, but hopefully once the views are cached, everything is still nice and performant.

moazam1:
After a week of troubleshooting I landed on medium post which led me here. I am hoping this will be tagged soon. Thank you for making this great package @calebporzio

joshhanley:
@moazam1 thanks for testing and that's great to hear! We've got one other PR that we want to get into this release, which should be merged early next week. Then we will tag a release.

zepfietje:
I'm running into this issue too and was wondering when this PR would make it into a release (since it's been a month now).

Any update on that, @joshhanley? 🙂

moazam1:
@joshhanley unfortunately this issue is back again.

joshhanley:
@zepfietje I think it's being delayed now until Laracon US.

@moazam1 not sure what you mean by it's back again, we haven't released this yet.

ziming:
any reason why it need to be delayed? its a bug fix, not a new feature after all and laracon US is another month away

thank you

joshhanley:
@ziming yep there are reasons, but none that I'm allowed to say at this stage, sorry!

zepfietje:
> @zepfietje I think it's being delayed now until Laracon US.

Thought so. 😉
No worries, using a fork with this fix for now. 👌

moazam1:
@joshhanley for like a week it started working fine on production. May be it's temperamental.

PR #8084, opened 2024-03-08
Added failing unittest for deprecation

areichardt-apo:
Added a unittest which fails with a deprecation.
Requested for this bug report: https://github.com/livewire/livewire/discussions/8057


calebporzio:
Thanks, however I believe this has been fixed here: https://github.com/livewire/livewire/pull/8099

If I'm wrong, please re-submit this PR

PR #8087, opened 2024-03-09
Enable cookie testing on subsequent request

fikri-kompanion:
This is a continuation from PR #6822. It was not completed because the cookie is not passed on the subsequent request.

The idea is to also able to read cookie on subsequent request, which is when calling the action method. When manually testing inside browser, the cookie works. But fail on Livewire unit test. This PR should fix the problem. Test is also included.


calebporzio:
Thanks @fikri-kompanion !

PR #8093, opened 2024-03-11
add cancelable alpine:navigate event

n1crack:
Hi, 

I think it is needed.

yes

no

yes

While integrating TinyMCE into my project, I encountered a situation where I needed to prevent navigation triggered by wire:navigate. Unfortunately, there was no built-in method to achieve this within Alpine.js.

This gives developers the power to deal with situations where they need to check certain conditions before letting navigation happen, just like the code snippet shows.

```js

    document.addEventListener("alpine:navigate", function handler(event) {
        if (tinymce.activeEditor?.isDirty()) {
            event.preventDefault();
            tinymce.activeEditor.windowManager.confirm('There are changes that have not been saved. Do you want to leave the page?', (state) => {
                if (state) {
                    tinymce.activeEditor.resetContent()
                    Alpine.navigate(event.detail.url);
                }
            });
        }
    });
```


joshhanley:
@n1crack thanks for the PR! Can you please add tests? Thanks!

n1crack:
I will, asap.

n1crack:
"livewire:before-navigate" version of the event is added.

the test is also included.

n1crack:
Im sure the tests are ok, I think I need to send the dist files too..  But my setup is ignoring latest changes in the dist folder, I don't want to break anything. Id be glad if someone checks and add the build files.. 

image


image

calebporzio:
Thanks @n1crack. This is probably the right move, I just want to make sure we're not adding an event we'll later regret.

Turbo Drive has: `turbo:click` -> `turbo:before-visit` -> `turbo:visit`
So we would have: nothing -> `livewire:before-navigate` -> `turbo:navigating`

Which I think makes sense. What do you think @joshhanley?

js/plugins/navigate/index.js
```diff
@@ -45,7 +45,18 @@ export default function (Alpine) {
             })
 
             whenItIsReleased(() => {
-                navigateTo(destination)
+                const cancelableEvent = new CustomEvent('alpine:before-navigate', {
```
calebporzio, line 48: Can this be cleaned up into a separate method?

n1crack: yes in a min.

calebporzio:
Another note: the code you have only dispatches this when a navigation is triggered from a link click.

I think it makes sense for this event to be triggered from anywhere a navigate happens (`Livewire.navigate()`) for example, but not for navigates triggered from history popstate events like the back and forward buttons right?

n1crack:
I think they are not the problem as those actions are preventable (with other events).

about the naming: before-navigate is not too weird, since there are examples
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation/onBeforeNavigate

joshhanley:
@calebporzio an ideal world to me would be that `alpine:navigating` event actually gets a `stop/cancel` callback passed to it, that someone could then listen for the navigating event and run some logic and call `cancel()` if they decided to stop the navigation.

Basically the same as how we have `preventDefault()` for page expiration https://livewire.laravel.com/docs/javascript#customizing-page-expiration-behavior

I'm not a fan of the multi-word `alpine:before-navigate`. Would it be bad just to make it `alpine:navigate` as it's just an event name? That way it goes `navigate -> navigating ->navigated`?

n1crack:
@joshhanley  yes I think alpine:navigate is better. i changed it.
 
Im not sure how should I implement the "stop/cancel callback".  I don't want to mess up the current codebase :( .

joshhanley:
@n1crack yeah I probably wouldn't add it without feedback from Caleb first to see if he likes the idea 🙂

n1crack:
> Whereas, if you ask someone about livewire:before-navigate, they would certainly say: "Before a page is navigated to, duh"

Actually, livewire:navigate seems ok to me, its like an intention and the event contains the url that is intended.

We have an example below, it does exact same thing. 

https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate 

```
// Intercept navigate events, such as link clicks, and
// replace them with single-page navigations
navigation.addEventListener("navigate", (event) => {
  event.intercept({
    async handler() {
      // Navigate to a different view
    },
  });
```

calebporzio:
Ok, yeah I think I agree, and good call re: the navigate API reference.

I refactored the code a bit and now I have the current hangup before merging:

## The scenario
There are 3 different "navigation" scenarios in the code path:
* A normal navigation (usually triggered by a link click or `Livewire.navigate()`)
* A history state navigation (back/forward button click) WITH the page cached in history state
* A history state navigation WITHOUT the page cached in history state (creates a new fetch request for the page)

According to the current code path (in this PR), the following events fire in these three scenarios:

**A normal navigation**
* `navigate` (cancellable)
* `navigating` (cancellable)
* `navigated`

**A history state navigation (WITH page cached)**
* `navigating`
* `navigated`

**A history state navigation (WITHOUT page cached)**
* `navigate` (cancellable)
* `navigating` (cancellable)
* `navigated`

## Some questions

At best, users might be confused about the varying event sequences, at worst, they can't achieve what they want and we regret locking ourselves into the current setup.

A few questions that come to mind are:
A) Should history state navigations (cached or not) fire the exact same events
B) Should ANY history state events be "cancellable"
C) Should the `navigate` event be fired for all navigations? including history ones? (I believe this is how the referenced browser Navigation API works (fires `navigate` on ALL navigations))
D) Should ALL navigations, history or not fire the exact same methods? Or Are there times people only want to respond to link clicks, and in that case, should we be instead adding a "navigate-click" event or something like that (Turbo has `turbo:click` for this case)

Hopefully we can get some clarity on this and get this past the finish line! Thanks for everyone's time and input.

edalzell:
> D) Should ALL navigations, history or not fire the exact same methods? Or Are there times people only want to respond to link clicks, and in that case, should we be instead adding a "navigate-click" event or something like that (Turbo has turbo:click for this case)

This one. When things aren't consistent it makes it very hard to figure out what's going on. People could be free to disregard in their own code, if they want (assuming there's a way to tell the different scenarios apart). But in general I don't care HOW something happens only that it's always the same cuz to me "navigating" is all the same, regardless of the "how".

calebporzio:
Good feedback @edalzell, I agree.

Also, @n1crack, you're right that your snippet would be an infinite loop. You would need to check the "event.detail.url" or something to make sure you're not infinite looping, idk. I imagine Turbo's "turbo:visit" or something would be the same and could trigger an infinite loop like that right?

So here's my final proposal:

Three events:
* `livewire:navigate`
  * Fires every-time a navigation is started
  * Cancellable
  * `event.detail = { url: '...', history: true/false }`
* `livewire:navigating`
  * Fires after a response is received but before it's handled on the page
  * Not cancellable
  * `event.detail = {}`
* `livewire:navigated`
  * Fires after a navigate has finished
  * Not cancellable
  * `event.detail = {}`

What do we think? Is this good? It's not a breaking change and allows for most things I can think of because we're providing `url` and `history` to the event.detail.

n1crack:
Well if there is an infinite loop, then the user should be responsible :) Like the first snippet I share: 
 something like `tinymce.activeEditor.resetContent()` should break the loop. 

I am very positive with the current state. Now that I think about it, what do you think about the 'livewire:unload' event? (Maybe instead of 'livewire:navigate'). I know I'm insisting on 'livewire:navigate', but I just wanted to mention it anyway. if you think it is something irrelevant, just ignore :).

calebporzio:
Ok, i finalized the implementation, added tests, and added docs.

Any last words before I merge this?

n1crack:
Thanks, I couldn't help much, I wish I could.

calebporzio:
You helped plenty @n1crack - thanks!

n1crack:
use `livewire:navigate` event instead `livewire:navigated`, And it is a native event, so you can cancel the event with `event.preventDefault();`

For example: 

```js
document.addEventListener("livewire:navigate", function handler(event) {
  event.preventDefault();
  // logic
});

```

[Check here](https://livewire.laravel.com/docs/navigate#javascript-hooks) for more info.

binaryfire:
Hi @n1crack. Sorry that was a typo in my original post. I am using `livewire:navigate`, not `livewire:navigated`. 

Actually it looks like the problem is with the logic I'm using to determine the page's dirty state. It's causing the event to fire multiple times for some reason. Need to do some more debugging. Thanks for the suggestion though!

PR #8099, opened 2024-03-12
BaseUrl Attribute properties are not filled similarly when using queryString() methods

robert-stanciu:
Yes, I have created a [discussion](https://github.com/livewire/livewire/discussions/8070). This is a fix for warnings that appear when using [query string trait hooks](https://livewire.laravel.com/docs/url#trait-hooks)

yes, fix-deprecated-null-param-on-BaseUrl

No

Yes

Just the fact that $subName is `null` by default will eventually cause this warning to appear.

Nullable property: https://github.com/livewire/livewire/blob/main/src/Features/SupportAttributes/Attribute.php#L19C70-L19C85

The warning appears here: https://github.com/livewire/livewire/blob/main/src/Features/SupportQueryString/BaseUrl.php#L42

warning message: ReflectionClass::hasProperty(): Passing null to parameter 1 ($name) of type string is deprecated

### DIFF
--- a/src/Features/SupportQueryString/BaseUrl.php
+++ b/src/Features/SupportQueryString/BaseUrl.php
@@ -39,7 +39,7 @@ class BaseUrl extends LivewireAttribute
         $reflectionClass = new ReflectionClass($this->getSubTarget() ?? $this->getComponent());
 
         // It's nullable if there's a nullable typehint like: public ?string $foo;
-        if ($reflectionClass->hasProperty($this->getSubName())) {
+        if ($this->getSubName() && $reflectionClass->hasProperty($this->getSubName())) {
             $property = $reflectionClass->getProperty($this->getSubName());
 
             return $property->getType()?->allowsNull() ?? false;
--- a/src/Features/SupportQueryString/BaseUrl.php
+++ b/src/Features/SupportQueryString/BaseUrl.php
@@ -39,7 +39,7 @@ class BaseUrl extends LivewireAttribute
         $reflectionClass = new ReflectionClass($this->getSubTarget() ?? $this->getComponent());
 
         // It's nullable if there's a nullable typehint like: public ?string $foo;
-        if ($this->getSubName() && $reflectionClass->hasProperty($this->getSubName())) {
+        if ($reflectionClass->hasProperty($this->getSubName())) {
             $property = $reflectionClass->getProperty($this->getSubName());
 
             return $property->getType()?->allowsNull() ?? false;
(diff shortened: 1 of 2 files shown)


PhiloNL:
Thanks for the PR @robert-stanciu. Any chance you could add a test as well? 🙏

robert-stanciu:
Hi @PhiloNL, I added a failing test. Assuming the method way of adding query string parameters must work the same as the property way, the tests should be correct.

calebporzio:
Thanks for the submission and for adding a test @robert-stanciu 

I pulled this down locally, ran the test, and received a failure? Was that intentional?

Couple other notes:
* Please drop the docblock typehints in the test (not consistent with the rest of the codebase)
* Please re-name the test or restructure it to be more clear as to what it's actually testing, I'm confused by it

Thanks!

robert-stanciu:
Hello @calebporzio, I recently updated Livewire from version 2 to 3, and I am currently exploring the new options for adding query string parameters.

Yeah, the failure is intentional. I am expecting the three new ways for adding a query string parameter to behave the same way in terms of attributes properties (the test showing they're not) so that php doesn't throw warnings when making [this check](https://github.com/livewire/livewire/blob/main/src/Features/SupportQueryString/BaseUrl.php#L42). (Unless this is intentional and maybe the fix I proposed is enough to avoid the warnings)

The three ways I am taking about are:
- component property + #[BaseUrl]
- component queryString() method
- component trait queryStringTraitName() method

My attempt to fix the warnings is more likely related to the fact that [$subName is nullable](https://github.com/livewire/livewire/blob/main/src/Features/SupportAttributes/Attribute.php#L19C70-L19C85) and php throws a warning [here](https://github.com/livewire/livewire/blob/main/src/Features/SupportQueryString/BaseUrl.php#L42C43-L42C62)

joshhanley:
@robert-stanciu I'm not sure I follow, is this PR a failing test PR or a fix? If it's both, can you split the fix (and fix test) into one PR and the failing test for what is not working into another PR? That way we can assess your current fix and potentially merge it before the rest of it gets investigated 🙂

robert-stanciu:
Hi @joshhanley, I will keep the failing test without that fix.

This failing test proves that the BaseUrl Attribute properties are not filled similarly when using queryString() methods.

I will create another PR for the nullable property fix.

calebporzio:
Ah, I see your other PR #8099, I'll close it and add the fix here. In the future please keep the fix and failing test together so the feature/fix is treated as a single unit.

I added the fix here but the test is still failing for me. Please tweak this PR until the failing test is fixed then we can review and merge. Thank you.

robert-stanciu:
I fixed and split the previous test into 3 smaller ones so it's easier to follow.

calebporzio:
Thank you!

PR #8106, opened 2024-03-12
Fix highlighting in blade samples

machour:
These are currently not being picked up by the documentation rendered

Screenshot 2024-03-12 at 4 37 47 PM


PhiloNL:
Thanks @machour!

PR #8108, opened 2024-03-12
Allow an array of possible URL parameters for Url::$as

TWithers:
This PR add additional functionality to URL parameters (`#[Url]`). 

We have different platforms linking to some Livewire pages, but the platforms pass different query string parameters. One platform might pass `first_name=Tim&last_name=Withers` while another may pass `fname=Tim&lname=Withers`

Currently, there isn't a great way to handle this directly with Livewire, and we need to write out own `request()` handler. 

This PR allows you to do something like:

```php
#[Url(as: ['fname', 'first_name'])]
public ?string $firstName;
```


joshhanley:
@TWithers thanks for the PR! That makes sense when pulling from the query string. But when a property is updated, how should Livewire know which name to use to push updates to the query string?

TWithers:
It is set to use the first in the list if none are present in the query string, or which ever is found in the query string first. On Mar 15, 2024, at 3:51 AM, Josh Hanley ***@***.***> wrote:﻿
@TWithers thanks for the PR! That makes sense when pulling from the query string. But when a property is updated, how should Livewire know which name to use to push updates to the query string?

joshhanley:
@TWithers ah ok sweet! 😁

calebporzio:
Thanks @TWithers, the use-case makes sense.

I'm hesitant because of the following reasons:
* It's not immediately obvious which one would be used to "set" the query string
* If `fname` is used to load the page, when the value changes, what happens to the url?:
  * A) `fname` is changed to `first_name`
  * B) the old `fname` value is left while the new `first_name` value is added
* I just have a vague sense this will introduce extra unforeseen complexity for us down the road as we can't point to a single alias anymore

I'm curious to see what your workaround is in the meantime (The `request()` manipulation you referenced).

I'm sympathetic to this cause because if you want to change a `#[Url]` alias, you are invalidating a bunch of old URLs and this would allow you to support old URLs that might still be out there.

I'm still on the fence, but want to talk more about it.

Thanks for your time!

TWithers:
@calebporzio Hey, thanks for taking a look. I did a quick update to the code and tests to address your issues from above. Here is a quick break down of the scenarios that the PR covers (which I think is all possible scenarios):

1. If `as:` is not passed or is a string, everything functions as normal.
2. If `as:` is an array, it searches the QS params in the order they are listed.
```php
// $queryString = "?baz=1";

#[Url(as: ['foo', 'bar'])
public ?string $foo; //$foo is empty

---
// $queryString = "?bar=1";

#[Url(as: ['foo', 'bar'])
public ?string $foo; //$foo is "1" because "bar" was found

---
// $queryString = "?bar=1&foo=2";

#[Url(as: ['foo', 'bar'])
public ?string $foo; //$foo is "2" because "foo" was found first and "bar" is ignored

```
3. If a parameter is found in the QS, it will use the found param for updating. 
```php
// $queryString = "?bar=1";

#[Url(as: ['foo', 'bar'])
public ?string $foo; //$foo is "1"

$this->foo = "3"; // $queryString = "?bar=3"

---
// $queryString = "?bar=1&foo=2";

#[Url(as: ['foo', 'bar'])
public ?string $foo; //$foo is "2"

$this->foo = "3"; //$queryString = "?bar=1&foo=3"
```
4. If no parameters are found in the QS, it will use the first parameter in the array for updating.
```php
// $queryString = "?baz=1";

#[Url(as: ['foo', 'bar'])
public ?string $foo; //$foo is empty

$this->foo = "3"; //$queryString = "?baz=1&foo=3"
```
5. If the QS was changed at some point manually or via another component or something, then yes, there could be an issue, which is I think what you addressed in your comment:
```php
// $queryString = "?bar=1";

#[Url(as: ['foo', 'bar'])
public ?string $foo; //$foo "1"

$this->foo = "3"; //$queryString = "?bar=3"

// $queryString updated "?bar=3&foo=1" somehow and component is $refreshed
// $this->foo finds "foo" first in the QS and it becomes "1" and the alias is now "foo"

$this->foo = "4"; // $queryString = ?bar=3&foo=4
```

To be fair though, I am not sure #5 is a super big problem. If another component was also using `#[Url]` and is setting `foo=1`, then `foo` is likely what is being used and relied on, so the other component grabbing `foo` from that point forward makes sense. If it wasn't desired, the user could swap the priorities (order) of the `as:` alias array and it would function correctly. 

Currently, in the project where I am wanting this, I am only reading in data, and not updating the URL, so my implementation has been fairly simple. I created a trait with the following method that I can throw onto a couple of my components, and it seems to do the trick for the one-way binding:
```php
    /**
     * $queryStringMap is the array that would be returned by "queryString<TraitName>()" 
     * with the "as" supporting an array
     * /
    public function findQueryStringParameters(array $queryStringMap): void
    {
        collect($queryStringMap)
            ->each(function ($options, $componentParam) {
                $urlParamValue = collect($options['as'])
                    ->map(function ($value) {
                        if (! app('livewire')->isLivewireRequest()) {
                            return request()->query->has($value) ? request()->query->get($value) : null;
                        }

                        $queryString = parse_url(request()->header('Referer'))['query'] ?? null;

                        if ($queryString) {
                            parse_str($queryString, $query);

                            return $query[$value] ?? null;
                        }

                        return null;
                    })
                    ->filter()
                    ->first();

                if ($urlParamValue !== null) {
                    $this->$componentParam = $urlParamValue;
                }
            });
    }
```

In the end though, the above code works just fine for me and the project as it stands now. Whatever way you decide to go with the PR, I am fine with it.

calebporzio:
Thank you for the thorough and clear response and thanks for providing the workaround for other's reference.

I think I'm going to close this FOR NOW. It's a good PR that we can re-open and merge if there's enough community demand, but I think I want to hold off on the risk of increasing the complexity of the history functionality in a way we regret down the road because there can be N aliases instead of simply 1 like right now.

Thanks for all your time, work, and attention!

TWithers:
No worries!

PR #8117, opened 2024-03-14
patch vulnerability

calebporzio:
This PR patches an XSS vulnerability in Livewire that was introduced in version v3.3.4.

The attack goes like this:
* On a Livewire page that uses `#[Url]` for a property
* Tweak the URL's query string to include malicious code
* Share that URL with other users in the app
* When they open it, it will inject HTML onto their pages, therefore allowing the malicious user to execute JavaScript in another user's environment (XSS vulnerability)

The reason this was happening was Livewire was using `let data = {}` and then building up a JS object directly from the query string using syntax like this: `data[key] = value`.

This is a problem because `let data = {}` creates an object that inherits from `Object.prototype` in the target's environment. Allowing injections such as this: `data['__proto__.html'] = "`

Now, when Livewire checks objects for "html" to morph onto the page, it will find this injected HTML in EVERY object in the app because it's attached to the prototype.

The fix is simple: instead of creating an object using an object literal: `let data = {}`, create a "pure" JS object using `let data = Object.create(null)` which DOES NOT inherit from `Object.prototype`.


caendesilva:
@calebporzio Could you please publish a security advisory so security services like Dependabot and Snyk can provide automated updates to projects?

Carnicero90:
May I ask why you are not using a Map? As far I see "data" is used just for map operations (has, get, set, delete), and you are actually reimplementing it (unless I'm missing something).

PR #8118, opened 2024-03-14
Add loading indicators for wire:navigate

patriktoth67:
Feature discussed here: #7578

Livewire allows you to indicate loading states on parts of your page when navigating between Livewire components using the wire:navigate directive. To utilize this feature, you can add the wire:loading.navigate directive to any element that you want to show during the navigation loading state.

To display an element while a navigation request is loading, add the wire:loading.navigate directive to your element:

```blade
<div wire:loading.navigate>
    Loading, please wait...
</div>
```

This element will become visible when a link with the wire:navigate directive is clicked and the request is in progress.

You can introduce a delay to the appearance of the loading indicator, which can help avoid a flickering effect for very fast navigation transitions. Livewire supports a range of predefined delays you can apply:

```blade
<div wire:loading.navigate.delay.shortest>...</div> <!-- 50ms delay -->
<div wire:loading.navigate.delay.shorter>...</div>  <!-- 100ms delay -->
<div wire:loading.navigate.delay.short>...</div>    <!-- 150ms delay -->
<div wire:loading.navigate.delay>...</div>          <!-- 200ms default delay -->
<div wire:loading.navigate.delay.long>...</div>     <!-- 300ms delay -->
<div wire:loading.navigate.delay.longer>...</div>   <!-- 500ms delay -->
<div wire:loading.navigate.delay.longest>...</div>  <!-- 1000ms delay -->
```


patriktoth67:
It is not yet finished it seems, I'm still finding some issues with it

patriktoth67:
Right now, if anywhere in the document a link with `wire:navigate` is clicked, all of the `wire:loading.navigate` divs will appear, so it's still a work in progress

patriktoth67:
Now it should be good

calebporzio:
Thanks for this submission.

I have a few thoughts/questions:
A) This feature is spread across the "loading" and "navigate" parts of the codebase, adding extra coupling/complexity we might regret
B) This feature adds extra CSS to Livewire's scripts that increases the overhead of using Livewire
C) The `wire:loading.navigate` API doesn't account for things like `wire:loading.navigate.flex` which users may want, but would yet again increase the complexity and CSS overhead

To address B and C, we might use an alternate API like:

```
wire:loading wire:target="$navigate"
```

I saw someone in the discussion suggested `wire:target="wire:navigate"`, and although it makes sense I wonder if it would be confusing and make people think they can do things like `wire:target="wire:click"` which they can't.

I'm also wondering if we instead add better lifecycle event hooks to the navigate system to allow users to manage this themselves (we're discussion better events in this PR https://github.com/livewire/livewire/pull/8093)

For example, if there were better events for wire:navigate, users could create and manage their own loading indicators:

```
let indicator = document.querySelector('#loading-indicator')

document.addEventListener('livewire:navigate-show-indicator', () => {
    indicator.style.display = 'block'
})

document.addEventListener('livewire:navigate-hide-indicator', () => {
    indicator.style.display = 'none'
})
```

I understand this puts a lot of complexity on users, but it could be a way to give the few people who desire this the ability.

I'm still on the fence about this for all the reasons above, but willing to here others thoughts about my hesitations.

patriktoth67:
@calebporzio 
All right, let's go with `wire:loading wire:target="$navigate"`

I just noticed, that I left this in `js/plugins/navigate/index.js` by mistake : 
```fireEventForOtherLibariesToHookInto('alpine:start-navigating')```

The feature works without listening for lifecycle event hooks from the navigate system. I tried to do it that way (hence the left in code), but I needed the livewire components ID as an event detail. That shouldn't be accessible in the Alpine navigate plugin I think.

Instead it works by firing a `livewire-navigating-start` event when a link with the `wire:navigate` directive is clicked, trying to go for the same setup as the file upload system with the `livewire-upload-start` event. This way I can send the components ID with the event.

And we need the components ID when displaying the loading indicator to only display it whenever a link in the same component is clicked. This way we can have different loading indicators for different livewire components on the same page, and we don't have to worry about all the loading indicators appearing at once whenever any navigate link is clicked.

The crossover into the navigate system should only be this part of `wire-navigate.js`, which only fires 2 events.
```
directive('navigate', ({ el, component }) => {
    whenThisLinkIsPressed(el, whenItIsReleased => whenItIsReleased(() => {
        window.dispatchEvent(new CustomEvent('livewire-navigating-start', { bubbles: true, detail: { id: component.id } }));
    }));

    document.addEventListener('alpine:navigating', () => {
        window.dispatchEvent(new CustomEvent('livewire-navigating-end', { bubbles: true, detail: { id: component.id } }));
    });
});
```

So to meet your concerns:
A) The crossover should be minimal I believe
B-C) I'll change to work with `wire:loading wire:target="$navigate"` to eliminate the CSS overhead and standard functionality

I'll commit the changed version soon, I believe it should make it more streamlined.

calebporzio:
Ok, thanks for the update, I'll look at the other PR. Closing this...

PR #8120, opened 2024-03-15
Fix file input clearing

joshhanley:
Today @jdiddydave and I ran into an issue with a file input not being cleared when we called `unset($this->fileProperty);`.

Resetting the property meant the file input was still showing the old filename instead of an empty file input.

This PR adds a fix for this, but I'm not sure if it's is a great solution. I suspect there might be a better Alpine way to handle the watching of the property and resetting of the file input.

One thing to note is we are only resetting the file input if the property value has been reset to `null` or an empty string.

@calebporzio any thoughts on a better way to handle this?


js/features/supportFileUploads.js
```diff
@@ -45,6 +45,13 @@ export function handleFileUpload(el, property, component, cleanup) {
 
     el.addEventListener('change', eventHandler)
 
+    // If the Livewire property has changed to null or an empty string, then reset the input
+    component.$wire.$watch(property, (value) => {
```
calebporzio, line 49: I don't love adding a $watch here. It feels like hooking in to the reactivity engine could cause headaches in the future (lack of being cleaned up or something weird like that)

Are there other, more "stateless" approaches to this we can try?

Worst case we can hook into "commit" and check if inputs should be cleared when requests come back from the server.

But I feel like there might be something clever we're missing.

js/features/supportFileUploads.js
```diff
@@ -45,6 +45,13 @@ export function handleFileUpload(el, property, component, cleanup) {
...
+        if (value === null || value === '') {
+            el.value = ''
```
calebporzio, line 51: We can use "clearFileInputValue()" here right? (if we extract it to a util fuction?)

calebporzio:
I did a little writeup and affirmed the proposed approach, just had to do some refactoring to make sure the watcher is cleaned up if the component or element is removed from the DOM.

Here's the writeup for the archives lol:

## The setup

Given the following part of a  Livewire component:

```php
public $photo;

public function resetInput()
{
    unset($this->photo);
}
```

With the following template:

```html
<div>

    <button wire:click="resetInput">Reset</button>
</div>
```

If a user selects a file, `$photo` is updated with a temporary file object and the file input on the frontend should show the selected filename.

Now, if they press the "Reset" button, `$photo` should be reset back to null, and the file input field on the frontend should be cleared and show something like "Choose file...".

## The problem

The problem is that when "Reset" is pressed, the `$photo` property IS cleared in the backend, but the `<input>` element on the frontend is NOT cleared.

Before we explore that problem, let's first understand how a normal "text" input with `wire:model` get's cleared when a property is "unset":

* Livewire initializes and sees an element with `wire:model` on it
* It attaches an Alpine `x-model` under the hood to the specified property
* That `x-model` internally uses a reactive `effect()` to force the input to reset itself to the stored value
* Therefore, when a value on the backend is changed, the reactive effect picks it up and resets the input

In Livewire's internals, `wire:model` for file inputs is handled differently than every other type of input.

When `wire:model` for a file input is initialized, Livewire does NOT create an `effect` to watch the value, therefore, there is nothing to "catch" the changed `$photo` value on the frontend.

## The solution

### Solution A) Add an effect/watcher
The simplest thing to do would be to add a watcher/effect similar to `x-model` in Alpine.

The only con of this approach is making sure it gets cleaned up when the element is removed from the DOM.

### Solution B) Something else?

Nah, I think solution A is the best. Especially after reminding myself that a normal `wire:model` effectively adds a watcher itself. There's a precedence for it.

PR #8122, opened 2024-03-15
Update actions.md

danie-ramdhani:
Based on discussion https://github.com/livewire/livewire/discussions/8119, Receiving return values with `$wire` object in `x-text` directly will make infinite request because `x-text` will re-evaluate again and again.

Maybe this is the better example using `$wire` object to assign element text (?)


calebporzio:
Oooof thanks!

PR #8126, opened 2024-03-15
Add float synthesizer to handle empty inputs in combination with Forms

lsmedia:
1️⃣ _Is this something that is wanted/needed? Did you create a discussion about it first?_
Please refer to the problem description in this discussion: https://github.com/livewire/livewire/discussions/8124

2️⃣ _Did you create a branch for your fix/feature? (Main branch PR's will be closed)_
Yes

3️⃣ _Does it contain multiple, unrelated changes? Please separate the PRs out._
No, my PR contains the Float synthesiser and related tests (please see my comment on the tests below).

4️⃣ _Does it include tests? (Required)_
Yes, I have written various tests. I consider the test cases in combination with Forms (FloatFormSynthUnitTest) to be particularly important, as the situation mentioned in the discussion is tested here (test `empty_string_hydrated_returns_null`). Please decide whether you want to use the test cases in combination with Forms.

5️⃣ _Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful._
In my opinion, this change makes sense and is useful, as otherwise a `TypeError` is thrown as soon as a float variable in a form is updated with an empty value (empty string).


src/Mechanisms/HandleComponents/Synthesizers/Tests/FloatFormSynthUnitTest.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Livewire\Mechanisms\HandleComponents\Synthesizers\Tests;
+
+use Livewire\Component;
+use Livewire\Form;
+use Livewire\Livewire;
+
+class FloatFormSynthUnitTest extends \Tests\TestCase
```
joshhanley, line 9: Can you merge these tests into the FloatSynthUnitTest file and instead add like a `hydrated_in_a_form` to the test names or something like that?

src/Mechanisms/HandleComponents/Synthesizers/Tests/FloatFormSynthUnitTest.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    /**
+     * @test
+     */
+    public function null_value_hydrated_returns_null()
+    {
+        Livewire::test(FormComponentWithNullableFloat::class)
+            ->set('form.floatField', null)
+            ->assertSet('form.floatField', null, true); // Use strict mode
```
joshhanley, line 18: There is now an `assertSetStrict('form.floatField, null);` method you can use which is descriptive meaning we can get rid of all the strict mode comments 🙂

joshhanley (changes requested):
@lsmedia thanks for the PR! Looks good, just a couple of small changes 🙂

src/Mechanisms/HandleComponents/Synthesizers/Tests/FloatFormSynthUnitTest.php
lsmedia: As I wrote in the PR description, I wasn't sure whether explicit tests of hydrating forms were desired, as they are not part of the other synthesizer Unit Tests. In my opinion, however, this should definitely be tested explicitly.

Done in commit [feb8cb8](https://github.com/livewire/livewire/pull/8126/commits/feb8cb8c0d7c71fcae4fa4f0c7f517e78d9ab33a)

src/Mechanisms/HandleComponents/Synthesizers/Tests/FloatFormSynthUnitTest.php
lsmedia: Done in commit [feb8cb8](https://github.com/livewire/livewire/pull/8126/commits/feb8cb8c0d7c71fcae4fa4f0c7f517e78d9ab33a)

calebporzio:
Thanks!

lsmedia:
We could remove the explicit casting to float in the hydrateFromType function. This is not necessary in order to handle empty strings correctly.


```
static function hydrateFromType($type, $value) {
        if ($value === '' || $value === null) return null;

        return (float) $value;
}
```

The question arises as to whether it might then also make sense to remove the explicit casting in the synthesiser for integers.

levu42:
@lsmedia @PerryvanderMeer Isn't the original problem, that something that's `string|float $foo = ''` is encoded to have a float type, so that the float synth tries afterwards to parse it into a float? Shouldn't this be always just a string? Only when it's `string|float $foo = 42.0`, i.e. when it's actually a float, it should be sent over in a way that it gets decoded again into a float.

lsmedia:
If I see this correctly, then there are two legitimate use cases here and, in my opinion, it all depends on where the data validation and transformation takes place. Does transformation take place before the component is hydrated/updated or do you want to allow (uncast) string values to be stored in the component in order to validate and cast them correctly afterwards.

We could, as I described above, remove the explicit float casting in the synthesizer. With this solution, for example, `TypeError` exceptions would have to be handled on hydrating/updating the component. But then the Float synthesizer would have has no negative effect on the use case described by @PerryvanderMeer .

If, as @levu42  suggested, we give the synthesizers some kind of prioritization and give preference to string for type declarations such as `string|float` (ignore the float synthesizer ), then this could be a simple solution to always hydrate the string value in order to be able to assign the data in the correct data type after the update (This would also enable the use case described by @PerryvanderMeer).
To be honest, however, I see the danger of "hidden features" and unwanted side effects due to prioritization with this proposed solution.

What is your opinion?

PR #8131, opened 2024-03-15
Update forms.md

danie-ramdhani:
Based on https://github.com/livewire/livewire/discussions/8128#discussioncomment-8808143, redirect chained `with` session doesn't exist in livewire.


PhiloNL:
Thanks!

PR #8140, opened 2024-03-17
Added warning about event listeners with wire:navigate

samlev:
This is just a documentation update for `wire:navigate`. There's a little "gotcha" that a few people in the livewire discord have encountered - because the document/window aren't reset when navigating, event listeners attached to the document or window aren't removed. In particular, when listening for the `wire:navigated` event you might be adding multiple of the same event listener, or leaving a page-specific listener in place for other pages.

Toy examples:

1. Same/similar event listener being added multiple times.
```js
    axios.post('https:://example.com/track', {page: document.location})
});
```

If this was added somewhere where it was evaluated on every page, then the first navigation event would fire a tracking event, but the second page will have added another tracking event, so it will track the view twice, and another navigation will add it _again_, etc. because the old listeners aren't removed, they just keep stacking up.

2. Page specific listeners not being removed.
```js
    Maps.mount('#map-container');
});
```

This works fine on the intended page, but navigating the another page might throw an exception because `#map-container` doesn't exist on the new page.

Anyway, the wording could probably be cleaned up, but it's definitely a gotcha that's worth noting, because I've seen a couple of people run into it.


danie-ramdhani:
Mmm... I don't know if this is bug or must be manually removed or documented as limitations.

Problem:

- navigate using `wire:navigate`
- notice the multiple event is running

### Case 1: `livewire:navigated / alpine:navigated` event

```blade
{{-- NOTE: this will run multiple times when user is navigated to another page --}}
<div x-data="{
    init() {
            console.log('foo');
        });
    }
}">
    Hello World
</div>

{{-- Alternative solution --}}
<div x-data="{
    init() {
        document.addEventListener('livewire:navigated', function handle() {
            console.log('bar');
            // you need to remove the event manually like this:
            document.removeEventListener('livewire:navigated', handle);
        });
    }
}">
    Hello World
</div>
```

### Case 2 (Custom Event)

Assuming we have button that will dispatching custom event:

```blade
<button type="button" x-on:click="$dispatch('some-event')">
    Dispatch
</button>
```

```blade
<div x-data="{
    init() {
        // if we add `{once: true}` the event will only run once when some-event dispatched (even the button is clicked multiple times)
        // if not using `{ once: true }`, when user navigated to other page using `wire:navigate`, the event will run multiple times
        window.addEventListener('some-event', () => {
            console.log('bar')
        });
        // }, { once: true });
    }
}">
    Hello World 2
</div>

{{-- Alternative solution --}}
<div x-data x-on:some-event.window="console.log('foo')">
    Hello World 1
</div>
```

### Case 3 event listener using `addEventListener`

Assuming you're navigate to the same page first for 10x or more, then resize you window

```blade
{{--  this will log `foo` for a thousands times 😅 --}}
<div x-data="{
    init() {
        window.addEventListener('resize', () => {
            console.log('foo');
        });
    }
}"></div>

{{-- Alternative solution --}}
<div x-data="{
    init() {
        window.onresize = () => {
            console.log('foo');
        };
    }
}"></div>
```

And i don't think adding `{once: true}` will be the solution for case 2.
i always using the alternative solution for now because it's not running multiple times when navigate using `wire:navigate`.
If it's a limitation, maybe the documentation look like the code I commented with `Alternative solution`.

samlev:
Yeah, I don't think that it's a bug, but definitely not expected behaviour for a lot of people. I wonder if there's another mechanism that could be created for global event listeners that could be properly cleaned up. Something like `Livewire.navigated(() => { /* event listener */}, {always: true});` that handles attaching the event listener once and only once.

danie-ramdhani:
using `Livewire.navigated(() => { /* event listener */}, {always: true});` is not helpful because it happen with event using `addEventListener`, look case 3.

samlev:
No, I understand that entirely. I'm saying that it might be worth creating another way to livewire to add window/document event listeners that cleans up after itself in navigation events.

danie-ramdhani:
mmm... this is interesting one https://github.com/livewire/livewire/pull/8147

calebporzio:
Thanks for this!

PR #8143, opened 2024-03-18
Add browser test for navigate page with hashtag

danie-ramdhani:
Failing test for this issue: https://github.com/livewire/livewire/discussions/8073

How to reproduce the problem: 
- On a site using wire:navigate with href `/foo#bar`
- the url is missing the hashtag `#bar`


calebporzio:
Thanks for submitting this, that is indeed a problem. I'll look into fixing soon.

calebporzio:
This has been fixed, thanks for the failing test. Along the way I learned that hash fragments aren't sent to the server so when I pull the URL from `response.url` it's missing the hash fragment. I had to add a little extra logic, but we should be good now!

sgtcoder:
Still having this problem.

danie-ramdhani:
> Still having this problem.

I think It because https://github.com/livewire/livewire/pull/8870 PR i guess. I suggest locking your version at 3.5.6

ghabriel25:
@calebporzio @danie-ramdhani Hi, any update about this or this will be fixed for V4?

PR #8146, opened 2024-03-18
POC: add wire.$loading property

AlexandreBonaventure:
This is a draft MR made to illustrate a POC. Waiting on maintainers to chime in before eventually finalizing it.

Discussion: https://github.com/livewire/livewire/discussions/8145

### DIFF
--- a/js/$wire.js
+++ b/js/$wire.js
@@ -9,7 +9,7 @@ import Alpine from 'alpinejs'
 let properties = {}
 let fallback
 
-function wireProperty(name, callback, component = null) {
+function wireProperty(name, callback) {
     properties[name] = callback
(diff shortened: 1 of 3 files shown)


calebporzio:
Yeah, so this makes a lot of sense, I initially explored something like this and backed out for the following reasons:
* This implementation doesn't account for file uploads processing
* There's no way to "scope" this boolean to specific targets

However, I could see adding this (as long as file uploads are accounted for) as a simple: "this component has a request out to the server"

Or like you mentioned in your discussion, maybe there's another API like:

```js
$wire.isLoading({ targets: ['foo', 'bar'] }) or something like that
```

PR #8147, opened 2024-03-18
Prevent listeners leak with once modifier

AlexandreBonaventure:
Browsing through the code, I realized some event listeners are not cleared in the loading directive, effectively leaking memory (correct my if I'm wrong). 
Adding once modifier mitigates this issue. 

I've added a once modifier option to the internal event bus to make things consistent (can be useful in other places as well).


calebporzio:
Good call, I think "once" is the wrong approach because we need those listeners to handle multiple times, but you want them removed when the element/components with `wire:loading` are removed.

I refactored things. If the tests pass, I'll merge this.

Thanks!

calebporzio:
Thanks!

PR #8154, opened 2024-03-19
Add browser test for binding class when navigate back

danie-ramdhani:
Based on discussion: #7947 and https://github.com/livewire/livewire/discussions/6686

Problem:

- Show then hide foo using button
- Navigate using wire:navigate
- Click the button again
- The element class is not changed

Apparently, it's not happened when using class object syntax `<span :class="{ 'hidden':!show }">foo</span>`.


calebporzio:
Thanks for the failing test. Here's a writeup of my current problem investigation before I attempt a solution.

## The setup

Given the following HTML on a page:

```html
<div x-data="{ show: false }">
    <button x-on:click="show = ! show">Toggle</button>

    <span x-bind:class="show || 'hidden'">...</span>
</div>
```

When the page is loaded, Alpine will set the `hidden` class on the `<span>` because `show` initially false.

Now, a user can toggle the `hidden` class on and off by pressing the "Toggle" button.

In theory, they should be able to navigate around the app using `wire:navigate` links, and any time they revisit this page, they should see the same behavior.

## The problem

However, this is not the case. If a user does the following:

* Visits the page and presses "Toggle" a few times (showing and hiding the `<span>`)
* Navigates to a new page in the app
* Presses the "back" button

They are on the original page, however they CANNOT toggle on and off the `<span>` by pressing "Toggle".

### The explanation

The reason is that when "Toggle" is pressed, Alpine adds and removes the `hidden` class from the `<span>`.

If the `<span>` contains the `hidden` class when the page is navigated away from, the following HTML will be cached in the history state for when the page is revisited:

```html
<span x-bind:class="show || 'hidden'" class="hidden">...</span>
```
(Notice that `class="hidden"` exists in the HTML and wasn't in the original template. It was added by Alpine)

Therefore, when the back button is pressed and the above HTML is used to initialize the page, Alpine will initialize as normal, however, by default, Alpine won't dynamically remove a class if it is already on an element.

This behavior results in the `<span>` appearing to be broken and not toggleable, however under the hood everything is working properly, `hidden` is just not being removed.

## The solution

### Solution A) Tear-down Alpine classes before navigating away

When a page is about to be navigated away from, Alpine will tear itself down before it caches it's HTML contents.

We could add code into `x-bind` that removes any classes that have been dynamically added (if it's even possible).

This way, the cached HTML wouldn't contain the `hidden` class and everything would work as normal.

**Pros:**
* Simple
* Uses existing code hooks

**Questions:**
* Is it easy to revert `class="..."` back to it's original state from the `x-bind` source code in Alpine?

### Solution B) Allow Alpine to overwrite existing HTML classes

Alternatively we could tweak `x-bind:class` in Alpine to overwrite existing classes if it notices that a certain class binding is `false`.

This would "fix" the problem by overwriting the `hidden` class after the back button is pressed and the page initializes.

**Pros:**
* Simple

**Questions:**
* Would this have adverse effects in other areas of Alpine usage? It might be a breaking change for other instances...

calebporzio:
I went with solution A. The fix is in the Alpine repo here: https://github.com/alpinejs/alpine/pull/4111

It will be available in the next Alpine/Livewire release.

PR #8157, opened 2024-03-19
[3.x] Make commands lazy

timacdonald:
This PR makes the Livewire commands "lazy", i.e., they do not need to be instantiated when running _other_ commands.

Currently, when you have Livewire installed in a Laravel project, if you were to run `php artisan make:model` all of the Livewire commands are instantiated. This is just how Symfony console commands work.

The `AsCommand` attribute was introduced to allow commands to be "lazy", i.e., they do not need to be instantiated when running _other_ commands.

Doing some work across the Laravel ecosystem to standardise this and thought I would drop by here and do the same.

See: https://github.com/laravel/framework/pull/50617

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,12 +10,13 @@
   ],
   "require": {
     "php": "^8.1",
-    "symfony/http-kernel": "^6.2|^7.0",
     "illuminate/database": "^10.0|^11.0",
     "illuminate/routing": "^10.0|^11.0",
     "illuminate/support": "^10.0|^11.0",
     "illuminate/validation": "^10.0|^11.0",
-    "league/mime-type-detection": "^1.9"
+    "league/mime-type-detection": "^1.9",
+    "symfony/console": "^6.0|^7.0",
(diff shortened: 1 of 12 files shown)


composer.json
```diff
@@ -10,12 +10,13 @@
...
     "php": "^8.1",
-    "symfony/http-kernel": "^6.2|^7.0",
     "illuminate/database": "^10.0|^11.0",
     "illuminate/routing": "^10.0|^11.0",
     "illuminate/support": "^10.0|^11.0",
     "illuminate/validation": "^10.0|^11.0",
-    "league/mime-type-detection": "^1.9"
+    "league/mime-type-detection": "^1.9",
+    "symfony/console": "^6.0|^7.0",
+    "symfony/http-kernel": "^6.2|^7.0"
```
timacdonald, line 19: Add `symfony/console` and sorted.

calebporzio:
Lovely @timacdonald - thanks a bunch!

PR #8159, opened 2024-03-19
Add browser test for `wire:target` multiple actions parameters

danie-ramdhani:
Failing test based on discussion https://github.com/livewire/livewire/discussions/8134

Apparently, if `wire:target` contains multiple actions with params will broke livewire with error message `someTarget is not defined`.

### DIFF
--- a/src/Features/SupportWireLoading/BrowserTest.php
+++ b/src/Features/SupportWireLoading/BrowserTest.php
@@ -269,7 +269,7 @@ class BrowserTest extends \Tests\BrowserTestCase
         ->assertSee('Processing...')
         ;
     }
-    
+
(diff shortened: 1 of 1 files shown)


calebporzio:
Ok, I dug in and confirmed this is a bug.

However, it would be quite a bit of added code/complexity to address this. Therefore I'm choosing to keep this as a limitation of Livewire and added a documentation warning on the main branch: https://github.com/livewire/livewire/commit/7503ef50d2b317cde63c8c1a201775723aecc1eb

Sorry I can't address this right now. It would require a change in the way we parse these methods/params and the complexity I don't think is warranted until there's more demand for it. Thanks for the report!

PR #8162, opened 2024-03-19
update forms.md

danie-ramdhani:
Discussion https://github.com/livewire/livewire/discussions/8158#discussioncomment-8838378

Improve `PostForm` create and update based on validated data.

I extract to `$validatedData = $this->validate();` because sometimes we need to manipulate the data after the data validated.

```php
public function store()
{
    $validatedData = $this->validate();
    
    // manipulate the data...

    Post::create($validatedData);
}
```

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -83,11 +83,9 @@ class CreatePost extends Component
 
     public function save()
     {
-        $this->validate(); // [tl! highlight]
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the contribution, and that's a fine way to do it, although I don't think it's distinctly better than what's there, so I'm going to close this. Thanks!

PR #8165, opened 2024-03-19
Wire loading for wire navigate links

patriktoth67:
Feature discussed here: https://github.com/livewire/livewire/discussions/7578

Made a previous PR (#8118), but livewire got updated since, and the feature needed some restructuring, so I made a new PR.

New features compared to the other one are, that every modifier for `wire:loading` works with it now (like .flex, .delay etc), and it doesn't have the CSS overhead.

Livewire allows you to indicate loading states on parts of your page when navigating between Livewire components using the wire:navigate directive.

To use this feature, simply pass `$navigate` to the `wire:target` directive on an element you wish to show during navigation:

```blade
<a href="/posts" wire:navigate>Posts</a>
<div wire:loading wire:target="$navigate">
    Navigating, please wait...
</div>
```

This element will become visible when a link with the wire:navigate directive is clicked **in the same Livewire component**.

To extend loading indicators across all Livewire components on a page, use the `wire:target="$navigate.global"` attribute. This makes the loading indicator visible whenever any wire:navigate directive is activated, regardless of which Livewire component it belongs to.


patriktoth67:
Also an extra feature, that you can use the same loader for a property change, a method running or the page navigating away like this: 
```
<a href="/posts" wire:navigate>Posts</a>
<div wire:loading wire:target="property, method, $navigate">
    Loading, please wait...
</div>
```

patriktoth67:
I'm not familiar how wire:stream works, and if there are any loading indicators while data is streaming, but if not, `wire:target="$streaming"` would be logical to implement, and it would fit the expression.

patriktoth67:
I'm not sure there is a need for it, but right now `wire:loading` fires whenever `wire:stream` is streaming something. Is it worth to separate it to only fire if it is targeted at `wire:target="$stream"`?

patriktoth67:
Right now using `wire:target="$navigate"` will only fire the loading indicator, when the link is clicked in the same livewire component. It is desired sometimes to have a global loading indicator for all navigate links, which could be placed onto any livewire component on a given page.

To solve this, I added `wire:target="$navigate.global"`, which shows the loading indicator for clicking any link with the `wire:navigate` directive in any Livewire component on the page.

The syntax can be changed of course, I'm open to ideas, it's an easy change.

calebporzio:
Hmmm yeah, I'm hesitant to expand this API even further.

At this point I'm leaning towards just dispatching events from the system and let people do the toggling in userland.

What do you think about that?

I'm thinking:

```
document.addEventListener('livewire:navigate-indicator-show', e => {...})
document.addEventListener('livewire:navigate-indicator-hide', e => {...})
```

Then we can just dispatch these two events when progress bars are shown/hidden.

I'd be fine with that PR, I think I want to head that direction instead of the API explored here. Sorry if I mislead but I think the complexity and API differences aren't worth it at this point.

calebporzio:
Yeah, I think it should be component-agnostic and just show for all navigate events right?

patriktoth67:
If the new event's syntaxes have been decided already, I think the loader should appear on `livewire:navigate` and disappear on `livewire:navigating`.

If you confirm, I'll commit the new changes.

bamboozled-panda:
Livewire internals noob here. Please disregard if this question is too out of scope: will this work with redirects? Even those that don't use `wire:navigate`? I think it would be great to have a built-in way to keep loading indicators visible and buttons disabled during a redirect. Like in this situation:

```html
<form wire:submit="save">
    ...

    <div wire:loading>Loading...</div>
    <button type="submit" wire:loading.attr="disabled">Save</button>
</form>
```
```php
{
    ...

    public function save()
    {
        ...

        $this->redirect('/sometimes-slow-page');
    }
}
```

The loading indicator disappears and the submit button becomes enabled between `save` completing and the new page loading.

patriktoth67:
@bamboozled-panda with the current implication, no. If we do it with the new events from #8093 then I believe yes.

patriktoth67:
Well, after some reconsideration this might not be worth to integrate into livewire based on my use cases. If we are not doing targeting, just always show the loader on any navigate event, that might look pretty bad on fast loading pages. The loader would always need a delay on it to only target the slow loading pages.

With the new `livewire:navigate` events the only thing we could target with is the URL, so if this feature is needed we could maybe do something like `wire:navigate="$navigate-/url/to/page"` or I don't know, but that would still expand the API too much.

@calebporzio what do you think, is this feature worth it without targeting specific links? If it only works with targets, should it be component based like it is right now in this PR, or URL based with the new `livewire:navigate` event?

patriktoth67:
@calebporzio The feature would show a loading indicator whenever we are redirecting to a new page which takes a bit longer to load. Right now only the progress bar shows, but with this we could show any type of loading indicator.

calebporzio:
Thanks @patriktoth67 - I understand that. I'm more curious about your exact use-case. Is it simply to show a page loading indicator of some kind instead of the progress bar? Or to keep an existing loading indicator alive if there is a redirect?

Anyhow, either way, I'm going to close this with the following suggestion:

```php
<script data-navigate-once>
document.addEventListener('livewire:navigate', () => {
    // Toggle your indicator on...
})

    // Toggle it off...
})
</script>
```

That should accomplish what you want.

Thank you for all your work and consideration on this, but I don't see an obvious feature to add that doesn't have other drawbacks and complexity. If there's more community demand we can revisit the problem.

As for the other problem mentioned in the discussion (Livewire stopping loading indicators when redirecting) that's another story that I'd be open to a failing test PR even about.

Thanks!

PR #8175, opened 2024-03-20
Make wire:navigate.hover cache temporary

stancl:
This PR is a sample implementation of https://github.com/livewire/livewire/discussions/7943. I'm opening this as a draft PR just to move the work forward.

Essentially, the `wire:navigate.hover` cache should be temporary instead of permanent like it is now. Permanent makes sense for applications like the Livewire docs, but not for many applications with rapidly changing data. There may be a difference between the state of a page 1 minute ago and now, and I want the user to see the up-to-date version. Which is why for instance in the app I'm working on now, I'm adjusting the prefetch duration to 10 seconds.

However, the code for monkeypatching it in is extremely ugly and involves doing unholy things with `Object.prototype`, so I'd rather have this become a first-party feature.

If this was merged as-is, it'd probably be better than this not being a thing at all, but I'd still like to address a few things:
- Components should be able to configure whether they're cached forever, or only for some amount of time (or maybe if at all?)
- There should ideally be some tests for this

The timeout value I used in the diff (10 seconds) works great for me, but probably not for many other apps. A single global setting controlling this duration would work okay, but I don't think you want JS-level configuration. And making this component-specific would involve adding more metadata to the payload. So I only submitted this simple diff until I hear how you'd like to proceed.


PerryvanderMeer:
Depending on the timeout, this might also solve problems such as https://github.com/livewire/livewire/discussions/8444.

muhammadsaeedparacha:
oh man. This is such an important feature. Wonder why this hasn't been made included in livewire officially.
Inertia v2 introduced ability to control cache time. 

That would be amazing to have in livewire

gijsbeijer:
Yeah we need this to be configurable or even disableable!

calebporzio:
@joshhanley  - let's make this the same as Inertia's default and merge it. thx

kachelle:
Great new feature coming up!

Is there a possibility to add a global parameter to change the default to for example 5 seconds for rapid changing applications?

PR #8187, opened 2024-03-21
[3.x] Better support for `BackedEnum` while testing

PerryvanderMeer:
The current approach for setting a `BackedEnum` while testing feels a little bit odd.

```php
Livewire::test(Component::class)
    ->set('property', Enum::FOO->value)
    ->assertSetStrict('property', Enum::FOO);
```

Notice the required `->value` while setting. Without, it will throw:
> TypeError: ...\Enum::from(): Argument #1 ($value) must be of type string|int, ...\Enum given

Better approach after this PR:

```php
Livewire::test(Component::class)
    ->set('property', Enum::FOO)
    ->assertSetStrict('property', Enum::FOO);
```


calebporzio:
Thanks for this, makes sense. I moved the logic into the testing code instead of muddying up the synth with testing concerns.

PR #8188, opened 2024-03-21
Re-enable Laravel 11 tests

PerryvanderMeer:
The Laravel 11 tests were disabled in 8e76c1a63f0093bbbb9c0d07a6f0f926a2d927d5, but since L11 seems quite stable right now, it's probably better to re-enable them.


calebporzio:
Good call. thank you

PR #8195, opened 2024-03-22
Update events.md

danie-ramdhani:
Based on discussion: https://github.com/livewire/livewire/discussions/8193 https://github.com/livewire/livewire/discussions/7795

Apparently dispatching event from `@script` immediately is not working if the code is not wrapped inside `$nextTick()`

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -135,7 +135,9 @@ Additionally, you can dispatch events from within a component's `@script` like s
 ```html
 @script
 <script>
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the report, I dug into this and fixed the core to make it work without calling `$nextTick`. Here's the PR: https://github.com/livewire/livewire/pull/8226

Going to close this, thanks again.

PR #8197, opened 2024-03-22
Fix - Handling query params without values

davidrushton:
I believe that #7868 introduced a bug by removing the following line of code in the history `fromQueryString` method:

```js
entries.forEach(([key, value]) => {
    if (!value)
      return;
    ...
```

I use a URL of `dashboard?wizard` to signify a state (i.e. query param without value), but this is causing a JS error which then breaks all of the page's Livewire functionality:

```
livewire.js?id=5d8beb2e:7881 Uncaught TypeError: Cannot read properties of undefined (reading 'replaceAll')
    at livewire.js?id=5d8beb2e:7881:40
    at Array.forEach (<anonymous>)
    at fromQueryString (livewire.js?id=5d8beb2e:7880:13)
    at has (livewire.js?id=5d8beb2e:7812:21)
    at track2 (livewire.js?id=5d8beb2e:7737:35)
    at livewire.js?id=5d8beb2e:8753:53
    at Array.forEach (<anonymous>)
    at Array.<anonymous> (livewire.js?id=5d8beb2e:8748:33)
    at trigger2 (livewire.js?id=5d8beb2e:3857:34)
    at Component.processEffects (livewire.js?id=5d8beb2e:4478:7)
```

To fix, this PR adds the following code within that closure:

```js
entries.forEach(([key, value]) => {
    // Query string params don't always have values... (`?foo=`)
    if ( typeof value == 'undefined' ) return;
    ....
```

This is my first JS PR to Livewire so apologies in advance if my `npm build` steps from https://livewire.laravel.com/docs/contribution-guide weren't 100% right.

Thanks 🙂


joshhanley:
@davidrushton thanks for the PR! Could you please add a test for this? Thanks!

davidrushton:
@joshhanley I've added a browser test now.

I had to split the `create` method in `DuskTestable` into a new `createBrowser` method as the current `Arr::query($queryParams)` handling didn't allow a URL such as `'/livewire-dusk/'.$id.'?flag'`.

The test fails on the master branch but passes on this PR branch.

calebporzio:
Thank you for the fix! Well done.

PR #8212, opened 2024-03-25
Add failing test for prefix config

assertchris:
Unfortunately, the testing setup doesn't appear to cover `GenerateSignedUploadUrl::forS3` and it seems impossible to test this (aside from the conditions that force a mock storage disk) without using S3 credentials. Please reach out to me and I can provide you with a set of credentials (to use locally), to reproduce this failing test result for yourself.

cgpitt@gmail.com + twitter.com/assertchris

Edit:
The error is happening because https://github.com/livewire/livewire/blob/83392ce47c0d53ca4383ebba7480486b44272594/src/Features/SupportFileUploads/GenerateSignedUploadUrl.php#L26 assumes `client` is defined, but when using prefix config; the adapter is nested. So, instead of `adapter->client` we need to reach for `adapter->adapter->client`. Unclear what other scenarios make this nesting deeper, so a hack I have used to get around this error is to recursively walk the adapter tree until I do find `client`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,9 @@
     "orchestra/testbench": "8.20.0|^9.0",
     "orchestra/testbench-dusk": "8.20.0|^9.0",
     "calebporzio/sushi": "^2.1",
-    "laravel/prompts": "^0.1.6"
+    "laravel/prompts": "^0.1.6",
+    "league/flysystem-aws-s3-v3": "^3.0",
+    "league/flysystem-path-prefixing": "^3.3"
   },
   "autoload": {
     "files": [
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -22,5 +22,14 @@
     <php>
         <env name="CACHE_STORE" value="file"/>
         <env name="SESSION_DRIVER" value="file"/>
+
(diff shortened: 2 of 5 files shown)


calebporzio:
Oh dang, yeah this stuff is not easy/possible to test, so thanks for taking it on @assertchris.

Do you have any suggestions? It's hard for me to get my brain fully into the remove file upload context without putting in a ton of work, so if you have a fix suggestion I'm all ears. 

Is there a better/simpler way than recursing through the adapter?

If not, maybe that's fine as long as there's some kind of guaranteed escape hatch in the recursion.

Thanks for your effort on this.

Btw, what introduced this bug? Is this is a recent introduction? If so, can we revert what caused it instead?

calebporzio:
Ok, closing this because inactivity. If someone wants to re-surface this issue, please re-submit this PR or a similar one. Thanks!

assertchris:
Heya. Sorry for not getting back to this sooner, work has been hectic. 😅 I have bypassed this for now, with a custom config option and extra path work at the upload code site. Not ideal, but I don't know more about the issue than when I posted so I can't suggest a better solution than the one I mentioned.

PR #8221, opened 2024-03-27
Update wire-loading.md

danie-ramdhani:
It missing comma like in this docs https://livewire.laravel.com/docs/wire-loading#targeting-multiple-actions


calebporzio:
Oooops! Thank you!

PR #8222, opened 2024-03-28
Add support for Validate message arrays

TWithers:
Rather than doing this:
```php
#[Validate('required', as: 'Input A', message: 'Input A is required')]
#[Validate('max:6', as: 'Input A', message: 'Input A should be no more than 6 characters')]
#[Validate('min:2', as: 'Input A', message: 'Input A should be at least 2 characters')]
public $inputA;
```

It would be nice to do this:
```php
#[Validate('required|max:6|min:2', as: 'Input A', message: ['required' => 'Input A is required', 'min' => 'Input A should be at least 2 characters', 'max' => 'Input A should be no more than 6 characters'])]
public $inputA;
```
Currently, if you attempt to do this with multiple properties, each message gets applied to the global validation message since it isn't prefixed. This makes it look like a bug at first glance. Here is a wirebox showcasing the problem/feature: https://wirebox.app/b/4oppg

This PR would allow the messages to be passed as an array and have their property prefix added on.

### DIFF
--- a/src/Features/SupportValidation/BaseValidate.php
+++ b/src/Features/SupportValidation/BaseValidate.php
@@ -63,9 +63,20 @@ class BaseValidate extends LivewireAttribute
 
         if ($this->message) {
             if (is_array($this->message)) {
+                if (count($rules) === 1) {
+                    $messages = [];
+                    foreach ($this->message as $key => $value) {
+                        if (! str_contains($key, '.')) {
+                            $messages[$name.'.'.$key] = $value;
+                        }
+                    }
+                } else {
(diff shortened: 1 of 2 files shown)


calebporzio:
Thank you for the contribution @TWithers, however, I don't want to make the implementation of `#[Validate]` any more complex than it already is. Especially because at this point, people can use the `rules()` method to define validation rules in bulk like this. Thank you, but closing this.

I won't see a reply to this likely, so if you disagree or have other ideas, you can re-open this or submit a new one.

Thanks!

PR #8226, opened 2024-03-28
Prioritize component listener registration

calebporzio:
## The setup

Let's say you wanted your Livewire component to dispatch an event to itself. You could do this:

```php
{
    public function mount()
    {
        $this->dispatchSelf('trigger');
    }

    #[On('trigger')]
    public function received()
    {
        dd('Event received!');
    }
}
```

When the page loads, the event will be dispatched, picked up by the same component, a request will be sent for the `received` action, and "Event received!" will show up on the page; all good.

## The problem

There are other scenarios where code similar to above doesn't work as expected. For example, take a look at the same component re-written to dispatch the event from the frontend using a `@script` directive:

```php
{
    #[On('trigger')]
    public function received()
    {
        dd('Event received!');
    }
}
```

```html
<div>
    <!-- ... -->
</div>

@script
<script>
    $wire.dispatchSelf('trigger')
</script>
@endscript
```

The above example does NOT work. Here's why...

## The explanation

To understand the problem, it is helpful to understand the "order of events" of a new Livewire component initializing on the page.

Here's an ordered list of relevant mechanisms in Livewire's JavaScript lifecycle:
1. `@asset` -> Blade assets are included and loaded
2. Morph html (if it's a subsequent request)
3. `@script` -> Blade scripts are executed
4. `$this->js(...);` -> One-off JavaScript snippets are executed
5. `#[On(...)]`-> Listeners are registered
6. `$this->dispatch(...)` -> Events are dispatched
* `x-init` -> Alpine directives like `x-init` are initialized
* `x-on:something` -> Alpine listeners are registered

As you can see, registering component event listeners is #5 in the above list.

The first example (calling `$this->dispatch(...)` from `mount()`) works because dispatches from the backend are processed on step #6, which is AFTER step #5.

Therefore, the first example will work because the listener is registered BEFORE the dispatch.

However, the failing example (calling `$wire.dispatch(...)` from `@script`) will fail because `@script` directives are processed on step #3, which is BEFORE the listener is registered in step #5.

### Solution A) Wrapping in `$nextTick(...)`
One workaround to this problem, would be to wrap the `$wire.dispatch()` call in something like `$nextTick()` or `setTimeout()` like so:

```html
@script
<script>
    setTimeout(() => {
        $wire.dispatchSelf('trigger')
    })
</script>
@endscript
```

By wrapping the dispatch in a `setTimeout()`, everything will work because the listener will have a chance to register well before `$wire.dispatchSelf()` is called.

**Pros:**
* Doesn't require any other changes

**Cons:**
* It doesn't "fix" the problem, just works around it

**Questions:**
* Is this behavior a "problem" that needs fixing? Or a side-effect of Livewire's ideal behavior that is unideal in certain circumstances?
* Are there other scenarios where the current behavior is ideal?

### Solution B) Prioritizing listener registration

It seems like registering component listeners should be one of the first things a Livewire component does.

Maybe we can tweak the codebase to register any event listeners BEFORE executing `@script` or calling dispatches.

This way you could freely dispatch from `@script` and it would be received by Livewire components without hiccups.

**Pros:**
* Fixes the original problem described

**Cons:**
* Requires a fundamental lifecycle change for Livewire. I don't see any problems currently especially if the test suite passes, however, it's possible this causes other unintended consequences for people

## The conclusion

I think **Solution B** is the clear winner here. I've separated out listener registration and dispatching in the codebase so that we can run listeners before everything else.


LordVeggie:
Hello just checking if this has been approved/working.

PR #8231, opened 2024-03-29
Fixed undefined key with unset method of DataStore

dmyers:
Ran into a new exception today after upgrading my app to the latest version.

The exception is `Undefined array key "computedProperties"` which I was able to trace back to #8048 and that there needed to be another `isset()` check on the `$key`.

The issue I was able to replicate that this fixes is with a legacy computed property (ex: `getUserProperty()`) which is using an event attribute `ex: [On('someEvent')]` and I guess it isn't in the data store for that kind of flow.

I updated the test added in #8048 and without the change I've made would fail, but continues to pass with the change, supporting both cases.

### DIFF
--- a/src/Features/SupportComputed/UnitTest.php
+++ b/src/Features/SupportComputed/UnitTest.php
@@ -8,6 +8,7 @@ use Tests\TestCase;
 use Livewire\Livewire;
 use Livewire\Component;
 use Livewire\Attributes\Computed;
+use Livewire\Features\SupportEvents\BaseOn;
 
 class UnitTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


joshhanley:
@dmyers thanks for the PR! Please don't modify the existing tests, just add new ones. What I was testing for was a different scenario. If I disable the changes from PR #8048, my test which you modified still passes which is incorrect.

Also don't make unrelated changes like removing the render methods, as we try to keep a PR focused just on the required change. You're welcome to submit a clean up PR if you feel it's necessary (but to me having the render methods are fine).

I have reverted those two changes.

dmyers:
@joshhanley Great feedback. I appreciate your help with that, sorry for the confusion. I'm new to the Livewire internals and was trying a few different approaches to getting a solid test set up for this.

joshhanley:
@dmyers no worries! I tested your changes, the test fails without it and passes with it, so all looks good to me 🙂

joshhanley (approved):

calebporzio:
Thanks!

PR #8235, opened 2024-03-31
Add failing test for array update hooks `$value`

joshhanley:
This PR adds a failing test for handling of update hooks for array properties (and possibly others).

The test is structured the way that V2 handles update hooks for arrays.

See discussions #7101 and #6665.

## The Scenario
When properties are updated on the front end, Livewire triggers `updatingProperty` and `updatedProperty` hooks on the backend (plus updating/ updated but they aren't needed for this example).

This allows developers to hook into the updates and perform other actions or side effects based on the values in the update hooks.

In particular, the `updatingProperty` hook allows your to inspect the new property value before the property has been updated.

When using hooks in V2, the `$value` passed into the hook was always the new value that a property is going to be set to and there was no `$key` passed in.

For example if an array gets updated by the front end (such as when using entangle), then that value is passed to the update hooks as the `$value` prop.

So if you had a property with an array `['one', 'two', 'three', 'four']` and it gets changed on the front end, to `['two', 'three']`, then the new value would be sent to the backend and the `$value` prop on the update hooks would contain `['two', 'three']`. 

Pasted image 20240331153421

It is to be noted that each of the updatingProperty/updatedProperty hooks would only be called once each for a single property change.

See below screen capture of this in action in V2.

Recording 2024-03-31 at 14 53 55

## The Problem
The problem is that in V3 the way the front end handles updating properties has changed.

Instead of sending the new value of a property to the backend, Livewire will now calculate the changes (diff) to a property (particularly when they are arrays/objects) and send each of the diffs as individual update calls.

Pasted image 20240331153447

What this means, is that for every update that happens to a property, each of the updating/updated hooks will be called.

In the screenshot above, this means for the `items` property `updatingItems` and `updatedItems` hooks will each be called 4 times.

Each time the update hooks are called the `$value` that is being passed in is the values in each of the updates, so for the first hooks call it's `"two"`, then second is `"three"`, and then `"__rm__"` and then `"__rm__"`. And the `$key` is now populated with the corresponding index of the update `1`, then `2`, then `3`, then `4`.

For users that were relying on this functionality in V2 this is an unexpected change.

It also means the internal `__rm__` value is being exposed, meaning a developer has to interact with it and decide what to do with it, which is not ideal.

See below screen capture of this in action in V3.

Recording 2024-03-31 at 14 58 41

### Solution A) Change the front end
We could go back to how the front end handles a property update when it gets sent to the server like in V2.

**Pros**
- Makes dealing with it on the backend really simple as there will only be one update for a property, so the appropriate hooks can be run once.

**Cons**
- Caleb has previously said that calculating the diffs on the front end is a better idea (for personal curiosity reasons - I'd love to know why that is the case 😁).
- Could require signficiant rework of the front end to get that to happen.

### Solution B) Only use the diff to determine if a request should be sent, but use the whole value for the update
This depends on the reasoning for the diff in the first place. My suspicion is that calculating the diffs is helpful for determining whether a request should even be sent at all. But we could still do that for handling of the request, but then send the new value as a single update call (not push the diff into the updates).

**Pros**
- In theory seems simpler than reworking all of the front end.

**Cons**
- May not be helpful if there is actually benefit to sending the diffs as part of the request to the backend.

### Solution C) Aggregate the changes on the backend
We could group all the changes for a particular property together on the backend and make the update to the property with the final value of the property - thus calling the update hooks once with the final value.

**Pros**
- Means a single set of update hooks will be run per property change.

**Cons**
- The code for this might not be great to maintain.
- What happens if the list of updates also has updated for properties nested under this property?

### Solution D) Do nothing
These changes have been in effect for a while now. So we could just leave it the way it is.

**Pros**
- Easy as there is nothing to do.

**Cons**
- Multiple update hooks run, which could lead to performance issues if it is a large array.
- `__rm__` is still exposed to the developer.

## Recomendation
I think if I were to choose, I would probably lean towards solution B, as it could be the best of both worlds. I hold this opinion loosly though as I don't have the full context on why diffs are good.

### DIFF
--- a/src/Mechanisms/HandleComponents/BrowserTest.php
+++ b/src/Mechanisms/HandleComponents/BrowserTest.php
@@ -149,6 +149,45 @@ class BrowserTest extends \Tests\BrowserTestCase
         ->assertSeeIn('@selected', 'D')
         ;
     }
+
+    /** @test */
+    public function it_does_not_show_rm_removal_value_in_update_hooks()
+    {
+        Livewire::visit(new class extends Component {
+            public $updates = [];
+            
+            public array $items = ['one', 'two', 'three', 'four'];
+
+            public function updatedItems($value, $key)
+            {
+                $this->updates[] = [
+                    'key' => $key,
+                    'value' => $value,
+                ];
+            }
+
+            public function render()
+            {
+                return <<<'HTML'
+                    <div>
+                        <div>Items: <span dusk="items">{{ json_encode($items) }}</span></div>
+                        <div>Updates: <span dusk="updates">{{ json_encode($updates) }}</span></div>
+
+                        <div x-data="{ values: [] }" x-modelable="values" wire:model.live="items">
+                            <button x-on:click="values = ['two', 'three']" dusk="changeArray">Change array</button>
+                        </div>
+                    </div>
+                HTML;
+            }
+        })
+            ->assertSeeIn('@items', '["one","two","three","four"]')
+            ->assertSeeIn('@updates', '[]')
+            
+            ->waitForLivewire()->click('@changeArray')
+            ->assertSeeIn('@items', '["two","three"]')
+            ->assertSeeIn('@updates', '[{"key":null,"value":["two","three"]}]')
+            ;
+    }
 }
 
 enum Suit: string


ITSYV:
Is this failing test related to this case as well (checking differences on updates): https://github.com/livewire/livewire/discussions/6905 ?

joshhanley:
@ITSYV nope that looks like a separate problem to me.

RenoLooijmans:
I am encountering this particular situation in some of my projects, which I recently migrated to Livewire v3. I also think option B would be the best way forward.

Would love to see a fix for the current implementation. The returned "__rm__"-value is unusable in the backend.

RenoLooijmans:
This bug is still present in 3.5.20. What's the way forward..?

RenoLooijmans:
I guess chances are getting slimmer by the day to get this resolved. Too bad, in my opinion this really is an undesired side-effect of Livewire V3.

GregoireGaonach:
Hi everyone,

Like always, thanks for the work of the maintainers.

Like many developers have mention (#7101, #6665, etc.), two years on, the issue is still here. Reading your publication, option B seems to be the most sensible.

As you mentioned @joshhanley; and maybe I am missing something too, but I cannot find a good use case for wanting to catch N events for one variable update.

For example, if I click on the checkbox #6, the public function `updatedFilters` will be fired six times with six times the same payload, which is weird behavior:
```blade
$filters = [1,2,3,4,5];
---
<input type="checkbox" wire:model="filters" value="6">
```

Is there any chance this could be further discussed? Can we expect changes in Livewire 4?

joshhanley:
Closing this as it has been fixed in #9691 which will be in v4.

PR #8238, opened 2024-04-01
Add browser test to handle `x-for` and `x-if` state inside `@teleport` when request happened

danie-ramdhani:
This PR is failing test to handle `x-for` and `x-if` state inside `@teleport` when request happened.

Scenario:

Do something that need to morph the livewire component (i.e. `wire:click="$refresh"`)

Problem:

- `Alpine Expression Error: item is not defined` in the first request
- Duplicate elements occurred error second request or more

This can be resolved with `wire:ignore` with div element as root template tag.

```blade
<div wire:ignore>
    <template x-if="open">
        <div>x-if open</div>
    </template>
</div>
```

```blade
<div wire:ignore>
    <template x-for="item in 3" :key="item">
        <span x-text="item"></span>
    </template>
</div>
```

If so, we need add this to documentation (i.e. use `@foreach` instead of `<template x-for>` inside `@teleport`).

### DIFF
--- a/src/Features/SupportTeleporting/BrowserTest.php
+++ b/src/Features/SupportTeleporting/BrowserTest.php
@@ -89,4 +89,48 @@ class BrowserTest extends BrowserTestCase
             ->assertSeeIn('@first-check', '2')
             ->assertSeeIn('@second-check', '2');
     }
+
+    /** @test */
+    public function can_handle_xfor_and_xif_state_when_request_happened()
+    {
+        Livewire::visit(new class extends Component
+        {
+            public function render()
+            {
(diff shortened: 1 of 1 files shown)


calebporzio:
Thank you for this report. Phew, that was a deep one. There were two bugs related to `x-teleport` in Alpine.

The fixes aren't actually in this repo at all, so I'm going to close this. Here are the two PR's in the Alpine repo resulting from this PR:
* https://github.com/alpinejs/alpine/pull/4169
* https://github.com/alpinejs/alpine/pull/4168

PR #8242, opened 2024-04-02
Fix listening multiple event using `@` directive attribute from child component

danie-ramdhani:
Scenario:

- livewire has child component that has multiple events using `@` directive (i.e. `<livewire:child @foo="foo" @bar="bar" />`)
- dispatching bar event is success but dispatching foo event to parent is failed.

Here's screenshot for more info (looks like the event is overwritten):

image


lrljoe:
Yeah, it's only binding the "last" event item ultimately.

A bit of a niche one, as I'd typically just dispatch the event to the parent from client-side anyway, but I'm sure someone will have a quick solution.

In the meantime, you can of course just use the "On" attribute which will work
```
use Livewire\Attributes\On;
```

danie-ramdhani:
is it breaking change? then, we should add explanation in the documentation that you only can bind one event and the last one will overwrite the previous event.
i made this failing test PR based on discussion but i can't remember that 😅 (only remember the issue).

PR #8254, opened 2024-04-03
[3.x] Clarify docs for validation utilities

PerryvanderMeer:
This PR contains a documentation-clarification for the issue described in https://github.com/livewire/livewire/discussions/8230.


lrljoe (changes requested):
Suggest "component " is replaced with "class", or "component or form" as the same applies to both.

Also suggest that "should **not**" is replaced with "must **not**"

calebporzio:
Thanks for this!

robsontenorio:
@PerryvanderMeer It also worth it to add that you cannot have a `messages()` method and related one. Faced this issue recently.

PerryvanderMeer:
@robsontenorio Got ya, #8348

PR #8257, opened 2024-04-04
Add test for lazy component property error

oddvalue:
This PR introduces a test relating to issue #7945 [Lazy components with listener causes render before mounted](https://github.com/livewire/livewire/discussions/7945)

The test fails but not for the reason I would expect. The error can also be seen here https://wirebox.app/b/xgrpd

### DIFF
--- a/src/Features/SupportLazyLoading/BrowserTest.php
+++ b/src/Features/SupportLazyLoading/BrowserTest.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features\SupportLazyLoading;
 
+use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\Facades\Route;
 use Tests\BrowserTestCase;
 use Livewire\Livewire;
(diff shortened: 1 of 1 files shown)


danie-ramdhani:
you forgot to import the blade class 😅 and you forgot to use quote in render function ($ sign will make the variable echoed).

and using `assertDontSee('#livewire-error')` instead of `$this->expectNotToPerformAssertions();` maybe is the best choice to test something error.

Ok. here's your problem. You need to add the default value to typed props.

```php
public string $message = '';
```

I don't see the the problem here except that. If that is resolve your problem, you can close this...

oddvalue:
> Uhh... you made PR from main branch. (see https://livewire.laravel.com/docs/contribution-guide#3-preparing-your-pull-request-branch)
> 
> You forgot to import the blade class 😅 and you forgot to use quote (<<<'HTML' instead <<<HTML ) in render function ($ sign will make the variable echoed).
> 
> and using `assertDontSee('#livewire-error')` instead of `$this->expectNotToPerformAssertions();` maybe is the best choice to test something error.
> 
> Ok. here's your problem. You need to add the default value to typed props.
> 
> ```
> public string $message = '';
> ```
> 
> I don't see the the problem here except that. If that is resolve your problem, you can close this...

Right, but what if that property is some sort of object that needs to be instantiated in the mount method? Why does having a listener on a lazy component cause it to try to render the actual component, not the placeholder, before it has been mounted?

PR #8260, opened 2024-04-05
Fix the possible side effects when set_error_handler() is used

skeemer:
If we use _set_error_handler()_ in our code, it can pick up errors from _SupportMultipleRootElementDetection_ when it runs `@$dom->loadHTML($html);`

I was using this with some API calls and it threw exceptions for something unrelated. Using it this ways is probably not best, just for this reason, but shouldn't cause Livewire to blow up.
```php
set_error_handler(
    function ($severity, $message, $file, $line) {
        throw new \ErrorException($message, $severity, $severity, $file, $line);
    }
);
```

I tried to create a failing test, but couldn't get it to work. I verified that the error function was being run by adding a `die()` inside, but the thrown error wouldn't surface in the test and returned normally.

This is my attempted test.
```php
/** @test */
function dont_throw_error_when_set_error_handler_is_used()
{
    config()->set('app.debug', false);

    Livewire::test(new class extends Component {
        function mount()
        {
            set_error_handler(
                function ($severity, $message, $file, $line) {
                    throw new \ErrorException($message, $severity, $severity, $file, $line);
                }
            );
        }
        function render()
        {
            return <<<'HTML'
            <div>
                <section>
                    An html5 element
                </section>
            </div>
            HTML;
        }
    })->assertSuccessful()->assertSee('html5 element');
}
```


calebporzio:
Gotcha, this makes sense. Is it possible to add this test to the PR? Also maybe extracting this to a wrapper function? or just a code comment to explain why it's necessary? Thanks!

skeemer:
I pasted it because I couldn't get the test to fail. I added a die statement just prior to the throw and that did quit the test. So I know that an error is being thrown, but it seems to be getting squashed and running normally. I don't know if it is related to how the testing captures errors or something else.

skeemer:
Also, I found out about `restore_error_handler()`, so setting our own to capture and restoring might be safer so we don't accidentally clear out libxml errors that were expected. I'll switch it if you agree with that.

calebporzio:
Yeah, that sounds better. You can switch.

calebporzio:
Hey @skeemer, I'm closing this because of inactivity. Feel free to PR the update we talked about. Thanks

PR #8261, opened 2024-04-05
Fix `assertViewIs()` for components without `render()`

PerryvanderMeer:
## Problem
The `assertViewIs()` method won't work as expected for components without the `render()` method. For components without the `render()` method, `assertViewIs()` expects the full path to the view, instead of just the name of the view.

## Cause
Based on whether the component has the `render()` method, the view will be rendered directly or via Laravel's `View` facade. When rendering directly, `$view->name()` will be the actual name. When rendering via the `View` facade, the 'name' will be the full path to the view.

https://github.com/livewire/livewire/blob/5dd96cfa9acd15cf71488d81d9e75affff15ffb0/src/Mechanisms/HandleComponents/HandleComponents.php#L269-L271

Laravel's default `assertViewIs()` method tries to match the given value with the view's name. Obviously, when the name contains the full path, it won't pass.

## Solution
I've extended Livewire with a custom assertion, wherein the assertion depends on whether the component has the `render()` method or not.


src/Features/SupportTesting/MakesAssertions.php
```diff
@@ -164,4 +165,28 @@ public function assertReturned($value)
...
+
+            return $this;
+        }
+
+        // When the render method does not exist on the component, the view name will be the full path...
+        $fileName = str($value)->replace('.', '/')->__toString();
+
+        PHPUnit::assertThat(
+            $this->lastState->getView()->name(),
+            PHPUnit::equalTo(resource_path('views/').$fileName.'.blade.php'),
```
calebporzio, line 187: I haven't dug too deep here, but is there a better way than hardcoding `resource_path('views/')`? Some apps change their view directory, and there can also be multiple. They can also use a non `.blade.php` suffix (just `.php` in some cases). I know this is edge casey-stuff, but I just want to make sure there's not some kind of Laravel method somewhere or precedent that we can use here.

PerryvanderMeer: While no `render()` method is given, we assume the view is located in: 
```php
View::file($viewPath . '/' . $fileName . '.blade.php')
```
In that regard, I think it is fine to use `.blade.php` as suffix.

PerryvanderMeer: As for `resource_path('views/')`, I'm not sure. As long as you customize your view directory to something within `resources/views` it should be fine (as covered within the test). Not sure what happens when you cutomize to something totally different, but that might be a very-very edge case.

calebporzio:
How often does someone run "assertViewIs" for a Livewire component test anyhow? I don't, for some reason I'm hesitant to merge this. Might just leave this little odd behavior out there as I can't really imagine myself using it? Let me know your thoughts.

PerryvanderMeer:
@calebporzio Hmm yeah I do see myself using it from time to time and it caused some problems in the past (https://github.com/livewire/livewire/discussions/8023, https://github.com/livewire/livewire/discussions/7756, https://github.com/livewire/livewire/discussions/6840), but it's up to you!

calebporzio:
Ok, thanks for the references. It seems two the three discussions were fixed by a different PR. I'd say to this and the remaining one: It feels slightly odd to `assertViewIs` when you don't provide a `render()` method.

If this comes back up and others think I'm wrong, we can re-evaluate and maybe merge this.

Thanks for your efforts!

PerryvanderMeer:
I see. For completeness, if you want to solve this problem in your own application, you can easily do it with a `macro` or a `mixin`. You can register this in a service provider for example:
```php
use Illuminate\View\View;
use Livewire\Features\SupportTesting\Testable;
use PHPUnit\Framework\Assert as PHPUnit;

Testable::macro('assertViewIs', function (string $value)
{
    if (! $this->lastState->getView() instanceof View) {
        PHPUnit::fail('The response is not a view.');
    }

    // When the render method exists on the component, the view name will be the actual name...
    if (method_exists($this->lastState->getComponent(), 'render')) {
        PHPUnit::assertEquals($value, $this->lastState->getView()->name());

        return $this;
    }

    // When the render method does not exist on the component, the view name will be the full path...
    $fileName = str($value)->replace('.', '/')->__toString();

    PHPUnit::assertThat(
        $this->lastState->getView()->name(),
        PHPUnit::equalTo(resource_path('views/').$fileName.'.blade.php'),
    );

    return $this;
});

```

PR #8262, opened 2024-04-05
[3.x] Fix persistance of validation errors in forms

PerryvanderMeer:
This PR fixes the problem described in discussion https://github.com/livewire/livewire/discussions/8168.

## Problem
When validating a single property in a form, validation errors for other properties won't persist and will disappear.

## Cause
Validating forms differs slightly from validating normal properties. When validating normal properties, we merge in existing validation errors:

https://github.com/livewire/livewire/blob/5dd96cfa9acd15cf71488d81d9e75affff15ffb0/src/Features/SupportValidation/HandlesValidation.php#L408-L410

When validating properties in a form, we do not.

## Solution
We can merge in existing validation errors when validating form properties via `validateOnly()`.


calebporzio:
This is perfect - good catch! Thanks!

PR #8264, opened 2024-04-05
Add support to invade `Testable`

PerryvanderMeer:
Every now and then I involve private and protected functions and properties in my tests.
It can be challenging to access those while testing, although `invade()` is the ideal solution.

This PR adds support for accessing protected and private functions and properties while testing via `->invade()`.

```php
$component = Livewire::test(new class extends Component {
    protected string $foo = 'bar';
});

$component->invade()->foo; // bar
```

I have this by default in my projects via a `Mixin`, but it may be useful to others as well.

### DIFF
--- a/src/Features/SupportTesting/Testable.php
+++ b/src/Features/SupportTesting/Testable.php
@@ -236,6 +236,11 @@ class Testable
         return $this->lastState->getComponent();
     }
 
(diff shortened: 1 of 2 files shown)


calebporzio:
This is pretty cool!

PR #8267, opened 2024-04-05
macro can be callable from view

AlexandreBonaventure:
Fixing a bug reproduced here: https://wirebox.app/b/4m82r

## Problem 
Since component macros are supported, one would expect that you can call the macro action/method from either the component methods or directly from the view (eg: with `wire:click="macroedMethod"`)
The latter usage throws an error and kinda defeat the purpose of using macros

image

This MR fixes the issue by adding a conditional check on the callMethods to account for macros there.

### DIFF
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -459,12 +459,16 @@ class HandleComponents extends Mechanism
             // @todo: put this in a better place:
             $methods[] = '__dispatch';
 
-            if (! in_array($method, $methods)) {
+            if (! in_array($method, $methods) && !$root::class::hasMacro($method)) {
                 throw new MethodNotFoundException($method);
             }
 
             if (config('app.debug')) $start = microtime(true);
-            $return = wrap($root)->{$method}(...$params);
+            if ($root::class::hasMacro($method)) {
+                $return = $root->macroCall($method, $params);
+            } else {
+                $return = wrap($root)->{$method}(...$params);
+            }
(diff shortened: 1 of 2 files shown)


calebporzio:
Hmmmm... At this point, you'd still have to enable the macro with the `->macroedMethod()` call, and I'd say you might as well put that macro inside a trait which would read a little better I think and not add an extra API. Going to close this, interesting idea though!

AlexandreBonaventure:
Hmm, I'm not sure if I follow you entirely, maybe I can explain a bit better:
The goal here is to use macro to provide a global api for every component as demonstrated here: https://fly.io/laravel-bytes/global-notifications-with-livewire/.  Coming from Vue, it is a very common pattern to add this kind of global functionality through global properties https://vuejs.org/guide/reusability/plugins#writing-a-plugin

I understand that trait is the most standard way to extend component functionality, but IMHO it fall short for a few specific use cases, namely global functionality that needs be accessible from anywhere in the app, even within nested blade views. eg: I have a blade layout that is being reused across many different routes I need to trigger actions from this view. If I use trait, not only I would need to make sure to import it on every route component, but more importantly it's also very brittle because this button can fail to trigger the action when switching route, if the trait was omitted. In other words, it makes the action tigthly coupled to the view's component, if you need to refactor, say use the layout in another nested component instead of top-level, then you need to make sure you also move the trait over. Tracking usage of such global functionality through the app is cumbersome and prone to breakage, especially when used frequently. 

I am fairly new to Livewire so maybe there are some better alternatives but it seems like using macro was the perfect way to do this kind of things. 

All the other ideas I have to workaround that also have issues: custom livewire directive etc.. 

I hope to hear back from you, and that maybe you reconsider if that makes sense :)

PR #8215, opened 2024-03-26
POC: add call and exception in lifecycle hooks

AlexandreBonaventure:
Discussion Here: https://github.com/livewire/livewire/discussions/8214

Let me know if this was intentionally not surfaced, if not, I can update this MR with tests!

PR #8268, opened 2024-04-05
Missing trait hooks

AlexandreBonaventure:
Follow up on this one: https://github.com/livewire/livewire/pull/8215


calebporzio:
Thank you for finishing this up @AlexandreBonaventure !

francoism90:
@AlexandreBonaventure Just an question about this commit, hope that's ok. :)

Is it also possible to add the exception inside the trait as well (e.g. `WithPost`), as can be seen in the examples later? I was thinking about catching model exceptions like 403/404, and than redirect to `/<-response-code>`.

AlexandreBonaventure:
@francoism90, sure that should work!

francoism90:
@AlexandreBonaventure that's awesome! 👍

PR #8269, opened 2024-04-06
Add failing test for wire loading works when target function was called by event

satulg01:

Yes, you can find the discussion here: https://github.com/livewire/livewire/discussions/8172

Yes, the branch is named `add-test-wire-loading-works-when-function-called-by-event`

No.

Yes.

I just create a test for a bug that i find.

Thanks for contributing! 🙌


src/Features/SupportWireLoading/BrowserTest.php
```diff
@@ -378,6 +381,64 @@ public function render()
...
+                    <div wire:loading wire:target="onFoo" dusk="loadingIndicator">
+                        Processing...
+                    </div>
+                </div>
+                HTML;
+            }
+        })
+        ->assertDontSee('Processing...')
+        ->waitForLivewire()->click('@button')
+        ->assertSee('Processing...');
```
danie-ramdhani, line 410:
```php
->click('@button')
->waitForText('Processing...')
->assertSee('Processing...');
```

src/Features/SupportWireLoading/BrowserTest.php
```diff
@@ -378,6 +381,64 @@ public function render()
...
+        });
+
+        $component->assertDontSee('Processing...')
+            ->press('@fooButton')
+            ->waitForText('Processing...')
+            ->assertSee('Processing...')
+            ->pause(3100)
+            ->assertDontSee('Processing...')
+            ;
+    }
```
danie-ramdhani, line 441:
```suggestion
```

I think you can delete this because it's not related with dispatching event (and this test is already exist)

satulg01: hi @danie-ramdhani, sorry for the ask but how that i build the livewire to test my changes made locally on js\directives\wire-loading.js ? I'v searched how to do this on Contribution Guide but not find. Sorry, i'm begginer on open source world's

src/Features/SupportWireLoading/BrowserTest.php
joshhanley: @danie-ramdhani FYI using `->waitForLivewire()` is fine and sometimes better as it resolves when the Livewire request actually finishes. I recently improved it to make it more stable too.

calebporzio:
Ok, you're right, this doesn't work and is somewhat expected behavior.

Currently, Livewire just sends the dispatched event name to the server and the server figures out who's listening.

To properly target a loading indicator like this, we would need the frontend to also know which method is being called by the dispatched event.
 
This is a fundamental shift from what we're currently doing and I'm hesitant to change that for a smaller feature like this.

Therefore, I'm going to close this until there is more demand and conversation around it (when that time comes, please re-submit this PR)

In the meantime, something like the following should work:
```
<div x-on:foo.window="$wire.onFoo()">
    <!-- ... -->
</div>
```

As you can see, instead of registering the listener in the backend, you can register it in the front-end and call the Livewire method manually. Then the loading indicator will react. Thanks!

PR #8282, opened 2024-04-10
Fix `@teleport` directive when view is cached

luanfreitasdev:

https://github.com/livewire/livewire/discussions/8130
Yes
No
No

When we run php artisan view:cache, the `@teleport` directive is broken. I didn't understand the reason in depth, however, when we create a directive similar to `@this` using the Facade Blade, the same problem does not occur.

- I noticed how it is called to create the directive for `@teleport` differs from `@this`. For this reason, I left it the same as `@this` and it worked.

- I was unable to create a test that runs view:cache and generates the cache for a dusk test at run time

### DIFF
--- a/src/Features/SupportTeleporting/SupportTeleporting.php
+++ b/src/Features/SupportTeleporting/SupportTeleporting.php
@@ -2,17 +2,18 @@
 
 namespace Livewire\Features\SupportTeleporting;
 
+use Illuminate\Support\Facades\Blade;
 use Livewire\ComponentHook;
 
 class SupportTeleporting extends ComponentHook
 {
     static function provide()
     {
-        app('livewire')->directive('teleport', function ($expression) {
+        Blade::directive('teleport', function ($expression) {
             return '<template x-teleport="<?php echo e('.$expression.'); ?>">';
         });
 
-        app('livewire')->directive('endteleport', function () {
+        Blade::directive('endteleport', function () {
             return '</template>';
         });
     }


PerryvanderMeer:
The problem is caused by the fact that the `@teleport` directive is registered via `app('livewire')->directive('teleport')`, which calls `app(ExtendBlade::class)->livewireOnlyDirective()` via the `LivewireManager`. So, the directive is only available while rendering via Livewire's own blade compiler. Which is obviously not used when compiling and caching the views via Laravel's compiler.

You can test if the `@teleport` directive is also available while caching the views via Laravel's compiler with:
```php
Blade::compileString(<<<'HTML'
    <div>
        @teleport('foo')
            bar
        @endteleport
    </div>
HTML);
```php

luanfreitasdev:
@PerryvanderMeer, do you believe that this PR should not continue?

PerryvanderMeer:
Well it solves the problem, but I'm not sure why @calebporzio initially limited the scope of the directive to Livewire's blade compiler.
Anyhow, you could extend your PR with a test based on the example code that I've given above.

luanfreitasdev:
This doesn't seem to work very well with testing. I also tried creating a component with a blade file and clearing the cache using `Artisan::call('view:cache')` and the problem didn't happen.

calebporzio:
Yeah, this is a problem with using `app('livewire')->directive()`. I initially introduced "livewire only" blade directives so that Livewire was less-invasive. But view caching kinda ruins that. I'm going to merge this and we should probably remove the concept of livewire-only Blade directives entirely. (This is the same problem we experience with the morph marker injection not happening when views are cached)

MrPunyapal:
All models and slideovers which were in @teleport() behaving odd after this update!

They use to stay on screen on subsequent requests now they disappear after a subsequent request!

joshhanley:
@MrPunyapal Can you make a https://wirebox.app that demonstrates the issue?

MrPunyapal:
somehow it is not working for me
I can create repo if you want?

joshhanley:
@MrPunyapal bugger, thanks! will see if I can get it fixed.

joshhanley:
@MrPunyapal wirebox should be fixed now 🙂

MrPunyapal:
> @MrPunyapal wirebox should be fixed now 🙂

thanks @joshhanley 

see the bug

3.4.10: https://wirebox.app/b/xeo72

3.4.11: https://wirebox.app/b/430dl

open the modal and click on add random
both codes are exact the same!

PerryvanderMeer:
@MrPunyapal Your 3.4.10 Wirebox is private, so we can't access.

MrPunyapal:
> @MrPunyapal Your 3.4.10 Wirebox is private, so we can't access.

done

PerryvanderMeer:
Tested a few things. Blade rendering and Livewire's update response are completely the same. Must be a problem since Alpine 3.13.9.

MrPunyapal:
> Tested a few things. Blade rendering and Livewire's update response are completely the same. Must be a problem since Alpine 3.13.9. Either way not related to this PR.

FYI: without @-teleport it has the same behavior on both versions!

luanfreitasdev:
What is causing this is Alpine, not the Blade directive. I made a version of the old directive with the new Alpine and it also fails

https://wirebox.app/b/4qp32

MrPunyapal:
But it's 3.4.11?

luanfreitasdev:
Yes
image

luanfreitasdev:
Blade Directive with version 3.4.10 (works)

https://wirebox.app/b/x6djd

MrPunyapal:
> Blade Directive with version 3.4.10 (works)
> 
> https://wirebox.app/b/x6djd

Btw I forgot one thing!

I was having this in my app service provider!

I was using a blade directive from 3.3 only! 🙂

https://twitter.com/MrPunyapal/status/1746556766278897995

MrPunyapal:
Solved it by tweaking some code at all the teleports

jeffchown:
> Solved it by tweaking some code at all the teleports

@MrPunyapal I know it's been a while since your post, but could you share a before and after example of your 'tweaking' to get your @teleports working?  Thanks!

joshhanley:
@jeffchown there was a bug with teleports, it should be fixed now, so you shouldn't need to do anything. Also I don't think the user TeleportS should be tagged, make sure to always escape blade directives in code blocks 🙂

jeffchown:
Gotcha.  Thanks, @joshhanley !

PR #8286, opened 2024-04-11
Fix hydrating union types

PerryvanderMeer:
This PR is a follow-up on #8126, where some problems were introduced (see: https://github.com/livewire/livewire/pull/8126#issuecomment-2032876585).

## Problem
Component properties cast to a union type like `string|int` or `string|float` are broken.
For example:
- Setting `(string) 'foo'` to `string|int $property` is cast to `(int) 0`.
- Setting `(string) 'foo'` to `string|float $property` is cast to `(float) 0.00`.

## Cause
Any property (union) casted to `int` or `float`, was aggressively casted to respectively an integer or float.

https://github.com/livewire/livewire/blob/cac2f7b28f37980653aaec99d9de55de21b2ebaa/src/Mechanisms/HandleComponents/Synthesizers/IntSynth.php#L20

https://github.com/livewire/livewire/blob/cac2f7b28f37980653aaec99d9de55de21b2ebaa/src/Mechanisms/HandleComponents/Synthesizers/FloatSynth.php#L21

## Solution 
We don't need to aggressively cast the passed values to an integer or float.
Casting is also already partially handled by PHP itself.
After this PR is merged, casting works as follows. 

`int $property`
| Input | Output | Behaviour |
| -------- | ------- |------- |
| (null) | (null) | ✅ Unchanged |
| (int) 3 | (int) 3 | ✅ Unchanged |
| (float) 3.00 | (int) 3| ✅ Unchanged |
| (float) 3.14 | (int) 3 | ✅ Unchanged |
| (string) '3' | (int) 3| ✅ Unchanged |
| (string) '3.14' | (int) 3 | ✅ Unchanged |

`float $property`
| Input | Output |Behaviour |
| -------- | ------- |------- |
| (null) | (null) | ✅ Unchanged |
| (int) 3 | (float) 3.00 | ✅ Unchanged |
| (float) 3.00 | (float) 3.00 | ✅ Unchanged |
| (float) 3.14 | (float) 3.14 | ✅ Unchanged |
| (string) '3' | (float) 3.00 | ✅ Unchanged |
| (string) '3.14' | (float) 3.14 | ✅ Unchanged |

`int|string $property`
| Input | Output |Behaviour |
| -------- | ------- |------- |
| (null) | (null) |✅ Unchanged |
| (int) 3 | (int) 3 |✅ Unchanged |
| (float) 3.00 | (int) 3|✅ Unchanged |
| (float) 3.14 | (int) 3|✅ Unchanged |
| (string) '3' | (int) 3 |✅ Unchanged |
| (string) '3.00' | (int) 3 | ✅ Unchanged |
| (string) '3.14' | (string) '3.14' | ⚠️ Before: `(int) 3`, can cause issues |
| (string) 'foo' | (string) 'foo' |✅ Before: `(int) 0`, solves issue |

`float|string $property`
| Input | Output |Behaviour |
| -------- | ------- |------- |
| (null) | (null) |✅ Unchanged |
| (int) 3 | (float) 3.00 |✅ Unchanged |
| (float) 3.00 | (float) 3.00|✅ Unchanged |
| (float) 3.14 | (float) 3.14|✅ Unchanged |
| (string) '3' | (float) 3.00 |✅ Unchanged |
| (string) '3.00' | (float) 3.00 |✅ Unchanged |
| (string) '3.14' | (float) 3.14 |✅ Unchanged |
| (string) 'foo' | (string) 'foo' | ✅ Before: `(float) 0.00`, solves issue|

I've also added some backwards compability for the old situation. The only rare situation in which some problems might occur, is when properties are cast to `int|string` and are set to a string-formatted float. Previously, the string was cast to an integer, now the actual string is set.


calebporzio:
This is great. Thank you for the in-depth work, description, and tests. Love it.

PerryvanderMeer:
@calebporzio Thanks! Can you please tag a new release with this included? This issue prevents our apps from updating from version 3.4.9.

PR #8288, opened 2024-04-11
Improve performance for nested component updates

alexwass-lr:
When a component with nested children is updated, and the children have reactive properties, all the components are sent with the request to be re-rendered. 

In a normal scenario this is ideal, as we obviously want all our children to receive the updated property and re-render. 

However, this becomes a problem when the parent filters the children, for example a search field on a table. As each row (child) is sent back in the request, all rows are computed, even if the result would only show one row.

This is probably something that would go unnoticed in normal use. However this becomes exponentially problematic when you have 100's of rows, all taking multiple seconds to load. Using a sample dataset, locally our application can take around 8 seconds to load, but a search which should yield one result would take around 14 seconds (because it also bundled all the children not loaded outside the viewport). After this change, the result now renders in around 800ms.

See https://github.com/livewire/livewire/discussions/8186 for more info.


calebporzio:
Gotcha, yeah, this makes sense. Thanks for the submission.

PR #8291, opened 2024-04-12
Add failing test for checkboxes with non boolean value

ITSYV:

There is a discussion about it:
https://github.com/livewire/livewire/discussions/6905

As stated in the original docs (https://livewire.laravel.com/docs/wire-model#multiple-checkboxes), checkboxes should allow non-boolean values, however when working with wire:model and (un)checking checkboxes, values are changed to booleans.

2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed) Y

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out. N

4️⃣ Does it include tests? (Required) Y

Only failing test in this PR.
The test checks if there is a valid non-boolean result when checking a checkbox and if the checkbox is checked if the value is the same as the wire:model value.

Thanks for contributing! 🙌


src/Features/SupportDataBinding/BrowserTest.php
```diff
@@ -130,4 +131,43 @@ public function render(): string
...
+
+            public function render(): string
+            {
+                return <<<'HTML'
+                <div>
+                    <input dusk="checkbox" type="checkbox" value="checked" wire:model.live="element">
+                </div>
+                HTML;
+            }
+        })
```
danie-ramdhani, line 170: you need to separate this test in each function 🙃

src/Features/SupportDataBinding/BrowserTest.php
```diff
@@ -130,4 +131,43 @@ public function render(): string
...
+        {
+            public $element = 'checked';
+
+            public function render(): string
+            {
+                return <<<'HTML'
+                <div>
+                    <input dusk="checkbox" type="checkbox" value="checked" wire:model.live="element">
+                </div>
+                HTML;
```
danie-ramdhani, line 168: you forgot to display the property `<span>{{ $element }}</span>` 🙃

danie-ramdhani:
livewire is using alpine js for binding data, so i think it's not related to livewire (see: https://github.com/alpinejs/alpine/discussions/4018). i think, you can close this PR.

src/Features/SupportDataBinding/BrowserTest.php
ITSYV: It's not needed in the second test, as it tests if the checkbox itself is checked or not

ITSYV:
> livewire is using alpine js for binding data, so i think it's not related to livewire (see: [alpinejs/alpine#4018](https://github.com/alpinejs/alpine/discussions/4018)). i think, you can close this PR.

Ok, but the Livewire docs are still different than the actual results

src/Features/SupportDataBinding/BrowserTest.php
ITSYV: Separated  👍
Wasn't sure if this was needed or not

danie-ramdhani:
can you give me the discussion link?

ITSYV:
> can you give me the discussion link?

Original discussion: https://github.com/livewire/livewire/discussions/6905
The PR question from Josh: https://github.com/livewire/livewire/pull/8235#issuecomment-2030851768

danie-ramdhani:
mmm... yeah this kinda complicated because livewire v3 is using alpine js for binding the data 😅...

src/Features/SupportDataBinding/BrowserTest.php
danie-ramdhani: mmm... yeah it's using `->assertChecked('@checkbox')`
image

joshhanley:
@danie-ramdhani yeah having the test here is fine even if the fix ends up being in Alpine. It's good we have tests too as Livewire can sometimes have some edge cases that the Alpine tests don't pick up (due to morphing, etc.)

gingebaker:
I see that the JS Code is directly in alpineJs wich I posted in the discussion [#6905](https://github.com/livewire/livewire/discussions/6905#discussioncomment-8557937)

Alpine Codebase:
https://github.com/alpinejs/alpine/blob/56649308eb686caa90cf156881dc3e19ac48efbf/packages/alpinejs/src/utils/bind.js#L60
and the .checked attribute of the checkbox is only set directly when the value is boolean.

Although I do not know what that Code Comment means in the alpine codebase:
// If the value is a boolean/array/number/null/undefined, leave it alone, it will be set to "on" automatically.

Is there a change that this can be changed in alpine? Should we report an issue in Alpine?
Thank you

joshhanley:
Just adding a note here - this behaviour was originally set in Alpine to match how VueJS handles checkboxes. In VueJS setting `value` on a checkbox does nothing, and it still just returns `true` or `false`.

But I have recently discovered that Vue offers a way to customise the `true-value` and the `false-value` https://vuejs.org/guide/essentials/forms#checkbox-1. Maybe that's something that can be done here. Although personally I think we should just support the `value` attribute and handle it the same way the browser does. If there is no value, we do true/false, but if there is a value we do `value` for true and an empty string `""` for false (or it could still just be false, although that seems strange alongside the value).

calebporzio:
Closing this PR as it's been open for a long time. If still relevant, please re-submit, ideally with a solution if you can. Thank you!

PR #8297, opened 2024-04-13
Use `assertSetStrict` in tests

PerryvanderMeer:
This PR replaces some obsolete `assertSet('property', $value, true)` with the new `assertSetStrict('property', $value)`.


calebporzio:
Thanks!

PR #8298, opened 2024-04-13
Use `#[WithConfig]` in tests

PerryvanderMeer:
This PR replaces some obsolete `config()->set('foo', 'bar')` with Orchestra's `#[WithConfig]`.


calebporzio:
I personally prefer what we have even though there is a helper available. It just feels more "native Laravel" or "bare-metal" which I prefer for test setup type things. Thank you anyways though!

PR #8300, opened 2024-04-13
Use `#[Test]` in tests

PerryvanderMeer:
I'm certainly not a fan of such big PRs, but it wouldn't be improved by splitting it up. So here is an attempt. Postponing doesn't make it better either.... The only change is replacing the obsolete `/** @test */` with PHPUnit's `#[Test]`.


calebporzio:
Gotcha. Yeah, I'm not sure actually. The Laravel framework internals use the `test` method name prefix like:

```php
public function testHandlerIgnoresNotReportableExceptions()
```

The Laravel docs use the same, but with snake case:

```php
public function test_exception_is_thrown(): void
```

And the PHPUnit docs use the prefix as well:

```php
public function testCanBeCreatedFromValidEmail(): void
```

Therefore, I think we should use the prefix internally. We can keep the snake case method names for now, but if you're willing, please change this PR to be the prefix instead and I can merge.

Thanks for working on non-sexy core things!

calebporzio:
Going to close this. If someone wants to PR the change I suggested, feel free. Thanks!

PR #8314, opened 2024-04-17
Add failing test for `wire:loading.remove` not working with other `wire:` directives

JayBizzle:
Not sure if this is a bug or if I m doing something wrong. I have no idea how I would go about fixing it, so have created a failing test for this scenario.

We are using Wire Elements Pro and we have a couple of buttons in a modal like this...

```html
<a wire:loading.remove wire:foo class="tw-inline-block tw-w-full tw-px-5 tw-py-3 tw-text-sm tw-font-semibold tw-text-center tw-text-red-500 tw-border tw-border-red-400 tw-rounded-lg sm:tw-w-auto" href="#">
    Cancel
</a>
```

The reason I noticed this is because we are using Wire Elements Pro and that allows you to do this `wire:modal="close, {force: true}"`.

When that directive is on a button, along with `wire:loading.remove`, the button does not get removed.

I have created a simple example that can be viewed that shows this in action here https://wirebox.app/b/x6dq1

Both blue buttons should hide after the red button is clicked, but only one of them does.

Thanks

### DIFF
--- a/src/Features/SupportWireLoading/BrowserTest.php
+++ b/src/Features/SupportWireLoading/BrowserTest.php
@@ -79,6 +79,37 @@ class BrowserTest extends \Tests\BrowserTestCase
     }
 
     /** @test */
+    public function wire_loading_remove_works_with_other_wire_directives()
+    {
+        Livewire::visit(new class extends Component
+        {
+            public function doSomething()
+            {
+                // Need to delay the update so that Dusk can catch the loading state change in the DOM.
+                usleep(500000);
+            }
+
+            public function render()
+            {
+                return <<<'HTML'
+                    <div>
+                        <button wire:click="doSomething" dusk="button">
+                            <span wire:loading.remove wire:foo>Do something</span>
+                            <span wire:loading>...</span>
+                        </button>
+                    </div>
+                HTML;
+            }
+        })
+        ->waitForText('Do something')
+        ->click('@button')
+        ->waitForText('...')
+        ->assertDontSee('Do something')
+        ->waitForText('Do something')
+        ->assertDontSee('...');
+    }
+
+    /** @test */
     function wire_loading_remove_works_with_renderless_methods()
     {
         Livewire::visit(new class extends Component {


Alexxlml:
Hi @JayBizzle , if your other buttons need to have a wire action, wrap that element in a div and to that div assign wire:loading.remove, this will correctly hide your element since it is wrapped in a div.

I don't know if this is correct but it is a solution.
``` 
<div wire:loading.remove>
    <button wire:foo
        class="text-white bg-blue-700 hover:bg-blue-800 focus:ring-4 focus:ring-blue-300 font-medium rounded-lg text-sm px-5 py-2.5 me-2 mb-2 dark:bg-blue-600">
        I should hide, but won't
    </button>
</div>

**Comment by JayBizzle** [comment]
Thanks, I know how to work around it, but it seems unexpected we can't do it the way I have highlighted above, hence this "bug" report 👍🏻

**Comment by JayBizzle** [comment]
Just a couple of updates on this after diving in to the source code...

When and element has any other custom directive specified, this code...

https://github.com/livewire/livewire/blob/fdddb53726a101e655615b9c18c9ccc2c8016d35/js/directives/wire-loading.js#L175-L181

...will make that targets object look like this...

```
[
    {
        "target": "foo"
    }
]
```
...which means that when we get here...

https://github.com/livewire/livewire/blob/fdddb53726a101e655615b9c18c9ccc2c8016d35/js/directives/wire-loading.js#L76

...we return early (because `targets.length` is now greater than zero) and we never call the `startLoading()` function.

I have no idea how to "fix" this though 😕

**Comment by joshhanley** [comment]
@JayBizzle and @PhiloNL we're in the process of trying to get through the back log of failing tests.

Can you provide some more insight into this and how Wire Elements Pro functions? By my understanding of this, Livewire is actually working the way that I would expect.

Now my reasoning for this is because Livewire looks for any `wire:*` directives - like `click`, `mousedown`, `mouseover`, etc. (except for a few non-action directives like `offline` or `poll`) -  on buttons and scopes loading down to the actions that those directives call. As `wire:modal` actually calls something, then I would actually expect it to be scoped to the action that `wire:modal` should call.

I don't have access to Wire Elements Pro though so don't know how it works under the hood.

Based on the information that I have, the only thing I think we can do is maybe add a `wire:loading.global` or something like that, that removes the button specific scoping. I have had need in the past for a button to show loading both for itself and for any other actions/requests.

**Comment by JayBizzle** [comment]
Just a quick response to your comment @joshhanley to make it clear that this issue isn't really related to Wire Elements Pro, although it is through that package where I encountered this issue.

As you can see, the problem occurs in this simple example I created - [wirebox.app/b/x6dq1](https://wirebox.app/b/x6dq1)

When clicking the red button, both blue buttons should hide, but they don't, only one buttons hides, because the other buttons looks like this....

```html
    <button
        wire:loading.remove
        wire:foo="test"
        class="text-white bg-blue-700 hover:bg-blue-800 focus:ring-4 focus:ring-blue-300 font-medium rounded-lg text-sm px-5 py-2.5 me-2 mb-2 dark:bg-blue-600"
    >
```

As we can see, this button has two `wire:` directives attached, and this is what causes the issue.

Not sure what else I can tell you at this moment, thus "bug" definitely caught me by surprise.

Happy to help more if I can, just let me know.

Thanks

calebporzio:
Closing this PR as it's been open for a long time. If still relevant, please re-submit, ideally with a solution if you can. thank you!

PR #8320, opened 2024-04-17
Add failing test for checkbox validation bug

aerni:
This PR adds a couple of failing and non-failing tests for the bug described in https://github.com/livewire/livewire/discussions/8310.

It turns out that validation with `validateOnly()` works as expected in a unit test. The issue only happens in a browser test. This leads me to believe that the issue is somehow connected to Alpine.

- The unit test `realtime_validation_works_as_expected` doesn't throw an error
- The browser test `realtime_validation_works_when_calling_validate_only` fails as per the bug report
- The browser test `realtime_validation_works_when_using_validate_attribute` passes as per the bug report

This issue might be connected to https://github.com/livewire/livewire/pull/8235. And maybe to https://github.com/livewire/livewire/pull/8291 as well.


joshhanley:
@aerni thanks for the PR. I've just had a chance to investigate this and what I found is that the issue is what is being passed into `validateOnly()`.

When we use the `#[Validate]` attribute, what gets called under the hood is `$this->validateOnly('items');`.

But when you are calling it in your updated hook, the `$field` property is set to `items.0`, so what is getting called is `$this->validateOnly('items.0');`

So what you need to do is actually pass `items` to the validateOnly method for it to work the same. I've pushed a fix into https://github.com/livewire/livewire/pull/8320/commits/05b1f9105a21b054d248c3de6d8e98501b45ac45 so you can have a look.

Going to close this for now as I don't think it is a bug. Feel free to open a new PR if you still think there is an issue.

aerni:
@joshhanley Thanks for looking into this and providing a workaround! 

I'm aware of your findings as per the "Why this happens" section in the issue original https://github.com/livewire/livewire/discussions/8310. The reason I reported this as a bug is that the behaviour is different to Livewire 2. I ran into this when upgrading to Livewire 3. If it's not a bug but a deliberate breaking change for Livewire 3, it should be documented as such.

If your fix https://github.com/livewire/livewire/commit/05b1f9105a21b054d248c3de6d8e98501b45ac45 could be applied in the `validateOnly` method, that would solve the issue. But not sure if it would cause other issues.

joshhanley:
@aerni no worries! Ah sorry I didn't look at the discussion. It's best to always add a small summary of any relevant finds to the PR descriptions along with the link to the discussion as we don't dig into discussions unless we need to.

I don't think it was an intentional change/break from V2, but to change it now would cause new breaking changes. So I think it would be best if it was documented in the upgrade guide instead. PRs welcome to fix that!

PR #8322, opened 2024-04-17
use setHTMLUnsafe in morphing for declarative shadow dom support

AlexandreBonaventure:
Discussion: https://github.com/livewire/livewire/discussions/8321


AlexandreBonaventure:
OK good to review now I've added a new test

AlexandreBonaventure:
not sure why tests are failing now ☹️

calebporzio:
Hey @AlexandreBonaventure - I triggered a re-build to see if the tests were just being flaky.

Also, I get the change, I just want to make sure there's a compelling reason. What are the real-life situations you are benefiting from this change in? Also, are there others that have voiced demand for this support?

It seems harmless enough, but I try not to make a habit of merging "sure, why not" PRs, and instead merge when the answer is "oh hell yeah" - make sense?

AlexandreBonaventure:
@calebporzio sure: 
quoting what I said in the discussion: 

> DSD support is pretty recent so that's not really a surprise but Livewire users could really benefit from it (easy encapsulation, server-rendered web components, etc..)

First of all, I understand this is a niche use-case at the moment, especially given how recent the broad support from browser is. That said declarative shadow DOM is part of the interop 2024 initiative https://wpt.fyi/interop-2024. It goes to show that this is not just like an "experimental" feature.

Now for the real "oh hell yeah" explanation you are looking for, I definetly want to write an article about that because I've been experimenting with it a lot recently, and again I think we'll have even more patterns to discover once adoption is growing, but in a nutshell:

- I think using web components in Livewire is pretty common and have been asked by others from what's I've seen in github discussions or blog articles. (https://mariohamann.com/livewire-web-components-attributes, https://github.com/alpinejs/alpine/issues/172)
- The reason why it is powerful is that Alpine is great but has some limitations when you need some complex frontend components (I myself need to build very specific frontend-heavy with drag and drop file inputs and recursive folder crawling etc). Browsers are offering an actual standard specifically for that, we can (should?) use it! 
- now we're getting into details but Async-alpine like functionality with a simple snippet (almost for free!!)! https://github.com/shoelace-style/shoelace/blob/next/src/shoelace-autoloader.ts
- and so much more really I'd like to share more but lacking time right now :) (things like interop, design system across different frameworks, scoped css) 
- really pair well with Livewire cause at the end of the day "it's just HTML" and enable some difficult frontend updates with just an attribute change (using Alpine bindings) that would otherwise be super difficult to achieve.

Now that was highlighting some of the benefits of web components that I'm sure you already aware of but for Declarative Shadow DOM specifically:
This is a just the perfect fit to use web components server-side. For instance, we are building a localized app and when building web components using DSD we can just render the copy with `__()` server side. Otherwise, we would have to use an i18n solution in the frontend as well. DSD really simplifies a lot of things and let you also leverage Blade compilation inside  web components!!! 🤯  (think about all the patterns that are emerging out of that: scoped slots, you name it!)  

If you want at some point, I'll be glad to share with you some snippets and screenshare if that is something you are interested in (our company is a livewire partner) 
Here's a POC/draft experiment that I'm playing with:

```
<axr-input-file {{ $attributes }}>
    <template shadowrootmode="open" shadowrootclonable>
        <div x-data="{ files: [] }" class="f-body border-border-divider border rd-2 text-center">
            <label class="p-2 block">
                <ul>
                    <template x-for="file in files">
                        <li
                            class="p-2 px-3 text-foreground-inverted bg-background-inverted flex items-center z-0 rd-2 relative">
                            <span x-text="file.name"></span>
                            <span x-text="$host.getFileSize(file)" class="ml-auto"></span>
                        </li>
                    </template>
                </ul>
                <input type="file" class="fileinput" @input="onInput" />
                <div class="p-3" x-show="!files.length">
                    <x-utils.translate keypath="__('Drag and drop here to add file. Or :browse.')">
                        <x-slot:browse>
                            <span class="link">
                                {{ __('browse') }}
                            </span>
                        </x-slot:browse>
                    </x-utils.translate>
                </div>
            </label>
        </div>
        <style>
            :host {
                display: block;
                font-family: var(--sans-serif);
            }

            input[type="file"] {
                position: absolute;
                left: -30000px;
                opacity: 0;
            }
        </style>
    </template>
</axr-input-file>

```
As you can see we can even Alpine inside the shadow DOM to power template update (so we don't need something like Lit or whatever) but the functionality itself is powered by the web component, that's an excerpt of the type of things we need to do:

```ts
function validateType (item: DataTransferItem): boolean {
      // first validate file/directory type
      // https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry
      const getAsEntry = item.webkitGetAsEntry.bind(item) || item.getAsEntry.bind(item);
      const fsEntry = getAsEntry() as FileSystemEntry | null;
      if (fsEntry?.isFile && props.directory)
        return false;
      if (fsEntry?.isDirectory && !props.directory)
        return false;

      if (!typesToValidate.value)
        return true;
      const { type, name } = item.getAsFile()!;
      const extension = name.replace(/^.*(\.\w*)$/, '$1');
      return typesToValidate.value.some((criteria) => {
        const isExtensionCriteria = criteria.startsWith('.');
        const isWildcardCriteria = [...criteria.matchAll(/(image|media|video)\/\*/g)];
        if (isExtensionCriteria) {
          return extension === criteria;
        }
        else if (isWildcardCriteria.length) {
          const matcher = isWildcardCriteria[0][1];
          return type.match(new RegExp(`${matcher}/.*`));
        }
        return type === criteria;
      });
    }
    function onDrop (e: DragEvent) {
      turnOffDragStyles();
      doHideDropHover(e);
      (async () => {
        if (props.active && e.dataTransfer) {
          const dt = e.dataTransfer;
          const items = Array.from(dt.items);

          const validatedFiles = { valid: [], invalid: [] } as { valid: File[]; invalid: DataTransferItem[] };
          const fsEntries = [] as FileSystemEntry[];
          for (const item of items) {
            if (validateType(item)) {
              const file = item.getAsFile()!;
              if (!props.directory) { validatedFiles.valid.push(file); }
              else {
                // handle dir tree
                // https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry
                const fsEntry = (item.webkitGetAsEntry?.() || item.getAsEntry?.()) as FileSystemEntry;
                if (!fsEntry)
                  return;
                fsEntries.push(fsEntry);
              }
            }
            else {
              validatedFiles.invalid.push(item);
            }
          }
          async function scanEntry (entry: FileSystemEntry) {
            if (entry.isDirectory) {
              const dirEntry = entry as FileSystemDirectoryEntry;
              async function * readAllEntries (reader: FileSystemDirectoryReader) {
                // To read all files in a directory, readEntries needs to be called repeatedly until it returns an empty array. In Chromium-based browsers, the following example will only return a max of 100 entries.
                let entries = null as FileSystemEntry[] | null;
                let done = false;
                while (!done) {
                  const _entries = await new Promise<FileSystemEntry[]>((resolve, reject) => reader.readEntries(resolve, reject));
                  entries = _entries;
                  if (!entries.length)
                    done = true;
                  for (const entry of entries!)
                    yield entry;
                }
              }
              for await (const entry of readAllEntries(dirEntry.createReader()))
                await scanEntry(entry);
            }
            else {
              const file = await new Promise<FileWithRelativePath>((resolve, reject) => (entry as FileSystemFileEntry).file(resolve, reject));
              if (validateFile(file)) {
                file._relativePath = entry.fullPath; // HACK set relative path ourselves which is generally prefilled when you use the <input webkitdirectory> picker
                validatedFiles.valid.push(file);
              }
            }
          }
          for (const fsEntry of fsEntries)
            await scanEntry(fsEntry);
```

This is definetly something we could achieve with an alpine component (`Alpine.data`) and there are some overlaps but it seems like using web component is:
- more portable
- compatibility with tooling (like HMR functionalities) 
- no need to add additional library like async-alpine to lazy load
- really encapsulate logic, using raw alpine component you can sometimes run into attributes conflicts when used within blade components.... 
 - plus all the things built on top of web component standard in the future
 
At the end of the day, I'm not yet sure this is necessarily something I will want to use everywhere, but it seems to almost "just work" and if that's a web standard, I'm sure others will benefit from using it.
Sorry for the long message, I'm just really excited about this new experimental ground

Thanks!

AlexandreBonaventure:
as for the tests, they are passing locally on my end...

calebporzio:
Ok, slept on this and I think I want to hold off still. Once there's more demand I'd be happy to pull it in.

PR #8328, opened 2024-04-19
Fix file uploads while time travelling

PerryvanderMeer:
See https://github.com/livewire/livewire/discussions/8307

## Problem
File uploads are broken while time travelling with `Carbon::setTestNow()` or `$this->travelTo()`.

## Cause
While uploading files, old uploads are cleaned up, based on the file's last modification time.

https://github.com/livewire/livewire/blob/ed89372eb6eee3952bf7ccd1a6a082b29a709a4b/src/Features/SupportFileUploads/WithFileUploads.php#L114

While uploading during time travelling, the filesystem does not respect the 'fake' timestamp and uses the 'normal' timestamp for setting the file's last modification time. While passing Livewire's `cleanupOldUploads()` method, it looks like the file needs to be cleaned up and the file will be removed immediately.

## Solution 
After uploading a file while running tests, we can touch it and update the file's last modification time to the current timestamp (based on Carbon, so it respects time travelling).

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -23,6 +23,14 @@ class TemporaryUploadedFile extends UploadedFile
         $tmpFile = tmpfile();
 
         parent::__construct(stream_get_meta_data($tmpFile)['uri'], $this->path);
+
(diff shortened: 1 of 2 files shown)


cdarken:
Tested it, the fix works.

calebporzio:
Dig it, thanks!

PR #8331, opened 2024-04-20
 Add support for "paginationSimpleView()" method in components 

iBotPeaches:
#1304

## Problem

If you switch to a paginator that uses simple or even the cursor offset based paginator. You lose the ability to configure the exact view used. Your only configuration option relies on the `paginationTheme` property which forces the `livewire` prefix, thus preventing the same override that exists on the `paginationView`.

## Cause

The function `paginationSimpleView` has no override, unlike the corresponding `paginationView` function: https://github.com/livewire/livewire/blob/main/src/Features/SupportPagination/SupportPagination.php#L127-L135

## Solution

Much like done in this [commit](https://github.com/livewire/livewire/commit/12130528b493cfe5260d989fc77f34ac30e746af), add override support for simple based paginators.

### DIFF
--- a/docs/pagination.md
+++ b/docs/pagination.md
@@ -380,6 +380,7 @@ public function paginationView()
 {
     return 'custom-pagination-links-view';
 }
+
 public function paginationSimpleView()
 {
     return 'custom-simple-pagination-links-view';
--- a/docs/pagination.md
+++ b/docs/pagination.md
@@ -359,7 +359,7 @@ Once the files have been published, you have complete control over them. When re
 If you wish to bypass Livewire's pagination views entirely, you can render your own in one of two ways:
 
 1. The `->links()` method in your Blade view
-2. The `paginationView()` method in your component
+2. The `paginationView()` or `paginationSimpleView()` method in your component
 
 ### Via `->links()`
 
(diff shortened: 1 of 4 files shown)


PerryvanderMeer (changes requested):
Can you please add a short description of the existence of `paginationSimpleView()` [to the docs](https://livewire.laravel.com/docs/pagination#using-custom-pagination-views)?

iBotPeaches:
> Can you please add a short description of the existence of `paginationSimpleView()` [to the docs](https://livewire.laravel.com/docs/pagination#using-custom-pagination-views)?

Adjusted some parts of docs to call out the option you have here.

docs/pagination.md
```diff
@@ -371,15 +371,19 @@ The first approach is to simply pass your custom pagination Blade view name to t
...
 
-The second approach is to declare a `paginationView` method inside your component which returns the name of the view you would like to use:
+The second approach is to declare a `paginationView` or `paginationSimpleView` method inside your component which returns the name of the view you would like to use:
 
 ```php
 public function paginationView()
 {
     return 'custom-pagination-links-view';
 }
+public function paginationSimpleView()
```
PerryvanderMeer, line 383: Please add a new line in between.

PerryvanderMeer (changes requested):

PerryvanderMeer (approved):

calebporzio:
Perfect, thank you!

PR #8340, opened 2024-04-23
Add request hook

joshhanley:
For Verbs Livewire, we need to be able to hook into the request before Livewire starts processing so we can set up the event queues and the response after Livewire has finished processing to finalise the event queues.

Currently, there is no hook for `request`, but there is one for `response`, so this PR adds a `request` hook in.

https://github.com/livewire/livewire/blob/1b1e0f02f2d23aaa8a5d7c8c977078aa3963206d/src/Mechanisms/HandleRequests/HandleRequests.php#L102

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -77,32 +77,34 @@ class HandleRequests extends Mechanism
 
     function handleUpdate()
     {
-        $components = request('components');
+        $requestPayload = request('components');
 
-        trigger('request', $components);
+        $finish = trigger('request', $requestPayload);
 
-        $responses = [];
+        $requestPayload = $finish($requestPayload);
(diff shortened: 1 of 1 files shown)


calebporzio:
Gotcha. I'm down, but I think the parameters are kind of a mis-match.

I'd say either we pass nothing into the `request` hook, or we pass `$components` in, to match the `response` hook. Also `$components` would be more useful than `request()` as you could always get the request out of the container yourself.

You down?

joshhanley:
@calebporzio Yeah makes sense. I was just matching the fact that the response hook was getting the response array.

How about both? 😁 `request, components`

But otherwise yeah I'm cool with just compoents.

calebporzio:
Yeah I still don't like how passing in request() and $components wouldn't be symmetrical with the response hook. I'll just rename "$components" to match better.

PR #8344, opened 2024-04-24
Fix Laravel 10.x dependency

tpetry:
Discussion #7786 talked about an issue when using Livewire 3 with Laravel 10: Livewire now depends on a new method introduced in Laravel `10.15.0` but still says it only needs `10.0`. So Livewire fails when using an older version. This fix corrects the dependency information to instruct composer to also upgrade Laravel to a newer version.


calebporzio:
Thanks @tpetry !

PR #8348, opened 2024-04-25
Clarify docs for validation utilities

PerryvanderMeer:
Follow-up on #8254.
Previously we only mentioned properties, but methods also conflict with Livewire's mechanisms.


PhiloNL:
Thanks!

PR #8350, opened 2024-04-25
Script URL generation with UrlGenerator

olehasoft:
Provide Script URL generation using UrlGenerator.

I'm sure it's correct when link generation is performed in a system-wide way, because its functionality can be changed in the project.

For example, my current project is located in a subfolder and I am forced to use the following hack (AppServiceProvider):

```
$this->app['livewire']->setScriptRoute(fn (array $handle) => tap(
    $this->app['router']->getRoutes()->getByAction(join('@', $handle)),
    fn ($route) => $this->app['config']->set(
        'livewire.asset_url',
        $this->app['url']->toRoute($route, [], false)
    )
));
```

### DIFF
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -150,7 +150,7 @@ class FrontendAssets extends Mechanism
     public static function js($options)
     {
         // Use the default endpoint...
-        $url = app(static::class)->javaScriptRoute->uri;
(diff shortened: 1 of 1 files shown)


PhiloNL:
I think you should also be able to set the URL using the Livewire config:

```php
[
    'asset_url' => 'https://my-domain.com/sub-path'
]
```

Or set it directly on the directive:

```blade
@livewireScripts(['url' => 'https://my-domain.com/sub-path/livewire.min.js'])
```

PR #8366, opened 2024-04-27
Add support for `#[On('updatedFoo')]`

PerryvanderMeer:
This PR adds support for using `#[On('updatedFoo')]`.
Previously no `updated*` events were fired, which prevented the usage of listeners.

Before:
```php
public function updatedDate()
{
    $this->calculate();
}

public function updatedFromCurrency()
{
    $this->calculate();
}

public function updatedFromAmount()
{
    $this->calculate();
}

public function calculate()
{
    //
}
```

After:
```php
#[On('updatedDate'), On('updatedFromCurrency'), On('updatedFromAmount')]
public function calculate()
{
    //
}
```

### DIFF
--- a/src/Features/SupportEvents/UnitTest.php
+++ b/src/Features/SupportEvents/UnitTest.php
@@ -3,6 +3,7 @@
 namespace Livewire\Features\SupportEvents;
 
 use Livewire\Component;
+use Livewire\Form;
 use Livewire\Livewire;
 
 class UnitTest extends \Tests\TestCase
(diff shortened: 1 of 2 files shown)


calebporzio:
Ah this is interesting and I can see how it's useful for not writing many hook methods.

However, currently "On" is restricted to the even system and this is more the "hook" system (not front-end dispatched events.)

Therefore I'm going to say no to this.

Another way you can make these methods a single method is using the more generic "updated" hook which receives the full property path name as a parameter. Then you can provide a match statement or something to target specific updates and run `->calculate()`. Thanks!

nuernbergerA:
@calebporzio would you be open to a new attribute `#[Updated('foo')]`?

PR #8369, opened 2024-04-28
Add failing test for Alpine error when using array_unshift with a loop

QuentinGab:
This PR adds a failing test for handling alpine components in a loop when using array_unshift or a similar operation.

When rendering a loop in Livewire, if a new item is unshifted, Alpine components (`x-data`) inside the loop will fail and throw errors.

The error only occurs after the second Livewire re-render. After the unshift operation, any other render will break alpine.

I suspect the problem to be related to the morph algorithm as unshifting in an HTML list causes the entire list to be recreated (new node).

However, the system functions correctly when using operations like push or when reordering the list. 

It's quite easy to reproduce this issue, making me wonder if I'm overlooking something or if there's an alternative solution.

### DIFF
--- a/src/Tests/AlpineMorphingBrowserTest.php
+++ b/src/Tests/AlpineMorphingBrowserTest.php
@@ -134,4 +134,53 @@ class AlpineMorphingBrowserTest extends \Tests\BrowserTestCase
             ->assertMissing('@hidden');
             // ->assertConsoleLogMissingWarning('show is not defined');
     }
+
+    /** @test */
+    public function alpine_property_persists_on_array_unshift()
+    {
+        Livewire::visit(new class extends Component {
+            public array $values = [2, 1, 0];
+
+            public function unshift()
+            {
+                array_unshift($this->values, count($this->values));
+            }
+
+            public function push()
+            {
+                array_push($this->values, count($this->values));
+            }
+
+            public function render(): string
+            {
+                return <<<'BLADE'
+                    <div>
+                        <div>
+                            @foreach($values as $value)
+                                <div wire:key="item-{{ $value }}" x-data="{ show: false }" >
+                                    <div>Value #{{ $value }}</div>
+                                    <span x-show="show" x-cloak>hidden</span>
+                                </div>
+                            @endforeach
+                        </div>
+
+                        <button type="button" wire:click="unshift" dusk="unshift">Unshift</button>
+
+                        <button type="button" wire:click="push" dusk="push">Push</button>
+                    </div>
+                BLADE;
+            }
+        })
+            ->assertSee('Value #2')
+            ->assertDontSee('hidden')
+            ->assertDontSee('Value #3')
+            ->waitForLivewire()->click('@unshift')
+            ->assertSee('Value #3')
+            ->assertDontSee('hidden')
+            ->waitForLivewire()->click('@unshift')
+            ->assertSee('Value #4')
+            // The second time: Error thrown in console, and Alpine fails and shows the hidden text when it should not.
+            ->assertDontSee('hidden')
+            ;
+    }
 }
--- a/src/Tests/AlpineMorphingBrowserTest.php
+++ b/src/Tests/AlpineMorphingBrowserTest.php
@@ -146,11 +146,6 @@ class AlpineMorphingBrowserTest extends \Tests\BrowserTestCase
                 array_unshift($this->values, count($this->values));
             }
 
-            public function push()
-            {
-                array_push($this->values, count($this->values));
-            }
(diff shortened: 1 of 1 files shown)


QuentinGab:
I should have mention that this bug also breaks nested Livewire components causing the whole app to freeze.
It's a real blocker for me as I'm heavily impacted by the problem in a chat feature (lazy loading previous messages would cause errors)

QuentinGab:
Hey @calebporzio and @joshhanley,

I should have mentioned that this bug also affects nested Livewire components, causing the entire app to freeze.

It's a real blocker for me as I'm heavily impacted by the problem my app. Unfortunately I have no time to help exploring the codebase and find a fix.

However, and I'm sorry if it's inappropriate as I would never want to push you, I am willing to offer a substantial financial incentive, through my company, to find a solution to this issue.

PerryvanderMeer:
@QuentinGab Your `wire:key` might need some more uniqueness.
In your example, Alpine actually tries to replace `item-0` with a new `item-0`.
If I modify your `wire:key` with `item-{{ $value }}-{{ Str::random(16) }}`, the test passes.

QuentinGab:
> @QuentinGab Your `wire:key` might need some more uniqueness. In your example, Alpine actually tries to replace `item-0` with a new `item-0`. If I slightly modify your example with `item-{{ $value }}-{{ Str::random(16) }}` as `wire:key`, the test passes.

Hi @PerryvanderMeer and thanks for the reply,

Actually by adding a random key, you are re-rendering the entire DOM without doing any smart morphing at all. As a consequence you will lose every Alpine/Livewire states. 
This is not how Livewire is supposed to work but I guess it could be a workaround for cases where you don't need to preserve states. Unfortunately I doubt that it can be applied to my app

PerryvanderMeer:
@QuentinGab I see. You can wrap your `div` inside another `div` with a [key](https://alpinejs.dev/plugins/morph#keys), like:

```blade
@foreach($values as $value)

    <div key="{{ $value }}">
        <div wire:key="item-{{ $value }}" x-data="{ show: false }">
            <div>Value #{{ $value }}</div>
            <span x-show="show" x-cloak>hidden</span>
        </div>
    </div>

@endforeach

```

QuentinGab:
@PerryvanderMeer It makes the test pass which is surprising to me, I was assuming `key` and `wire:key` would have the exact same effect and role. 

So I took a look at the morph Alpine plugin and found out that Livewire is using it the Apline plugin with a custom `key` getter.

https://github.com/livewire/livewire/blob/b818a5b85f5ee041b6568fea16a2926a7b91c193/js/morph.js#L81-L90

In the test, I changed `wire:key` to `key` without even wrapping the div in another one, and the test also passes!

This makes me think that there is definitely something in the Livewire `morph` plugin that is causing the unshift operation to fail.

QuentinGab:
Err

> @PerryvanderMeer It makes the test pass which is surprising to me, I was assuming `key` and `wire:key` would have the exact same effect and role.
> 
> So I took a look at the morph Alpine plugin and found out that Livewire is using the Apline plugin with a custom `key` getter.
> 
> https://github.com/livewire/livewire/blob/b818a5b85f5ee041b6568fea16a2926a7b91c193/js/morph.js#L81-L90
> 
> In the test, I changed `wire:key` to `key` without even wrapping the div in another one, and the test also passes!
> 
> This makes me think that there is definitely something in the Livewire `morph` plugin that is causing the unshift operation to fail.

Erratum, `key` on the child or a wrapper actually do nothing.

This would pass the test:
```blade
@foreach($values as $value)
    <div x-data="{ show: false }" >
        <div>Value #{{ $value }}</div>
        <span x-show="show" x-cloak>hidden</span>
    </div>
@endforeach
```

The only case where it fails is when using `wire:key`

PerryvanderMeer:
@QuentinGab Right. Did some digging (in Alpine's history) and definitely looks like an Alpine problem.
You can do either:
- Replace `x-cloak` with the traditional `style="display: none"`
- Add an additional `id` to your `div`, like `<div id="{{ $value }}" wire:key="{{ $value }}" x-data="{ show: false }">`

When doing either, your test will pass.

PerryvanderMeer:
@QuentinGab any chance you can rebase this PR? I've tried with version 3.5.17 and the test passes:

afbeelding

QuentinGab:
The added test in this PR is now passing on the latest version.

QuentinGab:
> @QuentinGab any chance you can rebase this PR? I've tried with version 3.5.17 and the test passes:
> 
> afbeelding

You are right, I have closed the PR.

PR #8380, opened 2024-04-29
Refactor variable names to use "article" instead of "post" to align w…

hrant1020:


hrant1020:
Despite the model being named "Article" several variables and properties were named "post".

calebporzio:
Thanks!

PR #8384, opened 2024-04-30
Remove redundant class_exists() check as is_subclass_of() already checks class existence

hrant1020:
Although `is_subclass_of()` can also check for interface implementations, in this case, we are explicitly working with the class `Component::class`.


calebporzio:
thanks!

PR #8385, opened 2024-05-01
Docs: Add tip for getClientOriginalName method

caseydwyer:
Highlights the `getClientOriginalName` method from the `Livewire\Features\SupportFileUploads\TemporaryUploadedFile` class. I couldn't find reference to that method anywhere in the docs. A handy little helper that deserves a spotlight, IMHO! 👏


calebporzio:
Thanks!

PR #8388, opened 2024-05-02
Prevent computed properties from being overridden

calebporzio:


sinhnv1991:
Can u help fix issue scope of computed property example public can access display value on blade by $this->foo->bar and private, protected should avoid.
And current version none computed property have bug example private $foo; on blade i can access private $foo property by $this->foo on blade

PR #8392, opened 2024-05-02
Fix Issue where S3 Cleanup Command Overwrites Existing Rules

iAmKevinMcKee:

I did not start a discuss, but it's a bugfix for an existing feature.

Yes

No, it's one single thing

I did not write any tests. 

I ran the `livewire:configure-s3-upload-cleanup` command today and when I did, I noticed that it overwrote my existing rules in my s3 bucket. I checked and the `putBucketLifecycleConfiguration` method from the S3 SDK overwrites the existing rules instead of appending the livewire rule to others. See: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycleConfiguration.html

Specifically: Creates a new lifecycle configuration for the bucket or replaces an existing lifecycle configuration. Keep in mind that **this will overwrite an existing lifecycle configuration, so if you want to retain any configuration details, they must be included in the new lifecycle configuration.**

This update first checks to see if there are any rules on the bucket. 

- If it finds a rule related to the livewire-tmp folder (or whatever was specified by the app in the config), then it returns and does nothing. 
- If it finds other rules, then those rules are appended to the new rule for the livewire-tmp folder before calling `putBuckerLifecycleConfiguration`
- If it does not find any other rules, the API throws an exception (which is a strange way for the AWS API to say there are no rules IMO), but we just catch that exception, ignore it, and proceed with adding the new rule as the only rule for the bucket.

I have tested this with all three scenarios listed above, and it all works.

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/S3CleanupCommand.php
+++ b/src/Features/SupportConsoleCommands/Commands/S3CleanupCommand.php
@@ -2,6 +2,8 @@
 
 namespace Livewire\Features\SupportConsoleCommands\Commands;
 
+use Aws\S3\S3Client;
+use function array_merge;
 use function Livewire\invade;
 use Livewire\Features\SupportFileUploads\FileUploadConfiguration;
 use Illuminate\Console\Command;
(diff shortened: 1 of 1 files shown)


calebporzio:
Dope! Thanks for this Kevin!

PR #8394, opened 2024-05-03
feat: Allow component to use the custom endpoint for update

sankarsuda:

 No. I have seen few requests and need for my microservice framework

 Yes. Separate branch created

 No

No

In the case of microservice architecture, we have to send the update to the different endpoints for components when it has its own middleware to handle. We have login component does not require auth middleware endpoint but other required auth middleware.

You can create a method ``endpoint` that return the array with URL and id

```
public function endpoint(): array
    {
        return [
            'url' => route('auth.livewire.update'),
            'id' => 'auth'
        ];
    }
```

Thanks for contributing! 🙌


calebporzio:
Thanks for this request, however changing the update endpoint system has almost always resulted in unforeseen errors for people so I'm always really hesitant to merge PRs like these. Thanks anyways!

PR #8399, opened 2024-05-04
Refactor Constructors

faissaloux:
Helloo 👋🏻 

In this PR I have refactored the contructors using Constructor Property Promotion.


calebporzio:
Hey @faissaloux - thanks for the contribution, however I think I'm going to keep things as-is for now. If we decide later to adopt this as a standard for the codebase, I'll re-open and merge this. Thanks again!

PR #8402, opened 2024-05-05
Remove redundant alias of Rule

alsterholm:
I noticed there was a redundant alias of the `Rule` class, which I'm guessing might've been a relic back from when the `Validate` attribute was called `Rule`.


calebporzio:
Ah yes? That's exactly right. Thanks!

PR #8403, opened 2024-05-07
Adding `wire:replace`

samlev:
Livewire morphing is great. The only issue is that sometimes certain elements, for one reason or another, aren't easily morphable. Maybe it's a custom component that has a shadow dom that livewire can't morph properly, or maybe it's an alpine compoinent that _needs_ to be re-initialised on each render, or maybe it's just an element that has event listeners attached that wouldn't be cleaned up properly if the element was re-used.

This isn't an entirely new concept - for livewire v2 I used the [`wire:replace`](https://github.com/archtechx/wire-replace) directive from @stancl, and I've implemented the basic concept in a number of livewire v3 projects as well.

Here are some "toy" examples that I've encountered recently where morphing caused more issues than it solved:

### Custom Web Components

```html
<json-viewer>@json($someContent)</json-viewer>
```
I ran into issues when using a [`<json-viewer>`](https://github.com/alenaksu/json-viewer) component. When it initializes, it will read the content in the slot, and re-render it as a collection of elements. Livewire can't dom-diff this as it's done inside the shadow dom, and even if it could the rendering is done on component initialisation. `wire:ignore` also wouldn't let you update the content of the `<json-viewer>`, and while adding `wire:key="{{ uniqid() }}"` would cause a similar result, it's just... messy.

Adding `wire:replace.self` to this component would instruct livewire to just replace the component with a new element, which will trigger the element initialization.

```html
<json-viewer wire:replace.self>@json($someContent)</json-viewer>
```

### Event listeners not getting cleaned up/pagination issues

```html
<table>
  <thead>
    <tr>
      <th>Name</th>
      <th>Email</th>
      <th><span class="sr-only">Actions</span></th>
    </tr>
  </thead>
  <tbody>
    @foreach($users as $user)
      <tr wire:key="user-row-{{ $user->id }}">
        <td>{{ $user->name }}</td>
        <td>{{ $user->email }}</td>
        <td>
          @can('archive', $user)
            <button
              wire:click="archive({{ $user->id }})"
              wire:confirm.prompt="Enter the user's email to archive|{{ $user->email }}"
            >Archive</button>
          @endcan
          @can('restore', $user)
            <button wire:click="restore({{ $user->id }})">Restore</button>
          @endcan
        </td>
      </tr>
    @endforeach
  </tbody>
  <tfoot>
    <tr>
      <td colspan="3">
        {{ $users->links() }}
      </td>
    </tr>
  </tfoot>
</table>
```
(This is possibly a bug in `wire:confirm` or `wire:key`, but it should illustrate my point)

When paging, even with `wire:key` on the row, the content from each row is seemingly morphed. If a button is "Archive" on one page and "Restore" on the next, it may still the `wire:confirm` trigger attached to the button even if the attribute is no longer there.

Any other state that ends up getting shoved onto the HTMLElement which isn't reflected in attributes will also persist, even if the element gets "re-used" for something else. Adding `wire:key` to the buttons as well is... not desirable.

Adding `wire:replace` to the `<tbody>` resolves the issue by avoiding the diff entirely, and just swapping in the new set of rows.

### Re-initializing components

Sometime you'll have an alpine component that needs to reset after each request - maybe it's an accordion, or something else collapsible. Livewire will helpfully attempt to keep alpine state between requests, but sometimes this isn't actually desirable. Again adding `wire:key` with a random value to elements just to force livewire to re-render them is messy.

## `wire:replace`

It's pretty simple - add this to an element, and during the DOM diffing phase, it will simply replace the inner HTML of the element with whatever new elements came from the server. By default, `wire:replace` only applies to _children_ of the element and not the element itself, but I've also added `wire:replace.self` so that the targeted element will also be replaced.

Technically DOM diffing still occurs for the element, but the `innerHTML` is replaced on the "from" element, so the diffing will result in "everything looks exactly the same". This allows the target element to still get attribute changes, etc.

`wire:replace.self` will replace the element itself (i.e. `outerHTML`) and then skip looking down the DOM tree.


calebporzio:
Interesting... I typically do the wire:key with unique ID trick when I need this kind of thing - but yeah it feels a little hacky and isn't obvious to most people.

I think I like this, just want to make sure before adding a new directive to the API. Thanks for the well-done submission!

samlev:
> I think I like this, just want to make sure before adding a new directive to the API. Thanks for the well-done submission!

I'll admit - this isn't actually how I _normally_ implement the directive. What I usually do is check for the existence of the attribute on the incoming element, so that I can choose when it is diffed, and when it's been replaced. e.g.;

```js
if (toEl.matches('[wire:replace]') el.innerHTML=toEl.innerHTML;
```

Which has the advantage that you can _conditionally_ replace the element by rendering `wire:replace` on it when you want it to be replaced. Ideally I would like `wire:ignore` to work the same way so that I could have finer control over the morphing from blade:

```html
<div>
  <json-viewer {{ $data->changed ? 'wire:replace.self' : 'wire:ignore' }}>@json($data)</json-viewer>
</div>
```

calebporzio:
Alright, let's do this thing. Thanks!

calebporzio:
(this is basically the perfect PR btw - fairly low-risk to merge, includes tests, good docs, etc.) - thanks so much

PR #8437, opened 2024-05-16
Fix - Handle encoded query string array keys

dllobell:

I came across this bug in a project for a client and I saw this discussion that shows and reproduces the bug: #8088

Yes, I created the branch: fix/encoded-query-string-keys

No, all changes all related

Yes

This could happen if the url comes from outside, an email for example, but also occurs when you generate a url with the laravel `route` function, since the query is encoded with the `Array::query` method that uses the `http_build_query` function, if you generate a url like this `route('name', ['foo' => ['bar']])` the result is `?foo%5B0%5D=bar`.

The solution I have found is to decode the key before checking if the key is an array inside the `fromQueryString` function of the history plugin, so before checking if the key includes `[` we decode the key `foo%5B0%5D` resulting in `foo[0]` so all works fine.

I didn't quite know how to test this efficiently, but I commited first a failing test that shows an example that, in the above case, after the component is loaded we set the query parameter `foo` to an empty array but it is still present when it should be removed from the query string.

Any help for testing this in a more robust manner would be appreciated.

Thanks for contributing! 🙌


MarcRO3:
Some news about this threat?

calebporzio:
Thank you for this. I did some digging on it today.

Your solution is what I would think to do, only problem is: Livewire will now convert all `%5D` and `%5B`s to `[`s and `]`s in the query string. Even for keys it's not tracking.

Make sense? Like if you load this URL: `?foo%5B0%5D=bar`

If there's ANY Livewire on the page, it will be converted to this in the query string on page load: `?foo[0]=bar`

Even if there is no `$foo` property in a component.

That feels a LITTLE heavy handed, but I'm not really sure. It's definitely a problem right now that is unaddressed.

Do you see any other solutions?

dllobell:
Yes, that makes sense, thanks for looking into this

The only solution I can think of is to pass the key to the `fromQueryString` function and check there if the decoded key includes the `[` and starts with the query key passed to the function, this way we only convert the tracked properties.

I commited the changes so you can check if it makes sense and I'm not forgetting anything that could break (all tests pass on my fork)

I don't think the solution is pretty but honestly I'm not seeing any other solution since I'm not familiar enough with the Livewire codebase at the moment

calebporzio:
Yeah, I think you're right -  if we can somehow just convert the keys that we're concerned with that should be good - even though ugly in the implementation. Worth it though to not break when routes are generated with array keys in them.

dllobell:
I already pushed the changes of the implementation for you to take a look

lrljoe:
Would be interested if this also addresses another issue that I've come across

QueryString is being incorrectly converted to Scientific Notation where it detects scientific notation syntax.  This can be impactful if you don't intend it as Scientific Notation

I've included an example component below.

If you set search to
13E7
Then refresh the page (so it uses the query string to populate)
It will convert to "130000000"

If you set search to 
12E33
Then refresh the page (so it uses the query string to populate)
It converts to "1.2E+34"

It's an interesting one in terms of when to/not to fix.

The simplest solution I found was to only decode valid json.  It does mean that any URL params that **are** scientific notation won't be converted, but I'd assume that fewer people need the automatic conversion vs don't need it.

```
use Livewire\Attributes\Url;

class TestComponent extends \Livewire\Component
{
    #[Url] 
    public string $search = '';

    public function mount()
    {

    }
    
    public function render()
    {
        return \Illuminate\Support\Facades\Blade::render(
            '<div><input type="text" wire:model.live="search" class="text-black" /></div>',
            deleteCachedView: true
        );
        
    }
}
```

The issue here appears to be related to 

Replacing original code:
```
        $decoded = is_array($initialValue)
            ? json_decode(json_encode($initialValue), true)
            : json_decode($initialValue, true);
```
With:
```
        $decoded = is_array($initialValue)
            ? json_decode(json_encode($initialValue), true)
            : (json_validate($initialValue) ? html_entity_decode($initialValue) : json_decode($initialValue, true));
 : json_decode($initialValue, true));
```
Appears to remediate the specific issue, but does break other things sadly.

AiAe:
Any update regarding this PR?

joshhanley:
@dllobell thanks for the PR! I was just working on PR #8819 and realised it is the same as your PR. As it was only failing tests, I've merged them in here so we can make sure everything is working as expected.

joshhanley (approved):

PR #8457, opened 2024-05-20
Use `assertSetStrict()` in tests

PerryvanderMeer:
The used  `assertSet()` does a very loose comparison and, for example, passes when `null` is expected and `(string) ''` is given.
I've replaced all the current usages with `assertSetStrict()`, except for assertions wherein two objects are being compared (strict mode will fail there). I encountered two issues while replacing.

---
### Decimal casting 
~~Tests `test_can_cast_decimal_attributes_with_one_digit_from_model_casts_definition` and `test_can_cast_decimal_attributes_with_two_digits_from_model_casts_definition` were expecting floats for the casted decimals, [but Laravel returns the decimal as a string](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1429-L1436).~~
See https://github.com/livewire/livewire/pull/8463

---

### Resetting forms
~~Test `test_can_reset_form_object_property` was failing, because resetting form properties via `$this->reset('form.title')` was broken. `$title` was always reset to `null`, and because of the loose comparison to `(string) ''`, it was passing. While resetting properties to the default value via `$freshInstance = new static;`, the form isn't initialized yet, so it'll always return `null` as "default" value. I got around by calling the reset method directly on the Form. I've also added a test to make sure that the properties are reset the the actual default values (instead of `null` or `(string) ''`).~~
See https://github.com/livewire/livewire/pull/8464


src/Concerns/InteractsWithProperties.php
```diff
@@ -59,6 +59,12 @@ public function reset(...$properties)
                 $propertyName = $property->afterLast('.');
                 $objectName = $property->beforeLast('.');
 
+                if(method_exists($this->{$objectName}, 'reset'))
+                {
+                    $this->{$objectName}->reset($propertyName);
+                    continue;
+                }
```
calebporzio, line 66: This seems unrelated to ->assertSet - please split different features into different PRs.

Also, please make the formatting match the rest of the code - your if statement formatting is different than the one above it. These small things make PRs harder to merge and make me more hesitant to merge them because if attention wasn't paid to details as superficial as formatting, it likely wouldn't be paid to more important factors.

Thanks!

calebporzio (commented):
If this is just a PR making internal assertions strict, I'm probably good to just merge it. Thanks!

src/Concerns/InteractsWithProperties.php
PerryvanderMeer: @calebporzio It is only related in that I encountered it while refactoring. Will move it to a new PR. Sorry about the formatting - am used to a slightly different style.

joshhanley:
@PerryvanderMeer can you move the decimal cast changes to a new PR too (leave those ones as `assertSet()` for this PR)?

PerryvanderMeer:
@joshhanley all good now 👍

calebporzio:
Thanks @PerryvanderMeer - makes it so much easier and less risky to merge stuff like this 👌

PR #8459, opened 2024-05-20
Added automated step of changing the namespace used in v2 for Livewire\TemporaryUploadedFile in the upgrade command

irfanm96:
The livewire upgrade command was a timesaver and this is going to be another addition to it.

I was working on a project which had livewire v2, and I was able to migrate and get all of it working by the upgrade command.

Once I migrated to v3 I found out that I was using the `Livewire\TemporaryUploadedFile` in some checks.

For example, 

```php

if($file instanceof Livewire\TemporaryUploadedFile){
  // return the preview url
}

```

I found out that this was broken and could be added to one of the automated to upgrades.

This PR contains the automated conversion of the namespace from `Livewire\TemporaryUploadedFile` to `Livewire\Features\SupportFileUploads\TemporaryUploadedFile`

Hope this would be useful for many as it would make one less step to move Livewire v3.

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/Upgrade/ReplaceTemporaryUploadedFileNamespace.php
+++ b/src/Features/SupportConsoleCommands/Commands/Upgrade/ReplaceTemporaryUploadedFileNamespace.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Livewire\Features\SupportConsoleCommands\Commands\Upgrade;
+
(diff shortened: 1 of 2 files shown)


irfanm96:
Attaching some screenshots of the new addition of changing the namesapce for the Livewire\TemporaryUploadedFile

image

image

calebporzio:
This is great, thanks!

PR #8464, opened 2024-05-21
Fix resetting forms properties via `$this->reset('form.title')`

PerryvanderMeer:
Follow-up on https://github.com/livewire/livewire/pull/8457.
Resetting form properties via `$this->reset('form.title')` was broken. `$title` was always reset to `null`, and because of the loose comparison to `(string) ''`, it was passing. While resetting properties to the default value via `$freshInstance = new static;`, the form isn't initialized yet, so it'll always return `null` as "default" value. I got around by calling the reset method directly on the Form. I've also added a test to make sure that the properties are reset the the actual default values (instead of `null` or `(string) ''`).

### DIFF
--- a/src/Concerns/InteractsWithProperties.php
+++ b/src/Concerns/InteractsWithProperties.php
@@ -59,6 +59,11 @@ trait InteractsWithProperties
                 $propertyName = $property->afterLast('.');
                 $objectName = $property->beforeLast('.');
 
+                if (method_exists($this->{$objectName}, 'reset')) {
+                    $this->{$objectName}->reset($propertyName);
+                    continue;
+                }
+
                 $object = data_get($freshInstance, $objectName, null);
 
                 if (is_object($object)) {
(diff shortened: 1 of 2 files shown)


calebporzio:
Perfect, thank you!

hapidjus:
This change break our code, not sure if we are using the method the wrong way of if its an oversight in this PR.
We've used it like this:
```    
$this->reset([ 
	'title',
	'description',
	'recipients.*',
]);
```
Where `'recipients.*'` is the culprit. 

I'll put together a test if that'll help. Happy to write up a fix it deemed a bug and not implementation error on our part.

Geries:
This change also breaks `a.b.c` resets, because `$objectName = $property->beforeLast('.');` will return `a.b` and referencing it with `$this->{'a.b'}` will throw exception.

Geries:
If you do, probably the best is to make test cases specific to this too. I am unsure what happens in the case of nested objects, nested arrays, or better yet, in mixed cases.

hapidjus:
Yes I've added tests see PR #8929 but not sure they cover all cases, feel free to have a look.

PR #8465, opened 2024-05-21
Add `assertSnapshotSetStrict()` and `assertSnapshotNotSetStrict()`

PerryvanderMeer:
Follow-up on https://github.com/livewire/livewire/pull/8457. Was working on strict types for `assertSnapshotSet`, but those methods were not available yet (and I don't prefer adding the `, true` parameter like `assertSnapshotSet('foo', 'bar', true)`.


calebporzio:
Thanks!

PR #8471, opened 2024-05-21
Fix flaky test for cancelling uploads

PerryvanderMeer:
The test related to cancelling file uploads fails fairly frequently:

https://github.com/livewire/livewire/blob/72e900825c560f0e4e620185b26c5441a8914435/src/Features/SupportFileUploads/BrowserTest.php#L65

Did some digging and the cancel button is only clicked after the file is uploaded (rather than while uploading).
A race occurs between uploading and cancelling, resulting in some errors, because `uploadItem.request` is `undefined`:

Schermafbeelding 2024-05-21 165622


calebporzio:
Wow, thanks for digging on that!

PR #8474, opened 2024-05-22
Fix typed nullable properties in forms

PerryvanderMeer:
See https://github.com/livewire/livewire/discussions/8473

## Problem
In forms, properties typed like `string|null` cannot be (re)set to `null`.
It will unset the whole property, causing problems while rendering the view.
 
> Typed property App\Livewire\Form::$foo must not be accessed before initialization

## Cause
While setting the properties, the property will be unset when the given value is `null` and the property is typed to anything.
So that also goes for types like `string|null`.

https://github.com/livewire/livewire/blob/72e900825c560f0e4e620185b26c5441a8914435/src/Features/SupportFormObjects/FormObjectSynth.php#L51-L52

## Solution
We should perform an additional check whether the property type allows a null value.
If so, we should just set the property as "normal".

### DIFF
--- a/src/Features/SupportFormObjects/FormObjectSynth.php
+++ b/src/Features/SupportFormObjects/FormObjectSynth.php
@@ -48,7 +48,7 @@ class FormObjectSynth extends Synth {
 
     function set(&$target, $key, $value)
     {
-        if ($value === null && Utils::propertyIsTyped($target, $key)) {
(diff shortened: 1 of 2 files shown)


joaopalopes24:
+1

calebporzio:
This makes perfect sense - thanks for the quick writeup and good PR!

PR #8480, opened 2024-05-23
Fix view components with parameters as `Layout`

PerryvanderMeer:
See https://github.com/livewire/livewire/discussions/8479

## Problem
When we use a view `Component` as layout, the view parameters set in the `render` method are not actually passed into the view.

## Cause
We merge in the data from the generated view:

https://github.com/livewire/livewire/blob/72e900825c560f0e4e620185b26c5441a8914435/src/Features/SupportPageComponents/PageComponentConfig.php#L49

This works perfectly for an `AnonymousComponent`, but not for a `Component`.
For those, we need to merge in the data from the resolved view like `$layout->resolveView()->getData()`.

## Solution
We should also merge in the data from a resolved `Component` view.


calebporzio:
Makes sense, thanks!

PR #8482, opened 2024-05-23
Cleanup render methods in internal tests

PerryvanderMeer:
In a lot of tests, a render method is passed with an empty div, because otherwise the components don't work. 
But, our default `TestComponent` already contains this empty render method.
This PR refactors the test components to this `TestComponent`, eliminating all empty render methods and making the tests a lot clearer and cleaner.

Note: both `<div></div>` as `view('null-view')` render an empty div.


calebporzio:
Beautiful, thank you so much for these @PerryvanderMeer !

PR #8496, opened 2024-05-24
Fix script re-evaluation on browser back

tlaverdure:
I'm currently using `wire:navigate` in an app I'm building. When I use `@script` in a component I'm noticing the script does not re-evaluate after navigating away from the page and then hitting the browser back button.

Looks like this discussion may be related:
https://github.com/livewire/livewire/discussions/6374

It looks like we need to inscribe the effect with scripts when navigating?

https://github.com/livewire/livewire/blob/b818a5b85f5ee041b6568fea16a2926a7b91c193/js/directives/wire-navigate.js#L16-L22

https://github.com/livewire/livewire/blob/b818a5b85f5ee041b6568fea16a2926a7b91c193/js/component.js#L151-L167

I've added a test using a JavaScript `confirm` that passes while using `wire:navigate` with these code changes.


calebporzio:
This makes perfect sense, thanks! I'll merge after the tests pass

PR #8511, opened 2024-05-29
HandleRequests: use empty array as default value to avoid error in foreach

Jonnx:

Thanks for contributing! 🙌

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -13,7 +13,7 @@ class HandleRequests extends Mechanism
 {
     protected $updateRoute;
 
-    public function boot()
+    function boot()
     {
         // Only set it if another provider hasn't already set it....
         if (! $this->updateRoute) {
@@ -25,14 +25,14 @@ class HandleRequests extends Mechanism
         $this->skipRequestPayloadTamperingMiddleware();
     }
 
-    public function getUpdateUri()
+    function getUpdateUri()
     {
         return (string) str(
             route($this->updateRoute->getName(), [], false)
         )->start('/');
     }
 
-    public function skipRequestPayloadTamperingMiddleware()
+    function skipRequestPayloadTamperingMiddleware()
     {
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::skipWhen(function () {
             return $this->isLivewireRequest();
@@ -43,7 +43,7 @@ class HandleRequests extends Mechanism
         });
     }
 
-    public function setUpdateRoute($callback)
+    function setUpdateRoute($callback)
     {
         $route = $callback([self::class, 'handleUpdate']);
 
@@ -55,12 +55,12 @@ class HandleRequests extends Mechanism
         $this->updateRoute = $route;
     }
 
-    public function isLivewireRequest()
+    function isLivewireRequest()
     {
         return request()->hasHeader('X-Livewire');
     }
 
-    public function isLivewireRoute()
+    function isLivewireRoute()
     {
         // @todo: Rename this back to `isLivewireRequest` once the need for it in tests has been fixed.
         $route = request()->route();
@@ -78,7 +78,7 @@ class HandleRequests extends Mechanism
         return $route->named('*livewire.update');
     }
 
-    public function handleUpdate()
+    function handleUpdate()
     {
         $requestPayload = request(key: 'components', default: []);
 
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -78,9 +78,8 @@ class HandleRequests extends Mechanism
 
     function handleUpdate()
     {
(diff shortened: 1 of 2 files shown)


Jonnx:
This PR addresses an bug where the livewire update endpoint throws an exception when the HandleRequests class cannot process the request payload due to `requestPayload` being null instead of an empty list. 

Here are some related discussions around this issue:
- https://github.com/livewire/livewire/discussions/8510
- https://github.com/livewire/livewire/discussions/7646

joshhanley:
@Jonnx Thanks for the PR! Could you please add tests? Thanks!

calebporzio:
Yes, tests please. Also, in what scenarios would a livewire request come back without a components key?

Jonnx:
Agreed @calebporzio I didn't expect it either.

I am seeing the error logs which is odd. That's also why my first instinct was to get to the latest version first to see if it would just go away. 

Will add tests today

Jonnx:
@joshhanley do you have a suggestion on where best to add a test for this?

Jonnx:
@calebporzio @joshhanley added a unit test, it will fail for commit `2ef362d` as I disabled the new default value to confirm the test was adequate and it will pass for commit `5b8f6ec` once the new default value was re-enabled

Jonnx:
> Yes, tests please. Also, in what scenarios would a livewire request come back without a components key?

I am guessing it may be crawler related activity (seo or automated exploit scanner)

maciek-szn:
> I am guessing it may be crawler related activity (seo or automated exploit scanner)

For me it looks like a legitimate traffic from real users. But I still haven't been able to find out the cause of the errors.

Jonnx:
> > I am guessing it may be crawler related activity (seo or automated exploit scanner)
> 
> For me it looks like a legitimate traffic from real users. But I still haven't been able to find out the cause of the errors.

we could create a custom branch on my fork that adds some additional logging of request context data when the case is hit that the components parameter is not provided to see if we can identify additional patterns. (we wouldn't ever open a PR for that logging pattern back to this repo, but if it happens regularly enough, we may be able to get a hint at what is happening)

@maciek-szn let me know if you are interested and / or able to install a forked dependency in your project

maciek-szn:
> @maciek-szn let me know if you are interested and / or able to install a forked dependency in your project

Sure, I'm interested. I'm curious what the source of the issue is.

joshhanley:
@Jonnx Thanks! Yeah I wonder whether it is a bot? You could use a custom scripts route instead to see if it's coming from legitimate users or whether it's just a bot scanning for the livewire/update route.

Jonnx:
> @Jonnx Thanks! Yeah I wonder whether it is a bot? You could use a custom scripts route instead to see if it's coming from legitimate users or whether it's just a bot scanning for the livewire/update route.

I am already using a custom route which is extra interesting. and as @maciek-szn mentioned, he is getting error reports from real users

Jonnx:
@joshhanley @calebporzio Is there anything else needed on this pr or can we get the tests run and potentially merge this?

maciek-szn:
Some statistics from me. Yesterday I had 386818 `POST /livewire/update` requests. Of these, 69 (1.78%) generated the error `foreach() argument must be of type array|object, null given`. Therefore, the problem is not of a very important nature, but it is frustrating because it is difficult to get to its essence. Perhaps it is even invisible to users (no 5xx errors in the server logs).

samlev:
I would be curious about what's causing the "broken" requests, and potentially fixing _that_. While it's good to make sure that there aren't unexpected errors thrown, the "unexpected" part is _actually_ a request being sent without components in the payload.

My _personal_ take is that setting an empty default here is masking the issue, so yes - set a default to prevent the unexpected error, but then throw an exception _with a better description_ when the payload has no components because that's still not something that Livewire should process.

Jonnx:
> I would be curious about what's causing the "broken" requests, and potentially fixing _that_. While it's good to make sure that there aren't unexpected errors thrown, the "unexpected" part is _actually_ a request being sent without components in the payload.
> 
> My _personal_ take is that setting an empty default here is masking the issue, so yes - set a default to prevent the unexpected error, but then throw an exception _with a better description_ when the payload has no components because that's still not something that Livewire should process.

should we maybe add request validation rules to the update route saying that at least one component should be part of the payload and have the endpoint respond with a 4XX code?

samlev:
I wouldn't bother with validation rules - I would just do something like this:

```php
$requestPayload = request('components');

throw_if(! $requestPayload, new InvalidRequestException('Livewire update called with no components'));
```

Jonnx:
thoughts on how we can get this moving forward? @calebporzio @samlev @joshhanley

edalzell:
> thoughts on how we can get this moving forward? @calebporzio @samlev @joshhanley

You can always composer patch this in until they release it

Jonnx:
> > thoughts on how we can get this moving forward? @calebporzio @samlev @joshhanley
> 
> You can always composer patch this in until they release it

already doing that, i just want to make progress here so i don't have to keep a fork around forever

PR #8527, opened 2024-06-03
Bugfix/url attribute existence

samlev:
Just a little fix for somethign that was previously... undocumented behabiour.

If you had a `#[Url]` attribute and you _legitimately_ wanted to have a value of "noexist", the `BaseUrl` attribute would silently discard it:

`https://example.com/some-path?filter=noexist`
`dump($this->filter); // string ''`

This is a tiny fix which changes the "parameter doesn't exist in the URL" string to something random so that you should _probably_ never have a conflict. Or if you did, you would have to be incredibly, imensly lucky, and you should probably go invest your life savings in lottery tickets.


calebporzio:
Gotcha, good catch - thank you!

PR #8551, opened 2024-06-10
Allow `update_uri` to be specified from config file

peter-mw:

Yes i needed to override this method `getUpdateUri` in [this file](https://github.com/microweber/microweber/blob/6da4049151554272a048a7a21263e916b1a8244f/src/MicroweberPackages/Livewire/LivewireServiceProvider.php#L125-L127)  by doing not so good thing to extend the main class here, so a config option will solve this issue  

Yes 

No

No, but it works for me manually testing

I would like to have the full URL with FQDN of the update route as here in the injected JS

image

 
Currently the function adds a slash at the start so its impossible to add FQDN via `app($this::class)->setUpdateRoute` because of [this line](https://github.com/livewire/livewire/blob/main/src/Mechanisms/HandleRequests/HandleRequests.php#L31)

image

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -157,4 +157,18 @@ return [
     */
 
     'pagination_theme' => 'tailwind',
-];
+
+
+    /*
+   |---------------------------------------------------------------------------
+   | Update URI
+   |---------------------------------------------------------------------------
+   |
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for contributing however, we have moved away from adding things like this to the config file so I'm going to close.

PR #8554, opened 2024-06-11
fix: Livewire published script load on Laravel Vapor

percymamedy:
This PR allows Livewire Scripts to be loaded when running on Laravel Vapor or Serverless environment. 

The Functionally seems to have been lost as of Version 3 of Livewire see https://github.com/livewire/livewire/pull/1693

This will allow livewire to run seemlessly on a Vapor environment when assets are published.

## Changes:

In `FrontendAssets` method `usePublishedAssetsIfAvailable` determine if we are in a serverless environment and if so use `config('app.asset_url')` to render the Url for the livewire.js script.

### DIFF
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -227,7 +227,11 @@ class FrontendAssets extends Mechanism
 
         $versionedFileName = "{$fileName}?id={$versionedFileName}";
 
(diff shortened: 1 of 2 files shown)


cisco6co (approved):

samlev:
I think that it's worth making this not just apply in serverless environments - vapor is where most people encounter this, but it would affect any system where the assets are hosted elsewhere, not _just_ vapor.

calebporzio:
Thanks!!

PR #8573, opened 2024-06-17
Add test method for asserting partial match in redirect path

mokhosh:
Laravel has `assertRedirect` and `assertRedirectContains` for test requests.
Livewire currently only has `assertRedirect` so there is no way for asserting a partial match on redirect path.

This PR adds `assertRedirectContains` by mimicking how `assertRedirect` has already been implemented for Livewire, and using the Laravel code in `Illuminate\Testing\TestResponse::assertRedirectContains`.


calebporzio:
Great! Thanks @mokhosh!

PR #8579, opened 2024-06-20
update custom asset url docs

zsp03:
Due to [#7928](https://github.com/livewire/livewire/pull/7928), the docs should reflect the changes and be clear. 

I was thinking of using 
> [!WARNING]
> You should use `livewire.min.js` when APP_DEBUG is disabled

But that would require changing the initial wording.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -84,10 +84,25 @@ Livewire::setUpdateRoute(function ($handle) {
 
 ## Customizing the asset URL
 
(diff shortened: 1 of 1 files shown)


calebporzio:
This adds complexity for users overriding the script route. I think they can just skip serving the minified version right? Livewire only offered a non-minified version for a while and most people go along fine right?

If I'm miss-understanding and what you're adding solves something that is causing breakages for people following the docs, please re-open this PR. Thanks again!

zsp03:
Yes exactly, it was on my project a while back. I'm running on custom domain and suddenly it does not work after upgrading, only when i read the PR i noticed the change since the docs were not clear enough.

PR #8585, opened 2024-06-21
add failing test for persist with livewire component

nleibbrandt:
Hey @joshhanley & @calebporzio,

I recently encountered this issue in one of my projects and wanted to write a browser test for it. However, I'm struggling to write a Dusk test for it. Could someone help me with this?

Thanks in advance!


calebporzio:
Thanks for the submission - I think we will need a deeper explanation. It's hard for me to know by looking at this PR, what the problem is exactly - also, it's hard for me to know if this test is working (failing where it should) in the first place.

Please re-submit this PR when you've addressed the above. Thanks!

PR #8592, opened 2024-06-23
Fix: Persistent middlewares not redirecting on subsequent requests

Ing-LuisGuerrero:
Yes, this change is needed. Although I didn’t start a discussion beforehand, I will provide a detailed explanation below.

Yes, the branch is named `fix/persitent-middleware-redirect`

No, these changes are specifically focused on fixing the issue I identified.

Yes, it does.

## Context
Livewire’s current implementation of persistent middleware handling has an issue with middlewares that perform redirects. This error occurs on subsequent requests, not the initial request. Specifically, when a middleware returns a RedirectResponse (instead of the next callback), the processing flow (Pipeline) is short-circuited. Also, this prevents subsequent middlewares from executing and can cause potential errors in the application.

For example, consider the middleware array [MiddlewareA, MiddlewareB], where MiddlewareA may perform a redirect if certain conditions are not met. Due to the short circuiting, MiddlewareB does not execute, and the redirect intended by MiddlewareA does not occur as expected, leading to unintended Livewire actions and potential internal errors.

```php
  protected function applyPersistentMiddleware()
  {
      $request = $this->makeFakeRequest();

      $middleware = $this->getApplicablePersistentMiddleware($request);

      // Only send through pipeline if there are middleware found
      if (is_null($middleware)) return;

      Utils::applyMiddleware($request, $middleware); // Returned response is not used
  }
```

```php
  static function applyMiddleware(\Illuminate\Http\Request $request, $middleware = [])
  {
      return (new \Illuminate\Pipeline\Pipeline(app()))
          ->send($request)
          ->through($middleware)
          ->then(function() {
              return new \Illuminate\Http\Response();
          });
  }
```

## Solution
This PR proposes a solution by checking the response returned by the middleware pipeline. If the response is an instance of `RedirectResponse`, it should be returned directly to the client, effectively handling the redirect correctly.

```php
  static function applyMiddleware(\Illuminate\Http\Request $request, $middleware = [])
  {
      $response = (new \Illuminate\Pipeline\Pipeline(app()))
          ->send($request)
          ->through($middleware)
          ->then(function() {
              return new \Illuminate\Http\Response();
          });

      if ($response instanceof \Illuminate\Http\RedirectResponse) {
          abort($response);
      }

      return $response;
  }
```

## About Auth Middleware
You might think that authentication redirects work perfectly, and you’re correct. This is because the Auth middleware doesn't redirect directly, it throws an `Illuminate\Auth\AuthenticationException`. This exception breaks the flow differently. The Laravel Exception Handler catches this exception, does not report it, and automatically performs a redirect. As a result, redirects appear to function seamlessly.

### See
- [`Illuminate\Auth\Middleware\Authenticate`](https://github.com/illuminate/auth/blob/master/Middleware/Authenticate.php#L102)
- [`Illuminate\Foundation\Exceptions\Handler`](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Foundation/Exceptions/Handler.php#L585)


calebporzio:
This makes total sense and is really well done. Thank you! As soon as the tests pass, I'll merge this.

PR #8616, opened 2024-07-01
Fix facade actingAs return type docblock

santigarcor:
This PR fixes the return type of the `actingAs` method defined in the docblock of the `Livewire` Facade to match the signature in the `LivewireManager` [class](https://github.com/livewire/livewire/blob/main/src/LivewireManager.php#L206C14-L206C22).


calebporzio:
Thanks!

PR #8620, opened 2024-07-02
Add array with property name as key for forms on $this->validate() result

schonhoff:

Yes, you can find the discussion here: https://github.com/livewire/livewire/discussions/8618

Yes, the branch is named `add-array-with-property-name-as-key`

No, the changes are related to my "bugfix".

Yes.

Hi!

Using Laravels `$validated = $this->validate()` has the possibility to exclude some properties of an array before using it to create a model for example. It avoids to use non-validated properties. If you are using multiple forms in your component the `$validated` isn't structured. If you are using the same form for multiple properties (for example a billing address and a shipping address) and you want to use only $validated for creating the last form validated is only present in the `$validated` variable instead of all data from all forms. 

I added all snippets in the discussion.

@calebporzio  Care! This can potentially break applications, but I don't know any other possible solution for this problem.


schonhoff:
I don't know why the test is failing. Locally the test didn't fail. In my opinion it has nothing to do with this PR. Maybe someone can check it?

lrljoe:
Given that this would potentially be a breaking change for some instances, why would you not just call validate() against the form objects individually?
```
$this->validatedArray['form1'] = $this->form1->validate();
```

schonhoff:
I understand your point. In my practical case we are using forms and properties in the same component. To validate each form and after this validate the component properties (and forms again) isn't that great, isn't it? I know, it is a special use case.
But the problem with overwritting the form data if you use one form within two component properties can be very difficult to get as a new Livewire (3) deverloper. It is a little bit counter intuitive. In my case I used a lot of array validation in Livewire v2 and would love to swap to those form validations to encapsulate my logic. But this makes it quite difficult.

After a lot of thinking about this problem maybe there is another solution to it. How about a new parameter for the `$this->validate()` method called `withFormsKeyed` (or something else) which is set to false as default. This would be the current live version behavior. If you toggle the flag to `true` your would get the forms keyed with their property name. This wouldn't be a breaking change anymore and would fix the overwritting. This could be potentially explained in the docs for validation or forms and would help out people. How about this @calebporzio and @lrljoe ? I would update this PR to the behavior mentioned if you are okay with the solution.

lrljoe:
I think you've got to preserve the current behaviour, otherwise a lot of websites are going to break

Adding in a non-breaking change is going to be much better received!

schonhoff:
@lrljoe Can you have a look into the new implementation? Would this make sense to you?
It shouldn't break current apps and add the possibility to key the forms if someone has trouble with the current behavior.

lrljoe:
> @lrljoe Can you have a look into the new implementation? Would this make sense to you?
> It shouldn't break current apps and add the possibility to key the forms if someone has trouble with the current behavior. 

Will give it a spin in a test environment in next day or so!

joshhanley:
@schonhoff thanks for the PR! This seems the same as #8634 which has been mered, so I'm going to close this PR. If you're still experiencing issues, please feel free to open a new PR and explain how it is different from #8634. Thanks!

PR #8621, opened 2024-07-02
Add failing test for nested dynamic properties

schonhoff:

Yes, you can find the discussion here: https://github.com/livewire/livewire/discussions/8318

Yes, the branch is named `failing-test-updated-hook-dynamic-nested-array`

No, the changes are only related to the failing test.

It is a failing test :D

In Livewire 2 it was possible to use Lifecycle Hooks that are a part of the nested array. For example an array property like this:
```php
public array $foo = ['bar' => []];
```
If this is now a dynamic checkbox it won't work to use `updatedFooBar` as a Lifecycle Hook. This worked in v2. Now you have to use either
```php
  public function updated($propertyName): void
  {
    if(Str::startsWith($propertyName, 'foo.bar')) {
      // do something
    }
  }
```
or you have to use a very specific Lifecycle Hook like `updatedFooBar0` etc. This is very complicated for dynamic nested arrays. . I would like to implement a fix for it but I couldn't find the implementation of the nested Lifecycle Hooks in v2, sorry for that. I didn't know where do I add this test that's why I added a new BrowserTest file in the `Livewire\Features\SupportLifecycleHooks`. I hope that is okay.

Thanks for contributing! 🙌
No problem :-D

### DIFF
--- a/src/Features/SupportLifecycleHooks/BrowserTest.php
+++ b/src/Features/SupportLifecycleHooks/BrowserTest.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Livewire\Features\SupportLifecycleHooks;
+
+use Livewire\Attributes\Layout;
+use Livewire\Component;
+use Livewire\Livewire;
+use Tests\BrowserTestCase;
+
+class BrowserTest extends BrowserTestCase
+{
+    public function test_livewire_updated_hook_with_dynamic_properties()
+    {
+        Livewire::visit(new class extends Component {
+            public array $foo = ['bar' => []];
+            public string $resultUpdatedFooBarHook = '';
+            public string $resultUpdatedHook = '';
+
+            public function updated($propertyName): void
+            {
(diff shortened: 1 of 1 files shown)


calebporzio:
I see, yes. I believe there is an open failing test PR for a related issue: https://github.com/livewire/livewire/pull/8235

What's happening is, in V2, changing an array would register an update for the key `items` (or whatever your array property is called).

In V3, when you manipulate an array, each key change is registered as an update. So instead of `items` changing, you might have `items.0` and `items.1`

We realize this is unintuitive, especially when wanting to use nice APIs like your nested update hook.

For now, I'd say do the workaround you mentioned when you use `Str::startsWith()`

In the meantime I think we'll have to dive deeper on the causes of this problem and if the solutions are possible with out tagging a major version or making a breaking change.

Going to close this because we have that other one open right now to represent the problem.

Thank you!

schonhoff:
Thank for the response!

PR #8622, opened 2024-07-02
Do nothing when href doesn't exist

danie-ramdhani:
This PR fix when user click `a` tag with `wire:navigate` will do nothing (before is navigate to null path)


calebporzio:
This looks good, thanks!

PR #8623, opened 2024-07-02
Don't fetch new page when user long click the link

danie-ramdhani:
This PR is fixing when user long click the link, don't fetch new page.

Before:

[before.webm](https://github.com/livewire/livewire/assets/58886605/8eeef202-4097-4980-a5fa-26e6cba2047a)

After:

[after.webm](https://github.com/livewire/livewire/assets/58886605/f5a338a5-984c-4064-8c41-f53e222cd70e)


calebporzio:
Thanks for this submission - looking over your code and it looks like your change would break prefetching.

The value of prefetching is starting to request a page BEFORE a link press is released. Your code changes appear to move prefetching code to AFTER a link is released.

Please let me know if I've missed something. Thanks

calebporzio:
I see the problem you're trying to solve, but I don't think this implementation is it. Thank you for the submission. But I'm going to close this. Feel free to re-open with a solution that doesn't break existing behavior and provides testing for it. Thank you!

PR #8624, opened 2024-07-02
Fix image upload extension validation

calebporzio:
Currently Livewire derives the extension of an uploaded file from it's MIME type.

This poses a problem for image validation, because, typically, image validation relies on two basic pillars:
1. Ensuring the mime type of the file is an image type
2. Ensuring the file extension is an image type

Currently Livewire creates an environment where you're only validating mime types and could potentially lead to a vulnerability if all of the following conditions were met:
1. You construct a stored filename based on `$file->getClientOriginalExtension()`
2. You are storing files directly on your server in a public storage disk
3. Your nginx configuration allows executing files with .php extensions that aren't index.php

To better explain, here's a potentially vulnerable Livewire component:

```php
{
    use WithFileUploads;

    #[Validate('extensions:png')]
    public $file;

    public function save()
    {
        $this->validate();

        $this->file->storeAs(
            path: 'images',
            name: $this->file->getClientOriginalName(),
            options: ['disk' => 'public'],
        );
    }
```

In this scenario, a user could upload a file called `malicious.php` with a `image/png` MIME type and execute it on your server (assuming your nginx configuration allows).

This PR fixes the vulnerability by no longer deriving the temporary file upload extension from the MIME type, but rather from the actual file extension uploaded instead.

This makes it so that `extensions:png` validation will work and catch non-PHP extension files.

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -168,7 +168,7 @@ class TemporaryUploadedFile extends UploadedFile
     {
         $hash = str()->random(30);
         $meta = str('-meta'.base64_encode($file->getClientOriginalName()).'-')->replace('/', '_');
-        $extension = '.'.$file->guessExtension();
+        $extension = '.'.$file->getClientOriginalExtension();
 
         return $hash.$meta.$extension;
     }
--- a/src/Features/SupportFileUploads/UnitTest.php
+++ b/src/Features/SupportFileUploads/UnitTest.php
@@ -13,6 +13,7 @@ use Illuminate\Validation\Rule;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Http\UploadedFile;
 use Facades\Livewire\Features\SupportFileUploads\GenerateSignedUploadUrl;
+use Illuminate\Http\Testing\FileFactory;
 use Tests\TestComponent;
 
 class UnitTest extends \Tests\TestCase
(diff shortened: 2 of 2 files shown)


angelej:
This looks good to me!
I wasn't able to exploit the vulnerability anymore.
Since we are now using the extension from the filename, Laravel's `Illuminate\Validation\Concerns\ValidatesAttributes::shouldBlockPhpUpload()` can take into action blocking PHP extensions, even if the developer only validates the mime type.

> [!CAUTION]
> **However, I highly recommend to always validate the mime type and the file extension!**

calebporzio:
Yep, my thinking as well!

beberlei:
Given https://github.com/livewire/livewire/security/advisories/GHSA-f3cx-396f-7jqp, any chance this could be backported to 2.x? This looks reasonably maintained branch and had a release around 3.5.2 as well.

RChutchev:
> Given [GHSA-f3cx-396f-7jqp](https://github.com/livewire/livewire/security/advisories/GHSA-f3cx-396f-7jqp), any chance this could be backported to 2.x? This looks reasonably maintained branch and had a release around 3.5.2 as well.

+1

jetwes:
If I'm not totally wrong this has been fixed for v2, too. So the vulnerability is only in 3.0 to 3.5.2

PerryvanderMeer:
Isn't this fixed in https://github.com/livewire/livewire/commit/cd168c6212ea13d13b82b3132485741f82d9fad9?

RChutchev:
> Isn't this fixed in [cd168c6](https://github.com/livewire/livewire/commit/cd168c6212ea13d13b82b3132485741f82d9fad9)?

Yep, you're right, looks like that's fixed and requires a https://github.com/livewire/livewire/security/advisories/GHSA-f3cx-396f-7jqp to be updated too.

RChutchev:
Hi, @calebporzio, do you have access to update https://github.com/livewire/livewire/security/advisories/GHSA-f3cx-396f-7jqp/?
This CVE was also fixed in 2.12.7 by you.
This advisory https://github.com/advisories/GHSA-f3cx-396f-7jqp was updated but the livewire repository-related one is not.

angelej:
> Hi, @calebporzio, do you have access to update [GHSA-f3cx-396f-7jqp](https://github.com/livewire/livewire/security/advisories/GHSA-f3cx-396f-7jqp)? This CVE was also fixed in 2.12.7 by you. This advisory [GHSA-f3cx-396f-7jqp](https://github.com/advisories/GHSA-f3cx-396f-7jqp) was updated but the livewire repository-related one is not.

Hi @RChutchev,
Thank you for your commitment!
I've now updated the original security advisory.

PR #8651, opened 2024-07-14
Add `$wire.$hook()`

joshhanley:
## Scenario

Livewire offers the ability to access javascript lifecycle hooks by using `Livewire.hook()`.

Typically these would get used like the below `commit` hook:

```javascript
Livewire.hook('commit', ({ component, commit, respond, succeed, fail }) => {
    // Runs immediately before a commit's payload is sent to the server...
 
    respond(() => {
        // Runs after a response is received but before it's processed...
    })
 
    succeed(({ snapshot, effect }) => {
        // Runs after a successful response is received and processed
        // with a new snapshot and list of effects...
    })
 
    fail(() => {
        // Runs if some part of the request failed...
    })
})
```

## Problem

The problem with the above `commit` hook, is you don't want your hook to fire everytime any component updates.

As such you typically have to add a check for the component first, like:

```javascript
Livewire.hook('commit', ({ component, commit, respond, succeed, fail }) => {
    // return early if the components don't match
    if (component.id !== this.$wire.id) return

    // do stuff
})
```

# Solution

This PR adds a `$hook` option to the `$wire` object, that ensures are component level hooks are scoped to the correct component.

```javascript
$wire.$hook('commit', ({ component, commit, respond, succeed, fail }) => {
    // do stuff only relevant to this component
})
```

Any request level hooks will also still work as expected, as they're not scoped to the component.
```javascript
$wire.$hook('request', ({ uri, options, payload, respond, succeed, fail }) => {
    // do request stuff
})
```

This also has the added advantage that if being used inside an `Alpine.data()` component, you're not having to reach for the Livewire window object and instead can just use `this.$wire.$hook` locally, which feels much nicer!

```javascript
Alpine.data('myComponent', () => ({
    init() {
        this.$wire.$hook('commit', ({ component, commit, respond, succeed, fail }) => {
            // do stuff only relevant to this component
        })
    }
}))
```

I have also added `$wire.$hook` to the `$wire` object docs.

Hope this helps!


calebporzio:
Thanks for this Josh! I think I'm down to merge this, it makes sense. Can you add a test for it as well as logic to "clean it up" (prevent the hook callback from running if the component is removed - Using something like: `component.addCleanup(() => ...)`)

joshhanley:
@calebporzio yep can do!

joshhanley:
@calebporzio this PR is now ready to go. I've added tests for this and added it to the component cleanup. I'm still returning the cleanup function as users may want to manually clean up the listener themselves while the component is still active.

RoccoHoward:
@joshhanley THANK YOU Been frustrated by this for so long and this is perect.

PR #8653, opened 2024-07-15
Fix noscript tag in head when using navigate

schonhoff:
Hey!

Yes, you can find the discussion here: https://github.com/livewire/livewire/discussions/6279

Yes, the branch name is `failing-test-for-noscript-tag`

No, the changes are only related to my feature.

Yes, it includes a failing test.

After changing to Livewire v3 our company tried to use wire:navigate on some pages and quickly run into an issue. If you are using the `noscript` HTML-Tag in `head`, you cannot use `wire:navigate` nor can you use `$this->redirect(URL, navigate:true)` because it will trigger the `noscript`-tag.

Thanks for contributing! 🙌
I could not find a solution for the problem. Maybe someone can help out? :-)


lrljoe:
I can agree that this behaviour occurs.

If I add:
```
<noscript> <meta http-equiv="refresh"  content="0;url=/dashboard"> </noscript>
```
Then it'll force to "/dashboard" regardless on wire:navigate.

schonhoff:
Hey @lrljoe ,

we are currently redirecting to a page without javascript used, because some clients wants a single page to be completely functional without javascript. Showing an element won't resolve the client request.

However, it is not the redirect, which isn't working. Even if your are using css to show/hide elements with the noscript tag it won't work. The fix from Caleb should to it. If you want me to add another test just for the css I would provide it asap.

I don't know why the workflow for PHP:8.3 isn't working as expected.

joshhanley (approved):
@schonhoff thanks for the PR! The failing test suite was just because we occasionally have flaky tests. I've merged main and everything looks good. I tested this without Caleb's fix and your test fails as expected, and then when I applied the fix, the test passes. So this PR is good to merge @calebporzio

schonhoff:
Hey @calebporzio,

I tried this out after upgrading and after using wire:navigate it works. But on the second page there is no noscript tag anymore. Here are two screenshots:

First Page:
image

On the second page (after using wire:navigate on a page component):
image

Maybe you need a second look at this?

joshhanley:
@schonhoff I would say it's fine, mainly because if someone can see the noscript tag, then they have javascript disabled and therefore `wire:navigate` won't work anyway (as it depends on javascript). What do you think?

schonhoff:
@joshhanley okay, that is a fair point! If they disable Javascript between the sides I cannot help them :D

PR #8657, opened 2024-07-17
Swallow error during test

nuernbergerA:
When faking `tmp-for-tests` a error is thrown that is handled but also bubbles up and leads to "ugly" output

```bash
[2024-07-17 13:08:13] testing.ERROR: Disk [tmp-for-tests] does not have a configured driver. {"exception":"[object] (InvalidArgumentException(code: 0): Disk [tmp-for-tests] does not have a configured driver. at /home/adrian/code/bauservices/vendor/laravel/framework/src/Illuminate/Filesystem/FilesystemManager.php:137)
[stacktrace]
...
```

This happens in a edgy-edge case when you have to call this in you test
```php
FileUploadConfiguration::storage();
TemporaryUploadedFile::createFromLivewire(...)
```

i've separated all refactoring steps into atomic commits:
* actual fix
* fixed a typo
* used arrow functions (available sind PHP 7.4)
* extracted disk to a variable (was called 3 times)

### DIFF
--- a/src/Features/SupportFileUploads/FileUploadConfiguration.php
+++ b/src/Features/SupportFileUploads/FileUploadConfiguration.php
@@ -11,7 +11,7 @@ class FileUploadConfiguration
     {
         if (app()->runningUnitTests()) {
             // We want to "fake" the first time in a test run, but not again because
-            // ::fake() whipes the storage directory every time its called.
+            // ::fake() wipes the storage directory every time its called.
             rescue(function () {
                 // If the storage disk is not found (meaning it's the first time),
                 // this will throw an error and trip the second callback.
--- a/src/Features/SupportFileUploads/FileUploadConfiguration.php
+++ b/src/Features/SupportFileUploads/FileUploadConfiguration.php
@@ -9,20 +9,22 @@ class FileUploadConfiguration
 {
     public static function storage()
     {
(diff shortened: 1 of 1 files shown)


src/Features/SupportFileUploads/FileUploadConfiguration.php
```diff
@@ -9,19 +9,22 @@ class FileUploadConfiguration
...
+            rescue(
                 // If the storage disk is not found (meaning it's the first time),
                 // this will throw an error and trip the second callback.
-                return Storage::disk(static::disk());
-            }, function () {
-                return Storage::fake(static::disk());
-            });
+                fn() => Storage::disk($disk),
+                fn() => Storage::fake($disk),
+                // swallows the error that is thrown on the frist try
```
edalzell, line 22: 'first'

edalzell:
Love this one Adrian, thanks

src/Features/SupportFileUploads/FileUploadConfiguration.php
nuernbergerA:
```suggestion
                // swallows the error that is thrown on the first try
```

nuernbergerA:
@calebporzio pls merge

PR #8664, opened 2024-07-19
Refactor cleanupOldUploads method to accept custom time intervals

ahmetsabri:
problem: The `cleanupOldUploads` method in the Livewire repository was originally designed to clean up old uploads based on a fixed threshold of one day. This approach lacked flexibility, limiting the ability to customize the cleanup interval based on different requirements.

solution: To address this limitation, the cleanupOldUploads method was refactored to accept two parameters: `$interval `and `$unit`. This change allows the method to dynamically adjust the cleanup threshold based on the specified time interval and unit.

### DIFF
--- a/src/Features/SupportFileUploads/WithFileUploads.php
+++ b/src/Features/SupportFileUploads/WithFileUploads.php
@@ -99,7 +99,7 @@ trait WithFileUploads
         }
     }
 
-    protected function cleanupOldUploads()
+    protected function cleanupOldUploads(int $interval = 24, string $unit = 'h')
     {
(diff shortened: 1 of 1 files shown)


lrljoe:
I think you'll probably need to add some tests that validate the functionality works as expected

joshhanley:
@ahmetsabri thanks for the PR! `cleanupOldUploads()` method is currently protected, so it can't be called from outside. So how would the changes in the PR be used? Please provide code examples of how you would make use of this. Thanks!

PhiloNL:
I am closing this PR until Josh's question is answered.

PR #8672, opened 2024-07-22
[V2] Fix fingerprint errors when testing

xtfer:

We use an call to $this->authorize() in every boot method to control access to our livewire resources. This has the unfortunate side effect of breaking testing wherever the call fails, due to how to TestableLivewire handles the fingerprint.

See attached

One change only

Not currently. Please advise what is required beyond existing TestableLivewire tests

TestableLivewire makes a mock request to a dummy page with the component mounted on it. This call ignores Exceptions, which should mean that it bypasses access tests in the boot hook, however it still tracks those Exceptions and doesn't set the Fingerprint if one is found. This prevents us from making subsequent calls to the component (e.g via call() or set()), which should throw an AuthorizationException(), but since the fingerprint is missing it fails on an internal error.

This PR fixes the issue with the fingerprint.

Thanks for contributing! 🙌

### DIFF
--- a/src/Testing/TestableLivewire.php
+++ b/src/Testing/TestableLivewire.php
@@ -83,7 +83,7 @@ class TestableLivewire
 
         $this->lastResponse = $this->pretendWereMountingAComponentOnAPage($name, $params, $queryParams);
 
-        if (! $this->lastResponse->exception) {
+        if (! $this->lastResponse->exception && !empty($this->rawMountedResponse)) {
             $this->updateComponent([
                 'fingerprint' => $this->rawMountedResponse->fingerprint,
                 'serverMemo' => $this->rawMountedResponse->memo,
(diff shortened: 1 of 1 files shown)


joshhanley:
@xtfer thanks for the PR! Could you please add a test for this (that fails without this fix) that is structured similar to the issue you were running into? Thanks!

xtfer:
@joshhanley Can you provide some guidance on how you'd expect that? I'm not sure how to test TestableLivewire _while testing._

joshhanley:
@xtfer do just the same as in your app. I would just create a component with the authorisation logic in boot like you have (but as simple as possible) and a test with `Livewire::test()` similar to one of your tests (again as simple as possible) that was failing due to the fingerprint error. That way the test should fail before your fix is applied and pass after the fix 🙂

xtfer:
@joshhanley Okay, I'll give it a shot.

joshhanley:
@xtfer thanks! Have any issues, just push up where you go to with the test so we can have a look.

joshhanley:
Closing due to inactivity. @xtfer please open a new PR if you manage to add a test. Thanks!

PR #8682, opened 2024-07-24
Support hooks within form classes

gdebrauwer:
If you have form property on your livewire component, you can only react to changes within that form by defining update hooks in the livewire component. That is all fine and good if you use that form class only in one livewire component, but if you use it in multiple livewire components, you must create a lot of duplicate code.

```php
// Before

class FirstComponent extends Component
{
    public SomeForm $myForm;

    public function updatingMyFormSomeProperty()
    {
		// ...
    }

    public function updatedMyFormSomeProperty()
    {
		// ...
    }
}

class SecondComponent extends Component
{
    public SomeForm $myForm;

    public function updatingMyFormSomeProperty()
    {
		// ...
    }

    public function updatedMyFormSomeProperty()
    {
		// ...
    }
}

class SomeForm extends Form
{
    public $someProperty;
}
```

This PR fixes that by allowing you to define hook methods in your form class.

```php
// After

class FirstComponent extends Component
{
    public SomeForm $myForm;
}

class SecondComponent extends Component
{
    public SomeForm $myForm;
}

class SomeForm extends Form
{
    public $someProperty;

    public function updatingSomeProperty()
    {
		// ...
    }

    public function updatedSomeProperty()
    {
		// ...
    }
}
```

The following hooks are supported:
- `hydrate`
- `hydrate{propertyName}`
- `dehydrate`
- `dehydrate{propertyName}`
- `updating`
- `updating{propertyName}`
- `updated`
- `updated{propertyName}`

```php
class SomeForm extends Form
{
    public $someProperty;

    public function hydrate()
    {
    }

    public function hydrateSomeProperty()
    {
    }

    public function dehydrate()
    {
    }

    public function dehydrateSomeProperty()
    {
    }

    public function updating($property, $value)
    {
    }

    public function updatingSomeProperty()
    {
    }

    public function updated($property, $value)
    {
    }

    public function updatedSomeProperty()
    {
    }
}
```


joshhanley:
@gdebrauwer thanks for the PR! Love the idea as I wish my FormObjects had update hooks. My only concern for this that if update hooks work then people might expect other lifecycle hooks to work when they don't (like mount).

lrljoe:
Definitely torn on this, agree with @joshhanley that there's an expectation that when two of the lifecycle hooks work, that they should all work, but logically some are irrelevant when it comes to a Form.

I think avoiding the confusion is probably smoother at this point in time though, as much as I'd love to have the hooks in my Form Objects.  The overhead of explaining why only some hooks work would probably suck up a lot of time.

gdebrauwer:
I added support for more hooks. The only ones that are not supported in this implementation are:
- mount
- boot
- exception

The mount and boot methods are a bit tricky. When do you call those: after or before the mount/boot methods of the component? 🤔

lrljoe:
> I added support for more hooks. The only ones that are not supported in this implementation are:
> 
> * mount
> * boot
> * exception
> 
> The mount and boot methods are a bit tricky. When do you call those: after or before the mount/boot methods of the component? 🤔

Why exclude exception out of interest?  I know some people had issues with Computed Properties, Forms and exceptions #8373

fotrino:
any updates on this ? 
it would be awesome to have the `updated` hooks logic within the forms.

francoism90:
@fotrino I think most developers were at Laracon.

Let's hope most developers have time the upcoming week. :)

SyedMuradAliShah:
@gdebrauwer Thank you, this will make our lives easier and keep the components clean. In my case, I’m using the `AdminForm`, which is designed for adding new admins, updating admins, and updating profiles. This helps avoid unnecessary uploads of `.php` files. However, I'm using `updatedImage()` in every component, which results in some repetitive code.

[Link to code](https://github.com/SyedMuradAliShah/Admin-Flare/blob/081b07705ea683a8e1c7f57120027eeec1f6aa3d/app/Livewire/Admin/Admins/EditAdmin.php#L62C1-L81C6)
 ```
   public function updatedImage(): void
    {
        try {
            $this->validateOnly('image', [
                'image' => 'image|max:1024',
            ], [
                'image.image' => 'The file must be an image.',
            ]);
        } catch (ValidationException) {
            if ($this->image instanceof TemporaryUploadedFile) {
                $this->image->delete();
            }

            $this->reset('image');

            return;
        }
    }
```

**A request**: Can you create a PR for dispatch a browser event form from???

lrljoe:
> @gdebrauwer Thank you, this will make our lives easier and keep the components clean. In my case, I’m using the `AdminForm`, which is designed for adding new admins, updating admins, and updating profiles. This helps avoid unnecessary uploads of `.php` files. However, I'm using `updatedImage()` in every component, which results in some repetitive code.
> 
> [Link to code](https://github.com/SyedMuradAliShah/Admin-Flare/blob/081b07705ea683a8e1c7f57120027eeec1f6aa3d/app/Livewire/Admin/Admins/EditAdmin.php#L62C1-L81C6)
> 
> ```
>   public function updatedImage(): void
>    {
>        try {
>            $this->validateOnly('image', [
>                'image' => 'image|max:1024',
>            ], [
>                'image.image' => 'The file must be an image.',
>            ]);
>        } catch (ValidationException) {
>            if ($this->image instanceof TemporaryUploadedFile) {
>                $this->image->delete();
>            }
> 
>            $this->reset('image');
> 
>            return;
>        }
>    }
> ```
> 
> **A REQUEST**: Can you create a PR for dispatch a browser event from FROM???

Are you not using a Trait for that?

calebporzio:
Sounds good! @joshhanley - can you review this and merge? thanks (also documentation will be needed)

PR #8690, opened 2024-07-25
Handle Long Filenames

samharvey44:

A user of my application came across this error today when trying to upload a file with a long filename:
`Unable to retrieve the file_size for file at location: livewire-tmp`
Telling users to reduce filename sizes before uploading doesn't make for great UX, so I went looking for a solution, and found this discussion: https://github.com/livewire/livewire/discussions/2813
The discussion led to a PR, but only to introduce validation for filename length, rather than dealing with the limitation itself. It was also never merged.

Yes

No

Yes

The idea behind this fix is to identify whether a file is going to exceed system filename size limits (typically 255 chars) when base64 encoded, and if so, truncate the name of the file before encoding and store the actual pre-truncation name of the file in a separate meta directory for usage in calls to `->getClientOriginalName()`.
The new directory for truncated filename references will be cleaned up on file 'expiry' the same way as uploaded files are now.
Note - I had to add an overriding `setFileWithoutFakeMeta` method to the base `Testable` class which I don't really like, but otherwise, the tests were appending fake meta parameters for hash and size to uploaded filenames, which made it impossible to properly test the filename length truncation changes I have made, due to these extra parameters being added on which will not happen in actual application environments.


samharvey44:
There is an open and much more recent discussion on this issue here too: https://github.com/livewire/livewire/discussions/3021

phattarachai:
Hi,

I'm facing this issue as well. Is there an update on this pull request?

Thanks.

DanielGSoftware:
Same issue here, would be great if this PR can be looked at 🙏

uandco:
Same here, uploading files with a filename that is 255 chars long results in `Livewire\Features\SupportFileUploads\TemporaryUploadedFile`'s `originalName` to default to `livewire-tmp` because the original name can't be hashed, nor the file can be stored., hence the filename becoming the `temporary_file_upload.directory` livewire's config value. Trying to retrieve it of course fails.

maikezan:
Same issue here, different error:

PHP 8.3.3
Laravel 11.34.2
Livewire 3.5.13

If the filename exceeds 155 characters i get the following error and there is no uploaded  file in livewire-tmp:

```
Livewire \ Features \ SupportFileUploads \ FileNotPreviewableException
File with extension "" is not previewable. See the livewire.temporary_file_upload.preview_mimes config.
```

OrakFR:
Same issue here. Livewire doesn't seem to handle the file when its name is too long.

From other discussions, it seems that the issue with long filenames could be related to how the original filename is embedded in the file path.
https://github.com/livewire/livewire/blob/0b5739cdca43114e0d8d699eb7b44e36b1478b60/src/Features/SupportFileUploads/TemporaryUploadedFile.php#L175-L182

Couldn't we avoid storing the original filename in the path itself and instead store it as a property on the [TemporaryUploadedFile ](https://github.com/livewire/livewire/blob/main/src/Features/SupportFileUploads/TemporaryUploadedFile.php)class?

For example, the generateHashNameWithOriginalNameEmbedded() function could be adjusted to only generate the hash, and the original filename could be stored separately as a class property for example.

This way we could simplify the file path generation and avoid issues with long filenames, while still retaining access to the original filename when needed and might help ensure compatibility with long filenames without causing issues with file path limits.

Naoray:
Same issue here.

@samharvey44 what's the status of this PR? Anything I can help you with?

samharvey44:
@Naoray the PR has been open a while, and the issue remains in Livewire (as far as I know anyway - I'm not using Livewire for work anymore). I guess the team don't view this bug as a high priority, so it's not been given any attention.

Naoray:
@samharvey44 alright. Then you don’t have any hard feeling when I fork your branch and see if I can get it merged?!

samharvey44:
@Naoray absolutely not - please go ahead!

calebporzio:
closing this because we can just work off the newer one. thanks everyone

PR #8712, opened 2024-07-31
Fix typo

oliver-dvorski:


PhiloNL:
Hi Oliver,
Thanks for the PR. I don't think this is a typo. It explains what the line does in the present tense. So the line below the comment will authorize if the user can delete.

PR #8721, opened 2024-08-04
Make reset except public

liamduckett:
Hello,

Please bear with me, fairly new to open source 😅 

Wanted. I did not, as there appeared at already be a [relevant one](https://github.com/livewire/livewire/discussions/6633)

Yes (although I wasn't sure of naming conventions etc?)

Nope.

Sort of. The functionality is already tested, using a proxy method to circumvent the visibility. I have changed the relevant tests to no longer use this proxy.

Currently, a Component cannot call the `resetExcept` method on any of its Form properties due to it being protected.

```php
<?php

namespace App\Livewire;

use Livewire\Component;

class Dashboard extends Component
{
    public FooForm $fooForm;

    public function storeFoo(): void
    {
        // has properties bar, baz and qux
        $validated = $this->fooForm->validate();

        Foo::create($validated);

        $this->fooForm->reset(['bar', 'qux']);
        
        // ideally could be
        // $this->fooForm->resetExcept('baz'); 
    }
}
```

The discussion mentioned queries for any security concerns. I don't see this as a problem due to the existing implementation of `reset` being public.

I didnt make this change in `src/Concerns/InteractsWithProperties.php` (despite it having the same body as its `src/Features/SupportFormObjects/Form.php` counterpart) as this is implemented on the Component itself, so protected visibility is not a hinderance here.

Thanks for any time spent looking at this, I'd appreciate any feedback 🙂


liamduckett:
Can't see how this failure could be related to this PR?

joshhanley (approved):
@liamduckett thanks for the PR! The failing test suite was just because we occasionally have flaky tests. I've merged main and everything looks good.

PR #8727, opened 2024-08-07
Added post-morph event for each Livewire component

navnavnav:

I didn't create a discussion, but there appear to be a number of discussions where this is needed:
- https://laracasts.com/discuss/channels/livewire/livewire-v3-does-not-wait-until-dom-changes-are-finished-to-dispatch-events-on-js-side
- https://github.com/livewire/livewire/discussions/7062

Yes, in my forked repo. Branch name: `post-morph-event`

No, just a single line of JS, and a few sentences in the docs.

I did look for some preexisting tests that test the firing of `morph.*` events, but I couldn't find any, so I assumed they were not required.

Currently, Livewire 3 does not appear to offer any way to perform an action after a component has been morphed. 

It fires a single `morph` event for the component, then a series of `morph.*` events for individual elements within the component. But what if we want to perform an action after a component has been morphed? There is no event for that. This PR addresses this by adding a new `morphed` event.

Since Livewire 3 replaced `dispatchBrowserEvent()` with `dispatch()`, there is no longer a guarantee that the browser events will be handled after Livewire has finished updating the DOM. I hope the changes in this PR will serve as a partial or complete solution for those of us who were dependent on that behavior.

Thanks for contributing! 🙌


aoxborrow:
+1. I was looking for a `rendered` event similar to the PHP lifecycle hook. ended up [adding a custom event](https://github.com/livewire/livewire/discussions/8578#discussioncomment-10401093), but feels like this should be baked in.

PhiloNL:
Thanks for the PR! 😄 Could you add some tests to cover the `morphed` event? Thanks!

calebporzio:
Merged thanks!

PR #8734, opened 2024-08-09
Add New assertDontSeeText Method

kellerjmrtn:

### 1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?
No previous discussion, willing to open one if desired

### 2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed)
Yes

### 3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
No

### 4️⃣ Does it include tests? (Required)
No, as it only changes test methods themselves

### 5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

Livewire already contains many useful testing assertions as well as their negative counterparts, `assertSee` / `assertDontSee`, `assertSeeHtml` / `assertDontSeeHtml`, etc. However, `assertSeeText` exists without an `assertDontSeeText`. Also Laravel itself maintains a very similar method https://laravel.com/docs/11.x/http-tests#assert-dont-see-text for testing responses. I figured this would be a useful addition to improve testability

The code I've added is exactly the same as `assertSeeText`, except it uses `assertStringNotContainsString` instead

I've also included an update to the docs. I noticed `assertSeeText` didn't have an entry there either, so I added that as well

#### Example Usage
```
<div data-attribute="some-example-string">
  Some text
</div>

// In tests
$testable->assertDontSee('some-example-string'); // Fails
$testable->assertDontSeeText('some-example-string'); // Passes

$testable->assertDontSee('Some text'); // Fails
$testable->assertDontSeeText('Some text'); // Fails
```

Thank you!


calebporzio:
Thanks!

PR #8744, opened 2024-08-12
Fixed bug in `mergeQueuedUpdates()` function

navnavnav:

https://github.com/livewire/livewire/discussions/7872
https://github.com/livewire/livewire/discussions/8736

Yes, branch name: `morph-queued-updates-fix`

No

No

The `mergeQueuedUpdates()` function is broken. It incorrectly discards queued updates, when one of them contains an empty `updateValue`. 

This bug is breaking a fundamental function in Livewire 3, when used in conjunction with Alpine. It results in calls to `$wire.set()` failing silently. See the linked discussions for more.


js/component.js
```diff
@@ -92,7 +92,7 @@ export class Component {
         // priority against ephemeral updates that have happend since them...
         Object.entries(this.queuedUpdates).forEach(([updateKey, updateValue]) => {
             Object.entries(diff).forEach(([diffKey, diffValue]) => {
-                if (diffKey.startsWith(updateValue)) {
+                if (typeof updateValue === 'string' && updateValue.length > 0 && diffKey.startsWith(updateValue)) {
```
navnavnav, line 95: BTW, as mentioned in the discussion I raised (https://github.com/livewire/livewire/discussions/8736), I'm still not very confident in this code. I don't know why it's checking for the `updateValue` in the `diffKey`. Are there instances where the `diffKey` can contain with the `updateValue`? If not, perhaps it should be checking if the `diffKey` starts with the `updateKey`? So `diffKey.startsWith(updateKey)`? 

Happy to update this, but I need confirmation from someone who knows the codebase better than me.

PR #6803, opened 2023-09-13
Implements 2 new helper methods in the LivewireManager to set the urls for upload-file and preview-file routes.

felahdab:
1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

Yes: I started discussion #6771. However, I received no answer but since I am confident that there are use cases for these 2 additional helpers, I am submitting this simple PR.

2️⃣ Did you create a branch for your fix/feature? (Master branch PR's will be closed)

Yes

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No. Only one change.

4️⃣ Does it include tests? (Required)

No: but the other helpers (setUpdateRoute and setScriptRoute) do not have their own tests neither.

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

The 2 new helpers allow the developer to do:

```php
Livewire::setUploadFileRoute(function ($handle) {
    return Route::post('/livewire/a/random/route/to/upload-file', $handle)
        ->middleware('web');
});

Livewire::setPreviewFileRoute(function ($handle) {
    // Here, the provided url should contain a {filename} parameter for model binding to work.
    return Route::get('/livewire/z/random/route/to/preview-file/{filename}', $handle)
        ->middleware('web');
});
```

This is a complement to the setUpdateRoute and setScriptRoute methods which already allow the developper to adjust the routing of livewire requests. I reproduced the style used in seUpdateRoute and setScriptRoute which makes the new helpers consistent and allow the developer to fully customize the route definition, adding whatever middleware he wants in particular.

These new methods register the new route into the Router and name them as livewire.upload-file and livewire.preview-file. The URL generation used in the SupportFileUpload and SupportFilePreview features actually rely on the fact that the generated URL depends on the last declared route (when multiple routes have the same name). However, Laravel is missing a feature to remove a Route from the Router (which would be a safe addition to these 2 helpers).

Despite that, I hope you will find this PR suitable for merging.

PR #8752, opened 2024-08-14
Made file-upload- and -preview-urls configurable

majortom731:
Configurable Upload URLs seem to be wanted/needed by others, see [https://github.com/livewire/livewire/pull/6803](https://github.com/livewire/livewire/pull/6803)

Yes

No

No

I have Laravel running under a subfolder and thus need to be able to customize my file upload and preview paths. So I just changed the hard-coded url paths to values fetched from config/livewire.php. I gave the formerly used fixed values as defaults, so if nothing is set in the config file, it will still work as expected.

### DIFF
--- a/src/Features/SupportFileUploads/SupportFileUploads.php
+++ b/src/Features/SupportFileUploads/SupportFileUploads.php
@@ -31,10 +31,10 @@ class SupportFileUploads extends ComponentHook
             }
         });
 
-        Route::post('/livewire/upload-file', [FileUploadController::class, 'handle'])
+        Route::post(config('livewire.upload_file_uri','/livewire/upload-file' ), [FileUploadController::class, 'handle'])
             ->name('livewire.upload-file');
 
-        Route::get('/livewire/preview-file/{filename}', [FilePreviewController::class, 'handle'])
+        Route::get(config('livewire.preview_file_uri','/livewire/preview-file/{filename}' ), [FilePreviewController::class, 'handle'])
             ->name('livewire.preview-file');
     }
 }


samlev:
Could this be solved by using `APP_DIR` in your `.env`? Surely your entire app would be running into issues with routing/url generation.

Alternatively, you could just define these routes yourself in your own `routes/web.php`, so long as you use the correct names and actions, it should override the ones defined by livewire.

emkcloud:
I've been working with Laravel, Livewire, Nova, and now also Flux PRO for a long time, and they all have the same issue: they don't consider route prefixes, which are crucial for those working with different servers and load balancers that distribute traffic. This is a problem that wastes a lot of time. From my experience, here's how I solved it—not a super elegant solution, but at least it works for me. I hope this can help someone partially solve their problem.

**Create un Custom Route Service Provider**

`class RouteServiceProvider extends ServiceProvider{}`

I work with 3 environment admin/backend/frontend and 3 distinct ELB (1 ELB for every environment)

```
protected function mapRoutesLivewire()
{
    /**
     * Application service provider function
     */
    protected function mapRoutesLivewire()
    {
        // Calculate current environment

        $environment = match (true)
        {
            eEnv::isAdmin()    => Environment::ADMIN,
            eEnv::isBackend()  => Environment::BACKEND,
            eEnv::isFrontend() => Environment::FRONTEND,

            default => false
        };

        // Define custom livewire routes

        if ($environment)
        {
            URL::defaults(['environment' => $environment]);

            foreach (Route::getRoutes() as $route)
            {
                if (Str::is('livewire/*',$route->uri()))
                {
                    $route->setUri('{environment}/'.$route->uri());

                    $route->where('environment',sprintf('%s|%s|%s', 

                        Environment::ADMIN,
                        Environment::BACKEND,
                        Environment::FRONTEND

                    ))->middleware(['livewire']);
                }
            }
        }

        // Define custom livewire namespace if necessary

        if (eEnv::isAdmin())
        {
            Config::set('livewire.class_namespace', 'App\\Nova\\Livewire');
        }
    }
}
```

I get with success for example

admin/livewire/upload-file
backend/livewire/upload-file
frontend/livewire/upload-file

php artisan route:list

```
GET|HEAD  {environment}/livewire/livewire.js .. Livewire\Mechanisms › FrontendAssets@returnJavaScriptAsFile
GET|HEAD  {environment}/livewire/livewire.min.js.map ............ Livewire\Mechanisms › FrontendAssets@maps
GET|HEAD  {environment}/livewire/preview-file/{filename} livewire.preview-file › Livewire\Features › FileP…
POST  {environment}/livewire/update livewire.update › Livewire\Mechanisms › HandleRequests@handleUpdate
POST  {environment}/livewire/upload-file livewire.upload-file › Livewire\Features › FileUploadControll…
```

felahdab:
Hi.

You solution is very elegant IMHO.

I actually didn't know that one could modify the uris of the already registered routes. Obviously, it depends on the service provider being executed after the original route registration bit that's rather easy to control.

It seems like a very nice workaround that one can use for a lot of use cases...

I happen to have the same issue again with the import and export routes in Filament. I will give it a try for that scenario and report my success here.

Thanks a lot for sharing this.

Regards.
Florian.

amimhayden22:
this is work. thank you sir

emkcloud:
After quite some time, I finally decided to package a solution for Livewire and Livewire Flux to enable more dynamic routing. I often work with clients where complex configurations are required to handle firewalls and load balancing based on URL prefixes, and this package aims to simplify that.

I've tested the solution in a few environments and it seems to meet my needs, though I’m not sure how it will perform in different setups. Hopefully, it will be useful to others as well — and any feedback on how it works in your context would be greatly appreciated.

Thanks!

https://github.com/emkcloud/livewire-tweak

calebporzio:
Closing this PR as it's been open for a long time and now has merge conflicts. If still relevant, please re-submit with an updated branch and tests. Thank you!

PR #8761, opened 2024-08-19
Docs: Remove commit prop from commit.prepare hook

wpoortman:
I've been reviewing the documentation for a feature I’m working on and needed to retrieve details from the commit property using `Livewire.hook('request.commit', ({ component, commit }) => {...})`. However, I noticed that commit is always undefined. After checking the code, this behavior makes sense.

As a result, I've decided to update the documentation to reflect this, at least until the core team decides whether to reintroduce the commit property in the hook callback in the future and to update to docs accordingly.

For anyone still looking to access commit details, you can use `Livewire.hook('commit', ({ component, commit, respond, succeed, fail }) => {...})` as an alternative.

### DIFF
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -552,7 +552,7 @@ These hooks expose `commit` objects. You can learn more about their schema by re
 The `commit.prepare` hook will be triggered immediately before a request is sent to the server. This gives you a chance to add any last minute updates or actions to the outgoing request:
 
 ```js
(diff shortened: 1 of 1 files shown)


samlev:
I wonder if it would be better to make this correction to the documentation, or to just update the line that fires the hook so that it includes the commit:
```js
hook('commit.prepare', {component: this.component, commit: this}}
```

PR #8766, opened 2024-08-19
Promote protected validation methods in docs

chillbram:
When the `rules()` method has public visibility, users can freely retrieve them. Simply typing `Livewire.first().call('rules')` in the console extracts the exact validation rules used by the component. This can leak valuable information and removes security through obscurity. I don't think this is expected behaviour as calling other public Livewire methods, like `mount()` or `reset()`, have been disabled.

That's why I propose promoting protected visibility for `rules()` (and related functions `messages()` and `validationAttributes()`). There is an alternative and that is to treat these methods differently from others and to prohibit calling these methods from the front-end similar to the previously mentioned lifecycle/internal methods. However, I do not know if the current behaviour is necessary for a valid use case somewhere.

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -443,7 +443,7 @@ class PostForm extends Form
 
     public $content = '';
 
-    public function rules()
+    protected function rules()
     {
         return [
             'title' => [
@@ -490,7 +490,7 @@ class PostForm extends Form
 
     public $content = '';
 
-    public function rules()
+    protected function rules()
     {
         return [
             'title' => [
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -301,7 +301,7 @@ class CreatePost extends Component
 
     public $content = '';
 
-    public function rules()
+    protected function rules()
     {
         return [
             'title' => 'required|min:5',
@@ -400,7 +400,7 @@ class CreatePost extends Component
 
     public $content = '';
 
-    public function rules() // [tl! highlight:6]
+    protected function rules() // [tl! highlight:6]
     {
         return [
             'title' => Rule::exists('posts', 'title'),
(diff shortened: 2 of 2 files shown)


lrljoe:
For the most part, a good UX will **tell** the end-user what is expected in a field, otherwise they're going to have a hard time determining what to put in the field.

I'm not sure what security element you're referring to really?

chillbram:
This is not about front-end UX; that remains unchanged. 

Malicious users, however, have the possibility of finding attack vectors by getting access to the exact validation rules that are normally obfuscated in error messaging. For example, the names of rule objects may reveal unknown business rules. It can leak the FQN for Rule objects. It reveals all valid values for rules such as `in:` or `regex:`. It reveals database table and column names in rules such as `exists:` and `unique:`.

I have no idea of the extent to which this could be abused in certain scenarios. I don't necessarily expect huge problems to stem from it, but setting a more secure default for no extra cost seems like a logical move to me.

lrljoe:
This aligns with the visibility of the Laravel "rules"/"attributes" etc, deviating from it may present issues in future.

Suggest you also raise this against the Laravel Framework.

chillbram:
As far as I know the implementation of `rules()` in Livewire is completely detached from any Laravel internals, although I might be mistaken on that. If it is, might need to take a look at that but I've never had problems with protected visibility in any of my projects.

I don't think there actually is a vulnerability in these methods in the Laravel contexts as these functions can't be called by a user. The issue arises because Livewire inadvertently exposes the method to the end user.

lrljoe:
Don't get me wrong, I agree that it SHOULD be protected, just as added safety.

However - consistency is key etc, and removing barriers to entry for migrating from Laravel to Livewire.

PR #8778, opened 2024-08-23
Added quotation marks to Dispatching events from Blade template Documentation

laurissss:
Added Quotation marks so the dispatch works with Uuids

Its needed if you follow the current docs it won`t work with Uuids

No

No

No

I added Quotation marks in the Code Snippet of the Documentation for Dispatching events from Blade template. So the dispatch will work if you are using a uuid

Thanks for contributing! 🙌

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -319,7 +319,7 @@ class CreatePost extends Component
 You can dispatch events directly from your Blade templates using the `$dispatch` JavaScript function. This is useful when you want to trigger an event from a user interaction, such as a button click:
 
 ```blade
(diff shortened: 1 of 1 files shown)


PhiloNL:
Thanks for the PR but in this case I'm going to leave it as is. There are already other examples that indicate that strings should be quoted, for example:

```blade
<button @click="$dispatch('post-created', { title: 'Post Title' })">...</button>
```

I would recommend the following as well, this will automatically ensure everything is escaped correctly:

```blade
<button wire:click="$dispatchTo('posts', 'show-post-modal', @js(['id' => $post->id]))">
    EditPost
</button>
```

PR #8784, opened 2024-08-26
Allow dynamic computed keys

francoism90:

Discussion already exists: https://github.com/livewire/livewire/discussions/8532

I also think it should be using `crc32` to avoid collisions when using the same component name and `cache: true`. But maybe there is a reason why the `crc32` isn't included? :)

Yes: `[dynamic-key-computers](https://github.com/francoism90/livewire)`

No

Yes, is it possible to actually test if caching is working? Like creating a counter?

See discussion. :)

For instance:

```php
public Page $page;

#[Computed(cache: true, key: 'page-components-{page.id}')]
public function pageComponents(): array
{
    // build something dynamically here
    return [..];
}
```
This would be useful to flush the cache when a page is updated in the admin panel.

It basically allows everything a session attribute already offers: https://livewire.laravel.com/docs/session-properties#setting-a-custom-key 

Thanks for contributing! 🙌


lrljoe:
So, I think it'd be worth (if extending this), to take into account the Auth ID, to mitigate against a User seeing something they shouldn't see.

LarsWiegers:
@francoism90 Thanks for creating the pr! Would love to see this merged as our current application could benefit from this.

lrljoe:
I think the other issue with this at present, is it requires that something is set, when that might not always be the case, in which case it should fail gracefully.

E.g. in your test example, you expect that $post has an "id", when that may not be the case

francoism90:
@lrljoe In the given example, I was assuming route-binding. Meaning it would fail before being able to call a computed value.

But you're indeed correct about certain situations this would fail. However this would always be for a computed value.

With this dynamic key approach, it would be possible to define the same caching key, based on an unique id. For instance, you could cache something for an user, or for model (e.g. tag-10-count).

lrljoe:
Apologies, on mobile at the mo, but is it not sensible to still prefix the dynamically generated key name with lw_computed or similar to avoid any cache conflicts and provide distinction?

Otherwise you may run into conflicting keys that return unexpected types.

francoism90:
@lrljoe You mean always prepend `lw_computed` to the key? Yeah, it's possible. :)

I copied the same logic from [Session](https://livewire.laravel.com/docs/session-properties#setting-a-custom-key), this also doesn't seem to prepend `lw` when setting a custom key, see: https://github.com/livewire/livewire/blob/58730e6053837aa889c19e59238e1b789bb87bf6/src/Features/SupportSession/BaseSession.php#L45.

If this is wanted, I could add them as well. But I think it's also valid to 'conflict' them.
You could setup a cache in an action/service for example, and letting Livewire calling that same key:

```php
#[Computed(cache: true, key: 'homepage-posts')]
public function posts()
{
    return app(GetHomePosts::class)->execute();
}
```

This is perfectly valid. If the `GetHomePosts` is called in a command, and holds the same data, it's works two-way.

francoism90:
@lrljoe I agree with this (also see my updated comment).

It was known to me that setting a session-key manually, it would simply make it global (without a prefix).

But I think indeed it may not be for everyone, and even myself could potentially cause an issue with something like this:

```php
#[Computed(cache: true, key: 'posts')]
public function posts()
{
    return Post::all();
}
```

And re-using the same key and thinking they are different.

I could add prefixes for both Session/Computed, but maybe adding a warning/note about setting a key manually would be better?

francoism90:
I don't know if I should report this as a bug @calebporzio @PhiloNL?

I'm thinking it could cause issues for users already now, e.g. using a custom key and the same one elsewhere, which shouldn't actually have the same output.

joshhanley:
@francoism90 I think a warning in the docs when manually setting a computed property key and also a session key would be sufficient if you don't mind PRing that 🙂

francoism90:
@joshhanley I'll do my best. :)

lrljoe:
Glancing at this again, it may be a nice to have  to also allow dynamic properties in the tags for cached props, not that this would block the PR

francoism90:
@lrljoe Sorry, at the moment I don't have time to contribute/improve the PR.

Thanks for creating the docs. :)

francoism90:
I don't see this getting merged any time soon, and unfortunately lack the time to improve this at the moment.

If someone ones to give this another shot, you can see the diff.

PR #8793, opened 2024-08-29
Fix csp nonce causing duplicate Alpine/Livewire instance

dorkster100:
As i understand CSP nonce causes same issues during wire:navigate as data-csrf in this PR(https://github.com/livewire/livewire/pull/6493), the fix is already there, need to just add the nonce into array.


dist/livewire.js
```diff
@@ -7624,6 +7624,7 @@ ${expression ? 'Expression: "' + expression + '"\n\n' : ""}`, el);
   var oldBodyScriptTagHashes = [];
   var attributesExemptFromScriptTagHashing = [
     "data-csrf",
+    "nonce",
```
danie-ramdhani, line 7627: `dist` folder is contains all scripts that already compiled. Maybe you mean this file?
https://github.com/livewire/livewire/blob/main/js/plugins/navigate/page.js

Try run `npm run watch` to see if the problem is fixed, if it fixed then run `npm run build` to compile all the assets.

Can you also make a test if the problem is fixed? if it's related with `wire:navigate` you can make the test [here](https://github.com/livewire/livewire/blob/main/src/Features/SupportNavigate/BrowserTest.php)

dorkster100: my bad, thank you for pointing me in the right direction

dorkster100:
I added the test and did the fix correctly.

However my small bit of criticism here is that it was really hard to get tests to work, would be nice to have a better docs on how to get started, for instance i found that i need chrome browser, not chromium, for dusk tests to run. And for some reason I have some issues with sqlite, for example:
`SQLSTATE[HY000]: General error: 1 no such table: paginator_post_test_models`
And I still have not figured out how to fix it, however it is not necessary for this PR.

joshhanley (approved):
@dorkster100 thanks for the PR! Makes sense to me.

Yeah the comments change depending on how Alpine is included, whether through node modules or symlinked. Caleb and I typically have it symlinked. It's not required, but it will be why you were seeing the differences.

calebporzio:
thanks! good to merge @joshhanley

PR #8801, opened 2024-09-01
Failing Test: Incorrect mime types guessing

Jubeki:

At least I need it for testing files with an allowed file extension, but an not allowed mime type of the file: https://github.com/filamentphp/filament/discussions/14102

I am not sure, if my way of testing this, should be done in this, way and I am also willing to simply update my test case, if there is a better method.

Yes.

No.

It includes a failing test.

I want to test in a form, that a user can not upload a file with a correct file extension, but in incorrect mime type. The issue is, that in the TemporaryUploadedFile Class of Livewire, the mime type is guessed incorrectly (based on the file extensions, instead of the provided mime type with UploadedFile).

I would expect that TemporaryUploadedFile uses the mime type which is provided during upload (In this case with UploadedFile), as an additional fallback, before going to guess the mime type based on the extension.

Here is the relevant code snippet in Livewire, were this probably must be adjusted.

https://github.com/livewire/livewire/blob/597a2808d8d3001cc3ed5ce89a6ebab00f83b80f/src/Features/SupportFileUploads/TemporaryUploadedFile.php#L55-L68

Thanks for contributing! 🙌


francoism90:
https://github.com/thephpleague/mime-type-detection

I think the check is actually correct. It first checks the file, and if it's getting steamed, it will try based on extension.

Jubeki:
What I mean is the following:

If you create a uploaded file for testing like the following

```php
$file = UploadedFile::fake()->create('file1.jpg', 1024, 'application/pdf');
```

And then call directly `getMimeType()`:
```php 
dd($file->getMimeType());
```

You will get `application/pdf`, were as if you use the same in an Livewire Component, it will create a copy and use the class `TemporaryUploadedFile`, which will not return the mime `application/pdf` but `image/jpg`.

That is the Problem I am having.

joshhanley:
I've worked out what is going on with this issue, what happens is Laravel's fake file no longer exists after Livewire uploads it, as it gets stored into Livewire's `livewire-tmp` directory as an actual file. The problem is the file is actually an empty file and the filesystem cannot use it to detect the mimeType so it falls back to using the file extension as the mimeType.

Thankfully we already had provision to send fake file data through the `TemporaryUploadedFile` path, so I have added the fake file mimeType to this. I then updated the `TemporaryUploadedFile` to use the fake mimeType data if running in unit tests (the same as how size and hashName work).

Jubeki:
Thanks @joshhanley !

PR #8819, opened 2024-09-08
Add tests for duplicated query strings when using BaseUrl(keep: true) with array-type parameters on link navigation

andrewdwallo:

Yes, this test highlights a potential issue in Livewire when using `BaseUrl(keep: true)` with array-type query parameters. While there is no discussion yet, this PR serves to demonstrate the bug for further investigation.

A key purpose of URL query parameters is to allow sharing or bookmarking of specific states. The duplication of query parameters breaks this functionality, as it leads to inconsistent or incorrect URLs. Specifically, this happens when clicking on a link that already contains query parameters and navigating to the page. The query parameters are duplicated, which causes unexpected behavior, and in my opinion, this is a significant issue.

Yes, a separate branch was created specifically for this issue.

No, this PR focuses solely on the failing and passing tests that demonstrate the bug.

Yes, this PR includes one failing test and two passing tests. The failing test demonstrates the bug where using the `BaseUrl(keep: true)` option with array-type query parameters results in duplicated query strings when navigating to a link that already contains query parameters. The two passing tests demonstrate that the issue is specific to the keep option with array-type parameters. When using string parameters with keep, or when not using keep, the issue does not occur.

This PR adds tests to demonstrate the issue where URL query parameters are duplicated when using the `BaseUrl(keep: true)` option with array-type query parameters, specifically when clicking on a link with query parameters already present. The query parameters are duplicated in the URL upon navigating to the page. For example, navigating to this link:
```
https://example.com/reports/account-transactions?filters[startDate]=2024-01-01&filters[endDate]=2024-09-05
```
results in the URL becoming:
```
https://example.com/reports/account-transactions?filters[startDate]=2024-01-01&filters[endDate]=2024-09-05&filters[startDate]=2024-01-01&filters[endDate]=2024-09-05
```

This issue only occurs when using the keep option with array-type parameters. The two passing tests show that the issue does not occur when using string-type query parameters with keep or when not using the keep option at all but with array-type parameters.

Thanks for contributing! 🙌


joshhanley:
@andrewdwallo thanks for the PR! I was investigating this issue and realised that it is the same issue raised in PR #8437. So I've merged these tests into that PR as it contains a fix. So going to close this PR and continue in PR #8437.

PR #8831, opened 2024-09-12
Add description for dispatching dynamic events

jaap:
Just added a bit of extra documentation around dispatching the dynamic event since I found myself searching for it for a couple of minutes. Seemed nice to have it in the documentation close to the listener example.


PhiloNL:
Thanks! :)

PR #8834, opened 2024-09-12
Correction getUpdateUri method parameter prompt

chuoke:
When using this method, the editor will have an error message underlined, although this will not cause the program to report an error, the visual sense is not good.

image


PhiloNL:
Thanks!

PR #8843, opened 2024-09-15
Add Methods for Component Attribute Assertion in Testing

CamKem:
This PR introduces the TestsAttributes trait, which provides a set of assertion methods to test the presence and values of attributes on Livewire components, including class-level, method-level, and property-level attributes.

It enhances the testing capabilities of Livewire by allowing developers to assert that specific attributes are present and have expected values, rather than traditionally having to use invade and loop through the values to assert the presence of the attributes. I believe it will really enhance component testing.

Methods that are being added by this PR:
- **Component Level (present anywhere in the component)**: assertComponentHasAttribute(string $attribute, string|array|null $value = null)
- **Class Level:** assertClassHasAttribute(string $attribute, string|array|null $value = null)
- **Method Level:** assertMethodHasAttribute(string $methodName, string $attribute, string|array|null $value = null)
- **Property Level:** assertPropertyHasAttribute(string $propertyName, string $attribute, string|array|null $value = null)

As the attributes can have a variety of strings, arrays & named properties placed within them, the logic in this PR is designed to handle all these edge cases in the new private method on the trait: attributeMatchesValue.

Note: the visibility on a couple of the Attribute properties have been changed to public in order to facilitate this to work.

Tests have been added to assert:
- Asserting attributes at different levels.
- Testing attributes with simple values, array values, and multiple fields.
- Handling multiple attributes of the same type on a single property or method.

The tests within the PR are the best example of how the code can be used, but I will place them here too:
```php
public function test_component_has_attribute()
{
    Livewire::test(NewComponent::class)
        ->assertComponentHasAttribute(Title::class, 'Class Level Attribute')
        ->assertComponentHasAttribute(Validate::class, ['required', 'integer'])
        ->assertComponentHasAttribute(On::class, 'fooEvent');
}

public function test_component_has_class_level_attribute_with_value()
{
    Livewire::test(NewComponent::class)
        ->assertClassHasAttribute(Lazy::class, ['isolate' => false])
        ->assertClassHasAttribute(Title::class, 'Class Level Attribute');
}

public function test_component_has_method_level_attribute()
{
    Livewire::test(NewComponent::class)
        ->assertMethodHasAttribute('jsMethod', Js::class)
        ->assertMethodHasAttribute('doubleFoo', On::class, 'fooEvent')
        ->assertMethodHasAttribute('barMethod', On::class, ['barEvent', 'bazEvent']);
}

public function test_component_has_property_level_attribute()
{
    Livewire::test(NewComponent::class)
        ->assertPropertyHasAttribute('bar', Session::class, ['key' => 'foo'])
        ->assertPropertyHasAttribute('foo', Validate::class, ['required', 'integer']);
}

public function test_component_has_property_with_multiple_attributes()
{
    Livewire::test(NewComponent::class)
        ->assertPropertyHasAttribute('bar', Validate::class, [
            'rule' => 'required',
            'as' => 'date of birth',
        ])
        ->assertPropertyHasAttribute('baz', Validate::class, [
            'rule' => 'required',
            'message' => 'This is a custom message',
        ])
        ->assertPropertyHasAttribute('baz', Isolate::class);
}
```

Yes, often I write manual tests using invade to assert the presence of attributes on & in a component class. No I didn't create a discussion first.

Yes

No

Yes

See above.

### DIFF
--- a/src/Features/SupportAttributes/TestsAttributes.php
+++ b/src/Features/SupportAttributes/TestsAttributes.php
@@ -3,6 +3,7 @@
 namespace Livewire\Features\SupportAttributes;
 
 use Illuminate\Support\Collection;
+use Livewire\Attributes\Computed;
 use Livewire\Attributes\Lazy;
 use Livewire\Attributes\On;
 use Livewire\Attributes\Session;
@@ -132,6 +133,7 @@ trait TestsAttributes
             Title::class => ['content'],
             Session::class => ['key'],
             Lazy::class => ['isolate'],
+            Computed::class => ['persist', 'cache', 'seconds', 'key', 'tags'],
         ];
 
         if (!isset($attributeComparisons[$attribute])) {
--- a/src/Features/SupportAttributes/TestsAttributes.php
+++ b/src/Features/SupportAttributes/TestsAttributes.php
@@ -143,8 +143,7 @@ trait TestsAttributes
         $fields = $attributeComparisons[$attribute];
 
         if (is_array($value)) {
-            $isAssociative = !array_is_list($value);
-            if (!$isAssociative) {
+            if (array_is_list($value)) {
                 $assigned = false;
                 foreach ($fields as $field) {
                     $actualValue = $this->getAttributeValue($item, $field);
--- a/src/Features/SupportAttributes/TestsAttributes.php
+++ b/src/Features/SupportAttributes/TestsAttributes.php
@@ -155,13 +155,13 @@ trait TestsAttributes
                     }
                 }
                 if (!$assigned) {
-                    PHPUnit::fail("Unable to assign flat array to any field of attribute {$attribute}.");
+                    return false;
                 }
             }
         } else if (count($fields) === 1) {
             $value = [$fields[0] => $value];
         } else {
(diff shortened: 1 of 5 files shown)


CamKem:
Open to any changes you think would be needed to facilitate this feature.

Also, I have written up some documentation to cover this, which I would like an opinion on before pushing it, however i'll wait to see if you want to proceed with this feature before doing any more work.

lrljoe:
My view is I'd exercise caution in reusing
assertClassHasAttribute

Some of these method names match up to deprecated phpunit method names, and may cause confusion down the line.

CamKem:
> My view is I'd exercise caution in reusing
> assertClassHasAttribute
> 
> Some of these method names match up to deprecated phpunit method names, and may cause confusion down the line.

Happy to adjust method names if/when Caleb or other maintainers weigh in & want to proceed.

joshhanley:
@CamKem thanks for the PR! Caleb and I discussed this and we think these assertions would be better off in userland. As such we're going to close this PR. Maybe it's worth PRing them to @christophrumpel's missing Livewire assertions package https://github.com/christophrumpel/missing-livewire-assertions

PR #8849, opened 2024-09-16
feat: Add support for dispatching Jobs from dispatch method

cconekin:
This is something that I needed due to a conflict upgrading from Livewire v2.5 to Livewire v3. When upgrading, I had many components that used the `DispatchesJobs` trait. When Livewire v3 introduced the `dispatch` method, this created a conflict because `$this->dispatch` was being used to dispatch jobs and not events. A discussion thread has not been created yet.

Yes. The branch is called `support-dispatching-jobs-with-dispatch-method`

The PR introduces a potential change to the `dispatch` method and a unit test.

Yes. There is a single unit test added. More testing should be added before merging, but I'm assuming that this PR will receive feedback with a better implementation approach, and 

I don't necessarily think that the approach in this PR is the best way to handle this and this definitely needs more testing before being accepted. The issue I experienced was a conflict with upgrading v2 components that used the `DispatchesJobs` trait. While I understand that using this trait in Livewire may not be common practice, it added to the complexity of upgrading Livewire. Additionally, to me, the `dispatch` method in v3 can be confusing when Events, Jobs, and Browser events are all using the same method name. I understand that this is what was decided in this discussion thread and I do not think that this change should be reversed. However, one way that the intended simplicity could be maintained would be to support multiple object types in the `dispatch` method, which is what I'm intending to introduce in this PR. 

Thank you 🌱

### DIFF
--- a/src/Features/SupportEvents/HandlesEvents.php
+++ b/src/Features/SupportEvents/HandlesEvents.php
@@ -3,6 +3,7 @@
 namespace Livewire\Features\SupportEvents;
 
 use function Livewire\store;
+use Illuminate\Contracts\Queue\ShouldQueue;
 
 trait HandlesEvents
 {
@@ -14,6 +15,10 @@ trait HandlesEvents
 
     public function dispatch($event, ...$params)
     {
+        if(method_exists($event, 'dispatch')) {
+            return $event->dispatch();
+        }
+
         $event = new Event($event, $params);
 
         store($this)->push('dispatched', $event);
--- a/src/Features/SupportEvents/HandlesEvents.php
+++ b/src/Features/SupportEvents/HandlesEvents.php
@@ -3,7 +3,6 @@
 namespace Livewire\Features\SupportEvents;
 
 use function Livewire\store;
-use Illuminate\Contracts\Queue\ShouldQueue;
 
 trait HandlesEvents
 {
--- a/src/Features/SupportEvents/UnitTest.php
+++ b/src/Features/SupportEvents/UnitTest.php
@@ -232,6 +232,7 @@ class UnitTest extends \Tests\TestCase
     public function test_dispatch_method_accepts_jobs()
     {
         Bus::fake();
+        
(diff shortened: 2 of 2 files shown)


cconekin:
> Why not just use the `dispatch()` helper method for dispatching jobs instead of overloading `$this->dispatch()`?

@samlev Thanks for the feedback here. Completely agree that using the helper is probably easier than making a carve-out in the `$this->dispatch` method. I'm just finding `dispatch` vs. `$this->dispatch()` from the `DispatchesJobs` trait somewhat confusing. For example, to me this would be confusing for a new developer:
```php
public function sendFoo(User $user): void
{
    // Dispatch the job
    $job = dispatch(new Foo($user));

   // Update the frontend
   $this->dispatch('success', $job->id);
} 
```

Maybe I'm complaining about nothing, but I think that there is a way to make this simpler, such as:
```php
public function sendFoo(User $user): void
{
    // Dispatch the job
    $this->dispatch(new Foo($user));

   // Update the frontend
   $this->dispatch('success', $job->id);
} 
```

My oppinion: Events, Browser Events, and Jobs are dispatched. Yet `$this->dispatch` only works for 2 of the 3. That seems to be conflicting. I understand that [this was discussed](https://github.com/livewire/livewire/discussions/5640) before the v3 change, so it's probably a little late in the game to introduce this change. 🤷

samlev:
Jobs, Batches, Chains, Notifications... There are a lot of things that are "dispatched" in Laravel, but I (personally) don't think that `$this->dispatch()` _in livewire_ should do anything other than events. It blurs the lines too much, and makes the behaviour unpredictable (or at least unexpected).

Aside from that, it would then be involving livewire in parts of Laravel that it has no reason to be involved in - firing events is one thing, but interacting with the queue is entirely different.

cconekin:
Closing after a convincing argument made by @samlev for not conflating the responsibilities of `$this->dispatch`.

Thanks @samlev for the feedback. 🙌

PR #8852, opened 2024-09-17
Fall back to page 1 if the page value is not numeric

mabdullahsari:
Last week, our application has been under heavy intrusion attempts. We received a lot of Livewire errors in Sentry, of which some could be fixed in Livewire Core. 

I think it only makes sense to reset the page number to 1 if an invalid page number is received, thus bulletproofing the incomplete validation in the `setPage` method.

This blind SQLi attempt eventually resulted in a page number of `-40` which kept crashing our Elastic instance. I'd say that in every pagination case, natural numbers is the only valid number set and thus integers should be excluded.

Thanks! 🫡


mabdullahsari:
I checked the failing browser test, but don't really know why it's failing. If someone could nudge me in the right direction, it'd be highly appreciated!

calebporzio:
This PR breaks cursor pagination where the page is not an integer but still needs to be set. Closing - please re-open with a different solution.

PR #8853, opened 2024-09-17
Fix TypeError on loadModel() /  Return null for deleted public legacy model

cbz1991:

Yes! No, but there are already discussions about this issue: [7987](https://github.com/livewire/livewire/discussions/7987), [7969](https://github.com/livewire/livewire/discussions/7969), etc

Yes!

No!

Yes!

I changed` $model = $query->first(); ` to the conditional `if ($query->first() !== null) { $model = $query->first(); }` to ensure that the query result is not null before assigning it to $model. This prevents the following error `Livewire\Features\SupportLegacyModels\EloquentModelSynth::setDataOnModel(): Argument #1 ($model) must be of type Illuminate\Database\Eloquent\Model, null given` triggered if the model doesn't exist anymore.

Thanks for contributing! 🙌


src/Features/SupportLegacyModels/EloquentModelSynth.php
```diff
@@ -210,7 +210,9 @@ protected function loadModel($meta): ?Model
                 $query->with($meta['relations']);
             }
 
-            $model = $query->first();
+            if ($query->first() !== null) {
```
joshhanley, line 213: `$query->first();` will already return `null` if nothing is found.

src/Features/SupportLegacyModels/Tests/PublicPropertyHydrationAndDehydrationUnitTest.php
```diff
@@ -876,6 +876,20 @@ public function test_it_ignores_the_key_if_the_model_does_not_exist()
...
     }
+
+    public function test_it_returns_null_if_the_model_was_deleted()
+    {
+        $this->expectNotToPerformAssertions();
+
+        $model = Author::create(['id' => 1, 'title' => 'foo', 'name' => 'bar', 'email' => 'baz']);
+
+        $component = Livewire::test(ModelsComponent::class, ['model' => $model])
+            ->call('$refresh');
```
joshhanley, line 887: Please change this to instead have an `->assertSuccessful()` after the call and remove the `$this->expectNotToPerformAssertions();` 

```suggestion
            ->call('$refresh')
            ->assertSuccessful();
```

joshhanley (changes requested):
@cbz1991 thanks for the PR!

I think it's probably better to add the check later instead of inside the loadModel method.

What I think you need to change is on line 74 add a check there like:

```php
if (is_null($model)) {
    return null;
}
```

Because returning an empty model would be unexpected when you were expecting an actual model. I think it would be better to return `null` instead if no model is found.

**Edit:** FYI - there was a big discussion on how to handle this error in PR #8047 but we never came to a conclusion. I was actually leaning towards we should throw a model not found error (instead of returning `null` like I have suggested above). On further thinking, I think throwing an error would be better than returning null.

src/Features/SupportLegacyModels/EloquentModelSynth.php
cbz1991: Ok, thank you for the suggestion!

src/Features/SupportLegacyModels/Tests/PublicPropertyHydrationAndDehydrationUnitTest.php
cbz1991: There may be use-cases when someone will prefer to handle the missing model different, like showing a custom message or re-creating the model.

cbz1991:
@joshhanley  There may be use-cases when someone will prefer to handle the missing model different, like showing a custom message or re-creating the model.

src/Features/SupportLegacyModels/Tests/PublicPropertyHydrationAndDehydrationUnitTest.php
cbz1991: Done!

luanfreitasdev:
Waiting for this 👍

cbz1991:
Hi @joshhanley! Any update regarding this issue?

lrljoe:
> Hi @joshhanley! Any update regarding this issue?

I'd suggest that there should be a check if it's nullable before returning null, to avoid any possibility of breaking existing implementations.

cbz1991:
> > Hi @joshhanley! Any update regarding this issue?
> 
> I'd suggest that there should be a check if it's nullable before returning null, to avoid any possibility of breaking existing implementations.
> 
> Otherwise, it should throw the standard model not found exception, which can be caught and handled by the developer

@lrljoe I am fine with any solution as long as it solves the problem.
@PhiloNL any update?

joshhanley:
@cbz1991 no updates at this stage. We're currently processing the backlog of PRs so should hopefully get to it soon

lrljoe:
Unless I'm missing something (on mobile), the test doesn't seem to validate behaviour post model delete.  

I may be wrong, and may have missed it. but I think it'd be best to assert for a behaviour, after deleting model and calling refresh, rather than just relying on no exception being thrown for the test to pass

christineRN:
@joshhanley Just checking in to see what the status is of this issue. I'm currently dealing with this and am hoping you've come to a consensus about how to fix. Any updates?

ghbob:
Is the current release related?

https://github.com/livewire/livewire/compare/v3.6.3...v3.6.4

cbz1991:
Hi @joshhanley, @PhiloNL ! Any update regarding this issue?

calebporzio:
Closing this because of it's age and our decision to move away from the legacy model support. Re-open if there is a better solution and renewed energy for the fix. thanks!

PR #8870, opened 2024-09-20
Fix scroll restoration

calebporzio:
Adjust scroll restoration so that it is the absolute LAST step in a page restoration (after Alpine initializes...)


anhhh11:
degrade style not update?

dist/livewire.esm.js
```diff
@@ -6816,13 +6820,17 @@ var require_module_cjs8 = __commonJS({
...
+            continue;
           if (!to.hasAttribute(name)) {
             from2.removeAttribute(name);
           }
         }
         for (let i = toAttributes.length - 1; i >= 0; i--) {
           let name = toAttributes[i].name;
           let value = toAttributes[i].value;
+          if (name === "style")
+            continue;
```
dauducmanh1812, line 6833: image
Degrade with inline styles

manavo: @calebporzio I think I'm finding the same, that we now can't modify inline styles anymore, based on this.

JayBizzle:
As others have said, this breaks updating `style` attributes :/

Totally broke a lot of our Livewire components

manavo:
I think newer versions have fixed this, but haven't confirmed 100%. But I think I saw a commit this week or last week that should fix it

manavo:
I think this commit, removing that check for "style": https://github.com/livewire/livewire/commit/774092003edb2670615ef09f3a9fbdd335d6d0d7

JayBizzle:
> I think this commit, removing that check for "style": https://github.com/livewire/livewire/commit/774092003edb2670615ef09f3a9fbdd335d6d0d7

Thanks 👍🏻

PR #8876, opened 2024-09-21
Pagination - Fixed three dot dark style

mattreesjenkins:
Removed duplicate tailwind dark model classes on the three-dot element and added the missing text colour


lrljoe:
Good catch!

PR #8883, opened 2024-09-23
Fix - v3.5.8 and v3.5.7 were causing some chart widgets in Filament to break

leandrocfe:
We noticed that Livewire versions v3.5.8 and v3.5.7 were causing some chart widgets in Filament to break.

After investigating, I couldn’t find any specific reason in the code for this behavior. However, after rebuilding the JS, the issue appears to be resolved.

This PR includes the rebuilt JavaScript, and the chart widgets are now functioning correctly.

> Before

image

> After

image


RazneshPrajapati:
Yes, I am also encountering same issue. In few pages the inline CSS gets added automatically in the output which is not in the code. Reverting to version 3.5.6 solves it all.

zepfietje:
@leandrocfe, this PR can be closed now since the assets have been rebuilt in the latest release.

leandrocfe:
> @leandrocfe, this PR can be closed now since the assets have been rebuilt in the latest release.

ahh yes, Thanks Zep!

PR #8886, opened 2024-09-25
enable autofocus after navigation

yilanboy:
There is an `autofocusElementsWithTheAutofocusAttribute` function to make input be focused after navigation.

```js
// js/plugins/navigate/index.js

function autofocusElementsWithTheAutofocusAttribute() {
    document.querySelector('[autofocus]') && document.querySelector('[autofocus]').focus()
}
```

But this feature is disable by default.

```js
// js/plugins/navigate/index.js

let autofocus = false
```

This PR change the `autofocus` value to `true` to enable this feature.


lrljoe:
How does this behave in conjunction with the advertised wire:scroll?

I suspect this is false by default due to that interaction

YilanBoy:
Thanks for your information, I will test in this situation.

PR #8898, opened 2024-09-29
display:contents for persist wrapper

mokhosh:
The `persist` directive adds a wrapper `div` around the code you want to persist, and this could create different issues specially with styling, since you have a wrapper which you can't control.

For example if you want to persist a `sticky` audio player, you won't be able to put the `sticky` style inside your component and you will have to wrap the whole `@persist()` directive inside a `sticky` wrapper because `sticky` is only sticky inside its parent. So, if you put the style inside a component it will do nothing since the parent is only wrapping your player.

What `display: contents` does is it makes css ignore your parent in terms of styling even though you still have the same js functionality. This way you can style the contents as if the parent doesn't exist.

Another example is if you have `flex` or something else on the actual parent (not the parent injected by persist). The `flex` will be broken by the wrapper, and won't work for the contents of the persisted code. With `display: contents` it does.

Hopefully this is a small change with big benefits.

### DIFF
--- a/src/Features/SupportNavigate/SupportNavigate.php
+++ b/src/Features/SupportNavigate/SupportNavigate.php
@@ -11,7 +11,7 @@ class SupportNavigate extends ComponentHook
     static function provide()
     {
         Blade::directive('persist', function ($expression) {
-            return '<?php app("livewire")->forceAssetInjection(); ?><div x-persist="<?php echo e('.$expression.'); ?>">';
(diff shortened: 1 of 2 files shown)


lrljoe:
Note that some CSP prohibit inline styles so this may break some apps!

mokhosh:
@lrljoe Nice. I moved the styles from the tag to `@livewireStyles`.

calebporzio:
Mmmm this makes sense however, this might mess with people's existing apps. Therefore, I recommend skipping `@persist` and using the `x-persist` attribute directly on an element. (`@persist` is just for vanity - it's a thin wrapper around the attribute)

PR #8905, opened 2024-10-02
allow to define a different middleware for file upload controller, th…

marianperca:
Allow to define a different middleware for FileUploadController the same as you can do for FilePreviewController.  
In a multi-tenant setup I don't have the `web` middleware and at the moment I'm getting an error when I upload a file.
We should be able to overwrite `web` middleware to something else.

No. I checked the other PRs and how this can be fixed,

Yes

No

No

This is so file uploads work if we don't have `web` middleware defined, usually in a multi tenant setup.
By having the `$middleware` defined statically on the controller, we would be able to overwrite it in a service provider: 
```
FileUploadController::$middleware = ['tenant', ...];
```

Thanks!

### DIFF
--- a/src/Features/SupportFileUploads/FileUploadController.php
+++ b/src/Features/SupportFileUploads/FileUploadController.php
@@ -8,10 +8,12 @@ use Illuminate\Support\Facades\Validator;
 
 class FileUploadController implements HasMiddleware
 {
+    public static array $middleware = ['web'];
+
     public static function middleware()
     {
         return array_map(fn ($middleware) => new Middleware($middleware), array_merge(
-            ['web'],
+            [static::$middleware],
             (array) FileUploadConfiguration::middleware(),
         ));
     }


PhiloNL:
Thanks for the PR @marianperca. 

1. Could you add some tests as well?
2. Using a static variable does not quite follow the code style Livewire has. I would suggest adding this to the config instead, for example:

```php
    'temporary_file_upload' => [
        // ...
        'preview_mimes' => [   // Supported file types for temporary pre-signed file URLs...
            'png', 'gif', 'bmp', 'svg', 'wav', 'mp4',
            'mov', 'avi', 'wmv', 'mp3', 'm4a',
            'jpg', 'jpeg', 'mpga', 'webp', 'wma',
        ],
        'preview_middleware' => null, // The middleware that is applied to the preview url | Default 'web'        
    ],
```

Thanks!

calebporzio:
Closing. Please re-submit if you want using some of the guidance provided by @PhiloNL

PR #8910, opened 2024-10-03
Fix php deprecation warning in BaseUrl class

gdebrauwer:
When the BaseUrl-class json-decodes `null` then that causes a deprecation warning:
```
json_decode(): Passing null to parameter #1 ($json) of type string is deprecated in /var/web/vd27736/releases/20240919140700/vendor/livewire/livewire/src/Features/SupportQueryString/BaseUrl.php on line 65
```

This PR fixes that by using an empty string if `$initialValue` is `null`. Json-decoding an empty string still results in `null`, so it does not change the behavior.


calebporzio:
Great, thank you @gdebrauwer

PR #8912, opened 2024-10-04
Handle null in dataGet helper to prevent TypeError

AlexandreBonaventure:
Hello, 
I have found myself using `$wire.$get('someObject.nested.property')` and when setting `someObject` to `null` I had a javascript error because the helper only handle `undefined`.
image

Also I could have worked around this issue if there was a $set API counterpart for deleting nested properties such as `$wire.$delete` (will set to undefined) but since it is not available, it would help me here.

I know I am not providing any repro nor tests but hopefully this is a very simple and not controversial change that does not require spending much time on it 🤞 

Thanks

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -84,9 +84,7 @@ export function dataGet(object, key) {
     if (key === '') return object
 
     return key.split('.').reduce((carry, i) => {
-        if (carry === undefined) return undefined
-
-        return carry[i]
+        return carry?.[i]
     }, object)
 }
 


js/utils.js
devajmeireles, line 1: Shouldn't you build assets since you changed that?

AlexandreBonaventure: @devajmeireles is it required ? I don't see any mentions of this in the contributing guide
I understand it is a frontend change and will eventually need an assets build but IMO this should be a release concern not a contribution concern. Ideally these things should be run in some kind of CI/release pipeline (more reliable, avoid "polluting" git diffs, etc) no ? 
Anyways, I am not going to mansplain you, let me know if this is a hard requirement and I'll commit the build

calebporzio:
This makes sense, thanks @AlexandreBonaventure

PR #8915, opened 2024-10-07
Update docs/properties.md

reppair:
Fixes the `Illuminate\Database\Model` class namespace reference at the Supported Property Types section.


reppair:
Oops! Will make another one :)

PR #8922, opened 2024-10-09
Update docs/properties.md 

reppair:
Fixes a class namespace reference at the Supported Property Types section.


PhiloNL:
Thanks!

PR #8930, opened 2024-10-11
Only merge the web middleware for file uploads when it is absent

ramonrietdijk:

The project that I'm working on requires some middleware to run before the `web` group. Currently, that is not possible as the `web` group will always be prepended.

I have not created a discussion for this matter.

Yes

No

Yes, I've included two test cases:
- The `web` group will be prepended in the middleware.
- The `web` group will only be added if it is absent.

Currently, the `web` group middleware will always be prepended in the `FileUploadController`. It's not possible to change the order. Middleware that has to be run before the `web` group can't be configured.

I've changed this logic so it will only prepend the `web` group if it's absent. This way, the `web` middleware can be added in the list at the desired position.

```php
'temporary_file_upload' => [
    'middleware' => [
        'custom',
        'web',
    ],
],
```

Without this change, the `web` group would be run twice with the configuration above.

### DIFF
--- a/src/Features/SupportFileUploads/FileUploadController.php
+++ b/src/Features/SupportFileUploads/FileUploadController.php
@@ -10,10 +10,13 @@ class FileUploadController implements HasMiddleware
 {
     public static function middleware()
     {
-        return array_map(fn ($middleware) => new Middleware($middleware), array_merge(
-            ['web'],
-            (array) FileUploadConfiguration::middleware(),
-        ));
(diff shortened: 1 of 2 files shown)


ramonrietdijk:
The test seems to fail prior to my commit

ramonrietdijk:
Hi @chillbram, great point! I've created a new pull request #8931 in order to make it more configurable. As I've chosen to use a new configuration key for this, it should not be a breaking change.

PR #8934, opened 2024-10-13
Introduction of $root for wire:model usage, taking $parent for a long spin

dimitri-koenig:

Yes, that's needed :-)
It helps avoid using Alpine with many workarounds and quirks, and use Livewire core features directly.

Yes

Nope, just this one

For sure

------

I want to introduce $root usage in templates within wire:model attributes which works similar to $parent usage but accesses the farthest root component it can find, which is usually the page component or the very first component within a normal blade template.

Example:
```html
<input type="checkbox" wire:model="$root.selectedIds" value="{{ $categoryId }}" />
```

That way you can use deeply nested components and still manipulate the root data, for which I have two use cases where this applies.

First use case: tree like recursive components which are often used in category like tree structures where one category contains multiple sub categories, which in turn contain multiple sub categories themselves. Current hacky workaround is using alpinejs and avoiding wire:model since its bound to the current or parent component.

Some pseudo code:
```html
<page-component>
  <first-level-component>
    <second-level-component>
      <third-level-component>
        <input type="checkbox" wire:model="$root.selectedIds" value="{{ $id }}" />
        
        Checkbox changes the $selectedIds array on <page-component />
      </third-level-component>
    </second-level-component>
  </first-level-component>
</page-component>
```

The other use case: A form component with a central values array, containing nested sub components which handle fine grained details. Same here: it's bound to the current or parent component. Everything deeper then that and you end up using workarounds and a lot of events.

// Update: tests are running fine locally. Trying to figure out how to make them run on CI. On it...


dimitri-koenig:
OK, some tests are failing due to issues in main branch, not due to these changes.

calebporzio:
Interesting. I will keep this in mind for a future version but I don't want to regret this precedent as sometimes $root can be interpretted as just the root element or scope of the current component you know/ thanks anyways

PR #8935, opened 2024-10-14
fix: function name typo

yilanboy:
There is a function name typo in `js/plugins/navigate/index.js`.

Change `Libaries` to `Libraries`.

```js
function fireEventForOtherLibrariesToHookInto(name, detail) {
    // ...
}
```


calebporzio:
Haha, that's pretty funny (just picturing myself saying "libaries" like a toddler). Good catch, thanks.

PR #8940, opened 2024-10-16
Add support for multiple action parameters to `wire:target`

blazery:
I was reading through the documentation and found a interesting warning about the wire:loading property.

> [!warning] Multiple action parameters aren't supported
> At this time, Livewire only supports targeting a loading indicator by a single method/parameter pair. For example `wire:target="remove(1)"` is supported, however `wire:target="remove(1), add(1)"` is not.

My first reaction was, "that can't be hard to fix", and so I dove into livewire for the second time. Load of fun solving this, and adding a possible new feature to the framework.

There was a [pending discussion](https://github.com/livewire/livewire/discussions/4441) for this, tho no response from other developers. 
Other bug reports and discussions that do include response from developers:
* [8134](https://github.com/livewire/livewire/discussions/8134)
* [8159](https://github.com/livewire/livewire/pull/8159#issuecomment-2022986030)

This now allows developers to use the wire:target attribute as follows
```php
<div wire:loading wire:target="processFunction('1', 2), processFunction('3', 4)">
    Processing...
</div>
```

Automated tests are included.

I don't know if I was supposed to push a new Js build into source control, but since it wasn't in the .gitignore I guessesd it can't hurt. Ill remove its if its not wanted.


calebporzio:
Love it thanks!

calebporzio:
@joshhanley  - merge conflicts then merge plz

PR #8953, opened 2024-10-21
Add default middleware to `FileUploadController` so it can be overridden

marianperca:
This PR allows to override the default `web` middleware for applications were this doesn't exist, specifically for multi-tenant setups.

Yes, customisation of the middleware is wanted.

Yes

No

Yes

For multi-tenant apps, the `web` middleware might not be present, at least in the implementation I'm using it's not. This poses an issue when uploading files, as the `web` middleware is applied automatically and can't be overwritten using the config.
In my previous PR #8905 you suggested to add a new config, but I think the majority of users won't need this, and adding a new default in config is not needed. The same approach was used for https://github.com/livewire/livewire/blob/3c8d1f9d7d9098aaea663093ae168f2d5d2ae73d/src/Features/SupportFileUploads/FilePreviewController.php#L11 that was introduced in #7911.
Since using a static variable, overriding the default middleware can be made in a service provider:
```
{
    public function boot(): void
    {
        FileUploadController::$default_middleware = 'tenant';
        ...
    }
```
If you think it's beneficial to have a new config value for this, I'll update the code.
Thank you!


lrljoe:
Is this not a duplicate of #8930?

marianperca:
@lrljoe there's a test and now the default middleware is a string, not an array. Also the code is updated to work with #8931.

lrljoe:
Looking at it, I can't help but think it's going to break something

lrljoe:
This would definitely need documentation writing for it.

src/Features/SupportFileUploads/FileUploadController.php
```diff
@@ -8,12 +8,14 @@
 
 class FileUploadController implements HasMiddleware
 {
+    public static string $default_middleware = 'web';
```
joshhanley, line 11: I think this should be an array to match `FilePreviewController`.

```suggestion
    public static array $defaultMiddleware = ['web'];
```

src/Features/SupportFileUploads/FileUploadController.php
```diff
@@ -8,12 +8,14 @@
...
+
     public static function middleware()
     {
         $middleware = (array) FileUploadConfiguration::middleware();
 
-        if (! in_array('web', $middleware)) {
-            $middleware = array_merge(['web'], $middleware);
+        if (! in_array(static::$default_middleware, $middleware)) {
+            $middleware = array_merge([static::$default_middleware], $middleware);
```
joshhanley, line 18:
```suggestion
        if (! in_array(static::$defaultMiddleware, $middleware)) {
            $middleware = array_merge(static::$defaultMiddleware, $middleware);
```

joshhanley (approved):
@marianperca thanks for the PR! Makes sense to me. Couple of small changes which I'll commit to make it match `FilePreviewController`. Only difference is it's called `$defaultMiddleware` instead of `$middleware` because there is already a `$middleware` being used in the middleware method and it looks funny both being `$middleware`.

https://github.com/livewire/livewire/blob/6cbb718e62267fad3e7ac4b8a770ea48e16c7cdf/src/Features/SupportFileUploads/FilePreviewController.php#L9-L16

calebporzio:
sounds good, thanks!

PR #8955, opened 2024-10-21
Add support for DI on computed properties

ryangjchandler:

Yes – discussion here: https://github.com/livewire/livewire/discussions/6967

Yes.

No,

Yes.

Currently it's not possible to resolve parameters from the container on computed property methods. This is inconsistent with actions which _do_ support DI parameters.

This pull request adds backwards compatible support for this behaviour by-way of a new `inject` argument on the `Computed` attribute.

This will change how computed properties are invoked and call the method through Laravel's container instead.

Here's what you need to do at the minute:

```php
#[Computed]
function foo() {
	$dep = app(MyDep::class);

	return $dep->something();
}
```

And with this change:

```php
#[Computed(inject: true)]
function foo(MyDep $dep) {
	return $dep->something();
}
```

### DIFF
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -8,6 +8,7 @@ use function Livewire\off;
 
 use Livewire\Features\SupportAttributes\Attribute;
 use Illuminate\Support\Facades\Cache;
+use Laravel\SerializableClosure\Support\ReflectionClosure;
 
 #[\Attribute]
 class BaseComputed extends Attribute
@@ -20,6 +21,7 @@ class BaseComputed extends Attribute
         public $cache = false,
         public $key = null,
         public $tags = null,
+        public $inject = false,
     ) {}
 
     function boot()
@@ -135,6 +137,10 @@ class BaseComputed extends Attribute
 
     protected function evaluateComputed()
     {
+        if ($this->inject) {
+            return app()->call([$this->component, parent::getName()]);
+        }
+
         return invade($this->component)->{parent::getName()}();
     }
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @ryangjchandler - wondering if we can add this functionality without it being a breaking change.

If we have to have an attribute called "inject", at that point it's already verbose enough to just use something like `app()` right?

Anyhow, because nothing can currently be passed into a `#[Computed]` method, I'm thinking we're safe. Let me know what you think.

ryangjchandler:
Yeah, the `inject` parameter on the attribute was more of a "let's be super safe and certain that there won't be any breaking changes" rather than a _necessity_ to avoid breaking changes.

I think we can remove it and everything will still work as expected – like you say, nothing can be passed through to a computed property anyway.

nuernbergerA:
why not somethink like this

```php
#[Computed]
function foo(#[Resolve] MyDep $dep) {
	return $dep->something();
}
```

ryangjchandler:
@nuernbergerA Doing this would required manually handling the reflection of parameters, etc.

ryangjchandler:
@calebporzio I've updated code to always run computed property methods through the container using Livewire's own `ImplicitlyBoundMethod` class.

I'm not super clued up on the differences with Laravel's `BoundMethod` class, so can change if need be.

calebporzio:
Thanks @ryangjchandler - I like this. Quick thought: I think your implementation won't work when computed properties are  protected methods because the removal of `invade()`.

ryangjchandler:
@calebporzio Good spot, I'll go back and add a test for protected methods too 👍

PR #8960, opened 2024-10-23
ref: Added Return Type

dipesh79:
Added return types to the methods in the LivewireServiceProvider class for better type safety and improved code readability. This update ensures that methods explicitly declare the expected return values, aligning with modern PHP standards.


calebporzio:
Thanks for being willing to contribute, however, I don't want to add this for only one file and I also don't want to add it for all files right now. Maybe we will add return types to the next version of Livewire. thanks

PR #8963, opened 2024-10-25
Allow rewriting of temporary upload/download urls for S3

austincarpenter:
1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first? ❌
2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed) ✅
3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out. ❌

No, due to the fact that the current implementation is not tested as per:
https://github.com/livewire/livewire/blob/82e2b40361f1779f541c18ee30d2d0d201ae768b/src/Features/SupportFileUploads/SupportFileUploads.php#L14-L20

Am somewhat unsure how best to test this so any help is appreciated.

Laravel's `AwsS3V3Adapter` allows for replacing of the base URL when temporary URLs are generated. 

This can be used to make the resulting url prettier or (in my case) avoid content blockers having a gripe at the URL containing 's3' or 'aws'.

Since https://github.com/laravel/framework/pull/45753 in L9, when using `Storage::temporaryUrl()` or `Storage::temporaryUploadUrl()` Laravel will look for the `temporary_url` key for an s3 disk in `config/filesystems.php`, i.e.:

```php
return [

    'disks' => [
        
        // ...

        's3' => [
            'driver' => 's3',
            // ...
            'temporary_url' => 'https://storage.foo.com'
        ],
    
    ],

];
```

and replace the scheme, host and port with what it can extract from the provided URL.

I've updated Livewire's implementation of generated upload URLs as per Laravel's.

See comment below for an example.

References:

https://github.com/laravel/framework/blob/5527e72efe5d86cf83afe21944a47fe9a0bdc406/src/Illuminate/Filesystem/AwsS3V3Adapter.php#L106-L130

https://github.com/laravel/framework/blob/5527e72efe5d86cf83afe21944a47fe9a0bdc406/src/Illuminate/Filesystem/FilesystemAdapter.php#L816-L824


austincarpenter:
By configuring something like CloudFront and pointing it to s3, any domain can be used.

A trivial example with a temporary download URL (query parameters omitted for brevity):

https://some-bucket.s3.us-east-1.amazonaws.com/file.txt
↓
https://abc123.cloudfront.net/file.txt
↓
https://storage.foo.com/file.txt

This is the approach that Vapor uses with its Custom Asset Domains. https://docs.vapor.build/projects/deployments.html#custom-asset-domains

austincarpenter:
One test is failing currently as it is on main.

calebporzio:
thanks!

PR #8978, opened 2024-11-01
Mount not running when a lazy loaded component reacts to an event

robert-stanciu:
This seems to be a bug already discussed. https://github.com/livewire/livewire/discussions/8430

Yes.

No, however, the changes done in the SupportEvents mechanism are just suggestions.

Yes.

Currently, I am facing an issue when I have multiple lazy components listening for a specific event. When I dispatch that event, all my not-yet-loaded lazy components will execute the callback attached to that event. Also, the mount method is not even executed.

I expect each component to react to the event only if they have been lazy-loaded.


robert-stanciu:
Hi @calebporzio, do you think the solution I suggested could cause any unexpected issues? My app behavior is broken because of the mentioned issue and I plan to have a temporary forked version in production until you give feedback.
Thanks, I appreciate it!

Grudok:
I think i have the same problem or an similar one

ypho:
Is there any update on this? We experience the same behaviour this PR fixes.

robert-stanciu:
Hello, I have updated the PR with the additional changes I have made on my fork.

I have been running these changes for a few months now and It seems to work pretty well.
There is however one thing I haven't managed to accomplish. The mount() method is not getting called. How does it work on my project? Well, I just don't have any logic in the mount method :).

I have tried to call the mount method (the commented out part in SupportsEvents.php) but it didn't work well.

calebporzio:
ahh intersting - i think if the component hasn't been initialized (because it's off the page for some reason) it shouldn't be responding to any events at all.

think about like a lazy component inside a modal - it should just be lying dormant until it enters the viewport and is fully loaded.

robert-stanciu:
Yes, that makes sense. By default should not listen at all, but it would be nice to have the option to control this behaviour.

e.g. you want to wake up the component with some user input which is available after the page loaded

joshhanley:
@robert-stanciu would you be willing to update this PR to match what @calebporzio has suggested? I think we should hold off on the control for now and see how this goes first.

joshhanley:
Closing this PR as it's been open for a long time and now has merge conflicts. If still relevant, feel free to re-submit with an updated branch and requested changes. Thank you!

PR #8985, opened 2024-11-04
Revert "added 'AllowDynamicProperties' to Component to silence log warning about dynamic property depreciation"

calebporzio:
Reverts livewire/livewire#8982

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -17,7 +17,6 @@ use Illuminate\Support\Traits\Macroable;
 use BadMethodCallException;
 use Livewire\Features\SupportFormObjects\HandlesFormObjects;
 
-#[\AllowDynamicProperties]
 abstract class Component
 {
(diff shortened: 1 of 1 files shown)


tristanisham:
Hey @calebporzio,

I saw you reverted my commit? I'm just curious as to why? Did I miss something in my original push? I'd like to learn so I can better contribute in the future.

joshhanley:
@tristanisham thanks again for your PR! We decided to revert it because applying the `#[\AllowDynamicProperties]` to the base Livewire component class means that all Livewire apps will be impacted by this. A lot of people may not want to allow dynamic properties as they are now deprecated in PHP.

It would be better if we could find what dynamic properties Livewire is using internally and actually create property definitions for them, that way it would stop this warning without needing the attribute override. We'd welcome a PR to fix this if you have the time 🙂

PR #9014, opened 2024-11-15
Fix optional route parameters

moisish:
At the moment, using optional parameters with route bindings returns a 404 as the query that runs will look like this

```
select * from `users` where `id` is null limit 1
```

Laravel ignore nullable parameters as you can find [here](https://github.com/laravel/framework/blob/835cb93346e63a33bfb3c73180a3a0e3d0e4e60c/src/Illuminate/Routing/ImplicitRouteBinding.php#L31) so I have applied the same on Livewire


joshhanley:
@moisish thanks for the PR! Can you please add some tests for this? Thanks!

lrljoe:
Would suggest you match the rest of the file, and return null, not $value (when value is null)

See the lines above your PR where it checks if $className is null.

moisish:
@joshhanley I have added a test but I am not sure if it's the right way. Can you please check

joshhanley:
@moisish looks ok to me (with a quick look). A good thing to make sure is that the test fails if you revert your changes. If that's the case, then the test should be all good 🙂

joshhanley (approved):
I've pulled this down and tested it. All looks good to me. I've actually run into this issue before, so I think it's a great addition!

schonhoff:
Would love to see this merged before the next release. I'm running into this frequently

calebporzio:
perfect thanks!

PR #9059, opened 2024-12-02
Fix query string hitting history state too many times

luanfreitasdev:

#7746

(cleanup of #8987 pr)

Fix Too many calls to location or history APIs in a short period.

Safari test (https://github.com/livewire/livewire/discussions/7746#discussioncomment-10389123): 

https://github.com/user-attachments/assets/ea87400a-7b2a-456d-8b46-0721891d06eb

With alias:

https://github.com/user-attachments/assets/f49d39a7-95e1-43e4-91ae-1a131d08fbb3


andrewdwallo:
Whats the holdup on this?

CalumBThomson:
> Whats the holdup on this?

Yeah, would be really great to see this fix pushed out asap . . . I have been running it locally on my code since @luanfreitasdev kindly raised this PR and it has resolved ALL the issues I was encountering within my Filament-based app.

mc-fu:
so still not merged, why?

CalumBThomson:
Does anyone here know what the criteria are for getting PRs accepted and merged in?

I have loved working with TALL stack over the past couple years, but without this fix I'm now having to reconsider my entire development approach - simply cannot wait forever to be able to deliver for my clients.

I'm more than happy to provide assistance / testing / evidence if someone can highlight what's holding this up. Otherwise, it's goodbye Livewire from me I'm afraid...

joshhanley:
@luanfreitasdev thanks for the PR! I'm still unable to replicate this issue, even in Firefox. Can someone please provide me a Volt component with instructions that demonstrates this issue so I can replicate it and so we can verify this PR? Thanks!

luanfreitasdev:
Hello @joshhanley, you can follow these steps:
* You must have 100 inputs with wire:model
* Open the console in Firefox
* Type some characters in any of the inputs

```
php artisan make:volt History
```

_History Component_
```blade
<?php
use function Livewire\Volt\{state};

state('search')->url();
?>

<div>
    <input wire:model.live.debounce.700ms="search" placeholder="Search" />
</div>

```

_welcome.blade.php_
```blade
<div>
    @foreach (range(1, 100) as $item)
        <livewire:history />
    @endforeach
</div>
```

benshawuk:
If you have a Filament app in SPA mode, you can just navigate through different pages a few times using Safari, and you will hit the problem very quickly.

lukas-frey:
Hi, I have the same issue. It's not a firefox issue only - safari on iOS and macOS has the same problem. And it is not only related to filament in SPA. I use filament actions on the frontend (no SPA or anything) and the error appears also.

andrewdwallo:
Yes it's not only SPA mode (wire:navigate), it's only more noticeable with Spa mode because it's easier to navigate between pages more quickly.

luanfreitasdev:
This doesn't happen only with filament or because filament uses too many resources simultaneously. Accepting this PR can fix the most significant bug in filament and improve the garbage thrown at the browser

joshhanley:
Thanks everyone for your input. I've managed to replicate this issue and get to the bottom of it.

I believe the solution provided here by @luanfreitasdev will help to reduce the number of requests hitting the browser history API. But there is still a chance that Livewire crashes due to the history state throttling error.

I've submitted a PR #9278 to add error handling to the history state calls.

I believe between this PR and my PR, the history state calls should be more robust.

joshhanley:
I've marked this as a draft as there are failing tests. Compiled assets weren't committed previously, hence why tests were passing before.

luanfreitasdev:
I submitted a possible fix and tested with this same script, but it would be good to have a more in-depth test using filament

luanfreitasdev:
Merged with https://github.com/livewire/livewire/pull/9278

luanfreitasdev:
I'm closing this until I know if it needs it.

benshawuk:
> I submitted a possible fix and tested with this same script, but it would be good to have a more in-depth test using filament

How can I test this?  Is it on a separate branch or merged into main?
Happy to give it a whizz with Filament - as it's quite easy to reproduce..
Thanks.

luanfreitasdev:
In version 3.6.1, a fix was sent that prevents Livewire from crashing, but it makes 100 attempts to fix errors. My version makes a comparison and prevents this behavior, making it faster.

You can test by comparing the two versions: 3.6.1 or composer require livewire/livewire dev-fix-firefox-history-replace


```json
"minimum-stability": "dev",
```

```bash
composer require livewire/livewire dev-fix-firefox-history-replace
```

@joshhanley . I'm reopening the pull request because on line 144, it is being overwritten several times with the same value and is resolved in the PR

https://github.com/livewire/livewire/pull/9278/files#diff-7c1bd01af20f289c6329b536209bc1258170460a3b911d350d2afaa8a6440d97R144

luanfreitasdev:
Tests passing: https://github.com/luanfreitasdev/livewire/actions/runs/14433481875/job/40471102060 😅

andrewdwallo:
I agree that while #9278 stops Livewire from crashing the browser, it kind of seems like a band-aid solution. The console error is still thrown in the browser because the history api is still being called unnecessarily. This PR seems like a suitable fix for that.

CalumBThomson:
> I agree that while #9278 stops Livewire from crashing the browser, it kind of seems like a band-aid solution. The console error is still thrown in the browser because the history api is still being called unnecessarily. This PR seems like a suitable fix for that.

Agreed - this PR is a no-brainer must-have.
Without this PR my console log is still littered with error "SecurityError: Attempt to use history.replaceState() more than 100 times per 10 seconds" (in Safari). When I apply this PR, no such errors are logged because the history is not being needlessly replaced hundreds of times.

avinmaster:
Waiting for merge :D

binaryfire:
As others have mentioned, unfortunately https://github.com/livewire/livewire/pull/9278 isn’t a complete fix. I’m still seeing lots of console errors.

@joshhanley Is there any reason this PR can’t be merged?

matalaweb:
+1 for merge

luanfreitasdev:
I'm closing this due to potential [conflicts](https://github.com/livewire/livewire/pull/9366/files#top). Thanks everyone! 🫶🏻

joshhanley:
Whoops didn’t mean to reopen this. Thanks @luanfreitasdev for your effort on this. I’ve got another idea for a fix for this that I’m working on. Looking at adding a buffer between our history plugin and the browsers history API to reduce the number of hits. It seems we are hitting it for every prop set/ change instead of batching them up.

robinvda:
> Whoops didn’t mean to reopen this. Thanks @luanfreitasdev for your effort on this. I’ve got another idea for a fix for this that I’m working on. Looking at adding a buffer between our history plugin and the browsers history API to reduce the number of hits. It seems we are hitting it for every prop set/ change instead of batching them up.

I was wondering if there's an update on this? Besides the errors in the console log, the url doesn't update anymore after this error is thrown, causing unexpected results for example if the user copies the url or refreshes the page.

Merdin:
> > Whoops didn’t mean to reopen this. Thanks @luanfreitasdev for your effort on this. I’ve got another idea for a fix for this that I’m working on. Looking at adding a buffer between our history plugin and the browsers history API to reduce the number of hits. It seems we are hitting it for every prop set/ change instead of batching them up.
> 
> I was wondering if there's an update on this? Besides the errors in the console log, the url doesn't update anymore after this error is thrown, causing unexpected results for example if the user copies the url or refreshes the page.


I have the problem as well, what stopped working for me are the notifications after saving/creating a resource.

binaryfire:
@robinvda @Merdin I opened a discussion re: URL updates here: https://github.com/livewire/livewire/discussions/9450. 

I think @luanfreitasdev's approach with this PR was the right solution. It's a shame that it (or something similar) wasn't merged. I'm doing the same thing in my apps i.e. using `JSON.stringify` to compare the old state and new state to make sure the history is only updated if something has changed. It's fixed everything and no history updates are lost. The URL is always updated correctly and I no longer get any browser throttling warnings.

pxlrbt:
Hey @joshhanley,

can you provide an update on your alternative solution? Is there a reason against this PR if it resolves the issue? Filament v4 uses many `#[Url]` attributes, making this a common problem on every Filament install (and any Livewire app that uses `#[Url]` or `wire:navigate`)

joshhanley:
@pxlrbt hey sorry I didn't get back to you on this. We now have a fix in v4 which fixes this at a deeper level that batches and then flushes any history state changes for both query strings and navigate. See this PR https://github.com/livewire/livewire/pull/9661

pxlrbt:
@joshhanley Thanks for the answer and for looking into this. I guess that means v3 won't be fixed?

joshhanley:
@pxlrbt no worries! Not likely, but I can check with Caleb.

PR #9061, opened 2024-12-02
Update discussion templates

joshhanley:
The existing bug discussion template only requires steps to reproduce but that isn't clear that they should describe the problem and include code snippets.

So this PR updates the bug discussion template to require people to actually describe the problem they are having, provide code snippets, and describe how they expect it to work. I've also removed repository URL as an option, and instead just made it Wirebox only, to make it easier to review the problem without needing to pull down repositories. We can always ask for a repository later if we cannot reproduce or if it's a complicated issue.

I've also created a similar template for the help discussion template and fixed the links on the "new issue" page to open a new discussion with the correct category selected.

Screenshot 2024-12-03 at 10 23 09 AM
Screenshot 2024-12-03 at 10 23 29 AM
Screenshot 2024-12-03 at 10 23 41 AM


nuernbergerA:
Was about time 👍

joshhanley:
@nuernbergerA yeah they needed a refresh!

PR #9075, opened 2024-12-09
adds note to validation doc

SilasOtoko:
Possibly avoids some confusion on expected functionality for the live-validation feature when using `.blur`. I personally thought the feature wasn't working and learned I misunderstood the intended behavior.

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -321,6 +321,9 @@ class CreatePost extends Component
 
 Now, in the above example, even though `#[Validate]` is empty, it will tell Livewire to run the fields validation provided by `rules()` everytime the property is updated.
 
+> [!warning] `.blur` does not trigger an error for an input with the "required" rule until an attempted form submit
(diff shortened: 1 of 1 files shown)


lrljoe:
This PR is misleading.  "blur" fires upon "blur", it does **not** do what you say:

_"does not trigger an error for an input with the "required" rule until an attempted form submit"_

lrljoe:
Rather than submitting a docs update, I'd suggest that you submit a failing test, so that the underlying issue(s) can be addressed

joshhanley:
@SilasOtoko thanks for the PR! I don't exactly follow what the issue is. Can you please open a discussion and include a Volt or Livewire component that demonstrates the issue, so we can investigate? I'm going to close this for now as we'd prefer to find a better way to handle it and if that's not possible, then we will revisit adding a warning to the docs. Thanks!

PR #9076, opened 2024-12-10
Add failing test for emoji aliases for url params

j3j5:

No, it's a bug I encountered when trying to use an emoji as an alias for a URL parameter.

Yes, it's called `emojis-as-aliases` and it's the target of this PR.

No, it's a single (failing) test for the bug I encountered.

Yes.

When trying to use the `Url` attribute on a public property, using an emoji as the alias (`as:` param on the constructor), together with a default exception, it doesn't seem to work (the query string param doesn't go away when using the default state). I've only tested with a boolean triggered by a checkbox model. The same component works fine if there's no alias or the alias is a string so it's probably related to the url encoding of the emoji.


calebporzio:
Closing this PR as it's been open for over a year without a fix. This was a failing test to demonstrate a bug rather than a solution. If this issue is still relevant, please open a new issue describing the problem and we can track it there. Thanks for reporting!

PR #9084, opened 2024-12-12
Add #[Required] attribute

oli-laban:
1️⃣ _Is this something that is wanted/needed? Did you create a discussion about it first?_

It's something I'd personally find useful, and imagine others would too. No discussion though.

2️⃣ _Did you create a branch for your fix/feature? (Main branch PR's will be closed)_

Yes: `required-attribute`

3️⃣ _Does it contain multiple, unrelated changes? Please separate the PRs out._

No, just this feature.

4️⃣ _Does it include tests? (Required)_

Yes

5️⃣ _Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful._

### Motivation

Due to the way attributes on Livewire components are set on the component's properties (i.e. not through a constructor), properties without default values can be left uninitialized if the corresponding attributes are missing from the component tag.

This can cause any number of different errors depending on how the property is used. Commonly something like:

* `Attempt to read property "foo" on null`
* `Typed property Foo::$bar must not be accessed before initialization`

These errors generally aren't too difficult to diagnose, but if another developer wrote the component, and particularly if the property is used in a less straightforward way (e.g. within an Alpine component), it can become much less obvious what is causing the error.

Also, there is no distinction between properties that need to be passed to the component, and properties that are simply set within the component itself (via actions etc.). So, again if another developer wrote the component, it's not always clear what needs to be passed via attributes unless it is documented.

### Solution

This PR adds a simple `#[Required]` attribute that can be added to properties that should be passed to the component. If any required attributes are missing, an error will be thrown on mount:

```
Missing required properties [foo, bar] in component [my-component]
```

### DIFF
--- a/src/Attributes/Required.php
+++ b/src/Attributes/Required.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Attribute;
+use Livewire\Features\SupportRequiredProperties\BaseRequired;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+class Required extends BaseRequired
+{
+    //
+}
--- a/src/Features/SupportRequiredProperties/BaseRequired.php
+++ b/src/Features/SupportRequiredProperties/BaseRequired.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Livewire\Features\SupportRequiredProperties;
+
+use Attribute;
+use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+class BaseRequired extends LivewireAttribute
+{
+    //
+}
--- a/src/Features/SupportRequiredProperties/RequiredPropertyNotProvidedException.php
+++ b/src/Features/SupportRequiredProperties/RequiredPropertyNotProvidedException.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Livewire\Features\SupportRequiredProperties;
+
+use Illuminate\Support\Str;
+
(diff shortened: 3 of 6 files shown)


francoism90:
I don't fully understand this PR. Hopefully you can give an example. :)

```php
class ChildComponent extends Components
{
	public Post $post; // throws error when not passed in parent
}
```
Do you mean without a typehint?

oli-laban:
> I don't fully understand this PR. Hopefully you can give an example. :)
> 
> ```
> class ChildComponent extends Components
> {
> 	public Post $post; // throws error when not passed in parent
> }
> ```
> 
> Do you mean without a typehint?

Both with and without. A typed property as in your example doesn't throw an error, it's simply left undefined.

lrljoe:
What's wrong with using Validate here?  Seems a duplication.

oli-laban:
@lrljoe 
> What's wrong with using Validate here? Seems a duplication.

Different use cases. This simply ensures that any required attributes are passed to the component on mount, whereas validation isn't run on mount.

You *could* get `Validate` to do the same thing by calling `$this->validate()` inside `mount()` but that adds weirdness if you want to validate any form data as well, and I wouldn't say it's a great solution regardless (you'd have to manually catch validation exceptions etc.).

> I think you'd need to add in capability for "sometimes" etc if implementing, as that'll be an almost immediate ask

Not really sure how that's applicable here. If the attribute isn't present (in the same vein as a field not being present with regular validation), then that's the whole point of `Required` and an exception will be thrown. There's no further validation here because as long as your component properties are typed, then PHP will handle that.

PR #9094, opened 2024-12-17
Add parent modifier to wire:loading.

slovenianGooner:
This functionality is related to a discussion opened and available here - luckily I found a solution for it by myself.

The feature is available on the `wire-loading-parent` branch.

It contains one change on one file in one commit on the `wire-loading.js` directive.

The tests are not included.

Please refer to the discussion for description of the need for the feature.
https://github.com/livewire/livewire/discussions/9093

What the condition does is to check if the parent component is "loading", everything else stays the same.
It only works for the immediate parent.

### DIFF
--- a/js/directives/wire-loading.js
+++ b/js/directives/wire-loading.js
@@ -8,6 +8,10 @@ directive('loading', ({ el, directive, component, cleanup }) => {
 
     let [delay, abortDelay] = applyDelay(directive)
 
+    if (component.modifiers.includes('parent')) {
+        component = component.parent;
+    }
+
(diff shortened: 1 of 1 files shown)


slovenianGooner:
FWIW this could also be solved if this is added. https://github.com/livewire/livewire/pull/8146

lrljoe:
I'm going to say it before someone else does, please can you add in relevant tests, otherwise it is unlikely to get merged in, as it creates a gap in the test suite, which if someone breaks it in future. won't get picked up.

joshhanley:
@slovenianGooner thanks for the PR! We do require tests though, so I'm going to close this. Feel free to resubmit with tests. Thanks!

PR #9096, opened 2024-12-17
Add missing semicolon

RDelorier:
Looks like I was doing too many things at once in https://github.com/livewire/livewire/pull/9080 and forgot the semicolon.


RDelorier:
Already fixed

PR #9097, opened 2024-12-18
Fix: Add cleanup to `$wire.$watch` and `$wire.$hook`

danie-ramdhani:
### TLDR

so, `$wire.$watch` is require to be cleanup after the element is removed (not just when the livewire component removed only). but not like `$wire.$hook` that return `unhook`, `$wire.$watch` is return nothing.

### Scenario

```html
<div x-data="{
    init() {
        this.$wire.$watch('foo', () => {
            console.log('watch model')
        })

        this.$wire.$hook('commit', ({ component, commit, respond, succeed, fail }) => {
            console.log('commit')
        })
    }
}" wire:replace.self></div>

<input type="text" wire:model="foo">
```

$wire.$hook:

- hit refresh button
- check console
- log is multiplied exponentially

$wire.$watch:

- hit refresh button
- check console
- typing inside input element
- log is multiplied exponentially

and worst scenario:

```html
<div x-data="{
    init() {
        this.$wire.$watch('foo', () => {
            this.$wire.$set('foo', 'bar', false)
        })

        this.$wire.$hook('commit', ({ component, commit, respond, succeed, fail }) => {
            this.$wire.$set('foo', 'bar', false)
        })
    }
}" wire:replace.self></div>

<input type="text" wire:model="foo">
```

this will make error `Uncaught (in promise) Could not find Livewire component in DOM tree` because the element is removed / not connected from DOM and can't find the closest component if i use things like this:

### Possible Fix

1. expose cleanup element (this PR)

create variable then assign it inside `Alpine.magic('wire', (el, { cleanup })` so we can use it inside `$wire.$hook` and `$wire.$watch`

2. add `return unwatch` to `$wire.$watch`

very simple fix for this is just return `unwatch` from `wireProperty('$watch', ...)` so we can unwatch it manually for every single time we use it:

```html
<button type="button" wire:click="$refresh">
    Refresh
</button>

<div x-data="{
    init() {
        const unwatch = this.$wire.$watch('foo', () => {
            console.log('watch model')
        });

        const unhook = this.$wire.$hook('commit', ({ component, commit, respond, succeed, fail }) => {
            console.log('commit')
        });

        Alpine.onElRemoved(this.$el, () => {
            unwatch()
            unhook()
        })
    }
}" wire:replace.self></div>

<input type="text" wire:model="foo">
```

but i don't think this is good idea to clean up it manually for every single time we use it.

### Test

Test included.


Willy-Bueno:
I am facing the same problem, I think it would be better if the element was disassembled automatically or the clock was cleaned

MaxencePaulin:
Hey, if $wire.$watch return unwatch, we can easily unwatch on Alpine component destroy() (with Alpine.data).

So, can i just ask pr for just return unwatch in $wire.$watch ?

calebporzio:
Closing this PR as it's been open for a long time and now has merge conflicts. If still relevant, please re-submit with an updated branch. Thank you!

PR #9099, opened 2024-12-18
Fix key component attribute in docs

angelej:
Fix the `:key` component attribute in the `components.md` documentation.


joshhanley:
@angelej thanks!

angelej:
Thank you for your time!

PR #9101, opened 2024-12-21
Additional test for morphing with `array_unshift`

PerryvanderMeer:
The failing test in this PR represents a use-case, resulting in the following error: `Uncaught Snapshot missing on Livewire component with id: xxx`

The test only fails when there is another `div` next to the child element.
Some other observations while debugging:
- It will also pass when you add a unique ID to the `div` (like the time).
- It will also pass when you remove `x-show="false"` from the div.
- It doesn't matter if the `div` comes before or after the child element.

(Follow-up on #8369)


nuernbergerA:
i can confirm this, have the same problem

PerryvanderMeer:
Looking good right now. Will leave this PR here as additional test.

joshhanley:
@PerryvanderMeer thanks for the PR! I'm going to close this as I've got a test for this in the snapshot fixes PR.

PR #9117, opened 2024-12-30
Fix `resetExcept()` to not reset all when all properties supplied

liamduckett:
Fixes bug discussed in https://github.com/livewire/livewire/discussions/8096.

Reproducing test has been added. Fix is very similar to one described in the discussion, although I chose a slightly 'stricter typing' approach.

This fix had to be applied in two places - you may notice `Form` declares most (although not all) functions from `InteractsWithProperties`. Once this PR is merged, it'd be good to see if `Form` can make use of `InteractsWithProperties` to prevent some of this duplication.


calebporzio:
Thanks!

PR #9119, opened 2025-01-02
Feat: Add `shouldPushToHistoryState` parameter to `Livewire.navigate`/`Alpine.navigate`

danie-ramdhani:
I think this feature can be useful for following scenario:

- change language without reloading page and push history state.
- tools like [vite livewire plugin](https://github.com/defstudio/vite-livewire-plugin) or [wire spy](https://github.com/wire-elements/wire-spy) can use this feature to refresh the layout and entire livewire components instead only refresh the entire livewire components.

### Test

test included.


js/plugins/navigate/index.js
```diff
@@ -16,7 +16,7 @@ let autofocus = false
 
 export default function (Alpine) {
 
-    Alpine.navigate = (url) => {
+    Alpine.navigate = (url, shouldPushToHistoryState = true) => {
```
joshhanley, line 19: I wonder if it would make more sense to make this parameter `replace = false`.

joshhanley (commented):
@danie-ramdhani thanks for the PR! See my comment re the parameter name.

I think this would also need to be added on the backend, so `$this->redirect('my-page', navigate: true, replace: true);` would be possible.

And also support some how for when `wire:navigate` is on a `<a>` tag. Maybe `<a href="my-page" wire:navigate.replace>`?

js/plugins/navigate/index.js
danie-ramdhani: idk if `replace` is good parameter. maybe `history = true` but in [this line](https://github.com/livewire/livewire/blob/e78ae3c4392e7a58e3b343634e0e428a52c506dc/js/plugins/navigate/index.js#L23) there's history parameter too and how's the `wire:navigate` modifier looks like if it's without history push state?

PR #9121, opened 2025-01-03
Update test stub to use test_ prefix

dwightwatson:
This simply updates the test stub used by the `livewire:make` command. It replaces the docblock syntax `/** @test */` with the `test_` method name prefix.

* Laravel's test stubs use this prefix so this makes it more consistent, and
* [Livewire itself made the move towards this prefix](https://github.com/livewire/livewire/pull/8365).

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/livewire.test.stub
+++ b/src/Features/SupportConsoleCommands/Commands/livewire.test.stub
@@ -10,8 +10,7 @@ use Tests\TestCase;
 
 class [testclass] extends TestCase
 {
-    /** @test */
-    public function renders_successfully()
+    public function test_renders_successfully()
     {
         Livewire::test([class]::class)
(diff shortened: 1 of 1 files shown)


joshhanley (approved):
@dwightwatson thanks for the PR!

Using the docblocks is deprecated in PHPUnit 11 and will be removed in PHPUnit 12, so I think this is a good change to make.

Personally I don't like having the `test_` prefix and would prefer the `#[Test]` attribute, but I'm also all for keeping everything consistent with Laravel.

Here is Laravel's test stub for reference https://github.com/laravel/framework/blob/11.x/src/Illuminate/Foundation/Console/stubs/test.stub

joshhanley:
@dwightwatson thanks for being willing to contribute! I just merged the same fix as part of PR #9213 so I'm going to close this. Thanks again!

PR #9122, opened 2025-01-05
Docs: Fix use named arguments for file uploads example

mike-healy:
The code in the docs right now give an error because it reverts from named to unnamed arguments.


PhiloNL:
Thanks!

PR #9123, opened 2025-01-05
Fix rendering styles and script multiple times

stayallive:
I am using Livewire, Flux UI and Wire Elements in my project and even though I think I did everything right and configured the right options (`inject_assets` as `false`) and made sure I had a single `@livewireStyles` and `@fluxStyles` I got rendered the inline CSS twice.

Looking at the code I noticed there is a flag to indicate the styles were already rendered, same for the scripts but only the scripts flag was used... and to top it of that script flag also applied to the CSS. This puzzled me so I tried to come up with a solution to use the flags as I think they were intended and stops rendering the styles multiple times for me, while everything else still works as expected. Maybe I missed something but this seems like a good start.

In addition, there is a check to see if there is any styles/scripts to be rendered before doing so after the request was handled but this almost always has a stray newline character resulting in the string not being empty. Added a `trim` to fix that.

### DIFF
--- a/src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
+++ b/src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
@@ -48,7 +48,7 @@ class SupportAutoInjectedAssets extends ComponentHook
                 $assetsBody .= FrontendAssets::scripts()."\n";
             }
 
-            if ($assetsHead === '' && $assetsBody === '') return;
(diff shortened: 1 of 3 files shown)


stayallive:
Still appreciate a look here, still maintaining a custom fork with this code 😄

calebporzio:
@claude - evaluate this PR for me please

calebporzio:
Thanks!

PR #9130, opened 2025-01-07
Add `data-current` attribute to active `wire:current` links

### LINKED ISSUE #918: Navigation Item Current Attribute dropped on dispatched event.
### Flux version

1.1

### Livewire version

3.5.17

### What is the problem?

When an event is dispatched from a component to another component, the :current Attribute on a navigation item seems to drop off. Attached is a demo video of this happening.

https://github.com/user-attachments/assets/9a06636f-1030-4121-a8e3-5b5569564429

### Code snippets

Attached are the files involved with this issue:

Navigation Badge Livewire Component

```php
<?php

namespace App\Livewire\Invites;

use App\Models\Invite;
use Livewire\Attributes\On;
use Livewire\Component;

class NavigationBadge extends Component
{
    public function render()
    {
        return view('livewire.invites.navigation-badge');
    }

    #[Computed]
    #[On('invitation-updated')]
    public function invitations()
    {
        return Invite::getPending();
    }
}
```

Navigation Component Template

```html
<div>
    <flux:navbar.item
        href="{{ route('invites') }}"
        :badge="$this->invitations"
        badge-color="sky"
        icon="envelope" 
        :current="request()->is('invites')"
    >
        Invitations
    </flux:navbar.item>
</div>
```

Model getPending query function.

```php
public static function getPending(): int
  {
      return self::query()->where('user_id', auth()->id())->where('status', self::PENDING)->count();
  }
```

Navigation Layout Header.

```html
<flux:header container>
    <flux:sidebar.toggle class="lg:hidden" icon="bars-2" inset="left" />

    <flux:navbar class="-mb-px max-lg:hidden">
        <flux:navbar.item icon="home" href="{{ route('welcome') }}" :current="request()->is('/')">
            Home
        </flux:navbar.item>

        @auth
            <flux:navbar.item icon="adjustments-horizontal" href="{{ route('dashboard') }}" :current="request()->is('dashboard')">
                Dashboard
            </flux:navbar.item>

            <livewire:invites.navigation-badge />
        @endauth

    </flux:navbar>

    <flux:spacer />

    <flux:button x-on:click="$flux.dark = ! $flux.dark" icon="moon" variant="subtle" aria-label="Toggle dark mode" />

    @auth
    
        <flux:dropdown position="top" align="start">
            <flux:profile avatar="{{ auth()->user()->avatar }}" />

            <flux:menu>
                <flux:menu.item icon="arrow-right-start-on-rectangle" wire:click="logout">
                    Logout
                </flux:menu.item>
            </flux:menu>
        </flux:dropdown>
    @else
        <flux:button size="sm" class="mx-1" href="{{ route('twitch.login') }}" icon-trailing="arrow-up-right">Login</flux:button>
    @endauth
</flux:header>
```

Invite approved function that dispatches event:
```php
public function update($inviteId, $status)
  {
      $this->form->update($inviteId, $status);

      $this->dispatch('invitation-updated');
  }
```

I expect it to when the event is dispatched, the count of the badge in the navigation item to be updated without the theming underline disappearing. The current route did not change. 

joshhanley:
This PR has been updated to add the `data-current` attribute automatically to any links that have `wire:current` that are active.

That way we can make use of this feature in Flux where styles are added based on the `data-current` attribute.

---
**Old PR description**
Currently if you use `wire:current` it adds whatever the value is to the elements classes list if there is a match.

But in Flux, we use an attribute `data-current` to denote whether an element is the current one or not and the styles are applied based on that attribute.

So this PR updates `wire:current` to add support for adding an attribute instead of classes.

```blade
<nav>
    <a href="/dashboard" ... wire:current.attr="data-current">Dashboard</a>
    <a href="/posts" ... wire:current.attr="data-current">Posts</a>
    <a href="/users" ... wire:current.attr="data-current">Users</a>
</nav>
```

This will allow us to use `wire:current` with Flux for more accurate active link detection. See livewire/flux#918 for where the idea came from.

I've not yet added documentation for this but will if happy with the concept.

Fixes livewire/flux#918


oli-laban:
I think it would be nice if this allowed for attribute values too, to cover all use cases. For example (outside of Flux), adding `aria-current="page"`.

It could perhaps use `=` as a separator since that isn't allowed in attribute names e.g.

```html
<a href="/dashboard" ... wire:current.attr="aria-current=page">Dashboard</a>
```

calebporzio:
Hmmm let's do this: for now, let's add `data-current` to the element automatically if wire:current exists on it. I'm hesitant to support .attr on this because it already diverges from other directives that support .attr and .class and it might confuse some.

joshhanley:
@calebporzio yep makes sense. I've updated the PR to automatically add the `data-current` attribute instead.

calebporzio:
Thanks Josh!

PR #9131, opened 2025-01-07
Docs: correct component name in redirect method

sajjadhossainshohag:
The `redirect()` method uses the correct Livewire component name, changing `ShowPage::class` to `ShowPosts::class` for consistency with the defined route.


lrljoe:
Looks good to me!

PhiloNL:
Thanks!

PR #9134, opened 2025-01-08
Check for ModelNotFoundException in request payload loop to avoid certain situations

matthewhaworth:
This will definitely not be accepted - but I wanted to raise it to flag a problem that exists and get your support in solving it.

Currently, if the following apply, there is a bug:

1. A livewire/update request bundles two component payloads into one request
2. The request path has a implicit binding on a model
3. The first request deletes the model

The issue is that the second loop in the handleUpdate method of `src/Mechanisms/HandleRequests/HandleRequests.php` tries to "subtituteImplicitBindings" for a model that no longer exists, because it was deleted in the first iteration, so it throws a 404.

It is my view that `livewire/update` should never throw a 404 as it effectively behaves like an RPC. The '404-ness' should be dealt with in the response so that the frontend can handle it more gracefully. 

I have outlined the specific scenario where I got this issue here: https://github.com/livewire/livewire/discussions/9126 

I know that this PR is not going to be the way to fix it, it's far too broad. However, I would prefer your take on this so I can amend it appropriately and learn along the way.

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Mechanisms\HandleRequests;
 
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Facades\Route;
 use Livewire\Features\SupportScriptsAndAssets\SupportScriptsAndAssets;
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Mmmmm this is interesting - you're right that I'm not comfortable merging this PR, but I definitely hear what you're saying -this seems like a bug to me.

PR #9135, opened 2025-01-09
Add support for using `@assets` outside of a Livewire component

### LINKED ISSUE #802: Editor cannot be used outside of a Livewire component

joshhanley:
# The scenario

Currently if a Blade component makes use of the `@assets` directive and that component is used outside of a Livewire component, then a `Using $this when not in object context` error is thrown.

image

# The problem

The problem is that the assets directive makes a call to the store and passes `$this` (the Livewire component) in as a reference.

```php
\Livewire\store(\$this)->push('assets', \$__output, \$__assetKey);
```

But when `@assets` is used outside of a Livewire component then `$this` doesn't exist, hence the error is being thrown.

# The solution

To fix this, I have added support to the assets directive for keeping track of assets that have been called outside of a Livewire component, and injecting them into the layout alongside any Livewire component assets.

I did review whether we should do the same with the `@scripts` tag, but those scripts are evaluated on the front end by Livewire. Because of that, it didn't make much sense to change it at the moment as it would require a re-think on how all of those scripts are processed.

Fixes livewire/flux#802

### DIFF
--- a/src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
+++ b/src/Features/SupportAutoInjectedAssets/SupportAutoInjectedAssets.php
@@ -29,6 +29,10 @@ class SupportAutoInjectedAssets extends ComponentHook
             $assetsHead = '';
             $assetsBody = '';
 
+            // If `@assets` has been used outside of a Livewire component then we need
+            // to process those assets to be injected alongside the other assets...
+            SupportScriptsAndAssets::processNonLivewireAssets();
+
             $assets = array_values(SupportScriptsAndAssets::getAssets());
 
             // If there are additional head assets, inject those...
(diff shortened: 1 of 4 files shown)


lrljoe:
Looks perfectly sane to me!  Nothing seems to break/behave unexpectedly.

calebporzio:
Thanks @joshhanley !!

PR #9140, opened 2025-01-13
Update enum query parameter to set `null` for invalid values when nullable

abduselam1:
The default implementation when the URL query parameter type is enum it uses `$enum::from()` method this is a good catch when the attribute is not nullable, but when the attribute is nullable it should use `tryFrom()` instead of `from()` method.

So when the page loads with different enum value on a query paramtere it checks if the attribute is nullable or not, if its nullable it uses `tryFrom()` and if not `from()` method

Here is the problem I was getting
image


calebporzio:
I like it - thanks

abduselam1:
> I like it - thanks

you are welcome. 🙏

PR #9143, opened 2025-01-16
Add support for non-backed enums in `EnumSynth`

liamduckett:
1️⃣ Is this something that is wanted/needed? 
Yes (although only by me so far!)

Did you create a discussion about it first?
No, but on reflection it would have been useful...

Yes

My goal here is to support Pure Enums (I've referred to them as Unbacked in this PR - happy to change this if desired) as valid Livewire properties. My usecase is for an Onboarding page. I have the following Enum:

```php
enum OnboardingStage: string
{
    case StageFoo = 'stage-foo';
    case StageBar = 'stage-bar';
    case StageBaz = 'stage-baz';
    case StageQux = 'stage-qux';
}
```

...and it is used as so:

```
@if($stage === OnboardingStage::StageFoo)
    <div>
        Basic Explanation of StageFoo, maybe with a livewire component or two
    </div>
@elseif($stage === OnboardingStage::StageFoo)
    <div>
        Different text and different components
    </div>
@endif

<button wire:click="nextStage">
    Next Stage
</button>
```

This enum is only backed (currently) for the purpose of working with Livewire: its string value has no real meaning. I fully appreciate this is a tiny inconvinience to developers (and is easily workable!).

I've achieved this by modifying the existing `EnumSynth`, although with hindsight maybe a separate synth would be cleaner? A lot of the added code is typechecking, which I'm not a huge fan of. Happy to take a second swing with a new Synth if desired (and if anyone has any name ideas they'd be appreciated).

### DIFF
--- a/src/Mechanisms/HandleComponents/Synthesizers/EnumSynth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/EnumSynth.php
@@ -6,22 +6,32 @@ class EnumSynth extends Synth {
     public static $key = 'enm';
 
     static function match($target) {
-        return is_object($target) && is_subclass_of($target, 'BackedEnum');
+        return is_object($target) && is_subclass_of($target, 'UnitEnum');
     }
 
     static function matchByType($type) {
-        return is_subclass_of($type, 'BackedEnum');
+        return is_subclass_of($type, 'UnitEnum');
     }
 
     static function hydrateFromType($type, $value) {
         if ($value === '') return null;
 
+        $backed = is_subclass_of($type, 'BackedEnum');
+
+        if(! $backed) {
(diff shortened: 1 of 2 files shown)


calebporzio:
I understand where you're coming from, however, there's all sorts of questions like: is ok to expose the actual enum values on the front-end? What IS the value? the fully qualified namespace and value? or just the value? What happens if someone binds it to a text field or uses "#[Url]" with it? 

This is why backed values are great: you can define human-readable, serializable, front-end friendly values to use elsewhere and keep the enum values reserved for PHP use.

Therefore going to stick with backed values for now. thanks!

liamduckett:
Yeah, looking back on this - I'm not even sure I'm in favour anymore 😅

PR #9148, opened 2025-01-21
Docs: Correct succeed callback argument from "effect" to "effects"

wpoortman:
Fix documentation: The upgrading docs stated that the succeed callback receives an "effect" argument, which returns undefined. Updated it to correctly reference "effects".


lrljoe:
Suggest you prefix the PR with "Docs: " or similar, just so it highlights that it's just a docs update (which may mean a faster resolution)

PhiloNL:
Thanks!

PR #9149, opened 2025-01-21
Add navigate support for replacing html attributes

### LINKED ISSUE #839: Scroll lock styles not being cleaned up on Navigate
### Flux version

v1.0.29

### Livewire version

3.5.17

### What is the problem?

If you use `flux:menu` within `flux:dropdown`, it'll add `overflow:hidden` to the root HTML element.
This style attribute persists if you then `wire:navigate` away from the page, which is probably not desirable behavior.

You may say "Well obviously! It says to utilize `flux:navmenu` on the [dropdown documentation page](https://fluxui.dev/components/dropdown#navigation-menus) for this exactly reason". Which is true, however, is pretty easy to miss if you are copying your navbar directly from the [header documentation page](https://fluxui.dev/layouts/header), which uses this exact type of menu (it doesn't include `wire:navigate` - but also doesn't include any warning about the `overflow:hidden` behavior)

### Code snippets

```blade
<flux:dropdown position="top" align="start">
    <flux:profile name="Sheldon" avatar="https://ui-avatars.com/api/?name=Sheldon&background=random" />
    <flux:menu>
        <flux:menu.item icon="user" :href="route('profile')" wire:navigate>{{ __('Profile') }}</flux:menu.item>
        <flux:menu.item icon="arrow-right-start-on-rectangle" wire:click="logout">{{ __('Log Out') }}</flux:menu.item>
    </flux:menu>
</flux:dropdown>
```

Results in `overflow:hidden` persisting on root html element when navigating to Profile

I'd either expect `overflow:hidden` to be removed from the element when navigating to a different page, OR for this possibly unexpected behavior to be more prominent within the documentation so that I could appropriately choose `flux:navmenu` as the sub-item and not experience this behavior.

joshhanley:
# The scenario

Currently if you use `wire:navigate` and change to a different page, the new pages `<html>` element attributes are not added and the old ones aren't cleaned up.

This is noticeable with Flux, when you have a `wire:navigate` link inside a modal. When you open a modal, styles are applied to the `<html>` element to stop the page in the background from scrolling. But when you navigate to the new page, the anti-scroll styles are still applied, stopping the page from being scrollable.

Recording 2025-01-21 at 19 40 53

```blade
<?php

use Livewire\Volt\Component;

new class extends Component
{
    //
};
?>

<div>
    <div class="h-screen">Some long content to demonstrate the scroll not working</div>

    <flux:modal.trigger name="example">
        <flux:button>Launch modal</flux:button>
    </flux:modal.trigger>

    <flux:modal name="example">
        <flux:button :href="request()->url()" wire:navigate>
            Go to me
        </flux:button>
    </flux:modal>
    <div class="h-screen">Some morelong content to demonstrate the scroll not working</div>
</div>
```

# The problem

The issue is that `wire:navigate` doesn't have any facility for replacing attributes on the `<html>` element.

# The solution

This PR adds support for replacing the page's `<html>` element attributes with the new pages `<html>` attributes.

I've structured this so it does a comparison between the old and new attributes, and updates the existing ones if there has been a change and removes any that are no longer needed. By doing it this way, we don't trigger any changes for attributes that are the same on both pages.

Fixes livewire/flux#839


miagg:
@joshhanley this broke dark mode for me 😬
On first visit "dark" class is applied on html tag based on user's schema or localStorage value but on wire:navigate it get's lost.

danharrin:
Yeah this also broke dark mode in Filament

AlexandreBonaventure:
it broke our custom dark mode handling as well. 

---
(unsolicited) sidenote: 
Flux could handle scroll lock with :has selector like 

```css
body:has(dialog[open]) {
   overflow: hidden;
}
```

joshhanley:
Hey guys, I sent this code to @danharrin  but though it's best to share it here too.

It's always a good practice to ensure that anything that gets applied dynamically, such as dark mode styles, is reapplied using the `livewire:navigated` event. https://livewire.laravel.com/docs/navigate#javascript-hooks

So for dark mode, if you currently have a script in your head tag that reads local storage/ etc to apply a dark class, you would need to ensure that it gets run on page load, but also gets run on the `livewire:navigated` event.

Below is simplified example of how I would apply dark mode for both page load and on any navigate requests.

```blade
<head>
    <script>
        function setDarkMode() {
            document.documentElement.classList.add('dark'); // Replace this with your dark mode detection logic.
        }

        setDarkMode(); // Make sure to call this manually as it stops initial page load FOUC

            setDarkMode(); // This re-applies dark mode styles if the page is navigated
        });
    </script>
</head>
```

I did some research into what Turbo/ Turbolinks does and it handles the `<html>` element attributes in the same as the change in this PR and they recommend listening to their `turbo:load` event to re-run/apply dynamic scripts/styles.

Hopefully this should help fix the issue, but if it's not working, please tag me and share some code and I will have a look to see if there is a better way to fix it!

miagg:
@joshhanley this was exactly what I did, but it doesn't look like an optimal/clean solution and it doesn't work properly since the event handler runs after the DOM has been loaded and it flashes for a fraction of a second on each navigation. 🤢

[see here](https://b2.apptime.gr/file/apptime/flashing.mov)

I had to rollback to 3.5.18 until a better solution is found.

joshhanley:
@miagg bugger 😕 I’m going to look further into this tomorrow. Can you share your layout file/ dark mode code so I can see if I can replicate it?

miagg:
I'm using the default 'class' strategy of tailwindcss  for dark mode. Then I have a triple state toggle (system, light, dark)

```
<html>
<head>
    <script>
        if (
            localStorage.theme === 'dark' ||
            (!('theme' in localStorage) && window.matchMedia('(prefers-color-scheme: dark)').matches)
        ) {
            document.documentElement.classList.add('dark');
        } else {
            document.documentElement.classList.remove('dark');
        }
    </script>
</head>
<body>
   ...
</body>
</html>
```

Then I tried
```
<html>
<head>
<script>
        function setDarkMode() {
            if (
                localStorage.theme === 'dark' ||
                (!('theme' in localStorage) && window.matchMedia('(prefers-color-scheme: dark)').matches)
            ) {
                document.documentElement.classList.add('dark');
            } else {
                document.documentElement.classList.remove('dark');
            }
        }
    
        setDarkMode();
            setDarkMode();
        });
    </script>
</head>
<body>
   ...
</body>
</html>
```


Since filament uses a similar approach, I guess it should be also testable with filament admin package.

danharrin:
I had to add another call to `setDarkMode()` in body to remove the flicker, but I don't think this should be required in an ideal world

edo2313:
@joshhanley
I think we should be able to exclude some attributes (class, data-xxxxx, etc..) from being removed/replaced as is the case with the dark mode class.

nicolagianelli:
> @miagg bugger 😕 I’m going to look further into this tomorrow. Can you share your layout file/ dark mode code so I can see if I can replicate it?

Any news about this? If you want to replicate it by force you could use a timeout to simulate a slow page load on navigated to see the page flickering from light to dark mode.
@edo2313 proposal could be good, a way to configure excluded attributes from the config file maybe?
So you could add a "data-theme" attribute to the exclusions and use it to toggle dark mode, or add the "class" attribute if no class are added dynamically...

`
<head>
    <script>
        function setDarkMode() {
            document.documentElement.classList.add('dark'); // Replace this with your dark mode detection logic.
        }

        setDarkMode(); // Make sure to call this manually as it stops initial page load FOUC

            setTimeout(setDarkMode, 1000); // Use a timeout to simulate a slow page load on navigated
        });
    </script>
</head>
`

PR #9150, opened 2025-01-21
Fixed asset loading in dynamically loaded components.

desarrolloalborada:
1️⃣ **Is this something that is wanted/needed? Did you create a discussion about it first?**

I haven't seen any prior discussion about this bug, likely because it occurs in a very specific scenario: when assets are included in a dynamically loaded component, and the assets include a non-script tag (e.g., a <link> tag) followed by additional assets. The issue may go unnoticed if the link tag is placed at the end of the asset list and the error does not manifest. Additionally, this issue can be easily worked around by preloading the assets in the initial page load, so anyone encountering the problem may have opted for a simpler solution instead of reporting it.

**EDIT**
The issue is discussed here: #8769 

2️⃣ **Did you create a branch for your fix/feature? (Main branch PR's will be closed)**

**Yes**, the branch is named` fix-assets-loading-in-nested-components`.

3️⃣ **Does it contain multiple, unrelated changes? Please separate the PRs out.**

**No**, this pull request is focused solely on fixing the issue with dynamically loaded assets after non-script tags.

Yes, `test_script_after_non_script_asset_is_loaded_in_nested_components` in `src/Features/SupportScriptsAndAssets/BorwserTest.php`

5️⃣ **Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.**

This pull request resolves an issue where dynamically loaded assets using the @assets directive would skip a tag if a `<link>` tag (or any non script tag) appeared earlier in the sequence. This occurred due to the use of a live DOM collection (newHead.children) during iteration. When elements were appended to the <head>, they were removed from the live collection, causing subsequent elements to be skipped in the loop.

For example:

```
<!--LivewireComponent-->
<div></div>

@assets
<link rel="stylesheet" type="text/css" href="/test.css">
<script src="/test.js"></script>
@endassets
```

If the above component were dynamically added to the page the `<link>` tag would be correctly appended to the <head>, but the `<script>` tag would be skipped due to newHead.children  shrinking during the iteration.

**How was it fixed?**
By converting the live newHead.children collection into a static array using Array.from. This ensures that  the iteration is not affected by DOM changes during the loop:

```
let headChildren = Array.from(newHead.children);
for (let child of headChildren) {
    await runAssetSynchronously(child);
}
```

instead of 

```
for (let child of newHead.children) {
    await runAssetSynchronously(child);
}
```

**Why is it useful?**

Usually, adding assets directly to the page where the assets are going to be used is often sufficient for most use cases, but there are scenarios where dynamically loading assets is important. 

For example, I frequently work with full page components that extend base components. Consider the following example:

```
AssistantSettingsForm Extends SettingsForm{
    public function render(){
           return parent::render()->layout('components.layout.admin-layout', [
                'title' => 'Assistant settings',
                'breadcrumb' => [...]
           ]
    }
}
```

I cannot simply add  a @myCustomAssets directive there, I would have to do it within the extended SettingsForm view, meaning that I am adding a resource to any components that extends SettingsForm, even if it is not needed.


kishannareshpal (approved):

jneto14:
will this be merged?

PR #9163, opened 2025-01-29
Fix error when there is a single parameter in `Livewire.dispatch()`

GC-Max:
This pull request addresses an issue that occurs when using Livewire.dispatch() inline without wrapping parameters in an array. Under current behaviour, passing parameters in non-array format leads to a page error. 

To replicate the error, you can run the newly added test while disabling Arr::wrap() in the SupportEvents class. This PR ensures that parameters passed to Livewire.dispatch() can be safely handled, regardless of whether they are provided as an array or as individual arguments, thereby preventing the page error.


joshhanley (approved):
@GC-Max thanks for the PR!

For anyone else looking, the error that is thrown is the below.

image

When I tested it, this PR fixes that issue and also allows `$this->dispatch('foo', 'bar');` to work. I've pushed an update to the test that includes that.

joshhanley:
So I've dug into this a little more and found that this fix doesn't fully fix the issue.

# The scenario

Currently if you call `$this->dispatch('test', 'new');` from a method in a Livewire component, then it get's picked up correctly by listeners and the `event.detail` is already an array `['new']`.

You can see below, we have 3 listeners (`#[On]` attribute, `x-on:`, and `@script` using `$wire.on()`) set up and they are all receiving `['new']` correctly.

image

But if we try to dispatch the event using `wire:click="$dispatch('test', 'other')"`, then we can see the two browser based listeners receive the data, but it's like `'other'` instead of `['other']` which is what is expected to match what happens above when dispatching from the backend. We can see the backend listener fails all together with an exception.

image

# The problem

The issue is that the front end `$dispatch()` is actually just using the params directly.

```js
export function dispatch(component, name, params) {
```

Where as the backend `$this->dispatch()` is getting the params as an array because of the `...`

```php
public function dispatch($event, ...$params)
```

# The solution

The solution proposed in this PR was to wrap the params in an array on the backend if they're not already an array.

The issue with that is that the results are inconsistent depending on whether the event if fired from the backend or the front end (top 2 below is backend and bottom 2 are front end).

image

So instead the fix is to add the check to the frontend, just before the frontend event is dispatched, check to see if the params are a string, and if they are wrap them in an array.

In theory we could update the frontend method signature to include `...params`

```js
export function dispatch(component, name, ...params) {
```

But when I tested this, I found it's actually a breaking change as it causes an extra layer of nesting.

Hence why the best solution is to only wrap params if it's a string. I've pushed these changes to the PR.

Now we can see no matter where the event is fired from, if there is only a single value in it, then it will be wrapped correctly.

image

calebporzio:
thanks a bunch

PR #9172, opened 2025-02-02
Fix/upload with long filenames

Naoray:
it;s another spin on #8690.

In my app we receive multiple reports of `Unable to retrieve the file_size for file at location: livewire-tmp` weekly. This PR aims to solve these issues,

yes

no

yes

This follows the same approach as #8690. We check if `$file->getClientOriginalName()` exceeds a **159-character** limit.

### Why is the max filename length **159**?
Most filesystems allow filenames up to **255 characters**, but we need space for:

- **30** → hash  
- **5** → `-meta` prefix  
- **1** → `-` separator  
- **5** → average file extension length  
- **159** → actual filename (base64 encoding increases length by ~33%)  
- **2** → buffer  

```
// Formula:
255 - 30 - 5 - 1 - 5 = 159  
Base64 increases by ~33% (159 * 4/3)  
+ 2 for safety buffer
```

### How does it work?
If the filename exceeds **159 characters**, we:
1. **Truncate** the base64-encoded filename.
2. **Store** the original filename in a separate meta file inside `FileUploadConfiguration::metaPath()`.

```php
public static function metaPath($filename = '')
{
    return static::path('meta' . DIRECTORY_SEPARATOR . $filename);
}
```

### Retrieving the original filename

When getClientOriginalName() is called:
- If the decoded base64 starts with [truncated], we fetch the full filename from the stored meta file.

### Handling in tests

When testing `getOriginalClientName()`, we remove the encoded `-hash`.., `-mimeType`, and `-size` from `$this->getFilename()` to correctly retrieve the original filename from the meta file.


bbredewold:
Very nice! I would like this to be merged @calebporzio ;)

benstjohn:
@calebporzio any idea when this might be merged in? Having some issues with Digital Ocean spaces backups due to this problem.

luisszzuniga:
@calebporzio please :)

jhjvandenbroek:
+1 for merge

alexsyvolap:
+1

bbredewold:
image

calebporzio:
had a talk with chat about this. what do we think of this option instead? this way we don't have all the extra complexity of storing a meta file? seems super easy.

CleanShot 2025-05-13 at 17 55 38@2x

calebporzio:
@joshhanley  - dig into this?

PerryvanderMeer:
Hmm in some cases we actually need to get the original file name via `getClientOriginalName()`. I think that might cause some problems with your solution?

Naoray:
> had a talk with chat about this. what do we think of this option instead? this way we don't have all the extra complexity of storing a meta file? seems super easy.
> 
> 
> 
> CleanShot 2025-05-13 at 17 55 38@2x
> 
> 

That removes the possibility of getting the original client file name, doesn't it?

joshhanley:
Thanks everyone for their input and @Naoray for the PR! I'm going to close this PR in favour of PR #9360 in which I've implemented a similar solution but one that always uses the meta file. This simplifies the implementation and also allows us to clean up how `Testable` stores meta information during testing.

I did experiment with option #2 as request by Caleb, but it didn't work out as easy as we'd hoped.

PR #9194, opened 2025-02-19
Remove unused `mount()` arguments from Livewire directive

AJenbo:
# The problem

The `mount()` method in `LivewireManager` is defined with 3 parameters:

https://github.com/livewire/livewire/blob/509f2258c51741f6d06deb65d4437654520694e6/src/LivewireManager.php#L71

However, the generated code calls this method with `5 arguments`, which causes PHPStan to report a mismatch. Since the method signature doesn't account for the extra arguments, static analysis sees it as an incorrect method call—even though those extra arguments are unused.

This shows up when analyzing Blade templates and can be confusing or noisy for users relying on stricter static analysis levels.

# The solution

This PR removes the extra, unused arguments passed to `mount()`, aligning the call with the method’s actual signature. This eliminates unnecessary static analysis warnings without changing any runtime behavior.

# The conversation

If the extra arguments are included intentionally—for example, to support future use cases or give extended components access to more context—an alternative would be to add the extra parameters to the `mount()` method signature instead. That would preserve compatibility and still satisfy static analysis tools.

I'm open to whichever approach makes the most sense long-term!


lrljoe:
This is for backwards compatibility I believe!

AJenbo:
In that case can we instead add the 2 unused arguments to the mount method?

AJenbo:
@calebporzio since you are the author of https://github.com/livewire/livewire/commit/3d478032f458ded68882c4e0dbe7938d9bacd05d could you look at this?
It's unclear if this is an unintended left over, or intentional.

calebporzio:
Thanks!

PR #9202, opened 2025-02-24
Add JavaScript actions using `$js()`

calebporzio:
This PR adds support for defining JavaScript actions, which are JavaScript functions that can be called from `wire:click` or `$wire`, to either run client side code only or to apply optimistic updates before triggering a server request.

How to define a JavaScript action:

```blade
<?php
use Livewire\Volt\Component;

new class extends Component {
    public function save()
    {
        // ...
    }
};
?>

<div>
    <button wire:click="doSomething">Do Something</button>
</div>

@script
<script>
    $js('doSomething', () => {
        console.log('doSomething');

        $wire.save()
    });
</script>
@endscript
```

Clicking on the `Do Something` button will call the `doSometing` JavaScript action defined in the script at the bottom. It will then call the `save()` method on the component when it is done.


joshhanley:
@calebporzio one thing to note, at the end I actually extracted the new JS actions code out of the main component class and am injecting it from the feature file instead. This required a new `component.register` trigger/listener, that runs before `$wire` is initialised.

If you don't like that idea, just revert the last two commits [159556a](https://github.com/livewire/livewire/pull/9202/commits/159556a4a6b3da9c1d287df4898941f337b99cae) and [80a7eff](https://github.com/livewire/livewire/pull/9202/commits/80a7eff1987b9f94b67f2173fcfae405bf5bcd98) as I have ensured those changes were applied separately.

sajjadhossainshohag:
Looks good

PR #9213, opened 2025-03-03
Converts `@test` to `test_` for PHPUNIT 12 support

realpoke:
**Problem**
When creating tests from the Livewire stubs, it adds the `@test` doc-comments to the method so PHPUNIT knows it's a test it should run. However, in PHPUNIT version 12 this is deprecated.

> Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.

**Solution**
This PR converts all  `@test` doc-comments to  `test_` methods, which is compatible with PEST and PHPUNIT. It also changes docs and mentions of the `@test` doc-comments.

**Conversation**
I converted to `test_` methods since this has been the standard in the past, but we could use PHP attributes instead. But since PHP attributes are newer, it might break backward compatibility. My opinion is that adding `test_` to methods is a great compromise!

### DIFF
--- a/docs/contribution-guide.md
+++ b/docs/contribution-guide.md
@@ -101,8 +101,7 @@ use Tests\TestCase;
 
 class UnitTest extends TestCase
 {
-    /** @test */
-    public function livewire_can_run_action(): void
+    public function test_livewire_can_run_action(): void
     {
        // ...
     }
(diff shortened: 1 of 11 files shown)


joshhanley:
Thanks!

PR #9218, opened 2025-03-05
Add `#[Persistent]` attribute for middleware

ralphjsmit:
Hello, thank you very much for all the great work you did and are doing on Livewire!

Quite often it happens that you add a new middleware to your routes, but don't immediately realize that it also needs to be marked as persistent. In my opinion, it is not super-convenient to create the middleware and _then_ look up a (big) service provider or something and _then_ add the middleware in `Livewire::addPersistentMiddleware()`.

This PR proposes a very easy solution to allow marking a middleware as `#[Persistent]` with a simple attribute, meaning that developers can just easily mark a middleware as persistent when creating them, without needing to open up a whole other file. It was actually pretty easy to do just inside the `filterMiddlewareByPersistentMiddleware()`, so I think this would be a great DX-improvement and would result in less middleware being "forgotten" to be marked as persistent. 

I also added tests for the method. Thanks!

### DIFF
--- a/src/Attributes/Persistent.php
+++ b/src/Attributes/Persistent.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Livewire\Mechanisms\PersistentMiddleware\BasePersistent;
+
+#[\Attribute]
+class Persistent extends BasePersistent
+{
+	//
+}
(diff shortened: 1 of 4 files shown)


maurocasas:
This actually is a pretty useful iteration!
Think of Multi-tenancy where you have a middleware which resolves tenant based upon subdomain.

calebporzio:
This is a decent idea, but going to pass on it for now. Let's keep an eye on demand for it in the future and maybe try again. thanks!

PR #9219, opened 2025-03-05
Add PHPDoc types to Testable

AJenbo:
# The problem

The `Testable` class is missing PHPDoc type annotations, which limits the ability of IDEs to provide contextual suggestions and autocomplete when writing tests. More critically, at higher analysis levels, tools like PHPStan will report issues when chaining methods, since the return types are undocumented. This forces developers to manually re-specify the object type in order to continue chaining, adding unnecessary friction.

# The solution

This PR adds PHPDoc return types to the `Testable` class, making it easier to chain methods in tests without triggering static analysis warnings. It also improves IDE support and makes tests more ergonomic to write.

# The conversation

These annotations don’t affect runtime behavior but significantly improve the developer experience—especially in projects using strict static analysis. It’s a small change with a high impact on usability.

### DIFF
--- a/src/Features/SupportTesting/Testable.php
+++ b/src/Features/SupportTesting/Testable.php
@@ -26,6 +26,15 @@ class Testable
         protected ComponentState $lastState,
     ) {}
 
+    /**
+     * @param string $name
+     * @param array $params
+     * @param array $fromQueryString
+     * @param array $cookies
+     * @param array $headers
+     *
+     * @return static
+     */
     static function create($name, $params = [], $fromQueryString = [], $cookies = [], $headers = [])
     {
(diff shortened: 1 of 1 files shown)


src/Features/SupportTesting/Testable.php
```diff
@@ -44,6 +53,11 @@ static function create($name, $params = [], $fromQueryString = [], $cookies = []
         return new static($requestBroker, $initialState);
     }
 
+    /**
+     * @param string $name
```
babeuloula, line 57: It can be an array, an object or a string.

AJenbo: Thanks, fixed

robingta:
- [ ] __bm~~Xbox~~

PR #9220, opened 2025-03-05
Change facade to reference class directly

AJenbo:
This makes it easier to navigate from  the facade to the underlying class since you can now click on it rather then have to guess at what the alias is for.


AJenbo:
It might for some, though I don't see why you would even use an alias here, what's the upside?

PR #8929, opened 2024-10-11
Fix bug in reset() when using nested dot notation or wildcard

hapidjus:
Review the contribution guide first at: https://livewire.laravel.com/docs/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?
Yes, you can find the discussion here https://github.com/livewire/livewire/discussions/8785 and related discussion here https://github.com/livewire/livewire/pull/8464
2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed)
Yes
3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
No
4️⃣ Does it include tests? (Required)
Yes
5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

After updating to Livewire v3.5 our code stopped working. I've traced it back to #8464 
After submitting a form we want to reset some input fields, where the recipients field are checkboxes.
```
 $this->reset([
    'title',
    'description',
    'recipients.*',
    'body',
);
```
It chokes on `'recipients.*',`  giving us  `method_exists(): Argument #1 ($object_or_class) must be of type object|string, array given` This would work before updating v3.5

It would also throw an error when using nested dot notation like
```
 $this->reset(['title.bar.baz']);
```

PR #9221, opened 2025-03-06
Fix `method_exists()` error when resetting arrays using nested syntax

danie-ramdhani:
follow-up on https://github.com/livewire/livewire/pull/8464. instead of checking the method from livewire component. we check the properties is form object or because `method_exists` will throw error on array:

image

i actually never use `$this->reset(...)` to reset array before. so, i don't know if the previous [PR](https://github.com/livewire/livewire/pull/8464) will throw error 😅.

Fixes #9217

@PerryvanderMeer @joshhanley


joshhanley:
@danie-ramdhani thanks for the PR! Can you explain how this PR is different to the fix, referenced in a comment on https://github.com/livewire/livewire/pull/8464#issuecomment-2407563769,  in PR #8929 that has already been merged? Thanks!

danie-ramdhani:
> @danie-ramdhani thanks for the PR! Can you explain how this PR is different to the fix, referenced in a comment on [#8464 (comment)](https://github.com/livewire/livewire/pull/8464#issuecomment-2407563769), in PR #8929 that has already been merged? Thanks!

in discussion https://github.com/livewire/livewire/discussions/9217. it throw error when reset array props in livewire component (not form object). simpler repro: https://github.com/livewire/livewire/discussions/9217#discussioncomment-12397428

joshhanley (approved):
Makes sense, thanks @danie-ramdhani!

calebporzio:
Thanks!

PR #9231, opened 2025-03-11
Add validation support to synths

joshhanley:
# The scenario

Currently if you have any custom objects that have their own Livewire synth, such as Flux's `DateRange` object and the corresponding `DateRangeSynth` class, those objects aren't able to be validated by Livewire.

For a full write up of this issue, please read this comment https://github.com/livewire/flux/issues/1222#issuecomment-2712542844

# The problem

The issue is that when Livewire prepares it's public properties ready for validation, it needs to convert those properties to strings or arrays of strings so Laravel's validator can correctly run validation against the data.

But when we are using a custom type, Livewire has no way to know how to convert those objects into arrays.

Livewire currently has this method which iterates through all public properties and attempts to convert them to a structure that Laravel can understand.

```php
protected function unwrapDataForValidation($data)
{
    return collect($data)->map(function ($value) {
        // @todo: this logic should be contained within "SupportWireables"...
        if ($value instanceof Wireable) return $value->toLivewire();
        else if ($value instanceof Arrayable) return $value->toArray();

        return $value;
    })->all();
}
```

But sometimes an object may not be a `Wireable` or `Arrayable`. Or maybe it is `Arrayable`, but what `toArray()` returns is different than the structure the data should be in for validation to work correctly. Again see here https://github.com/livewire/flux/issues/1222#issuecomment-2712542844 for more details on how Flux's DateRange object `toArray()` method doesn't return the same values that we want to validate.

The `@todo` for the `Wireable` also suggests that a better solution could be implemented that covers `Wireables` so there doesn't need to be feature specific code in the validation logic.

# The solution

One solution is to leverage the Synths to instruct Livewire on how to format the data ready for validation, which is implemented in this PR.

This PR adds support for adding a `unwrap()` method to Synths which returns the properties data in a format that Laravel's validation can support. It also update the validation logic to first try and match the object type with a synth and if found, then tries to call `unwrap()`. If a synth isn't found or the `unwrap()` method doesn't exist, then it falls back to the other options.

The `upwrapDataForValidation()` method now looks like this.

```php
protected function unwrapDataForValidation($data)
{
    return collect($data)->map(function ($value) {

        $synth = app('livewire')->findSynth($value, $this);
        
        if ($synth && method_exists($synth, 'unwrap')) return $synth->unwrap($value);
        else if ($value instanceof Arrayable) return $value->toArray();

        return $value;
    })->all();
}
```

I've also cleaned up and removed the `Wireable` logic from this method and moved it into the `unwrap()` method on the `WireableSynth`.

There have also previously been reports of a similar issue happening with `BackedEnum` not being able to be validated, so I've updated the `EnumSynth` to include the `unwrap()` method and added tests for it.

See livewire/flux#1222 for more details.


PerryvanderMeer:
@joshhanley For reference, this is a breaking change for us.

```php
public function rules() : array
{
    'level' => [
        'required',
        Rule::enum(UserLevel::class),
        static function (string $attribute, UserLevel $value, Closure $fail) : void
        {
            if ($value === UserLevel::MANAGER)
            {
                $fail('Whoops.');
            }
        },
    ],
}
```

The validation rule from above now throws the following exception, because the `unwrapForValidation` method now returns the enum value. 
`TypeError: App\Livewire\Forms\StoreOrUpdateUser::{closure:App\Livewire\Forms\StoreOrUpdateUser::rules():175}(): Argument #2 ($value) must be of type App\Enums\UserLevel, int given`

So, this might break the validation in some applications.

fotrino:
Hi @joshhanley like @PerryvanderMeer mentioned, after this update (from v3.6.1 to v3.6.2) it looses the casting of the typed enum.

Here are 2 examples in Wirebox.
[Enum in Rule v3.6.1](https://wirebox.app/b/4z6v8) -> The property returns an Enum on validation.
[Enum in Rule v3.6.2.](https://wirebox.app/b/47np0) -> The property returns a String.

This also breaks our application because we use custom rules to make different checks based on the enum type.

dalbo:
@joshhanley @calebporzio This breaking change in the treatment of Enums for validation has also broken things for us. Is there a plan to roll this back? Converting an Enum to it's backing value behind the user's back doesn't seem like a good idea, even if well intended.

joshhanley:
@PerryvanderMeer @fotrino @dalbo thank! Sorry I missed this! @dalbo has submitted a PR #9268 with a fix.

PR #9232, opened 2025-03-11
[Chore] Sync next page classes with previous page class

danie-ramdhani:
next page is white in dark mode

Screenshot 2025-03-11 134319


calebporzio:
Good catch, thanks!

PR #9240, opened 2025-03-17
Allow replacement of Drawer\Utils using the service container

felabrecque:

Might be, didn't asked.

Yes

One PR only

Not currently. Don't think it's required.

TLDR; Allow to use enhanced Utils class in my code.

Story:

In my project, I have livewire component inheritance.

The parent class is a "Student" class. The child is a "Teacher".
The "Teacher" share almost the same view and functionality as the "Student".

BUT, the "Teacher" has more features. This can easily be done by adding code to the component class.
Also the "Teacher", sometimes, must also have LESS features; something that the "Student" can see/do should not be done by the "Teacher".

In PHP, class overloading allows you to make a property/method MORE visible, not less.

The goal is to share as much code between the "Student" component and the "Teacher" component by using inheritance.

For example, we have a component representing a Questionnaire. The "Student" must be able to answer it.
The "Teacher" has pretty much the same template. But he must not be able to answer the questionnaire, not just by hiding the input fields/buttons, but by really blocking the execution of the code used by the student.

In a Livewire component, my understanding is that there is 2 main things that the javascript frontend can call/read: public methods and public properties.

For public methods, overloading the methods in the child component (Teacher) and throwing an exception block the teacher from calling the methods it should not be allowed to run. This covers all CRUD cases (Create, Read, Update & Delete) on the public methods.

But for public properties, it's a lot more tricky.

Throwing exceptions in livewire hooks doesn't work because the public property lookup is done even if it's not used anywhere in the "Teacher" template. The "Teacher" component becomes useless because while loading, the public property is found and the hook exception is thrown, crashing even the first rendering.

One way tthat I found that is working is to replace the "getPublicProperties" static method in Livewire\Drawer\BaseUtils class.
This makes the public property behave like a protected property. The blade can still render it (more on that later), but the javascript can no longer fetch/set the property value.

But the Livewire\Drawer\Utils class is called directly (without using the service container); not allowing me to replace it with my enhanced version.

Using Laravel service container to get the class fixes that problem; I can now simply bind my enhanced version in my application service provider.

@calebporzio I don't know if tests are required. Also, I could create documentation explaining a little bit Livewire component inheritance and how to add & remove feature in child components.

N.B. My enhanced version adds a filter that check the remaining public properties for an attribute I created called "Unreadable".
This way I can easily, in the child (Teacher) component, disable which public variables I don't want anymore by simply adding a PHP attribute. Obviously, the child (Teacher) blade template must also be adapted to remove all usage of these variables. It can be done through Laravel blade component slots or even better, by using policies/gates to check if the user has access to read the value.

### DIFF
--- a/src/ComponentHookRegistry.php
+++ b/src/ComponentHookRegistry.php
@@ -56,7 +56,7 @@ class ComponentHookRegistry
         }
 
         on('update', function ($component, $fullPath, $newValue) {
-            $propertyName = Utils::beforeFirstDot($fullPath);
+            $propertyName = app(Utils::class)::beforeFirstDot($fullPath);
 
             return static::proxyCallToHooks($component, 'callUpdate')($propertyName, $fullPath, $newValue);
         });
--- a/src/Concerns/InteractsWithProperties.php
+++ b/src/Concerns/InteractsWithProperties.php
@@ -9,15 +9,15 @@ trait InteractsWithProperties
 {
     public function hasProperty($prop)
     {
-        return property_exists($this, Utils::beforeFirstDot($prop));
+        return property_exists($this, app(Utils::class)::beforeFirstDot($prop));
     }
 
     public function getPropertyValue($name)
     {
-        $value = $this->{Utils::beforeFirstDot($name)};
+        $value = $this->{app(Utils::class)::beforeFirstDot($name)};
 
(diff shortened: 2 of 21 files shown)


calebporzio:
Sorry, going to hold off on this right now. thanks

PR #9246, opened 2025-03-24
Fix bug with component pagination when wire:navigate was previously used

N1ebieski:
Fixes https://github.com/livewire/livewire/discussions/6304

Yes, I've reported this bug right after the release Livewire v3 > https://github.com/livewire/livewire/discussions/6304, almost 2 years ago

This bug was reported partly in other discussions also, for example > https://github.com/livewire/livewire/discussions/6722#discussioncomment-10894770

Yes

No

It includes 2 failing tests.

The problem is complicated, so it's best to watch this short video first:

https://github.com/user-attachments/assets/c032855f-cd58-49b0-a914-3cd48983fd5c

What is the problem?

When we use links with `wire:navigate`, Livewire saves the entire html document to cache. So, when we use back/forward buttons, Livewire restores the entire html document from cache.

However, when we use pagination links, Livewire doesn't save document to cache, because pagination links only change the state of the component they are in. Despite the fact that clicks on pagination links generate pushState to browser history state. In their case, Livewire only saves the state of the component, not the entire html document.

Therefore, let's trace what happens in the video:

1. I'm on the "hello-page"
2. I click on the "pagination-page" `wire:navigate` link -> Livewire saves "hello-page" document to cache with the key: "hello-page-"`Math.random()`, Livewire saves page 1 document to cache with the key: "pagination-page-"`Math.random()`
3. I click on the page 2 in the pagination -> Livewire saves the alpine state to history state
4. I click on the page 3 in the pagination -> Livewire saves the alpine state to history state
5. I click on the page 4 in the pagination -> Livewire saves the alpine state to history state
6. I click on the back button -> Livewire tries to restore my entire document, but in the cache is only a snapshot of page 1, not 3

This PR adds a failing test for this problem and my fix. I suggest adding the previousUrl to the state. If the previousUrl is not set, then the popstate refers to the previous state of the component, not the return to the previous page. So Livewire should only restore the state, not the entire html document.

So after the fix:

https://github.com/user-attachments/assets/bd76e333-abdb-449e-8930-3768f1f85cc1

--------------

There is one more problem that will occur even if we fix the previous one. Livewire always replaces the entire html document to cache under the `snapshotCache.currentKey` if this one exists. 

Another trace (I warned that the problem is complicated! :P)

1. I'm on the "hello-page"
2. I click on the "pagination-page" `wire:navigate` link -> Livewire saves "hello-page" document to cache with the key: "hello-page-"`Math.random()`, Livewire saves page 1 document to cache with the key: "pagination-page-"`Math.random()`
3. I click on the page 2 in the pagination -> Livewire saves the alpine state to history state
4. I click on the page 3 in the pagination -> Livewire saves the alpine state to history state
5. I click on the page 4 in the pagination -> Livewire saves the alpine state to history state
6. I click on the back button -> now, after the fix, Livewire doesn't restore the entire document, only the state, so i back to page 3. Very good
7. I click on the back button -> i back to page 2. Good.
8. I click on the "hello-page" `wire:navigate` link -> Livewire saves "pagination-page?page=2" document to cache with the key: "pagination-page-"`Math.random()`, Livewire saves "hello-page" document to cache with the key: "hello-page-"`Math.random()`
9. I click on the back button -> i back to page 2 again. Good.
10. I click on the back button -> i back to page 1 but with the entire document from page 2. Bad.

Why did this happen? Because Livewire in point 8 replaced the entire html document into a cache under the same key as in point 2 -> "pagination-page-"`Math.random()`, despite the fact that the query param has changed - ?page=2 vs `snapshotCache.currentKey` for ?page=1

This PR adds another failing test for this problem and my fix. I suggest that Livewire checks that the `snapshotCache.currentKey` matches the current url. If it does not, need to use a new key.


luanfreitasdev:
This worked well in my testing

nicolagianelli:
I don't know if it's related or still something similar, but there is a bug with wire:navigate when the page loading throws an Exception or a 404 error.

As you can see in the two videos below, when you try to navigate to a page that throws an Exception (in dev mode), something goes wrong with the styles and scripts — the page’s style and functionality don't work as expected until it's reloaded. After that, the page works normally, but the back button doesn't correctly load the previous page.

The same happens with a 404 error page, although in this case only the styles are affected. If you watch the video, you can see how the styles aren’t applied correctly after navigating to and back from the 404 error page.

I think the problem happens because Livewire runs the inline styles and scripts from the Laravel error pages, and the inline styles are more specific than the Tailwind classes.

To reproduce the issue, I simply modified the Livewire starter kit by adding a mount method in the "register" page.
- In the first case, I threw an exception using **throw new \Exception('...')**.
- In the second case, I used a basic **abort(404)**.

https://github.com/user-attachments/assets/d19cf32c-0b30-4106-92f9-ab68b0af2836

https://github.com/user-attachments/assets/8677c443-a0c7-41e0-aefd-3194ab877deb

binaryfire:
@nicolagianelli I'm hitting the same issue. Styles and JS components are visually broken when using the back button from an error page.

kresnasatya:
Looking forward to this one since last year. Right now, the `wire:navigate` (I'm using Livewire version 3.6.4) doesn't have ability to sync with URL query params (the browser URL). If the `wire:navigate` has this ability, it will be more superpower! 🦸‍♂️⚡️
Here's the current situation.

https://github.com/user-attachments/assets/de3aa360-dbe0-440b-9bef-05a4e233c049

calebporzio:
Closing this PR as it's been open for a long time and now has merge conflicts. If still relevant, please re-submit with an updated branch and passing tests. Thank you!

Btw @binaryfire - the back button error page problem has been patched in v4 👌

PR #9251, opened 2025-03-26
show @livewireScriptConfig at top of page in documentation example

rowinf:

It seems to be an issue others are facing, I commented in this discussion https://github.com/livewire/livewire/discussions/9181

yes

no

No, it's a documentation change

I was importing Livewire into my app.js to listen for events, but this was causing livewire to boot twice with `Livewire.start()`. The auto-injected script appears to look for a window variable that was being declared too late when I put `livewireScriptConfig` at the bottom of the body.

Thanks for contributing! 🙌

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -127,12 +127,12 @@ To address this issue, we need to inform Livewire that we want to use the ESM (E
 <head>
     <!-- ... -->
     @livewireStyles
+    @livewireScriptConfig <!-- [tl! highlight] -->
     @vite(['resources/js/app.js'])
 </head>
 <body>
     {{ $slot }}
 
-    @livewireScriptConfig <!-- [tl! highlight] -->
 </body>
 </html>
 ```


PhiloNL:
Would you mind testing with a fresh installation of Livewire if this is an issue. It should not auto-inject the Livewire library whenever the `@livewireScriptConfig` is present. 

https://livewire.laravel.com/docs/installation#manually-bundling-livewire-and-alpine

> When Livewire detects the @livewireScriptConfig directive, it will refrain from injecting the Livewire and Alpine scripts. If you are using the @livewireScripts directive to manually load Livewire, be sure to remove it. Make sure to add the @livewireStyles directive if it is not already present.

joshhanley:
@rowinf thanks for the PR! I don't think moving this to the head tag is the solution, so I'm going to close this. Can you create a demo repo that shows the issue, so we can investigate? Thanks!

PR #9268, opened 2025-04-04
Revert Breaking Change to Enum Validation in 3.6.2

dalbo:
* Revert changes to EnumSynth.php from https://github.com/livewire/livewire/pull/9231

* Update EnumUnitTest.php to test bahaviour of Enums not being converted to their backing value

Yes, discussion at https://github.com/livewire/livewire/discussions/9236 and on the PR at https://github.com/livewire/livewire/pull/9231 confirms multiple people are affected by this change

Yes

No

Yes, changed the test that was added in 3.6.2 to test for proper Enum validation

A breaking change was introduced in 3.6.2 to the way that Enums are validated. The Enum was converted into its backing value (string or int) before being validated. As demonstrated in comments on the PR and in the discussion, this broke several different validation strategies for Enums.

This restores the previous and expected behaviour of validating Enums as Enums. 

Thanks for contributing! 🙌

### DIFF
--- a/src/Mechanisms/HandleComponents/Synthesizers/Tests/EnumUnitTest.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/Tests/EnumUnitTest.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Mechanisms\HandleComponents\Synthesizers\Tests;
 
+use Illuminate\Validation\Rule;
 use Livewire\Livewire;
 use Tests\TestComponent;
 use ValueError;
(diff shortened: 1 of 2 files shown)


src/Mechanisms/HandleComponents/Synthesizers/Tests/EnumUnitTest.php
```diff
@@ -34,7 +35,7 @@ public function test_an_enum_can_be_validated()
         Livewire::test(ComponentWithValidatedEnum::class)
             ->call('save')
             ->assertHasErrors('enum')
-            ->set('enum', ValidatedEnum::TEST->value)
+            ->set('enum', ValidatedEnum::TEST)
```
joshhanley, line 38: We need to change this back to the setting the value as a string, to ensure Livewire's Enum hydration actually functions. Setting to the Enum already will potentially skip that.

```suggestion
            ->set('enum', ValidatedEnum::TEST->value)
```

joshhanley (approved):
Ah shoot! Thanks for fixing it.

calebporzio:
Thanks!

PR #9277, opened 2025-04-08
Use JSON encoding instead of HTML encoding for the object ID

AJenbo:
# The problem

The `@this` Blade directive currently renders the component ID using:

```php
window.Livewire.find('{{ $_instance->getId() }}')
```

This uses HTML encoding (`e()`), which can lead to subtle issues if the component ID ever includes characters that get escaped (e.g., `&`, `"`, `<`, `>`). While unlikely in most cases, it's not robust.

# The solution

This PR switches the directive to use `@json()` instead of string interpolation:

```php
> window.Livewire.find(@json($_instance->getId()))
```

This ensures the component ID is safely and correctly encoded for use in JavaScript, avoiding any potential edge cases with special characters.

# The conversation

This is a safer default, especially as Livewire becomes more dynamic and interacts more deeply with JavaScript. It's a small change, but improves reliability and removes assumptions about what characters are safe in the ID.

### DIFF
--- a/src/Mechanisms/ExtendBlade/ExtendBlade.php
+++ b/src/Mechanisms/ExtendBlade/ExtendBlade.php
@@ -37,7 +37,7 @@ class ExtendBlade extends Mechanism
 
     function boot()
     {
-        Blade::directive('this', fn() => "window.Livewire.find('{{ \$_instance->getId() }}')");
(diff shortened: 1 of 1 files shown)


AJenbo:
Closing since I see the intent here is to use it in properties, my issue is that it doesn't work properly in the context of a script tag, but I don't see a good way of making it work for both.

PR #9281, opened 2025-04-10
[WIP] Adds type safe methods for properties

MrPunyapal:
let's say we are dispatching any event with the same name where we use any properties with concating with some prifix.

```php
$this->dispatch('chat:updated.'.$this->pull('chatId'));

// This will gonna fail PhpStan...
//  ------ -----------------------------------------------------------------------------
//  Line   Livewire\Chats\Save.php
//  ------ -----------------------------------------------------------------------------
//  65     Binary operation "." between 'chat:updated.' and mixed results in an error.
//         🪪  binaryOp.invalid
// ------ -----------------------------------------------------------------------------
```

but if we have some methods like `pullString`, `pullInt` and other similar methods, it can help to pass the PhpStan with convenient of dev.

```php
$this->dispatch('chat:updated.'.(string) $this->pullInt('chatId'));
$this->dispatch('chat:updated.'.$this->pullString('chatId'));
```

### DIFF
--- a/src/Concerns/InteractsWithProperties.php
+++ b/src/Concerns/InteractsWithProperties.php
@@ -113,6 +113,30 @@ trait InteractsWithProperties
         return $beforeReset;
     }
 
+    /**
+     * @return string
+     */
+    public function pullString($property)
+    {
+        $value = $this->getPropertyValue($property);
+
+        $this->reset($property);
+
+        return is_string($value) ? $value : '';
+    }
+
+    /**
+     * @return int
+     */
+    public function pullInt($property)
+    {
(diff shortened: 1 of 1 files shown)


MrPunyapal:
@calebporzio does this make sense to you? 🤔

MrPunyapal:
cc: @joshhanley

joshhanley:
@MrPunyapal thanks for the PR! We're going to hold off on this one for now as you could add an inline cast which should help. But we'd be open to adding a docblock with a return value to the `pull()` method if that helps static analysis.

PR #9282, opened 2025-04-11
Fix incorrect `getValue()` exception message labelled with `set`

rinodrummer:

It is needed to enhance the understanding of errors that happen when using Livewire, but the fix is so tiny that it shouldn't need further discussions.

Yes

No

No because there were no matching tests.

The LivewireAttribute class exposes a getter and a setter for its value, which may be unaccessible in same cases, but the thrown exceptions have the same message.


lrljoe:
Makes sense!

joshhanley:
@rinodrummer thanks!

PR #9284, opened 2025-04-12
Add synth for html string

mokhosh:
If you use the new `AsHtmlString` cast on a model attribute that was added in [v12.4](https://laravel-news.com/laravel-12-4-0) you will get:

```exception
Property type not supported in Livewire for property: [{}]
```

This makes Livewire work with `HtmlString`s seamlessly.

### DIFF
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -17,6 +17,7 @@ class HandleComponents extends Mechanism
         Synthesizers\CarbonSynth::class,
         Synthesizers\CollectionSynth::class,
         Synthesizers\StringableSynth::class,
+        Synthesizers\HtmlStringSynth::class,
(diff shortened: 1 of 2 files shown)


Cbrad24:
I believe the synthesizer needs to be loaded first, you can add it to the default synth list here:
https://github.com/livewire/livewire/blob/56aa1bb63a46e06181c56fa64717a7287e19115e/src/Mechanisms/HandleComponents/HandleComponents.php#L14-L25

calebporzio:
Ahhh yeah this is a good idea at first glance, but now I'm thinking a little deeper and i think it might be kind of attack vector if people don't lock the htmlstring property right? because someone could update teh value with wire:model on the frontend to include some XSS stuff in it?

idk, might be better left to userland right? people can always store the html in the public property and then echo without escaping: {!! !!} for the equivelant experience? thanks, but thinking I'll pass on this for now

PR #9286, opened 2025-04-14
Add phpdoc with types to helpers.php

AJenbo:
## The problem

The global helper functions defined in `helper.php` are missing type annotations. This limits the ability of IDEs to infer argument types, return values, and callable signatures, especially for functions like `once()`, `revert()`, `trigger()`, and event listener bindings such as `on()` and `after()`. Without these annotations, static analysis tools like PHPStan or Psalm struggle to validate or autocomplete these functions accurately, particularly when closures or generics are involved.

## The solution

This PR adds PHPDoc annotations to all exported functions in `helper.php`, including type-hinted parameters, return types, and closure signatures where applicable. Generic templates were used where appropriate (e.g. for `of()`), and event-related functions were aligned with the internal behavior of `Livewire\EventBus`.

## The conversation

While these annotations don't affect runtime behavior, they significantly improve the developer experience. They enable better autocompletion, reduce false positives in static analysis, and provide clearer context when reading or using these helpers—especially in codebases that use closures, middleware chaining, or custom event handling. This is a low-risk, high-reward improvement to DX.

## Looking ahead

It would be nice to introduce interfaces for the anonymous classes returned by some of these helpers—most notably `store()`. Currently, return types for those helpers are limited to `object`, even though they follow a predictable contract (e.g. duck-typing `DataStore` behavior). Defining a shared interface would allow us to hint these more explicitly and improve both tooling support and static analysis coverage.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -5,6 +5,10 @@ namespace Livewire;
 use ReflectionClass;
 use Illuminate\Support\Str;
 
+/**
+ * @param ?string $string
+ * @return object
+ */
 function str($string = null)
 {
     if (is_null($string)) return new class {
@@ -16,6 +20,10 @@ function str($string = null)
     return Str::of($string);
 }
 
+/**
+ * @param object $obj
+ * @return object
+ */
 function invade($obj)
 {
     return new class($obj) {
@@ -59,6 +67,10 @@ function invade($obj)
     };
 }
 
+/**
+ * @param callable(mixed ...$params): mixed $fn
+ * @return callable(mixed ...$params): mixed
+ */
 function once($fn)
 {
     $hasRun = false;
@@ -72,11 +84,20 @@ function once($fn)
     };
 }
 
+/**
+ * @template TType
+ * @param TType ...$params
+ * @return list<TType>
+ */
 function of(...$params)
 {
     return $params;
 }
 
+/**
+ * @param mixed &$variable
+ * @return callable(): void
+ */
 function revert(&$variable)
 {
     $cache = $variable;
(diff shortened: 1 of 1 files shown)


lrljoe:
Would suggest that while yes, off doesn't return anything, that you put that change in a separate PR, so that this one purely addresses docblock rather than removing a return.

Just prevents any confusion!

src/helpers.php
```diff
@@ -86,34 +107,71 @@ function revert(&$variable)
...
 }
 
+/**
+ * @param string $name
+ * @param callable(mixed ...$params): mixed $callback
+ * @return void
+ */
 function off($name, $callback) {
-    return app(\Livewire\EventBus::class)->off($name, $callback);
+    app(\Livewire\EventBus::class)->off($name, $callback);
```
AJenbo, line 168:
```suggestion
    return app(\Livewire\EventBus::class)->off($name, $callback);
```

joshhanley:
@AJenbo thanks for the PR! These helpers are used internally and aren't really something people should be using in their apps (they're not documented). As such I'm going to close this PR.

joshhanley:
@AJenbo can you show me an example of where that happens and what your static analyser shows? Thanks!

joshhanley:
@AJenbo please excuse my naivety, I don't use static analysers so don't particularly understand how they work, but why is the static analyser analysing vendor code? Can't you tell it to ignore it?

AJenbo:
@joshhanley Blade directives like `@endscript` generate PHP code that ends up directly in the user's application, so this isn't vendor code. When a static analyser like PHPStan checks the user's code, it sees whatever those directives expand into.

Think of directives as a shortcut for copy-pasting boilerplate into your app. When a directive like `@endscript` expands, it results in something like:
```php
\Livewire\store($this)->push('scripts', $__output, $__scriptKey);
```

That line becomes part of the user's codebase. So it’s their code making a call into Livewire's API, and the analyser wants to verify that the call is valid. It doesn’t care how `store()` works internally—it just needs to know what type it returns, since the user code is calling `push()` on the result.

Here’s a simplified example:
```php
VendorsBlackBox("bob") + 42;
```
To validate this, PHPStan needs to know:
- Does `VendorsBlackBox()` accept a string like "bob"? If it expects, say, a `VendorDto` instance, that would be flagged as an error.
- What does it return? If it returns an `int`, the addition is fine. But if it returns an `array` or `mixed`, it’ll raise a warning or error because the operation could fail at runtime.

So it’s not about analysing vendor internals. It’s about ensuring the user's code is calling the vendor’s public API correctly.

One of the main goals of a static analyser is to answer:
> "Am I using the vendor code like I’m supposed to?"

That’s why return and argument types matter on public APIs. They let tools help users write safer, more predictable, and easier-to-maintain code. If, in a future version, `VendorsBlackBox()` changes to require a different argument, the analyser can detect mismatches in user code without them needing to manually test every usage.

If we simply “ignored” vendor APIs, checks like that wouldn't be possible. It’s similar to how vendors document their public API on a website—those same entry points should be documented with types so that tools can reason about them.

It gets a bit quirky with Blade, since it’s more of a meta language than a traditional one. So I understand how the boundary can get a little fuzzy. I'd just suggest minimizing the generated code that gets injected into the users application. For example, wrapping it in a helper like this would reduce exposure of internals:
```php
\Livewire\directives\storeAndPush($this, 'script', ob_get_clean(), $__scriptKey);
```

For purely internal code that users aren't supposed to call, it's totally fine to rely on mental models. The analyser won’t be checking those paths anyway.

I think part of the confusion comes from the blurred line between what’s internal and what ends up in user code, which also helps explain the misalignment that’s led to some of my PRs being closed with feedback about user-documenting: https://github.com/livewire/livewire/pull/9312#issuecomment-2917694073
To clarify, I’m not trying to document all the internals—just the parts that effectively become part of the user’s application through directives or other indirect exposure.

PR #9252, opened 2025-03-27
[WIP] Add interruptible requests

calebporzio:

PR #9302, opened 2025-04-23
Ensure component exists before pooling children

AlexandreBonaventure:
Hello,
I'm creating this PR to workaround an issue we've been seing for a long time but was never able to reproduce reliably. 
Finally this morning I had some time to dig into this and was able to decypher what is happening.
🔴 **It's affecting a lot of users on production randomly and completely breaking the app because there is javascript error thrown synchronously (not inside a Promise). I have sentry reports to back it up.**

Capture d’écran, le 2025-04-23 à 11 19 20

Here's a diagram to describe the issue in a nutshell:
image

We have table components with polling updates that can run into race conditions when navigating between pages, because updates are not aborted when navigating between pages. 
Here's a couple screenshot that shows the issue in the debugger:
 
Capture d’écran, le 2025-04-23 à 11 17 52
Capture d’écran, le 2025-04-23 à 11 16 35

Sidenote: In general because Livewire does not abort fetch request when navigating away, or commiting more recent changes, we've seen a couple weird behaviour and race conditions happening. I'm guessing this is also what is motivating this PR: https://github.com/livewire/livewire/pull/9252 (despite not exactly being the same problem)
but some users also reported similar problems like:
https://github.com/livewire/livewire/discussions/9028
https://github.com/livewire/livewire/discussions/9030
https://github.com/livewire/livewire/discussions/2480
https://github.com/livewire/livewire/discussions/7476
https://github.com/livewire/livewire/discussions/7464

I believe there are a lot of different ways to fix the issue, and I wish we could have a more robust solution moving forward, however like I said, this issue is kind of critical for us, and hopefully this is a good workaround for the time being, as it breaks the app entirely and causes a lot of report of mad customers complaining about the app suddenly "not working" anymore.

I'm not sure this is easily testable due to the complexity of pooling and the specific timing, but please let me know what I can do to merge this fix as soon as possible. Thank you!


joshhanley:
@AlexandreBonaventure thanks for the PR! I didn't see this PR earlier, but we've also been working on a solution to the `component not found` error. Please see PR #9339.

Can you test out that PR and let us know if it fixes your issue or not?

If it does, we will close this PR in favour of that one.

If it doesn't fix the issue, would you be able to give me some detailed steps on how to reproduce it, so we can review this PR alongside the other changes we have made.

Thanks!

calebporzio:
Thanks for this - the real problem is we need "interuptible" requests as a concept in livewire - specifically in this case where a wire:navigate occurs while a long-running request is waiting to come back.

Going to close this and try solving the deeper problem.

Here's a WIP: #9252

AlexandreBonaventure:
@joshhanley https://github.com/livewire/livewire/pull/9302#issuecomment-2914827186 
We've tested main branch on staging and while difficult to say it 100% fix the issue, I'm fairly positive it did!

sidenote: we ran into an unrelated issue on main https://github.com/livewire/livewire/pull/9302#issuecomment-2914827186

joshhanley:
@AlexandreBonaventure awesome, thanks for testing!

No sure what the sidenote is, you seem to have linked my comment again 😅

PR #9306, opened 2025-04-24
Amend the existing comment style of livewire.js

no-onoma:
Existing comment style is offending livewire-starter-kit. In developer tools (console) of firefox, it's showing **Uncaught SyntaxError** with missing `)` and I'm unable to register or login. And laravel is not logging any error whatsoever. 
Maybe because of using  `(*!`,  js parser is having problem.


lrljoe:
dist/  are built files, you'll need to update the root js files instead

lrljoe:
Think you may want to do a pr against the relevant package

PR #9309, opened 2025-04-28
Remove invalid void return

AJenbo:
It's not valid to return void, in this case it was automatically cast to null since there isn't any return type on the parent function but better fix it now then it becoming an issue later.


joshhanley:
@AJenbo thanks!

PR #9312, opened 2025-04-30
Add types to HandlesEvents

AJenbo:
This resolves warnings when type checking is enabled on projects that use Livewire when they specify the $listeners for a component.


calebporzio:
thinking if we aren't adding them everywhere on Component.php we shouldn't only add them *somewhere*

PR #9313, opened 2025-05-01
Add .hash Modifier to wire:model

luanfreitasdev:

https://github.com/livewire/livewire/discussions/9311

Yes
No
Yes

---

### Add .hash Modifier to wire:model

This PR adds a **.hash** modifier to the wire:model, which transforms the attribute value into a hash when placed.

### Benefits:

* Enhanced Privacy – Masks sensitive wire:model values in the DOM without compromising functionality.
* Seamless Integration – Works alongside existing modifiers (e.g., wire:model.live.hash).
* Efficient Lookup – Uses a WeakMap for reversible mapping, ensuring performance.

### How it works:

* Alpine.js intercepts inputs with wire:model.hash.
* The original value is hashed and stored in a WeakMap.
* The hash is rendered in the DOM (e.g., wire:model="4ghta7nh943").
* Livewire resolves the original value via the WeakMap when needed.

### Example Usage:

```blade
<div>
    <p>Name: {{ $user->name }}</p>
    <input type="text" wire:model.live.hash="user.name" />
</div>
```

### DOM output:

```html
<input type="text" wire:model.live="4ghta7nh943">
```

https://github.com/user-attachments/assets/05f7491c-c7ce-4dcc-b0f9-71c9b417aa61


devajmeireles:
OMG! Very useful.

js/hash.js
```diff
@@ -0,0 +1,30 @@
...
+        })
+})
+
+export function get(el) {
+    return hashMap.get(el)
+}
+
+export function destroy(el) {
+    hashMap.delete(el)
+}
```
danie-ramdhani, line 30: block javascript from the site setting once and people can see the `wire:model` value that is not hashed... so, is it good to hash it from frontend side?

danie-ramdhani:
i have mixed feeling about this and i have some questions

- what happen when using `wire:model.hash` inside blade component that using data attribute from `wire:mpdel.hash`? (i.e. data-wire-model="{{ $attributes->wire('model')->value() }}") is the value is hashed too?
- if using `wire:model` is not safe because exposes database column names, should it just hashed by default?

js/hash.js
luanfreitasdev: This is possible, but there are also other things that we must consider besides the PR, such as the payload (snapshot) that has the decrypted values, and this is not the case with this PR. 😃

Another suggestion would be to use a blade directive as mentioned in the discussion, but it's not very pretty.

luanfreitasdev:
> i have mixed feeling about this and i have some questions
> 
> * what happen when using `wire:model.hash` inside blade component that using data attribute from `wire:mpdel.hash`? (i.e. data-wire-model="{{ $attributes->wire('model')->value() }}") is the value is hashed too?
> * if using `wire:model` is not safe because exposes database column names, should it just hashed by default?

1 - The value will always match the original wire:model, but we can test

2 - We don't always want to expose column names when looking at HTML, and in some cases, the names may reveal a pattern.

lrljoe:
Why are you exposing anything sensitive?  Name your properties appropriately, and don't directly bind to a Model.  This is disabled by default for a reason.

luanfreitasdev:
> Why are you exposing anything sensitive? Name your properties appropriately, and don't directly bind to a Model. This is disabled by default for a reason.
> 
> So what if a user figures out that you have named a column in your database"top_secret_password". Security through obscurity isn't security at all.
> 
> But using a proper Form approach, you could have wire:model='form.field1' then just do your mapping in your form.
> 
> But your security should be in your logic, not reliant on this kind of approach

The directive name is hash and not secure. It is disabled but not deprecated 🙂

danie-ramdhani:
> But using a proper Form approach, you could have wire:model='form.field1' then just do your mapping in your form.
> 
> But your security should be in your logic, not reliant on this kind of approach

Exactly, think `wire:model` is like `name` attribute for input elements.

Also this will make user think `wire:model` is insecure when using it inside form (see https://github.com/livewire/livewire/pull/8586#issuecomment-2206737500).

luanfreitasdev:
This should be the default behavior, though we can simplify its usage.
If someone includes a hash in wire:model, it’s because they needed to modify the front-end name.

lrljoe:
> This should be the default behavior, though we can simplify its usage.
> If someone includes a hash in wire:model, it’s because they needed to modify the front-end name.

But again, why put that overhead on the frontend?  Keep in mind, that the frontend will still know the mapping, so you're just obscuring it.  Someone who wants to know the mapped property can still discover it regardless.

luanfreitasdev:
How did you discover the property's name if it is not in the snapshot? The PR's proposal is only to obfuscate the name of the property as a directive, following the same concept of modifiers in Livewire.

Perhaps if this were the case, there could be an end-to-end obfuscator in another pull request, thus avoiding "fake" property names, and legacy_model_binding would no longer be a problem.

PR #9318, opened 2025-05-07
Fix form properties fill from URL

robert-stanciu:

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportAttributes/Attribute.php
+++ b/src/Features/SupportAttributes/Attribute.php
@@ -3,6 +3,7 @@
 namespace Livewire\Features\SupportAttributes;
 
 use Livewire\Component;
+use Livewire\Features\SupportFormObjects\Form;
 
 abstract class Attribute
 {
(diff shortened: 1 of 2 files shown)


robert-stanciu:
Hello @calebporzio, could you take a look?

lrljoe:
Would recommend not using fill in this approach.

Generically, to my mind, your Form is your reusable create/update interface.  It is how you map your frontend forms, against your component/backend logic.

Developers should map querystring against a form object themselves via the Component, rather than the Form.
Your approach also potentially has complexities/conflicts where there are several components on a page, each with Forms, naming conflicts etc.

Keeping querystring logic in the Component keeps it consistent, and simple.

Otherwise this is potentially a slippery slope, if Component level features start being added to the Form, then it defeats the purpose of the Form being a lightweight way to reuse frontend/backend interfaces.

joshhanley:
@robert-stanciu thanks for the PR! I'm going to close this as I've implemented the functionality for this in PR #9702 which makes use of the Synths system instead of using form->fill(). Thanks for being willing to contribute.

PR #9319, opened 2025-05-07
Add new hooks for navigation (navigate-swapping, navigate-swapped)

nicolagianelli:
There have been several discussions about the missing ability to update the document before the HTML swap occurs during navigation with "wire:navigate". This is particularly relevant for use cases like applying dark mode without causing flickering (example: #9230).
After support for replacing HTML attributes was introduced (#9149), some users encountered problems ensuring dark mode is applied correctly in time.

Currently, one way to handle dark mode is using a script inside the <body> tag. However, a more robust and flexible solution would be to provide hooks that allow developers to update the document before/after the HTML swap, but prior to the final page render.
This would give developers better control over document-level changes.

With this PR I introduced two new hooks: **livewire:navigate-swapping** and **livewire:navigate-swapped**.

- **livewire:navigate-swapping** is triggered before the HTML swap occurs. It also receives the new document, allowing developers to update it as needed before it's swapped into the DOM.
- **livewire:navigate-swapped** is triggered after the swap has taken place, but before Livewire proceeds with its internal logic and page rendering.


lrljoe:
livewire:navigating is fired immediately before the HTML is swapped

nicolagianelli:
> livewire:navigating is fired immediately before the HTML is swapped navigated is fired after
> 
> However your placement of your new events may result in oddities in results as scripts won't have run which may amend the dom.
> 
> Firing the event, doing a load of changes, then returning to complete the lifecycle is also going to cause a slow down.
> 
> Not sure this really adds much of a benefit just for the use case of toggling dark mode?

livewire:navigating is fired before, yes, and doesn't give you the ability to update the html that will be used for the new page.
As I said, right now, if you use the available hooks to add/remove classes in the DOM, in wire:navigating is useless, it's too early if you try, like the wire:navigate.
The only way is using wire:navigated but because Livewire uses a setTimeout to wait for scripts it will wait for the styles also before emitting wire:navigated producing flickering etc.
And as you said, scripts shouldn't have been run when using these hooks, because these are there to help you add something BEFORE Livewire evaluate the new DOM.
Can't tell if it's worth, but it's a missing piece, if you want to keep or remove something in the new DOM, now you can't do it clientside.

calebporzio:
to keep things light, sounds like we only need the swapping one and not the swapped one right?

lrljoe:
> to keep things light, sounds like we only need the swapping one and not the swapped one right?

Agreed, swapping makes sense, swapped doesn't.

nicolagianelli:
> to keep things light, sounds like we only need the swapping one and not the swapped one right?

Yes, potentially by using the swapping event and by updating the referenced new document we could edit clientside the new HTML before the swap happens and avoid darkmode flickering, injecting scripts etc.

joshhanley:
@nicolagianelli yesterday I was trying to replicate the issue with the darkmode flashing but couldn't replicate it and the author of #9230 said it was solved. Are you able to provide me a replication of the issue so I can test it? I'm in the process of reviewing all navigate issues and overhauling it.

nicolagianelli:
> @nicolagianelli yesterday I was trying to replicate the issue with the darkmode flashing but couldn't replicate it and the author of #9230 said it was solved. Are you able to provide me a replication of the issue so I can test it? I'm in the process of reviewing all navigate issues and overhauling it.

I tried again, and strangely, I can’t fully replicate the issue anymore.
However, part of the problem is still present if you apply `transition-colors` to elements like buttons or other components. An easy way to reproduce this is by adding `transition-colors` to the `<body>` tag. With a standard `navigated` event, you’ll notice a blink because dark mode gets reapplied after the page loads, causing a rerender. With my `navigate-swapping` approach, that doesn’t happen.

Even though the issue is now less noticeable, I still believe it’s worth using (and recommending) this event to avoid unnecessary rerenders for performance reasons.

Just a thought: maybe the issue is less noticeable now because of a change in Livewire’s behavior, perhaps it no longer removes attributes before receiving the new page? Not sure, but I remember that previously, it was easy to reproduce by simulating latency in the dev tools.

nox-4O4:
> I was trying to replicate the issue with the darkmode flashing but couldn't replicate it

@joshhanley The issue is still quite noticeable when using Bootstrap's input elements, as there is a `border-color` transition that leads to the input borders flashing white when navigating in dark mode.

See the following gif:
Flashing Input Fields


Switching between items "Test 1" and "Test 2" is done using `wire:navigate`. Dark mode is re-applied at the `livewire:navigated` event.

joshhanley:
@nox-4O4 yeah I actually just (only a few minutes ago!) managed to replicate this consistently when investigating a Flux bug. Going to run it past Caleb and see what he thinks.

joshhanley:
Going to close this as we've added an `onSwap()` hook to the `livewire:navigating` event in PR https://github.com/livewire/livewire/pull/9677

PR #9325, opened 2025-05-10
Add return docblock for Testable `invade()` method

PerryvanderMeer:
These docblocks provide some better autocomplete when using the `invade()` function during testing.

afbeelding

Thanks!


joshhanley:
@PerryvanderMeer thanks!

PR #9331, opened 2025-05-14
Add 'translate' prop to 'title' attribute

gdebrauwer:
Currently, you can not use the `#[Title]` if you want to show a translated title based on the user's locale. This PR fixes that.

```php
#[Title('some_component.title', translate: true)]
{
    // ...
}
```
For backward compatibility, it is disabled by default.

### DIFF
--- a/src/Features/SupportPageComponents/BaseTitle.php
+++ b/src/Features/SupportPageComponents/BaseTitle.php
@@ -9,5 +9,6 @@ class BaseTitle extends LivewireAttribute
 {
     function __construct(
         public $content,
+        public bool $translate = false,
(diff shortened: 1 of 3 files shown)


lrljoe:
You might want to update the docs so that this is a complete PR.  Otherwise nobody's going to know about this feature.

lrljoe:
> I think I'd rather just have people tackle translation at the blade component layout level like: `<title>{{ __($title) }}</title>` you know? thanks anyways.

Tbh - I'd agree with that!

Granted - there's nothing stopping @gdebrauwer from creating their own attribute and implementing translation.

PR #9341, opened 2025-05-23
Add navigate `preserve-scroll`

joshhanley:
# The scenario

Currently if you are using `wire:navigate` and you browse to a new page, navigate will always scroll the page to the top.

But there are scenarios where you don't want the page to scroll, for example if the page you're browsing to is a similar page and just the contents changes (see example of this in screen capture below).

Recording 2025-05-23 at 11 12 40

# The proposal

This PR adds support for a `preserve-scroll` modifier to the navigate feature.

It can be used like this:

```blade
<a href="/new-page" wire:navigate.preserve-scroll>New page</a>
```

Or if using javascript, like this:

```js
Livewire.navigate('/new-page', { preserveScroll: true })
```

I originally started with adding it as a modifier `wire:navigate.preserve-scroll`, but it needed to be registered manually with every different possible combination of modifiers which wasn't great. So decided to use an attribute instead. But it turns out that is a violation of the HTML spec and would need to be prefixed with `data-`, so have changed back to the modifier.

Will need to fix up this mess somehow...
```js
Alpine.addInitSelector(() => `[wire\\:navigate]`)
Alpine.addInitSelector(() => `[wire\\:navigate\\.hover]`)
Alpine.addInitSelector(() => `[wire\\:navigate\\.preserve-scroll]`)
Alpine.addInitSelector(() => `[wire\\:navigate\\.preserve-scroll\\.hover]`)
Alpine.addInitSelector(() => `[wire\\:navigate\\.hover\\.preserve-scroll]`)

Alpine.interceptInit(
    Alpine.skipDuringClone(el => {
        if (el.hasAttribute('wire:navigate')) {
            Alpine.bind(el, { ['x-navigate']: true })
        } else if (el.hasAttribute('wire:navigate.hover')) {
            Alpine.bind(el, { ['x-navigate.hover']: true })
        } else if (el.hasAttribute('wire:navigate.preserve-scroll')) {
            Alpine.bind(el, { ['x-navigate.preserve-scroll']: true })
        } else if (el.hasAttribute('wire:navigate.preserve-scroll.hover')) {
            Alpine.bind(el, { ['x-navigate.preserve-scroll.hover']: true })
        } else if (el.hasAttribute('wire:navigate.hover.preserve-scroll')) {
            Alpine.bind(el, { ['x-navigate.hover.preserve-scroll']: true })
        }
    })
)
```


lrljoe:
Given that the code supports both orders

navigate.hover.preserve-scroll
and
navigate.preserve-scroll.hover

Is it worth adding in the extra test, to ensure that nobody undoes the good work in a future change?

lrljoe:
Yep fair enough!

calebporzio:
good to merge

PR #9354, opened 2025-05-29
Fixed example in components doc

Ilyes512:
Fixed the example. It does not make sense to call `key()` on what is probably going to be a value of type int.


joshhanley:
@Ilyes512 thanks for the PR! But `key()` is used to signify the component's `wire:key` in the old `@livewire()` syntax in Blade. When it was written it was missed that there was a php function called `key()` hence the confusion. But that has been in production for years, so we can't change it now. As such I'm going to close this. Thanks for being willing to contribute!

joshhanley:
@Ilyes512 no worries! And yeah 😬

PR #9355, opened 2025-05-30
Allow `s3` to upload multiple files

utsavsomaiya:
### Why
Currently, when a user uploads multiple files, Livewire's global temporary file upload configuration does not support handling them properly. This pull request addresses that by enabling multiple file uploads using S3.

Now, onto the second issue: S3 doesn't support uploading multiple files in a single request. So, our current approach is to upload files one by one to S3 and send the file paths to the backend. Another idea I considered was to zip the files before uploading. However, managing zip files, including storing, extracting, and eventually providing access to users, adds unnecessary complexity. So, for now, uploading each file individually via separate requests is working as expected.

> [!IMPORTANT]
> One more thing to note: When uploading a single file to a local disk (e.g., the livewire-tmp directory), Livewire applies global validation before storing the file. However, when uploading to S3, we currently skip this validation step and directly store the file. The next focus should be to ensure the file is validated before being uploaded to S3 as well.

Okay, now, how do I want to test? I just made the public repo, which has a fresh Laravel app with a composer link to Livewire. Here is the repo: https://github.com/utsavsomaiya/demo

> Does it contain multiple, unrelated changes? Please separate the PRs out.
- Yes, this PR has fixed some typos on the specific related feature and also formatting

> Does it include tests? (Required)
- No, I need to write more dusk, but just fix the old test case

### Lifecycle of file upload
What is the file upload lifecycle in Livewire?
- When you upload a file, the first step is the startUpload call. If the configuration uses S3, it will generate the necessary headers and payload to avoid exposing environment variables. These headers and payload are then dispatched and shared with the frontend.
- The frontend then uses a PUT request to upload the file to AWS. The key difference is that AWS does not perform validation, whereas our local disks validate the file before storing it in the Livewire temporary directory.


utsavsomaiya:
composer link
image

utsavsomaiya:
https://github.com/user-attachments/assets/b47c0136-3b6c-4f1d-af6a-70ec328c1668

lrljoe:
Polite suggestion, avoid "fixing" whitepace correction, or fixing phpdoc typos in this type of PR, as it deviates attention from what is actually changing.

Keeping it to just "what is changing" makes it really easy for the core maintainers to easily review.

Also, suggest that you don't edit/PR changes to dist, dist is updated by the build script.  Your PR must work without any changes to dist. Avoiding PRing dist proves that your changes are in the right place, and will work longer term.

lrljoe:
Also avoid committing anything against dist/, as this is generated.

This minimises what needs to be reviewed, and ensures that your PR works long term

From experience, it's less likely that your PR will get merged if you have non-relevant code adjustments. 

 I'd recommend that you do not try to "fix" whitespace or typos in other areas when committing a new feature.

error1001es:
@utsavsomaiya

error1001es:
@lrljoe

calebporzio:

calebporzio:
Thanks for this PR @utsavsomaiya, however I'm just not comfortable merging this at this point in time. thank you

utsavsomaiya:
> Thanks for this PR @utsavsomaiya, however I'm just not comfortable merging this at this point in time. thank you

@calebporzio I will create new PR as claude code suggested this changes

PR #9356, opened 2025-05-30
Support wire:loading on parent methods

nicolagianelli:
Yes. There was a previous PR (#9094) introducing the idea of listening for parent method calls with `wire:loading`. This would be very useful, as we can currently call parent methods via `$parent.method`, but we cannot show a loading state for them using `wire:loading`.

Yes

No

Yes. Existing loading tests continue to pass, and a new test was added to validate the ability to listen to `$parent.method` calls.

The previous proposed solution introduced a .parent modifier for wire:loading, but that approach had some limitations:

- You could only listen to either parent methods or local component methods — not both.
- Additional frontend assets update was required to handle .parent variants (e.g. display: none rules).

✅ Proposed Improvement
This PR allows you to use $parent.method directly inside wire:loading, just like any other method.

Example Usage:

```
<button wire:click="$parent.save">Save</button>
<span wire:loading="$parent.save">Loading...</span>
```

This approach is more intuitive and consistent with current usage patterns. It enhances flexibility and simplifies the developer experience without requiring new syntax.


joshhanley:
@nicolagianelli thanks for the PR! Tests seem to be failing. I've run them all again to make sure it's not just something flaky and they're all failed again. Would you be able to look into that? Thanks!

nicolagianelli:
> @nicolagianelli thanks for the PR! Tests seem to be failing. I've run them all again to make sure it's not just something flaky and they're all failed again. Would you be able to look into that? Thanks!

Thanks for pointing that out! I checked and the issue was caused by the built assets, apologies for the oversight.

While reviewing the test, I realized I had implemented it incorrectly by mistake. After fixing the test and ensuring that all the loading indicators are covered, I noticed the implementation still wasn't working properly.

So I revised the logic, and here's the updated behavior:

If there are no `wire:target`, we only check whether the commit affects the component where `wire:loading` is used.
If there are targets (either from the component or the parent), we check the methods against the commit:

Use `parentTargets` when it’s a parent-level commit.
Use `componentTargets` when it’s a component-level commit.

This should now support all practical use cases. The only unsupported scenario is checking all methods on the parent (by using **$parent** without any specific method maybe...), but I don't see a strong case for adding extra logic just for that.

nicolagianelli:
@joshhanley
I've thought more carefully about the implementation, and I realized that with the current approach, if `wire:target` were to include both child and parent component methods, it wouldn't behave correctly. This is because they could be triggered simultaneously, potentially interfering with the loading behavior.

There are two possible paths forward:

1. Stick with the current implementation, but handle concurrency between parent and child explicitly. This would add complexity, but it would allow loaders that combine both parent and child methods.
2. Adopt the approach from the original PR, using a modifier like `wire:loading.parent`, which would strictly consider the parent context. In this case, targets would be declared as `wire:target="method"` (not `wire:target="$parent.method"`).

The advantage of option 2 is that it requires just a few small changes, avoiding unnecessary complexity. The only trade-off is the inability to mix parent and child methods in the same loader.

Personally, I lean toward **option 2**, as it offers a simpler and more maintainable solution for now.

With option 2, it's also possible to listen to all methods from the parent by simply using `wire:loading.parent` without specifying any target, since it would change the entire loading context to that of the parent.

Let me know your thoughts before I proceed. Thanks!

joshhanley:
@nicolagianelli hmm not sure, my thoughts would be that keeping it `wire:target="$parent.method"` keeps it consistent with the existing `$parent` syntax. In terms of the implementation, complexity is ok as I think it would be more beneficial for users to be able to mix and match than be restricted to just child or just parent. Plus it would require a bunch of extra CSS for all the `.parent` variants.

nicolagianelli:
> @nicolagianelli hmm not sure, my thoughts would be that keeping it `wire:target="$parent.method"` keeps it consistent with the existing `$parent` syntax. In terms of the implementation, complexity is ok as I think it would be more beneficial for users to be able to mix and match than be restricted to just child or just parent. Plus it would require a bunch of extra CSS for all the `.parent` variants.

I’ve updated the PR, and it should now support parallel requests. However, I wasn’t able to create a test to properly handle this, since the network requests are executed sequentially, at least on Windows.

sebGH06:
Hey guys !

I just found your discussion, and I am facing issue that seems similar

I have a livewire child component and inside the child is located an input text.

I want that input text to only be disabled when a particular parent method is executed

For informatio, that input text is also inside a wire:ignore directive.

At this time putting wire:loading.attr="disabled" on the input works well but reacts to every requests from any property or method.

Please let me know if a feature to target a specific parent method or property is soon available?

Many thanks!

joshhanley:
Closing this PR as it's been open for a long time with merge conflicts. If still relevant, please re-submit with the conflicts resolved. Thank you for the contribution!

PR #9350, opened 2025-05-29
Fix long filenames using `TemporaryUploadedFile`

joshhanley:
WIP

PR #9360, opened 2025-06-02
Fix long filenames in file uploads

joshhanley:
# The scenario

Currently if a user tries to upload a file with a long file name, a `UnableToWriteFile` exception is thrown.

image

Which is caught and then a `UnableToRetrieveMetadata` exception is thrown.

image

# The problem

The issue is that most file systems have a maximum filename length of 255 characters (macOS, Windows, and Linux).

But that is just for the filename length, full file-paths (file path + file name + file extension) can be longer. MacOS has a max full file-path of 1024 characters and Linux has a max of 4096 characters. But Windows has a max of 260 characters.

This issue occurs across all operating systems, but it's most common on Windows due to the small full file-path. What that means is typically a user's `livewire-tmp` directory, might have a path like `C:\laragon\www\myapp\storage\app\livewire-tmp\` which is already 46 characters out of 260 and that doesn't include the filename.

When Livewire uploads a file to the `livewire-tmp` directory, it generates a hased filename which includes the original filename embedded within it.

The hash consists of:
- 30 random characters
- `-meta` string
- a base 64 encoding of the original filename and extension
- the original file extension

For example if the original filename and extension is 150 characters long, when base64 encoded it becomes 200 characters long. Add the 30 random characters, `-meta` and the file extension, let's say `.png` and it's already 239 characters and that's before we even add the `livewire-tmp` directory path. So if we add the 46 characters from the example path above, we end up with 285 characters which is over the limited for Windows.

So why do we even need to encode and add the original filename to the temporary file name hash?

The reason is because it's the simplest way to store the original filename so it goes along with the temporary file. The temporary file name hash can be decoded to retrieve the original filename, which might be used in the UI, for saving, or for number of other reasons.

# The solution

To fix this problem, we need to remove the filename from the temporary file name hash and instead store it somewhere else.

There are two realistic options:
1. Store a mapping between the hashed temporary file name and the original filename in a meta data file
2. Find a way to store it on the `TemporaryUploadedFile` class

This PR implements option 1 and stores the original filename in a `.json` meta data file.

It's similar to PR #9172 which stored the data in a `meta/` folder in a file named the same as the original file. But the main difference is that PR would truncate the filename if it ended up too long. Where as this PR always uses the meta file.

I also explored option 2 in PR #9350 by storing the original file name as a property on `TemporaryUploadedFile`, but found that it required significant changes in multiple files and how data was structured, so decided to stop on that.

### Implementation details
With this PR, when a temporary file is uploaded, a corresponding meta file will be generated.

For example, when a file get's uploaded the `livewire-tmp` directory will now have two files in it:
- `L7obxpYUiNJmegxfKCkLPxEOWtGgZO1ZwXQgQeVS.png`
- `L7obxpYUiNJmegxfKCkLPxEOWtGgZO1ZwXQgQeVS.png.json`

The first file is the actual file. Where as the second (ending in `.json`) is the meta file.

The meta file contains the following information:
- name - original client name
- type - mimetype
- size - size of file
- hash - Laravel's hash of the original file

I have extracted the logic for creating a temporary file and it's meta file into a method on the `FileUploadConfiguration::storeTemporaryFile();`. I did this as it already has information like storage and disk, which simplifies the implementation. Ideally I would want it on the `TemporaryUploadedFile` class as a `store()` method, but `store` and `storeAs` already exist.

One thing to note is that I experimented with using Laravel's hash (as stored in the meta file above), but it attempts to read the file's mimetype from the file contents and changes the file extension on the end of the hash to suit. This is different to how Livewire has been handling it, so have kept our hash name implementation and just simplified it to not include the file name in the hash.

### Testable
I was able to clean up the `Testable` upload configuration to use the meta file now too. Previously it would inject file size,  mime type, and Laravel's file hash into the file name which ran the risk of the same file name too long issue.

I have made sure though that there is backwards compatibility with any temporary files, that may exist that won't have the new meta file generated, by ensuring all meta information look ups fall back to using the filename if no meta file is found.

### S3 compatibility

I have not touched the S3 implementation, so it is still encoding the filename in the hash. I tested this PR with S3 to ensure that everything is still working as expected and it is.

The reason I opted to leave filename encoding in S3, is because it has no restriction on the filename as such, instead it has a restriction of 1024 characters for the key name of the object in the store (which is essentially the relative path + file name), which I believe would be sufficient.

I also found that uploading a meta file to S3 would require it to be reworked as it would need to be uploaded in a separate API call, either when the signed URL is being generated, or as a second sideload (which would require a second signed URL).

Because of that, I have opted to leave it as is for now. I have added a test which ensures Livewire will happily read the original filename from the file hash if no meta file exists.

### DIFF
--- a/src/Features/SupportFileUploads/FileUploadConfiguration.php
+++ b/src/Features/SupportFileUploads/FileUploadConfiguration.php
@@ -124,4 +124,21 @@ class FileUploadConfiguration
     {
         return config('livewire.temporary_file_upload.max_upload_time') ?: 5;
     }
+
+    public static function storeTemporaryFile($file, $disk)
+    {
+        $filename = TemporaryUploadedFile::generateHashName($file);
+        $metaFilename = $filename . '.json';
+        
+        Storage::disk($disk)->put('/'.static::path($metaFilename), json_encode([
+            'name' => $file->getClientOriginalName(),
+            'type' => $file->getMimeType(),
+            'size' => $file->getSize(),
+            'hash' => $file->hashName(),
+        ]));
+
+        return $file->storeAs('/'.static::path(), $filename, [
+            'disk' => $disk
+        ]);
+    }
 }
--- a/src/Features/SupportFileUploads/FileUploadController.php
+++ b/src/Features/SupportFileUploads/FileUploadController.php
@@ -42,11 +42,7 @@ class FileUploadController implements HasMiddleware
         ])->validate();
 
         $fileHashPaths = collect($files)->map(function ($file) use ($disk) {
-            $filename = TemporaryUploadedFile::generateHashNameWithOriginalNameEmbedded($file);
-
-            return $file->storeAs('/'.FileUploadConfiguration::path(), $filename, [
-                'disk' => $disk
-            ]);
+            return FileUploadConfiguration::storeTemporaryFile($file, $disk);
         });
 
         // Strip out the temporary upload directory from the paths.
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -13,6 +13,7 @@ class TemporaryUploadedFile extends UploadedFile
     protected $disk;
     protected $storage;
     protected $path;
+    protected $metaFileData;
 
     public function __construct($path, $disk)
     {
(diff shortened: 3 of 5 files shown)


Naoray:
Thank you @joshhanley for tackling this issue 🙏

jhjvandenbroek:
Thanks for this @joshhanley (and others), but it seems there hasn't been a release with it yet. What's the ETA on that? @calebporzio

Naoray:
@jhjvandenbroek i think it's planned for the v4 release

vishal2931:
Hi @joshhanley 
Is there any quick fix for this in Livewire 3? Since this fix is coming in v4, but need a workaround for now. Any suggestions?
@calebporzio

JapSeyz:
We are in the same boat as well. It feels funky that images fail to upload based on the length of the file from the customers computer.

I would definitely appreciate a fix in Livewire 3, instead of having to wait to Livewire 4.

vishal2931:
@JapSeyz 
I found a quick fix for Livewire 3 until Livewire 4 is released. I’ve overridden Livewire’s `FileUploadController` `validateAndStore` method with custom file name logic. I know it’s not the best solution, but for now, this is the only workaround I’ve found.

JapSeyz:
Cool. How did you override it? I don't see it referenced in the config file.

Composer autoloading override or am I missing something?

ghabriel25:
> Cool. How did you override it? I don't see it referenced in the config file.
> 
> Composer autoloading override or am I missing something?

I think he is overriding route level

https://github.com/livewire/livewire/blob/main/src%2FFeatures%2FSupportFileUploads%2FSupportFileUploads.php#L33

vishal2931:
@JapSeyz @ghabriel25 
Yes, I'm overriding at Route level.

PR #9361, opened 2025-06-02
Consistency in naming the config tag

majidalaeinia:
Due to consistency on naming among similar Laravel related packages.

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -33,7 +33,7 @@ class LivewireServiceProvider extends \Illuminate\Support\ServiceProvider
     {
         $config = __DIR__.'/../config/livewire.php';
 
-        $this->publishes([$config => base_path('config/livewire.php')], ['livewire', 'livewire:config']);
+        $this->publishes([$config => base_path('config/livewire.php')], ['livewire', 'livewire-config']);
 
         $this->mergeConfigFrom($config, 'livewire');
     }


lrljoe:
Given the user base reach, I'd recommend a slow migration, rather than what could be seen by some as a breaking change.

By all means add the new approach, and encourage it via the docs, but I imagine many devs would be upset if suddenly their existing build processes broke because the existing approach was abandoned.

I'd suggest just **adding** the extra option for now, and adding in a reference to the relevant docs to state that the old syntax will become deprecated.
Then in a few versions, the older version can be removed.

majidalaeinia:
@lrljoe That's actually a valid consideration, but I prefer to let the maintainers handle that.

lrljoe:
> @lrljoe That's actually a valid consideration, but I prefer to let the maintainers handle that.

Then please make your PR fully backwards compatible.
I'd strongly recommend that you 
add sufficient docs explaining the change, 
and
add relevant tests.

Otherwise your PR basically changes the publish approach, and nobody is any the wiser, and you're asking an already busy core team to spend time writing tests and docs to satisfy your desired approach.

While I acknowledge that there's some reasonable standard to publish commands etc, there's no real meaningful standard across the spread of widely used packages.

My steer for your PR remains to update the docs, introduce the new method, and leave it to the core team to retire the old method as they see fit.

Keep in mind how many "intro to" tutorials removing it would break, which would be reputationally damaging, and demotivating to devs.  Removing any core install methods etc needs to be considered.  Adding a new one is much smoother.

majidalaeinia:
@lrljoe I saw your first message, I think you need to reconsider what you're typing before sending it, and welcome PRs in a more friendly manner.

@calebporzio

lrljoe:
Apologies if it came across poorly.

I'm not part of the core team to be clear.

I do however review PRs, and I try to consider all levels of users, and the widespread nature of the package which means that removing core "build" elements needs to be considered heavily.

Your PR just needs to address change and update the docs, and maintain historic deployment to mitigate the risk of breaking anyone's build scripts overnight.

majidalaeinia:
@lrljoe Completely valid statements. You're right.

lrljoe:
> @lrljoe Completely valid statements. You're right.

Please do make my recommended adjustments (or maybe wait for Josh/Caleb/Philo to review) and reopen the PR, there's definitely validity to it, it just needs that longer term/backwards support imho.  But they may disagree!

My view is avoiding anything that results in "learned behaviour"/historic scripts breaking.  But the guys may disagree!

PR #9364, opened 2025-06-03
Dispatch JavaScript event on each stream chunk

AndrewMonty:
# Background

Building a Chat-GPT like LLM interface with Livewire, using `wire:stream` to stream the response as it's generated.

# Problem

As the response comes in, I want to scroll the page down so the user can read along without having to keep manually scrolling. I could not find any way to hook in and execute some custom JavaScript.

# Solution

Tried adding a `wire:stream.scroll` modifier, felt clean at first, but seemed to rigid. I wanted more flexibility, like choosing which element to scroll, scroll behavior, canceling on scroll up, etc. without having to add a ton of modifiers.

Landed on just dispatching an event in the stream handler, then I can listen with Alpine and do whatever I want.

Like call a JavaScript Action:

```html
<div wire:stream="target" x-on:streamed="$wire.$js.scrollToBottom()"></div>
```

Call a method on a parent container element:

```html
<div @streamed="$el.scrollIntoView({behavior: 'smooth'})">
  <p wire:stream="answer"></p>
</div>
```

You get it.

https://github.com/user-attachments/assets/2730cccd-9746-459c-9a9d-050c3dc0f3b4


lrljoe:
Given a typical use case, why dispatch the content with the event, simply dispatching a streamed event on the element would be smoother.

I believe you could just capture the event content using existing logic tho,  unless I'm mistaken completely, there's a trigger firing for stream?

AndrewMonty:
> Given a typical use case, why dispatch the content with the event, simply dispatching a streamed event on the element would be smoother.

I figured just dispatch everything that's available. Maybe in the count-down streaming example you want to add a class when the number gets under 10: `Number(event.detail.content) < 10 && $el.classList.add('animate-pulse')`

> I believe you could just capture the event content using existing logic tho, unless I'm mistaken completely, there's a trigger firing for stream?

You're right, I just couldn't figure out how to listen for it.

I thought I tried every combo of `$wire.on('stream')`, `Livewire.on('stream')`, but it's a hook I needed: https://livewire.laravel.com/docs/javascript#javascript-hooks

It's not documented there, but this works! I'll have to do a little extra work to find the wire:stream element, but that's fine.

```js
    Livewire.hook('stream', ({name, content, replace}) => {
        //
    })
```

PR #9367, opened 2025-06-04
Toggle progressbar via configuration

utsavsomaiya:
### Why

There are three scenarios that need to be covered:

1. **Livewire automatically injecting the assets**  
   In this scenario, when the user sets `livewire.navigate.show_progress_bar` to `false`, a warning appears saying `Alpine is not defined`.

2. **Manually injecting the Livewire assets**

3. **Injecting the assets using Vite or Mix**  
   In the second and third scenarios, even if the user sets the progress bar to be disabled, it still appears. So the progress bar is not being properly hidden.

### Implementation

- As described in the documentation, if you want to tweak the features, you need to go into the respective folder and make the necessary changes. I encountered the first error while doing this.
- I looked into the `FrontendAssets` mechanism file and noticed that when the progress bar is disabled, it sets `data-no-progress-bar` in `window.livewireScriptConfig`.
- However, this doesn't match correctly in the frontend `supportNavigate.js` file. So, I made a tweak there and also used `Alpine.nextTick`.
- The reason for using `Alpine.nextTick` is that calling `Alpine.navigate` directly throws an `undefined` error. Therefore, I needed to hide the progress bar after the DOM has been rendered.

Maybe I'm wrong, but I’d love to hear your thoughts. Thanks!

Here is the minimal repo with the composer link: [https://github.com/utsavsomaiya/demo](https://github.com/utsavsomaiya/demo)


lrljoe:
Please don't try to commit changes to dist, as these are compiled scripts using the build method, and confuse any PR.

lrljoe:
#9379 addresses this

utsav-infynno:
@joshhanley Should I need to close this PR?

joshhanley:
@utsavsomaiya or @utsav-infynno (not sure which to tag haha) sorry, I missed this PR. Thanks for the PR! I'm going to leave it open for now and make sure I've got everything covered in mine. Then I will close this later. Thanks again for your effort on this!

utsav-infynno:
Okay I just rebased my both branch with main! Now you can check that out.

joshhanley:
@utsav-infynno thanks! I was missing the fix for the `@livewireScriptConfig` so I've added that to my PR too, thanks for that! I'm going to close this now.

PR #9376, opened 2025-06-10
Add rememberForever and flexible functionality to computed properties

schonhoff:
Introduce functionality to cache computed properties with flexible expiration and "forever" options. Corresponding tests demonstrate caching behavior using `seconds` arrays and the new `forever` attribute.

No, I don't create a discussion. In our project some DB calls will be the same in different components. Instead of writing:
```php
#[Computed]
public function foo(): void
{
    return Cache::rememberForever($key, fn () => 'extrem long db call');
}
```
every time we could just leverage the computed property attribute.

Yes. add-remember-forever-to-computed-properties

Yes and no. It includes the possibility to add flexible and rememberForever Cache functions with only modifying the attribute.

Yes, but I don't have a good way to test flexible and rememberForever in a Livewire way. I could just wait x seconds but that would slow down the test suite. Maybe someone has a better solution for tests. I'm willing to add them.

See point 1 for the problem. Here is the solution:

**Flexible**
Instead of providing an integer in the seconds parameter you can now provide an array with exactly two elements.
```php
#[Computed(seconds: [1, 2], cache: true)]
function foo() {
    return 'bar';
}
```

**Remember forever**
For the forever method we need a toggle. If te toggle is set it will discard the seconds parameter and only use the forever cache method.
```php
#[Computed(cache: true, forever: true)]
function foo() {
    return 'bar';
}
```

Thanks for contributing! 🙌

Thanks for making Livewire! <3


schonhoff:
I will mark this now as ready for review. Feedback is welcome and if somebody has an idea how to test the non existing flexible method or how to test the cache, you are very welcome.

schonhoff:
closing because Laravel 10/11 is supported at the moment.

PR #9377, opened 2025-06-10
`wire:target` failing when a string argument has parenthesis

AlexandreBonaventure:
This PR highlights a regression introduced by: https://github.com/livewire/livewire/pull/8940 (on main but not released yet)
with a failing test. 

It seems like the Regexp used is not greedy enough and will fail to parse correctly valid string arguments containing parenthesis (maybe also potentially `,` comma character) 

Capture d’écran, le 2025-06-10 à 15 00 14


joshhanley:
@AlexandreBonaventure ah was this the PR that you were meant to link the in the other comment? 😁 Thanks for the failing test, that's great!

AlexandreBonaventure:
> ah was this the PR that you were meant to link the in the other comment?

Yes, correct! sorry my mistake :)

joshhanley:
Going to close this as it has been fixed in https://github.com/livewire/livewire/pull/9513

PR #9378, opened 2025-06-11
Fix navigate loading conflict

joshhanley:
# The scenario

Currently if you have `wire:loading` and `wire:navigate` on the same element, the `wire:loading` isn't working at all.

```blade
<a href="/" wire:navigate class="text-blue-500" wire:loading.class="text-red-500">Link</a>
```

# The problem

The issue is that the `wire:loading` is using the `wire:navigate` value, which is an empty string, as the target, when it shouldn't have a target all.

The reason for this is because `wire:navigate` is not being filtered out as an non action/model directive in the automatic scoping, as it's not included in the list (see below).

```js
let nonActionOrModelLivewireDirectives = [ 'init', 'dirty', 'offline', 'target', 'loading', 'poll', 'ignore', 'key', 'id' ]
```

# The solution

The solution is to add `'navigate'` to the above list to ensure `wire:loading` doesn't try to automatically scope any navigate directives.

```js
let nonActionOrModelLivewireDirectives = [ 'init', 'dirty', 'offline', 'navigate', 'target', 'loading', 'poll', 'ignore', 'key', 'id' ]
```

Fixes #9057


ghabriel25:
@joshhanley You're a lifesaver. Great work!

Before, I use lazy load placeholder as workaround

PR #9379, opened 2025-06-11
Fix navigate disable progress bar

joshhanley:
# The scenario

Currently if you try to disable the navigate progress bar by setting `livewire.navigate.show_progress_bar` to `false` in config, a "Uncaught ReferenceError: Alpine is not defined" is thrown.

image

# The problem

The issue is that inside `js/features/supportNavigate.js` we have the following line to disable the progress bar.

```js
shouldHideProgressBar() && Alpine.navigate.disableProgressBar()
```

But it is the `Alpine.navigate.disableProgressBar()` that is throwing the error.

This is happening because the `supportNavigate.js` feature file is being loaded before the Alpine window object is defined by Livewire. I tried changing this to import Alpine, but the feature is also loaded before any Alpine plugins are registered by Livewire, so instead we were getting an undefined `Alpine.navigate` error.

# The solution

The solution is to defer this check until after Livewire has initialised.

I have done this by adding a listener for `livewire:initialized`.

```js
document.addEventListener('livewire:initialized', () => {
    shouldHideProgressBar() && Alpine.navigate.disableProgressBar()
})
```

There was also an issue with the progress bar not being disabled when using `@livewireScriptConfig` as the value returned from the backend when it is disabled is `data-no-progress-bar` not `false`.

```php
$progressBar = config('livewire.navigate.show_progress_bar', true) ? '' : 'data-no-progress-bar';
```

So I've fixed that too.

Fixes #6548


calebporzio:
great, thanks, merge it.

PR #9380, opened 2025-06-11
Fix navigate failed fetch handling

joshhanley:
# The scenario

Currently if a navigate fetch request fails, Livewire stops working and the progress bar never finishes. A failed fetch could happen because the browser is offline, something went wrong with the server request, or if the server does an external redirect (see screen capture).

Recording 2025-06-11 at 16 12 54

# The problem

The issue is that there is no error handling for a failed fetch request and because of that there is no clean up for the progress bar, so it never actually resolves.

# The solution

This PR adds error handling by adding a `errorCallback` to our fetch and prefetch methods, which can accept a callback to clean up anything in the event of an error. I have then added a callback which calls `finishAndHideProgressBar()` to resolve and remove the progress bar.

```js
.catch(error => {
    errorCallback()

    throw error
})
```

Now everything works happily, the progress bar resolves and Livewire still continues to function.

Recording 2025-06-11 at 16 23 44

For the moment, I am just re-throwing the error after the error callback has been run, but I believe we should probably dispatch an error which developers can listen to, to show an error toast or modal or something, to give users feedback that something has happened.

The question is what should that method be called?

For comparison, we currently have the following events:
- `wire:navigate`
- `wire:navigating`
- `wire:navigated`

But there is a case for adding more navigate methods (see PR #9319), so it feels like maybe we should change them to have a `navigate` prefix of some form, maybe like `wire:navigate:error` and the existing ones can be renamed to `wire:navigate:start`, etc.

That's probably beyond this PR, so just going to leave it with throwing the error for now.

Fixes #6267
Fixes #6609


calebporzio:
great, fix tests, don't worry about the event stuff right now, we'll just go with this fix. thanks mate

PR #9383, opened 2025-06-12
Fix Prevent header modification error when streaming responses end

MrCrayon:
## The Problem
When using `wire:stream`, the streaming functionality works correctly, but the final response triggers a PHP warning and causes a JavaScript `SyntaxError` in the browser.

**Error:** `Cannot modify header information - headers already sent`

This occurs because Livewire sends streaming content, and when the stream ends, it attempts to send a JSON response through Laravel's normal response pipeline. However, headers have already been sent during streaming, causing PHP to throw a warning that was previously suppressed by Symphony but is now visible since v7.2.7.

Discussion #9357

## The Solution
Detect when headers have already been sent and handle the final response appropriately:

- If headers are already sent (streaming mode): Return a custom response that skips header modification
- If headers haven't been sent (normal mode): Proceed with standard Laravel response handling

The fix uses an anonymous class extending Laravel's Response class, overriding only the `sendHeaders()` method to prevent header modification attempts.

## Screenshot
Basic countdown fails without fix
Screenshot From 2025-06-11 14-30-07


src/Mechanisms/HandleRequests/HandleRequests.php
```diff
@@ -104,8 +105,21 @@ function handleUpdate()
...
 
-        $finish = trigger('response', $responsePayload);
+        $response = $finish($responsePayload);
+
+        /**
+         * Headers have already been sent by stream
+         * we need to send the output without additional headers
+         **/
+        if (headers_sent()) {
+            return new class(json_encode($response)) extends Response {
```
joshhanley, line 115: If we return this new response here, none of the actual `$response` get sent though right? I wonder if that will break anything for Livewire as no component state, etc will be send. 🤔

src/Mechanisms/HandleRequests/HandleRequests.php
```diff
@@ -104,8 +105,21 @@ function handleUpdate()
             'assets' => SupportScriptsAndAssets::getAssets(),
         ];
 
-        $finish = trigger('response', $responsePayload);
```
joshhanley, line 107: I think this `trigger()` is still needed, I suspect that is why the other tests are failing.

joshhanley (commented):
@MrCrayon thanks for the PR! Added some comments below.

src/Mechanisms/HandleRequests/HandleRequests.php
MrCrayon: Absolutely, I forgot a line and it was kind of working just because `$finish` was already defined earlier.
That fixes the other tests.

kauffinger:
Just tested with `"symfony/http-foundation": "^7.3.6"` & `"livewire/livewire": "^3.6.4"` and the issue seems to be fixed :)

MrCrayon:
@kauffinger livewire behaviour should still be fixed.

https://github.com/symfony/symfony/issues/60603#issuecomment-2996908237
> The change has been reverted in [#60875](https://github.com/symfony/symfony/pull/60875) but will be reintroduced in Symfony 8.0 after [#60876](https://github.com/symfony/symfony/pull/60876) which is deprecating the behavior. Please tell Livewire and others they should accommodate for the change before 8.0 is out.

kauffinger:
Ah I see, thanks for the heads-up!

calebporzio:

calebporzio:
@claude - implement all your suggested changes. thanks

calebporzio:
Thanks for this however I'm going to close as especially in Livewire 4 people will intuitively use the keba-based component name with dot notation anytime they reference a component name. thanks!

calebporzio:
Ok, thanks everyone for your work on this. I couldn't re-produce the issue because Symfony reverted it, but I re-enabled it in Symfony core to replicate and saw what you were seeing.

I just added a wrapper response class instead of the anonymous class and improved the JS handling of the errors like this in general, so instead of getting that JSON.parse() error, you'll see a modal with the full Laravel one.

This was a bugger, good work 👊

PR #9384, opened 2025-06-12
Fix `wire:current` after morph

joshhanley:
# The scenario

Currently if you are using `wire:current` on a link inside a component and it makes a request, the `wire:current` styles disappear.

Recording 2025-06-12 at 15 51 18

```blade
<?php

use Livewire\Volt\Component;

new class extends Component {
    //
}; ?>

<div class="flex flex-col items-start">
    <a href="/playground" class="underline" wire:current="font-bold">Playground</a>
    <a href="/playground2" class="underline" wire:current="font-bold">Playground 2</a>
    <a href="/playground3" class="underline" wire:current="font-bold">Playground 3</a>

    <button type="button" wire:click="$refresh" class="rounded px-3 py-2 hover:bg-gray-200">Refresh</button>
</div>
```

# The problem

The issue is that the styles added by `wire:current` are being wiped out when Livewire morphs.

# The solution

The solution is to re-apply `wire:current` styles after morph has finished.

Recording 2025-06-12 at 16 02 44

Fixes #9074


danie-ramdhani:
@joshhanley should it applied after `Alpine.nextTIck` for link with alpine bind scenario (i.e. `<a x-bind:href="window.location.pathname" wire:current="text-red-500" x-text="window.location.pathname"></a>`)? Or it is works without `nextTick`?

ghabriel25:
@danie-ramdhani I just check with `x-bind:href` and it doesnt even work for intial render

calebporzio:
Why doesn't this run in the clone tree and work? like things like x-text and such all run in clone tree and initialize themselves.

it's possible that we ignore all `wire:` in the clone phase - but surely `wire:show` and `wire:text` would have problems if that was true. Maybe I made an exception for certain directives at some point? Please investigate mate.

joshhanley:
@danie-ramdhani and @ghabriel25 yeah that is a separate issue. This issue is specifically `wire:current` after a Livewire refresh.

joshhanley:
@calebporzio I investigated that and I found that when `wire:current` was added, you added a new concept of global directive. These are being initialised in `Livewire.start()` during `Alpine.interceptInit()` but it has a `Alpine.skipDuringClone()` inside of it.

The reason why `wire:show` and `wire:text` currently work is because under the hood, they map to Alpine's `x-show` and `x-text` which are always initialised even during clone. Each of these feature's have their own `Alpine.interceptInit()` inside of them, which always runs.

So I've added a fallback callback to `Alpine.skipDuringClone()` which allows the global directives be initialised during clone as well.

It seems to be working great. The disadvantage to this, is if any global directives ever add any event listeners, then they should be skipped during clone, but this implementation will prevent that.

Another alternative would be to change `wire:current` to use `Alpine.interceptInit()` inside of the feature file, similar to the other features, so then we don't need the fallback callback above. But then we would no longer need the global directive feature either.

Another other alternative would be to actually make an Alpine directive `x-current` and then update `wire:current` to map to it like the other features do.

Let me know what you think and I can update to suit 🙂

PR #9392, opened 2025-06-14
Class based computed handlers and dynamic computed keys

PhiloNL:
This PR adds the option to use class-based computed handlers and continues work on an existing PR to allow dynamic computed keys ([https://github.com/livewire/livewire/pull/8784](https://github.com/livewire/livewire/pull/8784)).

### Class-Based Handlers for Computed Properties

In some situations, you may want more control over the cache/persist behavior of computed properties. For example, you might want to use a custom cache key based on the currently authenticated user and adjust the cache duration based on specific conditions.

This PR allows you to pass a class to both the `cache` and `persist` attributes of computed properties.

```php
#[Computed(cache: ChangeLogCache::class)]
public function changeLogs()
{
    return auth()->user()->changelog()->latest()->take(8)->get(['title', 'created_at']);
}

#[Computed(persist: ChangeLogCache::class)]
public function changeLogs()
{
    // ...
}
```

The class must implement the `generateKey` method, where you can define your custom cache key.

```php
use Livewire\Features\SupportComputed\ComputedHandler;

class ChangeLogCache extends ComputedHandler
{
    public $seconds = 20;
    public $tags = 'example';

    protected function generateKey()
    {
        return 'changelog.' . auth()->id();
    }
}
```

This is particularly useful for caching component data across multiple requests and pages that rely on the same cached data—while giving you more control. For example, you can cache based on the current authenticated user, whereas the current implementation would serve the same cache to all users.

If you want, you can still set the `seconds` or `tags` inline as you normally would—the class-based properties will take precedence:

```php
#[Computed(cache: ChangeLogCache::class, seconds: 20, tags: 'example')]
public function changeLogs()
{
    // ...
}
```

### Dynamic Computed Keys

In addition, this PR includes support for dynamic computed keys, as introduced in [https://github.com/livewire/livewire/pull/8784](https://github.com/livewire/livewire/pull/8784):

```php
public Page $page;

#[Computed(cache: true, key: 'page-components-{page.id}')]
public function pageComponents(): array
{
    // Build something dynamically here
    return [..];
}
```

This PR does not introduce any breaking changes and is covered by existing tests. I’ll add extra tests for the class-based approach if we decide to move forward with this feature 😄

### DIFF
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -83,46 +83,46 @@ class BaseComputed extends Attribute
 
     protected function handlePersistedGet()
     {
-        if($this->persist === true) {
-            return (new DefaultPersistHandler($this))->handleGet();
+        $handlerClass = DefaultPersistHandler::class;
+
+        if (is_string($this->persist) && class_exists($this->persist)) {
+            $handlerClass = $this->persist;
         }
 
-        if(is_string($this->persist) && class_exists($this->persist)) {
-            return (new $this->persist($this))->handleGet();
-        }
+        return (new $handlerClass($this))->handleGet();
     }
 
     protected function handleCachedGet()
     {
-        if($this->cache === true) {
-            return (new DefaultCacheHandler($this))->handleGet();
+        $handlerClass = DefaultCacheHandler::class;
+
+        if (is_string($this->cache) && class_exists($this->cache)) {
+            $handlerClass = $this->cache;
         }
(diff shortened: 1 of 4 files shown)


lrljoe:
Oh now this is very shiny!  Nice job!

frcofrancisco444-blip (approved):

calebporzio:

calebporzio:
Thanks for this @PhiloNL, going to pass on it for now.

PR #9398, opened 2025-06-20
Fix sfc compiler clean up

joshhanley:
wip


calebporzio:
good with me! (probably? does this prevent route caching or whatever? idk)

joshhanley:
Just double checked and route caching is all good!

PR #9399, opened 2025-06-20
Fix morph markers generation on Octane

AlexandreBonaventure:
@joshhanley I think a bug was introduced as part of this MR https://github.com/livewire/livewire/pull/9310

When testing a few things on our app with some unreleased changes (see https://github.com/livewire/livewire/pull/9302#issuecomment-2963618427) , I realized morph markers were not generated anymore. I'm not sure why a flush-state listener was setup here because it is config driven, I can only assume that it was only for testing purposes, but during testing you are using `reloadFeatures` which effectively re-register hooks and reload config, which does not happen in a real-life situation with an Octane app. So what is happening really is that both `shouldInjectConditionalMarkers` and `shouldInjectLoopMarkers` are immediately set to `false` when using Octane, regardless of the config values.

### DIFF
--- a/src/Features/SupportMorphAwareBladeCompilation/SupportMorphAwareBladeCompilation.php
+++ b/src/Features/SupportMorphAwareBladeCompilation/SupportMorphAwareBladeCompilation.php
@@ -17,8 +17,8 @@ class SupportMorphAwareBladeCompilation extends ComponentHook
     public static function provide()
     {
         on('flush-state', function () {
-            static::$shouldInjectConditionalMarkers = false;
-            static::$shouldInjectLoopMarkers = false;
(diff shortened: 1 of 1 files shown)


AlexandreBonaventure:
not exactly sure it is the proper fix, because again not exactly sure why there's a flush state listener for other reasons than just testing. Maybe it is intentional to disable smart keys ? but in any case, disabling morph markers is unexpected

joshhanley:
@AlexandreBonaventure sorry for the delay. The flush state gets used to reset Octane which you can see here https://github.com/laravel/octane/blob/4ca38b90d76f31b8c1e27873316c2db34450151c/src/Listeners/PrepareLivewireForNextOperation.php#L20-L23

I believe that you are correct that the flush state should reset it back to the config values, not false. So I'm going to merge this.

PR #9407, opened 2025-06-24
#[Computed] Computed property dependency injection

luanfreitasdev:
https://github.com/livewire/livewire/pull/7841

This PR introduces the withContainer method, enabling objects to be accessed with both the container and their parameters when needed.

The `__call` method was intentionally left out of the base implementation to avoid unnecessary overhead. It remains scoped to the Computed class, where dynamic method resolution is required.

---

```php

    #[Computed]
    public function foo(FooDependency $dependency) {
        return $dependency->baz;
    }
})
    ->assertSetStrict('foo', 'bar');

class FooDependency {
    public $baz = 'bar';
}
```


joshhanley:
@luanfreitasdev thanks for the PR! There is another PR that addresses this issue, that Caleb has commented on. See his comment here https://github.com/livewire/livewire/pull/8955#issuecomment-2429978204. Can you make sure dependency injection will work with protected computed properties too as requested by Caleb? The original PR hasn't been updated, so happy to leave this one open. Thanks!

luanfreitasdev:
Added @joshhanley! 

https://github.com/livewire/livewire/pull/9407/commits/f180361a1cb5817887c0e42d3d9c73be7b69fbb6

joshhanley:
@luanfreitasdev wow quick work, awesome job! I'll try and review this fully tomorrow. Will be great to have it!

ghabriel25:
@luanfreitasdev does it work with
```php
#[Computed]
public function posts(#[CurrentUser] User $user)
{
    return $user->posts;
}
```

calebporzio:
This is actually something I would want however I've come to learn the performance impact of dependency injection and I'm a little concerned that if we add this, all computed properties would add significant overhead to big pages with lots of components. so I'm going to pass now until we gauge further demand or someone reports back about performance impact. thanks anyways!

PR #9412, opened 2025-06-27
Alpine Livewire sync property without assignig back to livewire?

utsavsomaiya:
While building a feature in my Livewire app, I noticed something surprising.

My Livewire form property (`$options`) was being updated, **even though I never explicitly pushed any data from Alpine back to Livewire**.

---

### Livewire Component

```php
<?php

namespace App\Livewire;

use Livewire\Component;

class Demo extends Component
{
    public array $options = [];

    public function render()
    {
        return view('livewire.demo');
    }
}
```

### Blade View (`livewire.demo`)
```blade
<div
    x-data="demo"
    class="max-w-xl mx-auto p-6 bg-white rounded-lg shadow-md mt-10"
>
    <div>@json($options)</div>

    <pre x-text="JSON.stringify(options, null, 2)"></pre>

    <div class="space-y-4 my-4">
        <template x-for="(option, index) in options" :key="index">
            <div class="flex items-center space-x-2">
                <input type="text" x-model="option.text" placeholder="Text" class="border rounded px-2 py-1" />
                <input type="text" x-model="option.explaination" placeholder="Explanation" class="border rounded px-2 py-1" />
                <input type="checkbox" x-model="option.is_correct" />
                <button type="button" @click="options.splice(index, 1)" class="text-red-500">Remove</button>
            </div>
        </template>
    </div>

    <div class="flex space-x-2">
        <button type="button" @click="addOption" class="bg-blue-500 text-white px-3 py-1 rounded">Add Option</button>
        <button type="button" @click="removeOption" class="bg-gray-500 text-white px-3 py-1 rounded">Remove Last</button>
        <button type="button" @click="$wire.$refresh" class="bg-green-500 text-white px-3 py-1 rounded">Save</button>
    </div>
</div>

@script
    <script>
        Alpine.data('demo', () => {
            return {
                options: [],
                init() {
                    // Sync Alpine with Livewire
                    this.options = this.$wire.options
                },
                addOption() {
                    this.options.push({
                        text: '',
                        explaination: '',
                        is_correct: false,
                    })
                },
                removeOption() {
                    this.options.pop()
                }
            }
        })
    </script>
@endscript
```
### What Surprised Me?
Even though I only updated the Alpine options array, and never manually assigned it back to Livewire, calling:`$wire.$refresh
` was enough to reflect all the changes back in Livewire.

### Why Does This Happen?
When you assign Alpine’s options from `this.$wire.options`, you are assigning a reference.

So any mutation to options in Alpine directly mutates the Livewire form property.

When you call `$wire.$refresh`, Livewire rehydrates the component and the mutated data is already there.

### Be Careful
- This behavior is powerful, but you should be cautious:
- You're mutating Livewire's state implicitly via Alpine.
- If you're expecting a clean or fresh state, it won’t be unless you clone it before using.
- This can lead to unexpected side effects if not handled intentionally.


lrljoe:
Javascript (at least historically) doesn't create a new instance or a property for an array, instead it creates a pointer in memory from the newProperty to the oldProperty, thus they sync up.

Whereas for a string/int, when creating a newProperty from an oldProperty, it creates a new property in memory set to the value of the oldProperty at the time of execution.


While I am very happy to be corrected (Livewire may natively be addressing it), or the core behaviour may have changed, I don't think this is livewire/alpine specific at a glance.  Worth testing with some bog standard js to validate.

Does the same behaviour occur when using a string property with Alpine? 

If you want to actually create a new instance of an array then you'd need to

```
newArray = oldArray.slice();
```

This would actually create a new array in memory, rather than a pointer.


Yep still the case according to mozilla references:
"JavaScript array-copy operations create shallow copies. (All standard built-in copy operations with any JavaScript objects create shallow copies, rather than deep copies)."

utsavsomaiya:
> Javascript (at least historically) doesn't create a new instance or a property for an array, instead it creates a pointer in memory from the newProperty to the oldProperty, thus they sync up.
> 
> Whereas for a string/int, when creating a newProperty from an oldProperty, it creates a new property in memory set to the value of the oldProperty at the time of execution.
> 
> While I am very happy to be corrected (Livewire may natively be addressing it), or the core behaviour may have changed, I don't think this is livewire/alpine specific at a glance. Worth testing with some bog standard js to validate.
> 
> Does the same behaviour occur when using a string property with Alpine?
> 
> If you want to actually create a new instance of an array then you'd need to
> 
> ```
> newArray = oldArray.slice();
> ```
> 
> This would actually create a new array in memory, rather than a pointer.
> 
> Yep still the case according to mozilla references: "JavaScript array-copy operations create shallow copies. (All standard built-in copy operations with any JavaScript objects create shallow copies, rather than deep copies)."

May be we need to try something other as well.

```js
this.options = JSON.parse(JSON.stringify(this.$wire.options))
```

I am thinking about this but should we do in the livewire by default to not sync that variable?

utsavsomaiya:
We’ve introduced a new magic property called `$clone`, which uses structuredClone under the hood. We’ve also written a test case for it. This will help with deep cloning of arrays or objects.

Check out the [demo repository](https://github.com/utsavsomaiya/demo)—inside the `Demo1` component, we’ve used `$clone` in action.

Thanks!

joshhanley:
@utsavsomaiya thanks for the PR! This is just how javascript works as you've added a reference to the `$wire` object. At this stage this isn't a feature we want to add to Livewire, so I'm going to close it. I would recommend creating a clone helper method in your app if you find yourself frequently wanting a copy of a Livewire value. Thanks for being willing to contribute.

PR #9435, opened 2025-07-28
[4.x] Single File Components now preserve @php blocks

luanfreitasdev:
### Problem

When compiling a Livewire component that starts with a @php block containing a class definition (e.g., an anonymous Livewire component), the compiler correctly extracts that block as the component definition.
However, if there are additional @php blocks inside the HTML (e.g., for local variables within loops or conditions), those blocks are unintentionally removed.

Example input:

```php
@php
    //
}
@endphp

<div>
    @php
        $i = 10;
    @endphp

    <p>{{ $i }}</p>
</div>
```

**Current behavior** (❌):
Only the first @php block is processed, and the rest are stripped from the view.

**Expected behavior** (✅):
Only the first @php block (containing the class) should be removed and extracted as component logic. All other @php ... @endphp blocks inside the HTML must be preserved in the compiled view.

### DIFF
--- a/src/V4/Compiler/SingleFileComponentCompilerUnitTest.php
+++ b/src/V4/Compiler/SingleFileComponentCompilerUnitTest.php
@@ -225,23 +225,6 @@ new class extends Livewire\Component {
         $this->compiler->compile($viewPath);
     }
 
-    public function test_throws_exception_for_invalid_php_block_content()
-    {
-        $componentContent = '@php
-echo "This is not a class";
-@endphp
-
-<div>Invalid content</div>';
-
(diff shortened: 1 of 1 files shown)


joshhanley:
@luanfreitasdev thanks for the PR! We're still in the process of working on v4, so I'm going to close this. Please resubmit if we haven't fixed it by the time we have tagged the beta release. Thanks!

luanfreitasdev:
I was sending the fixes!

joshhanley:
@luanfreitasdev thanks but at this stage we want to focus fixing things ourselves. Hope you can understand.

PR #9443, opened 2025-08-01
Replace isset checks with null coalescing in findDirective

xurshudyan:


joshhanley:
@xurshudyan thanks for the PR! I'm going to close this as the file has changed since then. If it still has problems, feel free to submit a new PR. Thanks again!

PR #9462, opened 2025-09-04
PersistantMiddleware returns 404 on unresolved route

blazery:
**Background**
I was working with Statamic CMS, which has a catch-all route to handle dynamic slugs for pages. For some reason, unrelated to Livewire, the catchall route wasn't being loaded on requests targeting /livewire/update. Due to this cms pages that included livewire widgets stopped responding to updates, and reported with a HTTP-404 instead. 

Debugging why this only happened on the cms pages took us a while. 

**Issue**
Persistent middleware attempts to find the page the Livewire component is being used on using the path property in the component's snapshot. It will then apply middleware for that route to the update request. If the route could not be found and null was returned, an empty list of middleware should be used instead (as indicated by this existing code in Persistantmiddleware).
``` php
{
 if (! $route) return [];
  // more code
}
```

While matching the route and request, a NotFoundHttpException is thrown instead of returning null. This bubbles up to Laravel's exception handling, and a 404 is returned instead of the Livewire JSON one might expect.

**Solution** 
Catch the exception and return null instead. 
This allows dynamic routes provided by CMS systems to work with Livewire components.


calebporzio:

calebporzio:
@claude - fix the spacing issue please. i think I'm fine with the rest of the PR. thanks

joshhanley:
@blazery thanks for the PR! I'm going to close this in favour of the v3 version PR #9464 and then will upmerge the fix into v4.

PR #9464, opened 2025-09-09
Add better PersistantMiddleware unresolved route handling

blazery:
This PR is the v3 version of https://github.com/livewire/livewire/pull/9462.
The PR to main is failing on the Browser test for V4 features. This PR hopefully doesn't have this issue.

**Background**
I was working with Statamic CMS, which has a catch-all route to handle dynamic slugs for pages. For some reason, unrelated to Livewire, the catchall route wasn't being loaded on requests targeting /livewire/update. Due to this cms pages that included livewire widgets stopped responding to updates, and reported with a HTTP-404 instead.

Debugging why this only happened on the cms pages took us a while.

**Issue**
Persistent middleware attempts to find the page the Livewire component is being used on using the path property in the component's snapshot. It will then apply middleware for that route to the update request. If the route could not be found and null was returned, an empty list of middleware should be used instead (as indicated by this existing code in Persistantmiddleware).

```php
{
 if (! $route) return [];
  // more code
}
```
While matching the route and request, a NotFoundHttpException is thrown instead of returning null. This bubbles up to Laravel's exception handling, and a 404 is returned instead of the Livewire JSON one might expect.

**Solution**
Catch the exception and return null instead.
This allows dynamic routes provided by CMS systems to work with Livewire components.


calebporzio:
@claude - please review

src/Mechanisms/PersistentMiddleware/UnitTest.php
```diff
@@ -33,4 +35,48 @@ public function test_it_does_not_have_persistent_middleware_memory_leak_when_add
...
+    {
+        $component = Livewire::test(EmptyComponent::class);
+        $snapshot = json_encode($component->snapshot);
+
+        // Remove the livewire testing route
+        // This should cause the persistent middleware to fail resolving the route
+        $existingRoutes = app('router')->getRoutes();
+        $runningCollection = new RouteCollection;
+        foreach ($existingRoutes as $route) {
+            if(str_contains($route->uri, 'livewire-unit-test-endpoint')) {
```
joshhanley:
```suggestion
            if (str_contains($route->uri, 'livewire-unit-test-endpoint')) {
```

joshhanley:
@blazery thanks for the PR!

PR #9470, opened 2025-09-14
Add documentation about the js directive

ramonrietdijk:
I've added documentation about using the `@js` directive. I'm aware that the directive is part of Laravel but I think it's still quite valuable to have documented in the context of Livewire and Alpine.

Also note that this was documented in the previous version of Livewire:

* [Inline Scripts](https://laravel-livewire.com/docs/2.x/inline-scripts#using-js-directive)
* [AlpineJS](https://laravel-livewire.com/docs/2.x/alpine-js#js-directive)


calebporzio:
Thanks!

PR #9473, opened 2025-09-15
Detect CSP Nonce from Vite and inject automatically

valorin:
Following on from https://github.com/livewire/flux/issues/1951, the issue is Livewire automatically injects itself onto the page without honouring the CSP Nonce defined in Vite.

This PR allows Livewire to detect when there is a Nonce (`Vite::cspNonce()` returns `null` if no Nonce has been set), and injects it into the script and style tags. This should only inject a nonce when one hasn't already been provided manually via `$options`, and when it has been initiated by the user, making it backwards compatible.

### DIFF
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Mechanisms\FrontendAssets;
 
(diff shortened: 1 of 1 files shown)


calebporzio:
This is great, thanks @valorin - @joshhanley can we also add this to v4 or target v4 instead. thanks

joshhanley:
Yep will merge into v3 and upmerge to v4.

PR #9474, opened 2025-09-15
Fix morph inside custom web elements

joshhanley:
# The scenario

Currently, if we have a Livewire component that is inside of a custom web element, when that component makes a request, it breaks and throws an error.

Screenshot 2025-09-15 at 07 00 23PM

```blade
<ui-sidebar>
    <livewire:my-component />
</ui-sidebar>
```

# The problem

The issue is that when Livewire goes to morph, for the clone tree, it detects what the parent element of the component is, and creates a new element of the same time to hang the clone tree off.

But if the parent element is a custom web element, it can't be instantiated using the hyphenated name like `ui-sidebar`, instead it needs to be instantiated using the class name like `UISidebar`.

# The solution

The solution is to add a check to morph to ensure that if the parent element is a custom web element, that we get the name off the web element and use that for instantiation.

```js
let customElement = customElements.get(wrapperTag)

// If the wrapper tag is a custom element, we can't instantiate it using the hyphenated
// tag name, so we need to get the name off the custom element instead...
wrapperTag = customElement ? customElement.name : wrapperTag
```

Now everything works as expected.

Screenshot 2025-09-15 at 07 19 25PM

Fixes livewire/flux#1937


calebporzio:
wonderful work, makes perfect sense

PR #9479, opened 2025-09-22
Update installation.md

jochensengier:
As of Laravel v12.29.0, you also need to modify the name of the custom update route.

https://github.com/laravel/framework/issues/57085

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -67,9 +67,11 @@ In those cases, you can register your own endpoint however you like, and as long
 
 ```php
 Livewire::setUpdateRoute(function ($handle) {
-	return Route::post('/custom/livewire/update', $handle);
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the report. even if we merge this will it still be broken with the optimize command as referenced here? https://github.com/laravel/framework/issues/57085#issuecomment-3637383189

jochensengier:
I don’t use the optimize command myself, so I’m not able to confirm.

joshhanley:
@calebporzio I tested this out and the optimize command didn't break during my testing.

I was working on updating this PR to include a `[!warning]` highlight in the docs when it got me wondering whether we could just fix this instead of requiring users to set a custom name and managed to come up with a solution.

I've submitted a fix for this targeting v3 https://github.com/livewire/livewire/pull/9699.

calebporzio:
Great, thanks Josh. closing this then

PR #9508, opened 2025-10-20
FIX #9507 : Split classname@methods according to last @ as classnames…

ju-gow:
… for anonymous objects have a @

No, it's a fix

Yes, I did

Non, just one change

Passes all tests

See Issue 9507 :

```php
<?php

use App\Models\User;
use Livewire\Component;


// create an anonymous class that extends Component
$anonymous = new class extends Component {

    public User $user;

    public function render()
    {
        return <<<HTML
        <div>
            <h1>User: {{ \$this->user->name }}</h1>
        </div>
        HTML;
    }
};


// get the class name of the anonymous class
$method = get_class( $anonymous ).'@__invoke';  // => Livewire\Component@anonymous\/tmp/laravel/routes/web.php:9$a0@__invoke

// create a route that uses the anonymous class, parameters are know bounds to Authorize middleware
Route::get('/buggy/{user}', $method)->middleware('can:view,user');
```

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportPageComponents/SupportPageComponents.php
+++ b/src/Features/SupportPageComponents/SupportPageComponents.php
@@ -233,8 +233,8 @@ class SupportPageComponents extends ComponentHook
         if (! $uses) return;
 
         if (is_string($uses)) {
-            $class = str($uses)->before('@')->toString();
-            $method = str($uses)->after('@')->toString();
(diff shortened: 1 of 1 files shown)


ju-gow:
Doesn't work when routes are cached

PR #9513, opened 2025-10-21
[3.x] Fix wire loading multiple methods and params

joshhanley:
# The scenario

Previously `wire:target` only supported either a single method with params `wire:target="foo(1, 2)"` or multiple methods but without params `wire:target="foo, bar".

We recently added support for using  `wire:target` with multiple methods and parameters `wire:target="foo(1, 2), bar(3, 4)"` in  PR #8940. 

But there has been a report from @danharrin that during his testing of the v3.7.0-beta.1 in Filament, this multiple methods and params support is broken.

The scenario that is broken for Filament is when a `wire:target` is generated that makes use of the `@js()` or `Js::from()` helper methods.

This is an example `wire:target` generated by Filament.

```html
<div wire:loading wire:target="mountAction('add', {{ \Illuminate\Support\Js::from(['block' => 'name']) }}, { schemaComponent: 'tableFiltersForm.queryBuilder.rules' })">
```

# The problem

When the `Js::from()` helper renders it looks like this:

```html
<div wire:loading wire:target="mountAction('add', JSON.parse('{\u0022block\u0022:\u0022name\u0022}'), { schemaComponent: 'tableFiltersForm.queryBuilder.rules' })">
```

This outputs `JSON.parse()` which naturally contains parens `()`.

The problem is the current implementation of multiple methods and params handling is using a regex that is looking for an opening paren followed by a closing paren, and everything inside of that is the parameters for one method in the string. But it's incorrectly matching on the closing paren on the `JSON.parse()` call.

# The solution

The solution in this PR is to create a more robust directives methods/params parser that iterates thorough and keeps track of the open and close parens to ensure that the closing paren is only matched if it's not a nested paren.

This PR will also need to be up-merged into v4 which I will do after this is merged.

I've added a bunch of unit tests in a Vitest test file called `directives.spec.js` which covers a wide range of scenarios, to ensure the parser is thoroughly tested. Now, currently we don't have Vitest installed in v3, so these tests won't run, but it is installed in v4. We can add Vitest to this PR, or delete the test file and just keep it for v4, or just leave it and it will be up-merged and run properly in v4. I've run tests tests in v4, so I know they all pass.


calebporzio:
Probably gonna break something, but we'll cross that bridge when we come to it. thanks!

nuernbergerA:
> Probably gonna break something, but we'll cross that bridge when we come to it. thanks!

will tell you tomorrow

joshhanley:
It's perfect, so no worries!

PR #9515, opened 2025-10-24
Fix redirect interceptor

joshhanley:
# The scenario

Currently if you call a Livewire component method that has a redirect on it, and you have a request redirect interceptor set up to block the redirect, the redirect happens anyway.

```blade
<?php

use Livewire\Component;

new class extends Component {
    public function redirectHome() {
        return redirect('/home');
    }
};

?>

<div class="p-8 max-w-2xl mx-auto">
    <button wire:click="redirectHome">Redirect Home</button>
</div>

<script>
    Livewire.interceptRequest(({ onRedirect }) => {
        onRedirect(({ url, preventDefault }) => {
            preventDefault()
        })
    })
</script>
```

# The problem

The issue is the request redirect interceptor is being called if the network request returns a redirect response.

```js
if (response.redirected) {
    handlers.redirect(response.url)
}

// Handlers
redirect: (url) => {
    let preventDefault = false

    request.onRedirect({ url, preventDefault })

    if (preventDefault) return

    window.location.href = url
},
```

But in our scenario, the component itself is returning an internal redirect. So the network response is actually an ok response rather than a redirect.

Also the above code doesn't work for `preventDefault` as just the value is being passed through, instead of a closure to set `preventDefault` to `true`.

# The solution

Currently, component level redirects are being processed in an effects hook in `js/features/supportRedirects.js`.

So to be able to fire off the request redirect interceptor, I had to pass the request through to the effects, so `supportRedirects` could access it and then call the `onRedirect()` helper method. I'm also passing through a callback which sets `preventDefault` to `true`.

```js
on('effect', ({ effects, request }) => {
    if (! effects['redirect']) return

    let preventDefault = false

    request.onRedirect({ url: effects['redirect'], preventDefault: () => preventDefault = true })

    if (preventDefault) return
    
    ...
```

We have no need of it in this PR, but I wonder whether we should also pass the message through to effects along with request, as it seems strange to have request but not message?

There could also be an argument that the component level redirect should be handled by a message interceptor instead. What do you think? Would still have the same issue that the effects will need something passed through for them to hook into.

I also fixed up other instances of `preventDefault` being passed into methods as a value and converted them to callbacks instead.


calebporzio:
amazing thank you josh

PR #9518, opened 2025-10-28
Add a custom exception for when property assignment has a type error

godismyjudge95:
> 1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

Yes this wanted/needed because the default error is ambiguous.  No I did not create a discussion.

> 2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed)

Yes

> 3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

Nope

> 4️⃣ Does it include tests? (Required)

Not sure how to add a test for an exception?

> 5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

I've run into this error multiple times before and seen a good number of people posting in Discord/X/etc about ambiguous errors like this being hard to pin down:

image

<br><br>

But with this custom exception we get a more exact error that leads straight to the issue:

image

---

I would be happy to submit a test for this but I didn't see any existing ones for testing the other exceptions so I wasn't sure it was needed?

Also, I realize this is messing with the stack frames (changing files and line numbers) but I wasn't sure how else to do it.  If there is a better way, I'd be happy to help implement it.

### DIFF
--- a/src/Exceptions/PropertyAssignmentTypeException.php
+++ b/src/Exceptions/PropertyAssignmentTypeException.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Livewire\Exceptions;
+
+use Exception;
+use Livewire\Component;
+use Livewire\Volt\Component as VoltComponent;
+use ReflectionClass;
+
+class PropertyAssignmentTypeException extends Exception
+{
+    use BypassViewHandler;
+
+    public function __construct(Component $component, string $levelName, mixed $value)
+    {
+        $value_string = gettype($value);
+
+        $reflection = new ReflectionClass($component);
+        $this->file = $reflection->getFileName() ?? $this->file;
+
+        $className = $reflection->getShortName();
+        if ($component instanceof VoltComponent) {
+            $className = str($this->file)->after(resource_path())->ltrim(DIRECTORY_SEPARATOR)->value();
+        }
+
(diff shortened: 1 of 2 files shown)


calebporzio:
Hmmm yeah I can see the need for something like this. Besides the implementation one my hesitancies is that there is more information in the original thrown error: like what the type expected.

I'm leaning towards just keeping it as-is and not introducing a custom exception.

@joshhanley ?

godismyjudge95:
> Hmmm yeah I can see the need for something like this. Besides the implementation one my hesitancies is that there is more information in the original thrown error: like what the type expected.
> 
> I'm leaning towards just keeping it as-is and not introducing a custom exception.
> 
> @joshhanley ?

If you are leaning against a custom exception, maybe just a more detailed error message with a proper path and line number for the affected property?  I think in general stack traces are kinda rough with multiple errors like this not pointing to where the actual issue is.  If there can be a standardized way to address them I'd be more than happy to submit PR(s) for whatever approach you think is best.

joshhanley:
Thanks for the PR! We're going to close this for now as we're working on trying to get better error message handling in Livewire which should link directly to the correct line in Livewire components. Thanks for being willing to contribute!

PR #9519, opened 2025-10-28
Return unwatch to unwatch manually if needed

MaxencePaulin:

Yes, I have seen the PR #9097, but is not necessary for me to automatically clean up after the element is removed because i will do it with the destroy() function of Alpine.data.

Yes but not in official livewire repository.

No, only the return unwatch

Yes, i have tried to do my best for dusk test (first time with dusk, and on windows...)

Sometimes, an Alpine component can use $wire.$watch to observe Livewire properties and execute certain actions related to the Alpine component. However, when the Alpine component is removed (for example, due to a parent Blade condition), the watcher continues to persist, causing errors. I attempted to fix this by using an Alpine.data variable called destroyed, but this only prevents the code from executing—it does not stop the watcher from persisting.

What i do right now to dodge some errors and re-executing:
```javascript
Alpine.data('myComponent', () => ({
    destroyed: false,

    init() {
        this.$wire.$watch('someProperty', (value, old) => {
            if (destroyed) return;
            this.executeSomething();
        });
    },

    destroy() {
        this.destroyed = true;
    };

    executeSomething() {
        if (this.$wire.someProperty == 'something') {
            // ...
        }
    }
}))
```

What i want to do by exemple after my PR: 
```javascript
Alpine.data('myComponent', () => ({
    unwatch: null,

    init() {
        this.unwatch = this.$wire.$watch('someProperty', (value, old) => {
            this.executeSomething();
        });
    },

    destroy() {
        if (typeof this.unwatch === 'function') {
            this.unwatch();
        }
    };

    executeSomething() {
        if (this.$wire.someProperty == 'something') {
            // ...
        }
    }
}))
```

Thanks for contributing! 🙌


danie-ramdhani:
i think this related (this one is automatically unwatch when element is removed): https://github.com/livewire/livewire/pull/9097

MaxencePaulin:
Yes, but not only in this case. Sometimes it's useful to call unwatch at personalized times I guess, even if the one I put as an example is during a destroy

danie-ramdhani:
I don't think so, because in alpine.js itself it's automatically unwatch: https://github.com/alpinejs/alpine/blob/92a9fd0195b06651e573bcbb0931533bddeb2c5e/packages/alpinejs/src/magics/%24watch.js#L17

That's what I'm doing in that PR, but let's see what mr caleb think (manually let the users unwatch the itself when element removed or automatically unwatch like in alpine.js)

MaxencePaulin:
Yes I completely agree, I just say that this is not the only time when an unwatch may be necessary depending on the use.

danie-ramdhani:
> Yes I completely agree, I just say that this is not the only time when an unwatch may be necessary depending on the use.

I mean `Alpine.onElRemoved(this.$el, () => {
            ...
        })` is same thing with `destroy` function. What kinda use case u need for unwatch prop hooks?

MaxencePaulin:
There are cases where calling unwatch without necessarily destroying the Alpine component can be useful.

In an inline script, depending on a condition, for example.

Or when changing the visibility of an element without destroying it via an x-show.

What I mean by this is that calling unwatch manually when necessary is not unreasonable, as alpinejs itself returns unwatch so that it can be called at any time.

But I agree that automatically unwatching when the element is removed from the DOM is also a good thing. It's just that my request doesn't only cover this point, even though it's the example I used in my description. The tests I've done are not related to the destruction of an element, for example.

danie-ramdhani:
> In an inline script, depending on a condition, for example.

for example??

> Or when changing the visibility of an element without destroying it via an x-show.

just use condition, u don't need unwatch it manually...

```js
  this.$wire.$watch('someProperty', (value, old) => {
            if (condition) {
            // ...
            }
        });
```

> What I mean by this is that calling unwatch manually when necessary is not unreasonable, as alpinejs itself returns unwatch so that it can be called at any time.

did you read the code in alpine js? https://github.com/alpinejs/alpine/blob/92a9fd0195b06651e573bcbb0931533bddeb2c5e/packages/alpinejs/src/magics/%24watch.js
it return nothing and just clean it up itself

MaxencePaulin:
You true i was looking on watch function used in de $watch.

```javascript
Alpine.watch(getter, callback)
```

It's what i do right now, i conditionnally execute some code in the watch. But for an Observer, i think it's more clean and more performant to unwatch instead of conditionnally execute the code.

If you don't agree with that it's your choice. I just give my opinion that a watch Observer should be unwatch whenever we want. I understand your position on only unwatch when element is destroyed but like the hook and unhook, is not only the case to unwatch something.

Watch my test, not my real case but we could sometimes watch or unwatch not depending on element itself.

MaxencePaulin:
I started web development with frameworks such as Vuejs.

In this framework, we can manually unwatch.

I was just surprised to see that this wasn't the case with Livewire, but apparently it's not the case with AlpineJS's magics either. Should I start with a pull request on AlpineJS?

[https://vuejs.org/guide/essentials/watchers#stopping-a-watcher](https://vuejs.org/guide/essentials/watchers#stopping-a-watcher)

calebporzio:
This makes a lot of sense and I would expect the same.

The only weird edge case that comes to mind is:

By default, if an Alpine expression returns a callback I believe it will be run right away?

Like `x-on:click="() => { console.log('foo) }"`

So if this returns an unwatch callback, will the following code watch and then immediately unwatch:

`x-init="$wire.$watch('foo', ...)"`??

So if you could test that locally or something that would be helpful.

Otherwise, i think it's a good addition and will merge.

Btw, if you don't mind removing the dist files from the PR, we will build those before tagging and this makes it easier to get around merge conflicts.

Thanks!

calebporzio:
(if this makes the test fail because you didn't commit the build files, we should add a build to the CI script @joshhanley but i bet he already did that)

thanks!

MaxencePaulin:
> By default, if an Alpine expression returns a callback I believe it will be run right away?
> 
> Like `x-on:click="() => { console.log('foo) }"`
> 
> So if this returns an unwatch callback, will the following code watch and then immediately unwatch:
> 
> `x-init="$wire.$watch('foo', ...)"`??
> 
Unfortunately, you're right. If we do that:

```x-init="$wire.$watch('foo', ...)"```

It immediately unwatch. I have to do :
```
x-init="() => {
    $wire.$watch('someProperty', (value, old) => {
        $data.somePropertyWatched = value;
    });
}"
```
And actually, it's the same for the `$wire.$hook`. If we do that :

```x-init="$wire.$hook('commit', ...)"```

It doesn't work, because magic $hook already return a callback (unhook).

I didn't realize that could be a problem in these cases. Do you know if there is a solution to this ?

I removed the files in “dist” from my PR.

Thank you for your reply. I will look into the Alpine code to see if I can come up with a solution.

**Comment by MaxencePaulin** [comment]
I can think of four things.
- Return the unwatch based on a parameter to $watch (doing this for $wire.$hook would change how it works, so it could be a breaking change).
```javascript
x-init="$wire.$watch('foo', ..., false)" // returning unwatch by default, have to mention "false" in third parameter to avoid it
// OR
x-init="() => {
    $data.unwatch = $wire.$watch('foo', ..., true) // not returning unwatch by default, have to mention "true" in third parameter
}"
```
- Instead of simply returning unwatch, put it in an object to prevent Alpine from being tempted to execute it inline
```javascript
return { unwatch }
```
- Condition the Alpine evaluator to avoid evaluating returns if they are functions such as $watch or $hook, but this must be more complex. I'll see if it's possible, but you know better than I do how it works, haha.
- Don't treat it as a problem, as is already the case for $wire.$hook, and mention in the documentation that if they want to use functions that return a callback, they must at least call them in an anonymous function. But I agree that if some people are already doing this, it could be a breaking change.

My opinion would be not to treat it as a problem because if Alpine is designed in such a way that a callback is executed if it is returned from a function, then using an anonymous function or an Alpine.data component to prevent execution seems fine to me.

PR #9524, opened 2025-10-29
Replace `lazy_placeholder` with `component_placeholder` in lazy docs

hebbet:
use correct config name in lazy docs

so it follows the stub file and https://livewire.laravel.com/docs/4.x/upgrading


joshhanley:
@hebbet thanks!

PR #9528, opened 2025-10-29
Update layouts docs

bookwyrm:
This is an initial pass at some updates to the docs related to layouts. Happy to iterate or make changes based on feedback.


calebporzio:
Thanks for the PR! I actually just made these tweaks, so closing, but good call and thanks

PR #9531, opened 2025-10-30
Fix composer fixtures issue

joshhanley:
# The scenario

Trying to install the Livewire v4 beta on a Mac through Composer is throwing the following error.

Screenshot 2025-10-30 at 11 43 30AM@2x

# The problem

The issue is that the macOS unzip doesn't seem to like the ⚡️ when composer tries to unzip the package.

According to ChatGPT:

> Yep, that’s macOS + unzip choking on your filename because of the fancy lightning bolt.
>
> What’s happening:
	1.	Your fixtures use ⚡︎ at the start of the filename / directory.
	2.	Git happily stores that in the repo using UTF-8 (cool).
	3.	Composer downloads a ZIP of the repo and calls the system /usr/bin/unzip to extract it.
	4.	/usr/bin/unzip on macOS tries to create those paths and blows up with Illegal byte sequence.
> 
> Why?
	•	The error Illegal byte sequence means: “I got bytes I can’t map to the current filesystem encoding.”
	•	macOS HFS+/APFS does Unicode but it normalizes filenames (NFD-ish). The lightning bolt you’re using has multiple visually similar codepoints and can include variation selectors.
	•	In the ZIP metadata, that character is coming through in a form that unzip (Info-ZIP) + your current locale can’t represent cleanly when it goes to actually open() the file. So it writes ������ in the log where the glyph should be and then gives up.

# The solution

Not sure how we can fix that exactly but we can work around it.

One issue I found is that the bolt emoji in the `Fixtures` directory is using `U+26A1 U+FE0F
` instead of using `U+26A1` which is a default when generated with `livewire:make`. This is likely because the files were created manually and the bolt emoji was added from the emoji picker, which adds `U+26A1 U+FE0F`.

But in testing this didn't fix the issue.

So the simpler solution is to exclude the `Fixtures` directory from the bundle. So that way the emoji files aren't included in the zip.

Now it installs without any errors.

Fixes livewire/livewire#9527


edo2313:
Doesn't this mean that the problem could arise again when zipping/unzipping a project that uses the manually-placed ⚡️ emoji in the filenames?
You shouldn't exclude the possibility that someone might decide to create a blade file and **then** implement livewire...

Personally, I don't really like the emoji in the filename because I think it will cause a lot of problems as not every system plays nice with Unicode characters (think about regexes, shell scripts, etc...): a suffix would be better imho (i.e. _component.live.blade.php_). Also when using a terminal writing the filename becomes a PITA :)

joshhanley:
@edo2313 yep possibly. That's something Caleb and I will have to discuss at a later date. This is just a fix for now to stop the composer error.

I love the emojis! Finding Livewire components in an IDE is much easier with the emoji 😁

PR #9549, opened 2025-11-02
Fix security tests by enabling snapshot tampering in test helper

yehuuu6:
The 3 security tests in ComponentsAreSecureUnitTest.php were skipped with @todo comments. These tests verify that the checksum validation properly detects and rejects tampered component payloads.

The issue was in the test infrastructure: The Testable class had a __get() magic method to read the snapshot, but no __set() method to write it. When tests tried to tamper with the snapshot using assignment ($component->snapshot = $tamperedSnapshot), PHP created a new property on the Testable object instead of updating the internal ComponentState. The subsequent request would use the original, untampered snapshot, so the validation never had a chance to detect the tampering.

The fix adds a __set() magic method to the Testable class that intercepts snapshot assignments and properly updates the internal ComponentState with the modified snapshot. This allows the security tests to actually tamper with the snapshot data, which then triggers the checksum validation (Checksum::verify() in HandleComponents.php line 213) as intended.

All 6 security tests now pass, and the full unit test suite (871 tests) continues to pass.

### 1️⃣ Is this something that is wanted/needed?

Yes. This re-enables 3 security tests that were skipped with `@todo` comments. These tests verify Livewire's checksum validation detects tampered payloads, critical security functionality that should be protected by tests.

### 2️⃣ Did you create a branch for your fix/feature?

Yes. Branch: `feature/fix-security-tests` (based on `v4.0.0-beta.2`)

### 3️⃣ Does it include multiple, unrelated changes?

**No.** Single focused fix: enable 3 skipped security tests by fixing the test infrastructure.

### 4️⃣ Does it include tests?

**Yes.** Re-enables 3 existing security tests that all now pass:
- `test_data_cannot_be_tampered_with_on_frontend`
- `test_id_cannot_be_tampered_with_on_frontend`
- `test_component_name_cannot_be_tampered_with_on_frontend`

Full suite: 871 tests pass ✅

### 5️⃣ Description

**Problem:** The `Testable` class had `__get()` for reading `snapshot` but no `__set()` for writing it. When tests tried to tamper with snapshots like `$component->snapshot = $tamperedSnapshot`, PHP created a new property instead of updating the internal `ComponentState`. Subsequent requests used the original snapshot, so checksum validation never detected tampering.

**Solution:** Added `__set()` magic method that intercepts snapshot assignments and properly updates the internal state:

```php
function __set($property, $value)
{
    if ($property === 'snapshot') {
        $this->lastState = new ComponentState(
            $this->lastState->getComponent(),
            $this->lastState->getResponse(),
            $this->lastState->getView(),
            $this->lastState->getHtml(),
            $value,
            $this->lastState->getEffects(),
        );
        return;
    }
    $this->instance()->$property = $value;
}
```

Now tampered snapshots are properly sent to the backend, allowing `Checksum::verify()` to detect mismatches and throw `CorruptComponentPayloadException`.


calebporzio:
Thanks for this. @joshhanley can you review? thanks

joshhanley (approved):
Looks good to me!

I just made one small change. At the end of `__set()`, instead of calling

```php
$this->instance()->$property = $value;
```

I have changed it to 

```php
$this->setProperty($property, $value);
```

This ensures the property update is handled properly.

It means that in tests doing `$component->foo = 'bar'` is now the equivalent of calling `$component->set('foo', 'bar')`

PR #9552, opened 2025-11-02
Add support for `DUSK_HEADLESS_DISABLED` in .env

ganyicz:
Currently, browser tests run in headed mode by default, which makes it impossible to use the computer while the tests are running. This can be annoying, especially since the test suite requires few minutes to run. Unless I'm missing something obvious, there was no easy way to run the tests in headless mode by default.

This addition makes it possible to add an `.env` file in the Livewire repo with `DUSK_HEADLESS_DISABLED=false`. The env file will be gitignored and the default remained the same (= headed mode). This approach makes it possible to define a default behavior for your local setup without affecting the repo itself.

I've also added instructions for running tests to the Contribution Guide.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ composer.lock
 .phpunit.result.cache
 .idea
 phpunit.xml
+.env
 
 # Files used to upload zips for early access folks
 upload-bin.sh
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks Filip, great addition.

Maybe would be good to mention alternatives to the .env file if that feels heavy:

- adding the env change directly in the bash command:
`DUSK_HEADLESS_DISABLED=false vendor/bin/phpunit`

- changing env in the phpunit.xml file:
`<env name="DUSK_HEADLESS_DISABLED" value="false"/>`

Makes sense. good addition.

PR #9553, opened 2025-11-02
Add support for testing SFC and MFC

ganyicz:
This PR makes it possible to test SFC/MFC components in the browser using `Livewire::visit('component-name')`, which was previously only possible by setting up a separate full-page component route, as can be seen in the modified test.

This is because strings passed into `Livewire::visit()` were considered to be a component class, which was then registered under a random unique id. The id was then used as a route parameter for the automatically generated dusk route.

This PR enables testing components by their registered name by checking if a string is a class. If not, the name of the component will be used instead of generating a unique id. The component name is then passed into the dusk route which will resolve the component using `Livewire::new()` in `Livewire\Features\SupportTesting\ShowDuskComponent`

### DIFF
--- a/src/Features/SupportSingleAndMultiFileComponents/BrowserTest.php
+++ b/src/Features/SupportSingleAndMultiFileComponents/BrowserTest.php
@@ -10,20 +10,14 @@ class BrowserTest extends \Tests\BrowserTestCase
     public static function tweakApplicationHook()
     {
         return function () {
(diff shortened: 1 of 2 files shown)


calebporzio:
Definitely cleaner. thanks!

The code change could benefit from a short comment explaining what different values are accepted.

If you get a sec, PR that as well.

PR #9554, opened 2025-11-02
Add comments to `Livewire::visit()`

ganyicz:
Adds comments to explain all values that are accepted by this method [as requested here](https://github.com/livewire/livewire/pull/9553#issuecomment-3478135441)

- Simple component name (eg. `counter`)
- Component class name (eg. `App\Livewire\Counter`)
- Anonymous class instance (eg. `new class extends Component {}`)


calebporzio:
hell ya thanks Filip

PR #9555, opened 2025-11-02
Fix `$wire` in bare scripts

ganyicz:
All properties and methods of the `$wire` object return `undefined` when using the bare script tag in SFC/MFC.

This is because in the generated script module, the `$wire` argument is an array: `[$wire]`.

```js
export function run($wire) {
    $wire.count // undefined
}
```

Which is caused by incorrect call to the `module.run.call()` when the script module is evaluated:

```js
import(path).then(module => {
    module.run.call(component.$wire, [
        component.$wire,
    ]);
})
```

Unlike `apply()`, `call()` expects the arguments to be passed directly:

```js
module.run.call(component.$wire, component.$wire);
```


ganyicz:
Included in https://github.com/livewire/livewire/pull/9556

PR #9556, opened 2025-11-02
Make `$js` available in scripts

ganyicz:
## The problem

The `$js` proxy is not accessible without `this.$js` or `$wire.$js` prefix, which caused the following test to fail:

```php
public function test_can_define_js_actions_though_dollar_js_magic_in_a_script()
{
    Livewire::visit(
        new class extends \Livewire\Component {
            public function render() { return <<<'HTML'
            <div>
                <button wire:click="$js.test" dusk="test">Test</button>
            </div>

            @script
            <script>
                $js('test', () => {
                    window.test = 'through dollar js'
                })
            </script>
            @endscript
            HTML; }
        }
    )
    ->click('@test')
    ->assertScript('window.test === "through dollar js"')
    ;
}
```

This could be a breaking change, since [the documentation](https://livewire.laravel.com/docs/3.x/actions#calling-from-php)  predominantly uses this syntax:

```html
<div>
    <!-- ... -->
 
</div>
 
@script
<script>
    $js('onPostSaved', () => {
        alert('Your post has been saved successfully!')
    })
</script>
@endscript
```

## The solution

This PR makes `$js` available in both inline `@script`s and `<script>` tags extracted from SFC/MFC.


joshhanley:
@ganyicz thanks for this! This is something I'd been waiting to talk with Caleb about, because he recently removed support for it in https://github.com/livewire/livewire/commit/ac3344e814a770f7b4da7b91f642d764ecd303b4, to see what he wanted to deprecate vs remove support for completely.

The v4 upgrade guide has details about changing from `$wire.$js('method', ...)` to `$wire.$js.method = ...`. Maybe it also needs an example of `$js()` too if that's what was documented in v3.

Actually looking at that actions page you linked on the v4 docs, it is documented as `$js.method = ...`. So maybe need a `$js('method')` => `$js.method` alongside what is already in the upgrade guide for `$wire.$js('method')` => `$wire.$js.method`.

Tests were recently updated from `$js()` to `this.$js()` in https://github.com/livewire/livewire/commit/8a3d5a9da5eea2117d985455d3ca10cb0f6a3f9d so maybe this one test was just missed.

@calebporzio thoughts?

ganyicz:
@joshhanley hey, thanks for checking this! I should have mentioned this in the OP - I actually checked in with Caleb before opening the PR. His response was "let's not make a breaking change, add $js without this/$wire prefix back"

I updated the docs to say that `$js` without prefix was also deprecated so that we're covering all legacy and new syntaxes.

calebporzio:
Perfect. thanks!

PR #9557, opened 2025-11-03
Fix failing browser tests

ganyicz:
Fixes two failing tests by updating legacy component registration and removing a `tinker()` call.

After https://github.com/livewire/livewire/pull/9556 and this PR are merged the entire test suite should run without errors.

### DIFF
--- a/src/Features/SupportQueryString/BrowserTest.php
+++ b/src/Features/SupportQueryString/BrowserTest.php
@@ -1041,11 +1041,8 @@ class BrowserTest extends \Tests\BrowserTestCase
 
     public function test_it_handles_query_string_params_without_values()
     {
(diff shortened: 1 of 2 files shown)


joshhanley:
@ganyicz thanks for the PR!

Just be careful with tests like this, what was being tested originally was that Livewire didn't break when the URL has a query string param without a value, like `?flag`. See the original PR #8197 for details.

But after your changes, the query string was being rendered like `?flag=1`.

I've fixed up the test now, so should all be good. Just something to be mindful of 🙂 I've added some comments and updated the test name to hopefully make it clearer for ourselves in the future!

ganyicz:
Oh shoot I totally missed that, thanks a lot @joshhanley!

PR #9561, opened 2025-11-03
Fix parallel requests to isolated components

ganyicz:
Split from https://github.com/livewire/livewire/pull/9550

This PR only solves the `#[Isolate]` bug, whereas the previous PR also included a change to `#[Async]`. The issue lies in multiple requests being fired at the same time, which is likely to happen when using events to communicate across components - so it makes more sense to to solve with `#[Isolate]`.

I could argue that it might be intuitive to solve the issue with `#[Async]` - it's what I did at first. But `#[Isolate]` is a more direct solution.

But let me know if you think that `#[Async]` should also prevent the following behavior and I will make a separate PR.

## The scenario

Imagine you have two components that both react to `countdown-started` event:

* A countdown that streams for 3 seconds when the event is triggered
* A component that displays start time when the event is triggered

```php
⚡countdown.blade.php

new class extends Component
{
    #[On('countdown-started')]
    public function startCountdown()
    {
        $value = 3;

        while ($value > 0) {
            $this->stream($value, true, ref: 'countdown');
            sleep(1);
            $value--;
        }
    }
};
?>

<div><span wire:ref="countdown"></span></div>
```

```php
⚡start-time.blade.php

new class extends Component
{
    public string $startTime;

    #[On('countdown-started')]
    public function updateTime() {
        $this->startTime = now()->format('Y-m-d H:i:s');
    }
};
?>

<div>{{ $startTime }}</div>
```

By default, the following happens when the event is triggered:

1. A request with both components is sent
2. Countdown starts and streams for 3 seconds
3. After 3 seconds, the start time updates

https://github.com/user-attachments/assets/5c821579-ae56-4988-9f9b-b6b82b57861a

This is expected, because Livewire bundles updates to multiple components into one request. But it should be preventable by marking one of the components with `#[Isolate]`, according to [the documentation](https://livewire.laravel.com/docs/3.x/bundling#isolating-component-requests):

> By using Livewire's #[Isolate] class attribute, you can mark a component as "isolated". This means that whenever that component makes a server roundtrip, it will attempt to isolate itself from other component requests.

However, adding `#[Isolate]` to just one of the components (either one), doesn't have any effect and the requests are always bundled together. This only starts working when **all** components are isolated.

## The problem

While Livewire creates a separate request for isolated components, it will later try to bundle outstanding messages into already existing requests, _even if those requests are to isolated components._

Eg. having 2 isolated components and 2 non-isolated components would result in just 2 requests being sent. Both isolated components will get their own request, but later the updates to non-isolated components will get added into one of the existing requests, which can be unexpected.

## The solution

Before adding an outstanding message to an existing request, we will first check if that request is to an isolated component, if it is, we will not bundle any more messages to it.


calebporzio:
@joshhanley - can you review this one? want to make sure we can't do this in the interactions.js file. Fine if this is the best approach, just want a second set of eyes that understands the request system. tyty

PR #9564, opened 2025-11-04
Change lazy bundle to be a modifier

joshhanley:
Instead of `<livewire:foo lazy lazy:bundle />` we now use `<livewire:foo lazy.bundle />`.
And same for defer `<livewire:foo defer.bundle />`.


calebporzio:
thanks mate!

PR #9568, opened 2025-11-04
Add CSP to docs navigation

danie-ramdhani:
I saw livewire V4 has CSP support now but not in livewire site yet.


joshhanley:
@danie-ramdhani thanks!

PR #9569, opened 2025-11-04
Support wire:model on input fields inside dialog elements

yehuuu6:
Fixes wire:model directives not working on input fields inside HTML5 <dialog> elements.

The issue: showModal() moves dialogs to the browser's top layer, breaking normal DOM parent traversal, causing wire:model to fail when finding the parent Livewire component.

Changes:
- js/lifecycle.js: Store component reference on dialog elements before top layer move
- js/directives/wire-model.js: Check for stored reference when element is inside a dialog
- js/morph.js: Preserve dialog open attribute during morphing to prevent auto-closing

Tests:
- Added 3 browser tests covering wire:model.blur, wire:model.live, and form submission
- All tests passing with CI=true for headless mode

Yes, this is from the V4 roadmap.

Yes, the branch is fix/wire-model-inside-dialogs

No, all changes are focused on making wire:model work inside dialog elements.
- lifecycle.js
- wire-model.js
- morph.js

Yes, added 3 browser tests in BrowserTest.php that cover wire:model.blur, wire:model.live, and form submission inside dialogs. All passing.

When you use wire:model on an input inside a element, it fails silently. The problem is that dialog.showModal() moves the dialog to the browser’s top layer, which is outside the normal DOM tree. When wire:model tries to find the parent Livewire component using closest() or parentElement traversal, it fails because the dialog is no longer a descendant of the component.

The fix stores a reference to the component on the dialog element before it moves to the top layer (in lifecycle.js). Then when wire:model runs and can’t find a component the normal way, it checks if the element is inside a dialog and uses that stored reference instead (in wire-model.js). There’s also a small change in morph.js to preserve the dialog’s open attribute during updates so it doesn’t randomly close when wire:model.live triggers updates.

https://github.com/user-attachments/assets/67f4f1b6-d4cc-4f01-8c72-9746cb5c7845

### DIFF
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -9,7 +9,16 @@ import Alpine from 'alpinejs'
 directive('model', ({ el, directive, component, cleanup }) => {
     // @todo: will need to probaby do this further upstream i just don't want to bog down the entire lifecycle right now...
     // this is to support slots properly...
-    component = findComponentByEl(el)
+    
+    // For elements inside a dialog, check if there's a stored component reference
+    // This is needed because showModal() moves dialogs to the top layer, breaking parent traversal
+    let dialog = el.closest('dialog')
+    
+    if (dialog && dialog.__livewireComponent) {
+        component = dialog.__livewireComponent
+    } else {
+        component = findComponentByEl(el)
+    }
 
     let { expression, modifiers } = directive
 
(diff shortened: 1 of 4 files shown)


joshhanley:
@yehuuu6 thanks for the PR! I just pulled down your tests so I could have a play to see what is going on, and the issue is actually that Livewire is wiping out the dialog state when it morphs. If you add `wire:ignore.self` to each of the `<dialog>` elements in your tests, everything works as expected and your tests pass. As such I'm going to close this for now. If you do find a scenario which doesn't work, then feel free to open another PR with a fix. Thanks for being willing to contribute!

Screenshot 2026-01-08 at 04 14 56PM@2x

PR #9578, opened 2025-11-06
Fix component discovery when `livewire.view_path` is customized

imhayatunnabi:

>    Yes—this fixes a real regression for anyone following the documented `livewire.view_path` override. No discussion thread was opened because it’s a straightforward defect with a direct fix.

>    Changes live on the feature branch `fix/view-path-finder` (no PRs from `main`).

>    Yes—only touches the service provider to wire up the correct config key plus adds this PR description.

>    Not yet. The existing suite doesn’t exercise custom `view_path` discovery, and adding a regression test would require plumbing config overrides into the finder boot process. I’ve added this as a follow-up TODO in the description below so it can be prioritized separately if desired.

>    Detailed below with context, impact, and reproduction steps.

## Problem
- `LivewireServiceProvider` registers the component finder with `config('livewire.component_path')` (src/LivewireServiceProvider.php:37-42), but Livewire doesn’t define that option anywhere. The documented knob for relocating Blade views is `livewire.view_path`.
- The make/convert commands already respect `view_path`, so any custom setup (e.g., `resources/views/app/livewire`) generates files in the new folder, but Finder never scans it. Components immediately fail to render until developers manually mirror the path inside `component_locations`, which is surprising and contradicts the docs.
- Teams that relocate Livewire views (a common pattern in modular apps) hit “view not found” errors right after running `php artisan make:livewire`.


joshhanley:
@imhayatunnabi thanks for the PR! I tested your fix, but then the make command tests failed. The `livewire.view_path` config is actually for class based components and as such I don't think it should be registered with the finder, otherwise view components (SFC/MFC) try to use this location when they shouldn't.

If you want to set where view components are found, use the `livewire.component_locations` config to set extra paths.

I've updated your PR to actually remove the `view_path` from the finder registration.

If you're still experiencing issues, please open a bug report and we'll investigate. Thanks!

imhayatunnabi:
Thanks to @joshhanley  for the review and refinements, and @calebporzio  for merging it.

PR #9581, opened 2025-11-08
Fixed mode not defined error

guratr:
Added mode variable declaration to fix 
Uncaught ReferenceError: mode is not defined


calebporzio:
Thank you

PR #9582, opened 2025-11-09
add redirectGuest() to HandlesRedirects

ghabriel25:

Yes, I've created a discussion and asked about this https://github.com/livewire/livewire/discussions/9575

Yes

No

No

This PR adds `redirectGuest()` method to `HandlesRedirects` class to provide redirect intended on subsequent request which is livewire request whenever its not satisfied the check. 

For example, there is a full page component which can be accessed from both authenticated user or guest but it contains a method that usually requires a user to be authenticated or others checking. Eventhough the component action is not visible for guest user using `@auth` directive, the method still can be called(tampered call) from client side.
```blade
<div>
    <livewire:features.like-post :$post />

    <livewire:features.comment-post :$post />
</div>
```
```php
class CommentPost extends Component
{
    public function saveComment(): void
    {
        if (! $this->isAuthenticated()) {
            return;
        }

        ...
    }

    protected function isAuthenticated(): bool
    {
        if (Auth::check()) {
            return true;
        }

        $this->redirectGuest(Login::class, navigate: true);

        return false;
    }
}
...
class LikePost extends Component
{
    public function liked(): void
    {
        if (! $this->hasPassword()) {
            return;
        }

        ...
    }

    protected function hasPassword(): bool
    {
        if (Auth::user()->password()) {
            return true;
        }

    $this->redirectGuest(CreatePassword::class, true);

        return false;
    }
}
```

Thanks for contributing! 🙌


calebporzio:
Thank you for this PR, I see what you're saying, but I think I'm going to close this in favor of people just writing the code you've written in the example themselves.

My reasoning is that the name `redirectGuest` implies that it will log a user out (at least it does to me). Maybe something like `redirectWithIntended()` or something is more relevant? However there isn't a method like that in Laravel to my knowledge and I'd rather preserve as much parity as possible.

Thanks for your efforts, but closing this.

ghabriel25:
@calebporzio I agree that the name implies like what you said and Its actually a method from laravel that I used. Like inside `EnsureEmailIsVerified` middleware
```php
    public function handle($request, Closure $next, $redirectToRoute = null)
    {
        if (! $request->user() ||
            ($request->user() instanceof MustVerifyEmail &&
            ! $request->user()->hasVerifiedEmail())) {
            return $request->expectsJson()
                ? abort(403, 'Your email address is not verified.')
                : Redirect::guest(URL::route($redirectToRoute ?: 'verification.notice'));
        }

        return $next($request);
    }
```

and the codebase behind it
```php
    public function guest($path, $status = 302, $headers = [], $secure = null)
    {
        $request = $this->generator->getRequest();

        $intended = $request->isMethod('GET') && $request->route() && ! $request->expectsJson()
            ? $this->generator->full()
            : $this->generator->previous();

        if ($intended) {
            $this->setIntendedUrl($intended);
        }

        return $this->to($path, $status, $headers, $secure);
    }

PR #9584, opened 2025-11-10
Add Volt to upgrade guide

joshhanley:
This PR documents the upgrade path from Volt class-based components to Livewire v4's single-file components.

Screenshot 2025-11-10 at 02 42 08PM@2x


utsav-infynno:
https://x.com/SomaiyaUtsav/status/1983367582607618267?s=20

I think, yess :smile:

PR #9586, opened 2025-11-12
exclude wire:scroll from wildcard handler

faisuc:
**1️⃣ Is this something that is wanted/needed?** Yes, fixes bug reported in discussion #8217

**2️⃣ Did you create a branch for your fix/feature?** Yes, branch: `fix/wire-scroll-alpine-expression-error`

**3️⃣ Does it contain multiple, unrelated changes?** No, single focused fix

**4️⃣ Does it include tests?** The fix prevents a runtime error. Manual testing confirmed the error no longer occurs. The navigate scroll plugin's existing functionality remains intact.

**5️⃣ Description:** Fixes Alpine expression error when using `wire:scroll` directive without a value (e.g., `<div wire:scroll>`).


calebporzio:
There are really two problems here:

A) The navigate feature uses `wire:scroll` to convey that a scrollable region should be preserved. This is a conflation because some folks might want to use `wire:scroll` to register a scroll event listener instead. I made a breaking change in Livewire 4 by changing to this: `wire:navigate:scroll`
B) Empty listener expressions currently throw an error. 

Both of these have been fixed in this PR: https://github.com/livewire/livewire/pull/9605

Thanks for putting some effort in here 👌

PR #9587, opened 2025-11-13
[3.x] add broadcastAs() documentation for Echo events

faisuc:
Added documentation for using `broadcastAs()` method with Laravel Echo events in Livewire components.

### What was added:

- New section "Customizing broadcast event names with `broadcastAs()`" in `docs/events.md`
- Examples showing how to implement `broadcastAs()` in Laravel event classes
- Documentation of the required dot prefix (`.`) when listening for custom broadcast names in Livewire
- Examples for both simple and dynamic channel names

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -543,6 +543,73 @@ public function notifyNewOrder($event)
 }
 ```
 
+### Customizing broadcast event names with `broadcastAs()`
+
+By default, Laravel broadcasts events using the event class name. However, you can customize the broadcast event name by implementing the `broadcastAs()` method in your event class.
+
+For example, if you have a `ScoreSubmitted` event but want to broadcast it as `score.submitted`:
+
+```php
+<?php
+
+namespace App\Events;
+
+use Illuminate\Broadcasting\Channel;
+use Illuminate\Broadcasting\InteractsWithSockets;
+use Illuminate\Contracts\Broadcasting\ShouldBroadcast;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Queue\SerializesModels;
+
+class ScoreSubmitted implements ShouldBroadcast
+{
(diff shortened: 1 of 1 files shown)


calebporzio:
Ooof, having to start an event name with a dot seems ick: `#[On('echo:scores,.score.submitted')]`

I bet it took you a while to find that.

I'm kind of rusty on all things Echo right now.

Is this surprising to users? How did you arrive at this conclusion that you needed the leading `.`?

Is there something we can do to get around this?

What would users expect you think? Any way we can support no leading `.`? without causing some other conflict? Thanks for your efforts here.

faisuc:
Yeah, the dot prefix is awkward. I found it after events weren't firing without it, then confirmed it in Echo's docs.

I think most users would be surprised by this. The dot tells Echo this isn't a namespaced class name.

We could detect custom broadcast names (like ones with dots that don't look like class names) and auto-add the dot prefix in Livewire. The question is whether that would break anything for people using actual namespaced events.

What do you think? Should we try to smooth this out, or just document it clearly?

nicolagianelli:
> Yeah, the dot prefix is awkward. I found it after events weren't firing without it, then confirmed it in Echo's docs.
> 
> I think most users would be surprised by this. The dot tells Echo this isn't a namespaced class name.
> 
> We could detect custom broadcast names (like ones with dots that don't look like class names) and auto-add the dot prefix in Livewire. The question is whether that would break anything for people using actual namespaced events.
> 
> What do you think? Should we try to smooth this out, or just document it clearly?

I think it should use the Echo convention with the dot prefix, people already using Echo would expect that pattern. You should only document it clearly by also linking the original Echo docs.

calebporzio:

calebporzio:
@claude - please update this PR to meet your recommendations - however your laravel docs link is broken - it goes to the right page but not the right section. please see if you can find the valid endpoint. thanks

PR #9594, opened 2025-11-14
Fix button wire directive for loading more items

aerni:
Took me a while to figure out that the docs are wrong.


calebporzio:
Sorry the docs are correct, but I think that change was on the main branch.

Try now with: v4.0.0-beta.3 and see if it works.

If not, re-open PR please thanks

aerni:
Works now, thanks!

PR #9597, opened 2025-11-15
Unify internal and user-defined debounce on `wire:model`

ganyicz:
## The scenario

Imagine you're building a custom text editor:

```html
<x-custom-editor wire:model.live="content" />

<!-- Preview -->
<div wire:text="content"></div>
```

The editor fires `input` events, which means we can use `wire:model.live` to sync the changes to the backend as the user types, and display a preview using `wire:text`.

## The problem

The above setup works well but sends a backend request with every keystroke.

This can be prevented by adding a `.debounce.150ms` modifier, mimicking the default Livewire behavior on a native `input` element with `wire:model.live` — where a 150ms debounce is added by default.

```html
<x-custom-editor wire:model.live.debounce.150ms="content" />

<!-- Preview -->
<div wire:text="content"></div>
```

However there is an important difference in how the "internal" and custom debounce works:

* Native input w/ internal debounce - `wire:text` is updated immediately after every keystroke
* Custom editor w/ user-defined debounce - `wire:text` is only updated when the user stops typing

This difference comes from the following code:

```js
let debouncedUpdate = isRealtimeInput(el) && ! isDebounced && isLive
    ? debounce(update, 150) // Internal debounce function
    : update

// ...

['x-model' + getModifierTail(modifiers)]() { // Custom debounce is forwarded to Alpine
    return {
        set(value) {
            dataSet(component.$wire, expression, value)

            isLive && (! isLazy) && (! onBlur) && debouncedUpdate()
        },
    }
}

```

Because the custom debounce is handled by Alpine, it will debounce the entire event handler, which includes the `dataSet` function that updates the ephemeral state on the frontend which is then displayed by `wire:text`.

This breaks the "optimistic update" effect of `wire:text`.

## The solution

This PR updates the `.debounce` and `.throttle` modifiers to use internal functions instead, and the modifiers will be removed from the modifier trail that is forwarded to Alpine.

### DIFF
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -30,6 +30,7 @@ directive('model', ({ el, directive, component, cleanup }) => {
     let isLazy = modifiers.includes('lazy') || modifiers.includes('change')
     let onBlur = modifiers.includes('blur')
     let isDebounced = modifiers.includes('debounce')
+    let isThrottled = modifiers.includes('throttle')
 
     // Trigger a network request (only if .live or .lazy is added to wire:model)...
     let update = () => {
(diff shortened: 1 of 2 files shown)


js/directives/wire-model.js
```diff
@@ -77,6 +82,22 @@ function getModifierTail(modifiers) {
         'lazy', 'defer'
     ].includes(i))
 
+    if (modifiers.includes('debounce')) {
```
calebporzio: Please add a comment here to describe why we are removing it (because we're using livewire debounce and don't want Alpine debounce)

calebporzio:
This is great, thanks Filip. Add a quick comment and we'll be good to merge 👌

PR #9599, opened 2025-11-16
Modify command for generating test files

techenby:
Updated command in the documentation to include '--mfc' flag for test file generation. Without adding this flag, no test file is generated, which was kind of confusing for me 😅.

I created a [discussion](https://github.com/livewire/livewire/discussions/9598) for commentary on how to handle generating `--test` flags for single-file components.


calebporzio:
I went ahead and now support support --test for single file components as well. Thanks for catching this Andy!

PR #9602, opened 2025-11-17
Fix ensureAnonymousClassHasReturn regex to check only class declaration

cyppe:
# Fix: Anonymous Class Return Statement Detection

## Problem

The `ensureAnonymousClassHasReturn()` method in `Parser.php` incorrectly checks if `return new` exists **anywhere** in the component file, not just on the anonymous class declaration line. This causes compiled files to be missing the `return` statement when components contain `return new SomeClass()` inside their methods.

## Error

```
TypeError: Cannot use "::class" on int
at vendor/livewire/livewire/src/Compiler/CacheManager.php:58
```

This occurs because `require` on line 56 of `CacheManager::getClassName()` returns `1` instead of the class instance.

## Example Component That Fails

```php
<?php

use Livewire\Component;

new class extends Component {
    public function getData()
    {
        return new Collection([]);  // ← This causes the bug
    }
};
```

**Current behavior:** The regex finds `return new Collection` and assumes the anonymous class already has `return`, so it skips adding it.

**Expected behavior:** Only check the anonymous class declaration line: `new class extends Component`

## Solution

Changed the regex to specifically match and check only the anonymous class declaration line using multiline mode and a negative lookbehind:

**Before:**
```php
if (preg_match('/\bnew\b/', $contents) && !preg_match('/\breturn\s+new\b/', $contents)) {
    return preg_replace('/\bnew\b/', 'return new', $contents, 1);
}
```

**After:**
```php
if (preg_match('/^(?!.*\breturn\s+).*\bnew\s+class\s+extends\s+Component/m', $contents)) {
    return preg_replace('/^(\s*)(new\s+class\s+extends\s+Component)/m', '$1return $2', $contents, 1);
}
```

## Testing

Tested with components that have `return new SomeClass()` in their methods. The compiler now correctly adds `return` to the anonymous class declaration while ignoring other `return new` statements in the file.

## Related Issues

Affects Livewire v4 beta when using multi-file components with methods that instantiate objects.

### DIFF
--- a/src/Compiler/Parser/Parser.php
+++ b/src/Compiler/Parser/Parser.php
@@ -52,9 +52,10 @@ class Parser
 
     protected function ensureAnonymousClassHasReturn(string $contents): string
     {
-        if (preg_match('/\bnew\b/', $contents) && !preg_match('/\breturn\s+new\b/', $contents)) {
-            return preg_replace('/\bnew\b/', 'return new', $contents, 1);
+        if (preg_match('/^(?!.*\breturn\s+).*\bnew\s+class\s+extends\s+Component/m', $contents)) {
+            return preg_replace('/^(\s*)(new\s+class\s+extends\s+Component)/m', '$1return $2', $contents, 1);
         }
+
         return $contents;
     }
 


calebporzio:
Great PR and writeup, thanks!

PR #9619, opened 2025-11-26
Fix missing named argument on registering class based component in docs

danie-ramdhani:
it will throw error if  `class` argument is not added

image

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -433,7 +433,7 @@ use Livewire\Livewire;
 // Register an individual class-based component
 Livewire::addComponent(
     name: 'todos',
-    \App\Livewire\Todos::class
+    class: \App\Livewire\Todos::class
 );
 
 // Register a location for class-based components


AtmoFX:
While you are at it, the few previous examples all use a parameter named `viewPath` when I think it should be named just `path`. Possible to correct it?

danie-ramdhani:
> While you are at it, the few previous examples all use a parameter named `viewPath` when I think it should be named just `path`. Possible to correct it?

Dude, this one is for class based component not view based. Please read the docs carefully.

joshhanley:
@danie-ramdhani thanks for the PR! Please don't be rude, they were trying to be helpful.

@AtmoFX yeah @danie-ramdhani is correct, the previous examples are for registering view based components thus require a `viewPath`. We elected to go with viewPath there because view could be ambiguous and mean a path or a dot notation view (like `livewire.admin.show-user`).

danie-ramdhani:
@joshhanley  sometimes I feel annoyed when i said it's in the docs and it's not bug but then he said (see: https://github.com/livewire/livewire/discussions/9615):

> My understanding from the Discord discussion is that Josh did expect my previous attempt(s) to work so I'll let him (or Caleb) decide what to do about it.

My bad, sorry.

PR #9621, opened 2025-11-27
Fix: keep version control over asset_url

Luc-cpl:
## This PR fixes an inconsistency in how Livewire applies versioning to its asset URLs.

The method `Livewire\Mechanisms\FrontendAssets\FrontendAssets::js()` always appends the build version (`?id=...`) to `config('livewire.asset_url')`.
However, the `usePublishedAssetsIfAvailable()` method does **not** apply the same versioning logic when a custom `asset_url` is configured. As a result, the published asset URL loses its version hash, causing stale assets to be loaded or cached incorrectly.

This update ensures that `config('livewire.asset_url')` also receives the `?id={version}` suffix when resolving published assets, restoring consistent versioning across both methods.

This brings the behavior of `usePublishedAssetsIfAvailable()` in line with the versioning logic already used in `FrontendAssets::js()`.

### DIFF
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -250,7 +250,7 @@ class FrontendAssets extends Mechanism
         }
 
         $publishedManifest = json_decode(file_get_contents(public_path('vendor/livewire/manifest.json')), true);
-        $versionedFileName = $publishedManifest['/livewire.js'];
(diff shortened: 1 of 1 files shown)


calebporzio:

calebporzio:
@claude please implement your suggestions

joshhanley:
@Luc-cpl thanks for the PR! Caleb and Claude have opened a new PR #9709 which is an extension of what you have done. So I'm going to close this one and finish it there. Thanks!

PR #9458, opened 2025-08-27
Fix bigint values from query string being truncated into scientific notation

doublerr999:
1️⃣ Is this something that is wanted/needed?
Yes. Currently, when a very large integer is passed via the query string and processed through `BaseUrl::setPropertyFromQueryString`, it is cast into **scientific notation** due to json_decode behavior. This causes data loss and unexpected values in Livewire components.

Did you create a discussion about it first?
No, should I still create a discussion?

2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed)
Yes, this PR was created from a feature branch, not main.

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
No

4️⃣ Does it include tests? (Required)
Yes

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

Currently, Livewire’s `BaseUrl::setPropertyFromQueryString()` decodes query string values using:
```
$decoded = is_array($initialValue)
    ? json_decode(json_encode($initialValue), true)
    : json_decode($initialValue ?? '', true);
```

When a large numeric string (e.g. `74350194073086909398128`) is passed in the query string,
**json_decode()** casts it into a PHP float, resulting in **scientific notation** (`7.4350194073087E+22`) and a loss of precision.

This PR fixes the issue by using **JSON_BIGINT_AS_STRING** when decoding non-array values:
```
$decoded = is_array($initialValue)
    ? json_decode(json_encode($initialValue), true)
    : json_decode($initialValue ?? '', true, flags: JSON_BIGINT_AS_STRING);
```

Why this is useful?
- Prevents silent data corruption when working with large IDs, transaction numbers, or other bigint values in query strings.
- Ensures consistency between original query string input and Livewire component state.

PR #9626, opened 2025-11-28
fixes a bug when big int is converted to float in scientific notation inside url query params

Vasiliy-Bondarenko:
The issue is pretty obvious to see on filament demo page https://demo.filamentphp.com/shop/products/products
try to search for this string "22000010620000001134" - it will search correctly and update url to `https://demo.filamentphp.com/shop/products/products?search=22000010620000001134` 
but then try to refresh the page (or copy paste the url to another tab) and the url will change to `https://demo.filamentphp.com/shop/products/products?search=2.200001062E%2B19` obviously - not what you would expect to happen. search string must stay string, not be converted to float in scientific notation. 

This PR fixes the problem, but I personally don't like the soultion. Whole approach of json-decoding url params seems off. What if i want to search actual json like string "[]"?.. well, right now now it will fail misarably with error 500. What about searching for some truth? :) let's search for string "true": `https://demo.filamentphp.com/shop/products/products?search=true` and right now it gets converted to "1": `https://demo.filamentphp.com/shop/products/products?search=1`. so this PR is an invitation to discuss better solution i guess :))
this PR fixes a problem with long ints (which is a serious problem for my project) but does not address bigger issue. it would be nice if someone familiar with overall architecture of livewire take a look at this more deeply.

it's a bug

yes, " bug-url-screws-big-ints"

nope. 

yes

above

### DIFF
--- a/src/Features/SupportQueryString/BaseUrl.php
+++ b/src/Features/SupportQueryString/BaseUrl.php
@@ -62,7 +62,7 @@ class BaseUrl extends LivewireAttribute
 
         $decoded = is_array($initialValue)
             ? json_decode(json_encode($initialValue), true)
-            : json_decode($initialValue ?? '', true);
+            : json_decode($initialValue ?? '', true, flags: JSON_BIGINT_AS_STRING);
 
         // If only part of an array is present in the query string,
         // we want to merge instead of override the value...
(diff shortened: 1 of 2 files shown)


joshhanley:
@Vasiliy-Bondarenko thanks for the PR! There is already an open PR #9458 with a fix for this, so I'm going to close this PR. Thanks for being willing to contribute!

PR #9632, opened 2025-12-01
Fix invalid `wire:stream` selector via escaping

godismyjudge95:
If you run this in your browser console you get an "is not a valid selector" error:
```js
document.querySelector(`[wire\\:stream.answer="answer"]`);
```

But if you change it to:
```js
document.querySelector(`[wire\\:stream\\.answer="answer"]`);
```

It works perfectly.


calebporzio:
Thanks!

PR #9637, opened 2025-12-03
Fix morph markers

joshhanley:
Fixes #9629
Fixes #9633


joshhanley:
Closing as I forgot to target 3.x initially, now it just wants to run tests against v4, even with the change.

PR #9643, opened 2025-12-04
Add lifecycle hooks for `#[reactive]` props

cyppe:
# RFC: Trigger Lifecycle Hooks for Reactive Props

## Summary

This PR proposes triggering `updating*()` and `updated*()` lifecycle hooks when `#[Reactive]` properties are updated by a parent component. Currently, these hooks are bypassed for reactive props, which can lead to unexpected behavior and hidden bugs.

## The Problem

When using `#[Reactive]` props, developers naturally expect that `updated*()` lifecycle hooks will fire when the prop value changes - similar to how `wire:model` updates work. However, the current implementation directly sets the value without triggering these hooks.

### Example of the bug/problem

```php
// Child component
#[Reactive]
public ?int $selectedVehicleId = null;

// ❌ This method NEVER fires when parent updates selectedVehicleId
public function updatedSelectedVehicleId(?int $value): void
{
    $this->checkCompatibility(); // Never runs!
}
```

This creates a "hidden bug" pattern where:
1. The code looks correct and follows Livewire conventions
2. The prop value does update correctly
3. But side effects in the lifecycle hook never execute
4. Developers spend hours debugging why their code doesn't work

## Current Behavior Analysis

Looking at `BaseReactive.php`:

```php
public function hydrate()
{
    if (SupportReactiveProps::hasPassedInProps($this->component->getId())) {
        $updatedValue = SupportReactiveProps::getPassedInProp(
            $this->component->getId(), $this->getName()
        );

        $this->setValue($updatedValue);  // ← Direct set, no hooks triggered
    }

    $this->originalValueHash = crc32(json_encode($this->getValue()));
}
```

Compare this to normal property updates in `HandleComponents.php`:

```php
public function updateProperty($component, $path, $value, $context)
{
    // ...
    $finish = trigger('update', $component, $path, $value);  // ← Triggers hooks
    // ...
    $this->setComponentPropertyAwareOfTypes($component, $property, $value);
    // ...
    return $finish;  // ← Calls updated* hooks
}
```

## Proposed Solution

Modify `BaseReactive::hydrate()` to trigger lifecycle hooks when the reactive prop value actually changes:

```php
public function hydrate()
{
    if (SupportReactiveProps::hasPassedInProps($this->component->getId())) {
        $updatedValue = SupportReactiveProps::getPassedInProp(
            $this->component->getId(), $this->getName()
        );

        $currentValue = $this->getValue();
        $propertyName = $this->getName();

        // Only trigger lifecycle hooks if value actually changed
        if ($currentValue !== $updatedValue) {
            // Trigger 'updating' and 'updated' lifecycle hooks
            $finish = trigger('update', $this->component, $propertyName, $updatedValue);

            $this->setValue($updatedValue);

            // Call the 'updated*' lifecycle hooks
            $finish();
        } else {
            $this->setValue($updatedValue);
        }
    }

    $this->originalValueHash = crc32(json_encode($this->getValue()));
}
```

### Key Points

1. **Only triggers when value changes** - If parent re-renders but the prop value is the same, no hooks fire (performance optimization)
2. **Uses existing `trigger('update')` mechanism** - Consistent with how other property updates work
3. **Backward compatible for display-only components** - Components that don't define `updated*()` methods continue to work identically
4. **Small, focused change** - ~15 lines of code

## Questions for Maintainers

1. **Was this behavior intentional?** Was bypassing lifecycle hooks a deliberate design choice, or an implementation detail that wasn't fully considered?

2. **Performance concerns?** Are there performance implications we haven't considered? The change only affects components with reactive props where the value actually changes.

3. **Philosophy of reactive props:** Should reactive props be "display-only" (parent owns state, child just renders), or should they support side effects via lifecycle hooks?

4. **Breaking change concerns:** Could this break existing code that relies on hooks NOT firing? (We think this is unlikely, as it would mean code that appears broken but "works" due to the bug.)

5. **Alternative: Opt-in behavior?** Would you prefer an opt-in approach like `#[Reactive(hooks: true)]` to maintain backward compatibility?

## Workarounds (Current Solutions)

Developers currently work around this limitation using:

### 1. The `hydrate()` Hook Pattern
```php
#[Reactive]
public ?int $selectedVehicleId = null;

public ?int $lastProcessedId = null;

public function hydrate(): void
{
    if ($this->lastProcessedId !== $this->selectedVehicleId) {
        $this->lastProcessedId = $this->selectedVehicleId;
        $this->doSideEffect();
    }
}
```

### 2. Events Instead of Reactive Props
```php
#[On('vehicle-state-updated')]
public function onVehicleChanged(array $data): void
{
    $this->selectedVehicleId = $data['selectedVehicleId'];
    $this->doSideEffect();
}
```

Both workarounds have downsides:
- `hydrate()` pattern is non-standard and requires tribal knowledge
- Events increase network requests (N+1 instead of 1)

## Tests Added

This PR includes browser tests that verify:

1. `updatedPropertyName()` fires when reactive prop changes
2. `updatingPropertyName()` fires before the value is set
3. Hooks do NOT fire when the value remains unchanged (performance)

## Documentation Impact

If accepted, the documentation for `#[Reactive]` should be updated to mention that lifecycle hooks are now triggered, matching developer expectations.


calebporzio:

calebporzio:
Seems like there is some validity to this PR and the implementation seems pretty sound. Because this is a foundational lifecycle hooks change, it would be good to introduce this before we tag official 4.0 release. My biggest concern with this one is performance impact to components that don't even use reactive. @joshhanley - can you take this one from here? thanks

joshhanley:
I've pushed a refactored implementation that addresses the performance concern raised by @calebporzio.

### The Problem

The original implementation modified `ComponentHookRegistry` to split hydration into two phases (init all hooks first, then boot/hydrate all hooks). This was needed because `BaseReactive::hydrate()` calls `trigger('update')`, which requires `SupportLifecycleHooks` to already be initialized. However, this change affected all components, even those without reactive props.

### The Solution

Instead of modifying `ComponentHookRegistry`, the new implementation:

1. **Queues updates** in `BaseReactive::hydrate()` instead of triggering them immediately
2. **Processes the queue** via an `after('hydrate')` listener in `SupportReactiveProps`

This ensures `SupportLifecycleHooks` is initialized before we trigger the update events, while keeping all the logic contained within the reactive props feature.

### Benefits

- No changes to `ComponentHookRegistry`
- Zero overhead for components without reactive props
- All reactive props logic stays within `SupportReactiveProps`

### How It Fits With Existing Patterns

This approach mirrors how Livewire already handles hydration and updates - all hydration finishes first, then updates run. With this implementation:

1. Component hydrates completely
2. After hydrate, reactive prop updates are triggered
3. Then normal component updates run

This keeps the lifecycle predictable and consistent with existing behavior.

joshhanley:
@cyppe thanks for the PR!

PR #9650, opened 2025-12-08
Fixes query string except bug after loading with existing value.

totov:

Yes, this fixes the issue outlined here: https://github.com/livewire/livewire/discussions/8791

Yes.

No.

Yes.

As per the discussion [here](https://github.com/livewire/livewire/discussions/8791), this resolves a bug where a property which is marked to be included in the URL query string would have strange behaviour when loading a page with an existing value and then changing it to the value used in `except`.

**Example:**

```php
...// within a livewire component
#[Url(except: false)]
public bool $foo = false;
...
```

**Expected Behaviour:**
Loading the page with the component present with the query string of `?foo=true` and then changing the value to false should hide it from the query string.

**Previous Behaviour:**
`?foo=true` would instead just change to `?foo=false`

**New Behaviour:**
`?foo=true` becomes `?`

I've kept the failing test showing this issue in a separate commit so it's easier to review/replicate.

The offending change which created this issue was part of the PR [here](https://github.com/livewire/livewire/pull/8528) but please note that the test related to this change did not require any changes as a result of this PR.

Please note, this did change some underlying behaviour and I therefore had to update some other tests in the same file which failed otherwise – I've kept these changes in a separate commit so it's easier to review. However, I think the new behaviour is actually what was originally intended anyway. The enum tests seemed to expect the value to be retained in the query string even when the value was effectively null – so I have updated these to use `keep`. The empty string being cast to null were essentially the same – an empty string being cast to null, so this should be tidied up as that would match the default `except` value.


calebporzio:

calebporzio:
@claude - can you remove the dist file from this PR? thanks

joshhanley:
@totov thanks for the PR!

totov:
No problem, is it helpful if I do the same for `main` since this was only aimed at `3.x` but is also a bug in `main`/v4?

https://github.com/livewire/livewire/blob/503908f15bef9a4e99c98e4b41ee9124766ea5b7/js/plugins/history/index.js#L50

joshhanley:
@totov I'll be up-merging all the changes from v3 to v4 shortly 🙂

PR #9655, opened 2025-12-12
Add wire:current.ignore

ganyicz:
# The scenario

Imagine you have a `x-nav-link` component that supports `wire:navigate`, meaning it adds active styling based on the `data-current` attribute. Now let's say you need to force the active styling for a specific use case, so you add an `current` prop:

```blade
<x-nav-link href="/homepage" wire:navigate :current="$isHomepageSubpage">Homepage</x-nav-link>
<x-nav-link href="/profile" wire:navigate>Profile</x-nav-link>
<x-nav-link href="/settings" wire:navigate>Settings</x-nav-link>

<!-- resources/views/components/nav-link.blade.php -->
@props(['current' => false])

<a {{ $attributes }} @if ($current) data-current @endif class="data-current:text-red-500">
    {{ $slot}}
</a>
```

# The problem

The above example won't work because in v4 `data-current` is automatically managed on all links that have `wire:navigate`. Therefore, even though the `data-current` attribute was manually added on to the element, it will be removed as soon as the page loads.

This came from an issue on Flux: https://github.com/livewire/flux/issues/2233

Where if you have a sidebar link with wire:navigate, overriding the `current` attribute will have no effect.

`<flux:sidebar.item href="/homepage" wire:navigate :current="$isHomepageSubpage">`

# The solution

One could refactor their code to apply the active styling without `data-current:` if the `current` prop is present, however that would lead to duplicated tailwind classes and goes against the idea of this feature, which encourages styling based on the data attribute.

I propose adding a new `ignore` modifier to `wire:current`.

This could be used in the following way:

```blade
<!-- resources/views/components/nav-link.blade.php -->
@props(['current' => false])

<a {{ $attributes }} @if ($current) data-current wire:current.ignore @endif class="data-current:text-red-500">
    {{ $slot}}
</a>
```

When this attribute is present on a link, Livewire will never add or remove the `data-current` attribute automatically.

### DIFF
--- a/docs/navigate.md
+++ b/docs/navigate.md
@@ -158,6 +158,12 @@ You can also use plain CSS to style active links:
 }
 ```
 
+If you would like to disable this behavior while still using `wire:navigate`, you may add the `wire:current.ignore` directive:
+
+```blade
+<a href="/posts" wire:navigate wire:current.ignore>Posts</a>
+```
+
 #### Using the `wire:current` directive
 
 Alternatively, you can use Livewire's `wire:current` directive to add CSS classes to the currently active link:
--- a/js/directives/wire-current.js
+++ b/js/directives/wire-current.js
@@ -15,7 +15,10 @@ globalDirective('current', ({ el, directive, cleanup }) => {
     let options = {
         exact: directive.modifiers.includes('exact'),
         strict: directive.modifiers.includes('strict'),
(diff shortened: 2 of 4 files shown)


willvincent:
This seems reasonable, unless there's a more automated way to determine if current is being explicitly set.

raz-iacob:
looks like a good solution to this problem!

PR #9665, opened 2025-12-17
Fix morph markers in scripts

joshhanley:
# The scenario

Currently if there are two Blade conditionals in a file, one in the HTML and the other inside a script tag, that are the same - the script tag won't function and morph markers will be output within the script. For example, the text in the screenshot below should be red if the script was actually running and not throwing errors.

Screenshot 2025-12-17 at 06 13 53PM@2x

```blade
<?php

use Livewire\Component;

new class extends Component
{
    public $today = 'Friday';
};
?>

<div>
    <div id="label">
        @if ($today === 'Friday')
            Friday
        @endif
    </div>

    <script type="text/javascript">
        @if ($today === 'Friday')
            document.getElementById('label').style.color = 'red';
        @endif
    </script>
</div>
```

# The problem

The problem is that we are individually matching directives and running a check to see if they are in a script tag. If they are, the match is skipped.

But when it comes to adding the morph markers to the directives, we are doing a `preg_replace` which replaces all instances of the match. This means that in the scenario above, even though we skipped the `@if` within the script tag for matching, it's still getting updated.

# The solution

The solution is to take the position of the match and us that to replace the string at that point.

To do this, I had to reverse the processing order, otherwise the matched positions will be incorrect after the first replacement.

Now the script to highlight the text in red is running because it no longer has morph markers in it.

Screenshot 2025-12-17 at 06 12 22PM@2x

### DIFF
--- a/src/Features/SupportMorphAwareBladeCompilation/SupportMorphAwareBladeCompilation.php
+++ b/src/Features/SupportMorphAwareBladeCompilation/SupportMorphAwareBladeCompilation.php
@@ -192,9 +192,6 @@ class SupportMorphAwareBladeCompilation extends ComponentHook
 
         $suffixEscaped = preg_quote($suffix);
 
-        // `preg_replace` replacement prop needs `$` and `\` to be escaped
-        $foundWithPrefixAndSuffix = addcslashes($prefix.$found.$suffix, '$\\');
-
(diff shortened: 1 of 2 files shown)


joshhanley:
Forgot to target 3.x branch, so tests ran against main instead.

PR #9668, opened 2025-12-17
Backport wire:current.ignore to v3

ganyicz:
In #9655, we added `wire:current.ignore` to v4, which is now also used in Flux.

Because this modifier wasn't implemented in v3, when a link has `wire:current.ignore`, it will act the same way as with `wire:current`, adding / removing the `data-current` attribute automatically.

### DIFF
--- a/js/directives/wire-current.js
+++ b/js/directives/wire-current.js
@@ -15,7 +15,10 @@ globalDirective('current', ({ el, directive, cleanup }) => {
     let options = {
         exact: directive.modifiers.includes('exact'),
         strict: directive.modifiers.includes('strict'),
+        ignore: directive.modifiers.includes('ignore'),
     }
+
+    if (options.ignore) return
 
     // Fragment hrefs aren't supported as of yet...
     if (expression.startsWith('#')) return


abishekrsrikaanth:
@ganyicz, @joshhanley, isn't this change (https://github.com/livewire/livewire/pull/9655/changes#diff-0f6641fa21a97249222ce2a84d2b17c82c4fb07099ec22098763143e9c1c46b4R23) not required as part of this backport? 

Not sure if the issue I am facing is related to this, but triggering a livewire action on volt page or a regular livewire component seems to remove the data-current attribute from the flux navbar links. So if a page has a navbar that uses the :current attribute, they loose the highlight. Not sure if it is just me or anyone else is having this issue? 

The only way to overcome this is to make the action `renderless`. But, `renderless()` actions is only possible on a need basis.

Here is an screencast indicating this behavior. If you notice on the screencast, the sidebar and the navbar both use the `:current` property to highlight the current page/tab the user is on. Post a `wire:submit` of a form, the navbar selection is lost. https://share.cleanshot.com/5tlxQnXT

ganyicz:
@abishekrsrikaanth I replied to your issue on Flux repo.

The code you're referencing doesn't exist in Livewire 3 so it wasn't required for this backport.

PR #9677, opened 2025-12-24
Add navigate `onSwap`

ganyicz:
When hooking into `livewire:navigated` to apply critical styles such as dark mode, the event might be delayed due to scripts being loaded.

This will result in flickering, which can be observed on Flux docs:

https://github.com/user-attachments/assets/c457f3c5-b4cb-4791-82ac-7dd84aa90d0f

The Editor page loads a separate script file for the component.

As such, the `Flux.applyAppearance()`, called in `livewire:navigated` listener, isn't called quickly enough.

This PR adds a new `livewire:navigating:swap` hook which is called immediately after the HTML is swapped but before scripts are fully loaded.

Using the new event fixes the issue:

https://github.com/user-attachments/assets/a3d6080c-ccd5-44d6-b801-f20c95af7cd9

### DIFF
--- a/docs/navigate.md
+++ b/docs/navigate.md
@@ -163,7 +163,6 @@ Each page navigation triggers three lifecycle hooks:
 
 * `livewire:navigate`
 * `livewire:navigating`
-* `livewire:navigating:swap`
 * `livewire:navigated`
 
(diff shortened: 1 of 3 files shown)


calebporzio:
Use this API:

```
document.addEventListener('livewire:navigating', (e) => {
    // Triggered when new HTML is about to swapped onto the page...

    // This is a good place to mutate any HTML before the page
    // is navigated away from...

    e.detail.onSwap(() => {

    })
})
```

ganyicz:
Done

PR #9678, opened 2025-12-26
Add support for converting Livewire class-based components to single-file and multi-file formats and vice-versa

dev-idkwhoami:
## Why

I was experimenting with Livewire v4.0.0-beta.5 by trying to "migrate" the Livewire Laravel starter kit to use v4. I noticed that while converting MFC to SFC worked, I was not able to convert class-based components to either.
A lot of developers use class-based components, since it was the default in v3. Its quite a lot of work to convert manually, especially in an existing project with dozens of components.

#### TL;DR
- convert command default conversion behavior
- conversion from & to class-based
- handling of full-page components
- added tests and moved some tests

The default behavior of the convert command was to simply swap between SFC and MFC. That had to change since we now have three different formats. So passing a format is now required and will be prompted if not passed via the cli. 

Additionally, I added a converter from SFC or MFC back to class-based, since class-based components are still supported and might be preferred by some developers. (for example at some point the starter kit will ship by default with MFC/SFC would be a pain to manually convert it to a class-based component)

Since class-based components all live in the same root directory, I had to differentiate between full-page components and simple components. As full-page components now need to live in the pages namespace in order to be rendered directly via a route. 

For this purpose, I added a flag --page, which the developer will need to pass if the component was used as a full page component. So the command knows to place the files in the correct folder (resources/views/pages). This flag WONT need to be passed if the developer is converting a SFC/MFC component into a class-based component if it is used as a full-page component, the finder checks both locations.

Since a few of the convert command tests were previously in the make command unit test, I moved those out into a new test file where the new convert functionality is.


calebporzio:
Thanks for the contribution. I can definitely see the utility with somethign like this. But making it deterministic seems pretty difficult because people have all sorts of custom set ups for their class components. For example, we would have to reference the view rendered in the render method and if they passed any data in we would have to put that into a computed property or something like that. That's a lot of parsing that will certainly be buggy.

For this reason I'm going to close this and wait and gauge more demand for this or a future attempt when that time comes.

Thanks anyways!

PR #9680, opened 2025-12-28
Auto-adjust pagination when current page becomes empty after deletion

mdiktushar:
Adds a new helper method `adjustPageIfEmpty` to the `HandlesPagination` trait.

- Automatically moves the paginator to the previous page if the current page becomes empty after deleting an item.
- Works with any `LengthAwarePaginator`, including filtered or searched queries.
- Ensures smooth UX without showing empty pages.

This prevents users from seeing empty pages after deletion when using Livewire pagination.

### Updated HandlesPagination Trait

```php
<?php

namespace Livewire\Features\SupportPagination;

use Illuminate\Pagination\Paginator;
use Illuminate\Support\Str;
use Illuminate\Pagination\LengthAwarePaginator;

trait HandlesPagination
{
    // ... existing pagination methods ...

    public function adjustPageIfEmpty(LengthAwarePaginator $paginator, string $pageName = 'page')
    {
        if ($paginator->isEmpty() && $this->getPage($pageName) > 1) {
            $this->previousPage($pageName);
        }
    }
    // ... existing pagination methods ...
}
```

### Example usage
In a Livewire component:
```php
use Livewire\Component;
use Livewire\WithPagination;
use App\Models\Post;
use Illuminate\Pagination\LengthAwarePaginator;

class PostsTable extends Component
{
    use WithPagination;

    #[Computed()]
    public function posts(): LengthAwarePaginator
    {
        return Post::paginate(10);
    }

    public function deletePost(Post $post): void
    {
        $post->delete();

        $this->adjustPageIfEmpty($this->posts);
    }
}

### DIFF
--- a/docs/pagination.md
+++ b/docs/pagination.md
@@ -426,9 +426,39 @@ For visual-only loading states (like opacity changes), you can use Livewire's au
 
 [Learn more about loading states →](/docs/4.x/loading-states)
 
+# Auto-adjust Pagination After Deleting Items
+
+When working with paginated lists in Livewire, deleting the **last item on the current page** can leave the page empty. This can lead to a confusing user experience where an empty page is displayed.
+
+The **`adjustPageIfEmpty`** helper automatically moves the paginator to the previous page if the current page becomes empty, ensuring users always see data when it exists.
+
+```php
+use Livewire\Component;
+use Livewire\WithPagination;
+use App\Models\Post;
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for the suggestion, however i think users can just have their own conditional in userland to keep the API surface area limited. thanks!

PR #9696, opened 2026-01-09
[4.x] Navigate view transitions

joshhanley:
WIP

Screenshot 2026-01-09 at 08 45 00PM

**First page**
```blade
<?php

use Livewire\Attributes\Url;
use Livewire\Component;

new class extends Component {
    //
}; ?>

<div>
    <flux:link href="/playground2" wire:navigate.transition>Playground 2</flux:link>

    <div class="h-screen flex items-center justify-center">
        <div wire:transition="bob">Bob</div>
    </div>
</div>

<style global>
  ::view-transition-group(bob) {
    animation-duration: 500ms;
    animation-timing-function: ease;
  }
</style>
```

**Second page**
```blade
<?php

use Livewire\Component;

new class extends Component {
    //
} ?>

<div class="flex items-center">
    <div wire:transition="bob">Bob</div>
</div>
```

### DIFF
--- a/js/directives/wire-navigate.js
+++ b/js/directives/wire-navigate.js
@@ -7,6 +7,17 @@ export let wireNavigateSelectors = [
     '[wire\\:navigate\\.preserve-scroll]',
     '[wire\\:navigate\\.preserve-scroll\\.hover]',
     '[wire\\:navigate\\.hover\\.preserve-scroll]',
+    '[wire\\:navigate\\.transition]',
+    '[wire\\:navigate\\.transition\\.hover]',
+    '[wire\\:navigate\\.hover\\.transition]',
+    '[wire\\:navigate\\.transition\\.preserve-scroll]',
+    '[wire\\:navigate\\.preserve-scroll\\.transition]',
+    '[wire\\:navigate\\.transition\\.hover\\.preserve-scroll]',
+    '[wire\\:navigate\\.transition\\.preserve-scroll\\.hover]',
+    '[wire\\:navigate\\.hover\\.transition\\.preserve-scroll]',
+    '[wire\\:navigate\\.hover\\.preserve-scroll\\.transition]',
+    '[wire\\:navigate\\.preserve-scroll\\.transition\\.hover]',
+    '[wire\\:navigate\\.preserve-scroll\\.hover\\.transition]',
 ]
 
 // Combined selector for querying all wire:navigate elements
@@ -19,6 +30,17 @@ let attributeMap = {
     'wire:navigate.preserve-scroll': 'x-navigate.preserve-scroll',
     'wire:navigate.preserve-scroll.hover': 'x-navigate.preserve-scroll.hover',
     'wire:navigate.hover.preserve-scroll': 'x-navigate.hover.preserve-scroll',
+    'wire:navigate.transition': 'x-navigate.transition',
+    'wire:navigate.transition.hover': 'x-navigate.transition.hover',
+    'wire:navigate.hover.transition': 'x-navigate.hover.transition',
+    'wire:navigate.transition.preserve-scroll': 'x-navigate.transition.preserve-scroll',
+    'wire:navigate.preserve-scroll.transition': 'x-navigate.preserve-scroll.transition',
+    'wire:navigate.transition.hover.preserve-scroll': 'x-navigate.transition.hover.preserve-scroll',
+    'wire:navigate.transition.preserve-scroll.hover': 'x-navigate.transition.preserve-scroll.hover',
+    'wire:navigate.hover.transition.preserve-scroll': 'x-navigate.hover.transition.preserve-scroll',
+    'wire:navigate.hover.preserve-scroll.transition': 'x-navigate.hover.preserve-scroll.transition',
+    'wire:navigate.preserve-scroll.transition.hover': 'x-navigate.preserve-scroll.transition.hover',
+    'wire:navigate.preserve-scroll.hover.transition': 'x-navigate.preserve-scroll.hover.transition',
 }
 
 // Register all selectors with Alpine
--- a/js/plugins/navigate/index.js
+++ b/js/plugins/navigate/index.js
@@ -13,11 +13,12 @@ let enablePersist = true
 let showProgressBar = true
 let restoreScroll = true
 let autofocus = false
+let enableTransition = false
 
 export default function (Alpine) {
 
     Alpine.navigate = (url, options = {}) => {
-        let { preserveScroll = false } = options
+        let { preserveScroll = false, transition = false } = options
 
         let destination = createUrlObjectFromString(url)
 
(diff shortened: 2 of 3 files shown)


calebporzio:
Mmmm yeah this is nice. we should probably use the same animation reset style element as the main wire-transition.md file - we can maybe even use the same method between the two.

like you have support for browsers that don't support view transitions and i don't in wire-transition.js which would benefit from that behavior as well.

As for the general vibe, I think this is good.

I have a FEW questions about API and expectations and defaults though like:

- [ ] Is `wire:navigate.transition` the right place for that API? Should it be on page components entirely? Like a `#[Transition]` attribute on the class? probably not... I feel like I would just want `wire:navigate.transition` on every navigation? maybe we do a similar strategy to wire:transition in general where we apply startViewTransition if we spot wire:transition on the page anywhere??? We can support a wire:navigate.no-transition, or a global Livewire.disableTransitions() or Livewire.navigate.disableTransitions() (which i think you have something like that). Just having a good API spec for all the affordances and interrogating it would be good
- [ ] Will conflating single-page usages of wire:transition with navigate ones be a problem? Will someone find themselves in a place where they need them to behave separately in some way? idk... probably not.
- [ ] Do we need to support transition types? like on the navigate level?
- [ ] If we add `wire:navigate.transition` we will also prolly need: `$this->redirect('...', navigate: true, transition: true)` which feels bad. should probably make the feature opt-out instead of opt-in you know... idk...
- [ ] If people aren't using navigate and want transitions for their multi-page app is there some future where we want to add apis for multi-page transitions and we painted ourselves into a corner with navigate transitions? i doubt it but worth thinking about.

Just some thoughts. Great idea, good start, makes sense.

calebporzio:
## PR Review: #9696 — [4.x] Navigate view transitions

**Type**: Feature
**Verdict**: Needs discussion

1. Currently, `wire:navigate` fetches a new page's HTML, then does `document.body.replaceWith(newBody)` — instant, no animation.
2. This PR wraps that body swap in the browser's [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/Document/startViewTransition) (`document.startViewTransition(() => swap())`), giving you cross-fade or custom CSS animations between page navigations.
3. Three ways to opt in: `wire:navigate.transition` on a link, `Alpine.navigate.enableTransition()` globally, or `Alpine.navigate(url, { transition: true })` programmatically.
4. This works *alongside* the existing `wire:transition` directive — `wire:transition="bob"` sets `viewTransitionName` on an element, and this PR makes `wire:navigate` actually use `startViewTransition()` so those names get animated.
5. Reduced-motion users get animations disabled via an injected `<style>` block.

1. **Auto-detect from `wire:transition` presence** — Caleb suggested: if any `wire:transition` element exists on the page, auto-enable view transitions for navigation. This would eliminate the need for `wire:navigate.transition` entirely and be opt-out instead of opt-in. This is worth serious consideration — it simplifies the API and avoids the `redirect(..., navigate: true, transition: true)` problem Caleb flagged.
2. **Page-level PHP attribute (`#[Transition]`)** — Caleb raised this and seemed to dismiss it. Agree — it should be a front-end concern.
3. **Always on, opt-out via `wire:navigate.no-transition`** — Mentioned by Caleb. Riskier since it could break existing apps, but simpler mental model for new projects.

### Unresolved API questions (from Caleb's comment)

These need answers before this can move forward:

1. **Opt-in vs opt-out?** — `wire:navigate.transition` per-link is opt-in, but if most users want transitions globally, opt-out (`Livewire.navigate.disableTransitions()` or `wire:navigate.no-transition`) may be better. The `redirect('...', navigate: true, transition: true)` argument is compelling — that API is ugly.
2. **Auto-detect from `wire:transition`?** — If `wire:transition` exists anywhere on the page, auto-wrap navigation in `startViewTransition()`. Elegantly sidesteps the whole opt-in question.
3. **Conflation with `wire:transition`?** — Will someone need navigate transitions and element transitions to behave independently? Probably not if the implementation just wraps the body swap — element-level `wire:transition` already sets `viewTransitionName` independently.
4. **Transition types on navigation?** — The View Transitions API supports [transition types](https://developer.mozilla.org/en-US/docs/Web/API/ViewTransition). `wire:transition` already supports types via `options.type`. Should `wire:navigate.transition` also? Not in v1 probably, but the `page.js` implementation currently uses the old callback-only form of `startViewTransition()`, not the config object form. Firefox 144+ only supports callback form anyway, so this is fine for now.
5. **Future multi-page transitions?** — Low risk of painting into a corner. The View Transitions API has a separate `@view-transition` CSS at-rule for cross-document (MPA) transitions.

### Code Review

**Combinatorial selector explosion** (`wire-navigate.js`): Adding `transition` blew the selectors from 5 to 16. If a 4th modifier is ever added, it's 64. This pattern needs to be replaced before merging — use a CSS attribute prefix selector or dynamically build the Alpine directive name from any `wire:navigate*` attribute. The `supportDataCurrent.js` consumer can use the same prefix approach.

**Duplicated reduced-motion style injection** (`page.js:43-49`): Identical pattern exists in `wire-transition.js:48-53`. If both fire simultaneously (navigate triggers a morph that also has `wire:transition` elements), you'll have two style blocks. Should be extracted into a shared utility.

**`transitionFinished` callback threading** (`index.js:120`, `page.js:62`): Clean approach — `swapCurrentPageWithNewHtml` passes `transitionFinished` (a `Promise`) as a second argument to the `andThen` callback. When no transition, it's `Promise.resolve()` so it's a no-op. Good pattern.

**Back/forward button uses global only** (`index.js:223`): `{ transition: enableTransition }` — correct, since there's no link to read `.transition` from. But if auto-detect is adopted, this should also auto-detect.

**Style cleanup**: `page.js:243` has a trailing semicolon (`return a & a;`). Minor, but project style is no semicolons in JS.

### Tests

No tests included. This is a WIP PR so that's expected, but before merge:
- Browser test: navigate with `.transition`, assert `document.getAnimations()` fires (or at minimum that `startViewTransition` was called).
- Browser test: navigate without `.transition`, assert no view transition.
- Browser test: `Alpine.navigate.enableTransition()` global opt-in.
- Browser test: `prefers-reduced-motion` disables animation.

### Security

No security concerns identified. This is purely client-side UI enhancement — no new server-side surface area.

### Verdict

**Needs discussion.** The implementation is solid and the feature is clearly wanted (2 reactions, natural progression of `wire:transition`), but Caleb raised 5 major API design questions in his comment that haven't been resolved yet. The biggest decision: **opt-in (`wire:navigate.transition`) vs auto-detect (use `startViewTransition` whenever `wire:transition` exists on the page) vs opt-out (always on, disable with config/directive)**. This shapes everything — the selector explosion, the `redirect()` API, and the mental model.

Recommend resolving the API direction before investing more implementation work. If auto-detect wins, this PR gets *simpler* (no `.transition` modifier needed on `wire:navigate` at all).

---
*Reviewed by Claude*

calebporzio:
closing because left in draft too long - thanks!

PR #9697, opened 2026-01-09
Fix incorrect variable in mergeQueuedUpdates causing queued updates to be lost

benbetterrx:
## Description

In `mergeQueuedUpdates()`, the condition `diffKey.startsWith(updateValue)` incorrectly uses the update's **value** instead of its **key** when checking if a diff entry should be removed.

This causes a critical bug when `updateValue` is an empty string: since every string starts with an empty string (`"foo".startsWith("")` returns `true`), ALL diff entries are incorrectly deleted, causing other queued updates to be lost.

### Example scenario that fails before this fix:

1. User has two inputs bound via `$wire.set('parent.foo', '')` and `$wire.set('parent.baz', '')`
2. Both are cleared (set to empty string `""`)
3. When the request is made, only the last update is sent because the first one was erroneously deleted from the diff

### The fix

Changes `diffKey.startsWith(updateValue)` to `diffKey.startsWith(updateKey)` so that nested properties are correctly identified by their key path (e.g., `'parent.foo'`) rather than their value.

## Checklist

- [x] Is this something that was discussed before? **Yes: https://github.com/livewire/livewire/discussions/7872**
- [x] Did you create a branch for your fix/feature? **Yes: `fix-merge-queued-updates-variable`**
- [x] Does this PR include unrelated changes? **No**
- [x] Does it include tests? **Yes - Browser test added to `SupportDataBinding/BrowserTest.php`**

### DIFF
--- a/js/component.js
+++ b/js/component.js
@@ -117,7 +117,7 @@ export class Component {
         // priority against ephemeral updates that have happend since them...
         Object.entries(this.queuedUpdates).forEach(([updateKey, updateValue]) => {
             Object.entries(diff).forEach(([diffKey, diffValue]) => {
-                if (diffKey.startsWith(updateValue)) {
+                if (diffKey.startsWith(updateKey)) {
                     delete diff[diffKey]
                 }
             })
--- a/src/Features/SupportDataBinding/BrowserTest.php
+++ b/src/Features/SupportDataBinding/BrowserTest.php
@@ -221,11 +221,6 @@ class BrowserTest extends BrowserTestCase
                 'baz' => 'qux',
             ];
 
-            public function showValues()
-            {
-                // This action triggers a server round-trip
-            }
-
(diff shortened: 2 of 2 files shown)


joshhanley (approved):
@benbetterrx thanks for the PR! Looks good to me.

To provide some context, before this fix, this is what the update payload looked like. You can see only `parent.baz` exists instead of both.

Screenshot 2026-01-12 at 05 21 20PM@2x


Where as after, you can see both keys are present.

Screenshot 2026-01-12 at 05 19 01PM@2x

calebporzio:
@claude can you review this for me please? these types of deep internal PRs are a bit nerve racking - so if you could advise on the safety of merge and the validity of the general approach and reframing the problem/solution so i can easily understand that would be great. thank you

PR #9699, opened 2026-01-10
Fix duplicate route registration when using custom Livewire update route

joshhanley:
## Summary
Fixes a break introduced in Laravel v12.29.0 which changed behavior so that two routes can no longer have the same name. This caused a collision when using a custom Livewire update route, as both the custom and default routes were being registered with the `livewire.update` name.

This PR ensures that only a single Livewire update route is ever registered, regardless of whether a custom route is configured or not.

- Wraps default route registration in `app()->booted()` to defer it until after all service providers have booted
- This deferral is necessary because custom update routes can be set in either service providers or routes files, both of which run after Livewire's service provider boots
- The deferred check skips default route registration if a custom route was already set, guaranteeing only one update route exists

## Test plan
- [x] Added test to verify only one route is registered when using custom update route
- [x] Added test to verify default route is registered when no custom route is set
- [x] Verified test fails when `app()->booted()` wrapper is removed

### DIFF
--- a/src/Mechanisms/HandleRequests/CustomUpdateRouteUnitTest.php
+++ b/src/Mechanisms/HandleRequests/CustomUpdateRouteUnitTest.php
@@ -1,0 +1,38 @@
+<?php
+
+use Illuminate\Support\Facades\Route;
+use Illuminate\Support\ServiceProvider;
+use Livewire\Livewire;
+use Tests\TestCase;
+
+class CustomUpdateRouteUnitTest extends TestCase
+{
+    protected function getPackageProviders($app)
+    {
+        return [
+            \Livewire\LivewireServiceProvider::class,
+            CustomUpdateRouteServiceProvider::class,
+        ];
+    }
+
+    public function test_only_one_livewire_update_route_is_registered_with_custom_update_routes(): void
+    {
+        $livewireUpdateRoutes = collect(Route::getRoutes()->getRoutes())->filter(function ($route) {
+            return str($route->getName())->endsWith('livewire.update');
+        });
+
+        $this->assertCount(1, $livewireUpdateRoutes);
+        $this->assertEquals('custom/livewire/update', $livewireUpdateRoutes->first()->uri());
+    }
+}
+
+// Service provider that sets custom route during boot (after Livewire's boot)
+class CustomUpdateRouteServiceProvider extends ServiceProvider
+{
+    public function boot()
+    {
+        Livewire::setUpdateRoute(function ($handle) {
+            return Route::post('/custom/livewire/update', $handle)->middleware('web');
+        });
+    }
+}
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -15,12 +15,14 @@ class HandleRequests extends Mechanism
 
     function boot()
     {
-        // Only set it if another provider hasn't already set it....
-        if (! $this->updateRoute) {
-            app($this::class)->setUpdateRoute(function ($handle) {
-                return Route::post('/livewire/update', $handle)->middleware('web');
-            });
-        }
+        // Only set it if another provider or routes file haven't already set it....
+        app()->booted(function () {
+            if (! $this->updateRoute) {
+                app($this::class)->setUpdateRoute(function ($handle) {
+                    return Route::post('/livewire/update', $handle)->middleware('web');
+                });
+            }
+        });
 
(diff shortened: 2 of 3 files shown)


calebporzio:
beautiful, thank you Josh!

parthibarath:
@joshhanley after updating livewire to v3.7.4 custom livewire update route not working in filament v4.5.3 

in previous livewire v3.7.3 i used to register the update route in boot method in appserviceprovider like below:
```
if ($this->app->isLocal()) {
        return Route::post('/{PROJECT_FOLDER}/public/livewire/update', $handle)->name('custom.livewire.update');
    });
        return Route::get('/{PROJECT_FOLDER}/public/livewire/livewire.js', $handle);
   });
}
```
but in current version i changed to
```
if ($this->app->isLocal()) {
       return Route::post('/{PROJECT_FOLDER}/public/livewire/update', $handle)->name('livewire.update')->middleware('web');
   });
       return Route::get('/{PROJECT_FOLDER}/public/livewire/livewire.js', $handle);
   });
}
```
in filament i click the  login it shows 404 not found livewire popup. Is i did anything wrong?

gehrisandro:
This PR has broken parallel test execution with route caching enabled: https://github.com/livewire/livewire/discussions/9757

oddvalue:
This PR causes a catch-all route in the application to be registered before the livewire update route so now all requests go to the catch-all.

Corys8646:
I am also seeing my update route return a 404 after upgrading the 3.7.5. I have had the following in my app service provider without issues. This seems to break after upgrading and clearing all the various cache files.

```
if (app()->environment('production')) {
                return Route::get('/portal/livewire/livewire.min.js', $handle);
            });
                return Route::post('/portal/livewire/update', $handle);
            });
        }
```

joshhanley:
@oddvalue and @Corys8646 I've just been testing this and I can't manage to replicate either of your issues. Can you confirm what Laravel versions you are on? If possible, could you provide a small example repo that demonstrates the issue so we can investigate? Thanks!

@parthibarath and @gehrisandro you issues should have been fixed in the latest update.

gehrisandro:
> @oddvalue and @Corys8646 I've just been testing this and I can't manage to replicate either of your issues. Can you confirm what Laravel versions you are on? If possible, could you provide a small example repo that demonstrates the issue so we can investigate? Thanks!
> 
> @parthibarath and @gehrisandro you issues should have been fixed in the latest update.

Thanks @joshhanley It works great!

oddvalue:
> @oddvalue and @Corys8646 I've just been testing this and I can't manage to replicate either of your issues. Can you confirm what Laravel versions you are on? If possible, could you provide a small example repo that demonstrates the issue so we can investigate? Thanks!
> 
> @parthibarath and @gehrisandro you issues should have been fixed in the latest update.

https://github.com/oddvalue/livewire-route-issue/commit/9551682a27d9cfa5024826afb806526b7ab5a7da

I appreciate this is a livewire 4 app but the issue exists the same in both versions.

Corys8646:
I'm on Laravel 12.48.1 (latest). When I view my route list with Livewire 3.7.3, there are two entries:

One for the default route 
**/livewire/update, route name livewire.update** 

and another for my custom route
**(/portal/livewire/update, route name custom-livewire.update).**

After upgrading to 3.7.5, the default route is no longer showing (i'm guessing by design). That is the only difference I can see between the two versions. 

The URL for the fetch request is the same for both versions. Just the 3.7.5 returns a 404

I will see if I can come up with an example repo.

Thanks for taking the time to look.

parthibarath:
> I'm on Laravel 12.48.1 (latest). When I view my route list with Livewire 3.7.3, there are two entries:
> 
> One for the default route **/livewire/update, route name livewire.update**
> 
> and another for my custom route **(/portal/livewire/update, route name custom-livewire.update).**
> 
> After upgrading to 3.7.5, the default route is no longer showing (i'm guessing by design). That is the only difference I can see between the two versions.
> 
> The URL for the fetch request is the same for both versions. Just the 3.7.5 returns a 404
> 
> I will see if I can come up with an example repo.
> 
> Thanks for taking the time to look.

@joshhanley I am updating my filament project from livewire v3.7.3 to v3.7.6. I am also getting the same issue

joshhanley:
@oddvalue thanks for that! I've submitted a new PR #9854 with a fix for catch-all routes. 

@Corys8646 or @parthibarath I'd still love to have a look at a reproduction of your issue if possible, just to make sure we have it covered before we tag another version. I'd expect that the custom route should be used for the fetch request, not the default update route, so I'm not sure why it would be 404ing.

PR #9702, opened 2026-01-12
Add synths to query string

joshhanley:
## The Problem

Previously, values loaded from the query string were always strings. If you had a typed property like `Carbon`, Livewire would throw an error when trying to assign the string value from the URL. Visiting `?created=2024-01-15` would result in:

Screenshot 2026-01-12 at 03 45 55PM@2x

```php
<?php

use Carbon\Carbon;
use Livewire\Attributes\Url;
use Livewire\Component;

new class extends Component {
    #[Url]
    public Carbon $created;
}; ?>

<div>
    {{ $created }}
</div>
```

## The Solution

Query string values now pass through the appropriate synth based on the property's type hint:

- `?page=5` → integer `5`
- `?price=9.99` → float `9.99`
- `?date=2024-01-15` → Carbon instance

This works for:
1. **Component properties** with `#[Url]` attribute
2. **Component properties** with legacy `$queryString` syntax
3. **Form object properties** with individual `#[Url]` attributes

### Small breaking change

At the moment, there are tests which check and ensure that if an invalid enum value is provided, then the property is set to null rather than throwing an error.

- test_can_use_url_on_string_backed_enum_object_properties_with_initial_invalid_value_on_nullable
- test_can_use_url_on_integer_backed_enum_object_properties_with_initial_invalid_value_on_nullable

This doesn't seem right to me. I think if the value provided is incorrect for the type, then it should throw the error, not mark it as null instead, as I feel that would be unexpected for a user rather than getting an error if their query string value is wrong.

So I've changed the tests but I'm not 100% on this change. I will document it in the upgrade guide if we decide to go with it.

The nullable was added in PR #9140.

### Form Object Hydration

The key addition in this PR is the ability to hydrate an **entire form object** from the query string with proper type conversion for its properties.

Previously, you could only bind individual form properties to the URL:

```php
class MyForm extends Form
{
    #[Url]
    public int $page = 1;  // Each property needs its own #[Url]
}
```

Now you can bind the entire form and have all typed properties hydrated correctly:

```php
class Component extends Livewire\Component
{
    #[Url]
    public MyForm $form;  // Bind entire form to URL
}

class MyForm extends Form
{
    public int $count = 0;
    public float $total = 0.0;
    public ?Carbon $created = null;
}
```

Visiting `?form[count]=10&form[total]=19.99&form[created]=2024-01-15` will hydrate all properties with their correct types.

This PR originated from the ideas presented in https://github.com/livewire/livewire/pull/9318

### DIFF
--- a/src/Features/SupportQueryString/BaseUrl.php
+++ b/src/Features/SupportQueryString/BaseUrl.php
@@ -4,6 +4,8 @@ namespace Livewire\Features\SupportQueryString;
 
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
 use Livewire\Features\SupportFormObjects\Form;
+use Livewire\Mechanisms\HandleComponents\ComponentContext;
+use Livewire\Mechanisms\HandleComponents\HandleComponents;
 use ReflectionClass;
 
 #[\Attribute]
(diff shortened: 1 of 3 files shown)


joshhanley:
Thinking about this some more, it will cause security concerns, especially if something is hydrated when it shouldn't like an Eloquent model. So going to close this for now.

PR #9713, opened 2026-01-13
Add token verification for file upload paths

joshhanley:
## Summary

This PR signs temporary file upload paths with a short HMAC token. The token is verified before accepting paths, ensuring they were generated by the server for the current session.

## How it works

- When a file is uploaded, the returned path is signed with an 8-character token: `{token}:{path}`
- The token is generated using `hash_hmac('sha256', $path, app_key + session_id)`
- When `_finishUpload` is called, it regenerates the token and compares it
- Invalid or missing tokens result in a 403 response

## Changes

- **TemporaryUploadedFile.php** - Added `generateToken()`, `signPath()`, and `extractPathFromSignedPath()` methods
- **FileUploadController.php** - Signs paths after local uploads
- **GenerateSignedUploadUrl.php** - Signs paths for S3 uploads
- **WithFileUploads.php** - Verifies signed paths in `_finishUpload`
- **BrowserTest.php** - Added tests for rejection of forged/unsigned paths

I've also manually tested with both local and S3 uploads to ensure everything works as expected.

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -249,6 +249,35 @@ class TemporaryUploadedFile extends UploadedFile
         return new static($filePath, FileUploadConfiguration::disk());
     }
 
+    /**
+     * Generate a short token for a given path using the app key and session ID.
+     * This ensures tokens are unique per session and cannot be forged without the app key.
+     */
+    protected static function generateToken(string $path): string
+    {
+        return substr(hash_hmac('sha256', $path, app('encrypter')->getKey() . session()->getId()), 0, 8);
+    }
+
+    public static function signPath(string $path): string
+    {
+        return static::generateToken($path) . ':' . $path;
+    }
+
+    public static function extractPathFromSignedPath(string $signedPath): string|false
+    {
+        if (! str_contains($signedPath, ':')) {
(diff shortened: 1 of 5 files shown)


calebporzio:
you're a legend Hanley

src/Features/SupportFileUploads/TemporaryUploadedFile.php
```diff
@@ -249,6 +249,35 @@ public static function createFromLivewire($filePath)
         return new static($filePath, FileUploadConfiguration::disk());
     }
 
+    /**
+     * Generate a short token for a given path using the app key and session ID.
+     * This ensures tokens are unique per session and cannot be forged without the app key.
+     */
+    protected static function generateToken(string $path): string
+    {
+        return substr(hash_hmac('sha256', $path, app('encrypter')->getKey() . session()->getId()), 0, 8);
```
casanasvictor: can you explain me how it supposed to work, like the session id is it supposed to be same when signPath function and extractPathFromSignedPath (_finishUpload), in my case im getting 2 different sessions and it result in error 403

i tested that in a multi-tenant app, no auth user, but i think none of that should be an issue, the url whne uploading a file is: ivewire-014d0d34/upload-file, different than the url i specify using  Livewire::setUpdateRoute.

also i testes using SESSION_DRIVER as file and database

PR #9720, opened 2026-01-14
Add failing test for `#[Js]` attribute 

sajjadhossainshohag:

No

Yes

No unrelated changes. 

Yes

This PR introduces a failing test to highlight an issue with Livewire’s #[Js] attribute when the JS-decorated method is invoked from another PHP method.

**Current behavior (problem)**

The incrementUpdateMessage() method is marked with `#[Js]`, but when it is called from inside another Livewire method (increment()),
the returned JavaScript is not executed on the client

As a result:

- assertDialogOpened('Count updated') fails
- No browser alert is triggered

**Expected behavior**

When a method decorated with `#[Js]` is called from another Livewire action, The JavaScript returned by the `#[Js]` method should be executed on the client.

Here's the wirebox link: https://wirebox.app/b/4q0je

Thanks for contributing! 🙌

### DIFF
--- a/src/Features/SupportJsEvaluation/BrowserTest.php
+++ b/src/Features/SupportJsEvaluation/BrowserTest.php
@@ -279,4 +279,42 @@ class BrowserTest extends \Tests\BrowserTestCase
         ->assertScript('window.test === "through backend js method with params: foo, bar"')
         ;
     }
+
+    public function test_can_call_a_defined_js_from_another_method_using_the_js_attribute()
+    {
+        Livewire::visit(
+            new class extends \Livewire\Component {
+                public $count = 0;
+
+                public function increment()
+                {
+                    $this->count++;
+                    $this->incrementUpdateMessage();
+                }
+
+                #[\Livewire\Attributes\Js]
+                public function incrementUpdateMessage()
+                {
+                    return "alert('Count updated') ";
+                }
+
+                public function render()
+                {
+                    return <<<'BLADE'
+                        <div>
+                            <h2>Counter : {{ $count }}</h2>
+                            <button wire:click="increment" dusk="increment">Increment</button>
+                        </div>
+                    BLADE;
+                }
+            }
+        )
+            ->waitForLivewire()
+            ->click('@increment')
+            ->pause(1000)
+            ->assertDialogOpened('Count updated')
+            ->acceptDialog()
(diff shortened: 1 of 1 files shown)


joshhanley:
@sajjadhossainshohag thanks for the PR! Is this something that worked in v3 but is no longer working in v4? In future, please be sure to add to the PR description why you have submitted it and the problem you are facing. Thanks!

sajjadhossainshohag:
@joshhanley Thanks for your feedback, I have updated PR description. 

I tried it on v3 too, but it’s not working.

Here’s the v3 Wirebox link: https://wirebox.app/b/x6p01

joshhanley:
Thanks for the detailed reproduction! You're right that this doesn't work as the documentation suggests — but after investigation, this is actually a **documentation issue** rather than an implementation bug.

### What `#[Js]` Actually Does

The `#[Js]` attribute is designed for creating **pure client-side methods** that bypass the server entirely. When you use `#[Js]` on a method:

1. During dehydration, Livewire pushes the method's return value as a `js` effect
2. On the client, this **registers** the method as a client-side override
3. Future calls to `$wire.methodName()` execute the JS directly without a server round-trip

It's meant for this pattern:
```php
#[Js]
public function toggle()
{
    return '$wire.show = !$wire.show';
}
```
```blade
<button @click="$wire.toggle">Toggle</button> <!-- No server request! -->
```

### Why Your Test Doesn't Work

When `increment()` calls `$this->incrementUpdateMessage()`, it just returns a string — nothing special happens. The `#[Js]` attribute only registers the method for client-side use; it doesn't execute the JS when called from PHP.

### The Correct Approach

For executing JavaScript **after a server action completes**, use `$this->js()` instead:

```php
public function increment()
{
    $this->count++;
    $this->js("alert('Count updated')");
}
```

This uses a different mechanism (`xjs` effect) that executes the JS immediately when the response arrives.

### The Documentation Issue

The example in `docs/attribute-js.md` showing `return $this->showSuccessMessage()` from a server action is misleading — this pattern doesn't work as documented. Interestingly, the `#[Js]` attribute was never documented in v3; this documentation was added for v4 with an incorrect example.

We'll need to either fix or remove that documentation. Thanks for surfacing this!

joshhanley:
I've updated the JS attribute documentation in #9723, hence am closing this. Thanks again!

PR #9733, opened 2026-01-15
docs: correct debounce and throttle modifier documentation

aarongmx:
Clarify that debounce and throttle modifiers require the .live modifier and update their descriptions for consistency.


calebporzio:
good call

PR #9735, opened 2026-01-15
Add callback support to Route::livewire

amdad121:
Allows passing a closure to `Route::livewire` for dynamic component selection based on runtime conditions like user roles.

Usage Example

```php
Route::livewire('/dashboard', function () {
    return match (auth()->user()->role) {
        'admin' => 'pages::admin.dashboard',
        'manager' => 'pages::manager.dashboard',
        default => 'pages::user.dashboard',
    };
})->name('dashboard');
```

### DIFF
--- a/src/Features/SupportRouting/LivewirePageController.php
+++ b/src/Features/SupportRouting/LivewirePageController.php
@@ -8,7 +8,11 @@ class LivewirePageController
     {
         $component = request()->route()->defaults['_livewire_component'];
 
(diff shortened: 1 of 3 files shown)


calebporzio:
While I think this would be nice, supporting closures can be tricky as I believe they aren't cached and can cause serializable problems and such. So to keep the framework running smoothly I'm going to close this. thanks anyways!

PR #9748, opened 2026-01-17
Fix ClassMorphViolationException for Eloquent collections when enforceMorphMap is active

calebporzio:
When using Relation::enforceMorphMap() in Laravel, the getMorphClass() method throws a ClassMorphViolationException for models that don't have a defined morph entry. This fix wraps the getMorphClass() call in EloquentCollectionSynth in a try-catch block to fall back to the class name, matching the existing behavior in ModelSynth and EloquentModelSynth.

Fixes #6862


calebporzio:
@claude - explain to me why this is necessary. why this is safe to do? what the problem in the first place is? if this PR addresses the exact issue referenced in the discussion in this PR description? Explain it to me simply. Thank you.

calebporzio:
@claude - gotcha. my question is, if a user enabled: Relation::enforceMorphMap(), doesn't that mean they WANT it to fail if there's no morph map? and that exception is a good thing? is ignoring it not a way around some desire for security or something? or maybe i missunderstand the intent of enforcing morph maps and somehow this doesn't side-step that. lmk

PR #9750, opened 2026-01-17
Fix stack overflow when transmitting large payloads

calebporzio:
The fingerprint getter in action.js used String.fromCharCode(...array) with the spread operator, which fails when payloads exceed ~125KB due to JavaScript's argument limit. This caused "Maximum call stack size exceeded" errors when sending large data like base64-encoded images.

The fix processes the encoded bytes in 8KB chunks, avoiding the spread operator limit while maintaining the same fingerprint output.

Fixes #9747


Youssef-Amjad:
Thank you, my friend ❤️

Youssef-Amjad:
Thank you, my friend ❤️

PR #9760, opened 2026-01-18
Fix parallel tests using WithCachedRoutes

calebporzio:
Original discussion: https://github.com/livewire/livewire/discussions/9757

When running parallel tests with WithCachedRoutes trait, routes are loaded from cache but HandleRequests instance is fresh (updateRoute is null). This caused duplicate route registration attempts since Laravel v12.29.0+ treats duplicate route names as an error.

The fix adds updateRouteExists() to check if any route with name ending in 'livewire.update' already exists in the router before attempting to register the default route. This handles both:
- Exact match: 'livewire.update'
- Prefixed custom routes: 'tenant.livewire.update'

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -21,7 +21,11 @@ class HandleRequests extends Mechanism
     {
         // Only set it if another provider or routes file haven't already set it....
         app()->booted(function () {
-            if (! $this->updateRoute) {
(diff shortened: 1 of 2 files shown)


calebporzio:
@claude - This makes sense however, can you test the performance impact of looping through all the routes every request from the service provider? If negligable, we're good, otherwise, can you propose another solution? thanks

joshhanley:
I have backported this to v3 as well in 9abb4cf

PR #9770, opened 2026-01-19
Fix class-based component generation when using root views path

kaue-f:
The Livewire class-based component was incorrectly loading the view using an absolute filesystem path.
This has been fixed by using the proper relative view path, as required by Laravel, ensuring correct component rendering across environments.
```php artisan livewire:make teste --class```

image

And also fixed how the Livewire config namespace is resolved.
The namespace was previously hardcoded `(App\\Livewire)` and is now correctly retrieved from the `livewire.class_namespace` config value, with a fallback to the default.

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
+++ b/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
@@ -2,18 +2,18 @@
 
 namespace Livewire\Features\SupportConsoleCommands\Commands;
 
-use Illuminate\Support\Str;
-use Livewire\Finder\Finder;
-use Illuminate\Console\Command;
-use Livewire\Compiler\Compiler;
 use function Laravel\Prompts\text;
-use Illuminate\Foundation\Inspiring;
 use function Laravel\Prompts\confirm;
-use Illuminate\Filesystem\Filesystem;
-use Livewire\Compiler\Parser\SingleFileParser;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by joshhanley** [comment]
@kaue-f Thanks for the PR!

*The following summary was generated by Claude:*

# The Scenario

When creating a class-based Livewire component with a custom configuration where views are stored directly in the `views` folder (not a subdirectory), the generated component contains an incorrect view path.

```php
// config/livewire.php
return [
    'class_namespace' => 'App\\Components',
    'view_path' => resource_path('views'), // views directly in views folder, not views/livewire
];
```

```bash
php artisan livewire:make teste --class
```

The generated component would have an incorrect `render()` method:

```php
// Generated with bug - absolute path or leading dot
public function render()
{
    return view('/Users/dev/project/resources/views.teste');
    // or
    return view('.teste');
}
```

# The Problem

Two issues in `buildClassBasedComponentClass()`:

1. **Hardcoded namespace**: The class namespace defaults to `'App\\Livewire'` instead of reading from `config('livewire.class_namespace')`.

2. **View path extraction fails for root views directory**: The `extractViewNamespace()` method doesn't handle the case where `view_path` equals `resource_path('views')`. When they match, the `str_replace()` calls don't find a match and return the full path unchanged.

3. **Leading dot in view name**: The view name was built using string concatenation (`$viewNamespace . '.' . ...`), which produces a leading dot when the namespace is empty.

# The Solution

1. Read the class namespace from config with a fallback: `config('livewire.class_namespace', 'App\\Livewire')`

2. Add an early return in `extractViewNamespace()` when the view path equals the base views path, returning an empty string.

3. Build the view name using collection methods with `->prepend($viewNamespace)->filter()->implode('.')` to automatically filter out empty values before joining.

PR #9772, opened 2026-01-19
[4.x] Move temporary file if disks are same

pxlrbt:
I realized that Livewire does not move temporary files even when they are on the same disk.

There was a [discussion](https://github.com/livewire/livewire/discussions/8994) for it, but since that one didn't get any response, I thought I'd open a PR

Currently, I am unsure how to test this properly, but I can investigate further with some guidance


calebporzio:
## PR Review: #9772 — [4.x] Move temporary file if disks are same

**Type**: Performance optimization (bug fix in implementation)
**Verdict**: Merge

1. When a user uploads a file in Livewire, it first gets stored on a "temporary" disk (e.g., `local`).
2. When the component calls `$this->photo->storeAs('photos', 'avatar.jpg', 'local')` — storing to the **same** disk — Livewire currently does a full copy: `readStream()` from the temp path, `put()` to the new path. The temp file stays behind until cleanup.
3. This PR makes Livewire use `move()` instead when the source and destination are the same disk and there are no extra storage options (like `visibility`). A move is essentially a rename — instant, no data copied.
4. For **different** disks (e.g., temp on `local`, destination on `s3`), the old `readStream()`/`put()` behavior is preserved because you can't rename across filesystems.

1. **Always move, even cross-disk**: Not possible — `Storage::move()` only works within the same disk.
2. **Move then delete the original**: Redundant — `move()` already removes the source. And `put()`+explicit `delete()` would be slower than `move()` for same-disk.
3. **Let the user decide**: Overengineered for what's a straightforward optimization with no behavioral tradeoff.

The PR's approach (conditional move vs copy) is the right one.

### Changes Made

**Fixed a critical bug in the condition.** The original code had:

```php
if ($this->disk === $disk && count($options) === 1 && isset($options['disk'])) {
```

This condition is **dead code** — it can never be true. `Arr::pull($options, 'disk')` on line 179 already removes the `'disk'` key from `$options` before this check. After `Arr::pull`, `$options` will never contain `'disk'`, so `isset($options['disk'])` is always `false`.

Fixed to:
```php
if ($this->disk === $disk && empty($options)) {
```

This correctly checks: same disk + no extra options (like `visibility`) remaining after extracting the disk name.

**Added two tests:**
- `test_same_disk_store_moves_file_instead_of_copying` — verifies the temp file is removed (moved, not copied) when storing to the same disk
- `test_different_disk_store_copies_file` — verifies the temp file persists (copied, not moved) when storing to a different disk

Both tests were verified to fail on `main` and pass with the fix.

### Test Results

All 57 file upload unit tests pass. CI was green on the original PR. The same-disk test correctly fails without the fix (temp file still present after `put()`).

### Code Review

- `src/Features/SupportFileUploads/TemporaryUploadedFile.php:184` — Fixed condition from dead code to working logic
- The optimization is surgical — only one conditional added around the existing `put()` call
- Edge case handled: if extra options like `visibility` are passed with same-disk, it still falls through to `put()` (correct, since `move()` doesn't accept storage options)

### Security

No security concerns. The `move()` call operates on the same Storage disk instance and respects the same access controls. No new user input is trusted. File path construction is unchanged.

### Verdict

**Merge.** The idea is sound — moving is faster than copying on the same disk. The original implementation had a fatal logic bug (dead condition) that I've fixed. Tests are now in place. 2 community thumbs-up. All CI green.

---
*Reviewed by Claude*

calebporzio:
sounds good, thanks!

PR #9773, opened 2026-01-19
fix: Target class [Livewire\Mechanisms\ComponentRegistry] does not exist.

sebacarrasco93:
This is my first PR 😁

1️⃣ I don't know if it is something is that you wanted, because I don't know if ComponentRegistry was accidentally deleted, but it is required in Livewire 4 when it throws an error when a method doesn't exists.

2️⃣ Yes, I've created fix-component-registry branch

3️⃣ It doesn't contains unrelated changes

4️⃣ Yep, it includes tests!

5️⃣ Steps to reproduce:

```bash
php artisan livewire:make my.component
```

my/component.blade.php:

```blade
<button wire:click="notExists">
  Click me
</button>
```

Current output:

```
[ReflectionException] > [BindingResolutionException]

Target class [Livewire\Mechanisms\ComponentRegistry] does not exist.
```

So, I bringed back the file.

Fixed output:

```
Unable to call component method. Public method [notExists] not found on component
```

### DIFF
--- a/src/Mechanisms/ComponentRegistry.php
+++ b/src/Mechanisms/ComponentRegistry.php
@@ -1,0 +1,217 @@
+<?php
+
+namespace Livewire\Mechanisms;
+
+use Livewire\Component;
+use Livewire\Exceptions\ComponentNotFoundException;
+
+class ComponentRegistry extends Mechanism
+{
+    protected $missingComponentResolvers = [];
+    protected $nonAliasedClasses = [];
+    protected $aliases = [];
+
+    function component($name, $class = null)
+    {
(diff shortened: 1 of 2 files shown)


joshhanley:
@sebacarrasco93 thanks for the PR! We actually remove the ComponentRegistry in v4 in favour of the new Finder class. Because of that, I'm going to close this PR. But thanks for being willing to contribute!

I believe this issue is actually coming from spatie/laravel-ignition which hasn't been updated to support Livewire v4, so I will be looking into PRing support to it today.

Thanks again!

PR #9790, opened 2026-01-20
Add `Route::macro()` return type

alloylab:
Adding the return types provides better support for larastan(phpstan), without it, you get the following phpstan error, phpstan: Cannot call method name() on mixed.


joshhanley:
@alloylab thanks!

PR #9326, opened 2025-05-12
Add release tokens for deployment invalidation

joshhanley:
This PR reintroduces the deployment invalidation feature from V2 as it will be needed with PR #9310. Original PR #4430.

PR #9795, opened 2026-01-21
Verify release token before checksum

joshhanley:
# The Scenario

After upgrading from Livewire v3.6.4 to v3.7.1, users began seeing a surge of `CorruptComponentPayloadException` errors in their error tracking systems. These errors appeared immediately after the upgrade with no code changes, affecting users who had browser tabs open from before the upgrade.

```php
Livewire\Mechanisms\HandleComponents\CorruptComponentPayloadException

Livewire encountered corrupt data when trying to hydrate a component.
Ensure that the [name, id, data] of the Livewire component wasn't tampered with between requests.
```

The errors naturally resolved after a couple of weeks as cached browser sessions expired.

# The Problem

PR #9339 removed `memo.children` from the checksum calculation:

```php
// In Checksum::generate()
unset($snapshot['memo']['children']);
```

When a user has a browser tab open from before the upgrade, their snapshot has a checksum that was computed WITH `memo.children` included. When this old snapshot is sent to the new Livewire version, the checksum verification fails because the new code strips `memo.children` before computing the comparison checksum.

PR #9326 introduced release tokens specifically to handle these version mismatch scenarios gracefully by throwing a 419 `LivewireReleaseTokenMismatchException` that prompts the user to refresh. However, the release token verification was implemented on the `snapshot-verified` hook, which fires AFTER checksum verification. This meant the checksum mismatch error was thrown first, before the release token could be checked.

# The Solution

Move release token verification from the `snapshot-verified` hook to the `checksum.verify` hook, which fires BEFORE the checksum comparison:

```php
// Before (fires after checksum passes)
on('snapshot-verified', function ($snapshot) {
    ReleaseToken::verify($snapshot);
});

// After (fires before checksum comparison)
on('checksum.verify', function ($checksum, $snapshot) {
    ReleaseToken::verify($snapshot);
});
```

Now when a stale browser session sends a request:
1. `Checksum::verify()` is called
2. The `checksum.verify` hook fires first
3. `ReleaseToken::verify()` detects the missing/invalid release token
4. `LivewireReleaseTokenMismatchException` is thrown (419 status)
5. The user sees a clear message prompting them to refresh

This ensures release tokens work as intended, catching stale sessions before the confusing `CorruptComponentPayloadException` can be thrown.

Fixes: #9663

### DIFF
--- a/src/Features/SupportReleaseTokens/SupportReleaseTokens.php
+++ b/src/Features/SupportReleaseTokens/SupportReleaseTokens.php
@@ -14,9 +14,9 @@ class SupportReleaseTokens extends ComponentHook
             $context->addMemo('release', ReleaseToken::generate($component));
         });
 
-        // Use `snapshot-verified` to run the check before any component properties are hydrated
-        // but after the snapshot has been verified to ensure it hasn't been tampered with...
-        on('snapshot-verified', function ($snapshot) {
+        // Verify the release token before verifying the snapshot checksum, as we don't want to trigger a checksum
+        // failure if the release token doesn't match as there may be intentional changes in the snapshot...
+        on('checksum.verify', function ($checksum, $snapshot) {
             ReleaseToken::verify($snapshot);
         });
     }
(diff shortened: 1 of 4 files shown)


joshhanley:
This should have been targeted at 3.x so closing.

PR #9798, opened 2026-01-21
Support wire:click.renderless modifier within islands

joshhanley:
# The Scenario

When using `wire:island.append` for patterns like infinite scrolling, clicking an action button inside the appended content (e.g., a bookmark button) causes the island to re-render, losing all the appended content.

```php
class PostFeed extends Component
{
    public $posts;

    public function mount()
    {
        $this->posts = Post::take(10)->get();
    }

    public function loadMore()
    {
        $this->posts = Post::skip(count($this->posts))->take(10)->get();
    }

    public function bookmark($postId)
    {
        // Toggle bookmark...
    }

    public function render()
    {
        return <<<'HTML'
        <div>
            @island(name: 'posts')
                @foreach($posts as $post)
                    <div>
                        {{ $post->title }}
                        <button wire:click="bookmark({{ $post->id }})" wire:island="posts">
                            Bookmark
                        </button>
                    </div>
                @endforeach
            @endisland

            <button wire:click="loadMore" wire:island.append="posts">Load More</button>
        </div>
        HTML;
    }
}
```

When "Load More" is clicked multiple times, additional posts are appended. However, clicking "Bookmark" on any post causes the entire island to re-render with only the first page of posts, losing pages 2+.

The workaround is to use `wire:click.renderless` to prevent the re-render, but this modifier was not being respected within islands.

# The Problem

The `wire:click.renderless` modifier correctly sets `metadata.renderless = true` in JavaScript, but `SupportIslands::call()` was not checking for this metadata flag. It only checked for the `#[Renderless]` PHP attribute (via `skipIslandsRender` in the component store).

# The Solution

Added a check for the `renderless` metadata flag in `SupportIslands::call()` before rendering the island. When the `.renderless` modifier is used on a `wire:click` (or similar) directive, the island rendering is now skipped, matching the behaviour of the `#[Renderless]` attribute.

```php
// Support `Wire:click.renderless`...
if ($metadata['renderless'] ?? false) {
    return;
}
```

Fixes: https://github.com/livewire/livewire/discussions/9786

### DIFF
--- a/src/Features/SupportIslands/BrowserTest.php
+++ b/src/Features/SupportIslands/BrowserTest.php
@@ -760,6 +760,39 @@ class BrowserTest extends BrowserTestCase
             ;
     }
 
+    public function test_renderless_modifier_skips_island_render()
+    {
+        Livewire::visit([new class extends \Livewire\Component {
+            public $count = 0;
+
+            public function increment()
+            {
+                $this->count++;
+            }
+
+            public function render() {
+                return <<<'HTML'
+                <div>
+                    @island(name: 'foo')
+                        <div dusk="island-count">Count: {{ $count }}</div>
+                    @endisland
+
+                    <button type="button" wire:click="increment" dusk="increment" wire:island="foo">Increment</button>
+                    <button type="button" wire:click.renderless="increment" dusk="increment-renderless" wire:island="foo">Increment Renderless</button>
+                </div>
+                HTML;
+            }
+        }])
+            ->assertSeeIn('@island-count', 'Count: 0')
+            ->waitForLivewire()->click('@increment-renderless')
+            // The count was incremented server-side but the island should NOT re-render...
+            ->assertSeeIn('@island-count', 'Count: 0')
+            ->waitForLivewire()->click('@increment')
+            // Now the island should show the updated count (including the renderless increment)...
+            ->assertSeeIn('@island-count', 'Count: 2')
+            ;
+    }
+
     public function test_more_than_ten_islands_using_single_file_component()
     {
         Livewire::visit('twelve-islands')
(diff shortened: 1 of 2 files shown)


ghabriel25:
@joshhanley I just tested `$this->skipRender()` not working too

### Workaround
```php
if (! ($this instanceof Bookmarks)) {
    store($this)->set('skipIslandsRender', true);
    $this->skipRender();
}
```
The reason behind this was I'm using the same component for different page (Dashboard, Published, Drafted, Bookmarks etc) and the bookmark toggle also shared for all of those components. I can dispatch an event to refresh Bookmarks page but it doesnt work with island (I dont know why but it only works for first page, not after that). Thats why I need to skip render on this specific page.

Its like applying `#[Renderless]` attribute but with condition.

This is just my point of view, maybe `$this->skipRender()` should check if `skipIslandsRender` is set or not
```php
function skipRender($html = null)
{
    if (store($this)->has('forceRender')) {
        return;
    }
    
    // or just set it to true without checking
    if (! store($this)->has('skipIslandsRender')) {
        store($this)->set('skipIslandsRender', true);
    }

    store($this)->set('skipRender', $html ?: true);
}

class BaseRenderless extends LivewireAttribute
{
    function call()
    {
        store($this->component)->set('skipIslandsRender', true); // and remove this

        $this->component->skipRender();
    }
}
```
### Update
Just watch Caleb's video on Laracast about little hack for this
```php
#[Renderless]
publif function toggleBookmark(bool $bookmarked)
{
    ...
    if (! $bookmarked && $this instanceof Bookmarks) {
        $this->renderIsland('posts');
    }
}
```
This will force Bookmarks page to render its island.

nicolagianelli:
@joshhanley 
Shouldn't `skipRender()` work even when executed from an island?

Just to know if it's intentional or not, because now we should use:
```
store($this->component)->set('skipIslandsRender', true);
$this->component->skipRender();
```
When we want to skip render conditionally from an island, correct?

PR #9800, opened 2026-01-21
[4.x] Configurable cache path

osmuhin:
## Description

This PR introduces a new `cache_path` configuration option that allows explicitly configuring where Livewire stores its compiled component views.

### Why this change is needed

Previously, Livewire always relied on an implicit path:

```php
storage_path('framework/views/livewire')
```

This path was hardcoded at the point of `CacheManager` instantiation, making it impossible to:

* Relocate Livewire’s compiled view cache
* Isolate Livewire cache from other framework caches
* Adapt Livewire to custom filesystem layouts (e.g. read-only containers, shared volumes, multi-tenant setups)

By exposing `cache_path` as a configuration option, Livewire becomes more flexible without changing any default behavior.

### What changed

A new config value was added:

```php
'cache_path' => storage_path('framework/views/livewire'),
```

And it is now used when initializing the compiler:

```php
$this->app->singleton('livewire.compiler', function () {
    return new Compiler(
        new CacheManager(
            config('livewire.cache_path', storage_path('framework/views/livewire'))
        )
    );
});
```

If the option is not set, Livewire falls back to the previous hardcoded path, ensuring full backward compatibility.

### Additional notes

* No breaking changes
* Default behavior remains unchanged
* This change is fully backward compatible

This makes Livewire’s caching behavior explicit, configurable, and easier to reason about for advanced setups.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -118,6 +118,23 @@ return [
     'view_path' => resource_path('views/livewire'),
 
     /*
+    |--------------------------------------------------------------------------
+    | Cache Path
+    |--------------------------------------------------------------------------
+    |
+    | This value is used to specify where Livewire stores compiled component
+    | views and other cached artifacts generated during the rendering process.
+    | These files are managed internally by Livewire and should not be edited
+    | manually.
+    |
+    | Changing this path may be useful if you want to customize storage layout
+    | or isolate Livewire cache from other framework caches.
+    |
+    */
+
+    'cache_path' => storage_path('framework/views/livewire'),
+
+    /*
     |---------------------------------------------------------------------------
     | Temporary File Uploads
     |---------------------------------------------------------------------------
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -43,7 +43,7 @@ class LivewireServiceProvider extends \Illuminate\Support\ServiceProvider
         $this->app->singleton('livewire.compiler', function () {
             return new Compiler(
                 new CacheManager(
-                    storage_path('framework/views/livewire')
+                    config('livewire.cache_path', storage_path('framework/views/livewire'))
                 )
             );
         });


calebporzio:
## PR Review: #9800 — [4.x] Configurable cache path

**Type**: Feature
**Verdict**: Close

1. Livewire compiles single-file components into PHP classes and Blade views, stored in `storage/framework/views/livewire/`.
2. That path is passed to `CacheManager` when the `livewire.compiler` singleton is created in the service provider.
3. This PR adds a new `livewire.cache_path` config option so users can change that path via config instead of rebinding the singleton.

1. **Rebind the singleton in a service provider** — This already works today. Any Laravel dev who needs a custom path can bind their own `livewire.compiler` in `AppServiceProvider`. This is the standard Laravel pattern for customizing framework internals and requires zero changes to Livewire.
2. **Environment variable** — Could use an env var, but that still adds surface area for a near-zero demand feature.
3. **Do nothing** — The default path works for virtually all deployments. Multi-tenant or read-only container setups that need a different path are advanced enough to rebind a singleton.

### Changes Made

No changes made.

### Test Results

No tests included in the PR. No tests to run.

### Code Review

- `config/livewire.php:120-135`: The 13-line comment block is disproportionately verbose for a single config value. Other entries in the file are more concise.
- `src/LivewireServiceProvider.php:46`: The actual code change is fine — `config('livewire.cache_path', storage_path('framework/views/livewire'))` with a fallback. No issues with the implementation itself.
- No test verifying that the config is respected.

### Security

### Verdict

**Close.** Zero community demand (0 reactions, no linked issue). The problem this solves — customizing the compiled view cache path — is already achievable by rebinding the `livewire.compiler` singleton, which is the standard Laravel approach. Every config key is a maintenance commitment, and this one doesn't clear the bar. If real demand surfaces (multi-tenant users requesting it, issues filed), it's trivial to add later.

---
*Reviewed by Claude*

calebporzio:
yeah, going to punt on this for now. thanks though!

PR #9806, opened 2026-01-21
Prevent HTML modal from showing after 419 session expired dialog

cyppe:
# Description

When a 419 (session expired) error occurs, the error handler shows a `confirm()` dialog. However, `showHtmlModal()` is still called afterwards, displaying the HTML error page in a modal overlay.

**Before:**
```javascript
if (response.status === 419) {
    confirm('This page has expired...') && window.location.reload()
}
// No return - continues to showHtmlModal()
showHtmlModal(responseBody)
```

**After:**
```javascript
if (response.status === 419) {
    confirm('This page has expired...') && window.location.reload()
    return  // Exit early
}
showHtmlModal(responseBody)
```

## Problem

Users see both:
1. The confirm dialog ("This page has expired. Would you like to refresh?")
2. The HTML error modal (showing the full error page)

## Solution

Added `return` after the 419 handling block to prevent `showHtmlModal()` from being called.

## Testing

1. Let a session expire (or manually clear the session cookie)
2. Interact with a Livewire component
3. **Before fix:** Confirm dialog appears, then HTML modal also appears
4. **After fix:** Only confirm dialog appears


joshhanley:
@cyppe thanks! FYI - we're no longer committing dist files to PRs as we build them as part of our GHA test runner.

cyppe:
Ah ok, sorry!  I peeked into https://livewire.laravel.com/docs/4.x/contribution-guide but maybe it also needs a PR ?

joshhanley:
@cyppe no worries! Yep might be worth adding a note there

PR #9809, opened 2026-01-21
[4.x] Fix form submit using readonly on input types not valid

nicolagianelli:
**Summary**
This PR fixes an issue where inputs that do not support the readonly attribute were not being correctly handled during form processing.

**Changes**
While investigating a bug regarding range input not being disabled while submitting, I noted that the current implementation only swaps readonly for disabled on checkbox and radio types. However, according to [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly), several other input types do not support the readonly attribute, including range, color...

I have updated the logic to ensure that all input types lacking readonly support are correctly set to disabled instead.


joshhanley:
@nicolagianelli thanks for the PR! Can you please add tests? Thanks!

nicolagianelli:
> @nicolagianelli thanks for the PR! Can you please add tests? Thanks!

Before adding the tests I checked what InertiaJS does with form submission to disable it and saw they are using the native form **inert** attribute.

What do you think?
Should we continue to manually handle inputs/submit button or change the PR to use the native inert attribute directly on form?

calebporzio:
## PR Review: #9809 — [4.x] Fix form submit using readonly on input types not valid

**Type**: Bug fix
**Verdict**: Merge

1. When a form with `wire:submit` is submitted, Livewire disables form inputs during the request to prevent double-submission
2. For text-like inputs (`<input type="text">`, `<textarea>`), it sets `readonly` — this is better UX because users can still select/copy text and the element doesn't visually gray out
3. For inputs that don't support `readonly` (checkboxes, radios), it sets `disabled` instead
4. **The bug**: the original code only handled `checkbox` and `radio` as types needing `disabled`. But per the HTML spec, many other input types also don't support `readonly`: `range`, `color`, `file`, `submit`, `image`, `reset`, `button`, `hidden`
5. **The consequence**: a `<input type="range">` inside a `wire:submit` form gets `readonly` set during submission, which the browser ignores entirely — the slider stays interactive during the request
6. **The fix**: expands the check from just `checkbox`/`radio` to the full list of input types that don't support `readonly`

1. **Use `disabled` for everything** — simpler, but worse UX for text inputs (can't select/copy, grays out). Current `readonly` approach for text inputs is correct.
2. **Use the `inert` attribute on the whole form** — the contributor mentioned this in comments (Inertia.js does this). Simpler implementation, but `inert` has worse browser support (Safari 15.5+) and removes the element from the accessibility tree entirely, which is heavier-handed than needed. Could be a future improvement but the current surgical approach is better for now.
3. **Allowlist types that DO support readonly** — inverse approach. Both work; the PR's approach (blocklist of types that don't) is fine since the list is well-defined by the HTML spec and unlikely to change.

### Changes Made

- Fixed variable name grammar: `inputTypesThatDontSupportsReadonly` → `inputTypesThatDontSupportReadonly`
- Removed trailing whitespace on line 60
- Added browser tests in new `src/Features/SupportDisablingFormsDuringRequest/BrowserTest.php`:
  - `test_range_input_is_disabled_not_readonly_during_form_submission` — verifies range gets `disabled`, text gets `readonly`, both cleared after response
  - `test_color_input_is_disabled_not_readonly_during_form_submission` — verifies color gets `disabled`, cleared after response

### Test Results

- 2 browser tests: **PASS** (12 assertions)
- Regression verified: both tests **FAIL** on `main` without the fix (range/color inputs get `readonly` which browser ignores)
- CI status: all checks pass except one flaky browser test (Browser 3 - L12/PHP 8.5) which is unrelated to this PR

### Code Review

- `js/features/supportDisablingFormsDuringRequest.js:59-62`: The type list is comprehensive and matches [MDN's readonly spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly). Good.
- Including `hidden` in the list is technically correct (hidden doesn't support readonly) and harmless — a hidden input doesn't need user interaction protection, but marking it disabled has no negative effects since Livewire uses JS for state management, not native form submission.
- The fix is surgical — only changes the condition in `shouldMarkDisabled()`, no other code touched.

### Security

### Verdict

Clean, correct bug fix with a well-defined root cause. The contributor's approach is the right one — expanding the blocklist to match the HTML spec. I've added tests and fixed minor style issues. Ready to merge.

---
*Reviewed by Claude*

calebporzio:
Thanks!

PR #9810, opened 2026-01-21
Fix getUpdateUri() failing when routes are cached

joshhanley:
# The Scenario

When using PestPHP's `livewire()` helper to test Livewire components (e.g., Filament resources), tests fail with an error after upgrading from v3.7.4 to v3.7.5:

```
Error: Call to a member function getName() on null

at vendor/livewire/livewire/src/LivewireManager.php:136
```

This occurs when routes are cached, which is common in parallel testing scenarios using Laravel's `WithCachedRoutes` trait.

```php
// Test that was working in v3.7.4 but fails in v3.7.5
livewire(WidgetResource\Pages\ListWidgets::class)
    ->assertSuccessful();
```

# The Problem

Commit `9abb4cf2` added a check in `HandleRequests::boot()` to prevent duplicate route registration when routes are cached:

```php
        return Route::post(EndpointResolver::updatePath(), $handle)->middleware('web');
    });
}
```

This correctly prevents duplicate registration, but has a side effect: when routes ARE cached and loaded by Laravel, the `livewire.update` route already exists in the router, so `setUpdateRoute()` is never called and `$this->updateRoute` remains `null`.

When `getUpdateUri()` is subsequently called (e.g., by Livewire's testing helpers), it tries to access `$this->updateRoute->getName()` which fails because `$this->updateRoute` is `null`.

# The Solution

Modified `getUpdateUri()` to fall back to finding the route from the router when `$this->updateRoute` is null:

```php
function getUpdateUri()
{
    // When routes are cached, $this->updateRoute may be null because
    // setUpdateRoute() was never called (the route already existed).
    // In this case, find the route from the router.

    return (string) str(
    )->start('/');
}
```

Also refactored `updateRouteExists()` to use the new `findUpdateRoute()` method to avoid code duplication.

Fixes: #9807


joshhanley:
Targeted 3.x instead of main, so closing.

PR #9813, opened 2026-01-22
[4.x] Cache resolved routes to prevent binding resolution issues for bundled child components

joshhanley:
# The Scenario

When a Livewire page component with route model binding has child components using `#[Reactive]` or `#[Modelable]`, deleting the route-bound model causes a 404 error. Clicking "Delete Post" in the example below results in a 404 instead of showing the updated title.

Screenshot 2026-02-05 at 12 56 04PM

```php
<?php
// Route: /posts/{post}

use App\Models\Post;
use Livewire\Attributes\Reactive;
use Livewire\Component;

new class extends Component {
    public Post $post;
    public $title = 'My Post';

    public function delete()
    {
        $this->post->delete();
        $this->title = 'Deleted';
    }
};
?>

<div>
    <livewire:child-component :$title />
    <button wire:click="delete">Delete Post</button>
</div>
```

```php
<?php

use Livewire\Attributes\Reactive;
use Livewire\Component;

new class extends Component {
    #[Reactive]
    public $title;
};
?>

<div>{{ $title }}</div>
```

# The Problem

During a Livewire update request Livewire creates a fake request and resolves the original route so any PersistentMiddleware, such as `SubstituteBindings`, can be reapplied to the current request,

When child components use `#[Reactive]` or `#[Modelable]`, they're bundled into the same HTTP request as the parent. For each component in the request, a fake request is created and middleware are applied.

Both parent and child components store the same original path in their snapshot (e.g., `/posts/1`), so both try to resolve that same route and bindings as part of their fake request.

The problem arises if the parent has resolved the route-bound model and then deleted it. When the child tries to resolve the route-bound model, as part of its fake request, the model can't be found and a 404 is thrown.

# The Solution

The solution is to allow the parent to resolve the route and any bindings (such as resolved model instances) and then cache that route in the PesistentMiddleware feature. Then when any child components go to resolve the route, it will check the cache first and use the already resolved route, if available.

Laravel's binding logic checks `$parameterValue instanceof UrlRoutable` and skips re-resolution for already-resolved models, that way the cached route's bindings are preserved even after the model is deleted.

Screenshot 2026-02-05 at 12 54 37PM

Fixes #9804

### DIFF
--- a/src/Features/SupportPageComponents/SupportPageComponents.php
+++ b/src/Features/SupportPageComponents/SupportPageComponents.php
@@ -207,19 +207,6 @@ class SupportPageComponents extends ComponentHook
             app('router')->substituteImplicitBindingsUsing(function ($container, $route, $default) {
                 // If the current route is a Livewire page component...
                 if ($componentClass = static::routeActionIsAPageComponent($route)) {
-                    // If we're processing a child component (not the page component),
-                    // skip binding resolution - child components don't need route bindings
-                    // and the models are already hydrated from the snapshot.
-                    $currentComponentName = \Livewire\Mechanisms\PersistentMiddleware\PersistentMiddleware::$currentComponentName;
-
-                    if ($currentComponentName) {
-                        $currentComponentClass = app('livewire.factory')->resolveComponentClass($currentComponentName);
-
-                        if ($currentComponentClass !== $componentClass) {
-                            return;
-                        }
-                    }
-
                     // Resolve and set all page component parameters to the current route...
                     (new \Livewire\Drawer\ImplicitRouteBinding($container))
                         ->resolveAllParameters($route, new $componentClass);
--- a/src/Features/SupportPageComponents/SupportPageComponents.php
+++ b/src/Features/SupportPageComponents/SupportPageComponents.php
@@ -207,6 +207,19 @@ class SupportPageComponents extends ComponentHook
             app('router')->substituteImplicitBindingsUsing(function ($container, $route, $default) {
                 // If the current route is a Livewire page component...
                 if ($componentClass = static::routeActionIsAPageComponent($route)) {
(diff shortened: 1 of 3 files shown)


calebporzio:
little more thinking on this

joshhanley:
@calebporzio I've reworked this to cache resolved routes instead, so the route binding doesn't need to get re-resolved for every component. I've updated the PR description.

stayallive:
I'm not really sure how to debug this but this commit / 4.1.3 seems to have broken parts of my application. I see it has something to do with a route binding considering these queries:

```sql
select * from `teams` where `slug` = 'aenqufbo' limit 1
select * from `teams` where `slug` = 'aenqufbo' limit 1
select * from `teams` where `slug` = 'Personal' limit 1 # wrong
​select * from `teams` where `slug` = 'Personal' limit 1 # wrong
select * from `teams` where `slug` = 'Personal' limit 1 # wrong
```

When I go back to 4.1.2 I can see the same 5 queries but with the correct slug. Upgrading to 4.1.3 or 4.1.4 breaks it again. I have a model setup for the binding resolution:

```php
$router->model('team_hint', Team::class);
```

I tracked it down to the `resolveRouteBinding()` for that model returns a model (as it should) but somehow this is cast to a string somewhere ("Personal" is the `__toString()` of the bound model). I'm not sure how this change causes that but it seems to be doing it.

Hope this report helps unlock some knowledge about the change that I'm missing that can cause this 😄 

(and yes; reverting this specific change by commenting out `$this->cachedRoutes[$cacheKey] = $route;` fixes the problem)

joshhanley:
@stayallive please open a bug discussion with code and reproduction steps or a small example repo that demonstrates the issue, so we can investigate. Thanks!

PR #9817, opened 2026-01-22
Fix race condition in cache directory creation

cyppe:
This PR fixes a race condition in `CacheManager` that can cause `mkdir(): File exists` errors during concurrent Livewire requests.

## Problem

When multiple Livewire requests concurrently compile components (e.g., after `view:clear`), the directory creation can fail:

1. Request A checks directory doesn't exist
2. Request B checks directory doesn't exist
3. Request B creates directory - succeeds
4. Request A tries to create directory - throws `mkdir(): File exists`

## Solution

Pass the `$force = true` parameter to `File::makeDirectory()`. This uses Laravel's built-in error suppression for mkdir operations, which is the intended way to handle this scenario.

```php
// Before
File::ensureDirectoryExists($this->cacheDirectory . '/classes');

// After
File::makeDirectory($this->cacheDirectory . '/classes', 0777, true, true);
```

The fourth parameter (`$force`) wraps the mkdir call with `@` to suppress errors when the directory already exists.

## Related

This is the same underlying issue addressed in https://github.com/laravel/framework/pull/58460 for `Filesystem::ensureDirectoryExists()`. If that fix is merged, these calls could be simplified back to `File::ensureDirectoryExists()`. Until then, this approach ensures Livewire works correctly regardless of the Laravel framework version.

## Impact

Backwards compatible. Directory creation behavior is unchanged, but concurrent requests no longer cause errors.

### DIFF
--- a/src/Compiler/CacheManager.php
+++ b/src/Compiler/CacheManager.php
@@ -114,7 +114,7 @@ class CacheManager
         $classPath = $this->getClassPath($sourcePath);
 
         $this->ensureCacheDirectoryExists();
-        File::ensureDirectoryExists($this->cacheDirectory . '/classes');
+        File::makeDirectory($this->cacheDirectory . '/classes', 0777, true, true);
 
         File::put($classPath, $contents);
     }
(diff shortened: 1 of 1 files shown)


joshhanley:
@cyppe thanks!

PR #9818, opened 2026-01-22
Fix missingComponentResolvers being emptied in Laravel Octane

sankarsuda:
### Problem
When running Livewire applications with Laravel Octane, the missingComponentResolvers array in the Factory class becomes empty after a few requests, causing component resolution failures. This issue does not occur with PHP-FPM/Nginx.
### Root Cause
The Livewire Factory is registered as a scoped instance in the service container, which means it gets flushed between requests by Octane's FlushTemporaryContainerInstances listener. This causes the missingComponentResolvers array to be reset to empty on each new request, even though these resolvers should persist throughout the application lifecycle.
### Solution
Modified the Factory class to persist missing component resolvers across Factory instances using a static property that survives Octane's container flushing:
### Changes Made

1. Added static storage property:

   ```protected static $persistentMissingComponentResolvers = [];```

2. Modified constructor to restore resolvers:

  ```   
public function __construct(
       protected Finder $finder,
       protected Compiler $compiler,
   ) {
       // Restore persistent missing component resolvers (important for Octane)
       $this->missingComponentResolvers = static::$persistentMissingComponentResolvers;
   }
```

3. Updated resolver registration to persist:

```
   public function resolveMissingComponent($resolver): void
   {
       $this->missingComponentResolvers[] = $resolver;
       static::$persistentMissingComponentResolvers[] = $resolver;
   }
```

### Why This Works

- **Static persistence:** Static properties persist across requests in Octane, unlike instance properties that get flushed
- **Instance restoration:** Each new Factory instance automatically restores the resolvers from static storage
- **Backward compatibility:** The fix maintains all existing functionality while solving the Octane-specific issue
- **Memory efficient:** Uses minimal additional memory since resolvers are typically registered once during bootstrap

**Impact**
This fix ensures that missing component resolvers registered during application bootstrap (typically in service providers) will now persist across requests in Laravel Octane, matching the behavior of PHP-FPM/Nginx environments.

### DIFF
--- a/src/Factory/Factory.php
+++ b/src/Factory/Factory.php
@@ -2,13 +2,15 @@
 
 namespace Livewire\Factory;
 
-use Livewire\Exceptions\ComponentNotFoundException;
 use Livewire\Compiler\Compiler;
-use Livewire\Finder\Finder;
 use Livewire\Component;
+use Livewire\Exceptions\ComponentNotFoundException;
+use Livewire\Finder\Finder;
 
 class Factory
 {
+    protected static $persistentMissingComponentResolvers = [];
+
     protected $missingComponentResolvers = [];
 
     protected $resolvedComponentCache = [];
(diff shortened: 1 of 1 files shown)


joshhanley:
@sankarsuda thanks for the PR! I've also been investigating this issue and found that the factory was a scoped singleton instead of a standard singleton, which was causing these issues on subsequent Octane requests. I've submitted a PR #9820 which changes that and fixes another bug I've found during my testing. Because of that, I'm going to close this PR. Thanks for being willing to contribute.

PR #9827, opened 2026-01-24
Fix persist directive docs

webard:
https://livewire.laravel.com/docs/4.x/navigate#preserving-scroll-position tells about `wire:navigate:scroll`, but in https://livewire.laravel.com/docs/4.x/directive-persist#preserving-scroll-position there is a `wire:scroll` attribute mentioned. First one is correct.


joshhanley:
@webard thanks!

PR #9832, opened 2026-01-26
Control wire:model client-side sync timing with modifiers

calebporzio:
## Summary

**This is a breaking change.**

`wire:model` modifiers like `.blur`, `.change`, and `.enter` now control **client-side** sync timing, not just network timing. The `.live` modifier acts as a delimiter between client-side and network modifiers.

## Breaking Change

**Before (v4.0):**
```blade
<input wire:model.blur="title">
```
- Client-side state (`$wire.title`) updates immediately as user types
- Network request fires on blur

**After (this PR):**
```blade
<input wire:model.blur="title">
```
- Client-side state updates on blur
- No network request (no `.live`)

**Migration:**
```blade
<!-- Old behavior: add .live before .blur -->
<input wire:model.live.blur="title">
```

> **Note:** `wire:model.lazy` remains backwards compatible—no changes needed.

## Why This Breaking Change?

Currently, there's no way to control **when client-side state syncs**. Alpine's `x-model` supports modifiers like `.enter`, `.blur`, and `.change` to control this, but Livewire's `wire:model` has always synced client-side state immediately—modifiers only affected network timing.

With Livewire 4's more powerful frontend (`wire:text`, `wire:show`, `$wire` reactivity), this limitation matters more than ever.

**Example use case:** Imagine width/height inputs for resizing an element. With the old behavior, as a user types "1200" into the width field, the UI reacts to "1", then "12", then "120", then "1200"—jarring and broken.

Now you can do:
```blade
<input wire:model.blur="width">
```
The value only updates when the user finishes typing and tabs away. Clean.

Or sync on Enter key:
```blade
<input wire:model.enter="search">
```

Or combine them:
```blade
<input wire:model.blur.enter="title">
```

## The New Mental Model

Modifiers **before** `.live` control client-side sync timing. Modifiers **after** `.live` control network timing.

```blade
wire:model.blur.live.debounce.500ms
         ^^^^      ^^^^^^^^^^^^^^
         │         └── Network: debounced 500ms
         └── Client-side: syncs on blur
```

## Why Now?

Livewire 4 is two weeks old. If we don't make this change now, we're locked into the current behavior for potentially years. The migration is simple (add `.live` before your modifier), and this unlocks an entire class of UI patterns that weren't possible before.

### DIFF
--- a/docs/alpine.md
+++ b/docs/alpine.md
@@ -118,7 +118,7 @@ Here is an example of using Alpine to listen for a "blur" event on an input and 
 </form>
 ```
 
-Typically, you would just use `wire:model.blur="title"` in this situation, however, it's helpful for demonstration purposes how you can achieve this using Alpine.
+Typically, you would just use `wire:model.live.blur="title"` in this situation, however, it's helpful for demonstration purposes how you can achieve this using Alpine.
 
 #### Passing parameters
 
--- a/docs/attribute-modelable.md
+++ b/docs/attribute-modelable.md
@@ -101,7 +101,7 @@ The parent can use wire:model modifiers, and they'll work as expected:
 <livewire:todo-input wire:model.live="todo" />
 
 {{-- Update on blur --}}
-<livewire:todo-input wire:model.blur="todo" />
+<livewire:todo-input wire:model.live.blur="todo" />
 
 {{-- Debounce updates --}}
 <livewire:todo-input wire:model.live.debounce.500ms="todo" />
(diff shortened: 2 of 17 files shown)


awcodes:
Why not leave everything as is and just add a new .client modifier so it doesn't have to be a breaking change?

calebporzio:
@claude - Livewire 4 was released about two weeks ago. There is a change I'd really like to make so that I'm not stuck with a limiting API (or awkward alternative), but it would cause a breaking change in one instance: folks using `wire:model.blur`. (which in hindsight now even feels like a bad API because it doesn't express that it's "live" in the sense that it will ever send to the server)

so let me know:
A) Are we in a reasonable window to still make a breaking change like this?
B) Just in case, what a few other options short of literally tagging a new major version just for this lol

PR #9833, opened 2026-01-26
[4.x] Fix Race Condition in Component Cache Writes

cyppe:
## The Problem

When multiple requests hit an uncached Livewire component at the same time, you get:

```
TypeError: Cannot use "::class" on int
```

This happens after fresh deployments when bots or users hit your site simultaneously.

## Why It Happens

In `CacheManager.php`, all write methods used `File::put()`:

```php
File::put($classPath, $contents);
```

This is not atomic. When two requests try to compile the same component:

1. Request A starts writing the file
2. Request B overwrites it mid-write
3. Request A reads the half-written file
4. PHP returns `1` instead of the class instance
5. `$instance::class` crashes

## The Fix

Use `File::replace()` which writes to a temp file first, then renames atomically:

```php
File::replace($classPath, $contents);
```

Rename operations are atomic on POSIX systems - the file is either fully there or not.

This fix applies to all cache write methods:
- `writeClassFile()` - the critical one causing the crash
- `writeViewFile()`
- `writeScriptFile()`
- `writeStyleFile()`
- `writeGlobalStyleFile()`
- `writePlaceholderFile()`
- `writeIslandFile()`

## Real Example

This error occurred in production with a footer component:

```
Cannot use "::class" on int (View: footer.blade.php)
```

The footer had `<livewire:footer-links />`. After deployment, Googlebot hit multiple pages simultaneously, triggering the race condition.

### DIFF
--- a/src/Compiler/CacheManager.php
+++ b/src/Compiler/CacheManager.php
@@ -116,7 +116,9 @@ class CacheManager
         $this->ensureCacheDirectoryExists();
         File::makeDirectory($this->cacheDirectory . '/classes', 0777, true, true);
 
-        File::put($classPath, $contents);
+        // Use atomic write to prevent race conditions when multiple
+        // concurrent requests try to compile the same component.
+        File::replace($classPath, $contents);
     }
 
     public function writeViewFile(string $sourcePath, string $contents): void
@@ -126,7 +128,7 @@ class CacheManager
         $this->ensureCacheDirectoryExists();
         File::makeDirectory($this->cacheDirectory . '/views', 0777, true, true);
 
-        File::put($viewPath, $contents);
+        File::replace($viewPath, $contents);
 
         $this->mutateFileModificationTime($viewPath);
     }
@@ -138,7 +140,7 @@ class CacheManager
         $this->ensureCacheDirectoryExists();
         File::makeDirectory($this->cacheDirectory . '/scripts', 0777, true, true);
 
-        File::put($scriptPath, $contents);
+        File::replace($scriptPath, $contents);
     }
 
     public function writeStyleFile(string $sourcePath, string $contents): void
@@ -148,7 +150,7 @@ class CacheManager
         $this->ensureCacheDirectoryExists();
         File::makeDirectory($this->cacheDirectory . '/styles', 0777, true, true);
 
-        File::put($stylePath, $contents);
+        File::replace($stylePath, $contents);
     }
 
     public function writeGlobalStyleFile(string $sourcePath, string $contents): void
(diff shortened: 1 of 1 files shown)


calebporzio:
## PR Review: #9833 — [4.x] Fix Race Condition in Component Cache Writes

**Type**: Bug fix
**Verdict**: Merge

1. Livewire compiles components into PHP class files and caches them to disk using `File::put()`.
2. `File::put()` wraps `file_put_contents()` — it writes directly to the target file, byte by byte.
3. When two concurrent requests hit an uncached component (e.g. Googlebot swarming after a deploy), both try to write the same class cache file simultaneously.
4. Request A starts writing. Request B overwrites mid-write. Request A then `require`s the half-written file.
5. PHP's `require` on a truncated/corrupt file returns `1` (int) instead of the expected class instance.
6. Line 60: `$instance::class` crashes with `TypeError: Cannot use "::class" on int`.

The fix: swap `File::put()` → `File::replace()`. Laravel's `File::replace()` writes to a temp file first, then atomically renames it into place. `rename()` is atomic on POSIX — readers always see either the old complete file or the new complete file, never a partial write.

1. **`File::put()` with `LOCK_EX`** — Would prevent concurrent writes, but `LOCK_EX` with `file_put_contents` is advisory on many systems and doesn't protect readers. Readers can still read a partially-written file.
2. **Locking with `flock()` on both read and write paths** — Would work but requires modifying `getClassName()` too and adds complexity. Overkill when atomic rename solves it cleanly.
3. **`File::replace()` (this PR)** — The standard atomic write pattern. Simplest, most robust, zero additional complexity. Clear winner.

### Changes Made

### Test Results

- No test files in the diff (race conditions are inherently non-deterministic and hard to unit test reliably).
- A mock-based test asserting `File::replace()` is called would be tautological — it would test the implementation, not the behavior.
- **CI: All 22 checks pass** (Unit, Browser, Legacy across L11/L12, PHP 8.4/8.5).

### Code Review

- `src/Compiler/CacheManager.php`: All 7 write methods correctly updated from `File::put()` → `File::replace()`. The change is consistent and complete.
- The comment on `writeClassFile()` (lines 119-120) explains the "why" for the most critical case. The other methods don't need separate comments — the pattern is established.
- `mutateFileModificationTime()` (called after some writes) works correctly with `File::replace()` — `rename()` sets a fresh mtime, then `touch()` adjusts it.
- No built assets in diff. No JS changes. No style issues.

### Security

No security concerns identified. This is a safer pattern than what existed before.

### Verdict

Clean, minimal, correct fix for a real production race condition. The contributor's diagnosis is accurate and independently verified. `File::replace()` is the right tool — it's Laravel's built-in atomic write primitive. Zero regression risk (same API surface, just atomic). No test needed — the bug is a timing issue that can't be reliably unit-tested, and the fix is trivially correct (one well-known API swap). All CI passes. Merge.

---
*Reviewed by Claude*

calebporzio:
great. thanks!

PR #9842, opened 2026-01-27
[4.x] Fix duplicate message error when event listeners have modelable children

joshhanley:
# The Scenario

When dispatching an event that multiple components in a hierarchy listen to, and one of those components has a nested child with `wire:model`, an error is thrown in the console:

```
Uncaught Error: A request already contains one of the messages in this array
```

```php
<?php

use Livewire\Attributes\On;
use Livewire\Component;

// Parent listens to the event
{
    #[On('test-event')]
    public function handleEvent() {}

    public function render()
    {
        return <<<'HTML'
        <div>
            <livewire:child />
        </div>
        HTML;
    }
}

// Child also listens to the same event AND has a modelable grandchild
class Child extends Component
{
    public $value = '';

    #[On('test-event')]
    public function handleEvent() {}

    public function render()
    {
        return <<<'HTML'
        <div>
            <button wire:click="$dispatch('test-event')">Dispatch</button>
            <livewire:modelable-child wire:model="value" />
        </div>
        HTML;
    }
}
```

# The Problem

The component hierarchy:

```
Parent (listens to 'foo')
└── Child (listens to 'foo')
    └── Modelable Grandchild (wire:model bound to Child)
```

**What happens when the event fires:**

1. Event dispatched → Both Parent and Child need to handle it
2. Messages created:
   - Message A for Parent
   - Message B for Child
3. Partition interceptor runs (in `interactions.js`) to bundle modelable children with their ancestors

**Here's where it breaks:**

4. Processing Message A (Parent):
   - Calls `Parent.getDeepChildrenWithBindings()`
   - Finds Modelable Grandchild (it's a deep descendant with a binding)
   - Creates Message C for Grandchild
   - Bundles them: `compileRequest([A, C])` → **Request 1 = {A, C}**
5. Processing Message B (Child):
   - Calls `Child.getDeepChildrenWithBindings()`
   - Finds the same Modelable Grandchild (it's Child's direct child)
   - Gets the same Message C (already exists in pending messages)
   - Tries to bundle: `compileRequest([B, C])` → **ERROR** because C is already in Request 1

# The Solution

Instead of throwing an error when a message is already in an existing request, we find that existing request and merge the new messages into it:

- Request 1 starts as `{A, C}`
- When processing B, we find C is already in Request 1
- We add B to Request 1 → **Request 1 = {A, B, C}**

This ensures all related components are sent together in the same request, maintaining data consistency.

Fixes: #9825

### DIFF
--- a/js/request/index.js
+++ b/js/request/index.js
@@ -190,10 +190,29 @@ function sendMessages() {
             callback({
                 message,
                 compileRequest: (messages) => {
-                    if (Array.from(requests).some(request => Array.from(request.messages).some(message => messages.includes(message)))) {
-                        throw new Error('A request already contains one of the messages in this array')
+                    // When multiple components in a hierarchy listen to the same event and one
+                    // has a modelable child, the partition interceptor may try to bundle the
+                    // same child message into multiple requests. For example: Parent and Child
+                    // both listen to 'foo', and Child has a modelable grandchild. Parent's
+                    // interceptor bundles [Parent, Grandchild] into Request 1. Then Child's
+                    // interceptor tries to bundle [Child, Grandchild], but Grandchild is already
+                    // in Request 1. Rather than throwing an error, we merge Child into Request 1
+                    // so all related components are sent together, maintaining data consistency.
+                    let existingRequest = Array.from(requests).find(request =>
+                        messages.some(message => request.messages.has(message))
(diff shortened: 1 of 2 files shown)


calebporzio:
merge when resolved conflict thanks

PR #9843, opened 2026-01-27
[4.x] Expose dispatchEl and dispatchRef on $wire object

joshhanley:
# The Scenario

A user wants to dispatch an event to a specific DOM element from the client side (JavaScript/Alpine). They can do this from PHP using method chaining:

```php
$this->dispatch('post-removed')->el("#post-{$post->id}");
$this->dispatch('foo')->ref('myRef');
```

However, when trying to do the same from the client side, the methods are not available:

```blade
{{-- These don't work --}}
<button x-on:click="$wire.dispatchEl('#target', 'foo', { message: 'bar' })">...</button>
<button wire:click="$dispatchRef('target', 'foo')">...</button>
```

# The Problem

The `dispatchEl` and `dispatchRef` functions exist in `js/events.js` and are used internally when processing server-dispatched events (via `->el()` or `->ref()`), but they were not exposed on the `$wire` object for client-side use.

# The Solution

Added `dispatchEl` and `dispatchRef` to the `$wire` object by:

1. Importing the functions in `js/$wire.js`
2. Registering them as wire properties (`$dispatchEl`, `$dispatchRef`)
3. Adding aliases so users can use either `dispatchEl` or `$dispatchEl`

Now users can dispatch events to specific elements from both Alpine and wire directives:

```blade
{{-- Using Alpine --}}
<button x-on:click="$wire.dispatchEl('#target', 'foo', { message: 'bar' })">...</button>
<button x-on:click="$wire.dispatchRef('target', 'foo', { message: 'bar' })">...</button>

{{-- Using wire: directives --}}
<button wire:click="$dispatchEl('#target', 'foo', { message: 'bar' })">...</button>
<button wire:click="$dispatchRef('target', 'foo', { message: 'bar' })">...</button>
```

Fixes: #9829

### DIFF
--- a/js/$wire.js
+++ b/js/$wire.js
@@ -1,5 +1,5 @@
 import { cancelUpload, removeUpload, upload, uploadMultiple } from './features/supportFileUploads'
-import { dispatch, dispatchSelf, dispatchTo, listen } from '@/events'
+import { dispatch, dispatchEl, dispatchRef, dispatchSelf, dispatchTo, listen } from '@/events'
 import { generateEntangleFunction } from '@/features/supportEntangle'
 import { findComponentByEl } from '@/store'
 import { dataGet, dataSet } from '@/utils'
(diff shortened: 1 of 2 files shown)


ghabriel25:
Thanks a lot!

PR #9862, opened 2026-01-29
[4.x] Add $wire.$island() API for calling methods scoped to islands

joshhanley:
# The Scenario

When using `wire:click` with `wire:island`, scoping an action to a specific island works seamlessly:

```blade
<button wire:click="loadMore" wire:island.append="feed">Load more</button>
```

However, when triggering component methods from Alpine via `$wire`, there's no equivalent way to scope the call to a specific island:

```blade
{{-- How do you scope this to only re-render the "feed" island? --}}
<button x-on:click="$wire.loadMore()">Load more</button>
```

Users are forced to reach for internal APIs like `Livewire.fireAction()` with manually constructed metadata:

```js
Livewire.fireAction(this.$wire.__instance, 'loadMore', [], {
    island: { name: 'feed', mode: 'append' }
})
```

# The Problem

`$wire.$island()` exists but is undocumented. Its current implementation only supports refreshing an island — it directly fires a `$refresh` action and returns a Promise, so there's no way to call an arbitrary method through it:

```js
wireProperty('$island', (component) => async (name, options = {}) => {
    return fireAction(component, '$refresh', [], {
        island: { name, ...options },
    })
})
```

# The Solution

Change `$wire.$island('name')` to set island metadata on the next action using the existing `setNextActionMetadata` mechanism, then return `$wire`. Since JavaScript is single-threaded, the chained method call picks up the metadata before anything else can fire:

```js
wireProperty('$island', (component) => (name, options = {}) => {
    setNextActionMetadata({ island: { name, ...options } })

    return component.$wire
})
```

This means every `$wire` method works naturally — `$set`, `$toggle`, `$refresh`, `$commit`, and arbitrary method calls all get island-scoped automatically:

```js
$wire.$island('posts').loadMore()
$wire.$island('feed', { mode: 'append' }).loadMore()
$wire.$island('revenue').$refresh()
$wire.$island('settings').$set('theme', 'dark')
```

No backend changes were needed — the PHP side already handles island metadata on any method call.

Also adds documentation for the `$island()` API in both `islands.md` and `javascript.md`.

Fixes: #9654

### DIFF
--- a/docs/islands.md
+++ b/docs/islands.md
@@ -293,6 +293,68 @@ You can use `wire:poll` within an island to refresh just that island on an inter
 
 The polling is scoped to the island — only the island will refresh every 3 seconds, not the entire component.
 
+## Triggering islands from JavaScript
+
+You can trigger island updates from Alpine or JavaScript using `$wire.$island()`. This returns a `$wire`-like object where any method you call will only re-render the named island:
+
+```blade
+<?php // resources/views/components/⚡activity-feed.blade.php
+
+use Livewire\Attributes\Computed;
+use Livewire\Component;
+use App\Models\Activity;
+
+new class extends Component {
+    public $page = 1;
+
+    public function loadMore()
+    {
+        $this->page++;
+    }
+
+    #[Computed]
+    public function activities()
+    {
+        return Activity::latest()
+            ->forPage($this->page, 10)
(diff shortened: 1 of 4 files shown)


ghabriel25:
Awesome!

PR #9863, opened 2026-01-29
[3.x] Fix smart wire keys regex incorrectly matching > in PHP expressions

joshhanley:
# The Scenario

When using `@foreach` loops containing nested directives with `>` comparison operators, users encounter errors like "Trying to access array offset on null":

```php
<?php

use Livewire\Component;

class BudgetList extends Component
{
    public $items = [];

    public function render()
    {
        return <<<'HTML'
        <div>
            @foreach($items as $item)
                {{ $item > 5 ? 'big' : 'small' }}
            @endforeach

            @foreach($items as $item)
                {{ $item }}
            @endforeach
        </div>
        HTML;
    }
}
```

The error occurs because the first `@foreach` doesn't get its `openLoop()` marker injected, but `closeLoop()` is still called, causing a stack underflow on the second loop.

# The Problem

The regex pattern `(?![^<]*(?<![?=-])>)` was used to detect if a Blade directive is inside an HTML tag (to avoid injecting markers there). This pattern looks **forward** from the directive position and fails if it finds a `>` not preceded by `?`, `=`, or `-`.

The bug: when processing `@foreach($items as $item)`, the lookahead sees `> 5` in the Blade echo `{{ $item > 5 ? ... }}` and incorrectly thinks the `@foreach` is inside an HTML tag, skipping the `openLoop()` call.

The pattern was designed to handle cases like `<div @if($x > 0) class="foo" @endif>` where directives ARE inside a tag, but it incorrectly matched `>` comparison operators in PHP expressions that appear later in the template.

# The Solution

Replace the forward-looking regex with a **backwards-looking** approach that checks if the directive position is inside an unclosed HTML tag:

1. Search backwards from the directive position to find the most recent `<` that could start an HTML tag
2. Skip PHP tags (`<?php`), HTML comments (`<!--`), and invalid tag openers (like `<` in `1 < 5`)
3. Check if there's a closing `>` between that `<` and the directive position, correctly ignoring `>` inside quoted strings, parentheses, and braces

This approach is more reliable because it examines the actual context before the directive, rather than looking forward at unrelated PHP expressions.

## Benchmark Results

| Template | Size | Directives | Tags | Original | Fix |
|----------|------|------------|------|----------|-----|
| simple | 65 | 2 | 4 | 0.17 ms | 0.17 ms |
| medium | 221 | 6 | 4 | 0.31 ms | 0.39 ms |
| complex | 1187 | 18 | 21 | 2.02 ms | 1.90 ms |
| many_directives | 2720 | 80 | 40 | 3.00 ms | 3.71 ms |
| deeply_nested | 443 | 12 | 2 | 0.62 ms | 0.64 ms |
| real_world | 8569 | 62 | 102 | 2.28 ms | 2.36 ms |

The fix adds minimal overhead (~0.1ms on real-world templates) and compilation only happens once per template (cached).

Fixes: #9642
Fixes: #9776
Fixes: #9706


ao-jhelmich:
This PR resolves the issue from #9776 in the 3.x branch. However, the same issue is still occurring in v4.

ao-jhelmich:
> This PR resolves the issue from #9776 in the 3.x branch. However, the same issue is still occurring in v4.

well looks like it was a me problem, optimize:clear saved the day

DirtyRacer1337:
guess I found broken template used in filament admin panel

```blade
<{{ $tag }}>
</{{ $tag }}>
```

where `$tag` might be `a`

joshhanley:
@DirtyRacer1337 thanks, working on it!

joshhanley:
@DirtyRacer1337 I've tagged a new release with a fix.

marin246:
Unfortunately, the fix seems to not apply to @for loops:
What works now:
```
<select class="custom-select" wire:model.live="year" id="year-selection">
@foreach ($loopYears as $loopYear)
    <option value="{{ $loopYear }}">{{ $loopYear }}</option>
@endforeach
</select>
```

What still does not work:
```
<select class="custom-select" wire:model.live="year" id="year-selection">
@for ($loopYear = 2024; $loopYear <= now()->format('Y'); $loopYear++)
    <option value="{{ $loopYear }}">{{ $loopYear }}</option>
@endfor
</select>
```
The latter throws the error: Attempt to read property "index" on null

joshhanley:
@marin246 we've just merged https://github.com/livewire/livewire/pull/10084 which adds for loop support (not sure if it will fix this issue you are saying). Can you test from the 3.x branch and let me know if you're still having issues?

PR #9873, opened 2026-01-30
[4.x] Add failing tests for `wire:model.blur` form submission sync

joshhanley:
# The Scenario

When using `wire:model.blur` on an input inside a form, submitting the form by pressing Enter from that input causes the input's value to be silently dropped. The server receives the submit action but with stale (empty) data for the blurred input.

```php
<?php

use Livewire\Component;

new class extends Component
{
    #[Validate('required')]
    public string $name = '';

    public function submit(): void
    {
        $this->validate();
    }
};
?>

<div>
    <form wire:submit="submit">
        <input wire:model.blur="name" type="text" />
        <button type="submit">Submit</button>
    </form>
</div>
```

# The Problem

In v4.1, `wire:model` modifiers like `.blur` were changed to control **client-side** sync timing (#9832). This was a coordinated change with Alpine `3.15.5`, which added explicit `.blur` support to `x-model` — previously Alpine ignored the `.blur` modifier and synced on every keystroke.

The new behaviour is intentional and useful: `wire:model.blur` now defers updating `$wire` until the input loses focus. However, when a user presses Enter to submit a form, the browser fires events in this order:

1. `keydown` (Enter) on the input
2. `submit` on the form
3. `blur` on the input (only after submit)

Since `blur` fires **after** `submit`, Alpine's `x-model.blur` listener never syncs the value to `$wire` before Livewire builds the request payload. The value is missing from the diff, so the server never receives it.

Note: `.change` and `.enter` modifiers are not affected. The browser fires a `change` event on text inputs when Enter is pressed (before submit), and `keydown` fires before submit too, so Alpine's listeners for those modifiers sync in time.

# The Solution

The root cause is in Alpine's `x-model.blur` directive — it doesn't account for form submission firing before blur. The fix belongs in Alpine: alpinejs/alpine#4729.

This PR adds browser tests for all three modifiers (`.blur`, `.change`, `.enter`) with form submission via Enter. The `.blur` test is currently **expected to fail** until the Alpine fix is merged. The `.change` and `.enter` tests pass and provide regression coverage.

Once alpinejs/alpine#4729 is merged, the `.blur` test will pass and this PR can be merged.

Fixes: #9858


jeffchown:
Great!  Thank you, @joshhanley 🙏

joshhanley:
Have managed to fix this in Alpine instead, see PR https://github.com/alpinejs/alpine/pull/4729, so I have removed the fix from this PR but left the tests.

If that PR get's merged, re-running this PR's tests should pull the latest changes from Alpine's main branch and pass.

Youssef-Amjad:
Great! Thank you, @joshhanley  🙏

PR #9876, opened 2026-01-30
[4.x] Fix form object `#[Validate]` attributes losing rules on consolidated updates

joshhanley:
# The Scenario

When using a form object with `#[Validate]` attributes and deferred `wire:model`, submitting the form for the first time with all fields correctly populated throws a `MissingRulesException` — even though the validation rules are properly defined on the form.

It works correctly when there's a validation error to fix first (e.g. leaving a required field empty), because in that case only the corrected field is sent as an individual update rather than a consolidated one.

```
Missing [$rules/rules()] property/method on: [App\Livewire\Forms\ExampleForm].
```

```php
<?php

use Livewire\Component;

new class extends Component
{
    public ExampleForm $form;

    public function submit(): void
    {
        $validated = $this->form->validate();
    }
};
?>
    <input wire:model="form.name" />
    <input wire:model="form.email" />
    <button type="submit">Submit</button>
</form>
```

```php
<?php

namespace App\Livewire\Forms;

use Livewire\Form;

{
    #[Validate(['required', 'string', 'min:10', 'max:75'])]
    public ?string $name = null;

    #[Validate(['required', 'email', 'max:255'])]
    public ?string $email = null;
}
```

# The Problem

When **all** form fields change from their initial values in a single request (common on first submit with deferred `wire:model`), the JavaScript diff algorithm consolidates the individual field updates (`form.name`, `form.email`) into a single parent-level update (`form`). On the server, this consolidated update triggers `FormObjectSynth::hydrate()` a second time during `updateProperties()` — after `SupportAttributes::boot()` has already run. This creates a **new** form object instance whose `#[Validate]` attributes never have their `boot()` methods called, leaving `rulesFromOutside` empty and causing `MissingRulesException`.

The consolidation in `diffAndConsolidate()` (`js/utils.js`) happens when all children of an object change, which is the common case when every form field goes from its default value (e.g. `null`) to a user-provided value on first submit.

When only some fields change (e.g. correcting a single validation error), the diff sends individual dot-notation updates like `form.name`, which go through `recursivelySetValue()` and modify the **existing** form object — so the booted attribute state is preserved and validation works.

# The Solution

In `FormObjectSynth::hydrate()`, before creating a new form instance, check if a properly initialised form already exists at the same path on the component. If it does, reuse it and just update its properties. This preserves the booted attribute state (validation rules, messages, custom attributes, etc.) that was set up during the initial hydration cycle, rather than discarding it by creating a fresh instance.

Fixes: #9860

### DIFF
--- a/src/Features/SupportFormObjects/FormObjectSynth.php
+++ b/src/Features/SupportFormObjects/FormObjectSynth.php
@@ -41,28 +41,14 @@ class FormObjectSynth extends Synth {
         $existing = data_get($this->context->component, $this->path);
 
         if ($existing instanceof Form && $existing instanceof $meta['class']) {
-            foreach ($data as $key => $child) {
-                if ($child === null && Utils::propertyIsTypedAndUninitialized($existing, $key)) {
-                    continue;
-                }
-
-                $existing->$key = $hydrateChild($key, $child);
-            }
-
-            return $existing;
+            return $this->hydrateFormProperties($existing, $data, $hydrateChild);
         }
 
         $form = new $meta['class']($this->context->component, $this->path);
 
         $callBootMethod = static::bootFormObject($this->context->component, $form, $this->path);
 
-        foreach ($data as $key => $child) {
-            if ($child === null && Utils::propertyIsTypedAndUninitialized($form, $key)) {
-                continue;
-            }
-
-            $form->$key = $hydrateChild($key, $child);
-        }
+        $this->hydrateFormProperties($form, $data, $hydrateChild);
 
         $callBootMethod();
 
@@ -87,5 +73,18 @@ class FormObjectSynth extends Synth {
         return function () use ($form) {
             wrap($form)->boot();
         };
+    }
+
+    protected function hydrateFormProperties($form, $data, $hydrateChild)
+    {
+        foreach ($data as $key => $child) {
+            if ($child === null && Utils::propertyIsTypedAndUninitialized($form, $key)) {
+                continue;
+            }
+
+            $form->$key = $hydrateChild($key, $child);
+        }
+
+        return $form;
     }
 }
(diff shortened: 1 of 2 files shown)


klmader:
Hello! Just popping in to mention that I believe I've experienced this exact error during an attempted Livewire 3 -> 4 upgrade, so I'm very appreciative of the great documentation of the error. In case it's helpful, I also wanted to confirm that when I temporarily updated my FormObjectSynth class with this change, the error was resolved.

joshhanley:
@klmader great, thanks for testing!

calebporzio:
deep one. will need to understand this a bit better and lets make sure this is bullet proof as it touches synth stuff.

joshhanley:
@calebporzio I've simplified the implementation by extracting the form property hydration into a method, so it's clear we're handling property updates the same way. Below is a better explanation of the issue.

### The Problem

During a subsequent request, the component lifecycle runs in this order:

1. `HandleComponents::fromSnapshot()` → `hydrateProperties()` → `FormObjectSynth::hydrate()` creates the form object
2. `trigger('hydrate')` → `SupportAttributes::boot()` calls `boot()` on all attributes (including the form's `#[Validate]` attributes, which populate `rulesFromOutside`)
3. `HandleComponents::updateProperties()` → processes property updates from the client

When the JS diff algorithm consolidates all form field changes into a single `form` update (which happens if every field in a form has changed), step 3 triggers `FormObjectSynth::hydrate()` again. The original code always created a new form instance, discarding the original form request from step 1, which had already had it's attributes booted properly in step 2.

### The Fix

If the form object already exists, then just update it's properties instead of creating a new one.

```php
$existing = data_get($this->context->component, $this->path);

if ($existing instanceof Form && $existing instanceof $meta['class']) {
    return $this->hydrateFormProperties($existing, $data, $hydrateChild);
}
```

dotryen:
experienced this error while working on my site as well. tested the latest commit and it seems to have resolved the issue just fine.

PR #9885, opened 2026-02-01
[4.x] Add setUploadRoute, setPreviewRoute, and interceptUpload support

wojt-janowski:
Adds Livewire::setUploadRoute(), Livewire::setPreviewRoute(), and Livewire.interceptUpload(), aligning file upload/preview route customisation and request interception with the existing patterns for setUpdateRoute() and interceptRequest().

**Problem**
- File upload and preview routes are hardcoded with no way to customise them. 
- interceptRequest() only hooks into fetch-based update requests, not the XMLHttpRequest used for file uploads. 
- This makes it impossible for multi-tenant apps (or anyone needing custom middleware/prefixes) to add headers or customise the upload endpoint.
- This is a problem particularly because in a db-per-tenant setup, the session is scoped to the tenant, and trying to upload to a 'shared' file upload path fails with 419s due to session issues.

**Changes**
- PHP: Defers default upload/preview route registration to app()->booted(), allowing setUploadRoute() and setPreviewRoute() to override them first (same pattern as HandleRequests for the update route)
- JS: Adds interceptUpload() which runs callbacks before the upload XHR, allowing modification of url and headers
- Tests: Unit test for custom upload/preview routes following the CustomUpdateRouteUnitTest pattern


calebporzio:
## PR Review: #9885 — [4.x] Add setUploadRoute, setPreviewRoute, and interceptUpload support

**Type**: Feature
**Verdict**: Request changes

This PR adds three new customization hooks for file uploads, following the existing patterns for `setUpdateRoute()` and `interceptRequest()`:

1. **`Livewire::setUploadRoute($callback)`** — lets you override the default `/livewire/upload-file` route (e.g., for multi-tenant apps that need custom prefixes or middleware)
2. **`Livewire::setPreviewRoute($callback)`** — same for the preview file route
3. **`Livewire.interceptUpload($callback)`** — JS-side hook that lets you modify the URL/headers of the XHR upload request before it's sent (e.g., adding auth headers for tenant-scoped sessions)

The PHP approach defers default route registration to `app()->booted()` so custom routes can be set first. The JS approach adds a simple interceptor array that mutates a `{ url, headers }` payload before the upload XHR fires.

### Critical Issue: Breaks 44 of 55 existing file upload tests

The `app()->booted()` deferral in `SupportFileUploads::provide()` causes a **regression**: routes aren't registered in time for unit tests. Running the existing `UnitTest.php`:

```
ERRORS! Tests: 55, Assertions: 15, Errors: 44, Failures: 1
Route [livewire.upload-file] not defined.
```

This is because `app()->booted()` fires after all service providers boot, but many tests resolve the upload route during component lifecycle calls within the test — before `booted()` fires. The original code registers routes eagerly in `provide()`, which works.

1. **Eager registration + `default-` prefix (like `HandleRequests`)** — Register default routes immediately with `default-livewire.upload-file` name. Custom routes get `livewire.upload-file`. Use a custom `findUploadRoute()` to prefer non-default. This is what `HandleRequests::setUpdateRoute()` does. **Problem**: `GenerateSignedUploadUrl::forLocal()` uses `URL::temporarySignedRoute('livewire.upload-file', ...)` which relies on Laravel's route name lookup. Laravel's `RouteCollection::refreshNameLookups()` always picks the *first* route registered with a given name — you can't override via a second registration. `HandleRequests` avoids this by using its own `getUpdateUri()` resolver, but uploads need signed URLs which require `URL::temporarySignedRoute()` with a name. This approach requires updating `GenerateSignedUploadUrl` to resolve the route name through `SupportFileUploads::getUploadRouteName()` instead of hardcoding it.

2. **Eager registration, no deferral, store custom route object** — Register default routes eagerly. When `setUploadRoute()` is called, store the custom route. Update `GenerateSignedUploadUrl` to use `SupportFileUploads::getUploadRouteName()`. This is the cleanest approach but requires touching `GenerateSignedUploadUrl.php` too.

3. **The PR's approach (`app()->booted()`)** — Defer default registration. **Breaks existing tests.** Not viable without significant rework.

### Changes Made

No changes pushed. The dist/ files were already cleaned up during the merge-main step. The PR's approach needs fundamental rework — I explored several alternatives but they all require additional changes beyond what the PR provides (specifically updating `GenerateSignedUploadUrl.php` to resolve the route name dynamically).

### Test Results

- **Custom route tests** (`CustomUploadRouteUnitTest.php`): 2 pass (but only because they don't trigger the upload flow that resolves the route by name)
- **Existing file upload tests** (`UnitTest.php`): **44 of 55 FAIL** with `Route [livewire.upload-file] not defined`
- **CI**: No checks reported on this branch

### Code Review

**PHP:**
- `SupportFileUploads.php:38-50` — `app()->booted()` deferral is the root cause of the regression. Routes must be registered eagerly.
- `SupportFileUploads.php:53-73` — `setUploadRoute`/`setPreviewRoute` methods follow the `HandleRequests` pattern correctly, but don't account for the name collision with the default route when using `URL::temporarySignedRoute()`.
- `SupportFileUploads.php:75-94` — `uploadRouteExists()`/`previewRouteExists()`/`findRouteByName()` are unnecessary if we use the eager registration approach. They add complexity without value.
- `LivewireManager.php:164-172` — `setUploadRoute`/`setPreviewRoute` facade methods are clean and follow existing patterns.
- Static `$uploadRoute`/`$previewRoute` properties need `flush-state` handling for test isolation (not present in PR).

**JS:**
- `supportFileUploads.js:3-7` — `interceptUpload` is much simpler than `interceptRequest`. It doesn't return a cleanup function, which is inconsistent. `interceptRequest` returns a function to unregister the interceptor; `interceptUpload` doesn't. This could cause memory leaks in SPA-like scenarios with `wire:navigate`.
- `supportFileUploads.js:167-170` — The interceptor only runs in `handleSignedUrl`, not in `handleS3PreSignedUrl`. If someone's using S3 uploads, the interceptor is silently ignored. Should this be documented or should both paths be intercepted?
- No semicolons, uses `let` — style is correct ✓

**Tests:**
- `CustomUploadRouteUnitTest.php` — Tests are minimal and only verify route registration, not that uploads actually use the custom route. A more thorough test would verify that `GenerateSignedUploadUrl::forLocal()` produces a URL using the custom route.

### Security

No security concerns identified. The `setUploadRoute`/`setPreviewRoute` methods pass the handler through the same controller classes (`FileUploadController`/`FilePreviewController`), preserving existing security checks. The `interceptUpload` JS hook only lets callers modify URL and headers on the client side — the server still validates the signed URL.

### Verdict

**Request changes.** The feature request is legitimate and well-motivated (multi-tenant apps need this), but the implementation has a critical regression that breaks 44 existing tests. The `app()->booted()` approach doesn't work. The fix requires:

1. Register routes eagerly (like the original code and like `HandleRequests`)
2. Add a `getUploadRouteName()` helper that returns the custom route's name when set
3. Update `GenerateSignedUploadUrl::forLocal()` to use `SupportFileUploads::getUploadRouteName()` instead of the hardcoded `'livewire.upload-file'` string
4. Add `flush-state` handling for the static properties
5. Fix `interceptUpload` to return a cleanup function (consistency with `interceptRequest`)
6. Remove built `dist/` files from the PR

Zero reactions on the PR suggest low community demand, but the use case is valid.

---
*Reviewed by Claude*

calebporzio:
Hey, thanks for putting this together — I appreciate the effort.

After digging into this I'm going to close it out though. Here's my reasoning:

## The upload URL is already controlled server-side

When a file upload starts, the JS calls `_startUpload`, which calls `GenerateSignedUploadUrl::forLocal()`, which does:

```php
URL::temporarySignedRoute('livewire.upload-file', ...);
```

That signed URL is sent back to the JS, and the JS just POSTs to whatever URL the server told it to. So if you register a custom route named `livewire.upload-file`, Laravel's `URL::temporarySignedRoute()` will resolve to your custom URL automatically. The JS doesn't need to know anything changed.

This means customizing the upload route is already possible today without any new API. In your `AppServiceProvider::boot()`:

```php
Route::post('/my/custom/upload', [FileUploadController::class, 'handle'])
    ->name('livewire.upload-file')
    ->middleware(['web', 'auth']);
```

Same goes for the preview route.

## `interceptUpload` on the JS side is solving a problem that doesn't exist

Since the URL comes from the server via the signed URL, there's no need to intercept and modify it client-side. The JS interceptor also doesn't cover S3 uploads (`handleS3PreSignedUrl`), so it would be an incomplete API.

## Why `setUpdateRoute` exists but `setUploadRoute` doesn't need to

`setUpdateRoute` has additional concerns — the update route URI needs to be resolved for the JS request payload, there's `default-livewire.update` fallback logic, etc. Upload routes don't have those concerns. They're resolved entirely by route name via `URL::temporarySignedRoute`.

## Cost

Three new public API methods maintained forever, route registration timing complexity, static state management, and more surface area for routing bugs — for something that's already achievable with standard Laravel routing.

---

Hope that makes sense. If you have a specific use case that the approach above doesn't cover, happy to hear about it!

wojt-janowski:
Thanks Caleb - fair enough! Will try that out shortly.

I may submit a docs PR for that if I can string the words together sensibly enough!

PR #9891, opened 2026-02-02
[4.x] Fix view resolution for self-named class components

joshhanley:
# The Scenario

When using a class-based component that follows the "self-named" folder convention (where the class name matches its parent folder), and the component is instantiated from its class name — such as in tests — the view file cannot be found.

For example, a component at `App\Livewire\Events\List\Filters\Filters` with its view at `resources/views/livewire/events/list/filters/filters.blade.php` throws an error when tested:

```
Illuminate\View\ViewException: File does not exist at path
/var/www/html/resources/views/livewire/events/list/filters.blade.php.
```

```php
Livewire::test(Filters::class)->assertSomething()
```

# The Problem

In v3, `generateNameFromClass()` produces the full name for self-named class components — `Filters\Filters` becomes `filters.filters`. View resolution then correctly finds `filters/filters.blade.php`.

The v4 single-file components work added logic to collapse self-named names (`filters.filters` → `filters`), which broke view resolution for class-based self-named components. The collapsed name `filters` causes `getView()` to look for `filters.blade.php` instead of `filters/filters.blade.php`.

# The Solution

Remove the self-named collapsing from `generateNameFromClass()` to preserve v3 behaviour:

- Self-named class components keep their full name (`Filters\Filters` → `filters.filters`)
- Index class components still have `.index` stripped (`IndexComponent\Index` → `index-component`) — same as v3
- View resolution works naturally without needing fallback paths

Fixes: #9869


calebporzio:
i think there's a deeper fix here related to class components and how they get their name. 

whatever v3 does for class components re: self/index-named, we should just preserve that.

joshhanley:
@calebporzio yeah the way it worked in v3 is self-named class components keep their full name (`Filters\Filters` → `filters.filters`) and index class components have `.index` stripped (`IndexComponent\Index` → `index-component`). I've updated the PR to match that.

PR #9892, opened 2026-02-02
[4.x] Invalid Typed Query String Check

ariaieboy:
As discussed [here](https://github.com/livewire/livewire/discussions/9451), the livewire throw 500 error when a user send array instead of a string or an int in the query string. 

Throwing a 500 error when a user sends invalid data is not a good idea. We must either return status code 400 or 422, or we just ignore that value and return an error to the user that your input is not correct.

I've added a strict mode to the URL attribute so users can change the behavior of the framework. By default, we abort with status code 400 instead of 500 since it's not a server error but a user input error and a bad request from the user. 

Or if it's not strict, we validate the type of the user input according to the type of the property.

Since I am not sure that this change is good enough,I did not add documentation for it. Please let me know if it's good or the api need changes so I can add this feature to the doc.

### DIFF
--- a/src/Features/SupportQueryString/BaseUrl.php
+++ b/src/Features/SupportQueryString/BaseUrl.php
@@ -95,7 +95,11 @@ class BaseUrl extends LivewireAttribute
             }
             // Validate the type compatibility non-strictly
             if (!$this->strict){
-                $validator = Validator::make([$this->getSubName()=>$value],[$this->getSubName()=>(string)$this->type]);
+                $rule = $this->type->getName();
+                if ($this->nullable){
+                    $rule .= '|nullable';
+                }
+                $validator = Validator::make([$this->getSubName()=>$value],[$this->getSubName()=>$rule]);
                 if ($validator->fails()) {
                     $this->component->addError($this->getSubName(),'The value type is not compatible.');
                     return;
--- a/src/Features/SupportQueryString/BaseUrl.php
+++ b/src/Features/SupportQueryString/BaseUrl.php
@@ -11,14 +11,17 @@ use ReflectionClass;
 class BaseUrl extends LivewireAttribute
 {
     protected $type = null;
+
     public function __construct(
         public $as = null,
         public $history = false,
         public $keep = false,
         public $except = null,
         public $nullable = null,
-        public $strict  = true
-    ) {}
+        public $strict = true
+    )
+    {
+    }
 
     public function mount()
     {
@@ -30,7 +33,7 @@ class BaseUrl extends LivewireAttribute
 
     public function dehydrate($context)
     {
-        if (! $context->mounting) return;
+        if (!$context->mounting) return;
 
         $this->pushQueryStringEffect($context);
     }
@@ -51,13 +54,16 @@ class BaseUrl extends LivewireAttribute
 
         return false;
     }
-    protected function determineType(){
+
+    protected function determineType()
+    {
         $reflectionClass = new ReflectionClass($this->getSubTarget() ?? $this->getComponent());
         if ($this->getSubName() && $reflectionClass->hasProperty($this->getSubName())) {
             return $reflectionClass->getProperty($this->getSubName())?->getType();
         }
         return null;
     }
+
     public function setPropertyFromQueryString()
     {
         if ($this->as === null && $this->isOnFormObjectProperty()) {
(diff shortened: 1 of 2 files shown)


calebporzio:
## PR Review: #9892 — [4.x] Invalid Typed Query String Check

**Type**: Bug fix + Feature (Mixed — this is a concern)
**Verdict**: Needs discussion

1. User has a Livewire component with `#[Url] public string $search = '';`
2. A malicious or malformed request sends `?search[]=fake` (an array instead of a string)
3. PHP's query parser hands Livewire an array `['fake']`
4. Livewire calls `setValue(['fake'])` which calls `data_set()` → PHP tries to assign an array to a `string` property → **TypeError → 500 error**
5. This is genuinely bad: user-controlled input should never cause a 500. It's a bad request (400), not a server error.

The bug is real and confirmed. I reverted the fix locally and verified the test produces a TypeError/500 without it.

1. **Just `abort(400)` — no new API surface (recommended).** When an array value arrives for a non-array typed property, abort(400). Simple, correct, no new parameters on `#[Url]`. Users who want custom query string validation can add it separately. This is the surgical bug fix.

2. **Silently ignore the bad value — keep the property default.** Instead of aborting, just `return` early without calling `setValue()`. The property keeps its default value. Less aggressive than a 400, but could hide real bugs in legitimate use.

3. **This PR's approach: `strict` parameter with two modes.** Abort(400) when `strict: true` (default), add validation error when `strict: false`. This works but adds new public API surface (`strict` param on `#[Url]`) that conflates a bug fix with a feature.

### Changes Made

No changes pushed — this needs API design direction from Caleb first.

### Test Results

- All 4 new PR tests pass with the fix ✅
- All 14 SupportQueryString unit tests pass ✅
- Verified test fails (TypeError/500) without the fix ✅
- CI: All 22 checks passing ✅

### Code Review

**Issues found:**

1. **`strict` is feature creep on a bug fix** (`BaseUrl.php:21`). The core bug is "500 on bad input." The fix should be one thing: don't 500. Adding a `strict` mode with validation errors is a new feature riding on a bug fix. Recommend: just abort(400), no new parameter.

2. **`strict` is an ambiguous name** (`BaseUrl.php:21`). Strict about what? Type checking? Validation? URL encoding? If this parameter is kept, it needs a more descriptive name.

3. **Unused import: `Validator`** (`BaseUrl.php:5`). `use Illuminate\Support\Facades\Validator;` is imported but never used.

4. **Unused import: `HttpException`** (`UnitTest.php:5`). `use HttpException;` is imported but never used.

5. **Gratuitous whitespace changes** (`BaseUrl.php:36,176`). `! $context->mounting` → `!$context->mounting` and similar. These are unrelated style changes that pollute the diff.

6. **`determineType()` duplicates reflection logic from `determineNullability()`** (`BaseUrl.php:58-65`). Both methods create a `ReflectionClass` and check `getSubName()/hasProperty()`. Could be consolidated, but this is minor.

7. **Constructor brace style changed** (`BaseUrl.php:22-24`). The `{}` after the constructor was reformatted to a multi-line block. Gratuitous.

8. **Snake_case variable in PHP** (`BaseUrl.php:107`). `$value_type` should be `$valueType` per PSR/Laravel conventions.

**What's good:**
- The bug identification is correct and the fix location is right
- Tests cover named types, union types, nullable types, and both strict/non-strict modes
- The reflection-based approach to check the property type is sound

### Security

The bug itself is a minor security concern — user-controlled input causing 500s could be used for targeted error-based probing or log pollution. The fix (returning 400) is the correct security posture. No new security issues introduced by this PR.

### Verdict

**Needs discussion.** The bug is real and should be fixed, but I'd recommend a much simpler approach:

```php
// In setPropertyFromQueryString(), right before setValue():
if (is_array($value) && $this->type instanceof \ReflectionNamedType && $this->type->getName() !== 'array') {
    abort(400);
}
// (Plus union type handling)
```

No `strict` parameter, no validation errors, no new API surface. Just: "you sent garbage → 400." If Caleb wants the `strict: false` validation error mode as a feature, that should be a separate PR.

The contributor (@ariaieboy) identified the right problem and the right fix location — the implementation just needs to be pared down to a pure bug fix. 3 heart reactions on the PR suggests community interest.

---
*Reviewed by Claude*

calebporzio:
Taking this over into a new PR: https://github.com/livewire/livewire/pull/9945

Thanks for this. think we're going to just fail silently instead of throwing any kind of error. Makes sense to me that if the data doesn't conform to the right type it's just silently skipped.

PR #9893, opened 2026-02-02
[4.x] Fix JS objects passed as functions in `wire:click` `$js` expressions

joshhanley:
# The Scenario

When iterating over a Livewire property with `x-for` and passing an iteration variable to a `$js` function via `wire:click`, the variable arrives as a function instead of the expected plain JavaScript object:

```php
<?php

use App\Models\User;
use Livewire\Component;

new class extends Component {
    public $users = [];

    public function mount()
    {
        $this->users = User::all()->toArray();
    }
};
?>
<div>
    <template x-for="user in $wire.users" :key="user.id">
        <div>
            <flux:heading size="lg" x-text="user.name"></flux:heading>
            <flux:button variant="primary" wire:click="$js.select(user)">
                Select
            </flux:button>
        </div>
    </template>
</div>

<script>
    this.$js.select = (user) => {
        console.log(typeof user) // "function" — expected "object"
    };
</script>
```

# The Problem

When `wire:click` processes an expression like `$js.select(user)`, the expression contextualiser in `js/evaluator.js` prefixes bare identifiers with `$wire.`, transforming it into `$wire.$js.select($wire.user)`.

Since `user` isn't a Livewire component property, alias, or registered `$wire` property, the `$wire` proxy's get handler falls through to the method fallback — which returns a callable function (designed to fire a server action). This callable is what gets passed to the `$js` function instead of the actual Alpine-scoped object from the `x-for` loop.

# The Solution

The expression contextualiser in `contextualizeExpression()` now receives the element and checks Alpine's scope chain via `Alpine.closestDataStack(el)` before prefixing an identifier with `$wire.`. If the identifier exists in Alpine's scope (e.g. an `x-for` iteration variable), it is left unprefixed so Alpine resolves it directly.

This keeps the fix isolated to `wire:click` expression processing rather than affecting all `$wire` property access.

Fixes: https://github.com/livewire/livewire/discussions/9881

### DIFF
--- a/js/$wire.js
+++ b/js/$wire.js
@@ -114,23 +114,7 @@ Alpine.magic('wire', (el, { cleanup }) => {
                 return generateEntangleFunction(component, cleanup)
             }
 
-            let result = component.$wire[property]
-
-            // If the property isn't a known $wire property, state, or alias,
-            // it will resolve to the method fallback (a function that fires a
-            // server action). Before returning this fallback, check if the
-            // property exists in Alpine's scope chain (e.g. an x-for variable).
-            // This allows wire:click="$js.select(user)" to correctly pass
-            // Alpine-scoped variables instead of treating them as server methods...
-            if (result?.__livewire_method_fallback) {
-                let scope = Alpine.mergeProxies(Alpine.closestDataStack(el))
-
-                if (property in scope) {
-                    return scope[property]
-                }
-            }
-
-            return result
+            return component.$wire[property]
         },
 
         set(target, property, value) {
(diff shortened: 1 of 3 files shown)


calebporzio:
very interesting mate. 

my first instinct is that you shouldn't mix x-for and wire:??

But that said, this is kind of a clean addition i think.

only hole in it is any alpine scope, even scope outside of this component will overide $wire properties from this component.

could also see adding wire:for at this point honestly lol 🤦‍♂️

So yeah, we probably need a shortcircuit callback added to Alpine.closestDataStack and stop collecting data proxies at he component barrier.

then we good.

joshhanley:
This PR now depends on alpinejs/alpine#4734, which adds a `shouldStopAtNode` callback to `closestDataStack`. This allows Livewire to stop collecting Alpine scope at the component boundary, so variables like `item` from an outer `x-data` don't shadow `$wire.item` inside the component.

Note: tests will fail in this PR until the Alpine PR is merged.

joshhanley:
We closed this as there's a chance that is behaviour might be unexpected. So going to hold off on this for now. The solution is, if you switch to Alpine, then use Alpine.

So for the example from above, because it's using `x-for` (aka switching to Alpine), then the click listener should also be in Alpine, like this:
```
<template x-for="user in $wire.users" :key="user.id">
    <button x-on:click="$wire.$js.select(user)">Select</button>
</template>
```

PR #9898, opened 2026-02-02
[4.x] Fix undefined array key 'errors' in file upload error handling

MrScript-DEV:
## Description

  This PR fixes an "Undefined array key 'errors'" error that occurs in the file upload error handling when `json_decode()` returns malformed JSON without the expected 'errors' key.

  ## Problem

  In `WithFileUploads::_uploadErrored()` method at line 93, the code directly accesses `json_decode($errorsInJson, true)['errors']` without checking if the 'errors' key exists:

  ```php
  $errors = json_decode($errorsInJson, true)['errors'];
```

  When the JSON is malformed or doesn't contain an 'errors' key, this throws:
  Undefined array key 'errors'

  ## Solution

  The fix safely checks if the 'errors' key exists and falls back to the same error handling logic that's already used when $errorsInJson is null:

```php
  $decoded = json_decode($errorsInJson, true);

  if (!isset($decoded['errors'])) {
      return $this->_uploadErrored($name, null, $isMultiple);
  }

  $errors = $decoded['errors'];
```

  This approach:
  - ✅ Reuses existing, well-tested error handling logic
  - ✅ Maintains consistent behavior (same error message whether JSON is null or malformed)
  - ✅ Preserves translation support
  - ✅ Follows DRY principles

  ## Checklist Response

  1️⃣ Is this wanted/needed? ✅ Yes - This fixes a runtime error that can occur during file uploads when server responses are malformed.

  2️⃣ Created a branch? ✅ Yes - fix-file-upload-undefined-errors-key

  3️⃣ Multiple unrelated changes? ❌ No - Single focused fix for the undefined array key issue

  4️⃣ Includes tests? ✅ Yes - Added test_file_upload_error_with_malformed_json_falls_back_to_default_message()

  5️⃣ Thorough description? ✅ Yes - See above with code snippets and reasoning


calebporzio:
## PR Review: #9898 — [4.x] Fix undefined array key 'errors' in file upload error handling

**Type**: Bug fix
**Verdict**: Merge

1. When a file upload fails on the JS side, the response body is sent to PHP's `_uploadErrored()` as `$errorsInJson`
2. If the HTTP response is 422, the raw response body string is passed. Livewire expects this to be `{"errors": {...}}` — Laravel's standard validation error format
3. If a 422 response comes back in a non-standard format (e.g., from a reverse proxy, custom middleware, or non-Laravel validation layer), `json_decode(...)['errors']` throws "Undefined array key 'errors'"
4. The fix: after decoding, check if the `errors` key exists. If not, fall through to the same default error message path that's already used when `$errorsInJson` is null

1. **Recursive call** (`return $this->_uploadErrored($name, null, $isMultiple)`) — cleaner than copy-paste but adds a stack frame for no reason
2. **Restructure to single fallback** (what I pushed) — invert the `is_null` check so the "try to decode errors" block is nested inside a `if (! is_null(...))` guard, with the default message as a single fallthrough at the bottom. Eliminates the code duplication entirely. This is the cleanest approach.
3. **Just use `?? null`** on the array access and leave the rest — the contributor's original approach, but it duplicated the entire 8-line fallback block verbatim

### Changes Made

- **Refactored `_uploadErrored()`** to eliminate duplicated error fallback block. The contributor's fix copy-pasted the entire null-handling block (translator lookup, attribute resolution, message building, throw). Restructured so the JSON decode is attempted first inside a `!is_null` guard, and the default message generation is the single fallthrough path at the bottom. Net: -11 lines, same behavior, zero duplication.

### Test Results

- All 55 unit tests in `SupportFileUploads/UnitTest.php` pass
- Verified the PR's test **fails on main** with `ErrorException: Undefined array key "errors"` (confirmed regression)
- Verified the PR's test **passes with the fix**
- No CI checks reported on this branch

### Code Review

- **`WithFileUploads.php:93`**: The `?? null` approach for the missing key is correct. The refactored structure is cleaner than the original PR's duplicated block.
- **`UnitTest.php:1045-1050`**: Test fixture `uploadErrorWithMalformedJson` correctly simulates the scenario — valid JSON but without the `errors` key. Follows the same pattern as the existing `uploadError` fixture above it.
- **Test naming**: `test_file_upload_error_with_malformed_json_falls_back_to_default_message` is descriptive and accurate.
- **Style**: PHP follows project conventions. No JS changes.

### Security

No security concerns identified. This is a defensive null-check in an error handling path.

### Verdict

Clean bug fix for a real edge case. The contributor correctly identified the problem — a missing array key when the JSON response doesn't contain the expected `errors` structure. The test precisely isolates the bug. I've pushed a refactor that removes the code duplication from the original fix (8 lines copy-pasted verbatim) by restructuring the method's control flow. All existing tests pass. Merge.

---
*Reviewed by Claude*

PR #9904, opened 2026-02-03
[4.x] Fix island polls triggering view transitions outside the island

joshhanley:
# The Scenario

When using `wire:transition` with a named view transition (e.g., `wire:transition="step"`) alongside an `@island` that polls, every poll request causes the named view transition to replay — even though the content with the transition is outside the island and hasn't changed.

Screenshot 2026-02-03 at 03 21 20PM

```blade
<?php
use Livewire\Component;

new class extends Component {
    public $step = 1;

    public function nextStep()
    {
        $this->step = 2;
    }
}; 
?>
<div>
    <div class="mx-auto max-w-2xl p-8 text-center text-3xl">
        @if($step === 1)
            <div wire:transition="step" wire:key="step-1">
                <h1>Step 1</h1>
                <button wire:click="nextStep">Next</button>
            </div>
        @elseif($step === 2)
            <div wire:transition="step" wire:key="step-2">
                <h1>Step 2</h1>
            </div>
        @endif

        @island(lazy: true)
        <div wire:poll.3s>{{ now()->toDateTimeString() }}</div>
        @endisland
    </div>
</div>
```

# The Problem

When morphing an island fragment, `morphFragment` in `js/morph.js` passes the island's parent element to `transitionDomMutation` to check for `wire:transition` elements. Since the parent contains everything — including `wire:transition` elements outside the island — the check incorrectly finds them and triggers a view transition for the island morph.

# The Solution

Build a temporary container with only the island's actual content (the nodes between the start and end markers) before passing it to `transitionDomMutation`. This ensures the transition check only sees `wire:transition` elements that are actually inside the island.

Transitions inside islands continue to work correctly.

Screenshot 2026-02-03 at 03 18 07PM

Fixes #9845


calebporzio:
better implementation here, but good otherwisr

calebporzio:
## PR Review: #9904 — Fix island polls triggering view transitions outside the island

**Type**: Bug fix
**Verdict**: Merge

1. A Livewire component has a `wire:transition="step"` element **outside** an `@island`, and the island has `wire:poll`.
2. When the island polls, `morphFragment()` is called. It passes `startNode.parentElement` (the entire parent div) to `transitionDomMutation()`.
3. `transitionDomMutation()` does `fromEl.querySelector('[wire\\:transition]')` — finds the `wire:transition` element that's *outside* the island, in the parent.
4. This triggers `document.startViewTransition()` on every poll cycle, even though nothing with `wire:transition` is changing.
5. Result: spurious view transition animations replay every poll.

1. **Clone island nodes into a temp container (original PR approach)** — Works but wasteful. Clones every node just for a `querySelector` check. Also subtly wrong: `setTransitionNames()` sets `viewTransitionName` on cloned elements, not the real DOM, so named transitions inside islands wouldn't get proper old-state snapshots from the View Transitions API (the test only checks transition count, not visual correctness).
2. **Walk island nodes to check, pass real parent when transitions exist (pushed approach)** — No cloning. When the island has no `wire:transition`, passes an empty div that short-circuits `transitionDomMutation`. When it does have transitions, passes the real `fromContainer` so `setTransitionNames` works on actual DOM elements.
3. **Refactor `transitionDomMutation` to accept fragment boundaries** — Cleaner API long-term but overkill for this fix.

Went with approach 2.

### Changes Made

- **Rewrote the JS fix** (`js/morph.js`): replaced cloneNode loop with a targeted walk that checks for `wire:transition` without cloning. Passes real parent or empty div based on result.
- **Simplified first test**: removed unused `nextStep()` method, `$step` property, CSS keyframes, `wire:key` — none needed to reproduce the bug. Stripped to minimum reproduction.
- **Simplified second test**: removed unnecessary CSS — `startViewTransition` fires regardless of `::view-transition` CSS rules.

### Test Results

- Both tests pass with fix ✅
- First test confirmed to fail on `main` without fix (`viewTransitionCount` = 1, expected 0) ✅
- All CI checks passed on original PR ✅

### Code Review

- `js/morph.js:104-117`: Clean implementation. The `nodeType === 1` check correctly skips text/comment nodes. Optional chaining on `hasAttribute`/`querySelector` handles edge cases. Early `break` avoids unnecessary iteration.
- `BrowserTest.php:926-956`: First test is now minimal — just a `wire:transition` outside an island + a polling island. No extra state, no CSS, no unused methods.
- `BrowserTest.php:958-990`: Second test correctly verifies the positive case (transitions inside islands still trigger).

### Security

No security concerns identified. The fix only changes which element is passed to `querySelector` — no new user input handling, no hydration changes.

### Verdict

Clean bug fix with proper regression tests. Root cause is correct: `morphFragment` was leaking the parent scope to `transitionDomMutation`. The improved implementation avoids unnecessary DOM cloning and correctly handles both the negative case (no transitions in island) and positive case (transitions inside island). Merge.

---
*Reviewed by Claude*

PR #9910, opened 2026-02-03
[4.x] Fix compatibility with iOS <16.4

vladrusu:
## Summary

Safari iOS < 16.4 doesn't support JavaScript lookbehind regex assertions, causing "Invalid regular expression: invalid group specifier name" errors that completely break Livewire on affected browsers.

## The Problem

The `contextualizeExpression` function in `js/evaluator.js` uses a lookbehind assertion:
```javascript
result = result.replace(/(?<![.\w$])(\$?[a-zA-Z_]\w*)/g, ...)
```

This pattern is not supported in:
- Safari iOS < 16.4
- Safari macOS < 16.4  
- Any WebKit-based browser without lookbehind support

See: https://caniuse.com/js-regexp-lookbehind

## The Fix

Replace the lookbehind with a capturing group that achieves the same result:
```javascript
result = result.replace(/(^|[^.\w$])(\$?[a-zA-Z_]\w*)/g, (m, pre, ident, offset) => {
    // ... same logic ...
    return pre + '$wire.' + ident
})
```

This approach:
- Captures the character before the identifier (or start of string)
- Prepends it to the result, maintaining the original behavior
- Works in all browsers

## Tested

✅ Verified working on Safari iOS (iPhone 13)

## Note for Maintainers

This PR only modifies the source file `js/evaluator.js`. The dist files will need to be rebuilt after merging.

🤖 Generated with [Claude Code](https://claude.ai/code)


calebporzio:
## PR Review: #9910 — [4.x] Fix compatibility with iOS <16.4

**Type**: Bug fix
**Verdict**: Merge

1. `contextualizeExpression` in `js/evaluator.js` rewrites Blade expressions like `foo` into `$wire.foo` so they reference the Livewire component.
2. The regex that does this uses a **lookbehind assertion** (`(?<![.\w$])`) to avoid prefixing identifiers that are already dot-accessed (like `.bar` in `foo.bar`).
3. Lookbehinds aren't supported in Safari iOS < 16.4 (or Safari macOS < 16.4). On those browsers, Livewire throws a regex parse error and **completely breaks**.
4. The fix replaces the lookbehind with a capturing group `(^|[^.\w$])` that captures the preceding character (or start of string), then prepends it back in the replacement. Functionally identical, universally supported.

1. **Polyfill/transpile** — Could use a Babel plugin to downcompile regex lookbehinds. Overkill for a single regex, adds build complexity.
2. **Two-pass approach** — Match all identifiers first, then filter out dot-accessed ones. More code, same result.
3. **Capturing group (this PR)** — Simplest, most direct. One line of regex change, same behavior. Clear winner.

### Changes Made
- Removed verbose `@see` and `Note:` comments — the regex is self-explanatory, and comments explaining "why we didn't use X" add noise.

### Test Results
- Vitest is broken in the project environment (jsdom/parse5 ESM incompatibility — pre-existing, not from this PR).
- Manually verified all 29 test cases (the full `evaluator.spec.js` suite + 7 edge cases) produce **identical output** between the old lookbehind and new capturing group approaches.
- Specifically verified the `offset + m.length` colon check still works correctly despite `m` now including the `pre` capture — object keys (`{ foo: bar }`) and ternaries (`foo ? bar : baz`) both handled correctly.
- No CI checks reported on this branch.

### Code Review
- `js/evaluator.js:51` — The regex change is correct and minimal. The capturing group `(^|[^.\w$])` is a textbook replacement for the lookbehind `(?<![.\w$])`.
- The `pre +` additions on lines 52-54 correctly propagate the captured preceding character.
- The `offset` parameter still refers to the full match position, so `result[offset + m.length]` correctly checks the character after the identifier even though `m` is now longer (includes `pre`).
- No semicolons, no `const` — style is clean.
- No dist files in the diff — contributor correctly noted these need rebuilding after merge.

### Security
No security concerns identified. This is a pure regex equivalence change with no behavioral difference.

### Verdict
Clean, correct, minimal fix for a real compatibility issue. Safari iOS < 16.4 still has meaningful market share (especially in enterprise/education where devices don't get updated). The regex change is provably equivalent — merge it.

---
*Reviewed by Claude*

PR #9918, opened 2026-02-04
[4.x] Fix "Route not defined" error for update route when route name index is stale

aydinza:
## The Problem

After the recent catch-all route fix (commit 02c17d2), `getUpdateUri()` can fail with:

```
Route [default.livewire.update] not defined
```

This happens in edge cases where:
- The route object exists and is returned by `findUpdateRoute()`
- But the route name isn't fully indexed in Laravel's named route collection
- Causing `route($route->getName(), [], false)` to fail

The issue manifests when rendering pages with Livewire components - the error occurs during HTML asset injection, even though `php artisan route:list` shows the route exists correctly.

## The Fix

Since we already have the route object, use `$route->uri()` directly instead of looking it up again by name via the `route()` helper:

```diff
-        return (string) str(
-            route($route->getName(), [], false)
-        )->start('/');
+        return (string) str($route->uri())->start('/');
```

## Why This Works

1. We already have a valid route object from `$this->updateRoute` or `findUpdateRoute()`
2. No need for redundant name-based lookup via `route()` helper
3. Avoids timing issues with route name indexing

## Testing

The existing test `test_get_update_uri_works_when_update_route_property_is_null()` continues to pass since we're still returning the same URI, just obtaining it more directly.


edalzell:
I hit this in v3 as well, any chance you could back-port this to v3 as well?

joshhanley:
@aydinza thanks for the PR! @aydinza and @edalzell are you able to provide me a way to replicate this problem, so I can verify it? Thanks!

edalzell:
> @aydinza thanks for the PR! @aydinza and @edalzell are you able to provide me a way to replicate this problem, so I can verify it? Thanks!

I am not, sorry. For us it only happens when I deploy to production, not even staging. And no issues locally. Still trying to nail down why it happens only there.

aydinza:
Hey @joshhanley, here's a reproduction of the underlying issue. It was really hard to reproduce it, because as I've mentioned, it's an edge case, which I believe directly related to the Octane worker resets.

The core problem is a timing issue in Laravel's RouteCollection. When Livewire registers its route in boot() with fluent
->name():

  Route::post(path, $handle)->middleware('web')->name('default.livewire.update');

  Laravel's Router::addRoute() calls RouteCollection::add() before the ->name() method executes. So the route enters
  allRoutes (findable by getRoutes() iteration) but not nameList (used by getByName() / the route() helper).

  Normally this is ok because Laravel calls refreshNameLookups() after loading route files, which rebuilds nameList. But
  if that refresh doesn't happen or the state diverges (Octane worker resets, compiled route edge cases, or service provider boot ordering), you get:

  - findUpdateRoute() → finds the route (iterates getRoutes())
  - route($route->getName()) → RouteNotFoundException (uses getByName() which checks nameList)

  Reproduction (paste into php artisan tinker):

  $router = app('router');
  $collection = new \Illuminate\Routing\RouteCollection();

  // Simulate what Livewire's boot() does:
  // Route::post(...)->name('default.livewire.update')
  $route = $router->newRoute(['POST'], '/livewire/update', ['uses' => fn() => null]);
  $collection->add($route);                    // add() runs first — name is null at this point
  $route->name('default.livewire.update');      // name set AFTER add(), so nameList misses it

  // findUpdateRoute() uses getRoutes() iteration — WORKS
  $found = collect($collection->getRoutes())
      ->first(fn($r) => str($r->getName())->endsWith('livewire.update'));
  dump($found?->getName());
  // "default.livewire.update" ✓

  // But route() helper uses getByName() — FAILS
  dump($collection->getByName('default.livewire.update'));
  // null ✗

  // The PR fix: $route->uri() works because we already have the valid route object
  dump($found->uri());
  // "livewire/update" ✓

  The fix is safe — getUpdateUri() already holds a valid route object from findUpdateRoute(), so there's no need for a redundant name-based lookup through route(). Using $route->uri() directly eliminates the dependency on nameList being in sync.

joshhanley:
@aydinza thanks for the info! I've pushed a test that covers this scenario and done a deeper analysis of the root cause.

# The Scenario

When using Livewire with Laravel Octane, pages with Livewire components can fail during HTML asset injection with:

```
Route [default.livewire.update] not defined
```

This happens despite `php artisan route:list` showing the route exists correctly. The issue only manifests in specific environments (e.g., production Octane workers) and is difficult to reproduce locally.

# The Problem

`getUpdateUri()` in `HandleRequests` finds the route object via `findUpdateRoute()`, which iterates all routes in the `RouteCollection`. It then passes the route's name to Laravel's `route()` helper to generate the URI:

```php

return (string) str(
)->start('/');
```

The `route()` helper internally calls `RouteCollection::getByName()`, which looks up the route in the `nameList` index — a separate data structure from the `allRoutes` array that `findUpdateRoute()` iterates. When `Route::post(...)->name('default.livewire.update')` is registered, `RouteCollection::add()` runs before the `->name()` fluent method executes, so the route enters `allRoutes` but may not be indexed in `nameList`. Normally `refreshNameLookups()` corrects this, but under Octane worker resets or certain boot ordering scenarios, the two can remain out of sync.

The result is that `findUpdateRoute()` successfully locates the route object, but the redundant name-based lookup via `route()` fails.

# The Solution

Since we already hold a valid route object from `findUpdateRoute()`, use `$route->uri()` directly instead of looking it up again by name:

```php

return (string) str($route->uri())->start('/');
```

This eliminates the dependency on `nameList` being in sync while returning the same URI.

A test has been added that simulates the out-of-sync state by removing the route from the `RouteCollection`'s `nameList` via reflection while keeping it in `allRoutes`, verifying that `getUpdateUri()` still works.

joshhanley (approved):

calebporzio:
## PR Review: #9918 — Fix "Route not defined" error for update route when route name index is stale

**Type**: Bug fix
**Verdict**: Merge

1. Livewire registers its update route during `boot()` via `Route::post(...)->name('default.livewire.update')`
2. `getUpdateUri()` needs to return this route's URI (e.g., `/livewire/update`) for injecting into the page HTML
3. Previously, it found the route object via iteration (`findUpdateRoute()`), then *re-looked it up by name* via Laravel's `route()` helper
4. The `route()` helper uses a separate index (`nameList`) which can get out of sync with the actual route list — particularly under Octane worker resets where `->name()` executes after `RouteCollection::add()` and `refreshNameLookups()` doesn't run
5. Result: the route exists and is findable by iteration, but `route('default.livewire.update')` throws `RouteNotFoundException`
6. Fix: since we already have the route object, call `$route->uri()` directly instead of the redundant name-based lookup

### Changes Made

- Removed verbose 4-line comment above the one-line fix (the code is self-evident)
- Removed unnecessary `setAccessible(true)` call in test (no-op on PHP 8.1+)
- Trimmed test comments for conciseness

### Test Results

All 6 HandleRequests unit tests pass. Verified regression: the new test **fails without the fix** with exactly the reported error (`Route [default.livewire.update] not defined`) and **passes with the fix**.

No CI checks reported on this branch.

### Code Review

- **Fix** (`HandleRequests.php:56`): Clean, surgical one-liner. `$route->uri()` is strictly more reliable than `route($route->getName(), [], false)` since we already hold the route object. No behavior change for the happy path.
- **Test** (`UnitTest.php:88-116`): Well-constructed. Uses reflection to desync `nameList` from `allRoutes`, which is the exact condition that triggers the bug. The `app('url')` pre-resolution prevents the URL generator's lazy boot from undoing the reflection hack — clever and necessary.
- Note: Josh Hanley (maintainer) already reviewed, approved, and pushed the test commit himself.

### Security

No security concerns. This is a read-only URI resolution change with no user input involved.

### Verdict

Merge. The fix is correct, minimal, and well-tested. It eliminates a redundant lookup that was the sole cause of the Octane edge case. The contributor's analysis of the root cause (nameList desync) is accurate and Josh has already approved.

---
*Reviewed by Claude*

calebporzio:
thanks everyone!

edalzell:
Did this ever get into v3 @joshhanley?

calebporzio:
@edalzell - backported and tagged 3.x 👌

edalzell:
> @edalzell - backported and tagged 3.x 👌

ferris buellers day off 80s GIF

PR #9926, opened 2026-02-05
[3.x] Add ddErrors and dumpErrors

AnOlsen:
I have not opened a discussion prior to this, no.

Yup, branch is for 3.x because that's what I currently use, but I can open one for 4.x if desired.

Edit: It seems the PR is targeting main, not sure if I messed something up or if it has to be changed manually.

No

Far as I know this is not testable, would love to know if I'm wrong though.

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why 

This PR adds two methods to the `Testable` class, `ddErrors` and `dumpErrors`, which would be convenient for debugging tests failing due to validation.

Today I added a new feature and a component started failing, and I found myself having to jump through some hoops to find the root cause. A simplified example of my test that failed:

```php
public function test_it_can_submit(): void
{
    Livewire::test(EditComponent::class)
        ->set('state', [
            'first_name' => 'Foo',
            'last_name'  => 'Bar',
            'email'      => 'foobar@example.com'
        ])
        ->call('save')
        ->assertHasNoErrors(); // `Component has errors: "email"`, but not the underlying validation error
}
```

The two methods introduced in this PR is just to make this debugging process easier.

```php
public function test_it_can_submit(): void
{
    // Option 1 (before)
    $errors = Livewire::test(EditComponent::class)
        ->set('state', [
            'first_name' => 'Foo',
            'last_name'  => 'Bar',
            'email'      => 'foo',
        ])
        ->call('save')
        ->errors();
//        ->assertHasNoErrors();
    
    dd($errors);

    // Option 2 (before) after remembering tap() exists
    Livewire::test(EditComponent::class)
        ->set('state', [
            'first_name' => 'Foo',
            'last_name'  => 'Bar',
            'email'      => 'foo',
        ])
        ->call('save')
        ->tap(fn ($component) => dd($component->errors()))
        ->assertHasNoErrors();

    // After
    Livewire::test(EditComponent::class)
        ->set('state', [
            'first_name' => 'Foo',
            'last_name'  => 'Bar',
            'email'      => 'foo',
        ])
        ->call('save')
        ->ddErrors()
        ->assertHasNoErrors();
}
```

All good if you don't want to merge this, I'll just keep tapping, but figured I wasn't the first wishing it was a little easier. :blush:


calebporzio:
## PR Review: #9926 — [3.x] Add ddErrors and dumpErrors

**Type**: Feature
**Verdict**: Needs discussion

1. When `assertHasNoErrors()` fails, it currently says: `Component has errors: "email"` — just the field name, not the actual validation message.
2. The contributor wants `ddErrors()` so they can dump the errors to see what the actual validation message is (e.g., "The email field must be a valid email address").
3. The PR adds `ddErrors()` and `dumpErrors()` to `Testable`.

### The real question

Could a code change eliminate the need for this workaround?

**Yes.** The root cause is that `assertHasNoErrors()` gives a bad failure message. I've pushed a fix that changes the failure message from:

```
Component has errors: "email"
```

to:

```
Component has errors: "email" => ["The email field must be a valid email address."]
```

This addresses the contributor's actual pain point directly. With this fix, most users won't need `ddErrors()` at all — the assertion failure already tells them what went wrong.

### Changes I pushed

1. **Improved `assertHasNoErrors()` failure message** in `TestsValidation.php` to include actual error messages, not just keys.
2. **Simplified `ddErrors()`/`dumpErrors()`** — removed the `$this->getData()` dump. When debugging validation errors, you want errors, not component data. The original code dumped both, which clutters the output.

### Other concerns

- **Branch mismatch**: PR targets `main` (4.x) but title says `[3.x]`. Contributor notes this too.
- **No tests**: Contributor says "not testable" but at minimum `dumpErrors()` could be tested for chainability. That said, `dd`/`dump` methods are inherently side-effect-y and the existing `dd()`/`dump()` methods also have no tests, so this is consistent.
- **Zero community engagement**: No reactions, no linked discussion/issue.

### Test Results

- `src/Features/SupportValidation/UnitTest.php`: 61 tests, 238 assertions — **all pass** (1 skipped, pre-existing)
- `src/Features/SupportTesting/UnitTest.php`: 3 pre-existing failures on `main` (unrelated to this PR)

### Security

### Verdict

The `assertHasNoErrors()` improvement (which I've pushed) is the higher-value change here and should ship regardless. It makes the existing assertion self-documenting and eliminates most of the need for `ddErrors()`.

The `ddErrors()`/`dumpErrors()` methods are fine as convenience sugar — they're tiny, consistent with existing `dd()`/`dump()`, and harmless. But they're also not necessary if the assertion message is good enough.

**Caleb's call**: merge as-is (improved assertion + ddErrors), or strip the ddErrors/dumpErrors and just ship the assertion improvement. Both are reasonable. I'd lean toward keeping them — they're 10 lines and follow existing patterns.

---
*Reviewed by Claude*

calebporzio:
Thanks! yeah i've wanted to know the actual errors myself. hopefully this is an even better solution for you.

AnOlsen:
@calebporzio That's a great solution, I agree! Thanks Caleb. 😊

PR #9929, opened 2026-02-05
[4.x] Docs - Adding new URL endpoints as Low Impact change

TWithers:
Document changes to Livewire asset URLs and endpoints in v4, including the new prefix format and customization options.

Thanks for contributing! 🙌


calebporzio:
## PR Review: #9929 — [4.x] Docs - Adding new URL endpoints as Low Impact change

**Type**: Docs
**Verdict**: Merge

In v4, Livewire changed all its URL paths from `/livewire/...` to `/livewire-{hash}/...` where the hash is derived from `APP_KEY`. This PR adds a note about that change to the upgrade guide so people who have firewall rules, CDN configs, or middleware referencing the old `/livewire/` prefix know to update them.

### Changes Made
Simplified the docs section:
- **Removed the exhaustive route table.** The original listed every single v4 route (9 routes) which was noisy and would go stale as routes change. Replaced with a focused before→after comparison showing the 3 most common endpoints (update, upload, script).
- **Removed inaccurate entries.** The original listed both `livewire.js` and `livewire.min.js.map` side-by-side, but in practice only one script route is registered depending on debug mode. The CSP source map was listed without the CSP script itself.
- **Tightened the prose.** "asset urls" → proper context, removed redundant explanations.
- **Fixed the link text.** Changed from "endpoint customization" to "customizing Livewire's endpoints" to better match the linked section heading.

### Test Results
Docs-only PR, no tests to run. No CI checks reported on the branch.

### Code Review
- `docs/upgrading.md`: The section is correctly placed under "Low-impact changes" — most users won't hit this since they don't reference Livewire URLs directly.
- The link to `/docs/4.x/installation#customizing-livewires-update-endpoint` is valid and the target section exists.
- Note: the linked customization section only covers the update endpoint and JS asset URL. Users who need to customize *all* prefixes (uploads, previews, CSS/JS modules) don't currently have a single config option for that — `EndpointResolver::prefix()` isn't configurable. This is fine for the upgrade guide but worth noting.

### Security

### Verdict
**Merge.** This is a legitimate upgrade guide addition. People with firewall rules or reverse proxy configs targeting `/livewire/` paths will hit a real problem during upgrade, and the existing installation docs don't frame this as a migration concern. The simplified version I pushed communicates the change clearly without the risk of a stale route table.

---
*Reviewed by Claude*

PR #9934, opened 2026-02-06
[4.x] `assertJs` method

gdebrauwer:
Currently, there is no way to test that a livewire component called the `js` method. This PR fixes that.
```php
new class extends Component
{
    public function someMethod()
    {
        $this->js("alert('do something');");
    }
}
```
```php
$component->call('someMethod')->assertJs("alert('do something');");
```


calebporzio:
## PR Review: #9934 — [4.x] `assertJs` method

**Type**: Feature
**Verdict**: Merge

### Summary
Adds `assertJs()` testing helper so users can verify that `$this->js()` was called with the expected expression and params. Clean, well-scoped addition that fills a real gap in the testing API.

### Changes Made
- **Removed multiline whitespace normalization** — The `preg_replace('/\n\s*/', '', ...)` was destructive. It collapsed `alert('\n                    do something\n                ')` into `alert('do something')`, making the assertion pass for fundamentally different JS strings. Whitespace inside string literals matters at runtime. If a user writes multiline JS, they should assert the exact string they wrote.
- **Added `assertNoJs()`** — Follows the existing pattern (`assertNoRedirect`, `assertNotDispatched`). Useful for verifying a code path did *not* trigger JS evaluation.
- **Improved error message** — Changed from generic `"Failed asserting that dispatched JS matches expected JS."` to `"Failed asserting that JS [{$expression}] was evaluated."` to match the events trait pattern and include the expected expression for easier debugging.
- **Added `test_assert_js_with_params`** — The `assertJs` method accepts `...$params` but had no test coverage for that path.
- **Removed `test_assert_js_multiline`** — This test encoded the flawed normalization behavior. It asserted that `alert('do something')` matches `alert('\n                    do something\n                ')` — these are different strings and should not match.

### Test Results
All 3 tests pass (unit): `test_assert_js`, `test_assert_js_with_params`, `test_assert_no_js`. CI was all green except one unrelated Browser 2 flake.

### Code Review
Good work by @gdebrauwer. The core implementation — a `TestsJsEvaluation` trait checking `$this->effects['xjs']` — follows existing patterns exactly (`TestsEvents`, `TestsRedirects`, `TestsFileDownloads`). Registration in `Testable.php` is correct.

### Security

### Verdict
Solid, well-structured contribution. The fixes above address the only issues: flawed multiline normalization, missing negative assertion, missing params test coverage, and a generic error message. Ready to merge.

---
*Reviewed by Claude*

PR #9935, opened 2026-02-06
[4.x] Fix duplicate query for route-bound models on update requests

joshhanley:
# The Scenario

When using `Route::livewire` with route model binding, every Livewire update request queries each route-bound model twice — once from `SubstituteBindings` middleware re-running via `PersistentMiddleware`, and again from `ModelSynth` restoring the model from the snapshot.

| Source | Query |
|---|---|
| `SubstituteBindings` | `select * from "posts" where "id" = '1' limit 1` |
| `ModelSynth` | `select * from "posts" where "posts"."id" = 1 limit 1` |

```php
// routes/web.php
Route::livewire('/posts/{post}', 'show-post');
```

```php
<?php

use App\Models\Post;
use Livewire\Component;

new class extends Component {
    public Post $post;

    public function save(): void {}
};

?>

<div>
    <p>Post: {{ $this->post->title }}</p>
</div>
```

# The Problem

During an update request, `HandleComponents::fromSnapshot()` triggers two independent systems that both query the same models:

1. `trigger('snapshot-verified')` fires `PersistentMiddleware`, which creates a fake request matching the original route and re-runs `SubstituteBindings`. This calls `ImplicitRouteBinding::resolveParameter()` which queries each route-bound model.

2. `hydrateProperties()` calls `ModelSynth::hydrate()`, which creates a lazy proxy that queries each model via `newQueryForRestoration()`.

These two systems have no shared state — the models resolved by `SubstituteBindings` are stored on the route parameters but are never made available to `ModelSynth`, so it independently queries the same models again.

# The Solution

After `PersistentMiddleware` runs `SubstituteBindings`, the resolved model instances are now collected from the route parameters and stored in a lookup table. `ModelSynth::hydrate()` checks this lookup before querying the database — if a matching model (same class and key) has already been resolved by route binding, it reuses that instance instead of making a duplicate query.

```php
// If this model was already resolved by route binding (via
// SubstituteBindings middleware), reuse it to avoid a duplicate query.
$resolvedModel = app(PersistentMiddleware::class)->getResolvedRouteModel($class, $key);

if ($resolvedModel) {
    return $resolvedModel;
}
```

Fixes https://github.com/livewire/livewire/discussions/9924

### DIFF
--- a/src/Features/SupportModels/ModelSynth.php
+++ b/src/Features/SupportModels/ModelSynth.php
@@ -4,6 +4,7 @@ namespace Livewire\Features\SupportModels;
 
 use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
 use Livewire\Mechanisms\HandleComponents\ComponentContext;
+use Livewire\Mechanisms\PersistentMiddleware\PersistentMiddleware;
 use Illuminate\Queue\SerializesAndRestoresModelIdentifiers;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Eloquent\Model;
@@ -75,6 +76,14 @@ class ModelSynth extends Synth {
         }
 
         $key = $meta['key'];
+
+        // If this model was already resolved by route binding (via
+        // SubstituteBindings middleware), reuse it to avoid a duplicate query.
+        $resolvedModel = app(PersistentMiddleware::class)->getResolvedRouteModel($class, $key);
+
+        if ($resolvedModel) {
+            return $resolvedModel;
+        }
 
         return $this->makeLazyProxy($class, $meta, function () use ($class, $key) {
             return (new $class)->newQueryForRestoration($key)->useWritePdo()->firstOrFail();
--- a/src/Features/SupportModels/UnitTest.php
+++ b/src/Features/SupportModels/UnitTest.php
@@ -5,6 +5,8 @@ namespace Livewire\Features\SupportModels;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Support\Facades\Route;
+use Livewire\Drawer\Utils;
 use Livewire\Livewire;
 use Sushi\Sushi;
 use Tests\TestComponent;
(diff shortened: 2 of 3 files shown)


calebporzio:
## PR Review: #9935 — [4.x] Fix duplicate query for route-bound models on update requests

**Type**: Bug fix (performance)
**Verdict**: Merge

### Summary
When using `Route::livewire` with route model binding, every update request queries each route-bound model twice — once from `SubstituteBindings` middleware (via PersistentMiddleware) and again from `ModelSynth` hydration. This PR caches models resolved by route binding and reuses them in `ModelSynth::hydrate()`, eliminating the duplicate query.

### Changes Made

### Test Results
- `test_route_bound_models_are_not_queried_twice_on_update` — **PASSES** ✅
- All 13 ModelSynth unit tests pass (1 pre-existing skip)
- CI: all 22 checks **PASS** ✅
- Verified the test **FAILS on main** (asserts 1 query, gets 2 without the fix)

### Bigger Picture
No deeper changes needed — this PR addresses the right level. The duplicate query is a genuine architectural gap: two independent systems (`PersistentMiddleware` → `SubstituteBindings` and `ModelSynth` → `newQueryForRestoration`) both query the same model with no shared state. The fix introduces a minimal bridge between them.

### Code Review
**What's good:**
- Surgical fix — only 3 files touched, ~30 lines of real logic
- Clean separation: `PersistentMiddleware` collects models after middleware runs (`PersistentMiddleware.php:111-121`), `ModelSynth` checks the cache before querying (`ModelSynth.php:80-86`)
- Proper cleanup on `flush-state` (`PersistentMiddleware.php:55`)
- Test is well-structured: mounts via GET, extracts snapshot, flushes state, POSTs update, asserts query count
- Route-bound models returned eagerly (bypassing lazy proxy) — correct, since they're already loaded
- Naming is clear: `getResolvedRouteModel`, `resolvedRouteModels`, `RouteModelBindingComponent`
- Follows project conventions: no unnecessary visibility keywords, consistent import patterns

**Timing verified:** `snapshot-verified` event fires synchronously before `hydrateProperties`, so `PersistentMiddleware` is guaranteed to have collected route models before `ModelSynth::hydrate()` runs.

### Security
No security concerns. Snapshots are HMAC-SHA256 checksummed before hydration, so `$meta['class']` and `$meta['key']` can't be tampered with. Even if they could be, a cache miss just falls through to the normal query path — no escalation possible.

### Verdict
Clean, minimal, well-tested performance fix from a trusted contributor. The approach is sound architecturally — it bridges two systems that were independently querying the same data. Test properly isolates the root cause (query count, not side effects). CI is fully green. Merge it.

---
*Reviewed by Claude*

calebporzio:
thanks mate!

PR #9936, opened 2026-02-06
[4.x] Fix `wire:transition` not animating correctly when components are swapped

joshhanley:
# The Scenario

When using `wire:transition` with components that get swapped (e.g., via `<livewire:is>`), the old and new content stack on top of each other instead of smoothly transitioning. For example, a parent component that swaps between two child components using transition types:

```php
<?php

use Livewire\Attributes\Transition;
use Livewire\Component;

class CommentSection extends Component
{
    public $current = 'comment-section.comments';
    public $parent = null;

    #[Transition(type: 'forward')]
    public function showReplies(int $id): void
    {
        $this->parent = $id;
        $this->current = 'comment-section.replies';
    }

    #[Transition(type: 'backward')]
    public function showComments(): void
    {
        $this->parent = null;
        $this->current = 'comment-section.comments';
    }
}
```

```blade
<div>
    <livewire:is :component="$current"
        :parent="$parent"
        :wire:key="$current" />
</div>
```

# The Problem

When a component is swapped during a morph (old component removed, new component inserted), the `wire:transition` global directive — which sets the `viewTransitionName` CSS property on elements — is normally initialised by Alpine's `MutationObserver`. However, the observer fires asynchronously as a microtask, and the View Transitions API captures the "new" DOM state before the observer has a chance to run.

This means the newly inserted element's `viewTransitionName` is not set when the new state is captured. Without a matching `viewTransitionName`, the View Transitions API can't pair the old and new snapshots, so the old content's snapshot animates while the new content appears as a raw DOM element — causing them to stack.

# The Solution

After the morph callback runs inside the view transition's `update` function, we now scan the live DOM (`fromEl` — the component's root element that Alpine morphed in place) for any `wire:transition` elements that don't yet have a `viewTransitionName` CSS property set, and apply it synchronously.

This is necessary because `Alpine.morph()` mutates the existing DOM tree in place — newly inserted elements (like a swapped child component's root) are added to `fromEl`'s subtree during the morph. These new elements have the `wire:transition` attribute from the server-rendered HTML, but Alpine's `MutationObserver` — which normally initialises the directive and sets `viewTransitionName` — fires asynchronously and hasn't run yet when the `update` callback returns. Since the View Transitions API captures the new state immediately after the `update` callback, the new elements are missing their `viewTransitionName` and can't be paired with the old snapshots.

The fix is applied in both the standard `startViewTransition` config path and the Firefox fallback path.

Fixes #9931


yehuuu6:
Hey, I’m running into another issue with wire:transition.

When you use wire:transition on elements inside a loop, and the same page also has a modal or a navbar, something weird happens during loading states.

During the animation or loading duration, the transitioning elements temporarily appear in front of the navbar or modal, even though their z-index should be lower. It’s like the z-index isn’t respected while the animation is happening.

I couldn’t find a proper fix for this. To reproduce it, just place a Flux modal on a page along with some wire:transition elements (for example, on a product listing page). Then open the modal using an event that preferably triggers a network request (like wire:click). You should see the issue immediately.

yehuuu6:
https://github.com/user-attachments/assets/4e42689c-1fb1-4521-a261-fc46ef673849

calebporzio:
## PR Review: #9936 — Fix `wire:transition` not animating correctly when components are swapped

**Type**: Bug fix
**Verdict**: Merge

### Summary
When components are swapped via `<livewire:is>`, the View Transitions API captures the new DOM state before Alpine's MutationObserver has a chance to set `viewTransitionName` on newly inserted elements. This PR fixes the timing issue by synchronously setting `viewTransitionName` on new `wire:transition` elements immediately after the morph callback runs.

### Changes Made
No changes made — PR is clean.

### Test Results
- `BrowserTest::test_can_transition_dynamic_component_swap` — pass
- All 3 existing transition browser tests — pass
- CI: all 22 checks green

### Bigger Picture
No deeper changes needed — this PR addresses the right level. The View Transitions API requires `viewTransitionName` to be set synchronously before it captures state. Alpine's async MutationObserver can't satisfy this timing constraint during morphs, so eagerly setting the property in the update callback is the correct and minimal fix.

### Code Review
- `js/directives/wire-transition.js:52-56` — The sync-fix logic is clean: query `fromEl` for `[wire:transition]` elements missing `viewTransitionName`, apply it from the attribute value. Correctly scoped to `fromEl` (the in-place morphed tree).
- `js/directives/wire-transition.js:78-82` — Same logic duplicated for the Firefox fallback path. The duplication is minor (5 lines) and keeps both code paths self-contained. Acceptable.
- `BrowserTest.php:106-222` — Excellent test. Monkeypatches `startViewTransition` to verify the property is set synchronously *within* the update callback, directly targeting the root cause rather than just asserting end-state. Tests both swap directions. Component names are clear.
- No style issues — JS uses no semicolons, `let` not `const`.
- No built assets in the diff.

### Security

### Verdict
This is a well-analyzed timing bug with a surgical fix. The root cause explanation in the PR body is thorough and accurate. The test is unusually good — it doesn't just check that transitions work, it verifies the specific timing guarantee that the fix provides. All CI green. Ready to merge.

---
*Reviewed by Claude*

calebporzio:
## PR Review (updated): #9936 — Fix `wire:transition` not animating correctly when components are swapped

**Type**: Bug fix
**Verdict**: Merge

### Summary
When components are swapped via `<livewire:is>`, the View Transitions API captures the new DOM state before Alpine has initialized `wire:transition` on newly inserted elements. The PR fixes this by synchronously setting `viewTransitionName` within the update callback.

### Root Cause (verified independently)
The issue is a microtask race between Alpine and the View Transitions API:

1. The `update` callback morphs the DOM — new elements with `wire:transition` are inserted
2. Browser queues a MutationObserver microtask
3. Callback returns — browser queues an "activate view transition" microtask
4. MutationObserver fires — but Alpine's `mutation.js:76` calls `queueMicrotask()` for internal batching, deferring actual processing by one more hop
5. "Activate view transition" fires — captures new DOM state. `viewTransitionName` is NOT set yet
6. Alpine's batched processing finally runs — sets `viewTransitionName`. Too late.

Test confirmed: fails on `main`, passes with fix.

### Changes pushed
- Extracted shared `defaultName` constant so the directive and `ensureTransitionNames` stay in sync
- Extracted `ensureTransitionNames` helper to eliminate duplication between the config-object path and Firefox fallback path
- Extracted `update` function shared by both `startViewTransition` call sites
- Rewrote comment to accurately describe the microtask race root cause

### Test Results
All 3 transition browser tests pass after changes.

### Verdict
Solid bug fix targeting the right layer. The contributor correctly identified the timing gap; the underlying mechanism (Alpine's `queueMicrotask` batching creating a two-hop microtask chain) is more subtle than described but the synchronous approach is the correct fix. Test directly verifies the timing guarantee. Ready to merge.

---
*Reviewed by Claude*

calebporzio:
@yehuuu6 - yeah I've heard about some weirdness with wire:transition and top-layer elements like dialog elements. might be a browser limitation. i'll try to explore it in a new branch. thanks

calebporzio:
@yehuuu6 - should be fixed in this PR. think we got it sorted: https://github.com/livewire/livewire/pull/9943

PR #9938, opened 2026-02-06
Suggested clarification for wire-sort documentation for sorting across groups…

adrelliott:
Hey Caleb,

The current documentation has caused several of us some head-scratching.  Whilst it would be blatantly obvious to a lot of developers, some of us had the same question:

'If we move an item across groups, how do we access the id of the target group?"

(see below for discussions on it)

So I added a small note to the section explaining that each group must be it's own component. I also added some example code for handling item sorting in a kanban board.

Please note: I've never done a PR before, so I'm prepared for feedback! 

https://laracasts.com/discuss/channels/livewire/livewire-4-sorting-across-groups?page=1&replyId=974085

https://github.com/livewire/livewire/discussions/9639

Thanks for contributing! 🙌


calebporzio:
## PR Review: #9938 — Suggested clarification for wire-sort documentation for sorting across groups

**Type**: Docs
**Verdict**: Needs discussion

### Summary
Adds a kanban board example to the `wire:sort` cross-group documentation, clarifying that each sortable group should be its own Livewire component so the handler knows which group an item was dropped into. This addresses confusion reported in Laracasts discussions and GitHub discussions.

### Changes Made
Pushed fixes to the example code:
- **Bug: `$this->columnId`** → `$this->column->id` (the component has `public Column $column`, not `$columnId`)
- **Bug: `ToDoItem`** → `TodoItem` (inconsistent casing)
- **Bug: reading `$item->column_id` after update** — saved `$oldColumnId` before the `update()` call so both columns get reordered correctly
- **Bug: `{{ $slot }}` usage** — Livewire components don't receive sortable children as slot content like that. Replaced with a proper `@foreach` loop inside the component's own view
- **Bug: `//` comment in Blade** — invalid syntax; removed
- **Missing `:key`** — added `:key="$column->id"` to the `<livewire:>` tag
- **Prose cleanup** — replaced `NOTE:` callout with prose matching existing doc style

### Test Results
No tests needed — docs-only PR. No CI checks configured on this branch.

### Code Review
The core idea is good: the existing docs explain *what* happens with cross-group sorting but don't explain the practical implication that each group needs to be its own component. That's a genuine gap.

However, the original example code had several bugs that would have confused readers more than helped them (property access errors, stale data reads, incorrect slot usage). I've fixed these.

### Security

### Verdict
The concept is solid and addresses a real documentation gap — multiple community discussions confirm the confusion. However, I'd flag this for Caleb to decide whether this level of detail belongs in the directive reference docs or if it should live in a separate guide/recipe. The `wire:sort` docs are currently concise and pattern-focused; a full kanban example is a different style. Could go either way.

---
*Reviewed by Claude*

adrelliott:
Wow. Good catch Claude! My code was a little buggy. 

And totally see the point that this code example may be more in depth than the docs actually need. 

But the explanation that it just be a separate component is important IMO. 🙂

klmader:
Hey! Just popping in to add support for some sort of addition to the docs on this front. Or, if not needed, I'd love to get clarity on how the current wire:sort:group example from the docs would work. As related discussions have concluded (here and on laracasts) it seems that creating a separate component is necessary. If that's indeed the case, this PR might work best if it also removes the existing wire:sort:group example.

calebporzio:
good call on this. just decided to add wire:sort:group-id, because I don't wanna force people to create separate components when they don't need to. thanks!

adrelliott:
Amazing. This solves the biggest issue of having to pass props from a kanban board, to a column component, to a card. 

You're awesome. Thanks Caleb. 

Ps. Is this live now?

joshhanley:
@adrelliott not yet, it will be available on the next release.

PR #9942, opened 2026-02-07
Fixes lazy loaded components with islands within blade components.

totov:

Fixes a bug – at least I believe the functionality I'm trying to achieve is intended. Here's the architecture:

Lazy loaded livewire component -> Blade component -> Island

In this situation, the island doesn't load regardless of whether it's lazy, deferred or not.

Yeah, new branch.

No, one commit for the test changes, one commit for the fix.

Yes.

Found a bug where if a lazy/deferred Livewire component has blade components within them which then also have islands (lazy, deferred or otherwise), the islands fail to load at all. See the failing test as an example. I can elaborate a bit more if needed.

### DIFF
--- a/src/Features/SupportIslands/BrowserTest.php
+++ b/src/Features/SupportIslands/BrowserTest.php
@@ -4,6 +4,7 @@ namespace Livewire\Features\SupportIslands;
 
 use Tests\BrowserTestCase;
 use Livewire\Livewire;
+use Illuminate\Support\Facades\Blade;
 
 class BrowserTest extends BrowserTestCase
 {
@@ -11,6 +12,7 @@ class BrowserTest extends BrowserTestCase
     {
         return function () {
             app('livewire.finder')->addLocation(viewPath: __DIR__ . '/fixtures');
+            Blade::anonymousComponentPath(__DIR__ . '/fixtures/components');
         };
     }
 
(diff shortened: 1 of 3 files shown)


calebporzio:
## PR Review: #9942 — Fixes lazy loaded components with islands within blade components.

**Type**: Bug fix
**Verdict**: Merge

### Summary
When a lazy-loaded Livewire component contains a Blade component that has an `@island` directive inside it, the island never renders. The fix correctly skips island hydration during the lazy placeholder phase (`lazyLoaded === false`), allowing islands to initialize properly once the parent component actually loads.

### Changes Made

### Test Results
- New test `test_island_inside_blade_component_renders_after_lazy_parent_load`: **passes** (3 assertions)
- Confirmed the test **fails on main** without the fix (island element never appears)
- All 25 existing island browser tests: **pass** (no regressions)
- CI: 21/22 checks pass. The single failure is a flaky `SupportQueryString` timeout — unrelated to this PR.

### Bigger Picture
No deeper changes needed — this PR addresses the right level. The root cause is that `SupportIslands::hydrate()` calls `markIslandsAsMounted()` even during the lazy placeholder phase (when the component hasn't actually mounted yet). This prematurely marks islands as mounted, so when the real lazy load happens, the islands don't get initialized. The 2-line guard clause is the correct and minimal fix.

### Code Review
- `SupportIslands.php:88` — Clean guard clause. Checks `$memo['lazyLoaded'] === false` with null coalescing so it's a no-op for non-lazy components. Correct.
- `BrowserTest.php:326-361` — Good test structure: uses `usleep(250000)` to simulate lazy load delay, asserts island is NOT present before load, then asserts it IS present after. Tests the exact bug scenario.
- `fixtures/components/island-wrapper.blade.php` — Minimal fixture, exactly what's needed.
- No style issues. No built assets committed.

### Security
No security concerns identified. The change only affects the timing of island initialization during hydration.

### Verdict
Clean, surgical bug fix. The test proves the bug, the fix is minimal (2 lines), all existing tests pass, and the approach correctly leverages the existing `lazyLoaded` memo flag. Ready to merge.

---
*Reviewed by Claude*

totov:
@calebporzio I think the failing test actually passed without it being wrapped in a blade component (weirdly). Maybe something weird I was doing locally though. Thanks for taking a look.

calebporzio:
yeah no problem. i think this has nothing to do with the "within blade components" part right? they're just broken with lazy components in general it seems?

PR #9948, opened 2026-02-09
[4.x] Added `wire:navigate.replace` modifiers

sajjadhossainshohag:

No

Yes

No, only related changes.

Yes

This PR introduces browser history replacement support for Livewire navigation and redirects.

It allows developers to `navigate` or `redirect` without pushing a new history entry, aligning Livewire navigation more closely with native browser behavior and real-world SPA needs.

I have added these modifiers.

`wire:navigate.replace`
`wire:navigate.replace.hover`
`wire:navigate.replace.preserve-scroll`
`wire:navigate.replace.preserve-scroll.hover`
`wire:navigate.replace.hover.preserve-scroll`

Also, support for the `redirect` method

 `$this->redirect(route('counter'), navigate: true, replace: true);`
 `$this->redirectRoute('counter', navigate: true, replace: true);`
 `$this->redirectIntended('/', navigate: true, replace: true);`
 
 **Validation Note:**
 
If `replace` is set to `true` while `navigate` is `false`, Livewire will throw an `InvalidArgumentException`.

``` 
$this->redirect('/counter', navigate: false, replace: true); // ❌ Invalid
```

**Reason**

`replace` controls how browser history is handled during client-side navigation.
Without navigation enabled, there is no browser history state to replace.

To use `replace`, `navigate` must be explicitly enabled:

```
$this->redirect('/counter', navigate: true, replace: true); // ✅ Valid
```

**No breaking changes.**


calebporzio:
## PR Review: #9948 — [4.x] Added `wire:navigate.replace` modifiers

**Type**: Feature
**Verdict**: Needs discussion

1. Currently, every `wire:navigate` click pushes a new entry onto the browser history stack (`history.pushState`).
2. This PR adds a `.replace` modifier so `wire:navigate.replace` uses `history.replaceState` instead — the current history entry gets swapped rather than a new one being added.
3. Practical use case: form submission → success page. Without `.replace`, hitting "back" takes the user back to the form. With `.replace`, the form page is removed from history.
4. The infrastructure already existed internally — `navigateTo()` had a `shouldPushToHistoryState` param used for popstate (back/forward button) handling, and `replaceUrl()` was already imported. The PR just wires this up to user-facing APIs.
5. Three touch points: (a) `wire:navigate.replace` modifier in the directive, (b) `Livewire.navigate(url, { replace: true })` in JS, (c) `$this->redirect($url, navigate: true, replace: true)` in PHP.

1. **Just use `window.history.replaceState()` manually** — Users could call this after navigation, but it's a timing nightmare with Alpine's async navigate pipeline. The in-framework approach is the right call.
2. **Handle at the Alpine level only** — Since Livewire bundles its own navigate plugin, this is effectively the same thing. No Alpine boundary concern here.
3. **Middleware-based approach** — Could set a header/meta tag to signal replace behavior, but that's over-engineered for what's essentially a one-line option pass-through.

### Changes Made
- Removed unused `RuntimeException` import from `HandlesRedirects.php`
- Removed the `InvalidArgumentException` validation for `replace` without `navigate`. The validation is unnecessary — `replace` is silently ignored when `navigate` is false since the JS only uses the `replace` effect in the `Alpine.navigate()` path. Throwing an exception for a no-op condition is surprising behavior. If someone passes `replace: true, navigate: false`, nothing bad happens.
- dist/ files were already handled via merge conflict resolution (restored from main)

### Test Results
- 5 browser tests: all pass ✅
- Verified regression: all 5 tests fail on `main` without the feature code (4 assertion failures + 1 timeout) ✅
- Existing `SupportNavigate/BrowserTest.php` tests still pass ✅
- No CI checks have run yet

### Code Review

**Positives:**
- The actual feature logic is minimal and clean — about 5 lines of real logic across 3 files
- Tests are well-structured: they verify history length, back-button behavior, and cover directive, redirect, preserve-scroll, hover-prefetch, and JS API paths
- Correctly uses the existing `replaceUrl()` infrastructure rather than reimplementing

**Concerns:**

1. **Zero community demand.** No linked issue, no discussion, no reactions on the PR. The contribution guide asks to create a discussion first. This is a reasonable feature (Inertia has it, it's a real SPA pattern), but the bar is higher without demonstrated demand.

2. **No documentation.** A feature PR needs docs. No `docs/wire-navigate.md` update, no `docs/__nav.md` entry, nothing explaining the `.replace` modifier or the `replace` parameter on redirect methods.

3. **Massive formatting diff.** The `navigate/index.js` diff is ~250 lines but the actual logic change is ~5 lines. The contributor ran prettier on the entire file. The formatting does match the project's `.prettierrc`, so it's arguably correct, but it makes the PR very hard to review. Ideally this would be a separate commit or PR.

4. **The `wire-navigate.js` selector explosion.** Adding `.replace` doubles the selector/attributeMap permutations from 5 to 10. The next modifier will make it 20. This is a pre-existing architectural problem (not introduced by this PR), but each new modifier makes it worse. Worth considering a dynamic approach.

5. **Effect key naming.** The PR uses `replace` as the effect key name, which is very generic. The store key is `redirectReplace` — the effect key should probably match for consistency (or at least be more specific like `redirectReplace`).

### Security
No security concerns identified. The `replace` option only affects browser history state management — it doesn't change what URLs can be navigated to or bypass any security checks.

### Verdict
**Needs discussion.** The feature itself is sound and well-implemented — `history.replaceState` during navigate is a legitimate SPA pattern. But:

- No community demand or prior discussion
- No documentation included
- The formatting noise makes the diff 50x larger than the actual change

Caleb, the key question: **Is this a feature you want in Livewire?** If yes, this PR is a solid implementation and I can add docs. If you're on the fence, the lack of community signal suggests "no for now" — it can always be added later when there's demand.

---
*Reviewed by Claude*

calebporzio:
think I'm going to pass on this for now. we can keep our eye on it for the future. thanks!

PR #9950, opened 2026-02-09
[4.x] Add missing stubs to `livewire:stubs` command

joshhanley:
# The Scenario

When running `php artisan livewire:stubs` to customise the stubs used by `make:livewire`, only the legacy stubs are published. The newer stubs for single-file components, multi-file components, and layouts are missing — so users can't customise them.

```bash
php artisan livewire:stubs
# Only publishes 7 of 14 stubs
```

# The Problem

The `StubsCommand` was last updated when Form and Attribute stubs were added, but it was never updated to include the stubs added since then. The `MakeCommand` and `LayoutCommand` both already check the user's `stubs/` directory for overrides (via `getStubPath()` and `stubPath()` respectively), but since these newer stubs are never published there, users have no way to customise them.

The missing stubs:

- `livewire.layout.stub`
- `livewire-sfc.stub`
- `livewire-mfc-class.stub`
- `livewire-mfc-view.stub`
- `livewire-mfc-test.stub`
- `livewire-mfc-js.stub`
- `livewire-mfc-css.stub`

# The Solution

Added the 7 missing stubs to the `StubsCommand` so that all 14 stubs are now published when running `php artisan livewire:stubs`.

Fixes #9939

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/StubsCommand.php
+++ b/src/Features/SupportConsoleCommands/Commands/StubsCommand.php
@@ -56,6 +56,41 @@ class StubsCommand extends Command
             file_get_contents(__DIR__.'/livewire.attribute.stub')
         );
 
+        file_put_contents(
+            $stubsPath.'/livewire.layout.stub',
+            file_get_contents(__DIR__.'/livewire.layout.stub')
+        );
+
+        file_put_contents(
+            $stubsPath.'/livewire-sfc.stub',
+            file_get_contents(__DIR__.'/livewire-sfc.stub')
+        );
+
+        file_put_contents(
+            $stubsPath.'/livewire-mfc-class.stub',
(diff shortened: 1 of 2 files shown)


Andyuu:
Very nice :)

Btw there is also a stubs unit test file that also hasn't been updated

joshhanley:
@Andyuu I thought I had all the stub files covered, which one is missing exactly?

Andyuu:
https://github.com/livewire/livewire/blob/750c9c04175808462d9dbfe518b8b6b5ce65f7ab/src/Features/SupportConsoleCommands/Tests/StubCommandUnitTest.php

This test here

joshhanley:
@Andyuu oh sorry! I thought you meant like a unit test stub haha.

joshhanley:
Have updated the tests.

PR #9960, opened 2026-02-10
[4.x] Update custom route docs to use EndpointResolver::updatePath() instead of hardcoded paths

joshhanley:
# The Scenario

When customizing Livewire's update endpoint using `Livewire::setUpdateRoute()`, the documentation examples use hardcoded paths like `/custom/livewire/update` and `/livewire/update`:

```php
// From docs/installation.md
Route::post('/custom/livewire/update', $handle);

// From docs/security.md
Route::post('/livewire/update', $handle);
```

Livewire v4 introduced hash-based endpoints to make installations harder to identify with automated scanners, but users following these examples would bypass that protection.

# The Problem

The `EndpointResolver` class was added in v4 to generate hash-based paths like `/livewire-{hash}/update`, but the documentation examples for `setUpdateRoute()` still referenced the old predictable paths from v3.

# The Solution

Updated the code examples in `docs/installation.md`, `docs/security.md`, and `docs/upgrading.md` to use `EndpointResolver::updatePath()`:

- **installation.md**: Both examples now use `/custom' . EndpointResolver::updatePath()` to demonstrate a custom route prefix while preserving the hash-based path
- **security.md**: The middleware example now uses `EndpointResolver::updatePath()` instead of `/livewire/update`
- **upgrading.md**: Added a before/after migration example in the "Livewire Asset and Endpoint URL Changes" section for users who have an existing `setUpdateRoute()` call

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -118,32 +118,30 @@ Livewire.start()
 
 ## Customizing Livewire's update endpoint
 
-**When you need this:** If your application uses route prefixes for localization (like `/en/`, `/fr/`) or multi-tenancy (like `/tenant-1/`, `/tenant-2/`), you may need to customize Livewire's update endpoint to match your routing structure.
+**When you need this:** If your application uses route prefixes for localisation (like `/en/`, `/fr/`) or multi-tenancy (like `/tenant-1/`, `/tenant-2/`), or you need to apply additional middleware to all Livewire requests.
 
-By default, Livewire sends component updates to a hash-based endpoint like `/livewire-{hash}/update`, where `{hash}` is derived from your application's `APP_KEY`. To customize this, register your own route in a service provider (typically `App\Providers\AppServiceProvider`):
+By default, Livewire sends component updates to a hash-based endpoint like `/livewire-{hash}/update`, where `{hash}` is derived from your application's `APP_KEY`. This unique-per-installation path makes it harder to target Livewire applications with automated scanners.
+
+To customise this, register your own route in a service provider (typically `App\Providers\AppServiceProvider`):
 
 ```php
 use Livewire\Livewire;
+use Livewire\Mechanisms\HandleRequests\EndpointResolver;
 
 class AppServiceProvider extends ServiceProvider
 {
     public function boot()
     {
         Livewire::setUpdateRoute(function ($handle) {
-            return Route::post('/custom/livewire/update', $handle);
+            return Route::post(EndpointResolver::updatePath(), $handle)
+                ->middleware(['web', 'auth']);
         });
     }
 }
 ```
 
-You can also add middleware to the update route:
-
-```php
-Livewire::setUpdateRoute(function ($handle) {
-    return Route::post('/custom/livewire/update', $handle)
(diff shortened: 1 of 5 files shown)


calebporzio:
## PR Review: #9960 — [4.x] Update custom route docs to use EndpointResolver::updatePath()

**Type**: Docs
**Verdict**: Needs discussion

1. Livewire v4 introduced hash-based endpoint paths (e.g., `/livewire-a1b2c3d4/update`) via `EndpointResolver` to make installations harder to fingerprint.
2. The docs examples for `setUpdateRoute()` still showed hardcoded paths like `/custom/livewire/update` from v3 — so anyone following the docs would bypass the hash protection.
3. This PR updates the docs to use `EndpointResolver::updatePath()` instead of hardcoded paths.

1. **Pass the path to the callback (code fix)** — Change `setUpdateRoute()` to pass `EndpointResolver::updatePath()` as a second argument: `function ($handle, $path)`. Users wouldn't need to import anything — the framework hands them the correct path. This is the laziest correct solution and eliminates the docs problem at the source.
2. **Add a facade method** — `Livewire::updatePath()` as a clean public API. Avoids exposing `Mechanisms\HandleRequests` internals to userland code.
3. **This PR's approach (docs-only)** — Tells users to import `Livewire\Mechanisms\HandleRequests\EndpointResolver` directly. Works, but leaks internal namespace structure into userland.

### Changes Made
No changes made — shell execution environment became unavailable during review.

### Test Results
No tests to run (docs-only PR). No CI checks reported on this branch.

### Code Review
The docs changes are factually correct — `EndpointResolver::updatePath()` returns the right path. However:

- **Leaky abstraction**: All three files (`docs/installation.md`, `docs/security.md`, `docs/upgrading.md`) ask users to import from `Livewire\Mechanisms\HandleRequests\EndpointResolver`. The `Mechanisms` namespace is an internal implementation detail. Livewire's public API surface is clean (`Livewire\Livewire`, `Livewire\Attributes\*`, etc.) — asking users to reach into `Mechanisms\HandleRequests` breaks that pattern.
- **A code change could eliminate this docs workaround entirely.** Passing the update path as a second argument to the `setUpdateRoute` callback is a ~1-line change in `HandleRequests.php:92` that makes the docs simpler AND nudges users toward the hashed path by default.

### Security
The underlying concern (hardcoded paths bypassing hash-based endpoint obfuscation) is valid. The fix direction is correct — the question is just about the right mechanism.

### Verdict
**Don't merge as-is.** The problem is real, but a ~1-line code fix in `HandleRequests::setUpdateRoute()` is better than making users import an internal class. Specifically, change:

```php
$route = $callback([self::class, 'handleUpdate']);
```

to:

```php
```

Then the docs become:

```php
Livewire::setUpdateRoute(function ($handle, $path) {
    return Route::post('/custom' . $path, $handle)
        ->middleware(['web', 'auth']);
});
```

No import needed. Framework hands you the correct path. Docs stay clean.

@joshhanley — Great catch on the security gap. Would you be open to adding the code fix instead? The docs changes can then reference `$path` rather than `EndpointResolver`.

---
*Reviewed by Claude*

PR #9961, opened 2026-02-10
[4.x] Expand SecurityPolicy denylist

joshhanley:
# The Scenario

Following Synacktiv's public disclosure of CVE-2025-54068 and associated exploitation tooling, the exploit chain documentation names several classes that can be leveraged to achieve remote code execution when an attacker has a valid `APP_KEY`. These classes were not on Livewire's synthesiser denylist.

# The Problem

The `SecurityPolicy` denylist only blocked 7 classes. Some of the classes explicitly used in the Synacktiv exploit chain were missing:

- `GuzzleHttp\Psr7\FnStream` — provides a `__toString` to arbitrary function call gadget
- `League\Flysystem\UrlGeneration\ShardedPrefixPublicUrlGenerator` — provides an `array_map` trigger gadget
- `Illuminate\Broadcasting\BroadcastEvent` — provides `dispatchNextJobInChain()` to `unserialize()` (notably, only `PendingBroadcast` was blocked, but `BroadcastEvent` is the class actually instantiated in the exploit chain)
- `Laravel\Prompts\Terminal` — used for clean `exit()` to avoid error logging
- `Illuminate\Queue\Jobs\Job` — queue job execution base class
- `Illuminate\Mail\Mailable` — could be used to send arbitrary emails

# The Solution

Added all 6 classes to the `SecurityPolicy` denylist in `src/Mechanisms/HandleComponents/SecurityPolicy.php`. Since `validateClass()` uses `is_a()` with `allow_string: true`, this also blocks any subclasses of these entries (e.g. any concrete queue job class is blocked via the `Job` base class entry).

Added unit tests verifying each new denylist entry is rejected, and updated the existing `test_get_denied_classes_returns_list` test to assert the new entries are present.

### DIFF
--- a/src/Mechanisms/HandleComponents/SecurityPolicy.php
+++ b/src/Mechanisms/HandleComponents/SecurityPolicy.php
@@ -19,13 +19,23 @@ class SecurityPolicy
 
         // Known serialization gadgets
         'Illuminate\Broadcasting\PendingBroadcast',
+        'Illuminate\Broadcasting\BroadcastEvent',
         'Illuminate\Foundation\Testing\PendingCommand',
 
         // Queue jobs - could execute arbitrary code
         'Illuminate\Queue\CallQueuedClosure',
+        'Illuminate\Queue\Jobs\Job',
+
+        // Mail - could send arbitrary emails
+        'Illuminate\Mail\Mailable',
 
         // Notifications - could send arbitrary notifications
         'Illuminate\Notifications\Notification',
+
+        // Known exploit chain gadgets (Synacktiv disclosure)
+        'GuzzleHttp\Psr7\FnStream',
+        'League\Flysystem\UrlGeneration\ShardedPrefixPublicUrlGenerator',
+        'Laravel\Prompts\Terminal',
     ];
 
     /**
--- a/src/Mechanisms/HandleComponents/SecurityPolicyUnitTest.php
+++ b/src/Mechanisms/HandleComponents/SecurityPolicyUnitTest.php
@@ -55,6 +55,54 @@ class SecurityPolicyUnitTest extends \Tests\TestCase
         SecurityPolicy::validateClass(\DateTimeImmutable::class);
     }
 
+    public function test_rejects_broadcast_event_class()
+    {
+        $this->expectException(\Exception::class);
+        $this->expectExceptionMessage('is not allowed to be instantiated');
+
+        SecurityPolicy::validateClass(\Illuminate\Broadcasting\BroadcastEvent::class);
+    }
+
+    public function test_rejects_queue_job_class()
+    {
+        $this->expectException(\Exception::class);
+        $this->expectExceptionMessage('is not allowed to be instantiated');
+
+        SecurityPolicy::validateClass(\Illuminate\Queue\Jobs\Job::class);
+    }
+
+    public function test_rejects_mailable_class()
+    {
+        $this->expectException(\Exception::class);
+        $this->expectExceptionMessage('is not allowed to be instantiated');
+
+        SecurityPolicy::validateClass(\Illuminate\Mail\Mailable::class);
+    }
+
+    public function test_rejects_guzzle_fn_stream_class()
+    {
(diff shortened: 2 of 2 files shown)


calebporzio:
## PR Review: #9961 — [4.x] Expand SecurityPolicy denylist

**Type**: Security hardening
**Verdict**: Merge

1. Livewire hydrates component state from JSON payloads. The payload includes `meta['class']` which tells synthesizers what class to instantiate (e.g., `CollectionSynth` does `new $meta['class']($value)`).
2. The primary defense is HMAC-SHA256 checksum verification — an attacker can't tamper with the payload without the `APP_KEY`.
3. If the `APP_KEY` is compromised, an attacker could craft payloads that instantiate dangerous classes through at least 5 synthesizers (`CollectionSynth`, `EnumSynth`, `ModelSynth`, `EloquentCollectionSynth`, `WireableSynth`) that use `meta['class']` for instantiation.
4. The `SecurityPolicy` denylist is the defense-in-depth layer — it blocks known-dangerous classes before any synthesizer touches them.
5. This PR adds 6 classes to that denylist, specifically ones named in the Synacktiv CVE-2025-54068 exploit chain that weren't previously blocked.

1. **Allowlist instead of denylist**: Each synthesizer could validate that `meta['class']` is a type it's designed to handle (e.g., `CollectionSynth` only accepts `Collection` subclasses, `EnumSynth` only accepts `BackedEnum` subclasses). This would be *far* more robust than a denylist since it blocks unknown future gadgets too. However, it's a bigger architectural change — worth pursuing separately, doesn't block this PR.
2. **Do nothing**: The HMAC checksum is the real protection. The denylist only matters if the `APP_KEY` leaks. But defense-in-depth is good security practice, and a leaked `APP_KEY` is a realistic scenario.
3. **Block at the synthesizer level**: Some synthesizers already validate (e.g., `WireableSynth` checks `is_a(Wireable)`, `FormObjectSynth` checks `is_a(Form)`). But `CollectionSynth` and `EnumSynth` don't — the centralized denylist is the right place for this until an allowlist approach is implemented.

The PR's approach (expanding the centralized denylist) is the right call for an immediate fix. The allowlist conversation should happen separately.

### Changes Made

No changes made. Code is clean as-is.

### Test Results

- **Regression verified**: Reverted `SecurityPolicy.php` to `main` version → all 6 new tests fail (expected exception not thrown). Restored PR version → all 12 tests pass. ✅
- **CI**: All 22 checks pass (Unit, Browser, Legacy across PHP 8.4/8.5 and L11/L12). ✅

### Code Review

- **`SecurityPolicy.php`**: Clean additions. Good use of `is_a()` with `allow_string: true` — the `Job` base class entry blocks all concrete queue job subclasses. The `Mailable` entry similarly blocks all mailable subclasses. Smart use of inheritance for broad coverage.
- **Tests**: Follow existing one-test-per-class pattern. Third-party classes (`FnStream`, `ShardedPrefixPublicUrlGenerator`) correctly use string literals instead of `::class` since those packages may not be installed in the test environment.
- **No style issues**: PHP follows project conventions. No JS changes.
- **No built assets**: Clean diff.

### Security

This PR is purely additive to security. Zero regression risk — it only adds entries to a static denylist array. The 6 classes are well-chosen: `BroadcastEvent` (was missing alongside `PendingBroadcast`), `Job` (broad base class coverage), `Mailable` (arbitrary email), and 3 specific Synacktiv gadget chain classes.

**Architectural note for Caleb**: The denylist approach is fundamentally reactive. Long-term, consider an allowlist where each synthesizer declares what classes it handles. The synthesizer set is bounded (~15 synthesizers), and most already have natural type constraints. `CollectionSynth` and `EnumSynth` are the main ones that would benefit from explicit class validation.

### Verdict

Merge. This is a clean, well-tested, zero-risk security hardening PR. The additions are precisely targeted at a published exploit chain. Tests are correct and verified to fail without the fix. No style issues, no built assets, no unnecessary complexity.

---
*Reviewed by Claude*

PR #9964, opened 2026-02-10
[4.x] Block lifecycle methods from being called via frontend calls

joshhanley:
# The Scenario

Livewire components can define lifecycle methods like `mount()`, `boot()`, `hydrate()`, and property hooks like `updatingFoo()` / `updatedFoo()`. These methods are intended to run automatically at specific points in the component lifecycle — they should never be callable from the frontend via `$wire.mount()` or similar.

# The Problem

Lifecycle methods defined on user subclasses pass the `getPublicMethodsDefinedBySubClass()` filter in `callMethods()`, since that filter only excludes methods from the base `Component` class. The existing `SupportLifecycleHooks::call()` hook blocked some lifecycle methods (`mount`, `exception`, `hydrate*`, `dehydrate*`, `updating*`, `updated*`) but was missing `boot`, `booted`, `rendering`, and `rendered`.

An attacker (or curious user) could call `$wire.boot()`, `$wire.rendering()`, or `$wire.rendered()` to execute these hooks out of sequence, potentially causing unexpected behaviour.

# The Solution

Extended the blocklist in `SupportLifecycleHooks::call()` to cover all lifecycle methods: `mount`, `boot`, `booted`, `rendering`, `rendered`, and `exception` (exact matches), plus `hydrate*`, `dehydrate*`, `updating*`, `updated*` (wildcard prefixes, unchanged from before).

For trait-suffixed lifecycle hooks (e.g. `mountWithFileUploads`, `bootMyTrait`), the blocklist is built dynamically by checking the component's actual traits via `class_uses_recursive()`. This ensures methods like `mountAction()` — legitimate user methods that happen to share a lifecycle prefix — are not falsely blocked.

Added a defense-in-depth check (`isLifecycleMethod()`) directly in `callMethods()` as a secondary guard, using the same trait-aware logic. It throws `MethodNotFoundException` — intentionally not revealing whether the method exists.

### DIFF
--- a/src/Features/SupportLifecycleHooks/SupportLifecycleHooks.php
+++ b/src/Features/SupportLifecycleHooks/SupportLifecycleHooks.php
@@ -99,12 +99,16 @@ class SupportLifecycleHooks extends ComponentHook
     public function call($methodName, $params, $returnEarly, $metadata)
     {
         $protectedMethods = [
-            'mount',
+            'mount*',
+            'boot*',
+            'booted*',
             'exception',
             'hydrate*',
             'dehydrate*',
             'updating*',
             'updated*',
+            'rendering',
+            'rendered',
             'scriptSrc',
         ];
 
--- a/src/Features/SupportLifecycleHooks/SupportLifecycleHooks.php
+++ b/src/Features/SupportLifecycleHooks/SupportLifecycleHooks.php
@@ -99,9 +99,9 @@ class SupportLifecycleHooks extends ComponentHook
     public function call($methodName, $params, $returnEarly, $metadata)
     {
         $protectedMethods = [
-            'mount*',
-            'boot*',
-            'booted*',
+            'mount',
+            'boot',
+            'booted',
             'exception',
             'hydrate*',
             'dehydrate*',
@@ -111,6 +111,20 @@ class SupportLifecycleHooks extends ComponentHook
             'rendered',
             'scriptSrc',
         ];
+
+        // Also block trait-suffixed lifecycle hooks (e.g. mountWithFileUploads, bootMyTrait)
+        $class = get_class($this->component);
+
+        if (! isset(static::$traitCache[$class])) {
+            static::$traitCache[$class] = class_uses_recursive($this->component);
+        }
+
+        foreach (static::$traitCache[$class] as $trait) {
+            $traitBasename = class_basename($trait);
+            $protectedMethods[] = 'mount'.$traitBasename;
+            $protectedMethods[] = 'boot'.$traitBasename;
+            $protectedMethods[] = 'booted'.$traitBasename;
+        }
 
         throw_if(
             str($methodName)->is($protectedMethods),
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -633,14 +633,6 @@ class HandleComponents extends Mechanism
             // @todo: put this in a better place:
             $methods[] = '__dispatch';
 
-            // Block lifecycle methods from being called via the frontend.
-            // These are defined on user subclasses so they pass the
-            // getPublicMethodsDefinedBySubClass filter, but they should
-            // never be invocable via $wire — they are internal hooks.
-            if (static::isLifecycleMethod($method, $root)) {
-                throw new MethodNotFoundException($method);
-            }
-
             if (! in_array($method, $methods)) {
                 throw new MethodNotFoundException($method);
             }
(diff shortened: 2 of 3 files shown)


calebporzio:
## PR Review: #9964 — [4.x] Block lifecycle methods from being called via frontend calls

**Verdict**: Merge

1. Livewire components can define lifecycle methods like `boot()`, `booted()`, `rendering()`, `rendered()` — these run automatically at specific points (hydration, render cycle, etc.)
2. There's already a guard in `SupportLifecycleHooks::call()` that blocks frontend calls (`$wire.mount()`, `$wire.hydrate()`, etc.) by checking method names against a blocklist
3. **The bug**: That blocklist was missing `boot`, `booted`, `rendering`, and `rendered`. These methods pass the `getPublicMethodsDefinedBySubClass()` filter since they're defined on user subclasses, so an attacker could call `$wire.boot()` or `$wire.rendering()` from the frontend to execute these hooks out of sequence
4. Additionally, trait-suffixed lifecycle hooks like `mountMyTrait` and `bootMyTrait` weren't blocked at all — the existing wildcards (`hydrate*`, `updating*`, etc.) only covered property-style hooks
5. This PR adds the missing methods to the blocklist and adds trait-aware blocking for `mount`, `boot`, and `booted` suffixes

1. **Just add the 4 missing names to the flat blocklist** — Simplest, but doesn't cover trait-suffixed variants like `bootMyTrait`. These are also callable from the frontend and should be blocked.
2. **Block all `mount*`/`boot*` prefixes with wildcards** — Would catch trait variants but also false-positive on legitimate user methods like `mountAction()` or `bootstrap()`. The PR's trait-aware approach (only block `mountFoo` if `Foo` is an actual trait) is correct.
3. **Defense-in-depth with a second guard in `callMethods()`** — The PR originally included a 57-line `isLifecycleMethod()` static method in `HandleComponents` as a secondary guard. I removed this because the `trigger('call', ...)` invocation in `callMethods` fires synchronously, and `SupportLifecycleHooks::call()` throws before execution ever reaches the secondary check — making it dead code that would diverge over time (it already diverged: `scriptSrc` was only in one list). The hook system IS the architecture — if someone disables the hook, Livewire is broken anyway.

### Changes Made

- **Removed** the redundant `isLifecycleMethod()` static method from `HandleComponents.php` and its call in `callMethods()` (-59 lines)
- **Removed** 7 unit tests that tested the removed static method
- **Added** missing `test_cannot_call_booted_from_frontend` integration test (was untested!)
- **Added** missing `test_cannot_call_trait_booted_variant_from_frontend` test
- **Added** `booted()` to `LifecycleMethodStub` fixture

### Test Results

- 21 tests pass in `ComponentsAreSecureUnitTest.php`
- Regression verified: reverted fix files to `main`, confirmed `boot`, `rendering`, `rendered` tests FAIL without the fix, and `mount` (already blocked) PASSES
- Trait-suffixed tests (`mountStubTrait`, `bootStubTrait`) also confirmed to fail without the fix
- CI: All 22 checks passing (Unit, Browser, Legacy across PHP 8.4/8.5, Laravel 11/12)

### Code Review

- `SupportLifecycleHooks.php:101-127` — The blocklist extension is clean and correct. The trait-aware logic uses `class_uses_recursive()` with the existing `$traitCache` for performance. Good.
- `SupportLifecycleHooks.php:130` — Uses `str($methodName)->is($protectedMethods)` which handles wildcard matching via Laravel's `Str::is()`. This is case-sensitive. The old `isLifecycleMethod()` was case-insensitive via `strtolower()`. However, PHP method calls are case-insensitive at the language level, so `$wire.Mount()` would still resolve to the method — but `str()->is('mount')` would NOT match `Mount`. This means case-variant attacks could bypass the hook guard. **This is a pre-existing issue on `main`** (the original blocklist also used `str()->is()` without lowercasing), not introduced by this PR.
- The `mountAction` false-positive prevention is smart and well-tested.

### Security

- This is a security fix. The missing lifecycle methods (`boot`, `booted`, `rendering`, `rendered`) could be called out-of-sequence from the frontend. While the practical impact is limited (these methods typically set up state or modify view output), allowing arbitrary lifecycle hook invocation violates the security model.
- **Pre-existing concern**: The `str()->is()` check is case-sensitive but PHP methods are case-insensitive. `$wire.Boot()` would bypass the guard. This exists on `main` already — worth a follow-up fix (lowercase `$methodName` before checking).

### Verdict

Merge. This is a legitimate security fix with good test coverage. The core change is surgical — extending an existing blocklist with missing entries and adding trait-aware blocking. The trait-aware approach is the right call to avoid false positives on user methods. I simplified the PR by removing the redundant defense-in-depth layer and added two missing test cases.

---
*Reviewed by Claude*

PR #9965, opened 2026-02-10
[4.x] Require `X-Livewire` and JSON content type on update requests

joshhanley:
# The Scenario

The Livewire update endpoint processes any POST request regardless of its headers. The legitimate Livewire JS client always sends both `X-Livewire: 1` and `Content-Type: application/json` on every update request, but the server never validates their presence.

# The Problem

Without header validation, the update endpoint is more exposed than necessary to non-client requests. Browsers can send cross-origin `application/x-www-form-urlencoded` POST requests without a CORS preflight check, which means a malicious page could potentially submit form-encoded POST requests to the Livewire endpoint. While Laravel's CSRF middleware provides the primary defence, requiring JSON content type adds defence-in-depth — browsers will trigger a CORS preflight for cross-origin `application/json` requests, providing an additional layer of protection.

Additionally, any POST request (from scanners, bots, or manual probing) reaches the handler logic regardless of whether it originated from the Livewire JS client.

# The Solution

Added a `RequireLivewireHeaders` middleware that rejects requests missing either the `X-Livewire` header or a JSON `Content-Type` header, returning a 404 response to avoid confirming the endpoint exists to scanners.

The middleware is registered on the default update route and is also automatically applied to custom routes defined via `setUpdateRoute()`.

The check uses Laravel's `$request->isJson()` method which matches any content type containing `/json` or `+json`, so it handles standard `application/json` as well as variants like `application/json; charset=utf-8`.

Since the legitimate JS client always sends both headers, this change will never affect real users.

### DIFF
--- a/src/Features/SupportModels/UnitTest.php
+++ b/src/Features/SupportModels/UnitTest.php
@@ -273,15 +273,16 @@ class UnitTest extends \Tests\TestCase
         Article::resolveConnection()->flushQueryLog();
 
         // POST to the update endpoint to trigger an update request
-        $this->post(app('livewire')->getUpdateUri(), [
-            'components' => [
-                [
-                    'snapshot' => $encodedSnapshot,
-                    'calls' => [['method' => '$refresh', 'params' => [], 'path' => '']],
-                    'updates' => [],
+        $this->withHeaders(['X-Livewire' => 'true'])
+            ->postJson(app('livewire')->getUpdateUri(), [
+                'components' => [
+                    [
+                        'snapshot' => $encodedSnapshot,
+                        'calls' => [['method' => '$refresh', 'params' => [], 'path' => '']],
+                        'updates' => [],
+                    ],
                 ],
-            ],
-        ])->assertOk();
+            ])->assertOk();
 
         // Count queries to the articles table — should only be one, not two
         $queryLog = Article::resolveConnection()->getQueryLog();
--- a/src/Features/SupportTesting/SubsequentRender.php
+++ b/src/Features/SupportTesting/SubsequentRender.php
@@ -33,7 +33,7 @@ class SubsequentRender extends Render
 
         [$response, $componentInstance, $componentView] = $this->extractComponentAndBladeView(function () use ($uri, $payload, $cookies) {
             return $this->requestBroker->temporarilyDisableExceptionHandlingAndMiddleware(function ($requestBroker) use ($uri, $payload, $cookies) {
-                return $requestBroker->addHeaders(['X-Livewire' => true])->call('POST', $uri, $payload, $cookies);
+                return $requestBroker->addHeaders(['X-Livewire' => true, 'Content-Type' => 'application/json'])->call('POST', $uri, [], $cookies, [], [], json_encode($payload));
             });
         });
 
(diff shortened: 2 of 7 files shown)


calebporzio:
## PR Review: #9965 — [4.x] Require `X-Livewire` and JSON content type on update requests

**Type**: Security hardening (defence-in-depth)
**Verdict**: Merge

1. Livewire's JS client always sends two headers on every update request: `X-Livewire: 1` and `Content-Type: application/json`.
2. The server endpoint never checked for these headers — any POST request (bots, scanners, cross-origin form submissions) reached `handleUpdate()`.
3. This matters because browsers can send `application/x-www-form-urlencoded` POST requests cross-origin **without** a CORS preflight. So a malicious page could submit form-encoded POST requests to the Livewire endpoint. CSRF is the primary defense, but requiring JSON content-type forces a CORS preflight for cross-origin requests — a free extra layer.
4. This PR adds a `RequireLivewireHeaders` middleware that aborts with 404 (not 403, to avoid confirming the endpoint exists) if either header is missing.
5. The middleware is registered on the default update route AND force-applied to any custom route defined via `setUpdateRoute()`, so users can't accidentally opt out.

1. **Guard inside `handleUpdate()` directly** — worse because it runs even when the test harness disables middleware, and it conflates request validation with business logic.
2. **Only check `X-Livewire` without JSON content-type** — misses the CORS preflight benefit, which is the main security win here.
3. **Do nothing, rely solely on CSRF** — valid, but this is cheap defence-in-depth that costs nothing for legitimate users (the JS client already sends both headers).

### Changes Made

### Test Results

All tests pass locally:
- `src/Mechanisms/HandleRequests/UnitTest.php` — 10 tests, 22 assertions ✅
- `src/Mechanisms/HandleRequests/PayloadGuardsUnitTest.php` — 11 tests, 22 assertions ✅
- `src/Features/SupportModels/UnitTest.php` (modified test) — passes ✅
- `src/Mechanisms/PersistentMiddleware/UnitTest.php` (modified test) — passes ✅
- CI: all 22 checks pass ✅
- Regression verified: all 3 rejection tests correctly fail (return 200 instead of 404) when middleware is removed.

### Code Review

**Middleware** (`RequireLivewireHeaders.php`): Clean, minimal, correct. Uses `$request->isJson()` which handles content-type variants like `application/json; charset=utf-8`. Returns 404 rather than 403 to avoid endpoint enumeration.

**Route registration** (`HandleRequests.php:27`): Middleware added to default route's middleware array — clean.

**Force-apply on custom routes** (`HandleRequests.php:94-95`): `setUpdateRoute()` auto-appends the middleware even if the user's callback doesn't include it. Smart design.

**Test updates**: Existing tests that POST directly to the endpoint now correctly use `->withHeaders(['X-Livewire' => 'true'])->postJson()` instead of `->post()`. The `handleUpdate()` unit test was also fixed — it was incorrectly creating a `new Request()` and passing it as an argument to a method that takes no parameters.

**New tests**: 4 well-structured tests covering all header combinations (missing X-Livewire, missing JSON, missing both, both present).

**No dist/ files in diff.** No style violations. No unrelated changes.

### Security

This PR *is* a security improvement. The middleware adds defence-in-depth against cross-origin form submissions and endpoint probing. No new security concerns introduced.

### Verdict

Clean, surgical, well-tested security hardening. The middleware approach is the right architecture choice. Tests properly verify both acceptance and rejection. All existing tests updated to include the required headers. No breaking changes for legitimate users since the JS client already sends both headers. Merge.

---
*Reviewed by Claude*

PR #9966, opened 2026-02-10
[4.x] Strengthen checksum failure rate limiting with lower per-IP threshold and global limiter

joshhanley:
# The Scenario

An attacker is attempting to brute-force or probe the Livewire update endpoint by sending requests with invalid checksums. With the previous rate limit of 10 failures per IP per 10 minutes, an attacker could make a significant number of probing requests before being blocked. Worse, by rotating across multiple IP addresses, they could bypass the per-IP limit entirely and sustain a high volume of invalid requests indefinitely.

# The Problem

The rate limiting in `Checksum.php` only tracked failures per individual IP address, with a threshold of 10 failures per 10 minutes. This had two weaknesses:

1. **The per-IP threshold was too generous** — 10 failures is more than a legitimate user would ever trigger (a real user might cause 1-2 checksum failures from a stale tab, never 10).

2. **No global limit existed** — an attacker using IP rotation (botnets, proxy pools, cloud functions) could distribute requests across many IPs, with each IP staying under the per-IP limit while the total attack volume remained high. There was no mechanism to detect or block this pattern.

Additionally, `request()->ip()` relies on correct trusted proxy configuration — applications behind reverse proxies or load balancers that haven't configured `TrustProxies` middleware would see all requests as coming from the same proxy IP, causing legitimate users to be incorrectly rate-limited.

# The Solution

- **Reduced per-IP threshold from 10 to 5** failures per 10 minutes — still generous enough for legitimate edge cases (stale tabs, key rotation) but significantly reduces the window for probing.

- **Added global rate limiting** — 50 failures per minute across all IPs. This catches distributed attacks where no single IP exceeds its individual limit. The global check runs first in `enforceRateLimit()` so it short-circuits before the per-IP check when the system is under distributed attack. Both `recordFailure()` and `enforceRateLimit()` now operate on both keys.

- **Documented trusted proxy requirements** — added an inline comment explaining that applications behind reverse proxies must configure `TrustProxies` middleware (or `TRUSTED_PROXIES` env var) for `request()->ip()` to return real client IPs, with a link to the Laravel documentation.

### DIFF
--- a/src/Mechanisms/HandleComponents/Checksum.php
+++ b/src/Mechanisms/HandleComponents/Checksum.php
@@ -8,8 +8,11 @@ use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
 use function Livewire\trigger;
 
 class Checksum {
-    protected static $maxFailures = 10;
+    protected static $maxFailures = 5;
     protected static $decaySeconds = 600; // 10 minutes
+
+    protected static $maxGlobalFailures = 50;
+    protected static $globalDecaySeconds = 60; // 1 minute
 
     static function verify($snapshot) {
         // Check if this IP is already blocked due to too many failures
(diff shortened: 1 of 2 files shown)


joshhanley:
Closing as this hands hackers a tool for DDOS

PR #9967, opened 2026-02-10
[4.x] Disable default update route when a custom route is registered

joshhanley:
# The Scenario

When a developer registers a custom update route with additional middleware (e.g. `auth`, tenant scoping), the default hashed route still accepts requests:

```php
// AppServiceProvider.php
    return Route::post('/tenant/{tenant}/livewire/update', $handle)
        ->middleware(['web', 'auth', 'tenant']);
});
```

An attacker can bypass the `auth` and `tenant` middleware by sending requests directly to the default `/livewire-{hash}/update` endpoint instead.

# The Problem

`HandleRequests::boot()` registers the default route before the user's service provider calls `setUpdateRoute()`. When the custom route is registered, both routes coexist in the router. While `findUpdateRoute()` already prioritises custom routes for URL generation, the default route remains active and continues processing requests — effectively making the custom route's middleware optional from an attacker's perspective.

# The Solution

At the top of `handleUpdate()`, the request's route name is checked against the actual route collection via `findUpdateRoute()`. If the request arrived via the default route (`default-livewire.update`) but a custom route exists, the request is rejected with a 404. This requires no additional state on the class and works correctly even with cached routes.

### DIFF
--- a/src/Mechanisms/HandleRequests/CustomUpdateRouteUnitTest.php
+++ b/src/Mechanisms/HandleRequests/CustomUpdateRouteUnitTest.php
@@ -18,6 +18,12 @@ class CustomUpdateRouteUnitTest extends TestCase
 
     public function test_custom_route_is_used_for_url_generation(): void
     {
+        // Both default and custom routes exist (with different names to avoid collision)
+        $livewireUpdateRoutes = collect(Route::getRoutes()->getRoutes())->filter(function ($route) {
+            return str($route->getName())->endsWith('livewire.update');
+        });
+
+        $this->assertCount(2, $livewireUpdateRoutes);
         $this->assertEquals('/custom/livewire/update', Livewire::getUpdateUri());
     }
 
--- a/src/Mechanisms/HandleRequests/CustomUpdateRouteUnitTest.php
+++ b/src/Mechanisms/HandleRequests/CustomUpdateRouteUnitTest.php
@@ -3,6 +3,7 @@
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\ServiceProvider;
 use Livewire\Livewire;
+use Livewire\Mechanisms\HandleRequests\EndpointResolver;
 use Tests\TestCase;
 
 class CustomUpdateRouteUnitTest extends TestCase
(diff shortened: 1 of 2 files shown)


calebporzio:
## PR Review: #9967 — [4.x] Disable default update route when a custom route is registered

**Type**: Security / Bug fix
**Verdict**: Merge

1. Livewire registers a default update route (`/livewire-{hash}/update`) during its service provider boot, with only `web` middleware.
2. A developer calls `Livewire::setUpdateRoute()` to register a custom route (e.g., `/tenant/{tenant}/livewire/update`) with additional middleware like `auth` and `tenant`.
3. Both routes now coexist in the router. `findUpdateRoute()` correctly prioritizes the custom route for URL generation (Livewire's JS will POST to the custom URL), but the default hashed route is still a live endpoint.
4. An attacker who discovers/guesses the hashed path can POST directly to it, completely bypassing the auth/tenant middleware the developer added to their custom route.

The fix: at the top of `handleUpdate()`, check if the request came in on the default route while a custom route exists. If so, 404.

1. **Don't register the default route when a custom one exists** — Can't. `boot()` runs before `setUpdateRoute()` is called by the app. The default route must be registered early to beat catch-all routes.
2. **Remove the default route from the router when `setUpdateRoute()` is called** — Laravel has no clean API for removing registered routes. Fragile.
3. **Handler-level check (this PR)** — Simplest correct solution. 4 lines of logic, works with cached routes, no route removal gymnastics. This is the right call.

### Changes Made

### Test Results

CI: All 22 checks pass (Unit, Browser, Legacy across PHP 8.4/8.5 and Laravel 11/12). One flaky Browser 2 failure on the first run that passed on retry — unrelated to this PR.

Could not run tests locally due to Chromedriver not being installed in this worktree, but the tests are straightforward HTTP assertions and CI confirms they pass.

### Code Review

- **`HandleRequests.php:125-131`**: The fix is surgical. Checks `request()->route()?->getName()` against `'default-livewire.update'` and compares with `findUpdateRoute()`. The `findUpdateRoute()` call is a linear route scan, but only fires on requests hitting the default endpoint — negligible overhead.
- **`CustomUpdateRouteUnitTest.php:30-45`**: Two clean tests. One proves the default route 404s when a custom route exists (the regression test), one proves the custom route still works (positive case). Both are minimal and clearly named.
- **The `findUpdateRoute()` method already existed and correctly prioritizes custom routes over the default** — the PR leverages this existing behavior cleanly.
- **Cached routes**: Works correctly. Route names survive caching, and `findUpdateRoute()` iterates the cached route collection.
- **Edge case — no custom route**: The condition is false (both sides equal `'default-livewire.update'`), so requests pass through normally. Correct.
- **Style**: Clean PHP, follows project conventions.

### Security

This PR *is* a security fix. The vulnerability is real: any app using `setUpdateRoute()` with auth middleware currently has a middleware bypass via the default hashed route. The fix correctly eliminates this bypass. The route name comparison is safe — set by the framework, not user input.

### Verdict

Merge. This is a clean, minimal, well-tested security fix from a trusted contributor (Josh Hanley). The implementation is the simplest correct approach. The tests are precise and CI is green. No changes needed.

---
*Reviewed by Claude*

PR #9968, opened 2026-02-10
[4.x] Fire ChecksumFailure event on checksum verification failure

joshhanley:
# The Scenario

When an attacker sends requests with tampered component payloads to the Livewire update endpoint, the checksum verification rejects them and rate-limits the IP. However, there's no way for applications to detect these attempts — the failures are silent beyond the rate limiter and the exception.

# The Problem

`Checksum::verify()` throws a `CorruptComponentPayloadException` and records a rate limit hit, but doesn't fire any Laravel event. Applications that want to monitor for potential attacks (e.g. to alert, log to a security dashboard, or trigger IP bans) have no hook to listen for.

# The Solution

A new `Livewire\Events\ChecksumFailure` event is now fired whenever a checksum mismatch is detected. The event includes:

- `ipAddress` — the request IP
- `userAgent` — the request user agent
- `componentName` — the component name from the snapshot (if available)

Applications can listen for this event using standard Laravel event listeners:

```php
Event::listen(ChecksumFailure::class, function (ChecksumFailure $event) {
    Log::warning('Livewire checksum failure', [
        'ip' => $event->ipAddress,
        'user_agent' => $event->userAgent,
        'component' => $event->componentName,
    ]);
});
```

### DIFF
--- a/src/Events/ChecksumFailure.php
+++ b/src/Events/ChecksumFailure.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Livewire\Events;
+
+class ChecksumFailure
+{
+    public function __construct(
+        public string $ipAddress,
+        public string $userAgent,
+        public ?string $componentName = null,
+    ) {}
+}
--- a/src/Mechanisms/HandleComponents/Checksum.php
+++ b/src/Mechanisms/HandleComponents/Checksum.php
@@ -3,6 +3,7 @@
 namespace Livewire\Mechanisms\HandleComponents;
 
 use Illuminate\Support\Facades\RateLimiter;
+use Livewire\Events\ChecksumFailure;
 use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
 
 use function Livewire\trigger;
@@ -25,6 +26,8 @@ class Checksum {
             trigger('checksum.fail', $checksum, $comparitor, $snapshot);
 
             static::recordFailure();
+
+            static::fireFailureEvent($snapshot);
 
             throw new CorruptComponentPayloadException;
         }
@@ -56,6 +59,17 @@ class Checksum {
     protected static function recordFailure()
     {
         RateLimiter::hit(static::rateLimitKey(), static::$decaySeconds);
+    }
+
+    protected static function fireFailureEvent($snapshot)
+    {
+        $request = request();
+
+        event(new ChecksumFailure(
+            ipAddress: $request->ip(),
+            userAgent: $request->userAgent() ?? '',
+            componentName: $snapshot['memo']['name'] ?? null,
+        ));
     }
 
     protected static function rateLimitKey(): string
--- a/src/Mechanisms/HandleComponents/ChecksumFailureEventUnitTest.php
+++ b/src/Mechanisms/HandleComponents/ChecksumFailureEventUnitTest.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace Livewire\Mechanisms\HandleComponents;
+
+use Illuminate\Support\Facades\Event;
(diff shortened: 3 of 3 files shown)


calebporzio:
## PR Review: #9968 — [4.x] Fire ChecksumFailure event on checksum verification failure

**Type**: Feature
**Verdict**: Close

1. When an attacker sends a tampered Livewire payload, `Checksum::verify()` detects the mismatch and throws `CorruptComponentPayloadException`.
2. This PR adds a new Laravel `event()` call (via a new `ChecksumFailure` event class) right before the exception is thrown.
3. The idea: let apps listen with `Event::listen(ChecksumFailure::class, ...)` to log/alert on these security events.

1. **Use the existing `trigger('checksum.fail')` hook (already works today).** Livewire already fires `trigger('checksum.fail', $checksum, $comparitor, $snapshot)` on line 26 of `Checksum.php` — it was always there. Any user can listen for it in their `AppServiceProvider::boot()`:
   ```php
   use function Livewire\on;

   on('checksum.fail', function ($checksum, $comparitor, $snapshot) {
       Log::warning('Checksum failure', [
           'ip' => request()->ip(),
           'component' => $snapshot['memo']['name'] ?? null,
       ]);
   });
   ```
   This gives you *more* data than the PR's event (the actual checksums + full snapshot) and requires zero framework changes. Internal code (`SupportChecksumErrorDebugging`) already uses this exact hook.

2. **Add a Laravel event on top of the existing trigger.** This is what the PR does. It's redundant — now there are two parallel event systems firing for the same thing. The Livewire trigger system is the idiomatic way to hook into Livewire internals.

3. **Document the existing hook.** If discoverability is the issue, a docs note about `checksum.fail` serves users better than a parallel event class.

### Changes Made

No changes made. This PR should not be merged.

### Test Results

CI: All checks passing. Tests are well-written but test a feature that shouldn't exist.

### Code Review

The code itself is clean and correct — no style issues, the event class is simple, tests are thorough. The problem isn't quality, it's justification.

Specific concerns:
- **Redundant API surface.** `trigger('checksum.fail')` already exists at `Checksum.php:26` and fires on every failure. Adding a parallel Laravel `event()` creates two ways to do the same thing.
- **New `src/Events/` directory.** This would be the first file in that namespace. Establishing a new namespace for something already covered by the existing event bus is unnecessary.
- **Less data than the existing hook.** The `checksum.fail` trigger passes `$checksum`, `$comparitor`, and the full `$snapshot`. The new event only passes IP, user agent, and component name — strictly less useful for debugging.
- **Test file in wrong location.** `ChecksumFailureEventUnitTest.php` lives in `src/Mechanisms/HandleComponents/` but doesn't follow the feature-folder convention. Minor point given the verdict.

### Security

No security concerns — the event doesn't expose anything that `request()` doesn't already have. However, the `checksum.fail` trigger exposes the full snapshot, which is also fine since it's server-side only.

### Verdict

**Close.** The use case is completely valid — monitoring checksum failures is a good idea. But the capability already exists via `trigger('checksum.fail')`. This PR adds a redundant Laravel event that duplicates the existing Livewire event bus hook, introduces a new `Events` namespace, and provides less data than what's already available. The laziest correct solution is to use what's already there. If discoverability is the concern, documentation is the answer, not new API surface.

---
*Reviewed by Claude*

calebporzio:
agreed @joshhanley ?

joshhanley:
@calebporzio the only reason I left this recommendation was to see if we are happy for end users to be using Livewire's `on('checksum.fail'...)` listeners? If so, then yep close it.

calebporzio:
mmm yeah it's a good question. I say fine on this one.

PR #9969, opened 2026-02-10
[4.x] Add type validation in `CollectionSynth` to prevent arbitrary class instantiation

joshhanley:
# The Scenario

When `APP_KEY` is compromised, an attacker can forge valid HMAC checksums for Livewire component snapshots. With a forged snapshot, they can manipulate the `CollectionSynth` metadata to specify an arbitrary class name, which gets instantiated via `new $meta['class']($value)` during hydration.

This was the primary gadget chain entry point described in [Synacktiv's disclosure of CVE-2025-54068](https://www.synacktiv.com/en/publications/livewire-remote-command-execution-through-unmarshaling) — classes like `GuzzleHttp\Psr7\FnStream` could be instantiated through `CollectionSynth` to achieve remote code execution.

# The Problem

`CollectionSynth::hydrate()` performs `new $meta['class']($value)` where `$meta['class']` comes from the signed snapshot. The only guards are the HMAC checksum (requires `APP_KEY` to forge) and the `SecurityPolicy` denylist. If the checksum is valid and the class isn't on the denylist, any class whose constructor accepts an array can be instantiated.

`FormObjectSynth` already validates that the class extends `Form` before instantiation:

```php
if (\! isset($meta['class']) || \! is_a($meta['class'], Form::class, true)) {
    throw new \Exception('Livewire: Invalid form object class.');
}
```

`CollectionSynth` had no equivalent check.

# The Solution

Added a type validation in `CollectionSynth::hydrate()` that ensures `$meta['class']` is `Illuminate\Support\Collection` or a subclass of it before instantiation — mirroring the same pattern `FormObjectSynth` already uses for `Form`:

```php
if (\! is_a($meta['class'], Collection::class, true)) {
    throw new \Exception("Livewire: Class [{$meta['class']}] is not a valid Collection type.");
}
```

This closes the primary gadget chain entry point even if the `APP_KEY` is compromised, as only legitimate Collection types can be instantiated through this synthesiser. The `is_a()` check with `true` as the third argument allows string class names to be validated without instantiation, and supports custom Collection subclasses that applications may use.

### DIFF
--- a/src/Mechanisms/HandleComponents/Synthesizers/CollectionSynth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/CollectionSynth.php
@@ -25,7 +25,7 @@ class CollectionSynth extends ArraySynth {
     }
 
     function hydrate($value, $meta, $hydrateChild) {
-        if (! is_a($meta['class'], Collection::class, true)) {
+        if (! isset($meta['class']) || ! is_a($meta['class'], Collection::class, true)) {
             throw new \Exception("Livewire: Class [{$meta['class']}] is not a valid Collection type.");
         }
 
--- a/src/Mechanisms/HandleComponents/Synthesizers/CollectionSynth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/CollectionSynth.php
@@ -25,6 +25,10 @@ class CollectionSynth extends ArraySynth {
     }
 
     function hydrate($value, $meta, $hydrateChild) {
+        if (! is_a($meta['class'], Collection::class, true)) {
+            throw new \Exception("Livewire: Class [{$meta['class']}] is not a valid Collection type.");
+        }
+
         foreach ($value as $key => $child) {
             $value[$key] = $hydrateChild($key, $child);
         }
--- a/src/Mechanisms/HandleComponents/Synthesizers/UnitTest.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/UnitTest.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace Livewire\Mechanisms\HandleComponents\Synthesizers;
+
+use Illuminate\Support\Collection;
+use Livewire\Component;
+use Livewire\Livewire;
+use Tests\TestComponent;
+
+class UnitTest extends \Tests\TestCase
+{
+    public function test_collection_synth_hydrates_a_collection()
+    {
+        Livewire::test(ComponentWithCollectionProperty::class)
+            ->assertSet('items', collect(['foo', 'bar']))
+            ->call('$refresh')
+            ->assertSet('items', collect(['foo', 'bar']));
+    }
+
+    public function test_collection_synth_hydrates_a_custom_collection_subclass()
+    {
+        Livewire::test(ComponentWithCustomCollectionProperty::class)
+            ->call('$refresh')
+            ->assertSet('items', new CustomCollection(['foo', 'bar']));
+    }
+
+    public function test_collection_synth_rejects_non_collection_class()
+    {
+        $synth = new CollectionSynth(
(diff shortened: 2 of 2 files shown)


calebporzio:
## PR Review: #9969 — [4.x] Add type validation in `CollectionSynth` to prevent arbitrary class instantiation

**Verdict**: Merge

1. When Livewire dehydrates a Collection property, it stores `['class' => 'Illuminate\Support\Collection']` in the snapshot metadata.
2. On hydrate, `CollectionSynth::hydrate()` does `new $meta['class']($value)` — trusting that `$meta['class']` is a Collection.
3. The snapshot is HMAC-signed, so normally this can't be tampered with. But if `APP_KEY` is compromised (as in CVE-2025-54068), an attacker can forge valid checksums and set `$meta['class']` to any class whose constructor accepts an array — like `GuzzleHttp\Psr7\FnStream` — achieving RCE.
4. This PR adds an `is_a()` check before instantiation, ensuring only `Collection` or its subclasses can be hydrated through this synth. This mirrors the existing pattern in `FormObjectSynth` and `WireableSynth`.

This is defense-in-depth. The HMAC is the primary guard, but validating the class type costs nothing and closes the gadget chain even when the key is compromised.

1. **Denylist approach** — Add dangerous classes to `SecurityPolicy`. Already exists but is whack-a-mole; new gadget classes can always be discovered. The allowlist approach here is strictly better since only legitimate Collection types can pass.
2. **Do nothing (rely on HMAC)** — The HMAC *should* prevent tampering. But defense-in-depth is the right call for a security-critical code path. The check is one line and zero-cost.
3. **Validate at the synth dispatcher level** — Could add a generic "each synth must validate its own class" contract. Overkill for now, but worth considering if more synths need the same treatment.

The PR's approach is the right one: surgical, minimal, follows existing precedent.

### Changes Made

- Added `isset($meta['class'])` guard before the `is_a()` check, matching the pattern used in `FormObjectSynth` (line 33) and `WireableSynth` (line 38). Without it, a forged snapshot with a missing `class` key would trigger a PHP warning before hitting the exception.

### Test Results

- All 5 unit tests pass with the fix in place.
- Verified regression: removed the `is_a()` guard and confirmed both rejection tests (`test_collection_synth_rejects_non_collection_class` and `test_collection_synth_rejects_arbitrary_class_with_array_constructor`) fail without the fix. Tests properly isolate the root cause.
- CI: All 22 checks pass (Unit, Browser, Legacy across PHP 8.4/8.5 and Laravel 11/12).

### Code Review

**The fix** (`CollectionSynth.php:28-30`): Clean, minimal, correct. Uses `is_a()` with `true` third arg for string class validation without instantiation. Supports custom Collection subclasses.

**The tests**: Well-structured. 5 tests covering: happy path (Collection round-trip), custom subclass round-trip, stdClass rejection, ArrayObject rejection (the actual attack vector), and direct subclass allowance. Test names are clear and descriptive. Test components are appropriately named.

**Note for follow-up**: `EnumSynth::hydrate()` does `$class::from($value)` with unvalidated `$meta['class']`. Lower risk (static method call, not constructor), but should get the same treatment: `if (! is_subclass_of($meta['class'], 'BackedEnum'))`. This is out of scope for this PR but worth a separate fix.

### Security

This PR **improves** security. The `is_a()` allowlist closes the primary gadget chain entry point described in CVE-2025-54068, even when APP_KEY is compromised. No new security concerns introduced.

### Verdict

**Merge.** This is a clean, minimal, well-tested security hardening fix that follows established patterns (`FormObjectSynth`, `WireableSynth`). The contributor correctly identified the vulnerability, wrote tests that actually verify the fix (confirmed by regression testing), and the implementation is the simplest correct solution. The only change I made was adding `isset($meta['class'])` for consistency with the other synths.

---
*Reviewed by Claude*

PR #9970, opened 2026-02-10
[4.x] Add payload schema validation and tiered response handling

joshhanley:
# The Scenario

When Livewire's update endpoint receives malformed or malicious requests, the responses leak information about the endpoint's existence and internal structure:

- A request with no `components` key returns `200` with empty JSON — confirming the endpoint exists to scanners
- A request with missing `snapshot`/`updates`/`calls` fields triggers a PHP warning and returns `500` with a stack trace (in debug mode)
- A request with invalid JSON in `snapshot` returns `500` with a type error
- A request with a bad checksum returns `500` with a `CorruptComponentPayloadException` message
- A request with type-mismatched update values (e.g., a string where an array property is expected) passes all structural validation but crashes during hydration with an unhandled `TypeError` → `500`
- A request that attempts to update a `#[Locked]` property returns `500` with a `CannotUpdateLockedPropertyException` message

The variety of distinct error responses also helps attackers fingerprint the endpoint and understand its internal structure.

# The Problem

`handleUpdate()` performs minimal validation of the incoming request structure. It defaults `components` to an empty array and iterates directly over whatever is provided, trusting that each element contains `snapshot`, `updates`, and `calls` with the correct types. There is no structural validation before data reaches the checksum verification or component hydration logic.

Additionally, `CorruptComponentPayloadException` (thrown on checksum failure) and `CannotUpdateLockedPropertyException` (thrown when a `#[Locked]` property is modified) both render as generic `500` errors. The JS client already has code to handle `419` responses with a "page expired, refresh?" prompt — which is exactly the right UX for stale snapshots — but it never receives that status code.

Even with structural validation in place, type-mismatched update values (e.g., sending `"not_an_array"` for an `array` property) pass validation because the request envelope is well-formed — `updates` is a valid array of key-value pairs. The type mismatch only surfaces when PHP attempts to hydrate the property, throwing a `TypeError` that results in a `500`.

# The Solution

Two layers of schema validation were added to `handleUpdate()`:

**Request-level** (before the processing loop):
- `components` must be a non-empty array
- Each element must be an array containing `snapshot` (string), `updates` (array), `calls` (array)

**Component-level** (after `json_decode`, before checksum verification):
- Decoded snapshot must contain `data` (array), `memo` (array), `checksum` (string)
- `memo` must contain `id` (string) and `name` (string)
- Each call entry must have `method` (string) and `params` (array)

The JS client always sends well-formed payloads — there are no code paths where any of these could be missing or have unexpected types — so these checks are safe to enforce.

**Type mismatch handling:**

The `app('livewire')->update()` call is wrapped in a try/catch for `\TypeError` — the exception PHP 8+ throws for type mismatches like assigning a string to an `array`-typed property. In production, this returns `419`. In debug mode, the exception propagates so developers see the real error. Only `\TypeError` is caught — `\ErrorException` (Laravel's wrapper for PHP warnings/notices) is intentionally excluded to avoid masking unrelated warnings as 419s.

**Tiered response handling:**

| Scenario | Response | Rationale |
|---|---|---|
| Missing/malformed structure | **404** | Request didn't come from the JS client. Avoids confirming endpoint exists. |
| Valid structure, bad checksum | **419** | JS handles 419 with "page expired, refresh?" prompt — correct UX for stale pages. |
| Valid structure, type-mismatched values | **419** | Same recovery action as checksum failure (see below). |
| Valid structure, locked property update | **419** | Same recovery action — only legitimate cause is a stale page after deployment. |
| Rate limited | **429** (unchanged) | Correct HTTP status. |
| Valid request | **200** (unchanged) | Normal operation. |

**Why 419 for type mismatches and locked property updates (not 400 or 422)?**

While `400` or `422` would be more semantically precise from an HTTP perspective, the user experience is the deciding factor. The only legitimate scenario where a real user triggers a type mismatch or locked property violation is after a deployment changes a property's type or adds `#[Locked]` — the user has a stale page with the old schema. The correct recovery action is the same as a checksum failure: refresh the page. Since the JS client already handles `419` with a "page expired, refresh?" prompt, using a different status code would require adding a second client-side handler that does exactly the same thing. Using `419` consistently for all "structurally valid but unprocessable" payloads also avoids giving attackers a way to distinguish how their payload failed.

The 419 for checksum failures and locked property updates is implemented via a `render()` method on `CorruptComponentPayloadException` and `CannotUpdateLockedPropertyException` respectively. The type mismatch 419 is handled by a try/catch in `handleUpdate()`. All three gate on `app.debug` so developers see full errors during development. All response bodies are generic to avoid leaking schema information.

Fixes #9997

### DIFF
--- a/src/Features/SupportLockedProperties/CannotUpdateLockedPropertyException.php
+++ b/src/Features/SupportLockedProperties/CannotUpdateLockedPropertyException.php
@@ -10,4 +10,13 @@ class CannotUpdateLockedPropertyException extends \Exception
             'Cannot update locked property: ['.$this->property.']'
         );
     }
+
+    // In debug mode, let Laravel render the full error page.
+    // In production, return a generic 419 to avoid leaking details.
+    public function render($request)
+    {
+        if (config('app.debug')) return false;
+
+        return response('', 419);
+    }
 }
--- a/src/Features/SupportLockedProperties/UnitTest.php
+++ b/src/Features/SupportLockedProperties/UnitTest.php
@@ -5,6 +5,7 @@ namespace Livewire\Features\SupportLockedProperties;
 use Livewire\Livewire;
 use Livewire\Component as BaseComponent;
 use Livewire\Form;
+use Livewire\Mechanisms\HandleRequests\EndpointResolver;
 use Tests\TestComponent;
 
 class UnitTest extends \Tests\TestCase
@@ -60,6 +61,25 @@ class UnitTest extends \Tests\TestCase
             ->set('form.foo', 'bar')
             ->assertSetStrict('form.foo', 'bar')
             ->assertOk();
+    }
+
+    function test_updating_locked_property_returns_419_in_production()
+    {
+        config()->set('app.debug', false);
+
+        $testable = Livewire::test(new class extends TestComponent {
+            #[BaseLocked]
+            public $count = 1;
+        });
+
+        $snapshotJson = json_encode($testable->snapshot);
+
+        $response = $this->withHeaders(['X-Livewire' => 'true'])
+            ->postJson(EndpointResolver::updatePath(), ['components' => [
+                ['snapshot' => $snapshotJson, 'updates' => ['count' => 2], 'calls' => []],
+            ]]);
+
+        $response->assertStatus(419);
     }
 }
 
--- a/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
+++ b/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
@@ -16,14 +16,6 @@ class CorruptComponentPayloadException extends \Exception
         );
     }
 
-    /**
-     * Render the exception as an HTTP response.
-     *
-     * Returns 419 ("Page Expired") — the JS client already handles this status
-     * with a "page expired, refresh?" prompt, which is the correct UX for stale
-     * snapshots (expired session, rotated key, stale tab). A generic body avoids
-     * leaking any internal details to attackers.
-     */
     public function render($request)
     {
         if (config('app.debug')) return false;
--- a/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
+++ b/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
@@ -16,6 +16,8 @@ class CorruptComponentPayloadException extends \Exception
         );
     }
 
+    // In debug mode, let Laravel render the full error page.
+    // In production, return a generic 419 to avoid leaking details.
     public function render($request)
     {
         if (config('app.debug')) return false;
--- a/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
+++ b/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
@@ -26,6 +26,8 @@ class CorruptComponentPayloadException extends \Exception
      */
     public function render($request)
     {
+        if (config('app.debug')) return false;
+
         return response('', 419);
     }
 }
--- a/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
+++ b/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
@@ -15,4 +15,17 @@ class CorruptComponentPayloadException extends \Exception
             "Ensure that the [name, id, data] of the Livewire component wasn't tampered with between requests."
         );
     }
+
+    /**
+     * Render the exception as an HTTP response.
+     *
+     * Returns 419 ("Page Expired") — the JS client already handles this status
+     * with a "page expired, refresh?" prompt, which is the correct UX for stale
+     * snapshots (expired session, rotated key, stale tab). A generic body avoids
+     * leaking any internal details to attackers.
+     */
+    public function render($request)
+    {
+        return response('', 419);
+    }
 }
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -199,8 +199,27 @@ class HandleComponents extends Mechanism
 
     public function update($snapshot, $updates, $calls)
     {
-        if (! is_array($snapshot) || ! isset($snapshot['data'], $snapshot['memo'])) {
-            throw new \InvalidArgumentException('Invalid Livewire snapshot');
+        if (! is_array($snapshot)
+            || ! is_array($snapshot['data'] ?? null)
+            || ! is_array($snapshot['memo'] ?? null)
+            || ! is_string($snapshot['checksum'] ?? null)
+            || ! is_string($snapshot['memo']['id'] ?? null)
+            || ! is_string($snapshot['memo']['name'] ?? null)
+        ) {
+            if (config('app.debug')) throw new \InvalidArgumentException('Invalid Livewire snapshot structure: expected [data], [memo], [checksum], [memo.id], and [memo.name].');
+
+            abort(404);
+        }
+
+        foreach ($calls as $call) {
+            if (! is_array($call)
+                || ! is_string($call['method'] ?? null)
+                || ! is_array($call['params'] ?? null)
+            ) {
+                if (config('app.debug')) throw new \InvalidArgumentException('Invalid Livewire call structure: each call must contain [method] (string) and [params] (array).');
+
+                abort(404);
+            }
         }
 
         $data = $snapshot['data'];
--- a/src/Mechanisms/HandleRequests/CustomUpdateRouteUnitTest.php
+++ b/src/Mechanisms/HandleRequests/CustomUpdateRouteUnitTest.php
@@ -5,6 +5,7 @@ use Illuminate\Support\ServiceProvider;
 use Livewire\Livewire;
 use Livewire\Mechanisms\HandleRequests\EndpointResolver;
 use Tests\TestCase;
+use Tests\TestComponent;
 
 class CustomUpdateRouteUnitTest extends TestCase
 {
@@ -37,8 +38,13 @@ class CustomUpdateRouteUnitTest extends TestCase
 
     public function test_custom_route_accepts_requests_when_registered(): void
     {
+        $testable = Livewire::test(new class extends TestComponent {});
+        $snapshotJson = json_encode($testable->snapshot);
+
         $response = $this->withHeaders(['X-Livewire' => 'true'])
-            ->post('/custom/livewire/update', ['components' => []]);
+            ->postJson('/custom/livewire/update', ['components' => [
+                ['snapshot' => $snapshotJson, 'updates' => [], 'calls' => []],
+            ]]);
 
         $response->assertOk();
         $this->assertArrayHasKey('components', $response->json());
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -133,17 +133,13 @@ class HandleRequests extends Mechanism
             }
         }
 
-        $components = request('components');
+        $requestPayload = request('components');
 
-        // Request-level schema validation: `components` must be a non-empty array,
-        // and each element must contain `snapshot` (string), `updates` (array), `calls` (array).
-        // The legitimate JS client always sends a well-formed payload, so any request
-        // failing these checks did not originate from the Livewire frontend.
-        if (! is_array($components) || empty($components)) {
+        if (! is_array($requestPayload) || empty($requestPayload)) {
             abort(404);
         }
 
-        foreach ($components as $component) {
+        foreach ($requestPayload as $component) {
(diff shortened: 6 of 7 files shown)


calebporzio:
## PR Review: #9970 — [4.x] Add payload schema validation and tiered response handling

**Verdict**: Merge

Currently, when someone sends a malformed or malicious request to Livewire's update endpoint:

1. A request with no `components` key returns `200` with empty JSON — confirming the endpoint exists to scanners
2. A request with missing `snapshot`/`updates`/`calls` fields crashes with a PHP warning → `500` with a stack trace
3. A request with invalid JSON in `snapshot` returns `500` with a type error
4. A request with a bad checksum returns `500` with a `CorruptComponentPayloadException` message
5. A request with type-mismatched values (e.g., string for an `array` property) crashes during hydration with an unhandled `TypeError` → `500`

Each of these distinct error responses helps attackers fingerprint the endpoint and map its internal structure.

This PR adds two layers of defense:

**Layer 1 — Structural validation**: Before processing begins, both the request envelope (`HandleRequests`) and the decoded snapshot (`HandleComponents`) are validated for correct types and required fields. Malformed requests get a `404` — as if the endpoint doesn't exist.

**Layer 2 — Tiered error responses**: Structurally valid requests that fail checksum verification or type hydration get `419` — which the JS client already handles with a "page expired, refresh?" prompt. This is correct UX because the only legitimate scenario for these failures is a stale page after deployment.

1. **Middleware-based validation**: Could validate the request shape in middleware before it reaches the handler. Adds complexity for no real benefit — the handler is the right place since it owns the schema contract.
2. **Generic 403/401 for all failures**: Simpler tiering but less useful. `404` for nonsense requests (hide the endpoint) vs `419` for stale-page failures (prompt refresh) is the right split.
3. **Do nothing / rely on error handler**: The status quo. Each failure mode produces a different error signature, which is the root problem.

The PR's approach is clean and correct.

### Changes Made

**Fixed `test_bad_checksum_returns_419`**: The original test was missing the `release` key in the hand-crafted snapshot memo. This caused `SupportReleaseTokens` to intercept the request with a `LivewireReleaseTokenMismatchException` (which already returns 419) *before* the checksum was ever verified. The test passed for the wrong reason entirely — it never exercised the `CorruptComponentPayloadException::render()` method it was supposed to test.

Added `'release' => $testable->snapshot['memo']['release']` to the snapshot so the release token check passes, the request actually reaches `Checksum::verify()`, and the test genuinely validates the new `render()` method.

Verified via stash-and-run:
- With fix reverted + release token in test → test correctly **fails** (gets 500, not 419)
- With fix applied + release token in test → test correctly **passes** (gets 419)

### Test Results

All 27 tests in `src/Mechanisms/HandleRequests/UnitTest.php` pass. CI is all green (22/22 checks passing).

Stash-and-run verification for each test category:

| Test | Without fix | With fix | Tests fix? |
|------|-------------|----------|------------|
| `malformed_request_payload_returns_404` (×9) | 500 | 404 | ✓ |
| `malformed_snapshot_returns_404` (×6) | 500 | 404 | ✓ |
| `malformed_calls_returns_404` (×4) | 500 | 404 | ✓ |
| `bad_checksum_returns_419` | 500 → 419 (after my fix) | 419 | ✓ (fixed) |
| `type_mismatched_update_value_returns_419` | 500 | 419 | ✓ |

### Code Review

- **`HandleRequests.php:136-150`**: Request envelope validation is clean. Uses `?? null` pattern consistently for null-safe type checks. Correct placement before the max_components limit check.
- **`HandleRequests.php:170-176`**: `TypeError` catch is appropriately narrow — only catches `TypeError`, not broader `Exception`/`Throwable`. Rethrows in debug mode. Good.
- **`HandleComponents.php:202-223`**: Snapshot and calls validation strengthens the existing weak check (`! isset($snapshot['data'], $snapshot['memo'])` → full type validation). The `abort(404)` in a mechanism-level class is a minor layering concern, but pragmatically fine since this method is only called from `HandleRequests`.
- **`CorruptComponentPayloadException.php:21-26`**: Using Laravel's renderable exception pattern is idiomatic. `return false` in debug mode to let Laravel's error page render is correct.
- **PHP style**: Follows project conventions. No issues.
- **Test quality**: Data providers are well-organized with descriptive case names. Good coverage of edge cases. The `test_valid_request_returns_200` sanity check is a nice touch.

### Security

This is a security hardening PR and it improves the security posture:
- Eliminates endpoint fingerprinting via distinct error signatures
- `404` for malformed requests avoids confirming endpoint existence
- No information leakage in production (empty response bodies)
- Debug mode still shows full errors for developer experience
- The `TypeError` catch is narrowly scoped — won't mask unrelated exceptions

No security concerns identified with the implementation.

### Verdict

Merge. This is a solid security hardening PR that eliminates endpoint fingerprinting and provides correct HTTP semantics for different failure modes. The code is surgical, well-tested, and follows existing patterns. The one test that was testing the wrong thing has been fixed and pushed.

---
*Reviewed by Claude*

joshhanley:
@calebporzio Added the same `render()` treatment to `CannotUpdateLockedPropertyException` — it was also returning 500 in production when someone attempted to update a `#[Locked]` property. Now returns 419 to match the rest. Also links the PR to discussion #9997 which reported this.

calebporzio:
## PR Review: #9970 — Add payload schema validation and tiered response handling

**Verdict**: Merge

### Summary

Livewire's update endpoint returns distinguishable error responses for different malformed payloads, letting attackers fingerprint the endpoint and infer internal structure. This PR adds two-layer schema validation (request-level and component-level) and unifies error responses: 404 for structurally invalid requests, 419 for valid-structure-but-unprocessable requests (bad checksum, type mismatch, locked property).

### What's happening

- **Before**: Sending different malformed payloads to the update endpoint produces different responses — empty 200, PHP warning 500, TypeError 500, checksum exception 500, locked property exception 500 — giving attackers a fingerprinting map of the endpoint's internals.
- **Before**: Users with stale pages (after deployment changes property types or adds `#[Locked]`) see a generic 500 error instead of the "page expired, refresh?" prompt the JS client already supports.
- **Root cause**: `handleUpdate()` assumed well-formed payloads from the JS client and had no structural validation before data reached the checksum or hydration layers.
- **Fix**: Two-layer validation rejects malformed requests early with 404 (hiding the endpoint from scanners). Checksum failures, type mismatches, and locked property violations now return 419 (triggering the existing client-side refresh prompt). All response bodies are empty to prevent information leakage.

1. **Just add `render()` to the two exceptions** (~8 lines) — fixes 500→419 for checksums/locked properties but leaves fingerprinting vectors open for structurally malformed requests.
2. **Wrap entire `handleUpdate()` in generic try/catch** — less code but dangerous: masks real bugs (DB failures, auth errors) as 404s.
3. **Do nothing** — legitimate users are unaffected, but the fingerprinting stays open and stale pages keep showing 500s instead of refresh prompts.

None — PR is clean as submitted. No bot-fixable issues identified. One design tradeoff (TypeError catch scope wrapping the entire update lifecycle) is acknowledged but acceptable — see verdict below.

### Test results

- **39 tests, 52 assertions** — all passing
- **Regression verified**: Stashed fix files, confirmed tests fail without the fix (malformed requests return 200/500 instead of 404, checksums return 500 instead of 419, type mismatches return 500 instead of 419, locked property updates return 500 instead of 419)
- **CI**: All 22 checks passing (Unit, Browser, Legacy across L11/L12 × PHP 8.4/8.5)

**Why now**: The endpoint fingerprinting is exploitable today. Issue #9997 confirms users are encountering this. No upstream fix is incoming — Laravel doesn't validate JSON payload structure at the framework level.

**Why here**: This is protocol-level validation specific to Livewire's wire format (snapshot/updates/calls structure, memo fields, checksum). Laravel can't know what shape a Livewire payload should be. The `render()` method on exceptions is standard Laravel convention.

**Why us**: Users can't wrap Livewire's internal update endpoint with their own validation — there's no hook for it. The framework registers the route, processes requests internally, and returns responses. This surface belongs to the framework.

<details>

**Framing**: Accurate. The PR is security hardening, not a bug fix per se — the endpoint works correctly for legitimate requests. The variety of distinct error responses for illegitimate requests is the real problem.

**Should this exist?**: Yes. ~50 lines of validation + ~10 lines of exception rendering. Maintenance burden proportional to value. The "no for now" bias doesn't apply — this is defense-in-depth that's harder to justify waiting on.

**Top concern — TypeError catch scope**: The catch at `HandleRequests.php:189` wraps the entire `app('livewire')->update()` call, including method calls, rendering, and dehydration. A `TypeError` from user component code (typed method parameters, synthesizer failures) gets silently returned as 419 in production. Practical risk is low but non-zero. See verdict for recommendation.

**404 for validation failures**: Semantically unusual (400 would be more correct) but defensible for scanner/bot resistance. Legitimate debugging in production is harder, but that's an acceptable tradeoff.
</details>

<details>

- **Phoenix LiveView**: WebSocket transport provides structural validation at the protocol layer — Livewire must build what Phoenix gets for free.
- **Inertia.js**: Uses 409 for version mismatches triggering full page reload — same pattern as Livewire's 419.
- **Turbo/HTMX**: HTML-over-the-wire avoids structured payload validation entirely — not applicable.
- **Universal pattern**: "Specific status code → client-side recovery action" is used by every framework surveyed. No framework returns 500 for stale-page scenarios.
- **Upstream fix viability**: None. This is Livewire-specific protocol validation.
</details>

<details>

**Blast radius**: Medium — `handleUpdate()` and `update()` are hot paths for all component updates, but changes are defensive (early returns/aborts before existing logic).

**High-risk areas checked**:
- File upload flow: `_finishUpload()` calls `update()` internally — safe, TypeError catch is appropriately narrow (only TypeError, not ErrorException)
- Component hydration: New validation runs before `fromSnapshot()` — JS client always sends well-formed payloads (verified in `js/request/index.js:264-275`)
- Exception render chain: `render()` method is standard Laravel pattern, matches existing `LivewireReleaseTokenMismatchException`
- 419 client handling: JS already handles 419 with "Page Expired" prompt (`js/request/index.js:374-379`)

**Backwards compatibility**: Fully compatible. Old clients send well-formed requests (pass validation). Server responses unchanged for valid requests.
</details>

<details>

No security concerns identified. This is a net security improvement:

- Eliminates endpoint fingerprinting (uniform 404 for structural failures, uniform 419 for semantic failures)
- Pre-checksum structural validation prevents invalid data from reaching hydration
- `json_decode` null safety handled correctly
- TypeError catch is appropriately narrow (only `\TypeError`, not `\Exception`)
- Empty response bodies prevent schema leakage
- `trigger('request')` hook bypass is theoretical only — hooks are server-side PHP code
</details>

<details>

**No new public API surface** — pure internal hardening.

**Pattern consistency**: `render()` methods match `CorruptComponentPayloadException` line-for-line. Debug-gating with `config('app.debug')` matches 34+ existing usages. `abort(404)` matches `RequireLivewireHeaders`. `try/catch (\TypeError)` matches existing pattern in `HandleComponents.php:603-614`.

**No registration changes needed** — no new classes, directives, attributes, or docs.

**Minor inconsistency**: `CannotUpdateLockedPropertyException` lacks the `BypassViewHandler` trait that 14 other Livewire exceptions use. Functionally irrelevant — exception is never thrown during Blade rendering.
</details>

<details>

**19 claims verified, 0 disproved.**

Key confirmations:
- TypeError catch scope concern is real — `callMethods()` and `render()` run inside the caught scope
- No debug mode tests exist — all tests set `config('app.debug', false)`
- `BypassViewHandler` inconsistency confirmed but irrelevant
- `trigger('request')` hook bypass is theoretical, not practical (no core code modifies payload after validation)
- All 39 tests pass locally on PHP 8.4.6
</details>

<details>

1. **Does the TypeError catch actually wrap method calls and rendering?** → Yes. `app('livewire')->update()` calls `callMethods()` at HandleComponents.php:250 and `render()` via lifecycle hooks, all inside the caught scope.

2. **Can `trigger('request')` modify the payload after validation?** → Theoretically yes (EventBus supports replacement), but no core code does this. Defense-in-depth validation in `update()` provides a second check.

3. **Is the `BypassViewHandler` trait needed on `CannotUpdateLockedPropertyException`?** → No. The trait prevents Blade's `ExtendedCompilerEngine` from wrapping exceptions during view compilation. `CannotUpdateLockedPropertyException` is thrown during the update lifecycle, never during Blade rendering.

4. **Does the JS client always send well-formed payloads?** → Yes. Verified in `js/request/index.js:264-275` — the commit function always constructs `{snapshot: string, updates: object, calls: array}`.
</details>

### Verdict

Merge. The security hardening is well-justified, well-implemented, and well-tested. The one substantive concern — the TypeError catch wrapping the entire update lifecycle — is a real but low-risk tradeoff. In practice, user TypeErrors during method calls or rendering are rare (PHP's type juggling handles most coercions), and when they do occur, a 419 with a refresh prompt is arguably better UX than a 500 with a stack trace. If this proves problematic in the wild, the catch can be narrowed in a follow-up without any API changes. Ship it.

---

PR #9971, opened 2026-02-10
[4.x] Enforce web middleware on custom update routes

joshhanley:
# The Scenario

When customising the Livewire update endpoint, a developer might forget to include the `web` middleware group:

```php
    return Route::post('/custom/livewire/update', $handle)
        ->middleware('auth');
    // Forgot ->middleware('web')!
});
```

Without `web` middleware, Laravel's CSRF protection (`VerifyCsrfToken`) is not applied, leaving the update endpoint vulnerable to cross-site request forgery attacks.

# The Problem

`setUpdateRoute()` accepts a callback that returns a route, but performs no validation on the returned route's middleware. The default Livewire route always includes `web` middleware, but custom routes rely entirely on the developer to include it. A single omission silently removes CSRF protection from the most security-sensitive endpoint in the application.

# The Solution

`setUpdateRoute()` now automatically ensures the `web` middleware group is present on custom routes. After the user's callback returns the route, Livewire checks the route's middleware list and adds `web` if it's missing. This is transparent — routes that already include `web` (either directly or via a route group) are unaffected.

The check uses `$route->middleware()` rather than `$route->gatherMiddleware()` to avoid polluting Laravel's computed middleware cache before the route is fully configured.

### DIFF
--- a/src/Mechanisms/HandleRequests/EnforceWebMiddlewareUnitTest.php
+++ b/src/Mechanisms/HandleRequests/EnforceWebMiddlewareUnitTest.php
@@ -1,0 +1,72 @@
+<?php
+
+use Illuminate\Support\Facades\Route;
+use Livewire\Livewire;
+use Tests\TestCase;
+
+class EnforceWebMiddlewareUnitTest extends TestCase
+{
+    public function test_web_middleware_is_automatically_added_to_custom_route_when_missing(): void
+    {
+        Livewire::setUpdateRoute(function ($handle) {
+            return Route::post('/custom/livewire/update', $handle);
+        });
+
+        $route = $this->findCustomUpdateRoute();
+
+        $this->assertNotNull($route, 'Custom livewire.update route should exist');
+        $this->assertContains('web', $route->middleware());
+    }
+
+    public function test_web_middleware_is_not_duplicated_when_already_present(): void
+    {
+        Livewire::setUpdateRoute(function ($handle) {
+            return Route::post('/custom/livewire/update', $handle)->middleware('web');
+        });
+
+        $route = $this->findCustomUpdateRoute();
+
+        $middlewareCount = count(array_filter($route->middleware(), fn ($m) => $m === 'web'));
+
+        $this->assertEquals(1, $middlewareCount);
+    }
+
+    public function test_web_middleware_is_detected_from_route_group(): void
+    {
+        Route::middleware('web')->group(function () {
+            Livewire::setUpdateRoute(function ($handle) {
+                return Route::post('/custom/livewire/update', $handle);
+            });
+        });
+
+        $route = $this->findCustomUpdateRoute();
+
+        $middlewareCount = count(array_filter($route->middleware(), fn ($m) => $m === 'web'));
+
+        $this->assertEquals(1, $middlewareCount);
+    }
+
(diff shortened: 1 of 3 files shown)


calebporzio:
## PR Review: #9971 — [4.x] Enforce web middleware on custom update routes

**Type**: Security hardening / Bug fix
**Verdict**: Merge

1. Livewire registers a default update route at `/livewire/update` with the `web` middleware group. This middleware group includes CSRF protection (`VerifyCsrfToken`), sessions, cookies, etc.
2. Developers can override this route using `Livewire::setUpdateRoute()`. The callback returns a Laravel Route object.
3. If the developer forgets to include `->middleware('web')` on their custom route, the update endpoint loses CSRF protection entirely. This is the most security-sensitive endpoint in the app — it executes public methods and updates component properties.
4. This PR adds a 4-line check in `setUpdateRoute()`: after the callback returns the route, it checks if `web` is in the route's middleware list and adds it if missing. Routes that already have `web` (directly or via a route group) are unaffected.

1. **Throw an exception** if `web` is missing — more explicit, but would break existing apps at runtime that happen to work (e.g., stateless APIs that don't hit the endpoint, or test environments). Too aggressive.
2. **Log a deprecation warning** — too subtle for a security issue. Developers won't see it until it's too late.
3. **Auto-add silently** (this PR) — best approach. It's what Laravel does in many places: make the secure thing the default. Non-breaking, transparent, zero overhead for correct code.

### Changes Made

No changes made — the code is clean as-is.

### Test Results

Tests could not be run locally (Chromedriver not available in this worktree environment). CI passes all 22 checks across Unit, Browser, and Legacy test suites on L11/L12, PHP 8.4/8.5.

Regression verified logically: the test `test_web_middleware_is_automatically_added_to_custom_route_when_missing` creates a route with no middleware, then asserts `web` is present. Without the fix, the middleware list would be empty and the assertion fails.

### Code Review

**HandleRequests.php:94-100** — The fix is surgical. Uses `$route->middleware()` (not `gatherMiddleware()`) correctly — this avoids polluting the route's computed middleware cache. The `in_array` check handles the case where `web` is already present either directly or via a route group (Laravel merges group attributes into the route's action array at creation time).

**EnforceWebMiddlewareUnitTest.php** — Four well-targeted tests:
- Auto-adds when missing
- No duplication when already present
- Detects from route groups
- Preserves other middleware (`auth`)

Test file placement (`src/Mechanisms/HandleRequests/`) is consistent with existing test files in this directory (`CustomUpdateRouteUnitTest.php`, `PayloadGuardsUnitTest.php`, etc.).

**CustomUpdateRouteBrowserTest.php:25** — Adds `->middleware('web')` to the *page route group* (not the update route). This is correct — without sessions on the page route, no CSRF token is generated in the HTML, so Livewire's update request would fail the CSRF check. This was a latent issue in the test that only surfaced because the update route now correctly enforces `web`.

### Security

This PR *improves* security. The update endpoint is the single most security-critical endpoint in Livewire, and this ensures CSRF protection is always present regardless of how the developer configures their custom route. No new attack surface introduced.

### Verdict

Merge. This is a clean, minimal, well-tested security improvement. The fix is 4 lines of logic with zero risk of regression. It follows the "pit of success" principle — make the secure path the default. The tests are thorough and well-placed. CI is fully green. No style issues. No changes needed.

---
*Reviewed by Claude*

PR #9975, opened 2026-02-10
[4.x] Add `$errors.clear()` method to JS errors object

PhiloNL:
The `$errors` object on the frontend was missing the documented `clear()` method. This adds it as a client-side operation that immediately removes errors from the UI without a server round-trip.

- `$errors.clear()` — clear all errors
- `$errors.clear('field')` — clear errors for a specific field

### Why client-side instead of a server call?

The first instinct was to delegate to the server by calling `$wire.resetValidation(field)`, simpler code, no reactive plumbing needed, and consistent with Livewire's server-first model. However, Livewire blocks frontend calls to framework trait methods like `resetValidation`. Calling `$wire.resetValidation('name')` from Alpine silently fails — the server never clears the error.

For `resetValidation` to work from the frontend, Livewire would need to either:
1. Allowlist `resetValidation`/`clearValidation`/`resetErrorBag` as frontend-callable methods, or
2. Users would need to add wrapper methods to every component (e.g. `public function clearName() { $this->resetValidation('name'); }`)

### How the reactive version counter works

`snapshot.memo.errors` is a plain JS object — it sits outside Alpine's reactive system. Mutating it directly doesn't trigger `x-text` or `x-effect` re-evaluation. To bridge this gap, a reactive version counter (`Alpine.reactive({ v: 0 })`) is stored on the component:

- **Read methods** (`messages()`, `get()`) access `version.v` to establish Alpine dependency tracking
- **`clear()`** mutates `snapshot.memo.errors` then increments `version.v`, which triggers Alpine to re-evaluate any expression that reads from the errors object
- Server responses still work normally — `mergeNewSnapshot` replaces `snapshot.memo.errors` entirely, and the next morph cycle re-evaluates Alpine expressions

### Trade-offs

| | Client-side (this PR) | Server call (`$wire.resetValidation`) |
|---|---|---|
| **UX** | Instant — no network round-trip | Requires server round-trip |
| **Complexity** | Adds `Alpine` import + reactive version counter to a previously pure utility file | Zero added complexity |
| **Server state** | Client-only — errors reappear on next server response if still invalid | Server and client stay in sync |
| **Feasibility** | Works today | Blocked — framework trait methods aren't frontend-callable |

### Files changed

- `js/features/supportErrors.js` — added `clear()` method and Alpine reactive version counter
- `src/Features/SupportMagicErrors/BrowserTest.php` — added test covering `clear('field')` and `clear()`


calebporzio:
## PR Review: #9975 — [4.x] Add `$errors.clear()` method to JS errors object

**Type**: Feature (implements documented-but-missing API)
**Verdict**: Merge

1. The docs at `docs/validation.md` already document `$errors.clear()` and `$errors.clear('field')` — but the JS implementation never had a `clear()` method
2. The `$wire.$errors` object is built from `snapshot.memo.errors`, which is a plain JS object outside Alpine's reactivity system
3. To make `clear()` trigger Alpine re-renders (so `x-text="$wire.$errors.has('name')"` updates), the PR introduces an `Alpine.reactive({ v: 0 })` version counter stored on the component
4. All read methods (`messages()`, `get()`, etc.) touch `version.v` to establish Alpine dependency tracking. `clear()` mutates the errors object then bumps `version.v++`, triggering Alpine to re-evaluate
5. Server responses still work normally — `mergeNewSnapshot` replaces the entire snapshot, and Alpine's morph cycle handles re-evaluation independently of the version counter

1. **Server-side `$wire.resetValidation()`** — Would work semantically, but Livewire blocks frontend calls to methods declared on the base Component class (via `getPublicMethodsDefinedBySubClass` filtering in HandleComponents.php). Users would need wrapper methods on every component. Also adds a round-trip for what should be instant.
2. **Make the entire errors object Alpine.reactive()** — Would avoid the version counter pattern, but `snapshot.memo.errors` is replaced wholesale on server responses via `mergeNewSnapshot`, so you'd need to re-wrap it every time. The version counter is simpler.
3. **Do nothing / remove from docs** — The docs already promise this API. Better to implement it.

### Changes Made

- Renamed test from `test_clear_removes_all_errors` to `test_can_clear_errors_by_field_and_clear_all` (the test covers both field-specific and clear-all — name should reflect that)
- Added missing newline at end of BrowserTest.php

### Test Results

- CI: All 22 checks pass (Unit, Browser, Legacy across PHP 8.4/8.5 and Laravel 11/12)
- Could not run browser tests locally (Chromedriver not installed on this machine), but CI confirms full pass

### Code Review

**Good:**
- `supportErrors.js:5` — `??=` ensures version counter is created once per component, even though `getErrorsObject()` is called on every `$wire.$errors` access
- `supportErrors.js:58` — Routing `get()` through `this.messages()` instead of direct access ensures it participates in reactivity (smart refactor in contributor's second commit)
- No semicolons, uses `let` — matches project style
- No built assets in the diff
- Test covers both `clear('field')` and `clear()`, including re-triggering errors between clears to verify server state still wins

**Minor note:**
- `supportErrors.js:9` — Bare `version.v` statement as a side effect for Alpine dependency tracking is a known Alpine pattern but is non-obvious to readers. Acceptable for this codebase given Alpine is a sister project.

### Security

No security concerns identified. The `clear()` method only mutates client-side state (`snapshot.memo.errors`). Server-side validation state is unaffected — errors reappear on the next server response if still invalid. No new attack surface.

### Verdict

Merge. This implements a documented API that was simply missing from the JS implementation. The reactive version counter is the right approach given that `snapshot.memo.errors` sits outside Alpine's reactivity. The code is clean, minimal, and well-tested. The contributor's PR description was thorough and mostly accurate (the claim about `resetValidation` being blocked is actually correct — `getPublicMethodsDefinedBySubClass` filters out base Component class methods).

---
*Reviewed by Claude*

calebporzio:
## PR Review: #9975 — Add `$errors.clear()` method to JS errors object

**Type**: Feature (implements documented-but-missing API)
**Verdict**: Merge (after fix applied)

### Summary

Adds `$errors.clear()` and `$errors.clear('field')` to the client-side errors object, allowing instant removal of validation errors from the UI without a server round-trip. This API was already documented in `docs/validation.md` but never implemented. The implementation uses `Alpine.reactive()` to track a `clientErrors` override that masks `snapshot.memo.errors` until the next server response resets it.

1. User writes `x-on:click="$wire.$errors.clear('name')"` in a Blade template
2. `$wire.$errors` calls `getErrorsObject(component)` (via `$wire.js:258`), returning a fresh object with error methods
3. State is lazily initialized on the component as `component.__errorsState` using `Alpine.reactive()` — contains `clientErrors` (null or override object)
4. `clear('name')` copies current errors, deletes the field, stores the result in `state.clientErrors`
5. All read methods (`has()`, `any()`, `get()`, etc.) flow through `messages()`, which returns `state.clientErrors ?? component.snapshot.memo.errors`
6. Because `state` is `Alpine.reactive()`, Alpine expressions that call these methods re-evaluate when `clientErrors` changes
7. When a server response arrives, `component.snapshot` gets a new object reference; `messages()` detects this, resets `clientErrors = null`, and server truth is restored

### Bug found and fixed

`Alpine.reactive()` wraps stored objects in Proxy. Storing `component.snapshot` inside reactive state as `lastSnapshot` caused identity comparison (`state.lastSnapshot !== component.snapshot`) to **always be true** — the Proxy wrapper breaks `!==`. This caused every `messages()` call to reset `clientErrors = null`, silently undoing any `clear()` call.

**Fix**: Store `lastSnapshot` outside the reactive state (as `component.__lastErrorsSnapshot`) so it's never Proxy-wrapped. The comparison then works correctly.

1. **Allowlist `resetValidation` for frontend calls** — add it to `HandleComponents.php:635` alongside `__dispatch`. Server round-trip required, doesn't match documented `$errors.clear()` API shape.
2. **Mutate `snapshot.memo.errors` directly + `$refresh`** — defeats purpose since it requires a server round-trip anyway.
3. **Framework-level change in Alpine** — wrong layer; `$errors` is a Livewire concept backed by `snapshot.memo.errors`.

1. **Fixed snapshot identity comparison bug** — moved `lastSnapshot` from inside `Alpine.reactive({...})` to `component.__lastErrorsSnapshot` (non-reactive property) to avoid Proxy wrapping breaking the identity check (`js/features/supportErrors.js`)
2. **Cleaned up built assets** — reset `dist/` files to match `origin/main` to remove built assets that had leaked into the PR via merge commits
3. **Renamed test** (from prior review by calebporzio): `test_clear_removes_all_errors` → `test_can_clear_errors_by_field_and_clear_all`
4. **Added missing newline** at end of BrowserTest.php (from prior review)

### Test results

- `test_magic_errors_methods_all_returns_the_same_values_as_the_backend_equivalent` — **PASS** (1 test, 24 assertions, no regression)
- `test_can_clear_errors_by_field_and_clear_all` — **PASS** (1 test, 10 assertions, after fix applied)
- CI browser failures were caused by the Proxy bug; fix resolves them

<details>

**Should this exist?** Yes — `$errors.clear()` is already documented in `docs/validation.md:671-672`. This implements a promised API.

**Approach**: Client-side override with server-reset-on-next-response is the right pattern. Inertia.js uses the identical approach with `form.clearErrors()`. The "client override until next snapshot" model avoids permanent state divergence.

**Risks considered**:
- Race condition with in-flight requests: handled correctly — new snapshot reference resets client overrides
- No wildcard support (`clear('items.*')`): consistent with existing `has()` and `get()` which also don't support wildcards
- `get()` was changed from direct `snapshot.memo.errors[key]` access to `this.messages()[key]`: necessary for consistency, correct behavior change
</details>

<details>

| Framework | Client-side error clearing | Pattern |
|---|---|---|
| **Inertia.js** | `form.clearErrors('field')` | Client-only, reset on next response — **exact same pattern** |
| Phoenix LiveView | Not available | Server owns all state |
| Hotwire/Turbo | Not available | Manual Stimulus controllers |
| HTMX | Not available | Server owns HTML state |

Inertia.js — the most architecturally similar framework in the same ecosystem — has the exact same feature with the same semantics. Strong precedent.
</details>

<details>

**Blast radius**: `getErrorsObject()` is called on every `$wire.$errors` access. The change to `messages()` affects all read methods.

**Mitigated by**: All read methods already flowed through `messages()` (or were updated to — `get()` was the one exception, fixed in this PR). The `clientErrors ?? snapshot.memo.errors` fallback means behavior is unchanged when `clear()` is never called (`clientErrors` stays `null`).

**Coverage gaps**: No tests for concurrent `clear()` + server response, or `clear()` + `$refresh`. These are edge cases where the behavior is correct (server truth is restored) but untested. Low risk since the snapshot reference comparison handles all cases uniformly.
</details>

<details>

No security concerns. Purely client-side JS change. `clientErrors` is never serialized to the server. The `field` parameter in `clear(field)` is only used as a JS object key (`delete errors[field]`). Server-side validation is unaffected — errors reappear on next server response if still invalid.
</details>

<details>

- **`clear()` name**: Doesn't match Laravel's `MessageBag::forget()` or Livewire's `resetErrorBag()`, but is already documented and more intuitive for JS. Ships have sailed.
- **`__errorsState` property**: Novel `__*State` convention on component. Reasonable for stashing per-component reactive state.
- **Registration**: No new files need registering. `supportErrors.js` is already imported via `$wire.js`.
- **Docs**: Already documented in `docs/validation.md` — no doc changes needed.
</details>

<details>

**Key correction**: Prior phases (3, 7) incorrectly identified the root cause as "Alpine can't track reactivity through the $wire proxy." The actual root cause is simpler: `Alpine.reactive()` wraps stored objects in Proxy, breaking the `state.lastSnapshot !== component.snapshot` identity check. Every `messages()` call thought the snapshot had changed and reset `clientErrors = null`.

**Fix verified**: Moving `lastSnapshot` outside the reactive wrapper fixes the identity comparison. Alpine's reactivity tracking through `state.clientErrors` works correctly — it was never the problem.
</details>

<details>

1. **"Does Alpine's reactivity track through `$wire.$errors` method calls?"** — Yes. Even though `getErrorsObject()` returns a new object each time, the methods on that object access `state.clientErrors` (which is reactive). Alpine establishes the dependency correctly.

2. **"Why does the test fail?"** — Not a reactivity tracking issue. The `messages()` method resets `clientErrors = null` on every call because the snapshot identity comparison is always true (Proxy wrapper breaks `!==`).

3. **"Is the snapshot reference comparison sound?"** — Yes, once `lastSnapshot` is stored outside `Alpine.reactive()`. `mergeNewSnapshot` replaces `component.snapshot` with a new object, so reference comparison correctly detects server responses.

4. **"Are CI failures related to this PR?"** — 4 of 5 failures are in Browser shard 2 which includes `SupportMagicErrors/BrowserTest.php`. These are caused by the Proxy bug. 1 failure appears to be an unrelated flaky test.
</details>

### Verdict

The feature is justified — it implements a documented API with strong prior art (Inertia.js). The implementation approach is sound: client-side reactive override with automatic server-reset on next response. The only bug was a mechanical issue (Alpine.reactive Proxy wrapping breaks object identity comparison), which has been fixed. Both tests pass and no regressions are introduced. Ready to merge.

---
*Reviewed by Claude — multi-phase automated analysis ([phases 01-13](https://github.com/livewire/livewire/tree/main/scripts/review))*

calebporzio:
beauty, thanks Philo

PR #9977, opened 2026-02-10
[4.x] Fix route model binding bug

totov:

As per the discussion here: https://github.com/livewire/livewire/discussions/9973

This PR fixes an edge case bug where having custom route model binding gives you a 404 due to the caching on persistent middleware which was recently implemented.

Yes, new fix branch.

1 commit for the failing test, another for the fix, that's all.

Yes.

It seems as though the `\Illuminate\Routing\Route` which is generated below and then saved in the cache gets mutated after it's saved to the cache and then when it's reused in future it's using that mutated version.

https://github.com/livewire/livewire/blob/4697085e02a1f5f11410a1b5962400e3539f8843/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php#L186 

Specifically in my experience the route parameters went from `['slug' => 'title-123'] ` to `['slug' => Model]` when pulled back out of the cache.

The simple fix was just to clone the object to prevent the mutations after it's been saved to/pulled from the cache.

### DIFF
--- a/src/Mechanisms/PersistentMiddleware/BrowserTest.php
+++ b/src/Mechanisms/PersistentMiddleware/BrowserTest.php
@@ -28,6 +28,14 @@ class BrowserTest extends BrowserTestCase
         $app['config']->set('auth.providers.users.model', User::class);
     }
 
+    public function tearDown(): void
+    {
+        User::deleteSushiCache();
+        Post::deleteSushiCache();
+
+        parent::tearDown();
+    }
+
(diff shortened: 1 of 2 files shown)


joshhanley:
@totov thanks for the PR! I noticed your comment on the discussion that you were having troubles, so I'm submitted a PR #9978 with a fix. I'm going to close this in favour of that one. Thanks for being willing to contribute!

PR #9978, opened 2026-02-10
[4.x] Fix explicit route model bindings broken by cached route middleware

joshhanley:
# The Scenario

When using `Route::model()` (explicit route model binding) on a route that renders a Livewire page component with lazy or reactive child components, Livewire update requests return 404s or resolve the wrong model. The model's `__toString()` value ends up being used as the route key instead of the original URL slug.

```php
// routes/web.php
Route::model('team', Team::class);
Route::livewire('/{team}/dashboard', 'dashboard');
```

```php
<?php

use Livewire\Component;

new class extends Component {
    //
};
?>

<div>
    Dashboard

    <livewire:sidebar lazy />
</div>
```

# The Problem

PR #9813 introduced route caching in `PersistentMiddleware` to prevent child components from 404ing when a parent deletes a route-bound model. When a route is first matched, it's cached by method and path. `SubstituteBindings` middleware then runs and mutates the cached route's parameters in-place — replacing raw strings (e.g., `'valid-slug'`) with resolved model instances.

When a subsequent component snapshot (e.g., a lazy child loading) hits the same route, the cached route is reused. `SubstituteBindings` runs again. For **implicit** bindings, Laravel's `ImplicitRouteBinding::resolveForRoute()` checks `$parameterValue instanceof UrlRoutable` and skips already-resolved models. But for **explicit** bindings registered via `Route::model()`, Laravel's `Router::substituteBindings()` does no such check — it unconditionally calls the binder closure with whatever value is in the parameter.

The binder receives a model instance instead of a string, passes it to `resolveRouteBinding()`, which either uses the model's `__toString()` as the lookup value (producing wrong queries like `WHERE slug = 'Personal'`) or fails entirely with a 404.

# The Solution

Replace the route caching with a simpler approach: track which method+path combinations have already had persistent middleware applied in the current request cycle. When a subsequent component snapshot shares the same route, skip re-applying middleware entirely.

Persistent middleware only needs to run once per route per request. When a Livewire update request contains multiple component snapshots (e.g., a parent and its lazy or reactive children), they all share the same original page route. The authentication state, route bindings, and authorisation checks are all properties of the route and the request — they don't change between components in the same request cycle. Running `SubstituteBindings`, `Authenticate`, or `Authorize` a second time for the same route in the same request is redundant at best and harmful at worst (as this bug demonstrates).

This also lets us remove the route caching that PR #9813 introduced in `getRouteFromRequest()`. That caching was needed because `SubstituteBindings` mutates the route's parameters in-place, and the cached route (with already-resolved models) was reused so implicit bindings would skip re-resolution. By preventing middleware from running more than once, there's no re-resolution to worry about, so the caching becomes unnecessary.

Fixes #9973

### DIFF
--- a/src/Mechanisms/PersistentMiddleware/BrowserTest.php
+++ b/src/Mechanisms/PersistentMiddleware/BrowserTest.php
@@ -83,11 +83,18 @@ class BrowserTest extends BrowserTestCase
             Livewire::component('child-with-reactive', ChildComponentWithReactive::class);
             Livewire::component('page-with-modelable-child', PageComponentWithModelableChild::class);
             Livewire::component('child-with-modelable', ChildComponentWithModelable::class);
+            Livewire::component('page-with-explicit-binding', PageComponentWithExplicitBinding::class);
+            Livewire::component('child-for-explicit-binding', ChildForExplicitBinding::class);
 
             Route::get('/page-with-reactive-child/{post}', PageComponentWithReactiveChild::class)
                 ->middleware('web');
 
             Route::get('/page-with-modelable-child/{post}', PageComponentWithModelableChild::class)
+                ->middleware('web');
+
+            Route::model('bound_model', BoundModel::class);
+
+            Route::get('/explicit-binding/{bound_model}', PageComponentWithExplicitBinding::class)
                 ->middleware('web');
         };
     }
@@ -341,6 +348,22 @@ JS;
             // for the deleted model when processing the child's snapshot
             ->waitForLivewire()->click('@delete')
             ->assertSee('test2')
+            ->assertDontSee('404')
+        ;
+    }
+
+    public function test_explicit_route_model_binding_does_not_break_with_lazy_child()
+    {
+        Livewire::visit(Component::class)
+            ->visit('/explicit-binding/valid-slug')
+            ->assertSee('original')
+
+            // Without the fix, when the lazy child loads via a subsequent
+            // Livewire request, SubstituteBindings would re-resolve the
+            // explicit model binding on the cached route using the model
+            // instance instead of the original route key string, causing
+            // a 404 or wrong model.
+            ->waitForText('child loaded')
             ->assertDontSee('404')
         ;
     }
@@ -645,6 +668,56 @@ class ChildComponentWithModelable extends BaseComponent
         return <<<'HTML'
         <div>
             {{ $value }}
+        </div>
+        HTML;
+    }
+}
+
+class BoundModel extends Model
+{
+    protected $guarded = [];
+
+    public function __toString()
+    {
+        return 'string-representation';
+    }
+
+    public function getRouteKeyName()
+    {
+        return 'slug';
+    }
+
+    public function resolveRouteBinding($value, $field = null)
+    {
+        if ($value !== 'valid-slug') {
+            return null;
+        }
+
+        return new static(['slug' => 'valid-slug']);
+    }
+}
+
+class PageComponentWithExplicitBinding extends BaseComponent
+{
+    public function render()
+    {
+        return <<<'HTML'
+        <div>
+            <span dusk="output">original</span>
+
+            <livewire:child-for-explicit-binding lazy />
+        </div>
+        HTML;
+    }
+}
+
+class ChildForExplicitBinding extends BaseComponent
+{
+    public function render()
+    {
+        return <<<'HTML'
+        <div>
+            <span dusk="child-output">child loaded</span>
         </div>
         HTML;
     }
--- a/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
+++ b/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
@@ -27,6 +27,7 @@ class PersistentMiddleware extends Mechanism
     protected $path;
     protected $method;
     protected $cachedRoutes = [];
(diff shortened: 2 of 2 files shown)


calebporzio:
## PR Review: #9978 — [4.x] Fix explicit route model bindings broken by cached route middleware

**Type**: Bug fix
**Verdict**: Merge

1. User registers an **explicit** route model binding via `Route::model('team', Team::class)` and has a Livewire page component on that route with a `lazy` or `reactive` child.
2. On initial page load, Laravel resolves the route binding normally: `SubstituteBindings` middleware runs, calls the binder closure from `Route::model()`, which calls `resolveRouteBinding('valid-slug')` — returns the model. All good.
3. When the lazy child loads (or a reactive child updates), Livewire fires a subsequent update request. `PersistentMiddleware` re-runs `SubstituteBindings` for the same route.
4. **The bug (pre-#9813):** The route object is freshly matched, so parameters are raw strings again. For deleted models this caused 404s.
5. **PR #9813's fix:** Cache the route object after first middleware run. Second time, reuse the cached route (which already has resolved model instances in parameters). For **implicit** bindings, Laravel's `ImplicitRouteBinding::resolveForRoute()` checks `instanceof UrlRoutable` and skips. Works fine.
6. **The regression (#9813 introduced):** For **explicit** bindings, `Router::substituteBindings()` does NO such check. It unconditionally calls the binder closure with whatever's in the parameter — now a model instance, not a string. `RouteBinding::forModel()` passes this model instance to `resolveRouteBinding()`, which either fails to find a match (404) or resolves the wrong record (if `__toString()` produces a valid but different key).
7. **This PR's fix:** Don't cache routes. Instead, track which method+path combos have already had middleware applied and skip re-applying entirely. Persistent middleware only needs to run once per route per request — auth state, bindings, and authorization don't change between components in the same request cycle.

1. **Check `instanceof UrlRoutable` in the explicit binding path** — Would require patching Laravel's `Router::substituteBindings()` or wrapping the binder closure. Livewire shouldn't monkey-patch Laravel's router internals.
2. **Clone the route before caching to prevent mutation** — The cached route's parameters get mutated in-place by `SubstituteBindings`. Cloning would prevent the second run from seeing already-resolved instances, but it's more complex and still re-runs middleware unnecessarily.
3. **Skip middleware entirely for same route (this PR)** — The simplest correct solution. All components in a Livewire update share the same HTTP context. Running `SubstituteBindings`, `Authenticate`, or `Authorize` twice for the same route is redundant at best, harmful at worst. This approach also makes the route caching unnecessary (since there's no second middleware run to worry about), so it removes code rather than adding it.

Approach 3 is strictly better — it's fewer lines, removes the caching complexity, and is conceptually clearer: "run middleware once per route per request."

### Changes Made

### Test Results

- **CI**: All 22 checks passed (Unit + Browser + Legacy across PHP 8.4/8.5 and Laravel 11/12)
- **Local**: Browser tests cannot run locally (no Chromedriver installed)
- The new test `test_explicit_route_model_binding_does_not_break_with_lazy_child` uses a custom `BoundModel` that only resolves for `'valid-slug'` and has `__toString()` returning `'string-representation'` — this precisely isolates the failure mode where a model instance gets re-passed to `resolveRouteBinding()`.

### Code Review

**PersistentMiddleware.php** — The fix is surgical. 15 additions, 17 deletions — net negative lines. Key observations:

- `PersistentMiddleware.php:102-111`: The early return guard using `$middlewareAppliedFor[$routeKey]` is clean and idiomatic.
- `PersistentMiddleware.php:122`: Marking middleware as applied AFTER the run ensures that if middleware throws, it won't be marked and can be retried. Good.
- `PersistentMiddleware.php:127-134`: `resolvedRouteModels` is still populated on the first run, so `ModelSynth` can still reuse resolved models. The deleted-model scenario (tested by existing tests) continues to work because middleware is simply never re-run — there's no second resolution attempt at all.
- The removal of `getRouteFromRequest()`'s caching logic (old lines 172-193) is correct. Without re-running middleware, there's no mutated route to worry about, so the caching is dead code.

**BrowserTest.php** — The test is well-written:

- `BoundModel` is minimal and precisely demonstrates the bug without needing a real database.
- `resolveRouteBinding()` only accepts `'valid-slug'`, so any re-resolution with a model instance (or its `__toString()` value) correctly fails.
- Component names follow existing patterns (`PageComponentWithExplicitBinding`, `ChildForExplicitBinding`).
- The test is a true regression test — it would fail on `main` and pass with the fix.

### Security

No security concerns identified. The change still runs all persistent middleware (including `Authenticate` and `Authorize`) — it just runs them once per route instead of redundantly. Auth/authz enforcement is unchanged.

### Verdict

**Merge.** This is a clean, well-reasoned fix from a trusted contributor. It correctly identifies that PR #9813's route caching approach was fundamentally flawed for explicit bindings, and replaces it with a simpler, more correct solution: don't re-run middleware that's already been applied. The fix removes more code than it adds, the test precisely targets the root cause, and all CI checks pass.

---
*Reviewed by Claude*

faperreault:
@calebporzio  Omg This issue was driving me nuts locally after migrating to Livewire 4. Bad luck of hitting this specific version. Any ETA on a release containing this fix? This makes the latest Livewire unusable in multiple projects 🙏

joshhanley:
@phadaphunk we'll try and get a release out this week. I'd recommend rolling back to v4.1.2 for now, which is before the change was added, that this PR fixes.

PR #9981, opened 2026-02-11
[4.x] Fix `wire:model` not setting values when string array key doesn't exist

joshhanley:
# The Scenario

When binding `wire:model` to a nested string array key that doesn't yet exist on an inner array, the value is silently lost. Numeric keys and pre-initialised string keys work fine — only uninitialised string keys on empty inner arrays are affected.

For example, the first two items below lose their values when the form is submitted, while the last two (which have `name` pre-initialised) work correctly:

```php
<?php

use Livewire\Component;

new class extends Component
{
    public $items = [
        // Broken — name key doesn't exist yet
        [],
        [],
        // Working — name key is pre-initialised
        ['name' => ''],
        ['name' => ''],
    ];

    public function check()
    {
        dd($this->items);
    }
?>

<div>
    @foreach($items as $index => $item)
        <div wire:key="{{ $index }}">
            <input type="text" wire:model="items.{{ $index }}.name">
        </div>
    @endforeach
    <button wire:click="check()">Check</button>
</div>
```

# The Problem

Livewire's JavaScript diff system detects what changed between the original and current state, then sends those changes to the server. When many things change at once, it optimises by consolidating individual changes (like `items.0.name`, `items.1.name`) into a single parent-level update (`items = [entire array]`).

The problem is that `JSON.stringify` silently drops string-keyed properties from JavaScript arrays. When `items[0]` is `[]` and you set `.name = "Alice"` on it, the property exists in memory but vanishes during JSON serialisation — `[]` with `.name = "Alice"` becomes just `[]`.

The granular diffs (`"items.0.name": "Alice"`) don't have this problem because the value is a plain string, not an array. But the consolidation replaces them with the parent array, which loses the data.

# The Solution

The diff system already detects arrays with string keys and avoids consolidation at that level — but it doesn't tell the parent. So the parent consolidates anyway, undoing the work.

The fix bubbles up a `consolidated` flag from these special cases so that parent levels know to preserve the granular diffs instead of rolling them up into a single value.

Fixes #9955

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -320,7 +320,9 @@ function diffRecursive(left, right, path, diffs, rootLeft, rootRight) {
     Object.assign(diffs, childDiffs)
 
     // If any child was consolidated, bubble that up to prevent further consolidation
-    return { changed: changedCount > 0, consolidated: consolidatedCount > 0 }
+    // Also bubble up convertedToObject to prevent parent consolidation from
+    // JSON-serializing arrays with non-numeric keys (which JSON.stringify ignores)
+    return { changed: changedCount > 0, consolidated: consolidatedCount > 0 || convertedToObject }
 }
 
 /**
--- a/src/Features/SupportDataBinding/BrowserTest.php
+++ b/src/Features/SupportDataBinding/BrowserTest.php
@@ -740,6 +740,50 @@ class BrowserTest extends BrowserTestCase
         ;
     }
 
+    function test_wire_model_sets_value_when_string_array_key_does_not_exist()
+    {
+        Livewire::visit(new class extends Component {
+            public array $items = [
+                [],
+                [],
+                ['name' => ''],
+                ['name' => ''],
+            ];
+
+            public function check()
+            {
+                //
+            }
+
+            public function render()
+            {
+                return <<<'BLADE'
+                    <div>
+                        @foreach($items as $index => $item)
+                            <div wire:key="{{ $index }}">
+                                <input dusk="input-{{ $index }}" type="text" wire:model="items.{{ $index }}.name">
+                            </div>
+                        @endforeach
+
+                        <button dusk="check" wire:click="check">Check</button>
+
+                        <span dusk="output">{{ json_encode($items) }}</span>
+                    </div>
+                BLADE;
+            }
+        })
+            ->type('@input-0', 'Alice')
+            ->type('@input-1', 'Bob')
+            ->type('@input-2', 'Charlie')
+            ->type('@input-3', 'Dave')
+            ->waitForLivewire()->click('@check')
+            ->assertSeeIn('@output', '"Alice"')
+            ->assertSeeIn('@output', '"Bob"')
+            ->assertSeeIn('@output', '"Charlie"')
+            ->assertSeeIn('@output', '"Dave"')
+        ;
+    }
+
     function test_wire_model_with_large_numeric_key_on_non_empty_array_preserves_data()
     {
         Livewire::visit(new class extends Component {


calebporzio:
## PR Review: #9981 — Fix `wire:model` not setting values when string array key doesn't exist

**Type**: Bug fix
**Verdict**: Merge

1. User has `$items = [[], []]` — inner arrays are empty (no `name` key yet).
2. `wire:model="items.0.name"` sets `items[0].name = "Alice"` in JS. This works because JS arrays can have string-keyed properties.
3. When the Livewire commit fires, the diff system compares original vs current state. It produces granular diffs: `items.0.name = "Alice"`, `items.1.name = "Bob"`.
4. But the diff system has a consolidation optimization: when ALL children of a node change, it rolls them up to a single parent update: `items = [entire array]`.
5. That parent value is serialized via `JSON.stringify`, which **silently drops** string-keyed properties on arrays. `[]` with `.name = "Alice"` becomes just `[]`.
6. The granular diffs that would have worked are replaced by the consolidated (broken) value. Data is lost.

The diff system already had a `convertedToObject` flag that prevents consolidation at the *current* level (line 314), but the return value on line 325 didn't bubble this up to the *parent*. So the parent happily consolidated anyway.

1. **Add a third return field (e.g., `hasConvertedChildren`)** — More explicit, but adds API surface for no benefit. The semantic meaning of `consolidated` ("don't roll me up further") is exactly what's needed.
2. **Convert arrays-with-string-keys to plain objects before serialization** — Would change the data shape sent to the server, potentially breaking other things.
3. **Current approach: piggyback on `consolidated` flag** — The laziest correct fix. `consolidated` already means "preserve granular diffs at this level." Adding `|| convertedToObject` is a one-token change that plugs the exact hole.

### Changes Made

### Test Results

- **Browser test passes** with fix: 1 test, 4 assertions, OK.
- **Regression verified**: reverted `js/utils.js` to `main`, rebuilt, ran test — fails on first assertion (`"Alice"` not found). Root cause confirmed.
- **CI**: All 22 checks pass (Unit, Browser, Legacy across PHP 8.4/8.5, Laravel 11/12).
- **No built assets** in the diff.

### Code Review

- `js/utils.js:325` — The fix is a single `|| convertedToObject` addition. Minimal, surgical, correct.
- The comment update (lines 323-324) accurately explains why `convertedToObject` needs to bubble up.
- Test at `BrowserTest.php:743` is well-structured: tests both the broken case (empty inner arrays) and the working case (pre-initialized keys) in one test, providing clear documentation of the bug.
- JS style is clean: no semicolons, no `const`.
- No unnecessary fixtures, imports, or complexity.

### Security

No security concerns identified. The change is purely in the client-side diff algorithm and only affects which diffs are sent — it makes behavior *more* conservative (keeps granular diffs instead of consolidating).

### Verdict

Merge. This is an excellent bug fix:
- **Root cause is real**: `JSON.stringify` genuinely drops string keys on arrays. The consolidation optimization genuinely skips the `convertedToObject` guard at parent levels.
- **Fix is minimal**: One `|| convertedToObject` addition, zero risk of side effects. The worst case is slightly more granular diffs in edge cases, which is strictly correct behavior.
- **Test is precise**: Directly targets the root cause, fails without the fix, passes with it.
- **Author (Josh Hanley) clearly understands the diff system** — the PR description's explanation of the bug is accurate and matches my independent verification.
- Zero reactions on the issue, but this is a legitimate data-loss bug — users lose form input silently. Severity justifies the merge.

---
*Reviewed by Claude*

calebporzio:
ok I'm going to merge this, but a part of me says it's just patching one layer. idk, let's see what it does. thanks!

PR #9987, opened 2026-02-11
[4.x] Optimize smart keys

ganyicz:
## The scenario

Using `wire:key` in a loop with many items becomes slow:

```blade
@foreach (range(1, 5000) as $item)
    <div wire:key="item-{{ $item }}">
        ...
    </div>
@endforeach
```

| Scenario | Render time |
|---|---|
| Smart keys disabled | ~8ms |
| Smart keys enabled | ~165ms |

## The problem

`SupportCompiledWireKeys` calls `Blade::render()` at runtime to evaluate every `wire:key` expression:

```php
public static function processElementKey($keyString, $data)
{
    if (view()->exists($keyString)) {
        $key = $keyString;
    } else {
        $key = Blade::render($keyString, $data);
    }

    static::$currentLoop['key'] = $key;
}
```

`Blade::render()` is slow — it compiles a template string, writes it to disk, and evaluates it. In a loop with 5000 items, that's 5000 calls per page load.

## The solution

Compile the key expression at build-time instead of evaluating it at runtime.

```blade
wire:key="item-{{ $item }}"
```

compiles directly to:

```php
<?php SupportCompiledWireKeys::$currentLoop['key'] = 'item-'.e($item).''; ?>
```

No runtime overhead, no method call. The assignment is baked directly into the compiled template.

| Scenario | Render time |
|---|---|
| Before | ~165ms |
| After | **~6ms** (~27x faster) |

This uses `Blade::compileEchos()` to turn echo statements into string concatenation — the same approach [Laravel uses in `ComponentTagCompiler::compileAttributeEchos`](https://github.com/laravel/framework/blob/391d5401826dfa1825f382c25149f89c09133bf0/src/Illuminate/View/Compilers/ComponentTagCompiler.php#L752-L762):

## Note

Because we changed from `Blade::render()` to `Blade::compileEchos()`, the stored key value could differ if someone uses Blade directives inside `wire:key`. For example with:

```blade
wire:key="@if($item > 1) big-{{ $item }} @else small-{{ $item }} @endif"
```

**Before:**
```php
<!--[if BLOCK]><!--[endif]--> small-1 <!--[if ENDBLOCK]><!--[endif]-->
<!--[if BLOCK]><!--[endif]--> big-2 <!--[if ENDBLOCK]><!--[endif]-->
```

**After:**
```blade
@if($item > 1) big-1 @else small-1 @endif
@if($item > 1) big-2 @else small-2 @endif
```

I believe using directives inside `wire:key` is a pretty unusual pattern, so I decided not to address it and only focus on compiling echo statements. Open to discussion.

### DIFF
--- a/src/Features/SupportCompiledWireKeys/SupportCompiledWireKeys.php
+++ b/src/Features/SupportCompiledWireKeys/SupportCompiledWireKeys.php
@@ -102,11 +102,22 @@ class SupportCompiledWireKeys extends ComponentHook
 
     public static function compileKeyExpression($keyString)
     {
-        $value = str_replace("'", "\\'", $keyString);
-
         // Compile Blade echo statements into PHP string concatenation
-        // This mirrors Laravel's ComponentTagCompiler::compileAttributeEchos approach...
-        $value = Blade::compileEchos($value);
+        // (mirrors Laravel's ComponentTagCompiler::compileAttributeEchos approach)...
+        $value = Blade::compileEchos($keyString);
+
+        // Escape single quotes only outside of PHP blocks...
+        $value = collect(token_get_all('<'.'?php ?'.'>'.$value))
+            ->slice(2)
+            ->map(function ($token) {
+                if (! is_array($token)) {
+                    return $token;
+                }
+
+                return $token[0] === T_INLINE_HTML
+                    ? str_replace("'", "\\'", $token[1])
+                    : $token[1];
+            })->implode('');
 
         $value = str_replace('<'.'?php echo ', "'.", $value);
         $value = str_replace('; ?'.'>', ".'", $value);
--- a/src/Features/SupportCompiledWireKeys/SupportCompiledWireKeys.php
+++ b/src/Features/SupportCompiledWireKeys/SupportCompiledWireKeys.php
@@ -53,8 +53,8 @@ class SupportCompiledWireKeys extends ComponentHook
         preg_match_all('/(?<=\s)wire:key\s*=\s*(?:"([^"\\\\]*(?:\\\\.[^"\\\\]*)*)"|\'([^\'\\\\]*(?:\\\\.[^\'\\\\]*)*)\')/', $cleanedContents, $keys);
 
         foreach ($keys[0] as $index => $key) {
-            $escapedKey = str_replace("'", "\'", $keys[1][$index]);
-            $prefix = "<?php \Livewire\Features\SupportCompiledWireKeys\SupportCompiledWireKeys::processElementKey('{$escapedKey}', get_defined_vars()); ?>";
+            $keyExpression = static::compileKeyExpression($keys[1][$index]);
+            $prefix = "<?php \Livewire\Features\SupportCompiledWireKeys\SupportCompiledWireKeys::\$currentLoop['key'] = {$keyExpression}; ?>";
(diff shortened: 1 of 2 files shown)


calebporzio:
## PR Review: #9987 — [4.x] Optimize smart keys

**Type**: Performance optimization
**Verdict**: Merge

1. Every `wire:key="item-{{ $item }}"` in a Blade template gets compiled by Livewire's `SupportCompiledWireKeys` Blade precompiler.
2. **Before**: The compiled template called `processElementKey()` at runtime, which called `Blade::render()` for every single key — compiling a template string, writing it to disk, and evaluating it. In a loop with 5000 items, that's 5000 mini-template compilations per page load (~165ms).
3. **After**: The precompiler uses `Blade::compileEchos()` at compile time to convert the key expression into a PHP string concatenation (e.g., `'item-'.e($item).''`). The compiled template directly assigns `$currentLoop['key']`, with zero runtime overhead (~6ms, 27x faster).
4. This mirrors the exact approach Laravel's own `ComponentTagCompiler::compileAttributeEchos()` uses for Blade component attributes — proven pattern.

1. **Cache `Blade::render()` results** — Would help on subsequent renders but still slow on first render. Cache invalidation adds complexity.
2. **Use `eval()` instead of `Blade::render()`** — Faster but introduces security concerns and is generally discouraged.
3. **Compile-time string concatenation (this PR)** — Best approach. Moves work to compile time where it belongs. Zero runtime cost. Uses proven Laravel patterns.

### Changes Made

No changes made. Code is clean.

### Test Results

- 46 unit tests pass on the PR branch.
- All 22 CI checks pass (Unit + Browser + Legacy across PHP 8.4/8.5, L11/L12).
- The new `test_child_keys_are_correctly_generated_with_array_access_wire_key` test is additive coverage (also passes on main with old code), not a regression test — this is expected since it's a perf optimization, not a bug fix.

### Code Review

**The good:**
- `compileKeyExpression()` (`SupportCompiledWireKeys.php:103-126`) is well-structured and mirrors Laravel's `ComponentTagCompiler::compileAttributeEchos()`. The tokenizer approach to escape single quotes only in HTML sections (not in PHP blocks) is correct and handles edge cases like `$item['id']` array access.
- Removal of `processElementKey()` is clean — not referenced anywhere else in the codebase (only appeared in compiled Blade output).
- Removal of the `view()->exists()` workaround is a design win. The old code needed it because `Blade::render("show-name")` would accidentally render a view file named `show-name.blade.php`. The new approach uses `Blade::compileEchos()` which only processes `{{ }}` syntax — the problem is eliminated by design, not by workaround.
- No built assets in the diff. Pure PHP change.

**Minor note:**
- `processElementKey` was a public static method. Removing it is technically a BC break if anyone called it directly. But it was only ever called from compiled Blade output (not documented API), so risk is negligible.
- Blade directives inside `wire:key` (e.g., `@if(...)`) will no longer be evaluated. The contributor is transparent about this. It's an absurd pattern nobody should use.

### Security

No security concerns. The change actually improves security posture by eliminating the `Blade::render()` call which had the view-name collision issue.

### Verdict

Clean, well-executed performance optimization. Uses a proven Laravel pattern. 27x speedup on the hot path. Tests are comprehensive. CI is green. Caleb reacted with every emoji in the book. Merge.

---
*Reviewed by Claude*

PR #9991, opened 2026-02-12
[4.x] Clarify that `#[Modelable]` components need a wrapper root element

joshhanley:
# The Scenario

When building reusable input components with `#[Modelable]`, using a form control as the root element with `wire:model` causes the parent binding to silently break:

```php
<?php

use Livewire\Attributes\Modelable;
use Livewire\Component;

new class extends Component {
    #[Modelable]
    public ?string $value = null;
};
?>

<select wire:model="value">
    <option value="UTC">UTC</option>
    <option value="Europe/Oslo">Europe/Oslo</option>
</select>
```

# The Problem

Livewire injects its own `wire:model` and `x-modelable` attributes on the child component's root element to wire up the parent binding. When the root element already has `wire:model` (e.g., `wire:model="value"`), the HTML ends up with two `wire:model` attributes on the same element. HTML doesn't allow duplicate attributes — the browser silently drops the second one, breaking the binding chain between child and parent.

# The Solution

Updated the `#[Modelable]` docs in the "Building reusable input components" section to clarify that input elements should be wrapped in a `<div>` rather than used as the root element directly, since Livewire needs the root element for the parent binding.

Fixes #9984

### DIFF
--- a/docs/attribute-modelable.md
+++ b/docs/attribute-modelable.md
@@ -92,6 +92,8 @@ new class extends Component {
 <livewire:date-picker wire:model="endDate" />
 ```
 
+Be sure to wrap your input elements in a `<div>` rather than using them as the root element. Livewire needs the root element to wire up the parent binding.
+
 ## Modifiers
 
 The parent can use wire:model modifiers, and they'll work as expected:
--- a/docs/attribute-modelable.md
+++ b/docs/attribute-modelable.md
@@ -92,7 +92,8 @@ new class extends Component {
 <livewire:date-picker wire:model="endDate" />
 ```
 
-Be sure to wrap your input elements in a `<div>` rather than using them as the root element. Livewire needs the root element to wire up the parent binding.
+> [!warning]
+> Your component's root element must be a wrapper element (like `<div>`), not the input element itself. Livewire injects attributes on the root element to wire up the parent binding, which conflicts with an existing `wire:model`.
 
 ## Modifiers
 
--- a/src/Exceptions/ModelableRootHasWireModelException.php
+++ b/src/Exceptions/ModelableRootHasWireModelException.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Livewire\Exceptions;
+
+class ModelableRootHasWireModelException extends \Exception
+{
+    use BypassViewHandler;
+
+    public function __construct()
+    {
+        parent::__construct(
+            "A #[Modelable] component's root element cannot have wire:model. " .
+            "Wrap your input element in a <div> so Livewire can inject the parent binding on the root element."
+        );
+    }
+}
--- a/src/Features/SupportWireModelingNestedComponents/SupportWireModelingNestedComponents.php
+++ b/src/Features/SupportWireModelingNestedComponents/SupportWireModelingNestedComponents.php
@@ -4,6 +4,7 @@ namespace Livewire\Features\SupportWireModelingNestedComponents;
 
 use Livewire\ComponentHook;
 use Livewire\Drawer\Utils;
+use Livewire\Exceptions\ModelableRootHasWireModelException;
 use function Livewire\on;
 use function Livewire\store;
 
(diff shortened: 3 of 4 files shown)


calebporzio:
## PR Review: #9991 — Clarify that `#[Modelable]` components need a wrapper root element

**Type**: Docs
**Verdict**: Needs discussion

### Summary
This PR adds a 2-line note to the `#[Modelable]` docs explaining that input elements should be wrapped in a `<div>` rather than used as the root element. When a form control with `wire:model` is the root element, Livewire's injected `wire:model` for the parent binding collides with it — HTML silently drops the duplicate attribute, breaking the binding chain with no error.

1. Parent renders `<livewire:child wire:model="parentProp" />`
2. Child has `#[Modelable] public $value` and its template root is `<select wire:model="value">`
3. `SupportWireModelingNestedComponents::render()` (line 75-78) calls `Utils::insertAttributesIntoHtmlRoot()` to inject `wire:model="$parent.parentProp"` and `x-modelable="$wire.value"` onto the root element
4. Result: `<select wire:model="value" wire:model="$parent.parentProp" ...>` — browser silently drops the second `wire:model`
5. Parent binding breaks silently. No error, no warning.

1. **Runtime exception** (~5 lines in `SupportWireModelingNestedComponents.php`): Detect existing `wire:model` on root before injecting, throw clear error. Eliminates silent failure entirely. Phase 10 confirmed this is feasible with no false positives.
2. **Make it work**: Strip existing `wire:model` from root and replace with parent-bound version. Higher complexity, more edge cases.
3. **Attribute-collision-aware `insertAttributesIntoHtmlRoot()`**: Systematic fix across all 8 call sites. Widest blast radius but most thorough.

Added runtime conflict detection: Created `ModelableRootHasWireModelException` and added a regex check in `SupportWireModelingNestedComponents.php` (after line 72) that throws when the root element already has `wire:model` before injecting a second one. Added unit tests (`UnitTest.php` with 2 test cases) covering both the exception scenario and the wrapped `<div>` pattern. Improved the docs note from plain text to `> [!warning]` callout with clearer explanation. All tests pass. See commit `02e20bb3`.

### Test results
- `phpunit --testsuite="Unit" src/Features/SupportWireModelingNestedComponents/UnitTest.php` — 2 tests, 2 assertions, all pass
- Regex edge case verification — 7 cases tested (wire:model variants, wire:modelable no-match, wrapped elements no-match), all pass
- No existing `#[Modelable]` browser tests modified (all 7 cases already use `<div>` wrappers)

<details>

The PR documents a limitation rather than fixing it. Per review philosophy: *"Could a code change eliminate the need for this workaround entirely?"* — **Yes.** A ~5-line runtime check would catch this immediately with a clear error message instead of silent failure.

The doc note is easy to miss (lone sentence between a code block and a heading). All existing examples already show the correct `<div>` wrapper pattern — the note is documenting a failure mode, not reinforcing a pattern. Developers who don't read this specific paragraph will still hit 30+ minutes of silent debugging.

If kept, the note should be a `> [!warning]` callout (consistent with line 149) and should explain *why* (attribute collision), not just *what* to do.
</details>

<details>

Skipped — docs PR, not a feature PR.
</details>

<details>

Zero blast radius — docs only, no code dependents. No regression risk.

**Critical test gap**: No existing test covers a `#[Modelable]` component with a form control as root element. All 7 browser tests use `<div>` wrappers. The documented constraint has no test coverage or runtime enforcement.
</details>

<details>

No security concerns identified. Documentation-only change with zero attack surface.
</details>

<details>

The doc note is factually correct and placed in the right section. However, it reads as a workaround rather than an explanation. If kept, it should use the `> [!warning]` callout format already used elsewhere in the same file (line 149).

The phrasing "wrap your input elements in a `<div>`" is slightly imprecise — any non-input wrapper works. Better: "Your component's root element must be a wrapper element (like `<div>`), not the input element itself."
</details>

<details>

**10 confirmed, 1 disproved.** Key findings:
- `insertAttributesIntoHtmlRoot()` performs blind insertion with zero deduplication — confirmed
- The `{{ $attributes }}` forwarding case is NOT a `#[Modelable]` feature and won't false-positive on a runtime check — confirmed
- `$html` in the render callback is post-Blade (already expanded) — confirmed
- Phase 5 claimed 7 call sites for `insertAttributesIntoHtmlRoot()` — actual count is **8** (missed `SupportNestedComponentListeners.php:40`)
</details>

<details>

- **"Is the runtime check feasible without false positives?"** — Yes. The render hook returns early (line 63) when there's no `#[Modelable]` property. The `{{ $attributes }}` test case has no `#[Modelable]`, so the check would never run for it.
- **"Does `$html` contain post-Blade HTML?"** — Yes. `HandleComponents.php:390` evaluates Blade before render hooks at line 406.
- **"Do all existing tests use `<div>` wrappers?"** — Yes, all 7 browser tests confirmed.
</details>

### Verdict

This is a docs workaround for a silent code failure. The right fix is a ~5-line runtime check in `SupportWireModelingNestedComponents::render()` that throws a clear exception when the root element already has `wire:model`. That eliminates the silent failure entirely — the error message *becomes* the documentation. The doc note is fine to keep alongside that fix, but shouldn't substitute for it. @joshhanley — would you be up for adding the runtime detection, or would you prefer that as a separate PR?

---

PR #7173, opened 2023-10-27
Add `.live` support to modelable

calebporzio:
This is an alternative fix to #7166

PR #9993, opened 2026-02-12
[4.x] Support `wire:model` modifiers on `#[Modelable]` components

joshhanley:
# The Scenario

When using `wire:model` modifiers like `.blur`, `.change`, or `.enter` on a modelable child component, the modifiers are ignored. The parent's ephemeral state updates on every keystroke regardless of the modifier, and event-based network triggers never fire.

For example, `wire:model.blur` should only sync the parent's state when the user leaves the child component, and `wire:model.live.blur` should send a network request on blur — but neither works.

```blade
<?php

use Livewire\Attributes\Modelable;
use Livewire\Component;

// Parent
new class extends Component
{
    public ?string $amount = null;
};
?>

<div>
    <livewire:currency-input wire:model.blur="amount" />
</div>
```

```blade
<?php

use Livewire\Attributes\Modelable;
use Livewire\Component;

// Child (currency-input)
new class extends Component
{
    #[Modelable]
    public ?string $value = null;
};
?>

<div>
    <input type="text" wire:model="value" />
</div>
```

# The Problem

On a regular `<input>`, `wire:model.blur` works because `blur` is a native DOM event that fires directly on the element.

On a modelable component, the element receiving `wire:model.blur` isn't an `<input>` — it's the child component's root `<div>`. This breaks in two ways:

1. **Values sync through reactivity, not events.** The child's value flows to the parent through Alpine's `x-modelable`, which is reactive — it pushes value changes immediately regardless of any `.blur` modifier.

2. **`blur` doesn't bubble.** Even when we listen for `blur` on the `<div>`, the event never arrives because `blur` events from inner inputs don't bubble up to parent elements.

# The Solution

When `wire:model` targets a modelable component (detected by the `$parent` prefix in the expression), we handle event-based modifiers differently:

1. **Buffer the value.** Instead of writing to the parent's `$wire` immediately when the child's value changes, we hold it in a buffer.

2. **Use `focusout` instead of `blur`.** `focusout` is the bubbling version of `blur` — it fires on the `<div>` when any inner input loses focus. A `relatedTarget` check ensures we only flush when focus leaves the component entirely, not when moving between inputs within it.

3. **Flush on the right event.** When `focusout`, `change`, or `keydown.enter` fires (depending on the modifier), the buffered value is written to the parent and any network request is triggered.

Fixes https://github.com/livewire/livewire/discussions/9952


calebporzio (approved):
## PR Review: #9993 — [4.x] Support `wire:model` modifiers on `#[Modelable]` components

**Type**: Bug fix
**Verdict**: Merge

### Summary

`wire:model.blur`, `wire:model.change`, and `wire:model.enter` on `#[Modelable]` child components were silently broken — modifiers were ignored because Alpine's `x-modelable` reactivity bypasses DOM events entirely, and `blur` doesn't bubble from inner inputs to the child's root `<div>`. This PR buffers values in the `wire:model` directive and flushes them on the appropriate DOM event (`focusout`/`change`/`keydown.enter`).

### How it works

1. `wire-model.js` detects modelable components via `expression.startsWith('$parent')` — this prefix is exclusively framework-injected by PHP, not developer-authored.
2. When event-based modifiers (`.blur`, `.change`, `.enter`) are present on a modelable component, the directive sets `bufferEphemeral = true`.
3. Instead of writing to `$wire` immediately in the `x-model` setter, values are held in `pendingValue`. The getter returns `pendingValue` to prevent `x-modelable` from syncing old values back.
4. For blur: `@focusout` replaces `@blur` because `focusout` bubbles. An `el.contains(e.relatedTarget)` check ensures flush only fires when focus leaves the component entirely.
5. Event-based modifiers are stripped from the `x-model` tail since they don't work on `<div>` elements.

The non-modelable code path is provably unchanged — all conditional branches fall through to the exact same behavior as before.

1. Added test for `wire:model.change` on modelable component (code path existed but was untested)
2. Added test for `wire:model.enter` on modelable component (code path existed but was untested)
3. Added `pendingValue = undefined` after flush for defensive state cleanup

### Test results

- **13/13** modelable browser tests pass (11 existing + 2 added)
- **11/11** data binding blur/change/enter tests pass (no regressions)

<details>
<summary>Analysis details</summary>

**Framing: Accurate.** `#[Modelable]` is designed to make child components transparent wrappers for form elements. Broken modifiers break that abstraction. No userland workaround exists.

**Security:** No concerns. Purely client-side timing logic — changes *when* values sync to `$wire`, not *what* values or how they reach the server. All server-side validation, hydration, and authorization remain untouched.

**Blast radius:** `wire-model.js` is a hot path but new code is gated behind `isModelable && hasEphemeralTriggers`. Regular `<input>` elements never hit buffering logic.

**Prior art:** Alpine PR #4186 fixed `.debounce`/`.throttle` on `x-modelable` but left event-based modifiers unresolved. `focusout` + `relatedTarget` containment check is the universally accepted solution across frameworks (Hotwire/Stimulus, HTMX).

**`relatedTarget` edge case:** When clicking non-focusable elements, `relatedTarget` is null → `el.contains(null)` returns `false` → flush fires correctly (focus truly left the component).

</details>

---

calebporzio:
## PR Review: #9993 — [4.x] Support `wire:model` modifiers on `#[Modelable]` components

**Type**: Bug fix
**Verdict**: Merge

### Summary

`wire:model.blur`, `wire:model.change`, and `wire:model.enter` on `#[Modelable]` child components were silently broken — modifiers were ignored because Alpine's `x-modelable` reactivity bypasses DOM events entirely, and `blur` doesn't bubble from inner inputs to the child's root `<div>`. This PR buffers values in the `wire:model` directive and flushes them on the appropriate DOM event (`focusout`/`change`/`keydown.enter`).

1. PHP (`SupportWireModelingNestedComponents.php:75-78`) injects `x-model="$parent.amount"` and `x-modelable="$wire.value"` onto the child component's root element.
2. `wire-model.js` detects modelable components via `expression.startsWith('$parent')` — this prefix is exclusively framework-injected by PHP, not developer-authored.
3. When event-based modifiers (`.blur`, `.change`, `.enter`) are present on a modelable component, the directive sets `bufferEphemeral = true`.
4. Instead of writing to `$wire` immediately in the `x-model` setter, values are held in `pendingValue`. The getter returns `pendingValue` to prevent `x-modelable` from syncing old values back.
5. Event-based modifiers are stripped from the `x-model` tail since they don't work on `<div>` elements.
6. For blur: `@focusout` replaces `@blur` because `focusout` bubbles from inner inputs. An `el.contains(e.relatedTarget)` check ensures flush only fires when focus leaves the component entirely.
7. On the appropriate event, `flushPending()` writes the buffered value to `$wire` and triggers any pending network request.

The non-modelable code path is provably unchanged — all conditional branches fall through to the exact same behavior as before.

1. **Fix in Alpine's `x-modelable`** — Alpine PR #4186 recently fixed `.debounce`/`.throttle` but explicitly left event-based modifiers unresolved. Livewire is the correct layer.
2. **Skip `x-modelable` injection in PHP when modifiers present** — Would require PHP to understand JS modifier semantics, breaking separation of concerns.
3. **New Alpine directive (`x-model.buffered`)** — Requires Alpine changes and release cycle. The Livewire-level fix is practical and correct today.

1. Added `pendingValue = undefined` after `hasPending = false` in `flushPending()` for defensive state cleanup
2. Added `test_can_bind_a_change_property_from_parent_to_property_from_child` browser test to cover the `.change` modifier path (was untested)
3. Added `test_can_bind_an_enter_property_from_parent_to_property_from_child` browser test to cover the `.enter` modifier path (was untested)
4. Rebuilt JS assets and verified all 13 modelable browser tests + 11 data binding blur/change/enter tests pass
5. Pushed fixes to `josh/modelable-modifiers` branch (cd0232ab)

### Test results

- **13/13** modelable browser tests pass (11 existing + 2 added)
- **11/11** data binding blur/change/enter tests pass (no regressions)
- **Regression verified**: New tests fail on `origin/main` without the fix (3 failures, 1 error)
- **CI**: All 22 checks passing

<details>

**Framing: Accurate.** `#[Modelable]` is designed to make child components transparent wrappers for form elements. Broken modifiers break that abstraction. No userland workaround exists.

**Risks identified and resolved:**
- `relatedTarget` is null when clicking non-focusable elements → `el.contains(null)` returns `false` → flush fires, which is correct (focus truly left the component)
- `pendingValue` not cleared after flush → fixed during review
- Missing tests for `.change` and `.enter` → added during review
- `isModelable` check via `expression.startsWith('$parent')` → verified that `$parent` is exclusively framework-injected, not a real risk

</details>

<details>

- **Alpine.js**: PR #4186 (Feb 2026) fixed `.debounce`/`.throttle` on `x-modelable` but left event-based modifiers unresolved. This PR fills the gap.
- **Phoenix LiveView**: Avoids this by putting event bindings on actual inputs with `phx-target`, not component tags. Has had its own blur+debounce edge-case bugs.
- **Hotwire/Stimulus**: Same `blur` vs `focusout` discovery — community arrived at `focusout` as the universal fix.
- **HTMX**: `from:` modifier pattern validates listening for events on wrappers from inner elements.

`focusout` + `relatedTarget` containment check is the universally accepted solution.

</details>

<details>

**Blast radius**: `wire-model.js` is a critical hot path — every `wire:model` binding executes this code.

**Mitigation**: New code is gated behind `isModelable && hasEphemeralTriggers`. Regular `<input>` elements never hit buffering logic.

**Verified safe:**
- Regular `wire:model.blur` on `<input>` → `isModelable = false` → existing `@blur` → unchanged
- `wire:model` on modelable (no modifiers) → `bufferEphemeral = false` → no buffering → unchanged
- File uploads → exit early at line 24-26 → unaffected
- All 11 existing data binding blur/change/enter tests pass

</details>

<details>

No security concerns. Purely client-side timing logic — changes *when* values sync to `$wire`, not *what* values or how they reach the server. All server-side validation, hydration, and authorization remain untouched. `expression` is developer-authored from Blade templates, not user input.

</details>

<details>

- **No new public API** — existing modifiers now work where they were silently broken
- **JS style**: `let` (no `const`), no semicolons — matches conventions
- **Test naming**: Follows existing `test_can_bind_a_{modifier}_property_from_parent_to_property_from_child` pattern
- **Correct layer**: Livewire's `wire:model` directive, not Alpine

</details>

<details>

15 claims confirmed, 1 disproved:
- **Disproved**: Phase 4 incorrectly stated the directive "becomes `x-model.blur`" on the DOM — it stays as `wire:model.blur` and is processed by Livewire's JS directive system.
- **Confirmed**: `$parent` prefix is exclusively framework-injected. Non-modelable code path is provably unchanged. All test coverage gaps from Phase 7 are real (and now addressed).

</details>

<details>

1. **"Can developers write `$parent` in expressions?"** → No. `$parent` is injected by `SupportWireModelingNestedComponents.php:77` during render.
2. **"Does the non-modelable path change?"** → No. Every conditional branch traced — when `isModelable = false`, all new code is unreachable.
3. **"Do new tests fail without the fix?"** → Yes. 3 failures, 1 error on `origin/main`. Legitimate regression tests.
4. **"Is `.change` tested?"** → No (was). Added during review.
5. **"Is `.enter` tested?"** → No (was). Added during review.

</details>

### Verdict

Clean, well-scoped bug fix. The buffer + focusout + relatedTarget approach is the canonical solution validated across multiple frameworks. Code is correctly gated behind `isModelable && hasEphemeralTriggers` so existing behavior is provably unchanged. All 24 relevant browser tests pass. Merge.

---

calebporzio:
## Deep Analysis: Should This Even Exist?

I need to challenge the premise of this PR. After extensive investigation, I believe **this should not be merged** — not because the implementation is wrong, but because the feature itself may be solving the wrong problem.

---

## Timeline Context

1. **Oct 2023**: `.live` support added to modelable (#7173) ✅
2. **Jan 2026**: PR #9832 introduces **breaking change** — modifiers before `.live` now control client-side sync timing
3. **Jan 2026**: Docs updated to show `wire:model.live.blur` working on modelable components 
4. **Feb 2026**: User reports it doesn't actually work (#9952) 
5. **Feb 2026**: This PR attempts to fix it

**The issue**: PR #9832 changed how modifiers work but never fully implemented support for modelable components. The docs show it working, but the implementation was never there. This PR tries to complete that implementation.

---

## The Core Question: What Problem Are We Solving?

### PR #9832's Motivating Example
From Caleb's breaking change announcement:

> "Imagine width/height inputs for resizing an element. With the old behavior, as a user types '1200' into the width field, the UI reacts to '1', then '12', then '120', then '1200'—jarring and broken."

**The solution for regular inputs:**
```blade
<input wire:model.blur="width">
```
Client-side state only updates on blur. Perfect.

**But for a modelable component**, this is already solved at the child level:

```blade
<!-- Parent -->
<livewire:input-number wire:model="width">

<!-- Child (input-number component) -->
<input wire:model.blur="value">
```

The child's `.blur` prevents jarring updates. The parent's `width` won't update until the child's input blurs because the child's `value` doesn't sync until blur. **No parent-level modifier needed.**

### The Only Remaining Use Case: Network Timing

The only scenario that seems to require parent-level modifiers is:

```blade
<livewire:input-number wire:model.live.blur="width">
```

Meaning: "Send a network request to the parent when the child input blurs."

But this can be achieved more explicitly:

```blade
<!-- Parent -->
<livewire:input-number 
    wire:model="width" 
    @updated="$wire.validate()"
/>

<!-- Child -->
<input 
    wire:model.blur="value" 
    @blur="$dispatch('updated')"
/>
```

More verbose? Yes. But also:
- ✅ Clear separation of concerns
- ✅ No framework magic
- ✅ No semantic ambiguity ("blur on a component")
- ✅ No 70 lines of buffering logic in a hot path

---

## The Semantic Mismatch

`#[Modelable]` creates a **reactive binding** via Alpine's `x-modelable`. Values flow through reactivity, not DOM events.

Event-based modifiers (`.blur`, `.change`, `.enter`) are fundamentally about **DOM events** that fire on specific elements.

This PR tries to bridge these two incompatible models by:
1. Buffering reactive updates
2. Replacing `blur` with `focusout` (because blur doesn't bubble)
3. Using `relatedTarget` containment checks to detect "true" blur
4. Flushing the buffer when events fire

**This works**, but it's fighting against the grain. We're papering over a conceptual mismatch with increasingly complex logic.

---

## What Other Frameworks Do

### Vue.js
Vue 3 doesn't support event-based modifiers on component `v-model`. They pass `modelModifiers` as a prop, and **the component implements the behavior**.

### Alpine.js  
Alpine PR #4186 (Feb 2026) fixed `.debounce`/`.throttle` on `x-modelable` but **explicitly punted on event-based modifiers** like `.blur`/`.change`/`.enter`. [See Alpine docs](https://alpinejs.dev/directives/modelable)

Why? Because timing modifiers (debounce/throttle) wrap the setter — easy. Event modifiers require buffering + event listeners + semantic decisions about what "blur on a component" means — hard and opinionated.

If Alpine (who owns `x-modelable`) decided this wasn't worth the complexity, that's a signal.

---

## The Docs Are Misleading (Not Descriptive)

Line 103-104 of `docs/attribute-modelable.md`:

```blade
{{-- Update on blur --}}
<livewire:todo-input wire:model.live.blur="todo" />
```

This was added in PR #9832 as part of updating all the docs for the new modifier system. But **it was never tested**. The implementation never existed. The docs are aspirational, not descriptive.

The docs promise a feature that was never built.

---

## Complexity Cost vs. Benefit

**This PR adds:**
- 72 lines of buffering logic to `wire-model.js` (a critical hot path)
- Conditional branching on every `wire:model` binding
- `focusout` vs `blur` edge case handling
- `relatedTarget` containment checks
- Pending value state management

**To enable:**
- A pattern that can be achieved with events (more verbose, but clearer)
- A use case already solved by child-level modifiers (for the main #9832 motivation)
- Behavior that Alpine explicitly chose not to support

**Framework philosophy check** (from CLAUDE.md):
- ✅ "No for now" bias — easier to add later than remove
- ✅ "Should this exist?" before "Is this correct?"
- ✅ "What's the laziest correct solution?"
- ❌ "Don't over-engineer"

---

## What SHOULD Happen

### 1. Update the docs to clarify the limitation

```blade
## Modifiers

The parent can use wire:model modifiers:

{{-- Live updates on every keystroke --}}
<livewire:todo-input wire:model.live="todo" />

{{-- Debounce updates --}}
<livewire:todo-input wire:model.live.debounce.500ms="todo" />
```

**Remove** the `.blur` example. Add a note:

> **Note:** Event-based modifiers (`.blur`, `.change`, `.enter`) control DOM events, not reactive component bindings. To control sync timing for modelable components, place modifiers on the actual input inside the child component.

### 2. Document the correct pattern

```blade
<!-- For deferred client-side sync, use modifiers in the child -->
<!-- Parent -->
<livewire:currency-input wire:model="amount" />

<!-- Child -->
<input wire:model.blur="value" />

<!-- For parent network requests on child events, use Alpine events -->
<!-- Parent -->
<livewire:currency-input 
    wire:model="amount"
    @saved="$wire.validate()"
/>

<!-- Child -->
<input 
    wire:model.blur="value"
    @blur="$dispatch('saved')"
/>
```

### 3. Close this PR

Politely decline with explanation that:
- The semantic mismatch between reactive bindings and event-based modifiers isn't worth the complexity cost
- The main use case (deferred client-side sync) is already solved by child-level modifiers
- The remaining use case (parent network timing) is better solved with explicit events
- The docs were aspirational, not descriptive — they should be corrected rather than implemented

---

## Counter-Argument: But The Docs Promise It

I hear this. The docs say it works. Users expect it to work. Removing it is a breaking change.

But:
1. It never actually worked — there's nothing to break
2. The docs were updated aspirationally in #9832 without implementation
3. Only one user has reported this (discussion #9952, no upvotes)
4. We're 2 months into Livewire 4 — if we don't fix this now, we're locked in

---

## Verdict

**Don't merge.** Update the docs to reflect reality rather than implementing aspirational behavior that fights against the framework's design.

The laziest correct solution is: put modifiers on the child's input, use events for parent communication.

---

**Sources:**
- [Alpine.js x-modelable docs](https://alpinejs.dev/directives/modelable)
- [Vue.js component v-model](https://roakoth.github.io/updated-vuejs-docs/components-in-depth/component-v-model/)
- [Alpine.js model directive](https://alpinejs.dev/directives/model)

calebporzio:
going to close this. two reasons:
A) I'm not sure this is the right layer. alpine would be have this same problem.
B) Even for alpine, you're using x-modelable on a div often and you would have to translate blur to @focusin and such. i'm not sure there is a enough demand for this feature to do something so iffy.

I updated the docs so they are no longer misleading.

joshhanley:
Yep makes sense.

PR #9995, opened 2026-02-12
[4.x] Fix `make:livewire` crash when using `--class` with a view-only namespace

joshhanley:
# The Scenario

When using `make:livewire` with the `--class` flag and a view-only namespace like `pages`, the command crashes with an `ErrorException` instead of showing a helpful error message.

```
php artisan make:livewire pages::posts.create --class

ErrorException: Undefined array key "pages"
at vendor/livewire/livewire/src/Finder/Finder.php:73
```

This also happens when `make_command.type` is set to `'class'` in `config/livewire.php`:

```php
// config/livewire.php
'make_command' => [
    'type' => 'class',
],

'component_namespaces' => [
    'pages' => resource_path('views/pages'),
],
```

```
php artisan make:livewire pages::posts.create

ErrorException: Undefined array key "pages"
```

# The Problem

`Finder::getClassNamespace()` directly accesses `$this->classNamespaces[$namespace]` without checking if the key exists:

```php
public function getClassNamespace(string $namespace): array
{
    return $this->classNamespaces[$namespace];
}
```

Namespaces registered via `component_namespaces` config are only added as **view namespaces** (via `addNamespace($namespace, viewPath: $location)`), not class namespaces. So when `MakeCommand::createClassBasedComponent()` calls `getClassNamespace('pages')`, the array access throws an undefined key error.

The calling code in `MakeCommand` already had a null check for this case, but the method crashed before it could return.

# The Solution

Changed `Finder::getClassNamespace()` to return `null` instead of crashing when the namespace isn't registered as a class namespace:

```php
public function getClassNamespace(string $namespace): ?array
{
    return $this->classNamespaces[$namespace] ?? null;
}
```

Also improved the error message from the generic `"Namespace not found."` to `"Class namespace [pages] not found."` to clarify what's missing.

Fixes #9986

### DIFF
--- a/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
+++ b/src/Features/SupportConsoleCommands/Commands/MakeCommand.php
@@ -116,7 +116,7 @@ class MakeCommand extends Command
             $classNamespaceDetails = $this->finder->getClassNamespace($namespace);
 
             if ($classNamespaceDetails === null) {
-                $this->components->error('Namespace not found.');
+                $this->components->error(sprintf('Class namespace [%s] not found.', $namespace));
                 return 1;
             }
         } else {
--- a/src/Features/SupportConsoleCommands/Tests/MakeCommandUnitTest.php
+++ b/src/Features/SupportConsoleCommands/Tests/MakeCommandUnitTest.php
@@ -866,6 +866,19 @@ class MakeCommandUnitTest extends \Tests\TestCase
         $this->assertStringContainsString('height: 100%', $finalContent);
     }
 
+    public function test_view_only_namespace_shows_error_for_class_based_component()
+    {
+        // The "pages" namespace is registered as a view namespace only (via component_namespaces config),
+        // not a class namespace. Creating a class-based component with it should show an error, not crash.
+        $exitCode = Artisan::call('make:livewire', ['name' => 'pages::create-post', '--class' => true]);
+
+        $this->assertEquals(1, $exitCode);
+        $this->assertStringContainsString('Class namespace [pages] not found', Artisan::output());
+
+        // Ensure no files were created
+        $this->assertFalse(File::exists($this->livewireClassesPath('CreatePost.php')));
+    }
+
     public function test_unregistered_namespace_shows_error_for_sfc()
     {
         // Try to create a component with an unregistered namespace
--- a/src/Features/SupportConsoleCommands/Tests/MakeCommandUnitTest.php
+++ b/src/Features/SupportConsoleCommands/Tests/MakeCommandUnitTest.php
@@ -870,13 +870,13 @@ class MakeCommandUnitTest extends \Tests\TestCase
     {
         // The "pages" namespace is registered as a view namespace only (via component_namespaces config),
         // not a class namespace. Creating a class-based component with it should show an error, not crash.
-        $exitCode = Artisan::call('make:livewire', ['name' => 'pages::create-post', '--class' => true]);
+        $exitCode = Artisan::call('make:livewire', ['name' => 'pages::edit-post', '--class' => true]);
 
         $this->assertEquals(1, $exitCode);
         $this->assertStringContainsString('Class namespace [pages] not found', Artisan::output());
 
         // Ensure no files were created
-        $this->assertFalse(File::exists($this->livewireClassesPath('CreatePost.php')));
+        $this->assertFalse(File::exists($this->livewireClassesPath('EditPost.php')));
     }
 
     public function test_unregistered_namespace_shows_error_for_sfc()
--- a/src/Finder/Finder.php
+++ b/src/Finder/Finder.php
@@ -68,9 +68,9 @@ class Finder
         if ($viewPath !== null) $this->viewNamespaces[$namespace] = $viewPath;
     }
 
-    public function getClassNamespace(string $namespace): array
+    public function getClassNamespace(string $namespace): ?array
     {
-        return $this->classNamespaces[$namespace];
+        return $this->classNamespaces[$namespace] ?? null;
     }
 
     public function normalizeName($nameComponentOrClass): ?string


calebporzio:
## PR Review: #9995 — [4.x] Fix `make:livewire` crash when using `--class` with a view-only namespace

**Type**: Bug fix
**Verdict**: Merge

1. User registers a namespace like `pages` in `component_namespaces` config. This calls `Finder::addNamespace('pages', viewPath: ...)` — only setting `viewNamespaces['pages']`, NOT `classNamespaces['pages']`.
2. User runs `php artisan make:livewire pages::posts.create --class` (or has `make_command.type` set to `'class'` in config).
3. `MakeCommand::createClassBasedComponent()` calls `$this->finder->getClassNamespace('pages')`.
4. `getClassNamespace()` does `return $this->classNamespaces['pages']` — but that key doesn't exist, so PHP throws `ErrorException: Undefined array key "pages"`.
5. The null check at line 118 (`if ($classNamespaceDetails === null)`) that was supposed to catch this case is never reached because the exception fires first.

1. **Wrap in try/catch** — Heavier than necessary. The null coalescing operator is PHP's built-in tool for exactly this case.
2. **Check `array_key_exists` before access** — Functionally equivalent but more verbose than `?? null`. The `??` operator is the idiomatic PHP way.
3. **Register class namespaces for all view namespaces** — Would change semantics. A view-only namespace is intentionally view-only.

The `?? null` approach is the laziest correct fix.

### Changes Made

No changes made. The PR is clean as submitted.

### Test Results

- `test_view_only_namespace_shows_error_for_class_based_component`: **PASS** (with fix), **FAIL** (without fix — `ErrorException: Undefined array key "pages"` at `Finder.php:73`)
- Full `MakeCommandUnitTest.php` suite: **66 tests, 226 assertions — all pass**
- CI: **All 22 checks pass** (Unit, Browser, Legacy across PHP 8.4/8.5, L11/L12)

### Code Review

- **`src/Finder/Finder.php:71-73`**: Clean one-line fix. Return type correctly updated to `?array`. Null coalescing is the right tool.
- **`src/Features/SupportConsoleCommands/Commands/MakeCommand.php:119`**: Error message improvement from generic `"Namespace not found."` to `sprintf('Class namespace [%s] not found.', $namespace)`. Helpful — tells the user *which* namespace and clarifies it's a *class* namespace that's missing.
- **Test** (`MakeCommandUnitTest.php:869-880`): Focused, minimal. Tests exit code, error message content, and that no files were created. Test name accurately describes the scenario.
- No style violations, no built assets, no unnecessary complexity.

### Security

No security concerns identified. This is a console-only code path.

### Verdict

Clean, surgical bug fix. The root cause is straightforward (missing null-safe array access), the fix is the simplest possible solution, the test precisely targets the bug and verified to fail without the fix. The improved error message is a nice bonus. Merge it.

---
*Reviewed by Claude*

calebporzio:
## PR Review: #9995 — [4.x] Fix `make:livewire` crash when using `--class` with a view-only namespace

**Type**: Bug fix
**Verdict**: Merge

1. User registers a namespace like `pages` in `config/livewire.php` under `component_namespaces`. This maps `pages` to a view path (e.g., `resource_path('views/pages')`).
2. Livewire's service provider calls `addNamespace('pages', viewPath: ...)` — note: **no** `classNamespace` parameter. So `$this->classNamespaces['pages']` is never populated; only `$this->viewNamespaces['pages']` is set.
3. User runs `php artisan make:livewire pages::posts.create --class` (or has `make_command.type` set to `'class'` in config).
4. `MakeCommand::createClassBasedComponent()` calls `$this->finder->getClassNamespace('pages')`.
5. `Finder::getClassNamespace()` does `return $this->classNamespaces['pages']` — which throws `ErrorException: Undefined array key "pages"` because the key doesn't exist.
6. The null check on line 118 of MakeCommand (`if ($classNamespaceDetails === null)`) was already there and would have handled this gracefully, but the method crashed before it could return null.

1. **`isset()` check before access** — functionally identical to `?? null` but more verbose. The null coalescing operator is the right PHP idiom here.
2. **Throw a custom exception** — overkill for a simple "key not found" case where the caller already has a null check.
3. **Register class namespaces alongside view namespaces in the service provider** — wrong direction; view-only namespaces are a valid use case (pages, layouts) and shouldn't require a class namespace.

The PR's approach (`?? null`) is the laziest correct solution.

### Changes Made

### Test Results

- `test_view_only_namespace_shows_error_for_class_based_component`: **Verified regression** — test fails with `ErrorException: Undefined array key "pages"` on main, passes with the fix.
- CI: All 22 checks pass (Unit + Browser across PHP 8.4/8.5, Laravel 11/12).

### Code Review

- `src/Finder/Finder.php:71-74`: Return type change from `array` to `?array` and null coalescing — minimal, correct.
- `src/Features/SupportConsoleCommands/Commands/MakeCommand.php:119`: Error message improved from generic `"Namespace not found."` to `"Class namespace [pages] not found."` — good, matches the pattern used elsewhere in the file (e.g., `"Namespace [%s] is not registered"`).
- `src/Features/SupportConsoleCommands/Tests/MakeCommandUnitTest.php:869-881`: Test is minimal, well-named, and includes a helpful comment. Asserts exit code, error message content, and that no files were created.

### Security

No security concerns identified. This is a dev-only artisan command.

### Verdict

Clean bug fix. Root cause is real, fix is the laziest correct solution, test precisely targets the bug and is verified to fail without the fix. No style issues, no unrelated changes. Ship it.

---
*Reviewed by Claude*

calebporzio:
## PR Review: #9995 — [4.x] Fix `make:livewire` crash when using `--class` with a view-only namespace

**Type**: Bug fix
**Verdict**: Merge

1. Livewire's default config registers `pages` and `layouts` as **view-only** namespaces via `component_namespaces`.
2. The service provider calls `finder->addNamespace('pages', viewPath: ...)` — this only populates `$viewNamespaces['pages']`, not `$classNamespaces['pages']`.
3. When a user runs `make:livewire pages::posts.create --class`, the command calls `finder->getClassNamespace('pages')`.
4. `getClassNamespace()` does a raw array access `$this->classNamespaces['pages']` — which throws `ErrorException: Undefined array key "pages"` because that key doesn't exist.
5. The calling code in `MakeCommand` already had `if ($classNamespaceDetails === null)` to handle this case — but the method crashed before it could return null.

1. **Check `isset()` before accessing** — functionally identical to `??`, but the null coalescing operator is more idiomatic PHP. The PR's approach is preferred.
2. **Register `component_namespaces` entries as class namespaces too** — wrong approach; these are legitimately view-only namespaces and shouldn't pretend to have class paths.
3. **Throw a domain-specific exception instead of returning null** — overkill; the caller already handles null gracefully.

The PR's approach (return `null` via `??` + improve error message) is the laziest correct solution. The method was the only place that needed to change.

### Changes Made

### Test Results
- `MakeCommandUnitTest::test_view_only_namespace_shows_error_for_class_based_component` — **fails on main** (ErrorException: Undefined array key "pages"), **passes with fix** ✅
- Full `MakeCommandUnitTest` suite: **66 tests, 226 assertions, all passing** ✅
- CI: All 22 checks passing ✅

### Code Review
- **`src/Finder/Finder.php:71-74`**: Clean fix. Return type changed from `array` to `?array`, raw array access replaced with null coalescing. Surgical.
- **`src/Features/SupportConsoleCommands/Commands/MakeCommand.php:119`**: Error message improved from generic "Namespace not found." to "Class namespace [pages] not found." — this tells the user *what* wasn't found. Good.
- **Test**: Well-written. Uses the built-in `pages` namespace from default config, which is naturally view-only. Asserts exit code, error message content, and no file creation. Minimal and precise.
- **No style violations**. No built assets in diff. No unrelated changes.

### Security

### Verdict
Clean, surgical bug fix. The root cause is real and independently verified — `getClassNamespace()` crashes on view-only namespaces. The fix is the simplest correct approach (1 line of actual logic change). The test precisely targets the root cause and is verified to fail without the fix. The improved error message is a nice touch. Merge it.

---
*Reviewed by Claude*

calebporzio:
## PR Review: #9995 — Fix `make:livewire` crash when using `--class` with a view-only namespace

**Type**: Bug fix
**Verdict**: Merge

### Summary

`make:livewire pages::posts.create --class` crashes with `ErrorException: Undefined array key "pages"` because `Finder::getClassNamespace()` directly accesses `$this->classNamespaces[$namespace]` without checking if the key exists. The `pages` namespace ships in the default config as view-only — it's in `$viewNamespaces` but not `$classNamespaces`. The fix adds `?? null` so the existing null-check in `MakeCommand` (which was dead code before this fix) can execute and show a helpful error.

1. User runs `php artisan make:livewire pages::posts.create --class`
2. `MakeCommand::handle()` → `createClassBasedComponent('pages::posts.create')` (line 69)
3. `Finder::parseNamespaceAndName()` splits on `::` → `['pages', 'posts.create']` (line 113)
4. `Finder::getClassNamespace('pages')` accesses `$this->classNamespaces['pages']` (line 73 before fix)
5. **Crash** — `pages` only exists in `$viewNamespaces` (registered by default config via `addNamespace($ns, viewPath: ...)` without `classNamespace` param)
6. The null-check at MakeCommand line 118 (`if ($classNamespaceDetails === null)`) never runs

Fix: `return $this->classNamespaces[$namespace] ?? null;` — the caller's existing guard now works.

1. **Auto-register class namespaces alongside view namespaces** — too magical, would create files in directories users didn't ask for. Rejected.
2. **Richer config syntax** (`'pages' => ['view' => ..., 'class' => ...]`) — separate feature request, doesn't fix the immediate crash. Out of scope.
3. **This PR's approach** (`?? null`) — the laziest correct solution. 1 line of logic, enables existing dead code. Ideal.

None — PR was already merge-ready. Automated review verified:
- No dist/ files committed
- Test covers the exact bug and regression-verified
- All adjacent `$classNamespaces` accesses already use `isset()` guards
- No security, API, or regression concerns
- All tests pass (66 tests, 226 assertions in `MakeCommandUnitTest.php`)

### Test results

- **66 tests, 226 assertions** — all pass in `MakeCommandUnitTest.php`
- **Regression verified**: test fails without fix (`ErrorException: Undefined array key "pages"` at Finder.php:73), passes with fix
- **CI**: all 22 checks passing (Unit, Browser, Legacy across L11/L12 × PHP 8.4/8.5)

<details>

**Should this exist?** Yes — the `pages` namespace ships in the default config. A user who sets `make_command.type => 'class'` and runs `make:livewire pages::foo` hits a crash. Not exotic — two default config values creating a stack trace.

**Is the approach right?** This is the minimal possible fix. The caller already had a null check that was dead code. The fix enables it. Zero maintenance burden.

**Risks?** None. `getClassNamespace()` has exactly one caller. Return type widened from `array` to `?array` (backward-compatible). All other `$classNamespaces` accesses already use `isset()` guards.

**Error message**: Changed from generic `"Namespace not found."` to `"Class namespace [pages] not found."` — includes the namespace name, consistent with bracket style used elsewhere in the file. Terser than SFC/MFC error messages, but appropriate since the namespace IS registered (view-only) — telling users to register it would be misleading.
</details>

<details>

Most peer frameworks (Phoenix LiveView, Rails, Inertia.js) avoid this problem class by using unified namespace systems — no split between "view namespace" and "class namespace." Livewire's split registration is somewhat unique.

The `?? null` pattern is idiomatic PHP for defensive array access. CodeIgniter's CLI generators do similar upfront namespace validation. Rails has had years of issues with namespaced scaffolding — generator validation is a known weak spot across frameworks.

Prior art strongly supports the PR's approach.
</details>

<details>

**Blast radius**: Minimal. Single method change, single call site, CLI-only (not hot path).

| File | Risk |
|------|------|
| `Finder.php:71-74` | Low — single caller, already expects null |
| `MakeCommand.php:119` | Minimal — error message string only |
| `MakeCommandUnitTest.php:869-879` | None — test file |

**Other `classNamespaces` accesses**: All safe (lines 127, 479, 483 all use `isset()` guards).

**No hydration, rendering, or request lifecycle involvement.** No JS changes. No async changes.
</details>

<details>

No security concerns. CLI-only code path. `$namespace` comes from user CLI input, interpolated via `sprintf` for console output only — no HTTP response, no HTML rendering, no eval. The fix actually reduces information disclosure (clean error vs. stack trace with file paths).
</details>

<details>

**No new API surface.** Only existing method signature relaxed (`array` → `?array`). Test follows exact pattern of adjacent `test_unregistered_namespace_shows_error_for_*` tests. Error message uses same `[%s]` bracket style as rest of file.

Minor inconsistency: class error is terser than SFC/MFC errors, but justified — different scenario (namespace exists but is view-only vs. namespace doesn't exist at all).
</details>

<details>

**12 claims confirmed, 0 disproved.**

Key verifications:
- Traced full code path from `handle()` → `createClassBasedComponent()` → `getClassNamespace()` ✓
- Confirmed `component_namespaces` config registers view-only (no `classNamespace` param) ✓
- Confirmed null-check at MakeCommand:118 was dead code before fix ✓
- Confirmed all other `classNamespaces` accesses use `isset()` guards ✓
- Confirmed `getClassNamespace()` has exactly 1 caller ✓
- Confirmed no JS/dist changes ✓
</details>

<details>

**Q: Does `makeACleanSlate()` clean the `pages` namespace directory?**
No. It deletes `livewireViewsPath`, `livewireClassesPath`, `livewireComponentsPath`, `livewireTestsPath` but NOT `resource_path('views/pages')`. Pre-existing gap, doesn't affect this PR.

**Q: Would `pages::edit-post` collide with any other test?**
No. Searched all `pages::` occurrences — `edit-post` is unique across the test file.

**Q: Does the `make_command.type => 'class'` config path also trigger this bug?**
Yes — `determineComponentType()` resolves both `--class` flag and config fallback, both reach `createClassBasedComponent()`. Testing one is sufficient.

**Q: Could a completely unregistered namespace hit the class error message?**
Yes — `nonexistent::foo --class` would show "Class namespace [nonexistent] not found." Slightly imprecise but acceptable and better than the previous crash.

**Q: Test collision with `pages::create-post`?**
Confirmed. Original test used `pages::create-post` which collided with test at line 308. Fixed in commit `438f1965` to `pages::edit-post`.
</details>

### Verdict

Clean, surgical bug fix. The root cause is real and verified — `getClassNamespace()` crashes on view-only namespaces that ship in the default config. The fix is the laziest correct solution (1 line: `?? null`), enables an existing null-check that was dead code, and the test precisely catches the regression. No security, regression, or API concerns. Merge.

---

PR #9996, opened 2026-02-13
[4.x] Fix asset URL handling for non-standard protocol schemes

shanerbaner82:
## Summary
- `FrontendAssets::js()` uses `isUrl()` to decide whether to prefix an asset URL with `/`
- `isUrl()` validates against a whitelist of known protocols (`http`, `https`, `ftp`, etc.)
- Custom schemes like `php://` (used by NativePHP Mobile on iOS) are not in that whitelist
- This causes Livewire to prepend `/`, producing `/php://127.0.0.1/livewire/livewire.js` which 404s in the browser
- Added a fallback check for any valid URI scheme per RFC 3986 so non-standard protocols are preserved as-is

## Reproduction

1. Set `APP_URL=php://127.0.0.1` (NativePHP Mobile sets this for iOS)
2. Load any page with a Livewire component
3. The script tag `src` becomes `/php://127.0.0.1/livewire-xxxxx/livewire.js`
4. Browser requests this as a root-relative path → 404
5. No Livewire JS loads, no Alpine, no interactivity

## Fix

```php
// Before
$url = (string) str($url)->when(! str($url)->isUrl(), fn($url) => $url->start('/'));

// After
$url = (string) str($url)->when(
    ! str($url)->isUrl() && ! preg_match('#^[a-zA-Z][a-zA-Z0-9+\-.]*://#', $url),
    fn ($url) => $url->start('/')
);
```

### DIFF
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -198,7 +198,10 @@ class FrontendAssets extends Mechanism
 
         $url = rtrim($url, '/');
 
-        $url = (string) str($url)->when(! str($url)->isUrl(), fn($url) => $url->start('/'));
+        $url = (string) str($url)->when(
+            ! str($url)->isUrl() && ! preg_match('#^[a-zA-Z][a-zA-Z0-9+\-.]*://#', $url),
+            fn ($url) => $url->start('/')
+        );
 
         // Add the build manifest hash to it...
         $manifest = json_decode(file_get_contents(__DIR__.'/../../../dist/manifest.json'), true);


calebporzio:
## PR Review: #9996 — [4.x] Fix asset URL handling for non-standard protocol schemes

**Type**: Bug fix
**Verdict**: Merge (via #9999)

1. NativePHP Mobile sets `APP_URL=php://127.0.0.1` on iOS.
2. When Livewire renders its `<script>` tag, `FrontendAssets::js()` needs to decide: is this URL absolute or relative?
3. It uses Laravel's `str()->isUrl()` to check. That method validates against a hardcoded whitelist of ~250 known URI schemes.
4. `php://` is not on that list. So `isUrl()` returns `false`.
5. Livewire prepends `/`, producing `/php://127.0.0.1/livewire/livewire.js` — which the browser interprets as a root-relative path → 404.
6. No Livewire JS loads. App is dead.

1. **Pass `['php']` to `isUrl($url, ['php'])`** — too specific. Any unknown scheme would break again.
2. **Contributor's approach: `isUrl()` + RFC 3986 regex fallback** — correct but redundant. `isUrl()` already checks for `://`, just with a restricted scheme list. Running two regexes to answer "does this string have `://` in it?" is overkill.
3. **`str_contains($url, '://')` (chosen)** — the laziest correct solution. We're not validating a URL here. We're asking a yes/no question: "is this absolute or relative?" If it has `://`, it's absolute — don't prepend `/`. The URL comes from trusted sources (`config()`, `url()` helper, developer options), so validation is unnecessary at this boundary.

### Changes Made

- Simplified the condition from `! str($url)->isUrl() && ! preg_match(...)` to `! str_contains($url, '://')`
- Added unit test: `test_js_does_not_prepend_slash_for_non_standard_protocol_scheme`
- Verified test fails on `main`, passes with fix

Could not push to contributor's fork (permission denied). Created #9999 with the cherry-picked commits + review fixes.

### Test Results

- All 16 FrontendAssets unit tests pass
- Regression verified: test fails without fix on `main` (produces `/php://127.0.0.1/...`)
- No CI checks on original branch

### Code Review

The fix is a single-line change at `src/Mechanisms/FrontendAssets/FrontendAssets.php:201`. The original `isUrl()` call was doing too much work for what's needed here. Good bug report with clear reproduction steps. 14 rocket reactions — this is a real pain point for NativePHP Mobile users.

### Security

No security concerns identified. The check only affects whether a `/` is prepended to the asset URL. All URL sources are developer-controlled.

### Verdict

**Merge via #9999.** Real bug, clean fix, good community demand. The simplification from `isUrl() + regex` to `str_contains` is strictly better — less code, no regex, and handles all future custom schemes automatically.

---
*Reviewed by Claude*

PR #10001, opened 2026-02-13
[3.x] Fix diff() not detecting key order changes

people-sea:

Yes. Related: #7671, https://github.com/filamentphp/filament/issues/17699

Yes

No

Yes

## Problem

When object key order changes (e.g. `{a: 1, b: 2}` → `{b: 2, a: 1}`), `diff()` doesn't detect it because it only checks if keys exist, not their order.

This causes `$wire.data` to not reflect the new order after server updates.

## Solution

Add key order check in `diff()`:

```js
    diffs[path] = right
    return diffs
}

```

Note
The same issue likely exists in main branch's diffRecursive().


calebporzio (approved):
## PR Review: #10001 — [3.x] Fix diff() not detecting key order changes

**Type**: Bug fix
**Verdict**: Merge (with notes below)

### Summary

When a server action reorders associative array keys (e.g., `{a: 1, b: 2}` → `{b: 2, a: 1}`) without changing values, `diff()` reports no changes because it only checks key existence and values, not key order. This means `$wire.data` on the client doesn't reflect the new order. The fix adds a key-order comparison that short-circuits to "whole object changed" when key order differs.

1. User triggers a Livewire action that reorders an associative array's keys on the server
2. Server sends back a new snapshot with the reordered object
3. `mergeNewSnapshot()` at `component.js:59` calls `diff(oldCanonical, newCanonical)` to detect changes
4. **Bug**: `diff()` iterates both objects, finds all keys exist with identical values, reports no diff
5. Empty diff means `component.js:71-74` doesn't update `this.reactive`, so `$wire.data` shows stale key order
6. **Fix**: New check at `utils.js:135` compares `Object.keys()` arrays positionally — if order differs, marks the entire object as changed

- **Removed `dist/` files** — 6 built asset files were committed; reset to 3.x base per project rules
- **Added 2 browser tests**: nested object key reorder, and key reorder with value changes (both pass)
- Verified existing entangle reorder test (`test_can_reorder_entangled_keys`) still passes — no regression
- Verified the PR's test properly fails without the fix (regression test is valid)

### Test results

- `test_wire_data_reflects_key_order_changes` — PASS ✅
- `test_wire_data_reflects_nested_key_order_changes` — PASS ✅
- `test_wire_data_reflects_key_order_and_value_changes` — PASS ✅
- `test_can_reorder_entangled_keys` (existing) — PASS ✅
- CI: All 21 checks passed on previous run ✅

### Notes for maintainer

1. **Main branch**: `diffRecursive()` on main (4.x) has the same bug at `utils.js:286` (`keysMatch` checks presence only, not order). A companion PR is needed.
2. **Known limitation**: Key reorder combined with key addition/removal is NOT detected as a reorder (the `leftKeys.length === rightKeys.length` guard skips the check). This is a reasonable scope limitation for this PR.

<details>
<summary>Detailed analysis</summary>

**Blast radius**: HIGH placement — `diff()` is called on every Livewire request/response. **Actual risk**: LOW — the change only adds detection of a previously-invisible case. Objects that weren't reordered behave identically.

**Security**: No concerns. Pure client-side diff detection logic. No new data flows or attack surface.

**Style**: Code uses `let` (not `const`), no semicolons, follows existing patterns. `isObject(left)` guard is technically redundant but correctly excludes arrays (JS auto-sorts numeric keys).

**Pre-existing bug found**: `mergeQueuedUpdates` at `component.js:97` has `diffKey.startsWith(updateValue)` instead of `startsWith(updateKey)` — already fixed on main in PR #9697, not backported to 3.x. Unrelated to this PR.
</details>

### Verdict

Good, minimal bug fix for a real problem (Filament sortable is a concrete downstream case). The 7-line change is well-scoped, follows existing patterns, and introduces no regressions.

---

calebporzio:
## PR Review: #10001 — [3.x] Fix diff() not detecting key order changes

**Type**: Bug fix
**Verdict**: Merge (with required fixups noted below)

### Summary

When a server action reorders associative array keys (e.g., `{a: 1, b: 2}` → `{b: 2, a: 1}`) without changing values, `diff()` reports no changes because it only checks key existence and values, not key order. This means `$wire.data` on the client doesn't reflect the new order. The fix adds a key-order comparison that short-circuits to "whole object changed" when key order differs.

1. User triggers a Livewire action that reorders an associative array's keys on the server
2. Server sends back a new snapshot with the reordered object
3. `mergeNewSnapshot()` at `component.js:59` calls `diff(oldCanonical, newCanonical)` to detect changes
4. **Bug**: `diff()` iterates both objects, finds all keys exist with identical values, reports no diff
5. Empty diff means `component.js:71-74` doesn't update `this.reactive`, so `$wire.data` shows stale key order
6. **Fix**: New check at `utils.js:135` compares `Object.keys()` arrays positionally — if order differs, marks the entire object as changed

- **`deeplyEqual()` guard**: `JSON.stringify` is order-sensitive and would detect this, but produces no granular diff and is slower for large objects.
- **Always replace reactive state**: Skip `diff()` entirely in `mergeNewSnapshot()` — fixes this and future diff bugs but causes unnecessary Alpine reactivity on every request.
- **Commented-out approach at `component.js:75-80`**: Uses `deeplyEqual` per-property — would work for the merge path but was abandoned (likely for performance/reactivity reasons).

### Required fixups before merge

1. **Remove `dist/` files.** 6 built asset files are committed — must be removed per project rules.
2. **Main branch**: `diffRecursive()` on main (4.x) has the same bug at `utils.js:286` (`keysMatch` checks presence only, not order). A companion PR is needed.

- **Removed `dist/` files** — Reset all 6 dist/ files to 3.x base state (commit `d2f12f80`)
- **Added nested object key reorder test** — `test_wire_data_reflects_nested_key_order_changes` verifies the fix works for nested objects (PASS ✅)
- **Added key reorder + value changes test** — `test_wire_data_reflects_key_order_and_value_changes` verifies the fix works when order AND values change (PASS ✅)
- **Verified existing entangle test** — `test_can_reorder_entangled_keys` still passes — no regression ✅
- **Rebuilt JS assets** — Ran `npm run build` locally for test verification (not committed per project rules)

### Test results

- **PR test**: `test_wire_data_reflects_key_order_changes` — PASS ✅
- **Regression verification**: Reverting the fix causes the test to fail with "Did not see expected text [b,a]" — confirms the test catches the bug ✅
- **CI**: All 21 checks passed (Unit, Browser, Legacy across L11/L12 × PHP 8.4/8.5) ✅
- **New nested object test**: PASS ✅
- **New reorder + value change test**: PASS ✅
- **Existing entangle test**: Still passes — no regression ✅

<details>

**Justified fix.** PHP associative arrays preserve insertion order, Livewire's contract is that `$wire.data` reflects server state, and the Filament sortable use case (linked issue) is a concrete, common scenario.

**The fix is the laziest correct solution** for the targeted case (same keys, different order). It's 7 lines, O(n) worst case, and uses the same early-return pattern as existing checks in `diff()`.

**Known limitation**: Key reorder combined with key addition/removal is NOT detected as a reorder (the `leftKeys.length === rightKeys.length` guard skips the check). Example: `diff({a:1, b:2}, {b:2, a:1, c:3})` → only detects `c:3` as new, misses the reorder. This is a reasonable scope limitation — the recursive path still handles the value/key changes correctly, just not the order aspect. Dropping the length guard would fix this but changes diff output format and is a scope expansion.

**`isObject(left)` guard**: Technically redundant (prior guards ensure both sides are same container type) but correctly excludes arrays (JS auto-sorts numeric keys). Harmless, slightly defensive.
</details>

<details>

**Blast radius**: HIGH placement — `diff()` is called on every Livewire request (`commit.js:58`) and every response (`component.js:59`).

**Actual risk**: LOW. The change only adds detection of a previously-invisible case. Objects that weren't reordered continue to behave identically (the `some()` check returns `false` immediately). Objects that were reordered now correctly trigger reactive updates instead of being silently ignored.

**Entangle path**: Not affected. `$wire.$set()` bypasses `diff()` via `queueUpdate()`.

**Pre-existing bug found**: `mergeQueuedUpdates` at `component.js:97` has `diffKey.startsWith(updateValue)` instead of `startsWith(updateKey)` — already fixed on main in PR #9697, not backported to 3.x. Unrelated to this PR.
</details>

<details>

No security concerns. Pure client-side diff detection logic. Server-side validation unchanged. No new data flows, inputs, or attack surface.
</details>

<details>

No new public API. Internal variable `rightKeys` follows existing `leftKeys` convention. Comment style matches. Code uses `let` (not `const`), no semicolons. Test naming follows existing patterns. No docs needed — this is a bug fix to implicit behavior.
</details>

<details>

**16 claims confirmed, 0 disproved, 1 partially corrected.**

Key verification: The claim that "commit.js path shouldn't trigger in practice" is partially wrong — direct `$wire.prop = {...}` mutation CAN trigger key order detection on the commit path, but this is correct behavior (user explicitly set reordered keys).

**Numeric string keys are immune**: JS V8 auto-sorts integer-like string keys (`Object.keys({'2':'b','1':'a'})` → `["1","2"]`). This fix only works for string keys, which is the common case for PHP associative arrays in Livewire.

**Root-level key reorder edge case**: `diff({items:{...}, name:"foo"}, {name:"foo", items:{...}})` would produce `diffs[""] = right` which could cause issues. However, this is unreachable in practice — Livewire component root keys are determined by PHP class property order, which is stable across requests.
</details>

<details>

| Question | Method | Result |
|----------|--------|--------|
| Does entangle reorder test regress? | Ran the test | PASS — no regression |
| Does PR test fail without fix? | Reverted fix, rebuilt, ran test | FAIL — valid regression test |
| Can numeric keys trigger false diffs? | Node.js: `diff({'2':'b','1':'a'}, {'1':'a','2':'b'})` | `{}` — V8 auto-sorts, no false positive |
| Key reorder + add? | Node.js: `diff({a:1,b:2}, {b:2,a:1,c:3})` | `{"c":3}` — reorder lost (known limitation) |
| Key reorder + remove? | Node.js: `diff({a:1,b:2,c:3}, {b:2,a:1})` | `{".c":"__rm__"}` — reorder lost (known limitation) |
| Is `deeplyEqual` order-sensitive? | Checked implementation | Yes — `JSON.stringify` preserves key order |
| Root-level reorder? | Analyzed code path | Produces `diffs[""]` but unreachable in practice |
| Does `mergeQueuedUpdates` interact badly? | Traced code | Pre-existing bug on 3.x makes interaction moot |
</details>

### Verdict

Good, minimal bug fix for a real problem (Filament sortable is a concrete downstream case). The 7-line change is well-scoped, follows existing patterns, introduces no regressions, and the test is valid. The `dist/` files have been removed and additional tests added. A companion PR for main branch's `diffRecursive()` is still needed. The key-reorder-plus-add/remove limitation is acceptable scope for this PR — the recursive path still handles value/key changes, just not the order aspect in that combined case.

---

PR #10004, opened 2026-02-14
[4.x] Fix diffRecursive() not detecting key order changes

people-sea:

Companion fix for #10001 (3.x). The same key-order issue exists in `diffRecursive()`.

Yes.

No.

Yes. Added a vitest case in `diff.spec.js`.

`diffRecursive()` checks key presence (`leftKeys.every(k => rightKeys.includes(k))`) but not key order. So `{a: 1, b: 2}` → `{b: 2, a: 1}` produces no diff.

`diff()` was fixed in #10001 but is `@deprecated`. Once it gets replaced by `diffAndConsolidate()`, the bug would come back without this fix.

```js

        if (path !== '') {
            diffs[path] = dataGet(rootRight, path)
            return { changed: true, consolidated: true }
        }
}
```

### DIFF
--- a/js/diff.spec.js
+++ b/js/diff.spec.js
@@ -285,4 +285,29 @@ describe('diffAndConsolidate', () => {
             data: { b: 2, a: 1 }
         })
     })
+
+    it('ignores key order changes at root level', () => {
+        expect(diffAndConsolidate(
+            { a: 1, b: 2 },
+            { b: 2, a: 1 }
+        )).toEqual({})
+    })
+
+    it('detects key order changes in deeply nested objects', () => {
+        expect(diffAndConsolidate(
+            { a: { b: { c: 1, d: 2 } } },
+            { a: { b: { d: 2, c: 1 } } }
+        )).toEqual({
+            'a.b': { d: 2, c: 1 }
+        })
+    })
+
+    it('detects key order changes with simultaneous value changes', () => {
+        expect(diffAndConsolidate(
+            { data: { a: 1, b: 2 } },
+            { data: { b: 3, a: 1 } }
+        )).toEqual({
+            data: { b: 3, a: 1 }
+        })
+    })
 })
--- a/js/diff.spec.js
+++ b/js/diff.spec.js
@@ -276,4 +276,13 @@ describe('diffAndConsolidate', () => {
             'form.text': 'Text'
         })
     })
+
+    it('detects key order changes in objects', () => {
+        expect(diffAndConsolidate(
+            { data: { a: 1, b: 2 } },
+            { data: { b: 2, a: 1 } }
+        )).toEqual({
+            data: { b: 2, a: 1 }
+        })
+    })
 })
--- a/js/utils.js
+++ b/js/utils.js
@@ -282,6 +282,14 @@ function diffRecursive(left, right, path, diffs, rootLeft, rootRight) {
         return { changed: true, consolidated: true }
     }
 
+    // Did the key order change?
+    if (isObject(left) && leftKeys.length === rightKeys.length && leftKeys.some((key, i) => key !== rightKeys[i])) {
+        if (path !== '') {
+            diffs[path] = dataGet(rootRight, path)
+            return { changed: true, consolidated: true }
+        }
+    }
+
     // Check if all keys are the same (no additions/removals)
     let keysMatch = leftKeys.every(k => rightKeys.includes(k))
 


calebporzio (approved):
## PR Review: #10004 — Fix diffRecursive() not detecting key order changes

**Type**: Bug fix
**Verdict**: Merge

### Summary

`diffRecursive()` (which powers `diffAndConsolidate()`) checks whether two objects have the same keys using membership comparison (`.includes()`), but never checks if those keys are in the same *order*. This means reordering an object's keys — like dragging rows in a sortable list — produces an empty diff, so the server never sees the change. This PR adds a positional key comparison that catches order-only changes.

### What's happening

- A user reorders items in a sortable/draggable UI (e.g., Filament's `KeyValue` field with `->reorderable()`)
- The server sends back the reordered data as a PHP associative array, which preserves insertion order
- On the client, the old state is `{a: 1, b: 2}` and the new state is `{b: 2, a: 1}` — same keys, same values, different order
- Livewire's diff algorithm only checked key *membership* (`['a','b'].every(k => ['b','a'].includes(k))` → `true`), so it saw no change
- The fix adds a positional comparison (`.some((key, i) => key !== rightKeys[i])`) that detects when keys are in a different order, even if values are identical
- When detected, the diff consolidates at the parent path level, sending the full reordered object to the server

- **Remove the keysMatch check entirely and let recursion handle it**: Doesn't work — recursion compares by key name (`left['a'] === right['a']`), so it can never detect pure reordering.
- **Fix in Alpine instead**: Alpine doesn't do data diffing; this is squarely Livewire's responsibility.
- **Do nothing**: Sortable/reorder features in Filament and custom Livewire components remain broken. No userland workaround exists since `diffRecursive()` is internal.

- Added 3 additional test cases: root-level key reorder (documenting expected no-op behavior), deeply nested reorder, and reorder-with-value-change.

### Test results

- **36/36 vitest tests pass** (including the new regression test and 3 additional cases)
- **Regression verified**: Removing the fix causes the test to fail with `{}` instead of the expected diff — confirming the bug
- **CI**: 21/22 checks pass. The single failure (`Browser 2 - L11.* - PHP 8.5`) is a `SupportQueryString` timeout unrelated to this JS-only change — all other Browser 2 configurations pass

**Why now**: PR #10001 just merged the same fix to the deprecated `diff()` on 3.x. The 4.x `diffRecursive()` has the identical bug. Filament users are hitting this in production ([filament#11128](https://github.com/filamentphp/filament/issues/11128), [livewire#7671](https://github.com/livewire/livewire/pull/7671)).

**Why here**: The diff lives in Livewire's own `js/utils.js`. Alpine.js has no data-level diffing — it uses Proxies for reactivity, which is downstream of Livewire's diff. No upstream library can fix this.

**Why us**: `diffRecursive()` is an internal implementation detail with no userland override mechanism. There is no workaround — the framework must fix this.

<details>
<summary>Critique</summary>

**Should this exist?** Yes — definitively. The diff system is the foundation of Livewire's change detection. A 6-line fix with a clear test that addresses a real, widely-reported bug class (sortable/reorderable UIs silently failing) is well-justified.

**Approach**: The fix is the laziest correct solution. It adds positional key comparison (`leftKeys.some((key, i) => key !== rightKeys[i])`) between the existing size-change check and the key-membership check. This mirrors the pattern established in PR #10001 for the deprecated `diff()`.

**Root-level gap**: The `if (path !== '')` guard means root-level key reorders are ignored. This is consistent with the adjacent `keysMatch` block and is not a practical concern — root-level keys are PHP property names (fixed declaration order), not user-reorderable data.

**Redundant length check**: `leftKeys.length === rightKeys.length` is technically always true at this point (the size-change check above would have returned early if lengths differed), but it's a harmless defensive guard.
</details>

<details>
<summary>Regression Analysis</summary>

**Blast radius**: MEDIUM — `diffRecursive()` runs on every Livewire request via `Component.getUpdates()`. However, the change only adds diff entries that were previously missed (never removes them), so existing behavior is strictly preserved.

**Query string risk**: LOW (verified). Query string tracking uses `JSON.stringify` comparison, not `diffAndConsolidate()`. The fix only affects server-bound request payloads.

**wire:dirty**: Correct behavior — `JSON.stringify` is already key-order-sensitive, so dirty detection was already working. The diff is now consistent with it.

**Performance**: The `.some()` check is O(n) per object comparison, but the existing `.every(k => rightKeys.includes(k))` is O(n²), so this is not a concern.
</details>

<details>
<summary>Security</summary>

No security concerns identified. This is a client-side-only change to a diff utility. No new data flows to the server that couldn't already be sent. No injection vectors. No prototype pollution risk.
</details>

<details>
<summary>API Harmony</summary>

No new public API. The fix follows the exact structural pattern of the two adjacent code blocks (size-change consolidation and key-membership consolidation). Comment style, return convention, and `dataGet()` usage all match. Test naming follows the established `diffAndConsolidate` describe block style.
</details>

### Verdict

Clean, minimal, well-motivated bug fix. The diff system must detect key-order changes for sortable/reorderable UIs to work — this is a correctness bug, not an edge case. The 8-line fix follows established patterns, has a valid regression test (plus 3 additional cases added during review), introduces no new API surface, and has no security implications. The single CI failure is an unrelated flaky test. Ship it.

---

calebporzio:
## PR Review: #10004 — Fix diffRecursive() not detecting key order changes

**Type**: Bug fix
**Verdict**: Merge

### Summary

`diffRecursive()` (which powers `diffAndConsolidate()`) checks whether two objects have the same keys using membership comparison (`.includes()`), but never checks if those keys are in the same *order*. This means reordering an object's keys — like dragging rows in a sortable list — produces an empty diff, so the server never sees the change. This PR adds a positional key comparison that catches order-only changes.

### What's happening

- A user reorders items in a sortable/draggable UI (e.g., Filament's `KeyValue` field with `->reorderable()`)
- The server sends back the reordered data as a PHP associative array, which preserves insertion order
- On the client, the old state is `{a: 1, b: 2}` and the new state is `{b: 2, a: 1}` — same keys, same values, different order
- Livewire's diff algorithm only checked key *membership* (`['a','b'].every(k => ['b','a'].includes(k))` → `true`), so it saw no change
- The fix adds a positional comparison (`.some((key, i) => key !== rightKeys[i])`) that detects when keys are in a different order, even if values are identical
- When detected, the diff consolidates at the parent path level, sending the full reordered object to the server

- **Remove the keysMatch check entirely and let recursion handle it**: Doesn't work — recursion compares by key name (`left['a'] === right['a']`), so it can never detect pure reordering.
- **Fix in Alpine instead**: Alpine doesn't do data diffing; this is squarely Livewire's responsibility.
- **Do nothing**: Sortable/reorder features in Filament and custom Livewire components remain broken. No userland workaround exists since `diffRecursive()` is internal.

- Added 3 additional test cases to `js/diff.spec.js`:
  - Root-level key reorder test (documents expected no-op behavior where root keys are ignored)
  - Deeply nested key reorder test (confirms consolidation at intermediate path)
  - Key reorder with simultaneous value change test (confirms correct handling of mixed changes)
- All 36/36 vitest tests pass
- Rebuilt JS assets (`npm run build`)
- Pushed test additions to contributor's branch

### Test results

- **36/36 vitest tests pass** (including the original regression test plus 3 new edge cases)
- **Regression verified**: Removing the fix causes the test to fail with `{}` instead of the expected diff — confirming the bug
- **CI**: 21/22 checks pass. The single failure (`Browser 2 - L11.* - PHP 8.5`) is a `SupportQueryString` timeout unrelated to this JS-only change — all other Browser 2 configurations pass

**Why now**: PR #10001 just merged the same fix to the deprecated `diff()` on 3.x. The 4.x `diffRecursive()` has the identical bug. Filament users are hitting this in production ([filament#11128](https://github.com/filamentphp/filament/issues/11128), [livewire#7671](https://github.com/livewire/livewire/pull/7671)).

**Why here**: The diff lives in Livewire's own `js/utils.js`. Alpine.js has no data-level diffing — it uses Proxies for reactivity, which is downstream of Livewire's diff. No upstream library can fix this.

**Why us**: `diffRecursive()` is an internal implementation detail with no userland override mechanism. There is no workaround — the framework must fix this.

**Prior art**: Phoenix LiveView, Hotwire/Turbo, and HTMX all diff at the HTML/DOM level, so they sidestep this class of problem entirely. Livewire is unique in doing client-side JSON state diffing, which means this bug class is Livewire-specific and must be fixed here.

<details>

**Should this exist?** Yes — definitively. The diff system is the foundation of Livewire's change detection. A 6-line fix with a clear test that addresses a real, widely-reported bug class (sortable/reorderable UIs silently failing) is well-justified.

**Approach**: The fix is the laziest correct solution. It adds positional key comparison (`leftKeys.some((key, i) => key !== rightKeys[i])`) between the existing size-change check and the key-membership check. This mirrors the pattern established in PR #10001 for the deprecated `diff()`.

**Root-level gap**: The `if (path !== '')` guard means root-level key reorders are ignored. This is consistent with the adjacent `keysMatch` block and is not a practical concern — root-level keys are PHP property names (fixed declaration order), not user-reorderable data.

**Redundant length check**: `leftKeys.length === rightKeys.length` is technically always true at this point (the size-change check above would have returned early if lengths differed), but it's a harmless defensive guard.
</details>

<details>

- **Phoenix LiveView**: Diffs rendered template output, not data structures. Key order is irrelevant.
- **Hotwire/Turbo**: HTML-over-the-wire. DOM morphing via Idiomorph. No JSON state diffing.
- **HTMX**: Same as Hotwire — HTML responses, no client-side state diffing.
- **Inertia.js**: Sends full page props; delegates reactivity to Vue/React.
- **Alpine.js**: Uses Proxies for reactivity, not state diffing. Not fixable upstream.

Livewire is unique in doing client-side JSON state diffing, making this bug class specific to Livewire.
</details>

<details>

**Blast radius**: MEDIUM — `diffRecursive()` runs on every Livewire request via `Component.getUpdates()`. However, the change only adds diff entries that were previously missed (never removes them), so existing behavior is strictly preserved.

**Query string risk**: LOW (verified). Query string tracking uses `JSON.stringify` comparison, not `diffAndConsolidate()`. The fix only affects server-bound request payloads.

**wire:dirty**: Correct behavior — `JSON.stringify` is already key-order-sensitive, so dirty detection was already working. The diff is now consistent with it.

**Performance**: The `.some()` check is O(n) per object comparison, but the existing `.every(k => rightKeys.includes(k))` is O(n²), so this is not a concern.
</details>

<details>

No security concerns identified. This is a client-side-only change to a diff utility. No new data flows to the server that couldn't already be sent. No injection vectors. No prototype pollution risk.
</details>

<details>

No new public API. The fix follows the exact structural pattern of the two adjacent code blocks (size-change consolidation and key-membership consolidation). Comment style, return convention, and `dataGet()` usage all match. Test naming follows the established `diffAndConsolidate` describe block style.
</details>

<details>

14 claims confirmed, 0 disproved, 1 partially true (the "redundant length check" claim was overstated — it's technically redundant but defensively reasonable).

Key verifications:
- Bug confirmed: `diffAndConsolidate({data:{a:1,b:2}}, {data:{b:2,a:1}})` → `{}` without fix
- Fix confirmed: Same call → `{data: {b:2, a:1}}` with fix
- Root-level gap confirmed and assessed as non-issue
- Query string risk downgraded from MEDIUM to LOW after tracing code paths
- CI failure confirmed unrelated (SupportQueryString timeout)
</details>

<details>

| Question | Method | Result |
|---|---|---|
| Does the deprecated `diff()` also have this bug? | Traced code in `js/utils.js:164-194` | Yes — `diff()` compares by key name, same bug |
| Is root-level key reorder silently ignored? | Ran `diffAndConsolidate({a:1,b:2}, {b:2,a:1})` | Confirmed — returns `{}` |
| Does query string use `diffAndConsolidate()`? | Traced `supportQueryString.js` | No — uses `JSON.stringify` comparison |
| Could PHP `json_encode` cause diff thrashing? | Checked PHP serialization behavior | No — PHP preserves insertion order stably |
| Is the `isObject` guard correct for arrays? | Checked `isObject([])` | Returns `false` — arrays excluded correctly |
</details>

### Verdict

Clean, minimal, well-motivated bug fix. The diff system must detect key-order changes for sortable/reorderable UIs to work — this is a correctness bug, not an edge case. The 8-line fix follows established patterns, has a valid regression test (plus 3 additional cases added during review), introduces no new API surface, and has no security implications. The single CI failure is an unrelated flaky test. Ship it.

---

calebporzio:
thanks!

PR #10018, opened 2026-02-16
[3.x] Fix reactive children re-rendering when props haven't changed

### LINKED ISSUE #18535: Dashboard select filter reloads all the widget data unnecessarily
### Package

filament/filament

### Package Version

v4.2.2

### Laravel Version

v12.38.1

### Livewire Version

v3.6.4

### PHP Version

8.4.14

### Problem description

I am filtering widget data by following this guide: https://filamentphp.com/docs/4.x/widgets/overview#filtering-widget-data. The widget data is reloaded before even selecting an option.

I have a Dashboard page with filters form. When opening the select dropdown, it reloads all the widgets available on page and runs all the queries even before showing the select drop down. This delays the select drop down options until all the widget data is reloaded. Once all the queries are run and the request finishes, thats when it populates the options for select field.

I have also tried [Filtering widget data using an action modal](https://filamentphp.com/docs/4.x/widgets/overview#filtering-widget-data-using-an-action-modal)
But same problem with Modal as well, even clicking the Modal action button reloads the data and runs all the queries before even opening the Modal.

Discord Discussion URL: https://discord.com/channels/883083792112300104/1440212167616892999

### Expected behavior

Clicking on select field should not reload the widget data unless an option selected.

### Steps to reproduce

1) Clone the repo
2) Follow the installation guide in README
3) Open Filament admin panel at /admin
4) Open Up DebugBar. Click on Group select field and you will see that it creates a new Livewire request which shows the query for getting groups but also a query for Users count.

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/mansoorkhan96/filament-widget-issue

### Relevant log output

```shell

```

people-sea:

Related to https://github.com/filamentphp/filament/issues/18535.

Yes.

No.

Yes.

When a parent component commits, all child components with `#[Reactive]` props are pooled into the request and re-rendered, even when their prop values haven't changed. This causes unnecessary work in `render()`, like expensive database queries, on every parent interaction.

This PR adds a hash comparison in `BaseReactive::hydrate()` to detect whether reactive prop values actually changed, and calls `skipRender()` when they haven't — as long as the component has no pending updates or method calls of its own.

The hash comparison approach is consistent with how v4 (`main` branch) already handles this in `BaseReactive::hydrate()`, but v4 currently only skips the `setValue()` call — it doesn't skip the `render()`. This PR extends that idea to also call `skipRender()` when no reactive props have changed.

Before:

// BaseReactive::hydrate() — always sets the value
```php
$this->setValue($updatedValue);
```

After:
```php
$currentHash = crc32(json_encode($this->getValue()));
$updatedHash = crc32(json_encode($updatedValue));

if ($currentHash !== $updatedHash) {
    $this->setValue($updatedValue);

    store($this->component)->set('reactivePropsChanged', true);
}
```

### DIFF
--- a/src/Features/SupportReactiveProps/BaseReactive.php
+++ b/src/Features/SupportReactiveProps/BaseReactive.php
@@ -28,7 +28,14 @@ class BaseReactive extends LivewireAttribute
                 $this->component->getId(), $this->getName()
             );
 
-            $this->setValue($updatedValue);
+            $currentHash = crc32(json_encode($this->getValue()));
+            $updatedHash = crc32(json_encode($updatedValue));
+
+            if ($currentHash !== $updatedHash) {
+                $this->setValue($updatedValue);
+
+                store($this->component)->set('reactivePropsChanged', true);
+            }
         }
 
         $this->originalValueHash = crc32(json_encode($this->getValue()));
--- a/src/Features/SupportReactiveProps/BrowserTest.php
+++ b/src/Features/SupportReactiveProps/BrowserTest.php
@@ -4,6 +4,7 @@ namespace Livewire\Features\SupportReactiveProps;
 
 use Livewire\Livewire;
 use Livewire\Component;
+use Livewire\Attributes\Renderless;
 
 class BrowserTest extends \Tests\BrowserTestCase
 {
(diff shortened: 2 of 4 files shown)


calebporzio:
doing some digging on this. got a different PR to solve this from a different angle that will result in lots more perf improvements: https://github.com/livewire/livewire/pull/10195 closing - thanks

PR #10019, opened 2026-02-17
[4.x] Reactive props during boot hooks

danharrin:

This is a bug fix from PR #9643.

Yes

No

Yes

PR #9643 added `updating*` / `updated*` lifecycle hook support for `#[Reactive]` props. To make that work, `SupportLifecycleHooks` had to be initialized before firing `trigger('update')`. The solution at the time was to defer the entire update, including setting the value, to an `after('hydrate')` callback.

`SupportLifecycleHooks::hydrate()` runs `boot()`, `hydrate()`, and `booted()` during the regular `on('hydrate')` phase, before any `after('hydrate')` callbacks fire. That means all of those hooks run with stale prop values. The new value does not actually land on the component until after all lifecycle hooks have already executed.

Before #9643 (beta.6 and earlier), `BaseReactive::hydrate()` called `$this->setValue()` immediately. Since `SupportAttributes` runs before `SupportLifecycleHooks` in the hook registration order, the value was already set by the time `boot()`, `hydrate()`, and `booted()` ran, so everything worked as expected.

This restores the pre-#9643 behavior by setting the value immediately again, but still queues `trigger('update')` for `after('hydrate')` so that `updating*` / `updated*` hooks continue to fire correctly.

During the deferred `trigger('update')`, the old value is temporarily swapped back in so `updating*` hooks still see the previous state via `$this->property`, then the new value is restored before `updated*` hooks run.

This affects anyone reading `#[Reactive]` prop values inside `boot()`, `hydrate()`, or `booted()`.

### DIFF
--- a/src/Features/SupportReactiveProps/BaseReactive.php
+++ b/src/Features/SupportReactiveProps/BaseReactive.php
@@ -37,22 +37,21 @@ class BaseReactive extends LivewireAttribute
             $updatedHash = crc32(json_encode($updatedValue));
 
             if ($currentHash !== $updatedHash) {
-                // Queue the update to be processed after all hooks are hydrated
-                // This ensures SupportLifecycleHooks is initialized before we trigger updates
-                // Don't set the value yet - it will be set when processing the queue
-                // so that updating* hooks see the old value
+                // Set value immediately so lifecycle hooks (boot/hydrate/booted) see updated data
+                $this->setValue($updatedValue);
+                $this->originalValueHash = crc32(json_encode($updatedValue));
+
+                // Queue the update trigger for after hydrate so updating*/updated* hooks fire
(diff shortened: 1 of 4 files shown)


calebporzio:
Pushed a commit adding unit tests (`src/Features/SupportReactiveProps/UnitTest.php`) that cover:

1. **`boot()`/`hydrate()`/`booted()` see fresh reactive prop values** — mounts child with `count=0`, simulates parent passing `count=5`, asserts all three hooks see `5` not `0`
2. **`updating*` sees old value, `updated*` sees new value** — verifies the value-swap contract is preserved for reactive props

The fix itself is clean — reviewed and matches independently-derived analysis. Merging main in to bring the branch up to date.

calebporzio:
thanks Dan!

PR #10021, opened 2026-02-17
[4.x] Failing test for nested `<script>` extracted as component script module

joshhanley:
Livewire's single file component compiler is wrongly identifying a script tag in the view portion of the template as a script tag intended for the script portion of the template.


calebporzio:
Thanks for the great failing test and bug report, @joshhanley! This clearly identified the issue.

I've opened #10023 which includes your test + the fix (removing `\s*` from the extraction regex so only root-level tags at column 0 are matched). The same fix was also applied to `<style>` and `<style global>` extraction which had the same vulnerability.

You're credited as co-author on the commit. Closing this in favor of the fix PR.

PR #10030, opened 2026-02-19
[4.x] Fix `wire:model` not updating custom elements when nested array data changes

### LINKED ISSUE #2414: Charts do not update when data is updated

joshhanley:
# The Scenario

When using `wire:model` to bind an array property to a custom element (such as Flux's `<flux:chart>`), the element doesn't update when nested values within the array change server-side.

The `:value` binding works correctly, but `wire:model` does not reflect the changes:

```php
<?php

use Livewire\Component;

new class extends Component {
    public array $data = [
        ['label' => 'Mon', 'revenue' => 1127],
        ['label' => 'Tue', 'revenue' => 249],
    ];

    public function change()
    {
        $this->data[0]['revenue'] = 0;
    }
}; ?>

<div>
    <flux:button wire:click="change">Change Data</flux:button>

    {{-- This does NOT update --}}
    <flux:chart wire:model="data">
        ...
    </flux:chart>

    {{-- This updates correctly --}}
    <flux:chart :value="$data">
        ...
    </flux:chart>
</div>
```

# The Problem

Since [#9897](https://github.com/livewire/livewire/pull/9897), `mergeNewSnapshot` applies data changes surgically at granular dot-notated paths (e.g., `data.0.revenue`) rather than replacing entire top-level properties. This preserves client-side ephemeral state during server updates.

However, Alpine's `bindInputValue` utility short-circuits with a reference equality check:

```js
if (el.value === value) return
```

When `wire:model`'s x-model getter returns the reactive proxy, and the custom element's `value` getter also returns the same reactive proxy (stored from a previous set), both sides are the same object reference — even though nested values within it have changed. The equality check passes, `el.value` is never re-set, and the custom element never sees the update.

# The Solution

Return a deep clone for non-primitive values from the `wire:model` x-model getter. This ensures Alpine's equality check sees a new reference each time nested data changes, allowing the update to propagate to `el.value`.

```js
get() {
    let value = dataGet(component.$wire, expression)

    if (typeof value === 'object' && value !== null) {
        return JSON.parse(JSON.stringify(value))
    }

    return value
},
```

Fixes https://github.com/livewire/flux/issues/2414


calebporzio:
This makes me nervous. JSON.parse/stringify is expensive, AND this feels patchy AND this feels like something that will have all sorts of ripple effects.

I request that we understand the landscape of the problem better and explore all solutions, even upstream ones. Maybe there's something in the reactivity engine that can help us here?

calebporzio:
bumping

joshhanley:
@calebporzio yeah turns out an upstream fix in Alpine is much cleaner. Have submitted an Alpine PR with a fix https://github.com/alpinejs/alpine/pull/4833. Have reverted my fix and marked this as draft for now (as this is just a failing test PR now) and will review it once fix is in Alpine.

calebporzio:
@joshhanley - merged the alpine PR. can we close this now?

joshhanley:
@calebporzio yep I just pulled the Alpine update into this branch locally and the test passes. So this will be fixed once we tag a new Alpine and Livewire releases. Will close this now.

PR #10032, opened 2026-02-21
[3.x] Add Laravel 13 support

joshhanley:
# The Scenario

Laravel 13 is releasing next month and Livewire 3.x doesn't yet declare compatibility with it.

# The Solution

- Added `|^13.0` to `illuminate/*` package constraints
- Added `|^8.0` to `symfony/console` and `symfony/http-kernel` (Laravel 13 uses Symfony 8)
- Added `|^13.0` to `laravel/framework`, `|^11.0` to `orchestra/testbench` and `orchestra/testbench-dusk`
- Added `|^12.5` to `phpunit/phpunit` (required by `orchestra/testbench-core` 11.x)
- Added `13.*` to the CI matrices with PHP 8.1/8.2 exclusions (Laravel 13 requires PHP 8.3+)
- Added `void` return type to `configure()` on `TouchCommand`, `MvCommand`, `RmCommand`, and `CpCommand` for Symfony 8 compatibility

PR #10033, opened 2026-02-21
[3.x] Fix session expiry detection for Laravel 13's `Sec-Fetch-Site` origin verification

joshhanley:
# The Scenario

Livewire's "page expired" dialog stops working on Laravel 13.

When a user's session is regenerated (e.g., they log out in another tab, or the session times out), the next Livewire request should trigger a 419 response and show a "This page has expired" dialog prompting the user to refresh. On Laravel 13, this dialog never appears — the request goes through as if nothing happened.

The two browser tests in `SupportErrorResponses/BrowserTest.php` reproduce this: they regenerate the session, trigger a Livewire update, and expect the page expired dialog. Both time out waiting for a dialog that never comes.

# The Problem

Laravel 13 ([laravel/framework#58400](https://github.com/laravel/framework/pull/58400)) introduced a new CSRF protection approach inspired by [Rails](https://github.com/rails/rails/pull/56350). The `VerifyCsrfToken` middleware was replaced by `PreventRequestForgery`, which checks the browser's `Sec-Fetch-Site` header **before** checking CSRF tokens:

```php
public function handle($request, Closure $next)
{
    if (
        $this->isReading($request) ||
        $this->runningUnitTests() ||
        $this->inExceptArray($request) ||
        $this->hasValidOrigin($request) ||   // NEW — checked before tokensMatch()
        $this->tokensMatch($request)
    ) {
        return tap($next($request), ...);
    }

    throw new TokenMismatchException('CSRF token mismatch.');
}
```

The `hasValidOrigin()` method trusts any request with `Sec-Fetch-Site: same-origin`:

```php
protected function hasValidOrigin($request)
{
    $secFetchSite = $request->header('Sec-Fetch-Site');

    if ($secFetchSite === 'same-origin') {
        return true;
    }
    // ...
}
```

Modern browsers automatically set `Sec-Fetch-Site: same-origin` on all same-origin `fetch()` requests — which is exactly what Livewire uses for its update endpoint. This means the middleware **always** trusts Livewire requests based on their origin and **never** checks the CSRF token. So when the session is regenerated and the CSRF token changes, the stale token sent by the browser is never compared, the middleware never returns 419, and Livewire's page expired dialog is never triggered.

This is the intended behaviour from Laravel's perspective — origin verification is a more reliable CSRF protection mechanism than tokens (the header can't be spoofed by JavaScript, and it eliminates false positives from stale tokens and cached pages). However, Livewire was relying on the CSRF token mismatch as a **session freshness indicator** — a side effect of the old CSRF protection that no longer occurs.

# The Solution

This PR adds an independent CSRF token check in Livewire's `handleUpdate()` method, before processing the update request:

```php
if (request()->hasSession() && ! (app()->runningInConsole() && app()->runningUnitTests())) {
    $sessionToken = request()->session()->token();
    $requestToken = request()->input('_token') ?? '';

    if (! is_string($sessionToken) || ! is_string($requestToken) || ! hash_equals($sessionToken, $requestToken)) {
        abort(419);
    }
}
```

Livewire's JS already sends the CSRF token (`_token`) with every update request (in `js/request/index.js`), and the token is embedded in the page via the `data-csrf` attribute on the Livewire script tag. When the session is regenerated, the token in the session changes but the page still has the old one — so the comparison fails and Livewire returns 419, restoring the page expired dialog.

The guards match the CSRF middleware's own behaviour:
- `request()->hasSession()` — skips the check when there's no session (e.g., `Livewire::test()` unit tests that call `handleUpdate()` directly without the `web` middleware)
- `! (app()->runningInConsole() && app()->runningUnitTests())` — skips the check during PHPUnit unit tests, matching the CSRF middleware's `runningUnitTests()` check. This is needed because unit tests that POST to `/livewire/update` don't send a `_token`. The check still runs in Dusk browser tests because the server process isn't a console process (`runningInConsole()` returns `false`).

On older Laravel versions this check is harmless but redundant — the CSRF middleware would have already rejected the request with a 419 before `handleUpdate()` is reached.

## What Happens in Standard Laravel Requests?

This isn't just a Livewire issue — it's a general behaviour change in Laravel 13. Consider a standard form submission where the user's session has been regenerated (e.g., they logged out in another tab):

**Before (Laravel 12 and earlier):**
- User submits form with stale CSRF token → middleware compares tokens → mismatch → **419 "Page Expired"**

**After (Laravel 13):**
- User submits form → browser sends `Sec-Fetch-Site: same-origin` → middleware trusts origin → **skips token check → request goes through**

From there, it depends on the route:
- **Route has `auth` middleware:** the user is no longer authenticated, so they get **redirected to login** — arguably a better UX than a cryptic "Page Expired" error
- **Route has no `auth` middleware:** the request processes with the new (guest) session as if nothing happened

So Laravel 13 no longer catches stale sessions at the CSRF layer at all. It relies on auth/authorisation middleware to handle the "user is no longer who they were" case.

## Why Does Livewire Need This When Standard Laravel Doesn't?

For **authenticated routes**, standard Laravel handles session loss naturally — the auth middleware redirects to login, the browser follows the redirect, and the user understands what happened. Livewire actually handles this too: `js/request/index.js` checks `response.redirected` and navigates the browser to the redirect URL. So for auth-protected Livewire components, the auth redirect would work without the page expired dialog.

But **Livewire routes can be guest routes**. A Livewire component on a public page has no auth middleware to catch the session change. Without the page expired dialog, the request would silently process with a new/empty session — no redirect, no warning, nothing.

Standard Laravel has the same gap on guest routes — a form POST to a guest route after session loss would also just process silently. But for a one-off form submission, that's usually harmless. Livewire components are **stateful and long-lived** on the page, so silently switching to a different session mid-interaction is more likely to cause unexpected behaviour (e.g., component state that depends on session data, actions that assume a particular session context, etc.).

So the page expired dialog isn't fighting the framework's direction — it's covering a gap that the framework also has, but that matters more in Livewire's context because of its persistent, stateful nature.

### We're checking the token that Laravel decided not to check

This is the core tension. Laravel 13 deliberately moved away from CSRF token validation for same-origin requests, and this fix re-introduces it for Livewire's endpoint. We're not using it for CSRF protection (origin verification handles that), but as a **session freshness indicator** — a different purpose that happens to use the same value.

### Is there a better approach?

Alternative approaches were considered:

1. **Livewire-specific session token** — store a separate `livewire_token` in the session when the page renders, embed it in the page, compare on each request. This is functionally identical to checking the CSRF token but with more moving parts (and JS changes).

2. **Session ID comparison** — hash the session ID when the page renders, embed it, compare on update requests. This checks the actual thing that changed (the session ID) rather than a proxy. But it also requires JS changes and is semantically the same operation.

3. **Opt the Livewire route into token checking** — there's currently no way in Laravel 13 to tell `PreventRequestForgery` to always check tokens for specific routes. If this existed, it would be the cleanest solution.

4. **Extend Laravel's CSRF middleware** — create a Livewire middleware that extends `PreventRequestForgery` and overrides `hasValidOrigin()` to always return `false`, forcing the token check to always run. This inherits all of Laravel's token checking logic (`tokensMatch()`, `getTokenFromRequest()` which checks `_token`, `X-CSRF-TOKEN`, and `X-XSRF-TOKEN`), the `runningUnitTests()` guard, `isReading()` skip, and `TokenMismatchException` on failure — all for free. The middleware would be applied to the Livewire update route alongside `web`. This is arguably the cleanest approach: one method override, proper separation of concerns, and no custom token comparison logic. The trade-off is that it essentially reverses what Laravel has done, but scoped to Livewire routes only.

The CSRF token approach was chosen because:
- The token is already embedded in the page and already sent with every Livewire request
- No JS changes required
- It changes reliably whenever the session is regenerated
- The comparison logic is well-understood and battle-tested

### Should this be raised with Laravel?

This feels like a gap that other packages relying on 419 responses for session expiry detection will hit. Possible framework-level solutions:
- A route-level option to force token checking regardless of origin verification
- A dedicated session freshness mechanism (separate from CSRF)
- A middleware hook that fires when origin verification bypasses token checking

### Scope

This fix only applies to Livewire v3 running on Laravel 13. Once merged the changes will be up-merged into Livewire v4 (main branch).

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -115,6 +115,24 @@ class HandleRequests extends Mechanism
 
     function handleUpdate()
     {
+        // Laravel 13 (laravel/framework#58400) introduced Sec-Fetch-Site origin
+        // verification in its CSRF middleware. When the browser sends
+        // "Sec-Fetch-Site: same-origin" (which it does for same-origin fetch
+        // requests), the middleware trusts the origin and skips CSRF token
+        // validation entirely. This means that after a session is regenerated,
+        // the stale CSRF token sent by the browser is never checked, so the
+        // middleware never returns a 419 — and Livewire's page expired dialog
+        // is never triggered. To ensure session expiry is still detected, we
+        // verify the CSRF token ourselves before processing the update.
+        if (request()->hasSession() && ! (app()->runningInConsole() && app()->runningUnitTests())) {
+            $sessionToken = request()->session()->token();
+            $requestToken = request()->input('_token') ?? '';
+
+            if (! is_string($sessionToken) || ! is_string($requestToken) || ! hash_equals($sessionToken, $requestToken)) {
+                abort(419);
+            }
+        }
+
         $requestPayload = request(key: 'components', default: []);
 
         $finish = trigger('request', $requestPayload);


calebporzio:
i think we should be fine to just do what laravel 13 does. auth'd people would be bounced, non auth people will still be fine.

calebporzio:
i say we just bypass these failing tests for laravel 13+ -> but maybe make sure we're not opening some door for csrf here, but i don't think we'd be opening a door that laravel hasn't already in this case.

joshhanley:
Yep, can do. I've updated it on the main Laravel 13 PR #10032 so going to close this one.

PR #10035, opened 2026-02-22
[4.x] Add configurable route prefix for all Livewire endpoints

kefivitch:
## Situation

Applications using Livewire need custom route prefixes for their Livewire endpoints for various reasons:

- Migrating legacy applications that need backward-compatible routes
  - Multi-tenant applications requiring tenant-specific route prefixes (e.g., `/{tenant}/livewire/...`)
  - API versioning schemes (e.g., `/api/v1/livewire/...`, `/api/v2/livewire/...`)
  - Compliance or organizational requirements for specific URL structures
## Problem

Livewire's routes are hardcoded to use `/livewire-{hash}/` as the base prefix with no configuration option to customize this. All 7 endpoint types (update, upload, preview, JS/CSS modules, frontend assets) use `EndpointResolver::prefix()` which returns a fixed path.

Developers currently have no way to prepend a custom prefix without:
- Modifying vendor code (breaks on updates)
  - Creating complex service provider workarounds that duplicate route registration logic
  - Manually registering custom routes and disabling Livewire's defaults
## Solution

Add a `route_prefix` configuration option to `config/livewire.php` that prepends a custom prefix to all Livewire routes when set.

**Implementation:**

Modified `EndpointResolver::prefix()` to check for and prepend the custom prefix:

```php
public static function prefix(): string
{
    $hash = substr(hash('sha256', config('app.key') . 'livewire-endpoint'), 0, 8);
    $livewirePrefix = '/livewire-' . $hash;
    
    $customPrefix = config('livewire.route_prefix');
    
    if (! empty($customPrefix)) {
        return '/' . trim($customPrefix, '/') . $livewirePrefix;
    }
    
    return $livewirePrefix;
}
```

**Why this approach:**

- **Centralized:** Single method change affects all 7 endpoint types uniformly
  - **Automatic:** All route methods (`updatePath()`, `scriptPath()`, etc.) inherit the prefix
  - **JavaScript compatible:** Frontend receives correct prefixed URLs from registered routes
  - **Route cache safe:** Works with `php artisan route:cache`
  - **Backward compatible:** Empty/null prefix maintains existing behavior
   - **Minimal:** ~14 lines of code vs 150+ for service provider approach
**Example usage:**

```php
// config/livewire.php
'route_prefix' => 'legacy',

// Results in:
// /legacy/livewire-{hash}/update
// /legacy/livewire-{hash}/upload-file
// /legacy/livewire-{hash}/livewire.js

// Nested paths work too:
'route_prefix' => 'api/v1',
// /api/v1/livewire-{hash}/update
```

**Test coverage:**

- ✅ 7 unit tests (custom prefix, nested paths, edge cases)
  - ✅ 10 integration tests (all endpoint types, consistency)
  - ✅ 7 browser tests (real components, wire:model, multiple components)
  - **Total: 24 new tests, all passing**
**Breaking changes:** None. Purely additive feature with default behavior unchanged.


calebporzio:
passing on this for now. historically endpoint related PRs create future headaches and too many customization paths confuse thigns. not opposed to a future with this kinda of thing in it, but would need more demand and exploration, etc.. thanks

PR #10036, opened 2026-02-23
[4.x] Enforce payload max size using actual request size

utsavsomaiya:
## Problem
Some oversized Livewire update requests could slip past the configured payload size guard in edge request scenarios.

## Solution
The payload limit is now enforced using the true request size so requests above the configured threshold are consistently blocked.

## Outcome
Payload protection is now reliable across normal and edge request patterns, while keeping existing application behavior intact.

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -145,10 +145,12 @@ class HandleRequests extends Mechanism
         $maxSize = config('livewire.payload.max_size');
 
         if ($maxSize !== null) {
-            $contentLength = request()->header('Content-Length', 0);
+            $actualPayloadSize = strlen(request()->getContent() ?? '');
+            $declaredPayloadSize = (int) request()->header('Content-Length', 0);
+            $payloadSize = max($actualPayloadSize, $declaredPayloadSize);
 
-            if ($contentLength > $maxSize) {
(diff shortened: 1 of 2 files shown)


nuernbergerA:
What about mb?

utsavsomaiya:
Good point. payload.max_size is still byte-based, so MB values are supported the same way as before (for example: 5 * 1024 * 1024).

This PR only changes how we measure request size (actual body size instead of trusting Content-Length), not the configured unit.

gp-lnuff:
The code seems to assume there can be a mismatch between the size declared in the header and the actual content size.
Is that allowed? If they're allowed to differ I'm a little confused about the utility of the header.
I feel like there would better be a check to ensure the declared payload size and the actual paylode size match.

utsavsomaiya:
Great question. In normal HTTP requests, Content-Length should describe the body size, but at the app layer we cannot fully trust it (or rely on it being present). Chunked requests may omit it, and proxies/middleware can affect what PHP ultimately sees.

This guard is focused on safely enforcing livewire.payload.max_size, not protocol-level framing validation. Using the measured request body bytes avoids bypasses when the header is missing or under-reported. Taking max(actual, declared) keeps the check conservative so spoofed small headers cannot slip through.

A strict equality check between declared and measured bytes can reject valid requests in some environments, while not improving the bypass protection this PR targets.

PR #10037, opened 2026-02-23
[4.x] Keep query string model values in sync on back/forward

utsavsomaiya:
## Problem
When navigating browser history, some query-string-bound fields could display stale UI values even though the URL/state had already changed.

## Solution
History restores now consistently keep model-bound field values in sync with the restored state across field types.

## Outcome
Back/forward navigation now behaves predictably for query-string state and avoids UI mismatches.

### DIFF
--- a/js/features/supportQueryString.js
+++ b/js/features/supportQueryString.js
@@ -41,10 +41,8 @@ on('effect', ({ component, effects, cleanup }) => {
             let forgetPopHandler = pop(async newValue => {
                 await component.$wire.set(name, newValue)
 
-                // @todo: this is the absolute worst thing ever I'm so sorry this needs to be refactored stat:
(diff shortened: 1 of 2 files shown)


joshhanley:
@utsavsomaiya thanks for the PR! Can you please revert the flaky test changes, as they aren't related to this PR? I've pushed some changes to main to hopefully try and fix that. Also can you fix the merge conflicts? Thanks!

P.s. if there are any flaky tests on only a couple of test runners, I can just re-run them, so no need to push again.

utsavsomaiya:
Thanks for the guidance. I’ve updated this branch by rebasing onto the latest `main`, dropping the unrelated flaky-test commits, and keeping only the query string popstate model sync fix. The merge conflict is now resolved.

PR #10038, opened 2026-02-23
[4.x] Prevent navigate crashes when head includes link tags without rel

utsavsomaiya:
## Summary

This change prevents navigate from failing when the destination page includes link tags in the head without a rel attribute.

Navigation now completes normally in this case, and a regression test keeps this behavior stable.

### DIFF
--- a/js/plugins/navigate/page.js
+++ b/js/plugins/navigate/page.js
@@ -199,7 +199,7 @@ function extractUriAndQueryString(el) {
 }
 
 function isAsset(el) {
-    return (el.tagName.toLowerCase() === 'link' && el.getAttribute('rel').toLowerCase() === 'stylesheet')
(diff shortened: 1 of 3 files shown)


joshhanley:
@utsavsomaiya thanks for the PR! Can you please revert the flaky test changes, as they aren't related to this PR? I've pushed some changes to main to hopefully try and fix that. And also remove any committed `dist/` files, as we build them in CI. Thanks!

P.s. if there are any flaky tests on only a couple of test runners, I can just re-run them, so no need to push again.

utsavsomaiya:
Thanks for the review. I’ve updated this branch by rebasing onto the latest `main`, removing the unrelated flaky-test changes, and keeping only the navigate fix (with no committed `dist/` files). The merge conflict is also resolved.

PR #10042, opened 2026-02-23
[4.x] Decode non-array query keys in history URL parsing

utsavsomaiya:
## Summary

This change ensures history URL parsing decodes non-array query parameter keys before tracking and updates.

It prevents missed initial values and duplicate query parameters when a tracked key contains encoded characters (for example spaces). A regression test is included.


joshhanley:
@utsavsomaiya thanks for the PR! Can you please remove any committed `dist/` files, as we build them in CI. Thanks!

utsavsomaiya:
Thanks for the review. I’ve updated this branch on top of the latest `main` and removed all committed `dist/` files. The PR now contains only the history query-key decoding source/test changes.

PR #10046, opened 2026-02-24
[4.x] Improve event testing assertions DX

utsavsomaiya:
## Summary
This improves Livewire test ergonomics for dispatched events.

## What this solves
- Event callback assertions now evaluate all matching dispatched events, avoiding false negatives when the same event is dispatched multiple times.
- Adds an explicit event count assertion so tests can clearly verify how many times an event was dispatched.

### DIFF
--- a/src/Features/SupportEvents/TestsEvents.php
+++ b/src/Features/SupportEvents/TestsEvents.php
@@ -44,6 +44,21 @@ trait TestsEvents
         return $this;
     }
 
+    public function assertDispatchedTimes($event, $times = 1)
+    {
+        $count = collect(data_get($this->effects, 'dispatches'))
+            ->where('name', $event)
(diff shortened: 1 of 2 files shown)


calebporzio:
Sorry @utsavsomaiya

I hate to do this to you, but I think I'm just going to close all these PRs

They were clearly vibe coded, which isn't necessarily disqualifier, but given that the descriptions are so minimal and not demonstrative of the contribution, I'm just going to close them.

Please, in the future be more considered about contributions. Choose one or two things you think are high value and make sure the description completely lays out the proposed APIs, provides reasoning, community demand, the problem it solves, everything.

A lot of these genuinely try to solve real problems or pain points, so I hope we can find solutions for them at some point.

Thanks mate 👌

PR #10047, opened 2026-02-24
[4.x] Improve checksum mismatch debugging experience

utsavsomaiya:
### Why
Checksum mismatch issues are painful to troubleshoot because the failure gives very little context about what changed between requests.

### Solution
In local/debug workflows, checksum failures now provide actionable context by showing the first mismatched payload path and the expected vs received value.
Production behavior is unchanged.

### Developer impact
This removes guesswork during debugging and makes it much faster to identify the exact state difference causing the failure.

### Safety
This is scoped to developer debugging behavior and does not change production error handling.

### DIFF
--- a/src/Mechanisms/HandleComponents/Checksum.php
+++ b/src/Mechanisms/HandleComponents/Checksum.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Mechanisms\HandleComponents;
 
+use Illuminate\Support\Facades\Cache;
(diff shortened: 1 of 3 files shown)


joshhanley:
@utsavsomaiya thanks for the PR! This has conflicts, can you please fix them? Thanks!

utsavsomaiya:
Thanks - I’ve resolved the conflicts by rebasing this branch onto the latest `main` and replaying the checksum debugging commit cleanly.

PR #10048, opened 2026-02-24
[4.x] Improve malformed update request debugging

utsavsomaiya:
## Why
When update requests are malformed, developers currently see a generic failure that is hard to diagnose during local development.

## Solution
In debug mode, malformed update requests now fail with clear, actionable error messages that identify what part of the payload is invalid.

## Impact
- Faster debugging of broken requests
- Less trial-and-error when integrating complex components
- No behavior change for production environments

## Validation
- HandleRequests unit coverage passes locally


calebporzio:
like i can't merge this without knowing what the actual error message is and what it looks liek to receive it and all that. right now i would have to dig into the code to assess. closing

PR #10049, opened 2026-02-24
[4.x] Add #[Sync] for consistent PHP/JavaScript property shaping

utsavsomaiya:
## Summary
This introduces an opt-in `#[Sync]` attribute that keeps Livewire property values aligned across PHP and JavaScript boundaries.

Instead of repeatedly normalizing payloads in each component, teams define one sync strategy and reuse it everywhere.

## What this solves for developers
- Preserve domain-first state (Money, DateRange, FilterQuery) in backend code while still having predictable browser payloads.
- Remove repetitive parse/format logic from components and Alpine blocks.
- Reduce drift bugs caused by ad-hoc conversions during request/response cycles.

## Real-world scenarios
- Checkout flows: totals, tax, and discount values stay consistent across server actions and client interactions.
- Booking forms: date-range and timezone values remain stable when users edit and resubmit.
- Admin analytics filters: complex nested filter state stays reliable without per-screen conversion helpers.

## Why this is safe
- Fully opt-in. Existing components are unchanged unless Sync is used.
- Supports both simple built-in strategies and custom strategies for rich domain values.
- Deep partial updates are intentionally guarded for custom strategies to prevent inconsistent state.

## Validation
- PHP unit coverage for built-in strategies, trait properties, custom strategy roundtrip, and guard behavior.
- JavaScript tests for built-in and custom client transforms.


joshhanley:
@utsavsomaiya thanks for the PR! Can you please remove any committed `dist/` files, as we build them in CI. Thanks!

utsavsomaiya:
Thanks for the review. I’ve updated the branch: all committed `dist/` changes have been removed from this PR, and I also merged the latest upstream `main` to resolve conflicts.

calebporzio:
these need better PR descriptions. i keep having to dig into the code and docs to learn full context and full api. closing

PR #10051, opened 2026-02-24
Throw exception on snapshot json_encode failure instead of silent corruption

sathia-musso:

**1. Is this something that is wanted/needed? Did you create a discussion about it first?**
Yes — https://github.com/livewire/livewire/discussions/10050

**2. Did you create a branch for your fix/feature? (Main branch PR's will be closed)**
Yes, branch: `fix/snapshot-json-encode-silent-failure`

**3. Does it contain multiple, unrelated changes? Please separate the PRs out.**
No. All changes address a single issue: silent `json_encode` failure during snapshot serialization.

**4. Does it include tests? (Required)**
Yes. A unit test in `src/Mechanisms/HandleRequests/UnitTest.php` that triggers the failure with non-UTF-8 data and verifies the exception is thrown.

**5. Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.**

When a Livewire component has public properties containing non-UTF-8 strings, `json_encode()` silently returns `false` and the framework doesn't check for it. This happens in two code paths:

1. **Update path** (`HandleRequests.php`): The response contains `"snapshot": false` instead of a valid JSON string
2. **Initial render path** (`Utils::escapeStringForHtml`): The snapshot is embedded as `wire:snapshot=""` (empty)

Both result in the same cryptic client-side error: `TypeError: can't access property "id", f.memo is undefined` — with zero server-side indication of the problem. No exception, no log, nothing.

This is common with legacy MySQL databases using `charset: latin1` (which is actually Windows-1252). Characters like curly quotes, em-dashes, and other typographic characters in the 0x80-0x9F byte range silently break the entire component. The bug appears "random" because it only triggers when a specific record with these characters is loaded.

**Reproduction:**

```php
class BrokenComponent extends \Livewire\Component
{
    public array $items = [];

    public function loadItems()
    {
        // 0x92 = right single quotation mark in Windows-1252, invalid in UTF-8
        $this->items = [
            ['name' => "Test\x92s Item"],
        ];
    }

    public function render()
    {
        return view('livewire.broken-component');
    }
}
```

**Fix:**

- Added `SnapshotEncodingException` with a clear message (component name, JSON error, UTF-8 suggestion)
- Added `json_encode()` return value check in `HandleRequests.php` (update path)
- Added `json_encode()` return value check in `Utils::escapeStringForHtml` (initial render path)

**Files changed:**

- `src/Exceptions/SnapshotEncodingException.php` (new)
- `src/Drawer/Utils.php`
- `src/Mechanisms/HandleRequests/UnitTest.php`

**Livewire version:** v4.1.4 (main branch)

### DIFF
--- a/src/Drawer/Utils.php
+++ b/src/Drawer/Utils.php
@@ -48,7 +48,15 @@ class Utils extends BaseUtils
             return htmlspecialchars($subject, ENT_QUOTES|ENT_SUBSTITUTE);
         }
 
-        return htmlspecialchars(json_encode($subject), ENT_QUOTES|ENT_SUBSTITUTE);
+        $encoded = json_encode($subject);
+
+        if ($encoded === false) {
+            throw new \RuntimeException(
+                'Failed to JSON-encode value for HTML attribute: ' . json_last_error_msg()
+            );
+        }
+
+        return htmlspecialchars($encoded, ENT_QUOTES|ENT_SUBSTITUTE);
     }
 
     static function pretendResponseIsFile($file, $contentType = 'application/javascript; charset=utf-8')
--- a/src/Exceptions/SnapshotEncodingException.php
+++ b/src/Exceptions/SnapshotEncodingException.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Livewire\Exceptions;
+
+class SnapshotEncodingException extends \Exception
+{
+    public function __construct(string $componentName, string $jsonError)
+    {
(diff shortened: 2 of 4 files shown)


joshhanley:
@sathia-musso thanks for the PR and for identifying this\! I've taken a slightly different approach in PR #10054 using `JSON_THROW_ON_ERROR` which avoids the need for manual checks and the custom exception class. Going to close this one in favour of that, but really appreciate you digging into this.

PR #10058, opened 2026-02-26
[4.x] Add optimistic actions with automatic rollback

utsavsomaiya:
## Summary
This introduces optimistic action support so Livewire UIs can respond instantly while still staying safe when requests fail.

Actions can now be marked optimistic using either:
- `wire:click.optimistic`
- `#[Optimistic]`
- `$wire.$set(..., { optimistic: true })`

If an optimistic request fails, Livewire rolls those optimistic client updates back to the last known server state.

## Real-world example
A common case is a "Like" button (or quick cart quantity toggle) on slower networks:

- User clicks and sees the count update immediately.
- Request succeeds: value stays.
- Request fails: value is automatically restored.

This gives the speed users expect without leaving the UI in a broken state.

## What changed
- Added optimistic action detection in the request/action pipeline.
- Added rollback capture/replay for optimistic updates on error/failure/cancel.
- Added new `#[Optimistic]` attribute that marks methods in snapshot memo.
- Added docs for optimistic usage and updated action/click docs.
- Added JS and PHP tests for optimistic detection, rollback behavior, and attribute memoing.

## Safety/compatibility
- Fully opt-in behavior.
- No changes for existing actions unless optimistic mode is explicitly used.

### DIFF
--- a/docs/__nav.md
+++ b/docs/__nav.md
@@ -63,6 +63,7 @@ PHP Attributes:
     Locked: { uri: /docs/4.x/attribute-locked, file: /attribute-locked.md }
     Modelable: { uri: /docs/4.x/attribute-modelable, file: /attribute-modelable.md }
     On: { uri: /docs/4.x/attribute-on, file: /attribute-on.md }
+    Optimistic: { uri: /docs/4.x/attribute-optimistic, file: /attribute-optimistic.md }
     Reactive: { uri: /docs/4.x/attribute-reactive, file: /attribute-reactive.md }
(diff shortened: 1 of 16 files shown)


calebporzio:
interesting idea... i think there's something here, but it's not mergeable as is

utsavsomaiya:
@calebporzio Sorry to tag you here again. I saw something today in Inertia 3.0 from @joetannenbaum. As you know, Inertia implements this approach, so I was wondering why we don't do something similar in Livewire without leaving PHP.

That’s why I introduced the same concept. If you’d like, I can reopen the PR.

PR #10060, opened 2026-02-26
[4.x] Add `wire:navigate.dirty-confirm` with custom dialog support

utsavsomaiya:
## Summary
- adds `wire:navigate.dirty-confirm` to protect users from accidentally leaving pages with unsaved changes
- supports custom modal flows by allowing an inline expression (for example opening a Tailwind dialog)
- keeps browser confirm as the default behavior and provides a fallback helper for custom flows

## Why this helps
- prevents accidental data loss in forms and editing screens
- lets teams use their own dialog UX while keeping navigation safety built in
- reduces repetitive hand-rolled unsaved-change handling across applications

## Validation
- JS tests for dirty detection, custom-expression behavior, and browser fallback behavior
- browser tests for both default confirm flow and custom dialog flow
- dist assets rebuilt

### DIFF
--- a/docs/navigate.md
+++ b/docs/navigate.md
@@ -197,6 +197,31 @@ To do this, you must add `wire:navigate:scroll` to the element containing a scro
 @endpersist
 ```
 
+### Confirming before leaving dirty state
+
+When users have unsaved Livewire changes, you can require confirmation before navigating away:
+
+```blade
+<a href="/settings" wire:navigate.dirty-confirm>Settings</a>
+```
+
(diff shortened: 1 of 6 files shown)


calebporzio:
Interesting idea @utsavsomaiya ! However I'm not i love with the final form of the feature. So I think I'm going to say "no" for now. 

Heads up, I'm going to be going through your pull requests soon and there are a lot so it will be exhausting to reply to each one. I will either merge or close without comment.

Thanks for your contributions! (next time, maybe not all at once though 😉)

PR #10061, opened 2026-02-26
[4.x] Fix stale reactive snapshots for overlapping `wire:model.live` requests

cyppe:

This fixes a real runtime failure (`Cannot mutate reactive prop [facets]`) we hit with rapid `wire:model.live` updates in a parent component that passes reactive data to a child. I do not have a prior discussion link for this specific case.

Yes, this is on `cyppe/reactive-prop-race-repro`.

No, this PR is focused on one race-condition fix plus regression coverage for that same bug.

Yes.

## Background
In production, we observed `Cannot mutate reactive prop [facets]` while typing quickly into a parent `wire:model.live` input. The parent updates facet data in `updatedSearch()`, and a child consumes that data via `#[Reactive]` (implemented internally by `BaseReactive`).

## Reproduction on `main`
When two `model.live` updates overlap:
1. Parent request is in-flight.
2. Second `model.live` action is allowed to run in parallel.
3. Because reactive/modelable children are bundled into parent requests, the second bundled request can carry a stale parent snapshot.

Both browser tests intercept outgoing requests and track the parent's snapshot `updateCount`/`count` value sent in each request:
- **stale** (without fix): `[0,0]` — both requests send the same initial snapshot
- **fresh** (with fix): `[0,1]` — second request waits for the first to complete, then sends the updated snapshot

Beyond the snapshot staleness, this also causes **silent state loss**: any server-side accumulation (counters, appended arrays, toggled flags) in `updatedSearch()` processes from stale state, so increments are lost and the child receives reactive prop values that don't reflect the full sequence of updates.

## Root cause
In `js/request/interactions.js`, the current fast-path allows overlapping `model.live` actions to run in parallel whenever both incoming and in-flight actions are `model.live`. That is safe for plain components, but unsafe when either side has reactive/modelable child bindings that depend on fresh parent snapshots.

## Fix
Keep the existing `model.live` parallel fast-path, but only when neither action scope has bound children.

Concretely:
- if both sides are `model.live` **and** neither has bound children: **keep parallel behavior (unchanged, no performance impact)**
- if either side has bound children: defer the incoming action to run after the in-flight message finishes

This uses existing defer/queue behavior and only narrows concurrency in the problematic nested-binding scenario. Components without reactive/modelable children are completely unaffected.

## What this solves
- Prevents stale parent snapshots from being bundled with reactive child commits.
- Makes overlapping requests carry fresh parent snapshot progression (`[0,1]` in the repro).
- Prevents the reactive mutation error path in the rapid-input reproduction.
- Prevents silent state loss from server-side logic processing stale data.

## Tests
Both browser tests type characters with a short pause between them to create overlapping requests while the server delay (`usleep`) keeps the first request in-flight. Each test verifies snapshot freshness deterministically via `Livewire.interceptRequest()`.

- `DUSK_HEADLESS_DISABLED=false phpunit --testsuite=Browser src/Features/SupportReactiveProps/BrowserTest.php`
- `phpunit --testsuite=Unit src/Features/SupportReactiveProps/UnitTest.php`
- `DUSK_HEADLESS_DISABLED=false phpunit --testsuite=Browser src/Features/SupportWireModelingNestedComponents/BrowserTest.php`
- `DUSK_HEADLESS_DISABLED=false phpunit --testsuite=Browser src/Features/SupportInterceptors/BrowserTest.php`
- `DUSK_HEADLESS_DISABLED=false phpunit --testsuite=Browser src/Features/SupportRequestInteractions/BrowserTest.php --filter='test_a_new_component_level_(user|poll)_action'`
- `npm run test:run`

## Notes
- I intentionally did not include built `dist/` assets in this PR; CI workflows build and use artifacts for JS/browser jobs.

If maintainers prefer a different concurrency boundary here (for example, a broader or narrower serialization rule), I would really value feedback and I am happy to revise quickly.


calebporzio:
@claude - can you please do the following:

- resolve merge conflicts
- clean up or simplify code if needed
- add a comment so that it's clear why these conditions exist

calebporzio:
@claude - try again, your last attempt failed in CI. just look at my last message to you and use that

calebporzio:
Thanks @cyppe! I pushed a few maintainer edits to take this over the finish line:

- **Merged `main` in to resolve the conflicts** (the branch had fallen behind; conflicts were just both sides adding sibling test methods to `UnitTest.php` / `BrowserTest.php` — kept both sets).
- **Simplified the `model.live` guard** in `js/request/interactions.js` — flattened the nested conditionals into a single `bothAreModelLive` check.
- **Added comments** explaining *why* a component with reactive/modelable children can't run overlapping `model.live` requests in parallel (otherwise both bundle the same stale parent snapshot).

Verified end-to-end in a real app before/after the fix: without it, two rapid keystrokes both send the stale snapshot (`[0,0]`) and an increment is silently lost; with it, the second request defers and sends fresh state (`[0,1]`). Unit + browser suites for `SupportReactiveProps` and `SupportRequestInteractions` all green.

The diff against `main` is back to the original 3 files. 🙏

joshhanley (approved):
Looks good to me! I do wonder whether the `componentHasBoundChildren()` helper should be on the component itself, so then we can call `action.component.hasBoundChildren()` or in the supports reactive code, rather than directly in `interactions.js`. But that's just a nitpick. Makes sense to not let requests run in parallel if they have reactive children.

calebporzio:
Traced this end-to-end — the `model.live` parallel fast-path was letting a second keystroke compile against a stale parent snapshot whenever reactive/modelable children were bundled into the request. Serializing just that case (and leaving plain components fully parallel) is the right boundary, and the deferred action can't be lost since every terminal path (`cancel`/`failure`/`error`) re-fires via `invokeOnFinish` after `mergeNewSnapshot`. Ready to merge; @joshhanley's `hasBoundChildren()` nitpick can land as a follow-up. Thanks @cyppe!

PR #10062, opened 2026-02-26
[4.x] Fix $-prefixed expressions not routing to Livewire inside Alpine scope

people-sea:

Yes. This fixes a regression in 4.2 where $-prefixed expressions (like $set, $toggle) inside Alpine scope get intercepted instead of routing to Livewire. Reported in            
  filamentphp/filament#19354.       

Yes

No


Yes

  getAlpineScopeKeys() in the evaluator was collecting $-prefixed keys (like $set, $get) from Alpine's data stack. This caused expressions like $set('foo', 'bar') to be treated as
   Alpine scope calls instead of being contextualized to $wire.$set('foo', 'bar').                                                                                                 
                                                                                                                                                                                   
  Fix: skip keys starting with $ when collecting Alpine scope keys.

```js
  // before
  if (! keys.includes(key)) keys.push(key)

  // after
  if (! keys.includes(key) && ! key.startsWith('$')) keys.push(key)
```

### DIFF
--- a/js/evaluator.js
+++ b/js/evaluator.js
@@ -9,7 +9,7 @@ function getAlpineScopeKeys(el) {
         if (currentEl._x_dataStack) {
             for (let scope of currentEl._x_dataStack) {
                 for (let key of Object.keys(scope)) {
-                    if (! keys.includes(key)) keys.push(key)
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks!

PR #10074, opened 2026-03-01
[4.x] Add native chunked file uploads

gregharmon:
Adds a \`wire:model.chunked\` modifier that automatically splits large files into smaller chunks client-side, uploads them sequentially via signed URLs, and reassembles them server-side.

Existing \`TemporaryUploadedFile\` flow is preserved — components receive the same object regardless of upload method.

## Usage

\`\`\`blade

<input type="file" wire:model.chunked="video">

<input type="file" wire:model.chunked.4096="video">

<input type="file" wire:model.chunked="files" multiple>
\`\`\`

## How it works

1. JS parses \`.chunked\` modifier, slices file with \`Blob.slice()\`
2. PHP \`_startChunkedUpload\` generates a UUID upload session stored in the user's session, returns a signed URL
3. JS POSTs chunks sequentially to \`/livewire/upload-chunk\` endpoint
4. Server stores chunks in \`livewire-tmp/chunks/{uploadId}/\` as zero-padded files
5. On final chunk: server reassembles, validates against global upload rules, stores as \`TemporaryUploadedFile\`
6. Normal \`_finishUpload\` flow resumes — component receives standard \`TemporaryUploadedFile\`

## Security

- **Signed URLs** — each upload session gets a tamper-proof signed URL
- **Session validation** — uploadId stored server-side, verified on every chunk
- **Per-chunk size limits** — configurable via \`livewire.temporary_file_upload.max_chunk_size\`
- **Reassembled file validation** — full file validated against \`FileUploadConfiguration::rules()\`
- **Chunk isolation** — UUID directories prevent collision/guessing

## Configuration

Two optional keys added to \`config/livewire.php\` under \`temporary_file_upload\`:

\`\`\`php
'max_chunk_size' => null,   // Default: 3MB (3145728 bytes)
'chunk_middleware' => null,  // Default: 'throttle:300,1'
\`\`\`

> **Note:** The global \`temporary_file_upload.rules\` validation (default \`max:12288\`) applies to the reassembled file. Increase this limit to match the maximum file size you want to support with chunked uploads.

## Constraints

- S3 chunked uploads are out of scope (S3 has its own multipart upload API)
- Chunks are sequential for simplicity and reliable progress tracking

## Tests

- 17 unit tests covering: session management, chunk storage, reassembly, validation, cancellation, size limits, error handling, middleware configuration
- 3 browser tests covering: single file upload, multiple file upload, upload cancellation


calebporzio:
closing because we're actively working on this. thank you.

PR #10076, opened 2026-03-01
[4.x] Fix namespace prefix missing in generateNameFromClass

mozex:
I ran into this while working on a modular Laravel package that uses `addNamespace()` to register Livewire components per module.

**The problem:**

When you register a component via `addNamespace()` and then reference it by class in `Route::livewire()`, it throws `ComponentNotFoundException`.

```php
// Register namespace in a service provider
Livewire::addNamespace('admin', classNamespace: 'App\\Admin\\Livewire');

// This works fine
Route::livewire('/dashboard', 'admin::dashboard');

// This throws ComponentNotFoundException
Route::livewire('/dashboard', Dashboard::class);
```

I traced it down to `Finder::generateNameFromClass()` when it matches a class against a registered namespace, it strips the `classNamespace` prefix but doesn't prepend the namespace key back. So `App\Admin\Livewire\Dashboard` becomes `'dashboard'` instead of `'admin::dashboard'`.

The issue is that the `foreach` loop iterates over a merged array of `classNamespaces` (string-keyed) and `classLocations` (numerically-keyed), but never uses the key:

```php
foreach ($classNamespaces as $classNamespace) { // key is ignored
    // ...
    return (string) $fullName->substr(strlen($namespace) + 1); // returns bare name
}
```

**The fix:**

Use the key in the loop. If it's a string key (from `classNamespaces`), prepend it. If it's numeric (from `classLocations`), return the bare name as before.

1. Yes - this breaks `Route::livewire()` with class references for any component registered via `addNamespace()`.
2. Yes - separate branch.
3. No - single focused fix.
4. Yes - two unit tests added to `src/Finder/UnitTest.php` covering both flat and nested namespace class normalization.
5. See above

### DIFF
--- a/src/Finder/Finder.php
+++ b/src/Finder/Finder.php
@@ -336,7 +336,7 @@ class Finder
             ->merge($this->classLocations)
             ->toArray();
 
-        foreach ($classNamespaces as $classNamespace) {
+        foreach ($classNamespaces as $key => $classNamespace) {
             $namespace = str_replace(
                 ['/', '\\'],
                 '.',
@@ -348,7 +348,9 @@ class Finder
                 ->implode('.');
 
             if ($fullName->startsWith($namespace)) {
-                return (string) $fullName->substr(strlen($namespace) + 1);
+                $name = (string) $fullName->substr(strlen($namespace) + 1);
+
+                return is_string($key) ? $key . '::' . $name : $name;
             }
         }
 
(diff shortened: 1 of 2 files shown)


mozex:
@calebporzio would you mind taking a look at this?

calebporzio:
makes sense to me. thanks!

bbredewold:
This is great :) For full-page components, the V3 docs show examples using the FQCN. notation when declaring routes.
In V4, the examples go with the string notation `namespace::component-name`. I was wondering if this now is the preferred way.

```php
Route::get('/posts/create', CreatePost::class); // v3
Route::livewire('/posts/create', 'pages::post.create'); // v4
```

PR #10082, opened 2026-03-02
[3.x] Fix `smart_wire_keys` breaking `@while` and `@for` loops

joshhanley:
# The Scenario

When `smart_wire_keys` is enabled (the default), using `@while` or `@for` loops in a Livewire component throws an `Undefined variable $loop` error, preventing the page from rendering.

```php
<?php

use Livewire\Component;

new class extends Component
{
    //
};
?>

<div>
    @while (true)
        <p>I'm looping forever.</p>
        @break
    @endwhile
</div>
```

# The Problem

In `SupportMorphAwareBladeCompilation`, the `isLoop()` method treats `@while` and `@for` as loop directives. When a loop directive is detected, the precompiler injects code that references `$loop->index`:

```php
$suffix .= '<?php \Livewire\Features\SupportCompiledWireKeys\SupportCompiledWireKeys::startLoop($loop->index); ?>';
```

The `$loop` variable is only available in `@foreach` and `@forelse` loops (provided by Laravel's Blade compiler). It does not exist for `@while` or `@for` loops, so the injected code throws an `Undefined variable $loop` error at runtime.

# The Solution

Remove `@while` and `@for` (and their closing counterparts `@endwhile` and `@endfor`) from the `isLoop()` and `isEndLoop()` methods. These directives still get conditional morph markers (`<!--[if BLOCK]><![endif]-->`) but no longer have loop tracking markers injected.

Fixes https://github.com/livewire/livewire/discussions/10078

### DIFF
--- a/src/Features/SupportMorphAwareBladeCompilation/SupportMorphAwareBladeCompilation.php
+++ b/src/Features/SupportMorphAwareBladeCompilation/SupportMorphAwareBladeCompilation.php
@@ -318,6 +318,9 @@ class SupportMorphAwareBladeCompilation extends ComponentHook
         $loopDirectives = [
             'foreach',
             'forelse',
+            // temp disabling because of "missing $loop" error
+            // 'for',
+            // 'while',
         ];
 
         $pattern = '/@(' . implode('|', $loopDirectives) . ')(?![a-zA-Z])/i';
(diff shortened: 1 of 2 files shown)


joshhanley:
Closing in favour of #10084

PR #10083, opened 2026-03-02
[4.x] Fix sessioned pagination not restoring query string on revisit

joshhanley:
# The Scenario

When persisting the current page to session and restoring it on revisit, the page content is correctly restored but the URL stays clean — the query string is not synced to reflect the restored page. This affects both standard and cursor pagination.

```php
<?php

use Livewire\Volt\Component;
use Livewire\WithPagination;
use App\Models\Post;

new class extends Component {
    use WithPagination;

    public function mount()
    {
        if ($page = session('pagination_page')) {
            $this->setPage($page);
        }
    }

    public function updatedPaginators($page, $pageName)
    {
        session(['pagination_page' => $page]);
    }

    #[Computed]
    public function posts()
    {
        return Post::paginate(10);
    }
}

?>

<div>
        <h1 wire:key="post-{{ $post->id }}">{{ $post->title }}</h1>
    @endforeach

    {{ $this->posts->links() }}
</div>
```

1. User navigates to page 2 — URL updates to `?page=2`, page stored in session
2. User navigates away
3. User revisits with a clean URL — page 2 content is shown (session restored) but the URL stays as `/posts` instead of `/posts?page=2`

# The Problem

Two issues in `SupportPagination::ensurePaginatorIsInitialized()` prevent the query string from syncing:

1. **URL hook skipped when paginator value is pre-set.** The method has an early return when `$this->component->paginators[$pageName]` is already set. This guards both value resolution and URL hook setup. When user code sets the paginator from session during `mount()` (before the paginator resolver runs), the `PaginationUrl` attribute is never created and no URL effect is sent to the frontend.

2. **Missing `except` value on `PaginationUrl`.** The `addUrlHook()` method creates a `PaginationUrl` without an `except` value (defaults to `null`). The push-mode sync logic added in #10028 compares the current value against `except` to detect session-restored values — when `except` is `null`, the comparison falls back to comparing the ephemeral value against itself, which is always equal, so the URL is never synced.

# The Solution

In `ensurePaginatorIsInitialized()`, the early return when the paginator is already set now only applies to subsequent Livewire AJAX requests (where the value came from hydration and JS already has URL tracking from the initial load), or when `SupportQueryString` has already registered a `BaseUrl` attribute for the paginator (e.g. via the legacy `$queryString` property). On the initial page load, when no existing URL tracking exists, the URL hook is added even when the paginator value was pre-set by user code (e.g. session restoration in `mount()`).

In `addUrlHook()`, when the paginator was pre-set, the `$defaultPage` is passed as `except` to `PaginationUrl`. This tells JS that `1` is the default page value (or `''` for cursor pagination), enabling the push-mode sync logic from #10028 to detect when the session-restored value differs from the default and sync the URL via `replaceState`.

Fixes https://github.com/livewire/livewire/discussions/10064

### DIFF
--- a/src/Features/SupportPagination/BrowserTest.php
+++ b/src/Features/SupportPagination/BrowserTest.php
@@ -1155,6 +1155,43 @@ class BrowserTest extends BrowserTestCase
             ;
         });
     }
+
+    public function test_sessioned_cursor_pagination_query_string_is_restored_on_revisit()
+    {
+        $this->beforeServingApplication(function () {
+            app('livewire')->component('session-cursor-pagination', SessionCursorPaginationComponent::class);
+
+            Route::get('/session-cursor-pagination', function () {
+                return app('livewire')->new('session-cursor-pagination')();
+            })->middleware('web');
+        });
+
+        $this->browse(function ($browser) {
+            $browser
+                ->visit('/session-cursor-pagination')
+                ->waitForLivewireToLoad()
+                ->assertSee('Post #1')
+                ->assertSee('Post #3')
+                ->assertDontSee('Post #4')
+                ->assertQueryStringMissing('page')
+
+                // Navigate to page 2 — stores cursor in session...
+                ->waitForLivewire()->click('@nextPage')
+                ->assertSee('Post #4')
+                ->assertSee('Post #6')
+                ->assertDontSee('Post #1')
+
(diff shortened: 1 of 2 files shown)


joshhanley:
I don't think this is the right solution, so closing this.

PR #10089, opened 2026-03-04
[4.x] Add `Authorize` attribute

PerryvanderMeer:
This PR introduces a declarative `#[Authorize]` attribute for Livewire actions, bringing the same elegant authorization DX recently introduced in the Laravel framework https://github.com/laravel/framework/pull/59048 directly into Livewire components.

Currently, authorizing actions in Livewire requires manual calls to `Gate::authorize()` or `$this->authorize()` inside the method body. This attribute allows you to lift that logic into the method signature, making the intent of your code clearer and more concise.

### Basic Gate Check

For simple permissions that don't require a model instance:

```php
#[Authorize('view-dashboard')]
public function showStats() { ... }
```

### Class-based Gate Check

Useful for 'create' permissions where no instance exists yet:

```php
#[Authorize('create', Post::class)]
public function create() { ... }
```

### Authorizing via Component Property

The attribute can automatically pull models from the component's state:

```php
public Post $post;

#[Authorize('update', 'post')]
public function save() { ... }
```

### Authorizing via Method Parameter

It also supports resolving models passed directly to the action, including support for route-model-binding resolution:

```php
#[Authorize('delete', 'comment')]
public function delete(Comment $comment) { ... }
```

### DIFF
--- a/docs/attribute-authorize.md
+++ b/docs/attribute-authorize.md
@@ -1,8 +1,8 @@
-The `#[Authorize]` attribute integrates Laravel’s Gate system directly into your Livewire actions. It ensures that an action is only executed if the user has the necessary permissions, throwing a `403 Forbidden` response otherwise.
+The `#[Authorize]` attribute integrates Laravel's Gate system directly into your Livewire actions. It ensures that an action is only executed if the user has the necessary permissions, throwing a `403 Forbidden` response otherwise.
 
 ## Basic usage
 
-Apply the `#[Authorize]` attribute to any action method. You can pass the ability name and an optional argument:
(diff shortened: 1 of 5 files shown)


calebporzio:
Very nice, makes good sense to me and I think I'd use it.

Only slightly odd thing would be the second parameter pointing to either a property or a method param. I assume Laravel just looks at params, but because Livewire uses properties, we of course need to support those.

But I think I'm fine with something like: `#[Authorize('update', 'post')]` first looking at method params, then falling back to property values.

PerryvanderMeer:
Makes sense @calebporzio. I've changed the resolution order to prioritize the parameter over the property.

calebporzio:
Love this! Thanks for the contribution @PerryvanderMeer !

PR #10108, opened 2026-03-10
[4.x] Fix form object properties bypassing hydration and lifecycle hooks during consolidated updates

joshhanley:
# The Scenario

When all properties of a form object are changed simultaneously (e.g., filling out every field before submitting), two things break:

1. **Enum-typed properties throw a `TypeError`** because the raw string value is assigned directly without casting:

```
TypeError
Cannot assign string to property App\Livewire\Foo\Form::$enumfield of type App\Enums\SomeEnum
```

2. **The `updated()` lifecycle hook on the form object is never called**, even though every property has changed.

Both issues only occur when *all* form fields change from their initial values. If even one field remains unchanged, everything works correctly.

```php
<?php

use App\Enums\SomeEnum;
use Livewire\Form;
use Livewire\Component;

class MyForm extends Form
{
    public string $title = '';
    public SomeEnum $status = SomeEnum::Draft;

    public function updated(string $property)
    {
        // This is never called when ALL fields change at once
    }
}

class Page extends Component
{
    public MyForm $form;

    public function save(): void
    {
        // TypeError: Cannot assign string to property MyForm::$status of type SomeEnum
    }
}
?>

<form wire:submit="save">
    <input type="text" wire:model="form.title" />

    <select wire:model="form.status">
        @foreach (SomeEnum::cases() as $enum)
            <option :value="$enum">{{ $enum->value }}</option>
        @endforeach
    </select>

</form>
```

# The Problem

When only some form fields change, the JavaScript diff algorithm sends individual dot-notation updates (e.g., `form.title: '...'`, `form.status: 'active'`). Each goes through `hydrateForUpdate()` (which handles type casting via synthesizers) and `updateProperty()` (which fires lifecycle hooks).

When **all** form fields change, the JavaScript `diffAndConsolidate()` function consolidates them into a single form-level update (e.g., `form: {title: '...', status: 'active'}`). This consolidated update bypasses the normal per-property hydration path entirely: `hydratePropertyUpdate()` replaces the whole form object via `FormObjectSynth::hydrate()`, which assigns raw values directly to typed properties (causing `TypeError` for enums) and skips the `trigger('update', ...)` call that fires lifecycle hooks.

# The Solution

Form objects are special; they should never be fully replaced during a property update. In `HandleComponents::updateProperties()`, detect when a consolidated update targets a form object and decompose it back into individual dot-notation property updates. This ensures each property goes through the normal `hydrateForUpdate()` and `updateProperty()` path, which properly handles type casting (enums, Carbon, etc.) and fires lifecycle hooks (`updating`/`updated`).

Fixes #10086
Fixes #10102

### DIFF
--- a/src/Features/SupportFormObjects/FormObjectSynth.php
+++ b/src/Features/SupportFormObjects/FormObjectSynth.php
@@ -3,10 +3,8 @@
 namespace Livewire\Features\SupportFormObjects;
 
 use Livewire\Drawer\Utils;
-use Livewire\Mechanisms\HandleComponents\HandleComponents;
 use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
 use Livewire\Features\SupportAttributes\AttributeCollection;
-use ReflectionUnionType;
 
 use function Livewire\wrap;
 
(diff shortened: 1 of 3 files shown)


lode:
Thanks for picking this up!

I'm not that much invested in the differences between components and forms. But I was wondering if it is form specific. If I have a component with the same setup (two properties of which one is an enum), I also want enum casting and update hooks called. Will that work?

joshhanley:
@lode no worries! It's definitely form specific due to the way updates are consolidated on the front end. For a component, even if you update all of it's properties in one go, it will still send individual changes to the backend. Where as with a form, if you update all it's properties, the component consolidates them all into an update called `'form'` (or whatever your form property name is) before sending them to the backend.

So a component like that will work fine as it is now 🙂

lode:
Thanks, great!

PR #8956, opened 2024-10-21
[3.x] Fix wire dirty

juliangarcess:
### Contribution Checklist

1️⃣ **Is this something that is wanted/needed?**  
   - Yes, this addresses a bug as discussed in the following issue:  
   [GitHub Discussion #8758](https://github.com/livewire/livewire/discussions/8758).

2️⃣ **Did you create a branch for your fix/feature?**  
   - Yes, I created a dedicated branch for this fix.

3️⃣ **Does it contain multiple, unrelated changes?**  
   - No, the PR only includes changes relevant to this issue.

4️⃣ **Does it include tests?**  
   - No new tests were added as the fix addresses a specific behavior, but no additional test cases are necessary for this change.

5️⃣ **Description of Improvement:**  
   - **What was changed:**  
     I modified `wire-dirty.js` to use the `succeed` hook of the commit instead of the `respond` hook. The `respond` hook caused the DOM to update before the values were flagged as dirty. By switching to the `succeed` hook, the update occurs at the right time, ensuring correct behavior for dirty state tracking.
     
   - **Why it's useful:**  
     This change ensures the DOM is only updated after all values have been marked as dirty, which resolves the issue of premature updates. For more details, refer to the [GitHub Discussion #8758](https://github.com/livewire/livewire/discussions/8758).

---

Thanks for contributing! 🙌

PR #10113, opened 2026-03-12
[4.x] Fix wire dirty (again)

juliangarcess:
### Contribution Checklist

1️⃣ **Is this something that is wanted/needed?**  
- Yes. This addresses a bug that had already been identified and discussed in the following issue:  
  [GitHub Discussion #8758](https://github.com/livewire/livewire/discussions/8758).  
  In fact, this was previously fixed in PR #8956, which was merged, but the failing code was later reintroduced.

2️⃣ **Did you create a branch for your fix/feature?**  
- Yes, I created a dedicated branch for this fix.

3️⃣ **Does it contain multiple, unrelated changes?**  
- No, the PR only includes changes relevant to this issue.

4️⃣ **Does it include tests?**  
- Yes.

5️⃣ **Description of Improvement:**  
- **What was changed:**  
  I modified `wire-dirty.js` to use the `succeed` hook of the commit instead of the `respond` hook. The `respond` hook caused the DOM to update before the values were flagged as dirty. By switching to the `succeed` hook, the update occurs at the right time, ensuring correct behavior for dirty state tracking.

- **Why it's useful:**  
  This change ensures the DOM is only updated after all values have been marked as dirty, which resolves the issue of premature updates. This bug had already been fixed previously in PR #8956, but the problematic code path was later reintroduced, causing the issue to appear again. For more details, refer to the [GitHub Discussion #8758](https://github.com/livewire/livewire/discussions/8758).

---

Thanks for contributing! 🙌


joshhanley:
@juliangarcess thanks for the PR! Can you please add new tests that cover this change, so it never gets broken again? Also please remove `dist/` files from the PR. They get built in CI for PRs now. Thanks!

juliangarcess:
@joshhanley done!

calebporzio:
makes sense, thanks!

PR #10117, opened 2026-03-16
[4.x] Ensure Livewire route with scoped bindings works regardless of order of properties in livewire component

gdebrauwer:
I was trying to use scoped bindings on a Livewire route in my project, but the route did not return a 404 when the child model did not belong to the provided parent.

```php
Route::livewire('scoped-bindings/{store}/{book}', 'pages::scoped-bindings')->scopeBindings();
```

After some debugging, I found out that the scoped bindings did not work because the child property was defined before the parent property in the Livewire component.

```php
// $book property is defined before $store property
// so scoped bindings currently do not work.

new class extends Component {
    public Book $book;

    public Store $store;
}
```

This PR fixes that.

```php
// $store property can be defined before or after $book property,
// scope bindings will work in both scenarios.

new class extends Component {
    public Book $book;

    public Store $store;
}

new class extends Component {
    public Store $store;

    public Book $book;
}
```

### DIFF
--- a/src/Drawer/ImplicitRouteBinding.php
+++ b/src/Drawer/ImplicitRouteBinding.php
@@ -69,6 +69,9 @@ class ImplicitRouteBinding
     {
         return $this->getPublicPropertyTypes($component)
             ->intersectByKeys($route->parametersWithoutNulls())
+            ->sortBy(function ($className, $propName) use ($route) {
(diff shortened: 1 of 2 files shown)


calebporzio:
great, thanks!

PR #10119, opened 2026-03-16
Fix wire:navigate CSP nonce mismatch violations

valorin:

This is a bugfix. I couldn't find any existing discussions, so attempted to fix.

Yes, own branch.

Single change to fix the issue.

Yes, although they were written by Claude - and won't run locally because I don't have Chrome set up in my dev env.

Thanks for contributing! 🙌

_Note, the fix and tests were implemented using Claude, because my JS skills are limited. I have full tested the fix in my own app, but it might need some wider testing, and as noted above, I haven't been able to run the tests._

## Problem

  When using `wire:navigate` with a strict nonce-based Content Security Policy, every SPA navigation triggers CSP violations.
  Each server response includes a unique CSP nonce, but during SPA navigation the browser continues enforcing the **original
  page's** CSP header. When Livewire fetches a new page and swaps the HTML, the new page's nonce-bearing `<script>`, `<style>`,  and `<link>` elements are rejected because their nonces don't match.

  This causes three distinct violations:
  1. **`DOMParser.parseFromString()`** reports violations when parsing HTML containing nonce'd elements — the browser enforces
  CSP even on inert parsed documents
  2. **`mergeNewHead()`** re-injects already-loaded assets because `outerHTML` comparison includes the nonce (which differs per
   request), making identical assets look "new"
  3. **`cloneScriptTag()`** copies nonces via `setAttribute()` which browsers ignore for CSP — the `.nonce` IDL property is
  required for dynamically inserted elements

  ## Solution

  Four changes to `js/plugins/navigate/page.js`:

  1. **Capture the initial CSP nonce** on first page load (via `livewireScriptConfig.nonce` or the `.nonce` IDL property) and
  store it for the SPA session lifetime
  2. **Replace nonces in raw HTML before parsing** — `replaceNoncesInHtml()` extracts the new page's nonce from the fetched
  HTML string and replaces all occurrences with the captured nonce *before* `DOMParser.parseFromString()` sees it
  3. **Compare head assets ignoring nonces** — `mergeNewHead()` uses the existing `ignoreAttributes()` function (already used
  for body script deduplication) to strip nonces before comparing
  4. **Use `.nonce` IDL property in `cloneScriptTag()`** — dynamically created scripts use `script.nonce = value` instead of
  `setAttribute('nonce', value)`

  ## Test

  Added `test_navigate_does_not_trigger_csp_violations_with_different_nonces` browser test that:
  - Renders two pages with different CSP nonces and enforcing `Content-Security-Policy` headers
  - Navigates between them via `wire:navigate`
  - Asserts no console errors and no CSP violation warnings

### DIFF
--- a/js/plugins/navigate/page.js
+++ b/js/plugins/navigate/page.js
@@ -6,7 +6,18 @@ let attributesExemptFromScriptTagHashing = [
     'aria-hidden',
 ]
 
+// Capture the current page's CSP nonce on first load, before the browser
+// clears it from the DOM. This nonce must be used for all subsequent
+// wire:navigate page swaps since the browser keeps enforcing the original
+// page's CSP header throughout SPA navigation.
+let currentPageNonce = captureInitialNonce()
+
 export function swapCurrentPageWithNewHtml(html, andThen) {
+    // Replace nonces in the raw HTML string before parsing so that
+    // DOMParser.parseFromString() doesn't trigger CSP violations.
+    // The browser enforces CSP even on inert parsed documents.
+    html = replaceNoncesInHtml(html)
+
     let newDocument = (new DOMParser()).parseFromString(html, "text/html")
     let newHtml = newDocument.documentElement
     let newBody = document.adoptNode(newDocument.body)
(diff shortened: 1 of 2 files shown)


calebporzio:
thanks for this, merged #10133 which supersedes this, made sure you have contribution credit in that one @valorin - thanks!

PR #10120, opened 2026-03-16
Fix SFC parser extracting style/script tags from inside @verbatim blocks

ganyicz:
# The scenario

A single-file component with a `<style>` or `<script>` tag inside a `@verbatim`/`@endverbatim` block has the tag incorrectly extracted and stripped from the view.

````blade
@verbatim
```html
<style>
    .example { color: red; }
</style>
```
@endverbatim
````

# The problem

We use a regex to find root-level `<style>` tags in the view portion, but it has no awareness of `@verbatim` blocks.

# The solution

Before running the tag-matching regex, strip `@verbatim`/`@endverbatim` blocks from the local copy of the view content used for matching.

```php
// Remove @verbatim/@endverbatim blocks (content inside is literal, not extractable)
$viewContents = preg_replace('/@verbatim\s*.*?@endverbatim/s', '', $viewContents);
```

This is the same pattern already used in `extractScriptPortion()` for `@script` and `@assets` directives.

### DIFF
--- a/src/Compiler/Fixtures/sfc-component-with-style-in-verbatim.blade.php
+++ b/src/Compiler/Fixtures/sfc-component-with-style-in-verbatim.blade.php
@@ -1,0 +1,19 @@
+<?php
+
+use Livewire\Component;
+
+new class extends Component
+{
+    public $message = 'Hello World';
+};
+?>
+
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks Filip!

PR #10124, opened 2026-03-18
[3.x] Backport Fix lazy components loading when reactive prop changes

Orrison:

Yes, this is a backport of https://github.com/livewire/livewire/pull/10079 for v4 to v3.

Yes

No

Yes

This backports a fix from Livewire v4 (done in https://github.com/livewire/livewire/pull/10079) that prevents lazy child components from being prematurely loaded when a parent updates a reactive prop.

The bug: When a parent component commits, `getDeepChildrenWithBindings()` bundles all child components with reactive props or wire:model bindings into the same network request — including lazy children that haven't been loaded yet (i.e., the user hasn't scrolled to them). This forces premature hydration of the lazy child, and can cause `$wire.$entangle()` errors when the child references properties that depend on mount(), which was skipped because the component was never actually lazy-loaded.

The fix: Added a guard in `getDeepChildrenWithBindings()` to skip any lazy child that hasn't completed its initial load. In v4 this uses isLazy / hasBeenLazyLoaded getters on the Component class; in v3 the equivalent check is `componentsThatAreLazy.has(child)`, which leverages the existing `WeakSet` that tracks lazy components until their first commit completes. Once the user scrolls to the lazy component and it loads normally, it will receive the current reactive prop value through normal rendering.

Thanks for contributing! 🙌


joshhanley:
@Orrison thanks for the PR! I've actually backported the v4 commit as part of our release process, so I'm going to close this. Thanks!

PR #10144, opened 2026-03-26
[4.x] Fix `<template x-if>` and `<template x-for>` content duplicating after morph

joshhanley:
# The Scenario

When using a `<template x-if>` or `<template x-for>` bound to a Livewire property and toggled via `wire:click`, the template content renders twice on the first interaction. Subsequent interactions render correctly.

The reporter confirmed this worked correctly on v4.1.0 and below.

```php
<?php

use Livewire\Component;

new class extends Component
{
    public bool $show = false;

    public function open(): void
    {
        $this->show = true;
    }
};
?>

<div>
    <template x-if="$wire.show">
        <p>Hello world!</p>
    </template>
    <button wire:click="open">Open</button>
</div>
```

# The Problem

During Livewire's morph, Alpine's `cloneNode` evaluates directives like `x-if` and `x-for` on the "to" element using already-updated reactive state. For `<template>` elements, these directives insert sibling DOM nodes via `el.after()`. This injects content into the "to" DOM that doesn't exist in the "from" yet. The morph algorithm adds this content, then Alpine's deferred effects add it again after the transaction commits, causing duplicates.

# The Solution

When both the "from" and "to" elements are `<template>` tags and the "from" template hasn't rendered content yet (checked via Alpine's internal `_x_currentIfEl` for `x-if` and `_x_prevKeys` for `x-for`), we stub `.after()` on the "to" element to a no-op. This prevents `cloneNode` from injecting sibling nodes during the morph. We also call `childrenOnly()` to preserve the "from" element's directives.

This approach handles both `x-if` and `x-for` templates, and avoids stripping directives from the "to" element.

Fixes https://github.com/livewire/livewire/discussions/10097

### DIFF
--- a/js/morph.js
+++ b/js/morph.js
@@ -146,6 +146,18 @@ function getMorphConfig(component) {
 
             if (isntElement(el)) return
 
+            // When an x-if template hasn't rendered its content yet (e.g. $wire.show just
+            // changed from false to true), prevent Alpine's cloneNode from evaluating x-if
+            // on the "to" element and injecting duplicate content into the "to" DOM. We strip
+            // the directive from "to" so cloneNode is inert, and use childrenOnly() to skip
+            // attribute patching so the "from" element keeps its original x-if directive...
+            if (el.tagName === 'TEMPLATE' && toEl.tagName === 'TEMPLATE'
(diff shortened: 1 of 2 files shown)


joshhanley:
Closing in favour of https://github.com/alpinejs/alpine/pull/4785

PR #10147, opened 2026-03-26
[4.x] Fix positional-after-named arg crash in callTraitHook

mariuscrisan:
`resolveMethodDependencies()` returns integer keys for optional parameters that have no matching route segment. When these mixed-key arrays are later unpacked with `...` in `callTraitHook()`, PHP throws `Cannot use positional argument after named argument during unpacking`.

Filter out integer-keyed entries in resolveMountParameters() before returning so only named (route-matched) parameters are passed downstream.

Fixes: https://github.com/livewire/livewire/discussions/10146

### DIFF
--- a/src/Drawer/ImplicitRouteBinding.php
+++ b/src/Drawer/ImplicitRouteBinding.php
@@ -62,7 +62,12 @@ class ImplicitRouteBinding
             throw $e;
         }
 
-        return new Collection($parameters);
+        // resolveMethodDependencies() adds default values for optional
+        // parameters with integer keys. Strip them so the array only
+        // contains named route-matched parameters and can be safely
+        // unpacked with `...` downstream (PHP forbids positional args
+        // after named args).
+        return new Collection(array_filter($parameters, 'is_string', ARRAY_FILTER_USE_KEY));
     }
 
     public function resolveComponentProps(Route $route, Component $component)
--- a/src/Features/SupportLifecycleHooks/SupportLifecycleHooks.php
+++ b/src/Features/SupportLifecycleHooks/SupportLifecycleHooks.php
@@ -197,11 +197,7 @@ class SupportLifecycleHooks extends ComponentHook
             }
 
             if (static::$methodCache[$cacheKey]) {
-                $keys = array_keys($params);
-                $paramsToSpread = array_filter($keys, 'is_string') && array_filter($keys, 'is_int')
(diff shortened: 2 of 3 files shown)


ganyicz:
Hey @mariuscrisan,

thanks for contributing! It seems that the CI failures are related to your changes.

Can you look into that please?

mariuscrisan:
Hey @ganyicz. Thanks for lookint into it! 

I found what was causing the L11 failures and pushed a fix.
 
The array_filter was in the wrong place. resolveMountParameters() feeds data back into Laravel's route binding, so stripping the integer-keyed entries there changed what L11's substituteImplicitBindings had to work with.

I reverted ImplicitRouteBinding.php back to returning the full Collection and I moved the integer-key filter to `callTraitHook()`, right at the `...$params` spread:

```php
wrap($this->component)->$method(...array_filter($params, fn ($key) => !  is_int($key), ARRAY_FILTER_USE_KEY));
```

Tested locally with Docker (PHP 8.4) and all relevant tests pass.

mariuscrisan:
I can see some some other unit tests are failing now. I will look into it and make sure to run all the unit tests locally before pushing again.

mariuscrisan:
My previous fix was a bit too broad. The integer-key filter now only applies when `$params` is a genuinely mixed
array (has both string and integer keys).

I squashed my commits into just one for brevity.

All unit tests passed successfully locally.

calebporzio:
Hey @mariuscrisan — thanks for this, the bug is real and the test is solid. I pushed a commit that moves the fix from `callTraitHook()` to `resolveMountParameters()` in `ImplicitRouteBinding.php`.

### What was happening

`resolveMethodDependencies()` returns a mixed-key array when there are optional mount params that don't match a route segment:

```php
// Route: /posts/{postId}, mount($postId, $tab = null)
// resolveMethodDependencies returns:
['postId' => '123', 0 => null]
```

When `callTraitHook()` tries to spread this with `...`, PHP throws `Cannot use positional argument after named argument`.

### Why I moved the fix

The root cause is the mixed-key array coming out of `resolveMountParameters()`. Cleaning it up at the source is simpler (one line instead of four) and keeps `callTraitHook()` generic — it shouldn't need to know about route parameter quirks.

```php
// Before
return new Collection($parameters);

// After
return new Collection(array_filter($parameters, 'is_string', ARRAY_FILTER_USE_KEY));
```

All 51 tests in SupportPageComponents + SupportLifecycleHooks pass.

PR #10153, opened 2026-03-27
[4.x] Fix file uploads failing with 403 when persistent middleware changes session ID

ganyicz:
## The scenario

File uploads use two separate requests: one to the **upload endpoint** and one to the **update endpoint**. If the session ID changes between them, the upload signature doesn't match and Livewire returns a 403.

The most documented case is **apps running inside an iframe** (e.g. Shopify embedded apps). Browsers block third-party cookies in iframes, so the session cookie never reaches the server. On each request, Laravel generates a fresh random session ID:

```php
// Illuminate\Session\Middleware\StartSession
public function getSession(Request $request)
{
    return tap($this->manager->driver(), function ($session) use ($request) {
        $session->setId($request->cookies->get($session->getName()));
    });
}

// Illuminate\Session\Store
public function setId($id)
{
    $this->id = $this->isValidId($id) ? $id : $this->generateSessionId();
}
```

When the cookie is `null`, `setId()` fails validation and generates a new ID — so each request gets a different session ID and the signatures never match.

[Discussion #9847](https://github.com/livewire/livewire/discussions/9847) also reports this in multi-tenant apps with custom middleware configurations. We investigated these cases but couldn't identify a specific middleware that would legitimately change the session ID without also breaking other session-dependent features. These may stem from incorrect middleware setup (e.g., a custom update route missing the `web` group, which Livewire now auto-adds).

## The problem

The upload path signature includes the session ID in the HMAC:

```php
hash_hmac('sha256', $path, app('encrypter')->getKey() . session()->getId())
```

1. The file is uploaded to the **upload endpoint**, which signs the temp path using session ID **A**.
2. The upload is finalized on the **update endpoint** — session ID is now **B**.
3. The signature is recomputed with session ID **B** — it doesn't match — Livewire aborts with 403.

## The solution

Remove the session ID from the upload path signature. The app key alone prevents forgery — this is the same approach used by Livewire's snapshot checksums.

The session ID was meant to scope tokens per user, but this is unnecessary:

- Temp file paths are 40-character random strings — unguessable across users.
- Finalizing an upload already requires a valid CSRF token and snapshot checksum.
- If an attacker can intercept signed paths, they already have the victim's session.

Fixes #9847

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -263,13 +263,9 @@ class TemporaryUploadedFile extends UploadedFile
         return new static($filePath, FileUploadConfiguration::disk());
     }
 
-    /**
-     * Generate a short token for a given path using the app key and session ID.
-     * This ensures tokens are unique per session and cannot be forged without the app key.
-     */
     protected static function generateToken(string $path): string
     {
-        return substr(hash_hmac('sha256', $path, app('encrypter')->getKey() . session()->getId()), 0, 8);
+        return substr(hash_hmac('sha256', $path, app('encrypter')->getKey()), 0, 8);
     }
 
     public static function signPath(string $path): string


ganyicz:
## Independent Claim Verification

I verified all claims in this PR against the codebase and external sources (CVE databases, GitHub advisories, discussion #9847).

### Core Claims (The Bug)

| # | Claim | Verdict |
|---|-------|---------|
| 1 | File uploads use two separate requests (upload endpoint signs path, update endpoint verifies) | **Verified** — `FileUploadController.php:52` signs, `WithFileUploads.php:36` verifies |
| 2 | Persistent middleware only runs on the update endpoint | **Verified** — `PersistentMiddleware.php:41-48` is gated by `isLivewireRoute()`, upload endpoint has its own middleware stack |
| 3 | Session ID mismatch between the two requests causes 403 | **Verified** — discussion #9847 confirms this exact bug |
| 4 | The code change is correct and minimal | **Verified** — removes `session()->getId()` from the HMAC key, nothing else |

### Security Claims

| # | Claim | Verdict |
|---|-------|---------|
| 5 | Snapshot checksums already use app key only (no session ID) | **Verified** — `Checksum.php:82` uses only `app('encrypter')->getKey()` |
| 6 | Temp file paths are 40-character random strings | **Verified** — `TemporaryUploadedFile.php:197` uses `str()->random(40)` (30 for S3) |
| 7 | `_finishUpload` requires valid CSRF token and snapshot checksum | **Verified** — goes through update endpoint with `web` middleware + `Checksum::verify()` |
| 8 | CVE-2025-54068 (Livepyre) is unrelated to file uploads | **Verified** — it's about property update hydration (CWE-94), not file upload signing |
| 9 | CVE-2024-47823 is about MIME type spoofing, not path signing | **Verified** — NVD confirms it's about file extension guessing from MIME type |
| 10 | PR #9681 attack scenario: forging paths requires app key, guessing paths infeasible | **Verified** — HMAC requires app key, filenames are 62^40 combinations |
| 11 | Leaked APP_KEY already gives full RCE via snapshot forgery | **Verified** — `Checksum::generate()` uses only app key |

### Deep Security Analysis: Is the session ID truly redundant?

I traced every attack vector to confirm removing the session ID doesn't weaken security:

**1. Forging a signed path** — Requires `APP_KEY` regardless of whether session ID is included. If `APP_KEY` is leaked, attacker already has full RCE via snapshot checksum forgery.

**2. Cross-user signed path reuse** (the one scenario session ID theoretically helps) — Intercepting another user's signed path requires MitM or XSS, which already gives the attacker the victim's session cookie and CSRF token (full access). Guessing another user's filename is infeasible (62^40 ≈ 10^71 combinations).

**3. Calling `_finishUpload` directly** — Protected by CSRF token (web middleware), snapshot checksum (requires `APP_KEY`), and HMAC path signature (requires `APP_KEY`). All independent of session ID.

**Conclusion:** The session ID is genuinely redundant. Its only real effect was causing the bug in #9847. All meaningful protections (CSRF, snapshot checksums, HMAC signing, random filenames) are independent of it.

joshhanley:
Thanks for the thorough analysis! Yeah makes sense. When I added the feature, I definitely didn't anticipate a scenario where the session might be different for every request 😅

PR #10156, opened 2026-03-30
Bump Alpine.js packages to v3.15.9

arjankapteijn:
Upgrade alpinejs and related @alpinejs/* dependencies from 3.15.8 to 3.15.9 in package.json and regenerate package-lock.json to capture updated transitive deps and integrity hashes.

Fix x-model.blur crash when input is removed from a form
https://github.com/alpinejs/alpine/pull/4783

Because of the ISO audit i'm running on Windows today so I'm unable to run te tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,18 +7,18 @@
             "name": "laravel-livewire",
             "license": "MIT",
             "dependencies": {
-                "@alpinejs/anchor": "^3.15.8",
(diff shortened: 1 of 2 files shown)


joshhanley:
@arjankapteijn thanks for the PR! We will be doing a release of Livewire soon with the updates, but are waiting to make sure there are no major issues in Alpine first. I'm going to close this as we will handle it as part of the release. Thanks!

PR #10159, opened 2026-03-30
[4.x] Fix blade compiler path resolution during optimize with Sentry

ganyicz:
[Merged into 3.x](https://github.com/livewire/livewire/pull/10160)

## The scenario

Running `php artisan optimize` with Sentry installed produces broken compiled templates — island paths collide and deterministic keys are wrong.

## The problem

Sentry forces early resolution of the blade engine during boot via `app('view')->getEngineResolver()->resolve('blade')`, which captures the current blade compiler instance. When `php artisan optimize` runs:

1. **`config:cache`** — swaps the application instance in the container.
2. **`view:cache`** — compiles views using the original compiler captured by the engine in step 0.

Inside compilation callbacks, `Blade::getPath()` and `app('blade.compiler')->getPath()` resolve a *new* compiler from the swapped container — one that was never given a path. The original compiler (the one actually compiling) has the path set, but nobody asks it.

## The solution

Capture the blade compiler reference once at registration time and thread it into compilation callbacks:

```php
$compiler = app('blade.compiler');

$compiler->prepareStringsForCompilationUsing(function ($content) use ($compiler) {
    $pathSignature = $compiler->getPath() ?: crc32($content);
    // ...
});
```

Applied to four files:
- `SupportIslands` — `Blade::getPath()` in `prepareStringsForCompilationUsing` callback
- `ExtendBlade` — `app('blade.compiler')` in `precompiler` callback
- `SupportScriptsAndAssets` — `app('blade.compiler')->getPath()` in directive callbacks
- `SupportMorphAwareBladeCompilation` — `app('blade.compiler')` in `precompiler` callback

Includes a regression test that forgets the compiler from the container before compilation and asserts it's never re-resolved.

Same fix as livewire/blaze#88.

Fixes livewire/blaze#169

### DIFF
--- a/src/Features/SupportIslands/SupportIslands.php
+++ b/src/Features/SupportIslands/SupportIslands.php
@@ -23,11 +23,13 @@ class SupportIslands extends ComponentHook
 
     public static function registerInlineIslandPrecompiler()
     {
-        Blade::prepareStringsForCompilationUsing(function ($content) {
+        $compiler = app('blade.compiler');
+
+        $compiler->prepareStringsForCompilationUsing(function ($content) use ($compiler) {
             // Shortcut out if there are no islands in the content...
             if (! str_contains($content, '@endisland')) return $content;
 
-            $pathSignature = Blade::getPath() ?: crc32($content);
+            $pathSignature = $compiler->getPath() ?: crc32($content);
 
             return IslandCompiler::compile($pathSignature, $content);
         });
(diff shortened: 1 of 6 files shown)


ziming:
Just curious, does this affect v3 as well?

ganyicz:
Hi @ziming,

great point. It does affect Livewire v3 as well, although the impact is much lower.

I'll submit a PR for 3.x as well anyway.

joshhanley:
Closing in favour of v3 PR #10160 which will get upmerged into v4.

PR #10165, opened 2026-03-30
[4.x] Add release tokens documentation

ganyicz:
Add a new `docs/release-tokens.md` page under the Advanced nav section covering:

- How the token is stored in snapshots and verified on each request
- The three-part token composition (internal + application + component)
- Changing `config('livewire.release_token')` on deploy
- Automating via `env('RELEASE_TOKEN')`
- Per-component tokens via the `releaseToken()` method override


joshhanley:
Might be worth adding a note that release tokens are in every request, so it's best to keep them as short as possible.

ganyicz:
Good point thanks!

PR #10172, opened 2026-04-01
Document wire:model bracket notation support

ganyicz:
## The scenario

`wire:model` in v4 supports bracket notation (`foo['bar']`, `items[0]`) for accessing nested properties, but this isn't documented anywhere. 

## The problem

Users upgrading from v3 who happen to have square brackets in their property key names hit unexpected behavior because `[]` is now parsed as an accessor.

## The solution

**`docs/wire-model.md`** — Added an "Accessing nested properties" section documenting both dot notation and bracket notation:

```blade
<input type="text" wire:model="address.city">
<input type="text" wire:model="address['city']">
<input type="text" wire:model="items[0].name">
```

Also expanded the Reference section to show all supported expression syntaxes.

**`docs/upgrading.md`** — Added a note under "Low-impact changes" explaining that `wire:model` now interprets square brackets as property accessors (breaking change from v3 where they were literal characters).

Fixes #10170

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -312,6 +312,20 @@ This change adds support for passing slots when mounting components and generall
 
 These changes only affect applications using advanced features or customization.
 
+### `wire:model` now supports bracket notation
+
+`wire:model` expressions now support bracket notation for accessing nested properties:
+
+```blade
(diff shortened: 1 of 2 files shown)


calebporzio:
great

PR #10180, opened 2026-04-02
Fix $refs access in JS actions called from PHP

ganyicz:
## The scenario

JS actions defined with `$js()` can't access `this.$refs` when called from PHP via `$this->js('actionName')`, even though the same action works fine when called directly from Blade via `wire:click="$js.actionName"`.

```blade
<script>
    $js("logRef", function () {
        console.log(this.$refs.myRef); // undefined when called from PHP
    });
</script>
```

## The problem

There are two paths for calling a JS action, and they handle `this` binding differently:

1. **From Blade** — the action is retrieved through `getJsAction()`, which binds `this` to `$wire`
2. **From PHP** — the action is retrieved through `getJsActions()`, which returned raw unbound functions

## The solution

Changed `getJsActions()` to return bound actions using `getJsAction()` internally, making both paths consistent.

Fixes livewire/livewire#10179

## Considerations

- **Breaking change risk**: the previous `this` binding was an Alpine internal proxy with no useful Livewire magics — no one could have relied on it intentionally
- **Extraction**: the binding logic lives in `getJsAction()` as the single source of truth; `getJsActions()` delegates to it rather than duplicating
- **`getJsActions()` behavior change**: only two source-level callers exist, both already wanted bound actions
- **`getJsAction()` still used**: the async script loading path in `$wire.js` needs lazy per-action binding, so the singular method is still needed

### DIFF
--- a/js/$wire.js
+++ b/js/$wire.js
@@ -157,7 +157,7 @@ wireProperty('$js', (component) => {
     let jsActions = component.getJsActions()
 
     Object.keys(jsActions).forEach((name) => {
-        fn[name] = component.getJsAction(name)
+        fn[name] = jsActions[name]
     })
 
     return new Proxy(fn, {
--- a/js/component.js
+++ b/js/component.js
@@ -295,7 +295,11 @@ export class Component {
     }
 
     getJsActions() {
-        return this.jsActions
+        let actions = {}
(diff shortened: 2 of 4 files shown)


calebporzio:
perfect thanks

PR #10189, opened 2026-04-03
Add configurable Livewire back button cache

prateekbhujel:
I’m adding a `back_button_cache` config option to Livewire.

Livewire currently disables the browser back-forward cache on pages with Livewire components. I want an opt-in way to change that behavior for apps that still want normal browser caching on back and forward navigation.

Why I think this is useful:
- I can enable bfcache without writing custom code in every component
- I still keep the existing per-component `disableBackButtonCache()` override when I need it
- it gives apps that are not fully using `wire:navigate` a simple config-based option

What I changed:
- added `back_button_cache` to `config/livewire.php` with a default of `false`
- updated `SupportDisablingBackButtonCache` to read that config value
- kept per-component overrides working
- added unit tests for the config default, the override, and the multi-component case
- documented the option in the navigate docs

I also fixed the request lifecycle behavior so one component cannot accidentally undo another component’s cache choice on the same page.

Tests:
- `./vendor/bin/phpunit --configuration phpunit.xml.dist --testsuite Unit --filter SupportDisablingBackButtonCache`

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -19,6 +19,19 @@ return [
     ],
 
     /*
+    |--------------------------------------------------------------------------
+    | Back Button Cache
+    |--------------------------------------------------------------------------
+    |
+    | Livewire normally disables the browser back/forward cache on pages with
+    | components so the browser always performs a full reload. Set this to true
+    | to opt into bfcache by default while still allowing per-component overrides.
+    |
+    */
+
+    'back_button_cache' => false,
+
+    /*
     |---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


ganyicz:
Hi @prateekbhujel, thanks for contributing!

there's already a PR open for the exact same feature: https://github.com/livewire/livewire/pull/10151

Feel free to leave your feedback in the comments

prateekbhujel:
Thanks @ganyicz. I see the overlap now, so I’ll end this one here. Appreciate it.

PR #10195, opened 2026-04-05
Skip entire lifecycle for unchanged reactive children

calebporzio:
## Summary

When a parent component re-renders, all children with `#[Reactive]` props are bundled into the request. Previously, every child went through the full server lifecycle (hydrate → render → dehydrate) even when its props didn't change.

This PR skips the **entire lifecycle** for reactive children whose props are unchanged — no hydrate, no render, no dehydrate, no snapshot in the response. The server returns a lightweight `{skip: true, id}` marker instead.

### Before
```
Request:  [parent, childA, childB, childC]  →  4 components processed
Response: [parent, childA, childB, childC]  →  4 full snapshots returned
```

### After (when no props changed)
```
Request:  [parent, childA, childB, childC]  →  1 component processed, 3 skipped
Response: [parent, {skip}, {skip}, {skip}]  →  1 full snapshot, 3 lightweight markers
```

## Changes

**`SupportReactiveProps.php`** — New `shouldSkipUpdate($snapshot)` method that checks:
1. Component has `#[Reactive]` properties (`memo.props`)
2. Parent already rendered and stored pending params (`$pendingChildParams`)
3. Component doesn't also have `wire:model` bindings (`memo.bindings`)
4. Every reactive prop value is unchanged (CRC32 hash comparison, same as `BaseReactive::hydrate()`)

**`HandleRequests.php`** — Before calling `update()` for each component, checks `shouldSkipUpdate()`. If true, returns `{skip: true, id}` and skips the entire `update()` call.

**`js/request/index.js`** — Handles `skip` markers in the response by resolving the message's action promises and finishing its lifecycle without updating the component's snapshot or morphing DOM.

## What gets skipped

| Layer | Before | After |
|-------|--------|-------|
| Server hydrate | Always | Skipped |
| Server render (Blade) | Always | Skipped |
| Server dehydrate | Always | Skipped |
| Response payload | Full snapshot + effects | `{skip: true, id}` |
| JS snapshot merge | Always | Skipped |
| JS DOM morph | Always | Skipped |

## Edge cases handled

- **`wire:model` children** — Never skipped (they have `memo.bindings`, excluded by check)
- **Components with JS-side dirty data** — Never skipped (`empty($updates)` check in HandleRequests)
- **Components with real method calls** — Processed normally (only `$commit`-only children are eligible)
- **Non-reactive components** — Never skipped (no `memo.props`)

## Test plan

- [x] `shouldSkipUpdate()` returns true when props unchanged
- [x] `shouldSkipUpdate()` returns false when props changed
- [x] `shouldSkipUpdate()` returns false without pending params
- [x] `shouldSkipUpdate()` returns false for non-reactive components
- [x] Integration test: bundled request returns `{skip}` for unchanged child
- [x] All 70 related tests pass (ReactiveProps + WireModeling + Nesting + HandleRequests)
- [ ] Browser tests


calebporzio:
## Note on Eloquent model prop comparison

The skip optimization compares reactive prop values to decide if a child can be skipped. For scalar/array props this is straightforward (CRC32 hash of json_encode). For Eloquent models it's more nuanced.

**Current approach:** Compare model class + primary key + `isDirty()` against the dehydrated snapshot form.

- Same model, unmodified → **skip** (class matches, key matches, not dirty)
- Same model, parent did `$model->name = 'New'` → **don't skip** (isDirty = true)
- Different model instance (different key) → **don't skip**

**The debate:** There are a few perspectives on what "unchanged" means for a model prop:

1. **Identity-based (current):** Same class + same key + not dirty = same model. This is the most permissive — it skips whenever the model _identity_ hasn't changed and no attributes were locally modified. But if the database row was updated externally between requests, the child still skips (it has the stale snapshot anyway — same as non-optimized behavior since Livewire hydrates models fresh from DB on both parent and child).

2. **Always re-render models:** The safest option — never skip children with model props. Just return `false` from `valuesMatch()` when the pending value is a Model. Simple, zero risk, but misses the optimization for a very common use case.

3. **Full attribute comparison:** Dehydrate the pending model using Livewire's synth system and compare the full serialized form. Most accurate, but requires synth infrastructure that isn't available pre-hydration (no component context). Would need a different architecture to support.

Open to input on which direction feels right. The current approach handles the common case (parent re-renders, passes the same model it had before) while staying safe for the "parent modified the model" case via `isDirty()`.

js/request/message.js
```diff
@@ -110,6 +116,22 @@ export default class Message {
...
     }
 
+    /**
+     * Mark this message as skipped. Used when the server returns a skip
+     * payload (e.g. for an unchanged reactive child) — or, in the future,
+     * when the JS side decides client-side that nothing needs to be sent
+     * for this component. After marking, the caller is responsible for
+     * firing `invokeOnSkipped()` and `invokeOnFinish()` to drive the
+     * lifecycle to completion.
+     */
```
calebporzio: remove comment

PR #10198, opened 2026-04-06
Add chunked file uploads

calebporzio:
PHP file uploads have hard ceilings — `upload_max_filesize`, `post_max_size`, `memory_limit`, `max_execution_time` — that make uploading large files (videos, datasets, archives) painful or impossible without per-server tuning. Today, a Livewire user wanting to upload a 2GB video has no good answer beyond "edit php.ini and pray your timeouts hold."

```php
new class extends Component {
    use WithFileUploads;

    public $video; // Today: dies on memory/timeout for large files

    public function save()
    {
        $this->video->store('videos');
    }
};
```

## Findings

- The existing single-request upload path can't handle files larger than `upload_max_filesize` (default 2MB) or `post_max_size` (default 8MB).
- Slow connections hit `max_execution_time` regardless of memory headroom.
- Failed uploads have no resume — the user starts over from 0%.
- Direct-to-S3 already exists but requires S3 (not everyone uses it) and still hits timeouts on slow connections without multipart uploads.
- Pulling in `tus-js-client` + `tus-php` would add ~8k lines of dependency code plus an external project's release schedule for what Livewire can do in ~600 production lines.

## Potential solutions

- **Tell users to bump PHP limits.** Doesn't scale past a few hundred MB. Doesn't help with timeouts. No resume.
- **Force users onto S3 + the existing direct-to-S3 path.** Doesn't fit every stack. Still hits timeouts on slow connections without multipart.
- **Adopt tus-js-client + tus-php as dependencies.** Adds ~8k lines + external dependency surface for features Livewire doesn't need.
- **Bespoke chunked upload support inspired by tus.** ~600 lines, no new dependencies, fits the existing `WithFileUploads` API surface exactly.

## Proposed solution

A bespoke chunked upload path inspired by the [tus protocol](https://tus.io). Large files are split into chunks and uploaded sequentially via PATCH requests with byte offsets. **The developer-facing API is unchanged** — same `WithFileUploads` trait, same `TemporaryUploadedFile`, same `wire:model`, same `livewire-upload-*` events. Chunking happens automatically when files exceed `chunk_size`. Smaller files continue to use the existing single-request path.

Opt-in by default. Set `chunk_size` in `config/livewire.php` to enable:

```php
'temporary_file_upload' => [
    // ...
    'chunk_size' => 5 * 1024 * 1024, // 5MB per chunk
],
```

### Flow

```
1. JS: POST /chunk-upload (Upload-Length, Upload-Name [base64])
   PHP: validates size, returns { transferId, patchUrl, offsetUrl }

2. JS: PATCH /chunk-upload/{id} (Upload-Offset: 0, body: chunk 1)
   PHP: 204, Upload-Offset: 2097152

3. JS: PATCH /chunk-upload/{id} (Upload-Offset: 2097152, body: chunk 2)
   PHP: 204, Upload-Offset: 4194304

   ... repeat for each chunk ...

4. JS: PATCH /chunk-upload/{id} (final chunk)
   PHP: validates assembled file against rules, returns
        204, Upload-Complete: true, X-Signed-Path: <signed-tmp-path>
        (or 422 with validation errors via ValidationException)

5. JS: calls existing _finishUpload(name, [signedPath]) — rejoins existing flow
   PHP: $this->video is now a TemporaryUploadedFile
```

If a chunk fails: client retries with backoff (`chunk_retry_delays`), querying `GET /chunk-upload/{id}/offset` for the server's authoritative offset before each retry.

### Security

| Threat | Mitigation |
|--------|------------|
| Validation bypass for large files | Assembled file is run through `temporary_file_upload.rules` at finalize time. Throws `ValidationException` on failure → 422 JSON |
| Storage exhaustion via huge `Upload-Length` claims | Init endpoint enforces `max:` from rules, falling back to `chunk_absolute_max_bytes` (5GB) if no `max:` is set |
| Storage exhaustion via abandoned chunks | 24-hour cleanup extended to chunk directories |
| Per-chunk size abuse | Chunks larger than `chunk_size` rejected with 413 |
| Total bytes abuse | Chunks pushing the running total past declared size rejected with 413 |
| Long-lived signed URL leakage | Signed URLs use `chunk_max_upload_time` (60 min default), not 24 hours |
| Path traversal via transfer ID | Validated against `^[A-Za-z0-9]{40}$` on every PATCH/GET |
| Path traversal via filename extension | Extension is sanitized to alphanumeric in finalize |
| Race conditions on manifest | `flock(LOCK_EX)` held through chunk write AND finalization |
| CSRF | Inherits the existing `web` middleware group with X-CSRF-TOKEN headers from JS |

### Limitations

- **Local disk only.** S3 falls back to the existing single-request path. Native S3 multipart support is a future PR — it maps cleanly to `CreateMultipartUpload` / `UploadPart` / `CompleteMultipartUpload` and would let chunks bypass PHP entirely via presigned URLs. R2, Spaces, B2, MinIO would all benefit since they implement the S3 API.
- **No cross-page-load resume.** Chunks resume after network failures within a single session, but a page refresh restarts. Adding fingerprint-based persistence is straightforward but out of scope.
- **PATCH method.** Some WAFs block PATCH by default. Either configure your infra to allow PATCH on the chunk endpoint or leave `chunk_size` at `null`.

### Test plan

32 new end-to-end tests in `ChunkedUploadTest.php`. Tests exercise the real HTTP endpoints, not internals:

- Init: signature validation, manifest creation, transfer ID format, size enforcement (both `max:` rule and `chunk_absolute_max_bytes` fallback), zero/negative size rejection
- PATCH: offset matching, 409 on mismatch, 404 on unknown transfer, transfer ID format validation, per-chunk size limit, total-bytes-claimed enforcement
- Offset endpoint: signature validation, returns current progress, `Cache-Control: no-store`
- Full upload flow: 3KB / 3 chunks → byte-for-byte assembled file → chunk dir cleaned up
- **Closed-loop test**: walk chunks through HTTP → feed signed path to `_finishUpload` → verify component property is a real `TemporaryUploadedFile` → verify `->store()` works
- Validation at finalize: mime mismatch returns 422, file is cleaned up on failure
- Resume flow: send chunk → query offset → resume → complete
- Filename sanitization: weird extensions are stripped to alphanumeric
- Unicode filenames decoded correctly from base64 `Upload-Name`
- Multiple concurrent uploads to different transfer IDs don't interfere
- `_startUpload` correctly skips chunking for small files, S3 disks, and `chunk_size: null`
- `chunk_middleware` config is applied to chunk routes
- `maxUploadSizeInBytes()` correctly parses from rules

**93 file upload tests passing locally** (32 new + 60 existing + 1 always-skipped). The single failing browser test (`test_can_upload_preview_and_save_a_file`) fails on `main` too — it's a pre-existing flaky Chromedriver test, unrelated to this PR.

Manual end-to-end browser verification: 5MB file → 3 PATCH chunks → component property populated → `->store()` works. 600MB file with `max:512000` rule → 413 at init endpoint → user-visible "file failed to upload" error.

---

Marked draft because it's a new feature surface that warrants discussion before merge — specifically:
1. Whether `chunk_size` should default to `null` (current, opt-in) or to a sensible value like 5MB (opt-out)
2. Whether the silent S3 fallback is the right call vs erroring loudly when chunked uploads are configured but the disk is S3
3. Whether `chunk_max_upload_time` of 60 minutes is the right default

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -140,6 +140,10 @@ return [
         ],
         'max_upload_time' => 5, // Max duration (in minutes) before an upload is invalidated...
         'cleanup' => true, // Should cleanup temporary uploads older than 24 hrs...
+        'chunk_size' => 5 * 1024 * 1024, // 5MB per chunk. Set to `null` to disable chunked uploads.
+        'chunk_retries' => 3, // Number of times to retry a failed chunk before giving up.
+        'chunk_retry_delays' => [500, 1000, 3000], // Backoff delays (in ms) between chunk retries.
+        'chunk_resumable' => true, // Allow resuming uploads across page loads.
     ],
 
     /*
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -142,7 +142,7 @@ return [
         'cleanup' => true, // Should cleanup temporary uploads older than 24 hrs...
         'chunk' => [
             'enabled' => false,                                   // Opt in to chunked uploads. When true, files larger than `size` are uploaded in chunks instead of a single request. Local disk only — S3 not yet supported.
-            'size' => 5 * 1024 * 1024,                            // Bytes per chunk. Files larger than this are split and uploaded sequentially. Defaults to 5MB.
+            'size' => 1024 * 1024,                                // Bytes per chunk. Defaults to 1MB to fit under nginx's default `client_max_body_size` on both Forge and Laravel Cloud — works out of the box. Bump alongside your server's body limit for better throughput on larger files.
             'retry_delays' => [500, 1000, 3000],                  // Backoff delays (in ms) between chunk retries. The number of entries determines the max retries.
             'max_upload_time' => 60 * 24,                         // Max duration (in minutes) for an entire chunked upload to complete. Used as the signed-URL expiry for chunk endpoints. Defaults to 24 hours to handle large uploads on slow connections.
             'middleware' => null,                                 // Middleware applied to chunk endpoints. Defaults to 'throttle:600,1' (looser than the global upload throttle since chunked uploads make many requests). Sites that expose uploads to anonymous visitors should tighten this — see the security note in uploads.md.
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -140,10 +140,10 @@ return [
         ],
         'max_upload_time' => 5, // Max duration (in minutes) before an upload is invalidated...
         'cleanup' => true, // Should cleanup temporary uploads older than 24 hrs...
-        'chunk_size' => 5 * 1024 * 1024, // 5MB per chunk. Set to `null` to disable chunked uploads.
-        'chunk_retries' => 3, // Number of times to retry a failed chunk before giving up.
-        'chunk_retry_delays' => [500, 1000, 3000], // Backoff delays (in ms) between chunk retries.
-        'chunk_resumable' => true, // Allow resuming uploads across page loads.
+        'chunk_size' => null, // Bytes per chunk. Set to a number (e.g., 5 * 1024 * 1024) to enable chunked uploads for files larger than the chunk size. Local disk only — S3 not yet supported.
+        'chunk_retry_delays' => [500, 1000, 3000], // Backoff delays (in ms) between chunk retries. The number of entries determines the max retries.
+        'chunk_max_upload_time' => 60, // Max duration (in minutes) for an entire chunked upload to complete. Used as the signed-URL expiry for chunk endpoints.
+        'chunk_middleware' => null, // Middleware applied to chunk endpoints. Defaults to 'throttle:600,1' (looser than the global upload throttle since chunked uploads make many requests).
     ],
 
     /*
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -140,11 +140,13 @@ return [
         ],
         'max_upload_time' => 5, // Max duration (in minutes) before an upload is invalidated...
         'cleanup' => true, // Should cleanup temporary uploads older than 24 hrs...
-        'chunk_size' => null, // Bytes per chunk. Set to a number (e.g., 5 * 1024 * 1024) to enable chunked uploads for files larger than the chunk size. Local disk only — S3 not yet supported.
-        'chunk_retry_delays' => [500, 1000, 3000], // Backoff delays (in ms) between chunk retries. The number of entries determines the max retries.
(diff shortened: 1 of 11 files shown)


joshhanley:
CloudFlare also has a 100mb request body limit, which also impacts file uploads, so benefits from chunked file uploads too!

Re the questions, my thoughts are:
1. Opt-in
2. Loud
3. 60 seems good. If you have a big file, like uploading a video, then having extra time do that by default would be useful.

calebporzio:
Thanks Josh! Going with all three of your answers:

1. **Opt-in** — already the default (`chunk_size: null`)
2. **Loud** — pushed [21bc8a3](https://github.com/livewire/livewire/commit/21bc8a38). Setting `chunk_size` while on the S3 disk now throws `S3DoesntSupportChunkedUploads` at upload time with a clear message telling you to either null out chunk_size or switch disks.
3. **60 minutes** — keeping the default

Also added a note in the docs about Cloudflare's 100MB body limit as another reason chunked uploads matter.

Marking this ready for review.

config/livewire.php
```diff
@@ -140,6 +140,10 @@
         ],
         'max_upload_time' => 5, // Max duration (in minutes) before an upload is invalidated...
         'cleanup' => true, // Should cleanup temporary uploads older than 24 hrs...
+        'chunk_size' => null, // Bytes per chunk. Set to a number (e.g., 5 * 1024 * 1024) to enable chunked uploads for files larger than the chunk size. Local disk only — S3 not yet supported.
+        'chunk_retry_delays' => [500, 1000, 3000], // Backoff delays (in ms) between chunk retries. The number of entries determines the max retries.
+        'chunk_max_upload_time' => 60, // Max duration (in minutes) for an entire chunked upload to complete. Used as the signed-URL expiry for chunk endpoints.
+        'chunk_middleware' => null, // Middleware applied to chunk endpoints. Defaults to 'throttle:600,1' (looser than the global upload throttle since chunked uploads make many requests).
```
joshhanley: Should this maybe be an array? Like:


```suggestion
        'chunk` => [
            'size' => null, // Bytes per chunk. Set to a number (e.g., 5 * 1024 * 1024) to enable chunked uploads for files larger than the chunk size. Local disk only — S3 not yet supported.
            'retry_delays' => [500, 1000, 3000], // Backoff delays (in ms) between chunk retries. The number of entries determines the max retries.
            'max_upload_time' => 60, // Max duration (in minutes) for an entire chunked upload to complete. Used as the signed-URL expiry for chunk endpoints.
            'middleware' => null, // Middleware applied to chunk endpoints. Defaults to 'throttle:600,1' (looser than the global upload throttle since chunked uploads make many requests).
        ]
```

joshhanley (approved):
Couple small things, but otherwise looks good!

I do wonder if 5gb is too big of a default limit?

PR #10201, opened 2026-04-07
[4.x] Fix ES import statements in multi-file component JS files

joshhanley:
# The Scenario

When using ES `import` statements in a multi-file component's `.js` file, the browser throws a syntax error:

```
Uncaught (in promise) SyntaxError: Unexpected token '{'
```

For example, a multi-file component with a `.js` file like this fails:

```js
import { OBJLoader } from "three/examples/jsm/Addons.js"

// Use OBJLoader...
```

Single-file components with `<script>` blocks handle imports correctly, but multi-file components do not.

# The Problem

When Livewire compiles a multi-file component's `.js` file, `MultiFileParser::generateScriptContents()` wraps the entire file contents inside an `export function run()` wrapper:

```js
export function run($wire, $js) {
    import { OBJLoader } from "three/examples/jsm/Addons.js"
    // ...
}
```

ES `import` statements are only valid at the top level of a module, not inside function bodies. This causes the browser to throw a `SyntaxError` when parsing the module.

The `SingleFileParser` already handles this correctly by hoisting `import` statements above the wrapper, but this logic was missing from the `MultiFileParser`.

# The Solution

The same import-hoisting logic has been applied to `MultiFileParser::generateScriptContents()`. Import statements are now extracted from the user's code and placed at the top of the module, with only the remaining code wrapped inside the `export function run()` body:

```js
import { OBJLoader } from "three/examples/jsm/Addons.js"

export function run($wire, $js) {
    // ...
}
```

### Tests

**Unit tests** (`src/Compiler/UnitTest.php`):

| Test | SFC/MFC |
|---|---|
| `test_script_hoists_imports_and_wraps_in_export_function` | SFC (existing) |
| `test_script_wraps_in_export_function_even_without_imports` | SFC (existing) |
| `test_mfc_script_wraps_in_export_function_even_without_imports` | MFC (new) |
| `test_mfc_script_hoists_imports_and_wraps_in_export_function` | MFC (new) |

**Browser tests** (`src/Features/SupportJsModules/BrowserTest.php`):

| Test | SFC/MFC |
|---|---|
| `test_nested_namespaced_component_loads_js_module` | SFC (existing) |
| `test_single_file_component_js_supports_es_imports` | SFC (new) |
| `test_multi_file_component_js_supports_es_imports` | MFC (new) |

Fixes https://github.com/livewire/livewire/discussions/10163

### DIFF
--- a/src/Compiler/Fixtures/mfc-component-with-imports/mfc-component-with-imports.blade.php
+++ b/src/Compiler/Fixtures/mfc-component-with-imports/mfc-component-with-imports.blade.php
@@ -1,0 +1,1 @@
+<div>{{ $message }}</div>
--- a/src/Compiler/Fixtures/mfc-component-with-imports/mfc-component-with-imports.js
+++ b/src/Compiler/Fixtures/mfc-component-with-imports/mfc-component-with-imports.js
@@ -1,0 +1,9 @@
+import { Alpine } from 'alpinejs'
+import { debounce } from './utils'
+
+console.log('Component initialized');
+Alpine.data('myComponent', () => ({
+    init() {
+        debounce(() => console.log('Debounced'), 100);
+    }
+}));
--- a/src/Compiler/Fixtures/mfc-component-with-imports/mfc-component-with-imports.php
+++ b/src/Compiler/Fixtures/mfc-component-with-imports/mfc-component-with-imports.php
@@ -1,0 +1,9 @@
+<?php
+
+use Livewire\Component;
+
+new class extends Component
+{
+    public $message = 'Hello World';
+};
+?>
--- a/src/Compiler/Parser/MultiFileParser.php
+++ b/src/Compiler/Parser/MultiFileParser.php
@@ -101,7 +101,25 @@ class MultiFileParser extends Parser
 
         $scriptContents = trim($this->scriptPortion);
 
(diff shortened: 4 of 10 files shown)


ghabriel25:
Finally, thanks @joshhanley . I'll test this in my local project

joshhanley:
@ghabriel25 no worries! That'd be great, thanks 😁

ghabriel25:
Hi, @joshhanley 

I try this on fresh Laravel + Livewire but there was an error (status 500) about Livewire scripts not found
```blade
<!-- Livewire Scripts -->
<script src="http://myapp.test/livewire-9ee34a36/livewire.js?id=dev" 
    data-csrf="A0rGG8roo31ntF7u5eWfyFEKvg6OjctKxAuaOk4l" 
    data-module-url="http://myapp.test/livewire-9ee34a36" 
    data-update-uri="http://myapp.test/livewire-9ee34a36/update" 
    data-navigate-once="true"></script>
```

Then I try manually bundling Livewire and Alpine but there is no `dist` folder in your branch.

joshhanley:
@ghabriel25 we don't commit dist files anymore, so you will need to pull Livewire branch, `npm install` and `npm run build` to build Livewire's assets.

PR #10204, opened 2026-04-08
Add TypeScript types

aguingand:
The goal of this PR is to add TypeScript types for JS exports of Livewire to:
- Give additional context to AI
- Enhance IDE support
- Give clear information of deprecated features (e.g. 'request' & 'commit' hooks)

Content:
- add `js/index.d.ts` : Types of exported `Livewire` methods
- update of `scripts/build.js` : Export the type file in `dist` to be in the composer package files (global types referencing) and to allow typing on ESM import

Potential downsides:
- Types must be updated on public JS API change. This does not seem to be a great deal especially with agents.

Potential improvements:
- Add definition for `$wire` magic variable.

### DIFF
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -1,0 +1,185 @@
+import type { Directive } from './directives'
+
+export interface Livewire {
+    // Register a custom Livewire directive
+    directive: (
+        name: string,
+        callback: (ctx: {
+            el: HTMLElement
+            directive: Directive
+            component: Component
+            $wire: any
+            cleanup: (cb: () => void) => void
+        }) => void
+    ) => void
+
+    // Dispatch an event to components by name
+    dispatchTo: (componentName: string, name: string, params?: any) => void
+
+    // Interception hooks from request layer
+    interceptAction: (
+        callback: (ctx: {
+            action: any
+            onSend: (cb: (ctx: { call: any }) => void) => void
+            onCancel: (cb: () => void) => void
+            onSuccess: (cb: (ctx: { result: any }) => void) => void
+            onError: (cb: (ctx: { response: Response; body: string; preventDefault: () => void }) => void) => void
+            onFailure: (cb: (ctx: { error: any }) => void) => void
+            onFinish: (cb: () => void) => void
+        }) => void
+    ) => () => void
+
+    interceptMessage: (
+        callback: (ctx: {
+            message: any
+            cancel: () => void
+            onSend: (cb: (ctx: { payload: any }) => void) => void
+            onCancel: (cb: () => void) => void
+            onFailure: (cb: (ctx: { error: any }) => void) => void
+            onError: (cb: (ctx: { response: Response; body: string; preventDefault: () => void }) => void) => void
+            onStream: (cb: (ctx: { json: any }) => void) => void
+            onSuccess: (cb: (ctx: {
+                payload: any
+                onSync: (cb: () => void) => void
+                onEffect: (cb: () => void) => void
+                onMorph: (cb: () => Promise<void> | void) => void
+                onRender: (cb: () => void) => void
+            }) => void) => void
+            onSkipped: (cb: () => void) => void
+            onFinish: (cb: () => void) => void
+        }) => void
+    ) => () => void
+
+    interceptRequest: (
+        callback: (ctx: {
+            request: any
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for this. I could see the value.

Two things I would want answered though:

A) Will my (or anyone's) IDE pick this up and use them when I'm writing Livewire JS (which is done from either script tag in SFC or .js file in MFC) -> attach a screen clip of an editor using this information

B) How will these things be kept up-to-date. If we leave it up to memory, updating them might slip through the cracks and we'll get API drift on it. We will need SOME kind of process before I'm comfortable merging this.

aguingand:
A) I've tested on PHPStorm & VSCode :

**PHPStorm** (the `vendor/livewire/livewire` directory must not be excluded from indexing).
➡️ Works in SFC + MFC

image

**VS Code** (the project must include a `tsconfig.json` or `jsconfig.json` file)
➡️ Works only in MFC (.js files)

image

---

B) The types can't be generated with TS (e.g. in a script), but it's possible to update Claude guidelines to also handle the `js/index.d.ts` file (if any JS  has changed). Would it be ok for you ?

calebporzio (changes requested):
Thanks for putting this together — having TypeScript types for the JS API is a nice idea.

However, there are some significant type inaccuracies that would actively mislead users. I verified these against the actual source code and confirmed at runtime:

### Wrong return types for `first()`, `find()`, `getByName()`

These are typed as returning `Component | undefined` / `Component[]`, but they actually return **`$wire` proxies**, not `Component` instances:

```js
// store.js
export function first() {
    return Object.values(components)[0].$wire  // ← returns $wire, not Component
}

export function find(id) {
    let component = components[id]
    return component && component.$wire  // ← returns $wire, not Component
}

export function getByName(name) {
    return componentsByName(name).map(i => i.$wire)  // ← returns $wire[], not Component[]
}

export function all() {
    return Object.values(components)  // ← this one IS Component[]
}
```

`$wire` is a proxy with a completely different API (`$get`, `$set`, `$toggle`, `$call`, `$refresh`, `$dispatch`, `$upload`, `$entangle`, etc.) plus it proxies arbitrary property access to server-side data/methods. Typing it as `Component` would be very confusing.

### Missing `key` property on `Component`

The `Component` constructor sets `this.key = el.getAttribute('wire:key')` but the interface doesn't include it.

### Missing lots of properties/methods on `Component`

The interface is also missing `reactive`, `originalEffects`, `queuedUpdates`, `jsActions`, and many methods like `mergeNewSnapshot`, `queueUpdate`, `processEffects`, `addJsAction`, `replayUpdate`, etc. Some of these are internal, but if you're going to type `Component`, it should be complete or clearly documented as partial.

### `$wire` is typed as `any` everywhere

The `$wire` proxy is the primary way users interact with Livewire from JS, and it's typed as `any` throughout. If the goal is "enhance IDE support," this is the most important thing to type — yet it's the one thing left untyped.

### General concern

These types would need to be manually maintained alongside the JS source, and there's no mechanism to verify they stay in sync. If they drift (which they will), they'd be worse than no types at all — giving false confidence. The Livewire JS API is still evolving (the `@todo` comment about v4 in index.js, the deprecated hooks, etc.), which makes this especially risky.

I'd want to see the return type issues fixed at minimum before considering this. The `$wire` typing gap is also a significant one — that's the API surface users actually touch most.

calebporzio:
Hey, thanks for the effort here — I appreciate the thought behind it.

After looking into this, I'm going to close it out. A few reasons:

The main JS surface users actually touch is `$wire`, and it's a dynamic proxy — `$wire.name` hits server data, `$wire.save()` fires a server method. You can't meaningfully type that without generics tied to each PHP component, which is a whole different problem. So the most important thing to type is essentially untypeable.

What's left (`hook()`, `interceptRequest()`, etc.) are niche power-user APIs. And a standalone `.d.ts` file that lives separately from the source will silently drift as the JS evolves — wrong types end up being worse than no types.

If we ever go down this road, the better approach would be JSDoc annotations inline in the actual source files so they stay in sync naturally. But it's not something I'm prioritizing right now.

Thanks again for the contribution!

aguingand:
Yeah, we can't type the `$wire` object completely. Laravel idea does it by default though.
The JSDoc approach could be enough I agree.

PR #10217, opened 2026-04-10
[4.x] Fix lazy components receiving events before mount

### LINKED ISSUE #10216: Event listeners on Lazy component broke Livewire lifecycle
### Livewire version

v4.2.4

### Laravel version

v13.4.0

### PHP version

8.4.14

Brave on Windows

## Description

When dispatching an event that is listened to by a lazy-loaded component that has not yet been initialized (e.g., not in the viewport or still loading), the listener action is executed and the component is rendered without going through the mount method first.

This leads to errors in the Blade view, since the component relies on data being initialized in mount in order to render correctly.

## Real-world use case (simplified)

In my case, I have:

- A list of items
- Each item contains a modal
- Inside each modal, there are tabs
- The second tab contains a lazy-loaded component
- This lazy component is itself a list with an event listener used to refresh its data

When one fully loaded modal dispatches the refresh event:

- Lazy components that are not yet loaded (e.g., their tab hasn’t been opened yet) still receive the event
- This triggers their listener and forces a render before they are initialized
- As a result, the component renders without mount being called, causing errors in the Blade view

In this example, I intentionally added enough scroll so that the lazy component is outside of the viewport on page load. This ensures that the component is not initialized before the event is dispatched.

A Wirebox link is provided to reproduce the issue.

https://wirebox.app/b/ywq7n

### Steps to reproduce : 
- Open the Wirebox link
- Do not scroll (the lazy component is intentionally outside of the viewport)
- Click on the "Refresh children" button

This will dispatch an event to a child component that has not yet been mounted.

### Result

The lazy component still receives the event and executes its listener, triggering a render before mount() has been called.

```php
# Parent component
namespace App\Livewire;

use Livewire\Component;

{
    public function render()
    {
        return <<<'BLADE'
            <div
                style="height: 2000px !important"
            >
                <div class="bg-red-600 mb-[1000px]">
                    <button
                        type="button"
                        class="bg-black m-2 p-2 rounded cursor-pointer"
                        wire:click="$dispatch('refresh-children')"
                    >
                        refresh children
                    </button>
                </div>
            
                <div class="bg-blue-600 mb-50px">
                    <livewire:children
                        :context="[
                            'title' => 'test',
                            'data_id' => 1,
                            'parent_id' => $this->getId(),
                        ]"
                        lazy
                    />
                </div>
            </div>
        BLADE;
    }
}
```

```php
# Child component
namespace App\Livewire;

use Livewire\Component;
use Livewire\Attributes\On;
use App\Models\SomeData;

class Children extends Component
{
    public string $title;
    public SomeData $someData;
    public string $parentId;

    public function mount(array $context): void
    {
        $dataId = $context['data_id'] ?? null;
        if (!$dataId) {
            throw new Exception('Data Id is required');
        }
        
        $this->title = $context['title'] ?? '';
        $this->someData = SomeData::findOrFail($dataId);
        $this->parentId = $context['parent_id'] ?? '';
    }

    #[On('refresh-children')]
    public function setData(): void
    {
        //
    }
    
    public function render()
    {
        return <<<'BLADE'
            <div class="flex w-full h-full flex-col items-center justify-center text-gray-300 px-16">
                <h1>Title : {{ $title }}</h1>
                <div>Some Data Id : {{ $someData->id }}</div>
                <div>Parent Id : {{ $parentId }}</div>
            </div>
        BLADE;
    }
}
```

Not necessary, the issue is fully reproducible via the provided Wirebox link and code snippet.

If needed, I will add them after opening the issue.

Lazy components that have not been initialized yet should either:

- Not react to events until they are mounted, or
- Ensure that mount is called before any render triggered by an event listener

Alternative proposal :

Another possible approach would be to allow dynamic event listeners on lazy components, but only register them after the component has been mounted.

But same issue if this can't be possible :

- Lazy components should not start listening to events until after mount has been executed
- This would prevent them from reacting prematurely and avoid inconsistent or broken state during rendering

joshhanley:
# The Scenario

When a lazy-loaded component has an event listener (via `#[On]` or `$listeners`) and hasn't been mounted yet (e.g., it's outside the viewport), dispatching that event causes the component to receive it and render without `mount()` ever being called. This leads to errors because properties initialised in `mount()` are uninitialised.

A common real-world case is a page with a list of items, each containing a modal with lazy-loaded tab content. When one item dispatches a refresh event, lazy components in other items that haven't been opened yet receive the event and break.

```php
<?php

use Livewire\Component;
use Livewire\Attributes\On;

// Parent component
new class extends Component {
    public function render()
    {
        return <<<'HTML'
        <div>
            <button wire:click="$dispatch('refresh-child')">Refresh</button>

            <div style="height: 200vh"></div>

            <livewire:child lazy />
        </div>
        HTML;
    }
};
```

```php
<?php

use Livewire\Component;
use Livewire\Attributes\On;

// Child component — errors because $title is never set
new class extends Component {
    public string $title;

    public function mount()
    {
        $this->title = 'Hello';
    }

    #[On('refresh-child')]
    public function refresh() {}

    public function render()
    {
        return <<<'HTML'
        <div>{{ $title }}</div>
        HTML;
    }
};
```

# The Problem

When a lazy component is initially rendered, `mount()` is skipped and a placeholder is output instead. However, `SupportEvents` still adds the component's event listeners to the dehydrated effects during this placeholder mount. On the JavaScript side, `supportListeners.js` registers global `window` event listeners and element-level listeners for "to"/"self" dispatches immediately, with no check for whether the component has actually been mounted.

When an event is dispatched, the JS handler fires `component.$wire.call('__dispatch', ...)`, sending a server request that calls the listener method on a component whose `mount()` was never executed.

# The Solution

Added a guard in both the global and element-level listener handlers in `supportListeners.js` to skip event dispatch when the component is lazy and hasn't been loaded yet:

```js
if (component.isLazy && ! component.hasBeenLazyLoaded) return
```

This uses the same pattern already established in `component.js` for skipping reactive prop updates on unmounted lazy children. Once the component is lazy-loaded via `x-intersect`, `hasBeenLazyLoaded` becomes `true` and events work normally.

Fixes #10216

### DIFF
--- a/js/features/supportListeners.js
+++ b/js/features/supportListeners.js
@@ -8,6 +8,8 @@ function registerListeners(component, listeners) {
     listeners.forEach(name => {
         // Register a global listener...
         let handler = (e) => {
+            if (component.isLazy && ! component.hasBeenLazyLoaded) return
+
             if (e.__livewire) e.__livewire.receivedBy.push(component)
 
             component.$wire.call('__dispatch', name, e.detail || {})
@@ -19,6 +21,8 @@ function registerListeners(component, listeners) {
 
         // Register a listener for when "to" or "self"
         component.el.addEventListener(name, (e) => {
+            if (component.isLazy && ! component.hasBeenLazyLoaded) return
+
             // We don't care about non-Livewire dispatches...
             if (! e.__livewire) return
 
--- a/src/Features/SupportLazyLoading/BrowserTest.php
+++ b/src/Features/SupportLazyLoading/BrowserTest.php
@@ -6,6 +6,7 @@ use Illuminate\Support\Facades\Route;
 use Tests\BrowserTestCase;
 use Livewire\Livewire;
 use Livewire\Component;
+use Livewire\Attributes\On;
 use Livewire\Attributes\Reactive;
 
 class BrowserTest extends BrowserTestCase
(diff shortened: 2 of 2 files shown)


calebporzio:
makes sense!

PR #10221, opened 2026-04-13
[4.x] Fix lazy dynamic event listeners set in mount

### LINKED ISSUE #10220: Lazy component doesn't fully support dynamic event
### Livewire version

v4.2.4

### Laravel version

v13.4.0

### PHP version

v8.4.14

Brave on Windows

Dynamic event listeners (e.g. `#[On('refresh-children:{parentId}')]`) are not fully supported when a component is marked as `lazy`. Specifically, if the property used in the dynamic event name is set inside `mount()`, this will cause the error `Unable to evaluate dynamic event name placeholder: {parentId}`

**It works** if the child component is not lazy.  
**It works** if `parentId` is set via implicit binding from the parent.
**It does not work** when the property is assigned inside `mount()` — which is required as soon as any logic needs to run before the property is set.

You can try directly here : https://wirebox.app/b/4zw7g

```php
# Parent component
{
    public function render()
    {
        return <<<'BLADE'
            <div
                style="height: 2000px !important"
            >
                <div class="bg-red-600 mb-[1000px]">
                    <button
                        type="button"
                        class="bg-black m-2 p-2 rounded cursor-pointer"
                        wire:click="$dispatch('refresh-children:{{ $this->getId() }}')"
                    >
                        refresh children
                    </button>
                </div>
            
                <div class="bg-blue-600 mb-[50px]">
                    <livewire:children
                        :context="[
                            'title' => 'test',
                            'parent_id' => $this->getId(),
                        ]"
                        lazy
                    />
                </div>
            </div>
        BLADE;
    }
}
```
```php
# Children component
<?php
namespace App\Livewire;

use Livewire\Component;
use Livewire\Attributes\On;
use App\Models\SomeData;

class Children extends Component
{
    public string $title;
    public SomeData $someData;
    public string $parentId;

    public function mount(array $context): void
    {
        $this->title = $context['title'] ?? '';
        $this->parentId = $context['parent_id'] ?? '';
    }

    #[On('refresh-children:{parentId}')]
    public function setData(): void
    {
        //
    }
    
    public function render()
    {
        return <<<'BLADE'
            <div class="flex w-full h-full flex-col items-center justify-center text-gray-300 px-16">
                <h1>Title : {{ $title }}</h1>
                <div>Parent Id : {{ $parentId }}</div>
            </div>
        BLADE;
    }
}
```

Not applicable i guess. Feel free to let me know if you really need to.

No error should occur and the dynamic event listener should be triggered correctly, even when the bound property is initialized in mount() and the component is marked as lazy. What do you think about that ?

prateekbhujel:
Closes #10220.

Dynamic listener placeholders were being resolved during the lazy placeholder mount, before `mount()` had a chance to populate the properties they depend on.

This skips listener serialization during the lazy placeholder mount and adds the listeners when the real lazy hydration runs instead. It also adds a browser regression test for the reported `#[On('refresh-child:{parentId}')]` case where `parentId` is assigned inside `mount()`.

## Tests

- `DUSK_DRIVER_URL=http://localhost:9515 ./vendor/bin/phpunit --configuration phpunit.xml.dist --filter 'test_lazy_component_(receives_events_after_being_loaded|can_use_dynamic_event_listeners_with_placeholders_set_in_mount)' src/Features/SupportLazyLoading/BrowserTest.php`


prateekbhujel:
Added a focused regression test for the reported lazy + dynamic listener case while working on this.

I also verified the flow locally end-to-end:
- the lazy child mounts without throwing when the listener placeholder is populated inside `mount()`
- clicking `Refresh` dispatches `refresh-child:{parentId}`
- the child listener runs and updates the rendered count from `0` to `1`

Before dispatch:
image

After dispatch:
image

Command used locally:
`DUSK_DRIVER_URL=http://localhost:9515 ./vendor/bin/phpunit --configuration phpunit.xml.dist --filter 'test_lazy_component_(receives_events_after_being_loaded|can_use_dynamic_event_listeners_with_placeholders_set_in_mount)' src/Features/SupportLazyLoading/BrowserTest.php`

src/Features/SupportLazyLoading/BrowserTest.php
```diff
@@ -733,6 +733,52 @@ public function render()
...
+                    HTML;
+                }
+            },
+            'child' => new class extends Component {
+                public string $parentId = '';
+                public int $count = 0;
+
+                public function mount(string $parentId): void
+                {
+                    $this->parentId = $parentId;
```
MaxencePaulin: This does not cover a potential regression, since `parentId` is passed implicitly to the child without necessarily going through the mount

MaxencePaulin:
Thanks for your fix ! If this PR gets merged, it would also cover the fix that Josh added previously for prevent executing event on lazy component not hydrated (#10217), so his fix would no longer be necessary i guess ? Might be worth cleaning it up at the same time !

prateekbhujel:
Good catch, I think you're right.

(https://github.com/livewire/livewire/pull/10217) guards things on the JS side so an unhydrated lazy component ignores events if listeners were already registered. This PR moves the fix a step earlier by not serializing/registering those listeners during the placeholder mount in the first place, so that guard may indeed end up being redundant.

I left this one focused because I wanted to cover the reported regression first, but I'm happy to fold in the cleanup as well if that's the preferred direction.

MaxencePaulin:
> I left this one focused because I wanted to cover the reported regression first, but I'm happy to fold in the cleanup as well if that's the preferred direction.

Sure, no problem — we'll wait for Josh or Caleb to take a look

joshhanley (approved):
@prateekbhujel thanks for the PR! @MaxencePaulin and @prateekbhujel I think it's ok to leave my fix in. It won't hurt to keep that check on the JS side.

# The Scenario

A child component marked as `lazy` uses a dynamic event listener like `#[On('refresh-children:{parentId}')]`, where `$parentId` is assigned inside `mount()`. When the parent renders, the page crashes with:

```
Unable to evaluate dynamic event name placeholder: {parentId}
```

It works when the child is not lazy, and it works when `$parentId` arrives via implicit prop binding, but it fails as soon as the property is initialised inside `mount()`.

```php
<?php

use Livewire\Component;

new class extends Component {} ?>

<div>
    <button wire:click="$dispatch('refresh-children:{{ $this->getId() }}')">
        Refresh children
    </button>

    <livewire:children
        :context="['parent_id' => $this->getId()]"
        lazy
    />
</div>
```

```php
<?php

use Livewire\Component;
use Livewire\Attributes\On;

new class extends Component {
    public string $parentId;

    public function mount(array $context): void
    {
        $this->parentId = $context['parent_id'] ?? '';
    }

    #[On('refresh-children:{parentId}')]
    public function refresh(): void
    {
        //
    }
} ?>

<div>Parent ID: {{ $parentId }}</div>
```

# The Problem

When a `lazy` component renders its placeholder, Livewire skips the child's `mount()` and dehydrates the component to take a snapshot for the placeholder HTML. The real `mount()` doesn't run until the placeholder later calls `$wire.__lazyLoad(...)`.

That's where the bug lives. Because `mount()` was skipped, `$parentId` is never assigned. But during dehydration, `SupportEvents` still tries to resolve every listener, including dynamic ones like `refresh-children:{parentId}`. It looks up `parentId` on the component, finds it uninitialised, and throws:

```php
return data_get($component, $matches[1], function () use ($matches) {
    throw new \Exception('Unable to evaluate dynamic event name placeholder: '.$matches[0]);
});
```

The component never even gets a chance to mount before it crashes.

# The Solution

`SupportEvents::dehydrate()` now defers listener serialisation until the component has actually mounted.

There are two checkpoints to handle:

1. **The placeholder pass.** Skip listener serialisation entirely. The component hasn't mounted yet, so dynamic placeholders can't be evaluated, and anything added here would be thrown away when the real hydration happens anyway.
2. **The lazy hydration pass (`__lazyLoad`).** This is when the real `mount()` runs, so it's the right moment to serialise listeners. Without this, lazy components would never register their listeners on the client at all.

`SupportJsModules` already follows this exact pattern for its own dehydration (skip during `isLazyLoadMounting`, run during `isLazyLoadHydrating`), so this change just brings `SupportEvents` in line with that precedent.

Fixes #10220

src/Features/SupportLazyLoading/BrowserTest.php
joshhanley: Ah @MaxencePaulin yeah you're correct. @prateekbhujel would you mind fixing that? Maybe just set parent ID to a random value in mount. Doesn't really need to be passed down from the parent.

prateekbhujel:
I went ahead and folded in the cleanup.

Since this PR now prevents listeners from being serialized during the lazy placeholder mount, the JS-side guard added in `#10217` ended up being redundant here. I removed that fallback and re-ran the lazy listener browser coverage.

Verified locally:
- lazy component outside the viewport still ignores a global dispatch before loading
- `dispatchTo()` is still ignored before loading
- events still work normally after the lazy component has loaded
- the dynamic `#[On('refresh-child:{parentId}')]` case still works when `parentId` is assigned in `mount()`

Command used:
`DUSK_DRIVER_URL=http://localhost:9515 ./vendor/bin/phpunit --configuration phpunit.xml.dist --filter 'test_lazy_component_(outside_viewport_ignores_global_dispatch|outside_viewport_ignores_dispatch_to|receives_events_after_being_loaded|can_use_dynamic_event_listeners_with_placeholders_set_in_mount)' src/Features/SupportLazyLoading/BrowserTest.php`

joshhanley:
@prateekbhujel please read my message above about the cleanup.

src/Features/SupportLazyLoading/BrowserTest.php
prateekbhujel: Fixed, thanks both.

I pushed an update that removes the implicit prop pass-through from the test. `parentId` is now created inside `mount()` only, and the dispatch reads that generated value from the lazy child after it has loaded.

I also restored the JS-side guard from `#10217` per Josh's note above, so the branch now keeps both protections in place.

Re-ran locally:
`DUSK_DRIVER_URL=http://localhost:9515 ./vendor/bin/phpunit --configuration phpunit.xml.dist --filter 'test_lazy_component_(outside_viewport_ignores_global_dispatch|outside_viewport_ignores_dispatch_to|receives_events_after_being_loaded|can_use_dynamic_event_listeners_with_placeholders_set_in_mount)' src/Features/SupportLazyLoading/BrowserTest.php`

prateekbhujel:
Yep, reread it and updated the branch accordingly.

I restored the JS-side guard from https://github.com/livewire/livewire/issues/10220, and I tightened the regression test so `parentId` is generated inside `mount()` instead of being passed from the parent. The focused lazy-event browser coverage is still green after that change.

Thanks for calling that out.

src/Features/SupportLazyLoading/BrowserTest.php
```diff
@@ -740,8 +740,8 @@ public function test_lazy_component_can_use_dynamic_event_listeners_with_placeho
                 public function render() {
                     return <<<'HTML'
                     <div>
-                        <button wire:click="$dispatch('refresh-child:{{ $this->getId() }}')" dusk="button">Refresh</button>
-                        <livewire:child :parentId="$this->getId()" lazy />
+                        <button @click="window.Livewire.dispatch('refresh-child:' + document.getElementById('child').dataset.parentId)" dusk="button">Refresh</button>
+                        <livewire:child lazy />
```
MaxencePaulin, line 744: Wire:click was good haha. You made things harder for yourself, but it works!

I think he was talking about something like that

```blade
<button wire:click="$dispatch('refresh-child:parent-xxx')" dusk="button">Refresh</button>
```

```php
public function mount(): void
{
    $this->parentId = 'parent-xxx';
}
```

joshhanley (approved):
@prateekbhujel thanks! I've pushed a fix to simplify the test.

src/Features/SupportLazyLoading/BrowserTest.php
joshhanley: @MaxencePaulin yep exactly! @prateekbhujel I've fixed it up. Thanks!

prateekbhujel:
Thanks @joshhanley, appreciate you cleaning that up. Much nicer this way.

prateekbhujel:
Just checking if there’s anything else needed from me on this one. Thanks.

joshhanley:
@prateekbhujel not from me. Caleb does the merging of most PRs.

calebporzio:
@claude - this PR makes good sense, however the logic in it: the early return and then the same condition used in an or statement later, reads a little weird to me. can you please confirm what im saying or challenge it and maybe refactor the conditions to read more logically/intuitively/clearly. thank you.

calebporzio:
@claude - try again, your last attempt failed in CI. just look at my last message to you and use that

calebporzio:
thanks!

prateekbhujel:
Thanks @calebporzio . Really appreciate you taking this one through.

PR #10223, opened 2026-04-13
[4.x] Fix wire:show with $errors.has()

### LINKED ISSUE #10222: wire:show with $errors.has() doesn't work
### Livewire version

v4.2.4

### Laravel version

v13.4.0

### PHP version

8.3.30

Chrome on WSL Ubuntu 24.04

Showing a validation error message using wire:show and $errors.has() doesn't work.
The div which has the wire:show directive has a style="display:none" applied to it, but it is never removed.
The wire:text directive with $errors.first() does work however.

Here's a fragment of the docs:

Image

```
<?php

use Livewire\Component;

new class extends Component {
    #[Validate('required')]
    public string $email;

    public function save()
    {
        $this->validate();
    }
};
?>

<div>
    <form wire:submit="save">
        <input type="email" wire:model="email">

        <div wire:show="$errors.has('email')">
            <span wire:text="$errors.first('email')"></span>
        </div>

    </form>
</div>
```

_No response_

I would expect the error message to be visible by removing the style="display:none" on the element

prateekbhujel:
Closes #10222.

## Summary

This fixes `wire:show="$errors.has(...)"` not updating when validation errors are added.

The issue was that the client-side `$errors` helper wasn't exposing a reactive dependency when the server error bag changed, so `wire:text="$errors.first(...)"` could update while `wire:show="$errors.has(...)"` stayed hidden.

This change keeps the current server errors in reactive state, resets any client-side overrides after each response, and adds a browser regression test for the reported repro.

## Testing

- `vendor/bin/phpunit src/Features/SupportWireShow/BrowserTest.php`
- `vendor/bin/phpunit src/Features/SupportMagicErrors/BrowserTest.php`

### DIFF
--- a/js/features/supportErrors.js
+++ b/js/features/supportErrors.js
@@ -1,23 +1,28 @@
 // This errors object has the most common methods from \Illuminate\Support\MessageBag class on the backend...
 import Alpine from 'alpinejs'
+import { interceptComponentMessage } from '@/request'
 
 export function getErrorsObject(component) {
     let state = component.__errorsState ??= Alpine.reactive({
         clientErrors: null,
+        serverErrors: component.snapshot.memo.errors,
     })
 
-    // Store lastSnapshot outside reactive state to avoid Proxy wrapping breaking identity comparison...
-    component.__lastErrorsSnapshot ??= component.snapshot
+    component.__errorsCleanup ??= interceptComponentMessage(component, ({ onFinish }) => {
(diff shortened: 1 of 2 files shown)


joshhanley:
@prateekbhujel thanks for the PR! I've submitted a different PR #10229 with a simpler fix, so I'm going to close this. Thanks for being willing to contribute!

PR #10225, opened 2026-04-14
Add .claude to export-ignore in .gitattributes

SRWieZ:
I saw my Claude going through those directories while trying to implement a new feature. 

I believe this directory should not be included when the project is imported with Composer.


joshhanley:
Thanks!

PR #10231, opened 2026-04-17
[4.x] Update `wire:key` documentation to include conditionals

### LINKED ISSUE #10227: wire:model in multi-step form triggers update of previous input
### Livewire version

v4.2.4

### Laravel version

v13.4.0

### PHP version

8.5.5

Mozilla on Linux Mint

I'm trying to make a multi-step form.

The problem that I encounter is that I fill the input of the first form and then proceed to the second form. **When I fill the input of the second form, the input of the first form is updated with the value of the second form.**

They both have different form names and their properties have different names too but I can't manage to find what is wrong and it seems to be a bug to me.

I reduced the original code to its basic in order to trigger the error with the same kind of multi-step form I try to make.

```blade
<?php

use App\Livewire\Forms\FirstForm;
use App\Livewire\Forms\SecondForm;
use Livewire\Component;

new class extends Component
{
    public FirstForm $firstForm;

    public SecondForm $secondForm;

    #[Locked]
    public string $current = 'user';

    #[Locked]
    public array $steps = [
        'user' => 'current',
        'company' => 'incomplete',
        'review' => 'incomplete',
    ];

    public function validateAndNextStep()
    {
        if ($this->current === 'user') {
            // $this->firstForm->validate();
        } elseif ($this->current === 'company') {
            // $this->secondForm->validate();
        }

        $this->nextStep();
    }

    protected function nextStep()
    {
        $currentIndex = array_search($this->current, array_keys($this->steps));
        if ($currentIndex !== false && $currentIndex < count($this->steps) - 1) {
            $this->steps[$this->current] = 'complete';
            $this->current = array_keys($this->steps)[$currentIndex + 1];
            $this->steps[$this->current] = 'current';
        }
    }

    public function previousStep()
    {
        $currentIndex = array_search($this->current, array_keys($this->steps));
        if ($currentIndex !== false && $currentIndex > 0) {
            $this->steps[$this->current] = 'incomplete';
            $this->current = array_keys($this->steps)[$currentIndex - 1];
            $this->steps[$this->current] = 'current';
        }
    }
};
?>

<div class="flex flex-col justify-center items-center gap-6 my-6">

    <flux:timeline horizontal>
        <flux:timeline.item status="{{ $steps['user'] }}">
            <flux:timeline.indicator>
                <flux:icon.user-round-plus variant="micro" />
            </flux:timeline.indicator>

            <flux:timeline.content>
                <flux:heading>First Form</flux:heading>
            </flux:timeline.content>
        </flux:timeline.item>

        <flux:timeline.item status="{{ $steps['company'] }}">
            <flux:timeline.indicator>
                <flux:icon.building-2 variant="micro" />
            </flux:timeline.indicator>

            <flux:timeline.content>
                <flux:heading>Second Form</flux:heading>
            </flux:timeline.content>
        </flux:timeline.item>

        <flux:timeline.item status="{{ $steps['review'] }}">
            <flux:timeline.indicator>
                <flux:icon.list-checks variant="micro" />
            </flux:timeline.indicator>

            <flux:timeline.content>
                <flux:heading>Review and Validate</flux:heading>
            </flux:timeline.content>
        </flux:timeline.item>
    </flux:timeline>

    @dump($firstForm->email)
    @dump($secondForm->name)
    @dump($current)

    @switch($current)
        @case('user')
            <flux:card class="flex flex-col justify-center items-center w-full">
                <flux:input autofocus badge="Required" clearable label="Email" placeholder="email placeholder" type="email"
                    wire:model.live.debounce="firstForm.email" />
            </flux:card>
        @break

        @case('company')
            <flux:card class="flex flex-col justify-center items-center w-full">
                <flux:input class="max-w-sm" label="Company Name" placeholder="company placeholder" type="text"
                    wire:model.live.debounce="secondForm.name" />
            </flux:card>
        @break

        @case('review')
            <div class="flex flex-col justify-center items-center gap-2 sm:gap-6">
                <flux:card class="flex flex-col justify-center items-center w-full">
                    <flux:input autofocus badge="Required" clearable disabled label="Email" placeholder="email placeholder"
                        type="email" wire:model.live.debounce="firstForm.email" />
                </flux:card>

                <flux:card class="flex flex-col justify-center items-center w-full">
                    <flux:input class="max-w-sm" disabled label="Company Name" placeholder="company placeholder" type="text"
                        wire:model.live.debounce="secondForm.name" />
                </flux:card>
            </div>
        @break

        @default
            Default
        @break
    @endswitch

    @switch($current)
        @case('user')
        @case('company')
            <div class="flex justify-between gap-2">
                <flux:button wire:click="previousStep">Back</flux:button>
                <flux:button wire:click="validateAndNextStep">Next</flux:button>
            </div>
        @break

        @case('review')
            <div class="flex justify-end gap-2">
                <flux:button wire:click="previousStep">Back</flux:button>
                <flux:button variant="primary">Save</flux:button>
            </div>
        @break


    @break

    @default
@endswitch
</div>

``` 

_No response_

Since I have two different forms, with two different properties, I'd like to be able to update one or another without impacting the other.

joshhanley:
# The Scenario

A user is rendering one of two structurally similar inputs based on a property, each bound to a different `wire:model` target:

```blade
<div>
    <button wire:click="$set('current', 'company')">Company</button>

    @switch($current)
        @case('user')
            <flux:card>
                <flux:input wire:model="email" />
            </flux:card>
        @break

        @case('company')
            <flux:card>
                <flux:input wire:model="name" />
            </flux:card>
        @break
    @endswitch
</div>
```

After clicking the button to switch from `'user'` to `'company'`, typing into the visible input updates `$email` instead of `$name`. The same problem occurs with `@if/@else` branches that contain structurally similar elements bound to different properties.

# The Problem

Livewire's morph engine sees the same `<flux:input>` element at the same DOM position before and after the switch, so it patches attributes in place rather than replacing the element. The `wire:model` directive's bound property reference is set when the directive initialises and isn't re-resolved when the attribute value changes during morph, so it stays pointing at the original property.

This is the same class of issue as missing `wire:key` on `@foreach` items, but the existing `wire:key` troubleshooting docs only mention loops, not conditionals.

# The Solution

Update the `Adding wire:key` section of `docs/troubleshooting.md` to mention that `@if/@else` and `@switch/@case` branches need `wire:key` for the same reason as loops.

## Why a docs change rather than a code fix

We prototyped an automatic fix where the precompiler injects `<!--[if BRANCH:hash]-->` / `<!--[if ENDBRANCH:hash]-->` markers inside each branch of `@if/@else/@elseif/@endif`, with the intent to extend the same approach to `@switch/@case/@default` and other conditional directives. PHP's natural branching gates which markers actually render, so morph could compare hashes and replace branches wholesale on mismatch (avoiding the stale `wire:model` binding problem entirely, with no user-supplied `wire:key` required).

The PoC compiled correctly but added meaningful overhead to Blade compilation. Benchmark on a representative template (10 top-level `@if`s, mix of standalone / `@else` / `@elseif`, with nesting up to 3 levels deep, ~1.5 KB), 1000 iterations:

| Approach                                       | Per-compile | Overhead vs baseline |
|------------------------------------------------|-------------|----------------------|
| Baseline (no BRANCH markers)                   | ~2.7 ms     |                      |
| Standalone precompiler (chunked, O(n) build)   | ~3.5 ms     | **+730 µs (~28%)**   |
| Integrated into `compileDirectives` reverse loop | ~3.7 ms   | **+985 µs (~35%)**   |

A more sustainable home for this kind of compile-time transformation would be an AST-based pipeline (e.g. extending `livewire/blaze`'s tokenizer + parser with directive nodes), where BRANCH markers, BLOCK markers, and smart wire-keys all become trivial visitors over a single parse. That's a larger investment than this issue warrants, so for now we just document the workaround.

Fixes #10227

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -18,7 +18,7 @@ There are lots of reasons why you may encounter these messages, but the most com
 
 ### Adding `wire:key`
 
-Any time you have a loop or a conditional in your Blade templates using something like `@foreach`, `@if`/`@else`, or `@switch`/`@case`, you need to add `wire:key` to the opening tag of the first element:
+Any time you have a loop or switch case in your Blade templates using something like `@foreach`, or `@switch`/`@case`, you need to add `wire:key` to the opening tag of the first element:
 
 ```blade
 @foreach($posts as $post)
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -18,7 +18,7 @@ There are lots of reasons why you may encounter these messages, but the most com
 
 ### Adding `wire:key`
 
-Any time you have a loop in your Blade templates using something like `@foreach`, you need to add `wire:key` to the opening tag of the first element within the loop:
+Any time you have a loop or a conditional in your Blade templates using something like `@foreach`, `@if`/`@else`, or `@switch`/`@case`, you need to add `wire:key` to the opening tag of the first element:
 
 ```blade
 @foreach($posts as $post)
@@ -28,7 +28,7 @@ Any time you have a loop in your Blade templates using something like `@foreach`
 @endforeach
 ```
 
-This ensures that Livewire can keep track of different elements in the loop when the loop changes.
+This ensures that Livewire can keep track of different elements in the loop or conditional when it changes.
 
 The same applies to Livewire components within a loop:
 


calebporzio:
thanks, i tweaked the docs to not say "a conditional" and instead just talk about switch cases because people should only really need this with long ifelse and switch cases which are much less common than simply conditionals and this could be missinterpreted as all conditionals. anyways. thanks!

Issue #2717, opened 2021-04-13
Livewire encountered corrupt data when trying to hydrate when a variable that has -0.0


ChrisSantiago82:
This might be an issue that almost never happens.

I have a positive number from a Database and I need to have it in negative. I'm multiply it with -1 and then store in a public variable for Livewire. Everything works fine unless the number was 0. 0 times -1 returns in PHP -0.0

Then when Livewire tries to hydrate it I'm getting a "Livewire encountered corrupt data when trying to hydrate". I'm guessing the -0.0 is being changed to 0 on the way and when it comes back livewire thinks the component was tampered with.
I don't have a problem if -0.0 changes to 0, but then livewire should also know about that change.

Of course it's easy for me to fix this and in this case just send a 0. But it took me a while to find the issue because it doesn't happen often that this number is 0 and maybe somebody else will encounter the same issue. But I also would understand if it's not worth fixing it....

**Context**
- Livewire version: 2.42
- Laravel version: [e.g. 8.36.2]
- Browser: Chrome

calebporzio:
Thanks for reporting this. I think your diagnoses of the problem is spot on.

I'm going to close this as it's fairly edge casey, but would love it if you PRd a fix (which should be fairly straightforward)

Thanks!

squishythejellyfish:
👋 Oh Hi! I'm Squishy, the friendly jellyfish that manages Livewire issues.

I see this issue has been closed.

Here in the Livewire repo, we have an "issues can be closed guilt-free and without explanation" policy.

If for ANY reason you think this issue hasn't been resolved, PLEASE feel empowered to re-open it.

Re-opening actually helps us track which issues are a priority.

Reply "REOPEN" to this comment and we'll happily re-open it for you!

(More info on this philosophy here: https://twitter.com/calebporzio/status/1321864801295978497)

dac514:
I had this issue today.

As explained by original submitter, checksum breaks because `-0 != 0`

For me, it took a long time to figure it out by manually comparing variable differences in: 

`Livewire\Mechanisms\HandleComponents\Checksum::verify`

Only after I fixed it did I manage to find this GitHub issue, because I knew the exact problem and was going to report it. 

No Google search helped.

>  would love it if you PRd a fix (which should be fairly straightforward)

Maybe modifications to?

`Livewire\Mechanisms\HandleComponents\HandleComponents::snapshot`
`Livewire\Drawer\BaseUtils::getPublicProperties`

For now, I did it in my model.

[...]

PR #10233, opened 2026-04-17
Livewire encountered corrupt data when trying to hydrate when a variable that has -0.0

dac514:
Fix for: Livewire encountered corrupt data when trying to hydrate when a variable that has -0.0

See: https://github.com/livewire/livewire/issues/2717


calebporzio:
thanks!

PR #10234, opened 2026-04-17
[4.x] Add a Compiled Path option for the config

matt-h:
This is a complement to the config option for `view.compiled` in Laravel: https://github.com/laravel/framework/blob/13.x/config/view.php#L20-L35

Right now the compiled path is configured to be hardcoded to `storage_path('framework/views/livewire')` which doesn't match the customization Laravel has for blade views.

The other option instead of this or as a default in this config would be to default it to `config('view.compiled').'/livewire'`

---

No, but matches Laravel's config.

Yes

No

Yes

Included at top.

Thanks for contributing! 🙌

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -119,6 +119,19 @@ return [
 
     /*
     |---------------------------------------------------------------------------
+    | Compiled Path
+    |---------------------------------------------------------------------------
+    |
+    | This value determines where Livewire stores compiled view-based component
+    | artifacts such as generated classes, extracted Blade views, placeholders,
+    | scripts, styles, and island templates.
+    |
+    */
+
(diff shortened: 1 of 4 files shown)


joshhanley (changes requested):
@matt-h thanks for the PR! Something similar has previously been submitted but was closed. See PR #9800.

In light of that, I think it would be better to not add an ENV/config for Livewire's compiled path, as we want it to remain inside Laravel's. But I think it'd be ok to update our compiler path to reference the Laravel config instead of hardcoding it.

Are you able to make those changes?

matt-h:
Thanks @joshhanley, not sure how I missed that other PR, but reading through that that makes sense.

I have made the changes and made a new PR for the new work. https://github.com/livewire/livewire/pull/10240

Closing this one out.

PR #10237, opened 2026-04-20
[4.x] Add documentation reference for `$set` and `$toggle` live parameter

ghabriel25:
Magic actions `$set` and `$toggle` accepts third parameter (boolean) whether it should trigger network request or not. 

However, there is no information about third parameter of `$set` and `$toggle` magic actions on the **Magic Actions** documentation. This PR adds reference link to `$wire` object details.

The third parameter often used for custom select dropdown which usually set the value based on `wire:model` or `wire:model.live`. With this information, user can use the magic actions and customize the syncing should be live or not.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -611,14 +611,6 @@ The `$set` magic action allows you to update a property in your Livewire compone
 
 In this example, when the button is clicked, a network request is dispatched that sets the `$query` property in the component to `''`.
 
-`$set` action accepts a third boolean parameter that determines whether the update should trigger a network request:
-
-```blade
-<button wire:click="$set('query', '', false)">Reset Search (no request)</button>
-```
-- `true` (default): Sends a network request and updates the property on the server.
-- `false`: Updates the property without making a network request.
-
 ### `$refresh`
 
 The `$refresh` action triggers a re-render of your Livewire component. This can be useful when updating the component's view without changing any property values:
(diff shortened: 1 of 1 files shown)


joshhanley:
@ghabriel25 thanks for the PR! We sort of have this documented on the `$wire` object docs https://livewire.laravel.com/docs/4.x/javascript#reference but yeah as they can be used directly in `wire:click` there should probably be a reference for it somewhere.

I personally think maybe we should have a "reference" section at the bottom of the "Magic actions" section (or maybe at the bottom of the page, although that might feel disjointed) which documents all the magic actions and their parameters like we have for `$wire`, rather than adding more to the `$set` and `$toggle` sections.

ghabriel25:
Hi, @joshhanley 

What about include the parameters information after the magic action name?

For example

#### `$set ` parameters: (name, value, live = true)

joshhanley:
@ghabriel25 what do you mean? Can you show me what that might look like?

ghabriel25:
Hi, @joshhanley 

I have updated and simplify things with just referencing the Javascript section

calebporzio:
Thanks.

I removed the bit where you switched Asc/Desc because what we had was actually correct if you think about it.

Also note: next time, keep unrelated changes as separate PRs to avoid things like this.

Thanks!

ghabriel25:
Ah my bad. I think the AI autocomplete that part and didnt notice that change

PR #10239, opened 2026-04-21
[4.x] Fix the url reference of wire object in Javascript documentation

ghabriel25:
There was a wrong url reference on `$wire` object in Javascript section

image

The url should be referencing to `#wire-object-1` instead `#wire-object` which was self referencing.

### Before
**`[$wire reference](#the-wire-object)`**

### After
**`[$wire reference](#the-wire-object-1)`**


joshhanley:
Thanks!

PR #10240, opened 2026-04-21
[4.x] Use Laravel compiled view path for Livewire compiler

matt-h:
A follow up to https://github.com/livewire/livewire/pull/10234

This uses Laravel's value of `config('view.compiled')` which defaults to `storage_path('framework/views')` instead of hardcoding the path to be `storage_path('framework/views/livewire')`.

This preservers the customization of your `view.compiled` path while still putting it in the `livewire` path after that by setting the path to `config('view.compiled').'/livewire'` directly.

### DIFF
--- a/src/Features/SupportIslands/Compiler/IslandCompiler.php
+++ b/src/Features/SupportIslands/Compiler/IslandCompiler.php
@@ -178,7 +178,7 @@ PHP;
     {
         $instance = new class (
             app('files'),
-            storage_path('framework/views/livewire'),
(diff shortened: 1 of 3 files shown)


joshhanley (approved):
@matt-h thanks for the PR! Looks good to me. Only small nitpick would be the islands test using `app('livewire.compiler')->cacheManager->cacheDirectory` should probably just use the config path the same as the other change. But it's a test, so doesn't overly matter.

calebporzio:
This feel great, thank you.

PR #10242, opened 2026-04-22
[4.x] Fix nested unnamed wire:transition not following parent's typed swap

### LINKED ISSUE #10232: Transition type not working on element that has `wire:transition`
### Livewire version

v4.2

### Laravel version

v12.0

### PHP version

8.4

Chrome on Windows

When using livewire dynamic component with transition type, if there is an element inside the component using `wire:transition` (used for smooth transition when form has validation error), the transition type doesnt work on that element

```php
<?php

use Livewire\Component;

new class extends Component
{
    #[Locked]
    public string $mode = 'profile.view-mode';

    #[Transition('backward')]
    public function viewMode(): void
    {
        $this->mode = 'profile.view-mode';
    }

    #[Transition('forward')]
    public function editMode(): void
    {
        $this->mode = 'profile.edit-mode';
    }
}
```
```blade
// pages.profile.blade.php
<div>
    <livewire:dynamic-component :is="$mode" :wire:key="$mode" />
</div>

// livewire.profile.view-mode.blade.php
<div wire:transition="slide">
    <flux:button wire:click="$parent.editMode()">Edit</flux:button>
</div>

// livewire.profile.edit-mode.blade.php
<div wire:transition="slide">
    <form class="space-y-6" wire:submit="save">
        <flux:input label="Email" wire:model="email" />

         // this get skip from transition type
        <div class="flex items-center gap-2" wire:transition>
            <flux:button wire:click="$parent.viewMode()">Cancel</flux:button>
            <flux:button type="submit">Save changes</flux:button>
        </div>
    </form>
</div>
```

### Workaround
Setting the `wire:transition` attribute using Alpine and use `wire:ignore.self` on that element
```blade
<div x-init="$el.setAttribute('wire:transition', '')" wire:ignore.self>
    <flux:button wire:click="$parent.viewMode()">Cancel</flux:button>
    <flux:button type="submit">Save changes</flux:button>
</div>
```
This works for general html element, however I think this will cause some issue if applied on flux components

https://github.com/user-attachments/assets/496bef13-a9ec-4869-b15e-a42582e79ea4

It shouldnt get skipped and respect the transition type but still working as usual

joshhanley:
# The Scenario

When a parent swaps children with a typed transition (`#[Transition('forward')]`), any nested element with `wire:transition` (no value) animates with the browser's default fade instead of sliding along with the parent's animation.

Screenshot 2026-04-23 at 09 47 07AM

```php
<?php

use Livewire\Attributes\Transition;
use Livewire\Component;

new class extends Component
{
    public string $mode = 'profile.view-mode';

    #[Transition('forward')]
    public function editMode(): void
    {
        $this->mode = 'profile.edit-mode';
    }
}
```

```blade
{{-- pages/profile.blade.php --}}
<div>
    <livewire:dynamic-component :is="$mode" :wire:key="$mode" />
</div>

{{-- livewire/profile/edit-mode.blade.php --}}
<div wire:transition="slide">
    <form wire:submit="save">
        <flux:input label="Email" wire:model="email" />

        {{-- Animates independently with the default fade during the parent swap --}}
        <div wire:transition>
            <flux:button type="submit">Save changes</flux:button>
        </div>
    </form>
</div>
```

# The Problem

In the View Transitions API, any element with a `view-transition-name` is captured as its own group. Named groups animate independently, and an unnamed element rides along with its parent's snapshot.

`setTransitionNames` always assigns a name to every `wire:transition` element, falling back to `match-element` when none is given:

```js
function setTransitionNames(root) {
    root.querySelectorAll('[wire\\:transition]').forEach(el => {
        if (! el.style.viewTransitionName) {
            el.style.viewTransitionName = el.getAttribute('wire:transition') || defaultName
        }
    })
}
```

So during the typed parent swap, the inner `<div wire:transition>` becomes its own group. It only exists in the new snapshot, has no user CSS targeting it, and falls back to the browser's default ~250ms fade while the surrounding panel slides 600ms with the user's `:active-view-transition-type(forward)` rules.

# The Solution

When a transition has an explicit type, the developer is orchestrating the swap as a single visual unit. Inside that orchestration, only explicitly named `wire:transition` elements should animate independently. Unnamed ones should ride along with the parent.

`setTransitionNames` now skips the `match-element` fallback when `options.type` is set:

```js
function setTransitionNames(root, options = {}) {
    root.querySelectorAll('[wire\\:transition]').forEach(el => {
        if (el.style.viewTransitionName) return

        let name = el.getAttribute('wire:transition')

        if (! name && ! options.type) name = defaultName

        if (name) el.style.viewTransitionName = name
    })
}
```

Named transitions still get their explicit name and animate independently as before. Unnamed transitions during a regular morph (e.g. a layout shift when a validation error appears) continue to be auto-named so they animate independently for the smooth in-place transition users already expect.

Screenshot 2026-04-23 at 09 34 46AM

Fixes #10232

### DIFF
--- a/docs/wire-transition.md
+++ b/docs/wire-transition.md
@@ -155,6 +155,29 @@ class Wizard extends Component
 }
 ```
 
+### Unnamed transitions during a typed swap
+
+When a typed transition is active, Livewire treats the swap as a single orchestrated unit. Any *unnamed* `wire:transition` element inside that swap stays unnamed and rides along with its parent's snapshot, instead of becoming an independent group with the browser's default fade.
+
+```blade
+<div wire:transition="slide">
+    <p>Step content</p>
+
+    {{-- Unnamed: rides along with the parent's "slide" during a typed swap --}}
+    <div wire:transition>
+        <button>Save</button>
+    </div>
+</div>
+```
+
+If you need an inner element to animate independently during a typed swap, give it an explicit name:
+
+```blade
+<div wire:transition="badge">...</div>
+```
+
+Outside of a typed transition (a regular morph, like a validation error appearing), unnamed `wire:transition` elements continue to use `match-element` and animate independently as before.
+
 ## Skipping transitions
 
 Sometimes you may want to disable transitions for specific actions—for example, a "reset" button that should instantly jump to the first step without animation.
--- a/js/directives/wire-transition.js
+++ b/js/directives/wire-transition.js
@@ -14,13 +14,12 @@ function setTransitionNames(root, options = {}) {
 
         let name = el.getAttribute('wire:transition')
 
-        // When an explicit transition type is active (e.g. via #[Transition('forward')]),
-        // the developer is orchestrating the swap as a single visual unit. Unnamed
-        // wire:transition elements should ride along with their parent's snapshot
-        // rather than animate independently and pop with the browser's default fade...
-        if (! name && ! options.type) name = defaultName
+        // When the developer orchestrates a typed swap (e.g. #[Transition('forward')]),
+        // unnamed wire:transition elements stay unnamed so they ride with the parent's
+        // snapshot instead of becoming independent groups with the browser's default fade...
+        if (! name && options.type) return
 
-        if (name) el.style.viewTransitionName = name
+        el.style.viewTransitionName = name || defaultName
     })
 }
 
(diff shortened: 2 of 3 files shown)


ghabriel25:
Thanks! I'll try this out.

**Update:**
@joshhanley I have test this fix and it works as expected 👍

joshhanley:
@ghabriel25 great, thanks for testing!

calebporzio:
made a few tweaks for clarity and sanity checked the implementation as a whole. thanks!

PR #10245, opened 2026-04-24
[4.x] Fix unchanged wire:model.live.blur requests

people-sea:

Yes. This is related to filamentphp/filament#19764, where an unchanged `live(onBlur: true)` field can still cause an unexpected Livewire round-trip.

I did not open a Livewire discussion first because this is a small fix for a no-op model request.

Yes.

No.

Yes.

This skips clean model-triggered commits from `wire:model.live.blur`.

A blur still sends a request when there are pending model updates on the component. It only skips the request when the commit would have no updates to send.

### DIFF
--- a/js/component.js
+++ b/js/component.js
@@ -131,6 +131,11 @@ export class Component {
         return this.mergeQueuedUpdates(propertiesDiff)
     }
 
+    hasUpdates() {
+        return Object.keys(this.queuedUpdates).length > 0
+            || Object.keys(diffAndConsolidate(this.canonical, this.ephemeral)).length > 0
+    }
+
     applyUpdates(object, updates) {
         for (let key in updates) {
             dataSet(object, key, updates[key])
--- a/js/request/index.js
+++ b/js/request/index.js
@@ -189,8 +189,20 @@ export function createOrAddToOutstandingMessage(action) {
 
 function sendMessages() {
     let requests = new Set()
+    let skippedMessages = new Set()
 
     messageBus.eachPendingMessage(message => {
+        if (! messageIsSkippableModelLiveCommit(message)) return
+
+        if (message.component.hasUpdates()) return
+
+        skippedMessages.add(message)
+        message.resolveActionPromises([], [])
+    })
+
+    messageBus.eachPendingMessage(message => {
+        if (skippedMessages.has(message)) return
+
         partitionInterceptors.forEach(callback => {
             callback({
                 message,
@@ -230,7 +242,7 @@ function sendMessages() {
         })
     })
 
-    let messages = messageBus.getPendingMessages()
+    let messages = messageBus.getPendingMessages().filter(message => ! skippedMessages.has(message))
 
     messageBus.clearPendingMessages()
 
@@ -480,6 +492,13 @@ function sendMessages() {
             },
         })
     })
+}
+
+function messageIsSkippableModelLiveCommit(message) {
+    let actions = message.getActions()
+
+    return actions.length > 0
+        && actions.every(action => action.name === '$commit' && action.metadata.type === 'model.live')
 }
 
 async function sendRequest(request, handlers) {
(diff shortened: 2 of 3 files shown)


js/directives/wire-model.js
```diff
@@ -166,12 +168,17 @@ function isRealtimeInput(el) {
         || el.tagName.toUpperCase() === 'UI-COMPOSER' // Flux UI
 }
 
-function isDirty(subject, dirty) {
-    // Check for exact match: wire:model="bob" in ['bob']
-    if (dirty.includes(subject)) return true
+function hasDirtyUpdates(component, property) {
```
calebporzio: This function makes me nervous. There's a lot of custom logic and heavy lifting.

js/directives/wire-model.js
```diff
@@ -166,12 +168,17 @@ function isRealtimeInput(el) {
...
+    if (property.startsWith('$parent')) {
+        let parent = findComponentByEl(component.el.parentElement, false)
+
+        if (! parent) return false
 
-    // Check case of parent: wire:model="bob.1" in ['bob']
-    return dirty.some(i => subject.startsWith(i))
+        return hasDirtyUpdates(parent, property.slice(7).replace(/^\./, ''))
```
calebporzio: for example, having this conditional for `$parent` is good, but it makes me think: "what other prefixes are we missing or will later encounter that this doesn't account for?" you know?

js/directives/wire-model.js
```diff
@@ -166,12 +168,17 @@ function isRealtimeInput(el) {
...
+    }
+
+    return Object.keys(component.queuedUpdates).length > 0
```
calebporzio: surgically accessing queuedUpdates and depending on it in here doesn't feel great. I'm also not sure there is a precedent for this anywhere else in the codebase.

This logic feels like something that should live inside the component.

js/directives/wire-model.js
```diff
@@ -166,12 +168,17 @@ function isRealtimeInput(el) {
...
+        || Object.keys(diffAndConsolidate(component.canonical, component.ephemeral)).length > 0
```
calebporzio: diffAndConsolidate is not a cheap operation. And also not one that feels like it should live outside the component or request areas of the codebase.

calebporzio (changes requested):
Overall, I want this to exist because it cuts down on unnescessary requests.

However, the implementation makes me nervous. I suspect it may fall short in three areas:

- robustness/future-proofness
- performance
- maintainability

So, maybe you can explore an alternative approach. Thank you.

calebporzio:
@claude - wanna address my messages on this PR? thank you

People-Sea:
Agreed, I’m looking for a better approach.

calebporzio:
thank you. the code looks nicer. However can you please let me know how you've addressed each of my concerns. I'd like to see your level of understanding and reasoning behind this before i feel comfortable merging thank you!

people-sea:
@calebporzio 

Yeah, I should have updated the explanation when I changed the code.

I think my first pass was looking at this from the wrong place. I was trying to answer “did this input change?” inside the directive, which is what led to the path handling, `$parent` cases, and the extra dirty-checking logic.

The approach I’m trying now is narrower: let the normal commit be created, then drop it right before Livewire turns queued commits into a request if that component has nothing to send. That keeps real pending updates working, but avoids the empty blur commit before it creates extra reactive/modelable child commits.

So the intended behavior is only to skip a model-triggered no-op commit. Plain `$commit()`, `$refresh()`, and real actions should keep behaving normally.

The part I’m still thinking through is whether the skip should use a separate preflight check like `hasUpdates()`, or be based on the same update preparation path that builds the request payload. The latter feels cleaner, especially for the performance and maintainability concerns you raised.

people-sea:
I’m going to close this implementation PR and open a smaller one with only the failing test. I’m not confident the current approach preserves the existing child commit semantics well enough.


Thanks for taking the time to review this.

PR #10249, opened 2026-04-28
[4.x] Allow overriding payload guards per component

FursanAbdulhak:
Overview: This PR introduces the ability to override global payload security limits on a per-component basis. This is particularly useful for complex components that may require deeper property nesting or more method calls than the global defaults defined in config/livewire.php.

Key Changes:

New Attributes: Added `#[MaxNestingDepth]` and `#[MaxCalls]` PHP attributes to allow declarative limit overrides.
Property Support: Added support for protected $maxNestingDepth and protected $maxCalls properties within component classes.
Mechanism Update: Updated HandleComponents to check for these overrides (Attribute first, then Property) before falling back to the global configuration.
Improved DX: Updated exception messages to inform developers about the new ways they can adjust these limits for specific components.
Testing: Added comprehensive unit tests to ensure overrides work correctly for both attributes and properties.

Example Usage:
```
use Livewire\Attributes\MaxNestingDepth;
use Livewire\Attributes\MaxCalls;
#[MaxNestingDepth(15)]
#[MaxCalls(100)]
class ComplexDashboard extends Component 
{
    // Or via property:
    // protected $maxNestingDepth = 15;
}
```

### DIFF
--- a/src/Attributes/MaxCalls.php
+++ b/src/Attributes/MaxCalls.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+class MaxCalls extends LivewireAttribute
+{
+    public function __construct(
+        public int $maxCalls
+    ) {}
+}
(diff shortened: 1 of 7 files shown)


calebporzio:
hmmm, I say for now just make the guards way more permissible or remove them entirely at a global config level. I added them as a cautionary measure. you're good to just bust through them thanks.

PR #10252, opened 2026-04-28
[4.x] Fix `#[Session]` attribute on non-primitive properties under JSON session serialisation

### LINKED ISSUE #10250: Session attribute on Collection property dont work with default json session serialization
### Livewire version

v4.2.4

### Laravel version

v13.6.0

### PHP version

8.3.30

Firefox on Arch Linux

For new applications, JSON is now the default config for session serialization.
laravel/laravel@75cef50

The #[Session] attribute on a Collection property will throw an error because JSON serialization turns the collection data to an array.

```blade
<?php

use Illuminate\Support\Collection;
use Livewire\Component;

new class extends Component {

    #[Session]
    public ?Collection $list = null;

    public function mount(): void
    {
        if(is_null($this->list)) {
            $this->list = collect();
        }
    }

    public function add(): void
    {
        $this->list->push(random_int(0, 100));
    }

};
?>

<div>
    <button wire:click="add">Add</button>

    <p>{{ $this->list->implode(', ') }}</p>
</div>
```

Image

Ideally, Collections should also work with JSON serialization. Otherwise, a note in the documentation that session serialization must be changed to "php" in the configuration file when using Collections would be very helpful.

joshhanley:
# The Scenario

Fresh Laravel applications (as of [laravel/laravel@75cef50](https://github.com/laravel/laravel/commit/75cef50)) use JSON session serialisation by default. With this default, `#[Session]` on a non-primitive property (e.g. `Collection`, `Carbon`, an Eloquent model) throws on the next request:

```
TypeError: Cannot assign array to property ::$list of type ?Illuminate\Support\Collection
```

```php
<?php

use Illuminate\Support\Collection;
use Livewire\Component;

new class extends Component {
    #[Session]
    public ?Collection $list = null;

    public function mount(): void
    {
        $this->list ??= collect();
    }

    public function add(): void
    {
        $this->list->push(random_int(0, 100));
    }
}
?>

<div>
    <button wire:click="add">Add</button>
    <p>{{ $this->list->implode(', ') }}</p>
</div>
```

# The Problem

`BaseSession` writes and reads the property value straight from the session without any marshalling. Under PHP serialisation, `serialize()`/`unserialize()` round-trips the object intact. Under JSON serialisation, `json_encode` strips the type (a `Collection` becomes its items, a `Carbon` becomes its ISO string), and `json_decode` brings it back as a plain array or string. The next request hands that decoded value to a typed property and PHP rejects the assignment.

# The Solution

Route the value through Livewire's existing synth-based dehydrate/hydrate pipeline before writing to / reading from the session. The synth system already produces JSON-safe tuples (`[data, meta]`) that round-trip through any session driver and reconstruct the original type on the way back, covering every type Livewire supports (`Collection`, `Carbon`, Eloquent models, `Stringable`, `BackedEnum`, custom collections, etc.).

`HandleComponents::dehydrateValue()` and `HandleComponents::hydrateValue()` are added as thin public wrappers around the existing internal pipeline so `BaseSession` can dehydrate or hydrate a single property value. `hydrateValue()` is a no-op for plain scalars, so primitive-typed `#[Session]` properties and existing PHP-serialised sessions keep working unchanged.

Note: this technically weakens Laravel's JSON session hardening (which exists to keep forged session data inert if `APP_KEY` leaks) because we're instantiating classes again on read. But that's only exploitable if an attacker has `APP_KEY`, and at that point they can already forge Livewire snapshots and hit the same hydrate pipeline on every request — so a Livewire app has bigger problems than `#[Session]` either way.

Fixes #10250

### DIFF
--- a/src/Features/SupportSession/BaseSession.php
+++ b/src/Features/SupportSession/BaseSession.php
@@ -47,7 +47,9 @@ class BaseSession extends LivewireAttribute
     {
         // Dehydrate to a JSON-safe tuple so the original type is restored
         // on read regardless of the session driver's serialisation format...
-        Session::put($this->key(), app(HandleComponents::class)->dehydrate($this->getValue(), new ComponentContext($this->component), ''));
+        $value = app(HandleComponents::class)->dehydrate($this->getValue(), new ComponentContext($this->component), '');
+
+        Session::put($this->key(), $value);
     }
 
     protected function key()
--- a/src/Features/SupportSession/BaseSession.php
+++ b/src/Features/SupportSession/BaseSession.php
@@ -36,26 +36,18 @@ class BaseSession extends LivewireAttribute
 
     protected function read()
     {
+        $value = Session::get($this->key());
+
         // Hydrate via Livewire's synth pipeline so non-primitive types
         // (Collection, Carbon, models) survive JSON-serialised sessions...
-        return app(HandleComponents::class)->hydrate(
-            Session::get($this->key()),
-            new ComponentContext($this->component),
-            '',
-        );
+        return app(HandleComponents::class)->hydrate($value, new ComponentContext($this->component), '');
     }
 
     protected function write()
     {
         // Dehydrate to a JSON-safe tuple so the original type is restored
         // on read regardless of the session driver's serialisation format...
-        $value = app(HandleComponents::class)->dehydrate(
-            $this->getValue(),
-            new ComponentContext($this->component),
-            '',
-        );
-
-        Session::put($this->key(), $value);
+        Session::put($this->key(), app(HandleComponents::class)->dehydrate($this->getValue(), new ComponentContext($this->component), ''));
     }
 
     protected function key()
--- a/src/Features/SupportSession/BaseSession.php
+++ b/src/Features/SupportSession/BaseSession.php
@@ -3,6 +3,7 @@
 namespace Livewire\Features\SupportSession;
 
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
+use Livewire\Mechanisms\HandleComponents\ComponentContext;
 use Livewire\Mechanisms\HandleComponents\HandleComponents;
 use Illuminate\Support\Facades\Session;
 use Attribute;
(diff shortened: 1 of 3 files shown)


src/Features/SupportSession/BaseSession.php
```diff
@@ -34,12 +36,20 @@ protected function exists()
 
     protected function read()
     {
-        return Session::get($this->key());
+        $value = Session::get($this->key());
+
+        // Hydrate via Livewire's synth pipeline so non-primitive types
+        // (Collection, Carbon, models) survive JSON-serialised sessions...
+        return app(HandleComponents::class)->hydrate($value, new ComponentContext($this->component), '');
```
ghabriel25, line 43: Hi, @joshhanley This means, primitive types also hydrated via Livewire's synth pipeline ?

joshhanley: @ghabriel25 yep, but that's not the issue, the issue is primitive types can json serialise fine but non-primitives can't.

ghabriel25: I means the performance impact. If it doesnt affect the performance that much then its good 👍

joshhanley: @ghabriel25 all Livewire properties go through synth system at the moment for every hydrate and dehydrate, that part hasn't changed. This is for properties with the session attribute, so the difference will be negligible.

calebporzio:
Hmmm I see the problem.

My hesitations with merging:
A) This introduces another pathway to our hydration system (which could introduce vulnerabilities (RCE stuff) if careless). I would want to make 1000% sure there is zero risk of that here.
B) Feels like we're leaking an abstraction with that surgical hydrate/dehydrate call. Don't see an obvious way around it but this feels like an AI fix and we should at least think for a bit about how this could fit harmoniously into our system. For example, we've had related issues with #[URL] properties and serializing/deserializing them. There might be a need for a new part of our system that can be used nicely by attributes like this?
C) Will impact performance right? probably no way around it, but we could at least do a check for the session serialization config and only do this if it's json? (thinking about eloquent model hydration cost)

I'll re-review when these concerns are addressed. Thanks @joshhanley !

calebporzio:
Closing for now. lets just open up a new PR to address my notes as well. thanks!

PR #10254, opened 2026-04-29
[4.x] Always recompile generated Blade files

ganyicz:
# The scenario

When a SFC is updated twice within the same second, it will not get recompiled:

```
5:30:00 PM - save ⚡example: foo
5:30:00 PM - visit /example: foo
5:30:00 PM - save ⚡example: bar
5:30:00 PM - visit /example: foo
```

# The problem

When Livewire compiles a SFC, it creates the view file with modification time set to -1 second:

| file | mtime |
| --- | --- |
| resources/views/components/⚡example.blade.php | 5:30:00 PM |
| storage/framework/views/livewire/classes/example.php | 5:30:00 PM |
| storage/framework/views/livewire/views/example.blade.php | 5:29:59 PM |

> Without this, the compiled file would never cache and recompile on every render because the compiled file is created in the same second as the source. Laravel's `isExpired` checks `filemtime($path) >= filemtime($compiledPath)` — equality passes this check, forcing recompilation, forever.

So when a blade file is saved within the same second as it was compiled, Livewire will make the source look like it was created 1 second ago, which makes the pre-existing compiled file look newer, skipping compilation.

# The solution

Force compilation by deleting the compiled file.

```diff
public function writeViewFile(string $sourcePath, string $contents): void
{
    $viewPath = $this->getViewPath($sourcePath);

    File::replace($viewPath, $contents);

    touch($viewPath, filemtime($viewPath) - 1);

+   File::delete($this->blade->getCompiledPath($viewPath));
}
```

> Also updates the compiler instance test in `ExtendBlade/UnitTest.php` to be narrower - instead of testing that the blade compiler instance is never resolved from the container, it now only tests that we never use `getPath()` from the new instance — which is the problematic pattern that causes [issues with Sentry](https://github.com/livewire/livewire/pull/10159). Without this test change we would have to thread the blade compiler to the CacheManager via constructor.

### DIFF
--- a/src/Compiler/CacheManager.php
+++ b/src/Compiler/CacheManager.php
@@ -4,12 +4,10 @@ namespace Livewire\Compiler;
 
 use Illuminate\Support\Facades\File;
 use Illuminate\Support\Str;
-use Illuminate\View\Compilers\BladeCompiler;
 
 class CacheManager
 {
     public function __construct(
-        public BladeCompiler $blade,
         public string $cacheDirectory,
     ) {}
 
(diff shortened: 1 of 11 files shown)


joshhanley:
Thanks!

PR #10265, opened 2026-05-06
[4.x] Fix unhandled console error when deferring components with non-div root

### LINKED ISSUE #2598: Console errors when using lazy/defer on components with flux:modal as root element
### Flux version

v2.14.1

### Livewire version

v4.2.4

### Tailwind version

v4.1.17

Firefox on Ubuntu

### What is the problem?

I load a modal with defer and it logs an error in the console. The modal seems to work fine but the error is annoying, also because it is very vague without any file reference.

I managed to work around it by not having the modal as root element, e.g. wrapping it in a `<div>`. But it would be nice if I don't have to have an unclear div lying around.

It doesn't show when:
- loading the component from another Livewire component (e.g. a regular Livewire page)
- or removing defer

I managed to bring it back to a small repro.

In a layout file (or maybe any file without Livewire itself or as parent):
```blade
<flux:modal.trigger name="wrapped-modal">
	<flux:button>Foo</flux:button>
</flux:modal.trigger>
<livewire:wrapped-modal defer />
```

In the wrapped-modal component:
```blade
<flux:modal name="wrapped-modal">
    Bar
</flux:modal>
```

The console shows `Uncaught (in promise) Component not found: undefined`, without any file reference. The modal doesn't need to open, it shows it directly on page load.

Image

The error shouldn't be there.

joshhanley:
# The Scenario

A Livewire component loaded with `defer` (or `lazy`) that has a non-`<div>` root element logs `Uncaught (in promise) Component not found: undefined` in the browser console once the component finishes loading. The component still works, but the error is noisy and references no source location.

```blade
<livewire:my-component defer />
```

```php
new class extends Component {
    public function render() { return <<<'HTML'
        <section id="my-component">
            Hello
        </section>
        HTML; }
}
```

# The Problem

When a deferred component finishes loading, Livewire morphs the placeholder `<div>` into the real root. Because the root tag differs, Alpine's `morph` falls back to `swapElements`, which clones the new root with `cloneNode(true)`. The clone does not carry custom JS properties, so the `__livewire` reference that `morph.js` set on the original target element is lost.

The morph config's `added` callback then calls `findComponentByEl(el).id`:

```js
added: (el) => {
    if (isntElement(el)) return

    const findComponentByElId = findComponentByEl(el).id

    trigger('morph.added', { el })
},
```

The clone has no `__livewire`, so `findComponentByEl` walks up looking for one. If the parent chain has no Livewire ancestor, the lookup returns `undefined`, `findComponent(undefined, true)` throws, and the throw surfaces as an unhandled promise rejection.

# The Solution

Remove the dead line. `findComponentByElId` is never read (leftover from the Slots refactor), and the `trigger('morph.added', { el })` call below it is preserved, keeping the public `Livewire.hook('morph.added', ({ el }) => { ... })` API contract intact.

The missing `__livewire` reference on the swapped root is harmless: as soon as `Alpine.morph` returns, Livewire's `wire:id` add/remove hooks fire, destroying the placeholder's `Component` and re-initialising one on the swapped root with `__livewire` set. The validation only ran during the brief gap between swap and re-init.

Fixes livewire/flux#2598


calebporzio:
Overall good that we're removing the dead code.

However, have some thoughts overall:

- this is a symptom of a deeper problem: when a placeholder has a different root element, the element gets swapped via morph and loses attributes like .__livewire causing a double init of the component. this can cause weird edge cases and problems.
- would be nice to have some kind of error around element type mismatches like this. probaby fine to have just a console.warn message somewhere that's like: hey, your placehollder and your main element should be the same root element. or something idk.

thanks!

joshhanley:
@calebporzio I believe this is no longer a problem, I fixed it / added support for it in PR #9310. So we do support using different elements for the placeholder vs the component root now. That PR set `wire:snapshot` and `wire:effects` on the new root before morph runs, and guarded `initComponent` so the destroy then re-init sequence works cleanly. The new root has everything needed to rebuild the Component from the snapshot.

That's why I only removed the invalid code in this PR.

I can still add the warn if you'd prefer it, but I think the change in this PR is good and all that is needed.

calebporzio:
interesting... i believe i had an ai pull this pr down and explore potential bugs caused by the double init and it found some. i think i destroyed the demo, but basically the diffferent elements cause a double init so if you dispatch an event in mount that calls something you can get a component to infinitiely loop requests? maybe something to just sanity check before merging? actually i'll just merge this, but if you get a sec please check.

joshhanley:
@calebporzio I had a dig into this and couldn't reproduce the loop. Had AI try dispatching from `mount` (simple and recursive), `@script` blocks calling `$wire`, and a few other variations on the swap path. The action fingerprint dedupe in `Message.addAction` squashes identical calls so they never escape the component, and I couldn't find a way around it. I'm happy that we're good here.

calebporzio:
great, thanks mate!

PR #10266, opened 2026-05-07
[4.x] Fix slots disappearing on lazy and deferred components

### LINKED ISSUE #10255: Slots don't work on lazy/defered components.
### Livewire version

v4

### Laravel version

v13

### PHP version

8.4

Chrome Mac

Create a very simple component with a view that uses the `{{$slot}}`, add the Lazy or Defer attribute to the component. Watch the slot disappear.

```html
<livewire:hello-world>
    <h1>Hello World From Slot</h1>
</livewire:hello-world>
```

```php
#[Lazy]
class HelloWorld extends Component
{
    public function render()
    {
        return view('livewire.hello-world');
    }
}
```

```html
<div>
    <div>Hello World From Component</div>
    {{ $slot }}
</div>
```

_No response_

I would expect Slots to behave the same regardless of lazy vs normal.

joshhanley:
# The Scenario

Slot content passed into a lazy or deferred component disappears. The child component itself loads correctly, but anything between its opening and closing tags is silently dropped.

```blade
<livewire:hello-world>
    <h1>Hello World From Slot</h1>
</livewire:hello-world>
```

```php
new class extends Component
{
    #[Lazy]
    public function render()
    {
        return <<<'HTML'
        <div>
            <div>Hello World From Component</div>
            {{ $slot }}
        </div>
        HTML;
    }
};
```

When the component lazy-loads, only `Hello World From Component` renders. `Hello World From Slot` never appears.

# The Problem

Slot content is captured by the parent at render time and passed to the child via `mount()`, where `$component->withSlots($slots, $parent)` attaches it to the child instance. For a normal component, the child's render is called next and `Slot::toHtml()` inlines the content into the child's HTML output.

For a lazy component, `SupportLazyLoading::mount` calls `skipMount()` and `skipRender()` to swap the child's render for placeholder HTML before the slot ever gets rendered. `SupportSlots::dehydrate` only persists slot **metadata** (`name`, `componentId`, `parentId`) into the child's memo, so by the time `__lazyLoad` fires and the child renders for real, the slot content is gone and `{{ $slot }}` produces empty fragment markers.

# The Solution

Reuse the existing `slotFragments` effect mechanism that's already used to deliver slot updates to a stubbed child on parent re-renders.

When the lazy child dehydrates during its initial mount (the moment the placeholder is being generated), `SupportLazyLoading::dehydrate` now also emits the child's own slots as `slotFragments` on its effects, alongside the existing `lazyLoaded`/`lazyIsolated` memos. So the slot HTML rides along in the lazy placeholder's `wire:effects` — no parent indirection, no encoded snapshot bloat, no new memo fields.

On the JS side, `js/features/supportSlots.js` now also listens on `effect` (not just `interceptMessage`), so `slotFragments` delivered in a component's own initial `wire:effects` are processed too. When a fragment arrives but its target slot markers don't exist yet (because the lazy placeholder hasn't been replaced), it's stashed against the target's component id. After each subsequent message's morph completes, any fragments stashed for that component are retried — so the moment the lazy load morph produces the matching slot markers, the stashed content is morphed in via the existing `morphFragment` helper.

Fixes #10255

### DIFF
--- a/js/features/supportSlots.js
+++ b/js/features/supportSlots.js
@@ -6,7 +6,13 @@ import { on } from '@/hooks'
 
 let pending = {}
 
-on('effect', ({ effects }) => (effects.slotFragments || []).forEach(f => renderSlot(null, f)))
+on('effect', ({ effects }) => {
+    let fragments = effects.slotFragments || []
+
+    fragments.forEach(async fragmentHtml => {
+        await renderSlot(null, fragmentHtml)
+    })
+})
 
 interceptMessage(({ message, onSuccess, onStream }) => {
     onSuccess(({ payload, onMorph }) => {
@@ -17,7 +23,12 @@ interceptMessage(({ message, onSuccess, onStream }) => {
                 await renderSlot(message.component, fragmentHtml)
             })
 
-            for (let f of pending[message.component.id] || []) await renderSlot(null, f)
+            let pendingFragments = pending[message.component.id] || []
+
+            pendingFragments.forEach(async fragmentHtml => {
+                await renderSlot(null, fragmentHtml)
+            })
+
             delete pending[message.component.id]
         })
     })
--- a/js/features/supportSlots.js
+++ b/js/features/supportSlots.js
@@ -2,23 +2,48 @@ import { extractFragmentMetadataFromHtml, extractInnerHtmlFromFragmentHtml, find
 import { interceptMessage } from '@/request'
 import { findComponent } from '@/store'
 import { morphFragment } from '@/morph'
+import { on } from '@/hooks'
 
-interceptMessage(({ message, onSuccess, onStream }) => {
+// Slot fragments that arrived before their target's slot markers existed in
+// the DOM (e.g. the target is a lazy child whose placeholder hasn't been
+// replaced yet). Keyed by the target component's id...
+let pendingSlotFragments = {}
+
+// Initial component mount: process any `slotFragments` effects delivered in
+// the parent's full-page render. This is the path lazy components rely on...
+on('effect', ({ component, effects }) => {
+    let fragments = effects.slotFragments || []
+
+    fragments.forEach(fragmentHtml => renderSlotOrStash(fragmentHtml))
+})
+
+// Subsequent messages: process this message's `slotFragments` effects, and
+// retry any fragments that were stashed for this component's id (e.g. its
+// markers have just appeared after a lazy load morph)...
+interceptMessage(({ message, onSuccess }) => {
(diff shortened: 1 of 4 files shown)


js/features/supportSlots.js
```diff
@@ -2,6 +2,17 @@ import { extractFragmentMetadataFromHtml, extractInnerHtmlFromFragmentHtml, find
 import { interceptMessage } from '@/request'
 import { findComponent } from '@/store'
 import { morphFragment } from '@/morph'
+import { on } from '@/hooks'
+
+let pending = {}
+
+on('effect', ({ effects }) => {
```
ghabriel25, line 9: Hi, @joshhanley I really want to understand Livewire js flow so I just ask here instead. Why you're listening outside interceptor instead using `onEffect` callback inside `onSuccess` callback? Appreciate it

joshhanley: @ghabriel25 no worries. It's mainly for consistency. You'll find all our internal effect listeners use on('effect'). Interceptor should work too.

ghabriel25: I see. Thanks for the information 👌

calebporzio:
Ok, I did some thinking on this one to see if there was a better way. Here's what I got: https://github.com/livewire/livewire/pull/10272

I think this is a better and more "pure" solution over all.

However, there is definite con: The slot contents passed to the lazy component will go BACK To the server when the lazy component fully renders (it will travel as part of the snapshot (in the memo)).

Slots can be pretty big so this could be bad performance wise. But it is in theory more pure because the lazy component render would actually have `$slot` props that render fully (like in a normal non-lazy scenario).

And the code is cleaner IMO. And no risk of JS memory leaks.

What do you think?

I think I'm ok with either approach (though i slightly prefer mine). If you think it's better to go the more performant route that you have, I'm down, but it hink the code can be refactored to be a little cleaner (making sure the skipped concept is isolated behind an abstraction of some sort and not spread out)

Thanks mate!

joshhanley:
@calebporzio yeah what you implemented was my first thought but didn't go with it because of the overhead. I think I'm the same as you, in two minds.

The javascript management side of things isn't ideal though, so I would actually lean towards your solution (it's cleaner and I prefer the change being in SupportSlots on the backend instead of SupportLazyLoading). I think the performance hit of sending the slot html back once is fine to keep it simple. We can always refactor it later if we need.

calebporzio:
Ok great, thanks @joshhanley - lets do that.

joshhanley:
@calebporzio yep, no worries, closing this in favour of #10272

PR #10267, opened 2026-05-07
[4.x] Fix CSS/JS module 404 by normalising component names with slashes

### LINKED ISSUE #10261: MFC sibling .css file not injected when component is resolved via component_namespaces (only works via component_locations)
### Livewire version

v4.3

### Laravel version

v12.58

### PHP version

5.4

Chrome on Windows

MFC (Multi-File Component) CSS co-location does not work when the component is resolved through a custom component_namespaces entry in config/livewire.php. The sibling .css file is silently ignored — no error, no warning, styles simply never reach the browser.
The exact same component with the exact same .css file works correctly when the component's directory is added to component_locations instead. This means the two resolution paths behave inconsistently: component_locations injects the sibling CSS, component_namespaces does not.

Steps to Reproduce

Register a custom namespace in config/livewire.php:

```
php'component_namespaces' => [
    'pages' => resource_path('views/pages'),
],
```

Create a MFC component under that path with a sibling .css file:

```
resources/views/pages/
└── my-component/
    ├── my-component.blade.php
    ├── my-component.php
    └── my-component.css   ← this file is never injected
```

Reference the component anywhere via its namespace:

blade`<livewire:dynamic-component component="pages::my-component" />`
{{-- or --}}
`<livewire:pages::my-component />`

Load the page and inspect the page source or browser `DevTools.`

Expected: The contents of my-component.css are injected into the page (same as documented MFC CSS co-location behavior).
Actual: The .css file is completely ignored. No styles are applied, no error is thrown.

Additional Context
The failure is completely silent. There is no exception, no log entry, and no browser console warning. This makes it very hard to diagnose — developers will assume their CSS is wrong and spend time debugging styles rather than the resolution path.
The documented MFC CSS co-location feature should work consistently regardless of how the component is resolved. If a component is found successfully (blade renders correctly), its sibling .css should be injected.

Directory structure used during investigation:
```
resources/views/pages/lastmile/invoices/
└── ⚡accountant-management/
    │   accountant-management.blade.php   ← CSS injected ✅ (page route)
    │   accountant-management.css
    │   accountant-management.php
    │
    └── modals/
        └── ⚡invoices-detail/
                invoices-detail.blade.php  ← CSS ignored ❌ (namespace-resolved child)
                invoices-detail.css
                invoices-detail.php
```
Parent loads the child as:
blade`<livewire:dynamic-component
    component="pages::lastmile/invoices/⚡accountant-management/modals/⚡invoices-detail"
    key="invoices-detail-modal" />`
The parent's CSS loads fine. The child's CSS never does.

_No response_

Workaround
Adding the same path to component_locations fixes the CSS injection:
```
php'component_locations' => [
    resource_path('views/components'),
    resource_path('views/livewire'),
    resource_path('views/pages'), // ← adding this makes the CSS work
],
```
However this is not a real fix — it requires duplicating the path in both config keys, and it may still not work for deeply nested child components loaded via <livewire:dynamic-component component="namespace::..."> inside a parent that was itself namespace-resolved.

joshhanley:
# The Scenario

A developer has multi-file components nested under bolt-prefixed directories at multiple levels, and references a deep child via `<livewire:dynamic-component>` with a slash-style component name. The component renders, but its sibling `.css` file is silently never loaded. No console warning, no exception, just no styles.

```blade
<livewire:dynamic-component
    component="pages::lastmile/invoices/⚡accountant-management/modals/⚡invoices-detail"
    key="invoices-detail-modal" />
```

# The Problem

The resolver accidentally tolerates non-canonical component names that contain `/` or `⚡` characters. The component renders fine and emits `wire:effects` with the slash-laden name intact. Downstream, the JS encoder at `js/features/supportCssModules.js` only substitutes `.`, `::`, and `:`:

```js
let encodedName = component.name
    .replace(/\./g, '--')
    .replace(/::/g, '---')
    .replace(/:/g, '----')
```

Slashes pass straight through into the asset URL. Laravel's `{component}` route parameter doesn't match across `/`, so the request 404s silently.

# The Solution

Component names should always be dotted with no `⚡` markers. Names that come in any other form (slash-separated paths, bolt-prefixed segments) are now rewritten to canonical form inside `Finder::normalizeName()`, just before the function returns:

```php
// Rewrite slash-style nested paths to canonical dot-notation and
// strip ⚡ markers so the same component is referenced by a
// single canonical name regardless of how it was written.
$nameComponentOrClass = preg_replace('/' . self::ZAP . '[\x{FE0E}\x{FE0F}]?/u', '', $nameComponentOrClass);
$nameComponentOrClass = str_replace('/', '.', $nameComponentOrClass);
```

That single rewrite happens before path resolution, name caching, and `$component->setName()`. So the canonical name is what flows through the snapshot, the asset URL, and the cache. The asset routes themselves are unchanged.

A developer who writes `pages::a/⚡b/⚡c` ends up with a component named `pages::a.b.c` everywhere it's surfaced.

Fixes #10261


calebporzio:
hmmm - i don't think we should be supporting that syntax for referencing a component in the first place right? we use dot notation and non zap emoji.

so if anything, i would expect this: `pages::lastmile/invoices/⚡accountant-management/modals/⚡invoices-detail` to be normalized to `pages::lastmile.invoices.accountant-management.modals.invoices-detail` internally.

So I say we don't take this patch ->where() route approach and instead ensure that component names are normalized internally.

make sense? thanks!

joshhanley:
@calebporzio yep, makes sense! I've updated it to normalise the name upfront and everything else now just works happily.

calebporzio:
great, thanks mate!

PR #10269, opened 2026-05-07
[4.x] Fix `$wire.$errors` reactivity when using `skipRender()`

### LINKED ISSUE #10263: $wire.$errors.all() manually triggers error DOM updates when skipRender() is used`
### Livewire version

4.2.4

### Laravel version

13.7

### PHP version

8.3

chrome on ubuntu 24.04

When using skipRender() inside the updated() lifecycle hook, validation errors are not reflected in the DOM through wire:show and wire:text.

However, if I manually access the error bag from JavaScript using: `$wire.$errors.all()`

inside a `$wire.intercept()` `onFinish()` callback, the DOM updates correctly and the validation error becomes visible.

So the behavior changes depending on whether `$wire.$errors.all()` is called or not.

- install a fresh livewire project 
- put the code in SFC compoent 
- interact with input you should see the error
- and now comment out the line `$wire.$errors.all()`
- and inertact with input again you will no longer see the error on dom

```php

use Livewire\Component;

new class extends Component {
    #[Validate(['required', 'min:5', 'max:10'])]
    public $title = '';

    public function updated()
    {
        $this->skipRender();
    }
};
```

```html
<div>
    <flux:input label="Title" wire:model.live="title"/>

    <div wire:show="$errors.has('title')">
        <span wire:text="$errors.first('title')"></span>
    </div>
</div>

<script>
    $wire.intercept(({ onFinish }) => {
        onFinish(() => {
            $wire.$errors.all()
        })
    })
</script>
```

https://github.com/user-attachments/assets/c7d446b3-b9d6-4d9d-98c6-b83618d339a6

In the above screenshot i am using the exact same code as the reproduce section

https://github.com/user-attachments/assets/baeda449-fa3d-4995-8f9a-da2af96caa93

on in the one the only change is i comment out this line

```js 
onFinish(() => {
            // $wire.$errors.all()
        })
```

as you can i no longer see the error on the dom

wire:show="$errors.has('title')" and wire:text="$errors.first('title')" should update consistently after validation, even when skipRender() is used.

Accessing $wire.$errors.all() manually from JavaScript should not be required to trigger the DOM update.

joshhanley:
@ganyicz suggested we make errors a reactive object on the component, I have done that in a alternative PR #10275

# The Scenario

When `skipRender()` is called inside a component (e.g. from `updated()`), validation errors arriving in the response are not reflected in the DOM through `wire:show` and `wire:text` directives that depend on `$errors`.

```php
<?php

use Livewire\Component;

new class extends Component {
    #[Validate(['required', 'min:5', 'max:10'])]
    public $title = '';

    public function updated()
    {
        $this->skipRender();
    }
}; ?>

<div>
    <input type="text" wire:model.live="title">

    <div wire:show="$errors.has('title')">
        <span wire:text="$errors.first('title')"></span>
    </div>
</div>
```

# The Problem

In `js/features/supportErrors.js`, `$wire.$errors` caches its messages in `Alpine.reactive` state via `clientErrors`. The cache is only invalidated lazily, *when something reads `messages()`*:

```js
messages() {
    if (component.__lastErrorsSnapshot !== component.snapshot) {
        state.clientErrors = null
        component.__lastErrorsSnapshot = component.snapshot
    }
    return state.clientErrors ??= component.snapshot.memo.errors
},
```

In the normal flow, the morph step after a request causes Alpine to re-evaluate bindings, which calls `messages()`, which triggers the reactive update. With `skipRender()`, no morph runs, so nothing reads `messages()` and `state.clientErrors` is never reassigned, leaving dependent effects with stale errors.

# The Solution

Register an `effect` hook in `supportErrors.js` that invalidates the reactive errors cache after every server response, so any Alpine effect depending on `$wire.$errors` (`wire:show`, `wire:text`, custom Alpine bindings) re-evaluates regardless of whether a morph runs:

```js
on('effect', ({ component }) => {
    if (! component.__errorsState) return

    component.__errorsState.clientErrors = null
})
```

The `__errorsState` guard skips components that have never read `$wire.$errors`, since the reactive state is created lazily.

Fixes #10263

### DIFF
--- a/js/component.js
+++ b/js/component.js
@@ -38,6 +38,8 @@ export class Component {
         // "reactive" is just ephemeral, except when you mutate it, front-ends like Vue react.
         this.reactive = Alpine.reactive(this.ephemeral)
 
+        this.errors = Alpine.reactive({ messages: this.snapshot.memo.errors ?? {} })
+
         this.queuedUpdates = {}
 
         this.jsActions = {}
@@ -84,6 +86,8 @@ export class Component {
         this.snapshotEncoded = snapshotEncoded
 
         this.snapshot = snapshot
+
+        this.errors.messages = snapshot.memo.errors ?? {}
 
         this.effects = effects
 
--- a/js/features/supportErrors.js
+++ b/js/features/supportErrors.js
@@ -1,5 +1,16 @@
 // This errors object has the most common methods from \Illuminate\Support\MessageBag class on the backend...
 import Alpine from 'alpinejs'
+import { on } from '@/hooks'
+
+// After every server response, invalidate the cached errors so any Alpine
+// effects depending on $wire.$errors (like wire:show / wire:text) re-read
+// the fresh errors from the new snapshot. Without this, components that
+// skip rendering would never trigger that re-read...
+on('effect', ({ component }) => {
+    if (! component.__errorsState) return
+
+    component.__errorsState.clientErrors = null
+})
 
 export function getErrorsObject(component) {
     let state = component.__errorsState ??= Alpine.reactive({
--- a/js/features/supportErrors.js
+++ b/js/features/supportErrors.js
@@ -1,35 +1,8 @@
 // This errors object has the most common methods from \Illuminate\Support\MessageBag class on the backend...
-import Alpine from 'alpinejs'
-import { on } from '@/hooks'
-
-// After every server response, invalidate the cached errors so any Alpine
-// effects depending on $wire.$errors (like wire:show / wire:text) re-read
-// the fresh errors from the new snapshot. Without this, components that
-// skip rendering would never trigger that re-read...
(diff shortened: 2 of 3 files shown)


ganyicz:
What if we make errors an actual reactive object on the component?

```js
// js/component.js

constructor(el) {
    this.errors = Alpine.reactive({ messages: this.snapshot.memo.errors ?? {} })
}

mergeNewSnapshot() {
    this.errors.messages = snapshot.memo.errors ?? {}
}
```

Then the entire `supportErrors.js` could be simplified:

SCR-20260508-qrfk

This fixes the issue from this PR and seems to pass all the existing tests.

joshhanley:
@ganyicz yeah I like that. I wonder why it wasn't already a reactive object in the first place? 🤔 Will submit a different PR with that, as an option.

PR #10272, opened 2026-05-08
[4.x] Fix slots disappearing on lazy/deferred components

### LINKED ISSUE #10255: Slots don't work on lazy/defered components.

calebporzio:
## Summary

Slot content passed into a lazy or deferred component disappears after the component loads. The child renders correctly but `{{ $slot }}` produces empty output.

```blade
<livewire:child lazy>
    <h1>This disappears</h1>
</livewire:child>
```

Fixes #10255

## The problem

During a normal mount, slot content flows through cleanly: the parent captures it via `ob_start()`, `withSlots()` attaches it to the child, the child renders `{{ $slot }}`, and `Slot::toHtml()` outputs it wrapped in fragment markers.

For lazy components, `SupportLazyLoading::mount()` calls `skipMount()` and `skipRender()` before the child ever renders. The slot content is attached to the component but never rendered. Then `SupportSlots::dehydrate()` saves only slot **metadata** (name, componentId, parentId) into the memo — not the HTML content. When `__lazyLoad` fires and the child hydrates, it reconstructs empty `PlaceholderSlot` objects from the metadata, and `{{ $slot }}` renders nothing.

The content is lost in the dehydrate → hydrate cycle because the memo was never designed to carry it.

## The fix

When a component's render is skipped (the `skipRender` flag is set), `SupportSlots::dehydrate()` now also persists the slot's HTML `content` in the memo alongside the existing metadata. On hydrate, if `content` is present, full `Slot` objects are restored instead of empty `PlaceholderSlot`s. The `__lazyLoad` render then calls `{{ $slot }}` and gets real content.

Three files changed:

- **`SupportSlots.php`** — `dehydrate()` conditionally includes `content` when render was skipped; `hydrate()` restores full `Slot`s when content is present
- **`HandlesSlots.php`** — new `withHydratedSlots()` method that creates `Slot` or `PlaceholderSlot` per entry depending on whether content exists
- **`BrowserTest.php`** — test that slot content survives lazy loading

## Why this approach

The alternative is to emit slot HTML as a `slotFragments` effect on the lazy placeholder and handle it in JS — stashing the fragments until `__lazyLoad` morphs in the real DOM, then retrying them. That works but introduces a new JS-side timing pattern (stash + retry after morph) and renders slot content in two passes, which can flash empty.

This approach keeps the fix entirely in PHP within the existing render cycle. The slot content rides in the snapshot for exactly one round-trip (the `__lazyLoad` request), then on subsequent renders the parent takes over delivering fresh content via `slotFragments` as usual. No JS changes, no timing complexity.

## Test plan

- [ ] Browser test: `test_slot_content_is_preserved_on_lazy_component`
- [ ] Manual: render a `#[Lazy]` component with slot content, verify it appears after lazy load
- [ ] Manual: render a `#[Defer]` component with slot content, same verification
- [ ] Verify normal (non-lazy) slot behavior is unchanged

### DIFF
--- a/src/Features/SupportSlots/HandlesSlots.php
+++ b/src/Features/SupportSlots/HandlesSlots.php
@@ -55,5 +55,18 @@ trait HandlesSlots
         return $this;
     }
 
+    public function withHydratedSlots(array $slots): self
+    {
+        foreach ($slots as $slot) {
+            if (isset($slot['content'])) {
+                $this->slots[] = new Slot($slot['name'], $slot['content'], $slot['componentId'], $slot['parentId']);
+            } else {
+                $this->slots[] = new PlaceholderSlot($slot['name'], $slot['componentId'], $slot['parentId']);
+            }
+        }
+
+        return $this;
+    }
+
 
 }
--- a/src/Features/SupportSlots/HandlesSlots.php
+++ b/src/Features/SupportSlots/HandlesSlots.php
@@ -58,11 +58,7 @@ trait HandlesSlots
     public function withHydratedSlots(array $slots): self
     {
         foreach ($slots as $slot) {
-            if (isset($slot['content'])) {
-                $this->slots[] = new Slot($slot['name'], $slot['content'], $slot['componentId'], $slot['parentId']);
-            } else {
-                $this->slots[] = new PlaceholderSlot($slot['name'], $slot['componentId'], $slot['parentId']);
-            }
+            $this->slots[] = new Slot($slot['name'], $slot['content'], $slot['componentId'], $slot['parentId']);
         }
 
         return $this;
--- a/src/Features/SupportSlots/SupportSlots.php
+++ b/src/Features/SupportSlots/SupportSlots.php
@@ -5,6 +5,7 @@ namespace Livewire\Features\SupportSlots;
 use Livewire\ComponentHook;
 
 use function Livewire\on;
+use function Livewire\store;
 
 class SupportSlots extends ComponentHook
 {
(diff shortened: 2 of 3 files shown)


src/Features/SupportSlots/SupportSlots.php
```diff
@@ -44,11 +45,17 @@ public function renderIsland($name, $view, $properties)
...
+
+        // If slot content was persisted (e.g. from a lazy component that skipped its
+        // initial render), restore full Slots so {{ $slot }} renders the content...
+        $hasContent = collect($slots)->contains(fn ($s) => isset($s['content']));
+
+        if ($hasContent) {
+            $this->component->withHydratedSlots($slots);
+        } else {
             $this->component->withPlaceholderSlots($slots);
         }
```
ghabriel25: `withHydratedSlots` already has `content` check so why not just calling that instead redundant check
```php
$this->component->withHydratedSlots($slots);
```

src/Features/SupportSlots/HandlesSlots.php
```diff
@@ -55,5 +55,18 @@ public function withPlaceholderSlots(array $slots): self
         return $this;
     }
 
+    public function withHydratedSlots(array $slots): self
+    {
+        foreach ($slots as $slot) {
```
ghabriel25, line 60: This check could be placed here
```php
$hasContent = collect($slots)->contains(fn ($s) => isset($s['content']));

if ($hasContent) {
    // loop to create new slot
} else {
    $this->withPlaceholderSlots($slots);
}

return $this;

calebporzio:
@joshhanley - would you mind taking this PR the rest of the way? thanks mate

src/Features/SupportSlots/HandlesSlots.php
joshhanley: @ghabriel25 the loop needs to check each one, one by one, so can't use the collection contains here. We could probably use each or something like that, but I think the foreach reads clearly.

joshhanley:
@calebporzio have updated this PR. It's good for review!

calebporzio:
thanks!

PR #10276, opened 2026-05-11
[4.x] Extract synth orchestration into a `HandleSynths` mechanism

joshhanley:
# The Scenario

`HandleComponents` currently owns both the component lifecycle (`mount`, `snapshot`, `update`, `render`) and the synth-based serialisation pipeline (registry, resolution, recursive dehydrate/hydrate). Any feature (like `#[Session]`) that wants to use synths at the moment has to reference the `HandleComponents` mechanism to do it. Came up during review of #10252.

# The Solution

This PR extracts the synths system into its own mechanism class `Livewire\Mechanisms\HandleSynths\HandleSynths`.

`Livewire::propertySynthesizer(...)` and `Livewire::findSynth(...)` now delegate to `HandleSynths`, so the public API is unchanged.

This will let the fix for #10250 leverage `HandleSynths` directly, which is cleaner than reaching into `HandleComponents`.

One thing to note: all the built-in synth classes (`CarbonSynth`, `CollectionSynth`, the `Synth` base, etc.) are still in their original location (`Livewire\Mechanisms\HandleComponents\Synthesizers\*`). The docs demonstrate extending `Synth` from userland, so moving them would be a breaking change. We could move them and add shell classes at the old location if you'd prefer.

### DIFF
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -3,9 +3,8 @@
 namespace Livewire\Mechanisms\HandleComponents;
 
 use function Livewire\{on, store, trigger, wrap };
-use ReflectionUnionType;
 use Livewire\Mechanisms\Mechanism;
-use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
+use Livewire\Mechanisms\HandleSynths\HandleSynths;
 use Livewire\Exceptions\PublicPropertyNotFoundException;
 use Livewire\Exceptions\MethodNotFoundException;
 use Livewire\Exceptions\MaxNestingDepthExceededException;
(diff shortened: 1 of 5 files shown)


ghabriel25:
This looks cleaner 👍 

Make sense to me to move all synth-classes to `HandleSynths` and create shell-class at old location

calebporzio:
Looking good. Pretty straightforward.

One performance note: calling `app(HandlesSynths::class)->` is meaningfully slower than just `$this->`. So If we can, let's instantiate and pass HandlesSynths into HandlesComponents in the service provider so that it can directly access the instance without going to the container every time it needs it. (especially in mount and especially in loops)

src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -16,23 +15,16 @@
...
-        Synthesizers\FloatSynth::class
-    ];
-
-    // Performance optimization: Cache which synthesizer matches which type
-    protected $synthesizerTypeCache = [];
-
     public static $renderStack = [];
     public static $componentStack = [];
 
+    // Holding the HandleSynths instance directly avoids resolving it from the
```
calebporzio: this comment feels unescessary

calebporzio:
thanks!

PR #10280, opened 2026-05-12
[4.x] Add "@return $this" docblocks to MakesAssertions for static analysis tools to work better in userland tests

imliam:
Running PHPStan over a project's tests directory causes some issues with Livewire tests - because the assertions have no return type:

```
 ------ ------------------------------------------ 
  Line   tests/Feature/ApiTokensSettingsTest.php   
 ------ ------------------------------------------ 
  :121   Cannot call method call() on mixed.       
         🪪  method.nonObject                      
  :122   Cannot call method assertSet() on mixed.  
         🪪  method.nonObject                      
 ------ ------------------------------------------ 
```

As `Testable` has docblocks denoting `@return $this`, I figure it's also appropriate to add them to the `MakesAssertions` trait too, so it covers that section of userland-used code.

### DIFF
--- a/src/Features/SupportTesting/MakesAssertions.php
+++ b/src/Features/SupportTesting/MakesAssertions.php
@@ -8,7 +8,9 @@ use Illuminate\Support\Arr;
 
 trait MakesAssertions
 {
-    /** @return $this */
+    /**
+     * @return $this
+     */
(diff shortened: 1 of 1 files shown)


calebporzio:
thanks mate!

PR #10283, opened 2026-05-14
[3.x] Fix file uploads hanging silently on network-level failures

### LINKED ISSUE #10281: File Upload makeRequest missing onerror handler causes silent hang on S3 upload network/CORS failures
### Livewire version

4.0

### Laravel version

v13.*

### PHP version

4.x

Chrome on Windows

##  Description
When uploading directly to S3 (`LIVEWIRE_TEMPORARY_FILE_UPLOAD_DISK=s3`), any network-level failure — including a non-existent bucket, missing CORS configuration, or dropped connection — causes the upload to hang indefinitely with no feedback. The livewire-upload-error event never fires, and the $wire.upload() error callback is never called.

## Root cause

In `makeRequest` (dist/livewire.js), the S3 XHR only registers two event listeners:

```JS
request.upload.addEventListener("progress", (e) => { ... });
request.addEventListener("load", () => { ... });
```

The load event only fires when the server returns a response. CORS failures, non-existent buckets, and network errors fire the error event instead — which has no handler. The error2 function that dispatches livewire-upload-error exists in the codebase but is never wired to `request.addEventListener("error", ...)` for the S3 upload path.

## Steps to reproduce

  1. Configure `LIVEWIRE_TEMPORARY_FILE_UPLOAD_DISK=s3` with an invalid or non-existent bucket name
  2. Use `$wire.upload()` or `wire:model` on a file input
  3. Select a file to upload

## Expected behavior

`livewire-upload-error` dispatches and/or the `$wire.upload()` error callback fires, allowing the UI to surface the failure to the user.

## Actual behavior

The upload silently hangs at 0% progress. No error event fires. No callback is called. The UI has no way to detect or report the failure.

##  Suggested fix

Add an error listener alongside the existing load listener in **makeRequest**:

```JS
request.addEventListener("error", () => {
    this.component.$wire.call("_uploadErrored", name, null, this.uploadBag.first(name).multiple);
});
```

Please see description for further info

_No response_

Please see description for further info

joshhanley:
# The Scenario

When uploading a file directly to S3, any network-level failure of the upload request causes the upload to hang indefinitely with no feedback. A missing CORS configuration on the bucket, a non-existent bucket, or a dropped connection all leave the loading state stuck on screen. `livewire-upload-error` never fires, the error callback is never called, and the user has no idea anything went wrong.

For example, with a bucket that has no CORS rule for the app's origin, selecting a file in the component below leaves "Uploading…" showing forever:

Screenshot 2026-05-14 at 06 08 33PM

```php
use Livewire\Component;
use Livewire\WithFileUploads;

class UploadPhoto extends Component
{
    use WithFileUploads;

    public $photo;

    public function render()
    {
        return <<<'HTML'
        <div>

            <div wire:loading wire:target="photo">Uploading…</div>

            <span>@error('photo') {{ $message }} @enderror</span>
        </div>
        HTML;
    }
}
```

# The Problem

`makeRequest()` in `js/features/supportFileUploads.js` registers only two listeners on the `XMLHttpRequest`:

```js
request.upload.addEventListener('progress', e => { /* ... */ })

request.addEventListener('load', () => { /* ... */ })
```

The `load` event only fires once an HTTP response is received, so it correctly handles S3 responding with a `4xx`/`5xx` status. But network-level failures, such as CORS blocks, DNS failures, and dropped connections, fire the `error` event instead, and there is no `error` listener. The request is left hanging, so the upload bag entry is never shifted, the loading state never clears, and neither `_uploadErrored` nor any callback runs.

Both `handleSignedUrl` (local `POST`) and `handleS3PreSignedUrl` (direct-to-S3 `PUT`) flow through `makeRequest`, so both are affected.

# The Solution

Add an `error` event listener to `makeRequest` that calls `_uploadErrored` with `null` errors:

```js
request.addEventListener('error', () => {
    this.component.$wire.call('_uploadErrored', name, null, this.uploadBag.first(name).multiple)
})
```

This mirrors the existing non-`2xx` branch of the `load` handler, which already calls `_uploadErrored` with `null` errors for any non-`422` response.

Because both upload paths share `makeRequest`, the single listener covers local and direct-to-S3 uploads alike.

Screenshot 2026-05-14 at 06 05 55PM

---

This was reported against v4, but the same `makeRequest` code exists on `3.x`, so the fix is being made here first and will be up-merged into v4.

Fixes #10281

### DIFF
--- a/js/features/supportFileUploads.js
+++ b/js/features/supportFileUploads.js
@@ -211,6 +211,10 @@ class UploadManager {
             this.component.$wire.call('_uploadErrored', name, errors, this.uploadBag.first(name).multiple)
         })
 
+        request.addEventListener('error', () => {
+            this.component.$wire.call('_uploadErrored', name, null, this.uploadBag.first(name).multiple)
+        })
+
         this.uploadBag.first(name).request = request
 
         request.send(formData)


calebporzio:
makes sense, thanks mate

PR #10284, opened 2026-05-15
[4.x] Prevent `wire:model.live.blur` from triggering redundant requests

### LINKED ISSUE #19764: ->live(onBlur: true) ignores partial rendering constraints when state is not updated
### Package

filament/schemas

### Package Version

v5.6.1

### Laravel Version

v13.6.0

### Livewire Version

v4.2.4

### PHP Version

PHP 8.5.3

### Problem description

When a field is set to `live()` with `onBlur: true`,
any restrictions on partial rendering configured through `partiallyRenderComponentsAfterStateUpdated()`
will be ignored if the field is blurred without updating the state.

### Expected behavior

The field would partially render the components it's configured to when it loses focus.
The field would _not_ render other components when it loses focus.

### Steps to reproduce

- Navigate to the User resource in the reproduction repository
- Create a new User (or Edit the test user)
- In the `belowContent()` of both the `email` and `password` fields, a `Closure` is configured to display the value of the user `name`. However, the `name` field is set to only partially render the `email` field.
- Change the `name` field
- Have the field lose focus
- Observe how the `email` field's `belowContent()` changed, but `password`'s does not.
- Focus the `name` field anew
- Have the field lose focus without changing the state of the field
- Oserve how both the `email` and `password` fields' `belowContent()` is updated

### Reproduction repository (issue will be closed if this is not valid)

https://github.com/gp-lnuff/filament-issue/tree/live-onblur

### Relevant log output

```shell

```

people-sea:

Yes. This adds coverage for `wire:model.live.blur` sending another request when the value has not changed. I did not open a discussion first because this PR only adds a failing test.

Yes.

No.

Yes.

This adds a browser test for an unchanged `wire:model.live.blur` input.

The first blur sends the expected model update. The second blur leaves the value unchanged, but still sends another Livewire request. This PR only captures that behavior so the request handling can be fixed separately.

### DIFF
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -1,4 +1,5 @@
 import { directive } from '@/directives'
+import { checkDirty } from '@/directives/wire-dirty'
 import { handleFileUpload } from '@/features/supportFileUploads'
 import { findComponentByEl } from '@/store'
 import { dataGet, dataSet } from '@/utils'
@@ -95,7 +96,17 @@ directive('model', ({ el, directive, component, cleanup }) => {
 
     // Network event listeners (for modifiers after .live, or .lazy backwards compat)
     if (shouldSendNetwork && networkOnBlur) {
-        bindings['@blur'] = () => update()
+        bindings['@blur'] = () => {
+            // Wait a microtask so an ephemeral blur sync (`.blur.live.blur`) lands before the dirty check...
+            queueMicrotask(() => {
+                let target = expression.startsWith('$parent') ? component.parent : component
+
+                // Skip the request if nothing has changed since the last server sync...
+                if (target && ! checkDirty(target)) return
+
+                update()
+            })
+        }
     }
 
     if (shouldSendNetwork && networkOnChange) {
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for the failing test @people-sea — good artifact. I pushed a fix to your branch.

The gap was in `wire-model.js`: the `@blur` network binding called `update()` unconditionally, so an unchanged field still sent a `$commit` on every blur. The fix reuses `checkDirty()` from `wire-dirty` and skips the request when nothing on the component has changed since the last server sync. Two deliberate choices:

- The check is component-wide rather than scoped to the bound property, so a blur still flushes other deferred `wire:model` properties that are waiting to piggyback on the next request.
- The check runs in a microtask so that ephemeral blur syncs (`wire:model.blur.live.blur`) land in the reactive state before we decide whether to skip.

Your test plus the other 18 `wire_model` browser tests in `SupportDataBinding` pass locally. Might be worth retitling the PR now that it carries the fix.

people-sea:
Thanks @calebporzio! I tested the fix and it resolved my issue. Appreciate it!

PR #10288, opened 2026-05-16
[4.x] Fix `RequireLivewireHeaders` middleware applied twice

### LINKED ISSUE #10287: `RequireLivewireHeaders` middleware applied twice on subsequent request
### Livewire version

v4.3

### Laravel version

v13.0

### PHP version

v8.4

Chrome on windows

When testing some mechanism, I found that **RequireLivewireHeaders** applied twice for page component action or subsequent request by lazy/defer page component.

This only happened when not using custom update endpoint. Otherwise, all is good.

**General page component with action**
- create page component
- call any action

**Lazy/Defer page component**
- create page component
- use `#[Lazy]` or `#[Defer]` attribute

inspect both cases middleware using https://github.com/fruitcake/laravel-debugbar

**General page component with action**
| --- | --- |
| Image | Image |

**Lazy/Defer page component**

Image

It should only applied once.

If **RequireLivewireHeaders** middleware only for Livewire request, why not just include this middleware in **PersistentMiddleware** class

ghabriel25:
`RequireLivewireHeaders` middleware always applied twice on component action or lazy/defer page component

### Files changes
- src/Mechanisms/HandleRequests/HandleRequests.php
- src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php

### Summary
- Remove `RequireLivewireHeaders` middleware on `boot()` method in `HandleRequests` class
- Add `RequireLivewireHeaders` as persistent middleware in `PersistentMiddleware` class

### Before

https://github.com/user-attachments/assets/31fe7eb8-70a8-4c8b-9470-11bba270ff43

### After

https://github.com/user-attachments/assets/65a3a7c7-2a6b-43ab-9d82-cc8e2d4a010d

Fix: https://github.com/livewire/livewire/issues/10287

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -25,7 +25,7 @@ class HandleRequests extends Mechanism
         if (! $this->updateRoute && ! $this->updateRouteExists()) {
             app($this::class)->setUpdateRoute(function ($handle) {
                 return Route::post(EndpointResolver::updatePath(), $handle)
-                    ->middleware(['web', RequireLivewireHeaders::class])
+                    ->middleware('web')
                     ->name('default-livewire.update');
             });
         }
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -25,7 +25,7 @@ class HandleRequests extends Mechanism
         if (! $this->updateRoute && ! $this->updateRouteExists()) {
             app($this::class)->setUpdateRoute(function ($handle) {
                 return Route::post(EndpointResolver::updatePath(), $handle)
-                    ->middleware('web')
+                    ->middleware(['web', RequireLivewireHeaders::class])
                     ->name('default-livewire.update');
             });
         }
--- a/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
+++ b/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
@@ -22,6 +22,7 @@ class PersistentMiddleware extends Mechanism
         \Illuminate\Auth\Middleware\Authenticate::class,
         \Illuminate\Auth\Middleware\Authorize::class,
         \App\Http\Middleware\Authenticate::class,
+        \Livewire\Mechanisms\HandleRequests\RequireLivewireHeaders::class
     ];
 
     protected $path;
--- a/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
+++ b/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
@@ -22,7 +22,6 @@ class PersistentMiddleware extends Mechanism
         \Illuminate\Auth\Middleware\Authenticate::class,
         \Illuminate\Auth\Middleware\Authorize::class,
         \App\Http\Middleware\Authenticate::class,
-        \Livewire\Mechanisms\HandleRequests\RequireLivewireHeaders::class
     ];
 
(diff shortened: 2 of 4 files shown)


ghabriel25:
@calebporzio @joshhanley I'm not sure how to test this but happy to add some test with guidance.

src/Mechanisms/HandleRequests/HandleRequests.php
```diff
@@ -25,7 +25,7 @@ function boot()
         if (! $this->updateRoute && ! $this->updateRouteExists()) {
             app($this::class)->setUpdateRoute(function ($handle) {
                 return Route::post(EndpointResolver::updatePath(), $handle)
-                    ->middleware(['web', RequireLivewireHeaders::class])
```
joshhanley, line 28: @ghabriel25 this should remain here.

src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
```diff
@@ -22,6 +22,7 @@ class PersistentMiddleware extends Mechanism
         \Illuminate\Auth\Middleware\Authenticate::class,
         \Illuminate\Auth\Middleware\Authorize::class,
         \App\Http\Middleware\Authenticate::class,
+        \Livewire\Mechanisms\HandleRequests\RequireLivewireHeaders::class
```
joshhanley: See if there is a way we can exclude the `RequireLivewireHeaders` middleware for persistent middleware as it shouldn't be applied to the faux request.

joshhanley (changes requested):
@ghabriel25 thanks for the PR!

The middleware must be applied to the update endpoint, so are you able to change it back?

Instead it should be excluded from the persistent middleware so it doesn't run on the faux request and only the update request.

Re testing: I'm not sure if we have persistent middleware tests, but we if do, try seeing if we already have a test for excluding middleware from the faux request or not.

src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
ghabriel25: Note that. I'll try reject the middleware from persisten middleware

ghabriel25:
@joshhanley Need clearer mind to solve this. I'm gonna close this for now.

joshhanley:
No worries! Take your time

ghabriel25:
This is really embarassing for me. I just close other PR after testing this PR.

For this PR: After some debug using logger inside **`RequireLivewireHeaders`**, it only triggered once per request. It seems a bug from Laravel Debugbar, not from Livewire

joshhanley:
@ghabriel25 no worries!

ghabriel25:
Hi, @joshhanley Could you check this failing test? I tried this on `HandleRequests` unit test and failed
```php
public function test_require_livewire_headers_middleware_is_not_duplicated_on_update(): void
{
    $testable = Livewire::test(new class extends TestComponent {});
    $encodedSnapshot = json_encode($testable->snapshot);

    // Livewire's update route should still be matched, not the catch-all
    $response = $this->withHeaders(['X-Livewire' => 'true'])
        ->postJson(EndpointResolver::updatePath(), ['components' => [
            ['snapshot' => $encodedSnapshot, 'updates' => [], 'calls' => []],
        ]]);

    $response->assertOk();

    $route = collect(Route::getRoutes()->getRoutes())->first(function ($route) {
        return $route->getName() === 'default-livewire.update';
    });

    $middlewareCount = count(
        array_filter($route->middleware(), fn ($m) => $m === RequireLivewireHeaders::class)
    );

    $this->assertEquals(1, $middlewareCount);
}
```

PR #10289, opened 2026-05-16
[4.x] Re-order custom route endpoint middleware

ghabriel25:
## Scenario
For general livewire update route endpoint, the middleware always returns
```php
array:2 [▼
    0 => "web"
    1 => "Livewire\Mechanisms\HandleRequests\RequireLivewireHeaders"
]
```
However, for custom update route endpoint, the middleware order becomes
```php
array:2 [▼
    0 => "Livewire\Mechanisms\HandleRequests\RequireLivewireHeaders"
    1 => "web"
]
```

## Problem
In `HandleRequests::setUpdateRoute()`, `RequireLivewireHeader` was append first then `web` middleware. This PR to ensure that middleware always applied at same order for general update endpoint or custom.
```php
{

    // Ensure the header guard middleware is always present, even on custom routes.
    $route->middleware(RequireLivewireHeaders::class);

    // Ensure the route includes the `web` middleware group.
    // Without it, CSRF protection is lost entirely on the update endpoint.
    // Note: we use middleware() (not gatherMiddleware()) to avoid polluting
    // the route's computed middleware cache before it's fully configured.
    if (! in_array('web', $route->middleware())) {
        $route->middleware('web');
    }

    ...
}
```

## Solution
Change the order middleware applied. First must be `web` middleware then `RequireLivewireHeader`.
```php
{

    // Ensure the route includes the `web` middleware group.
    // Without it, CSRF protection is lost entirely on the update endpoint.
    // Note: we use middleware() (not gatherMiddleware()) to avoid polluting
    // the route's computed middleware cache before it's fully configured.
    if (! in_array('web', $route->middleware())) {
        $route->middleware('web');
    }    

    // Ensure the header guard middleware is always present, even on custom routes.
    $route->middleware(RequireLivewireHeaders::class);
    ...
}
```

### File changes

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -92,9 +92,6 @@ class HandleRequests extends Mechanism
     {
         $route = $callback([self::class, 'handleUpdate'], EndpointResolver::updatePath());
 
-        // Ensure the header guard middleware is always present, even on custom routes.
-        $route->middleware(RequireLivewireHeaders::class);
-
         // Ensure the route includes the `web` middleware group.
         // Without it, CSRF protection is lost entirely on the update endpoint.
         // Note: we use middleware() (not gatherMiddleware()) to avoid polluting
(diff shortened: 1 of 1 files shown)


joshhanley (approved):
@ghabriel25 thanks for the PR! Makes sense to me.

calebporzio:
good call thanks

PR #10290, opened 2026-05-17
[4.x] Fix checking empty array against `is_null` instead empty-array comparison

ghabriel25:
## Problem
In **PersistentMiddleware** class, applicable middleware was checked using `is_null` eventhough it returns an array
```php
protected function applyMiddleware()
{
    ...

    $middleware = $this->getApplicablePersistentMiddleware($request);

    // Only send through pipeline if there are middleware found
    if (is_null($middleware)) return;

    ...
}

{

    if (! $route) return [];

    $middleware = app('router')->gatherRouteMiddleware($route);

    return $this->filterMiddlewareByPersistentMiddleware($middleware);
}

protected function filterMiddlewareByPersistentMiddleware($middleware)
{
    $middleware = collect($middleware);

    $persistentMiddleware = collect(app(PersistentMiddleware::class)->getPersistentMiddleware());

    return $middleware
        ->filter(function ($value, $key) use ($persistentMiddleware) {
            return $persistentMiddleware->contains(function($iValue, $iKey) use ($value) {
                // Some middlewares can be closures.
                if (! is_string($value)) return false;

                // Ensure any middleware arguments aren't included in the comparison
                return Str::before($value, ':') == $iValue;
            });
        })
        ->values()
        ->all();
}
```
This returns **false** therefore this check will always passed

## Solution
Change the checking using comparison
```php
if (is_null($middleware) || $middleware === []) return;
```

## File Changes
- `src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php`

### DIFF
--- a/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
+++ b/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
@@ -115,7 +115,7 @@ class PersistentMiddleware extends Mechanism
         $middleware = $this->getApplicablePersistentMiddleware($request);
 
         // Only send through pipeline if there are middleware found
-        if (is_null($middleware)) return;
+        if ($middleware === []) return;
 
         Utils::applyMiddleware($request, $middleware);
 
--- a/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
+++ b/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
@@ -112,6 +112,7 @@ class PersistentMiddleware extends Mechanism
 
         $request = $this->makeFakeRequest();
 
+        // If no middleware found, this returns `[]`
         $middleware = $this->getApplicablePersistentMiddleware($request);
 
         // Only send through pipeline if there are middleware found
--- a/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
+++ b/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
@@ -116,7 +116,7 @@ class PersistentMiddleware extends Mechanism
         $middleware = $this->getApplicablePersistentMiddleware($request);
 
         // Only send through pipeline if there are middleware found
-        if ($middleware === []) return;
+        if (is_null($middleware) || $middleware === []) return;
 
         Utils::applyMiddleware($request, $middleware);
 


src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
```diff
@@ -112,10 +112,11 @@ protected function applyPersistentMiddleware()
 
         $request = $this->makeFakeRequest();
 
+        // If no middleware found, this returns `[]`
         $middleware = $this->getApplicablePersistentMiddleware($request);
 
         // Only send through pipeline if there are middleware found
-        if (is_null($middleware)) return;
+        if ($middleware === []) return;
```
joshhanley: I wonder if it would be better to leave the null check there too?

joshhanley (changes requested):
@ghabriel25 thanks for the PR! I wonder if it would be better to leave the `is_null()` check there and just add the empty array check (or just use `empty($middleware)`)?

src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
ghabriel25: Make sense. I'll push it right away

ghabriel25: Wait, the variable is not the same. Other PR use property. Should this PR closed?

joshhanley: @ghabriel25 not sure what you mean. But you should leave this fix open, because if that PR doesn't get accepted, it will get lost. If anything this PR should be reviewed and merged/ closed, before that one is reviewed.

ghabriel25: @joshhanley I mean the other PR
```php
$this->applicableMiddleware = $this->getApplicablePersistentMiddleware($request);

if ($this->applicableMiddleware === []) return;
```
while this PR
```php
$middleware = $this->getApplicablePersistentMiddleware($request);

if ($middleware === []) return;
```

joshhanley: @ghabriel25 that PR isn't the concern of this PR, so I would update this PR accordingly. If that PR needs updating separately, then do that too.

It's generally why we try not to have the same fix in two PRs, better to keep them as two separate issues.

ghabriel25: Thanks, now I understand. I have pushed new commit there to restore original `null` check

calebporzio:
agree re: keeping null check in addition to new `[]` check

ghabriel25:
Done 👍

calebporzio:
Thanks @ghabriel25 !

ghabriel25:
@calebporzio Great! I still need more context though, is it for backward compatibility? (for future reference) 😅

joshhanley:
@ghabriel25 as we support Laravel L10-L13, we're never sure if there might be a difference back in L10 than what it is now in L13. So we just prefer to add the change, but not remove, just in case. Especially when people might not be even running the latest L10 and there could have been a change in an older point release.

ghabriel25:
@joshhanley I see. So prioritize adding changes rather than remove existing flow if possible 👍

joshhanley:
@ghabriel25 yep! Sometimes we have to change/remove things and that's fine, we just have to be more careful when doing so. For a small PR like this, theres no negative to leaving the null in, so just easier to do so.

PR #10292, opened 2026-05-17
[4.x] Fix authorization checked twice when applied on route and component action

### LINKED ISSUE #10291: Authorization checked twice when applied on route and component action
### Livewire version

v4.3

### Laravel version

v13.0

### PHP version

v8.4

Chrome on Windows

When using `#[Authorize]` attribute on component (child component) action and the same authorization middleware already applied on route level. The Authorization getting checked twice.

The root cause was there are no guard in **BaseAuthorize::call()** whether ability and model already applied on persistend middleware.

```php
<?php

use App\Models\User;

Route::livewire('/', 'pages::home')
    ->can('create', User::class)
    // ->middleware('can:create,'. User::class)
    ->name('home');
```
**Page component**
```blade
<div>
    <livewire:child />
</div>
```
**Child component**
```php
<?php

use App\Models\User;
use Livewire\Attributes\Authorize;
use Livewire\Component;

new class extends Component
{
    #[Authorize('create', User::class)]
    public function check(): void
    {
        //
    }
};
```
```blade
<div>
    <flux:button color="blue"
                 variant="primary"
                 wire:click="check">Test persistent middleware</flux:button>
</div>
```

**`Without route model binding`**
| Before Action | After action |
| --- | --- |
| Image | Image |

**`With route model binding`**
| --- | --- |
| Image | Image |

It should only applied once

ghabriel25:
## Scenario
When user applied authorization on route level as middleware for page component then inside child component also using **`#[Authorize]`** attribute with the same ability and class/model. The authorization checked twice See (https://github.com/livewire/livewire/issues/10291). This is confirmed using debugbar and logger inside policy and turns out its logged twice.

## Problem
In **`BaseAuthorize::call()`** there are no guard against authorization middleware already applied using persistent middleware therefore its getting checked twice in single request

## Solution
- Store applicable middleware before applying it inside **PersistentMiddleware** class
- Reset the applicable middleware on `flush-state` callback
- Create dedicate function to retrieve only authorization middleware from applicable middleware
- Before running authorization in **`BaseAuthorize::call()`**, check if authorization middleware already has the same ability and class/model.
- If its the same `ability` and `class` / `model`, early return;

## Files Changes
- `src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php`

If there are more clean way to solve this. Feel free to close this, thanks!

Fix: https://github.com/livewire/livewire/issues/10291

### DIFF
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -34,13 +34,7 @@ class BaseAuthorize extends LivewireAttribute
         $arguments = Arr::wrap($this->argument);
         
         // Check if authorization already applied on route level
-        $abilityModel = app(PersistentMiddleware::class)->getAuthorizeMiddleware();
-        
-        if ($abilityModel !== []) {
-            [$ability, $model] = $abilityModel;
-            
-            if (enum_value($this->ability) === $ability && in_array($model, $arguments)) return;
-        }
+        if ($this->isAlreadyAuthorizedByMiddleware($arguments)) return;
 
         // Resolve each argument (prioritize method parameters first, then component properties)
         $resolved = [];
@@ -51,9 +45,6 @@ class BaseAuthorize extends LivewireAttribute
         Gate::authorize($this->ability, $resolved);
     }
 
-    /**
-     * Resolve a single argument.
-     */
     protected function resolveArgument(string $arg, array $parameters): mixed
     {
         // Action that does not require a model, for example a 'create' action...
@@ -79,5 +70,41 @@ class BaseAuthorize extends LivewireAttribute
 
         // Fall back to component property
         return data_get($this->component, $arg);
+    }
+
+    protected function isAlreadyAuthorizedByMiddleware($arguments): bool
+    {
+        $middleware = app(PersistentMiddleware::class)->getAuthorizeMiddleware();
+
+        if ($middleware === []) {
+            return false;
+        }
+
+        return collect($middleware)
+            ->map(fn (string $m) => $this->parseMiddlewareArgument($m))
+            ->filter()
+            ->contains(fn (array $item) =>
+                $item['ability'] === enum_value($this->ability) 
+                && in_array($item['model'], $arguments, true)
+            );
+    }
+
+    protected function parseMiddlewareArgument(string $middleware): ?array
+    {
+        $parts = explode(':', $middleware, 2);
+        if (count($parts) !== 2) {
+            return null;
+        }
+
+        $abilityAndModel = explode(',', $parts[1], 2);
+
+        if (count($abilityAndModel) !== 2) {
+            return null;
+        }
+
+        return [
+            'ability' => trim($abilityAndModel[0]),
+            'model'   => trim($abilityAndModel[1]),
+        ];
     }
 }
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -3,10 +3,8 @@
 namespace Livewire\Features\SupportAuthorization;
 
 use Attribute;
-use Illuminate\Auth\Middleware\Authorize;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Gate;
-use Illuminate\Support\Str;
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
 use Livewire\Mechanisms\PersistentMiddleware\PersistentMiddleware;
 use UnitEnum;
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -10,7 +10,6 @@ use Livewire\Mechanisms\PersistentMiddleware\PersistentMiddleware;
 use UnitEnum;
 
 use function Illuminate\Support\enum_value;
-use function Livewire\str;
 
 #[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_METHOD)]
 class BaseAuthorize extends LivewireAttribute
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -11,6 +11,7 @@ use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
 use Livewire\Mechanisms\PersistentMiddleware\PersistentMiddleware;
 use UnitEnum;
 
+use function Illuminate\Support\enum_value;
 use function Livewire\str;
 
 #[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_METHOD)]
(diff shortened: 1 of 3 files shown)


src/Features/SupportAuthorization/BaseAuthorize.php
```diff
@@ -3,11 +3,17 @@
...
+use Illuminate\Auth\Middleware\Authorize;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Gate;
+use Illuminate\Support\Str;
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
+use Livewire\Mechanisms\PersistentMiddleware\PersistentMiddleware;
 use UnitEnum;
 
+use function Illuminate\Support\enum_value;
+use function Livewire\str;
```
ghabriel25: Todo: remove unused import

src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
```diff
@@ -71,6 +75,38 @@ function getPersistentMiddleware()
...
+    {
+        $middleware = $this->applicableMiddleware;
+
+        if ($middleware === []) {
+            return [];
+        }
+
+        $authorizeMiddleware = Arr::first($middleware, function ($m) {
+            return Str::startsWith($m, Authorize::class);
+        });
```
ghabriel25: Todo: use `Arr::where()` instead

ghabriel25 (commented):
To do: remove unused import

ghabriel25:
More test coming to check if multiple **`#[Authorize]`** used in component, only deduplicates the one that has same ability and model/class.

joshhanley:
@ghabriel25 thanks for the PR! Can you clarify for me why we need this? Why can't the developer just use one or the other?

ghabriel25:
Hi, @joshhanley This comes to me when using nested child component where page component already applied the authorization. Because the "**same authorization on both levels**" case in Livewire (deep nested components) is surprisingly common and annoying:

- Developers naturally put route protection for the whole page.
- They also put **`#[Authorize]`** on actions inside nested children for explicitness / self-documenting code / future-proofing.
- Because it's nested, often developer forgot (or maybe its just me) that this is already applied on route level

I understand that **`#[Authorize]`** attribute was for explicit authorization. However, this still open some _neglectfullness_ on developers side.

src/Features/SupportAuthorization/BaseAuthorize.php
```diff
@@ -65,4 +69,37 @@ protected function resolveArgument(string $arg, array $parameters): mixed
...
+            return false;
+        }
+
+        return collect($middleware)
+            ->map(fn (string $m) => $this->parseMiddlewareArgument($m))
+            ->filter()
+            ->contains(fn (array $item) =>
+                $item['ability'] === enum_value($this->ability) 
+                && in_array($item['model'], $arguments, true)
+            );
```
ghabriel25: Parse the attribute ability and argument properties into string auhtorization middleware and check using `in_array` instead running through collection.

```php
use Illuminate\Auth\Middleware\Authorize as AuthorizeMiddleware;

...

$authorizeMiddleware = AuthorizeMiddleware::using($this->ability, ...$arguments);

return in_array($authorizeMiddleware, $middleware);
```
reference: https://github.com/laravel/framework/blob/13.x/src/Illuminate/Routing/Attributes/Controllers/Authorize.php

ghabriel25:
Failing tests from workflow but passed on local
- Livewire\Features\SupportTesting\UnitTest::test_testing_does_not_create_dynamic_properties_on_test_response

> ErrorException: Since symfony/http-foundation 8.1: Directly setting property "headers" of "Illuminate\Http\Response" is deprecated; pass the header bag as a constructor argument instead.

Laravel already has fix for this (See https://github.com/laravel/framework/pull/60318)

joshhanley (changes requested):
@ghabriel25 thanks! This PR concerns me. Have a look at Codex's findings below. I'm not sure stopping the double authorisation check is worth the potential security risks. But if you feel you can get it working happily, then I'd be open to it.

--- 
_Codex review_

**Scenario**

This PR is trying to avoid a duplicate authorisation check when the same ability is applied at both the route level and the Livewire action level.

For example, the route may already protect the page with `can:edit,post`:

```php
Route::livewire('/posts/{post}', EditPost::class)
    ->can('edit', 'post');
```

Then the component action may also declare the same check with `#[Authorize]`:

```php
#[Authorize('edit', 'post')]
public function save(Post $post): void
{
    //
}
```

On the initial page load, Laravel runs the route middleware normally. On later Livewire requests, Livewire replays persistent route middleware so the update request is still protected.

After that, Livewire runs method attributes before calling the component method. `BaseAuthorize::call()` resolves the attribute arguments and calls `Gate::authorize()` again.

So the duplicate check this PR is addressing is real when the route middleware and action attribute are protecting the exact same ability and the exact same model.

**Issue With The Current PR**

The concern is that the current dedupe check compares the middleware string before resolving the real action arguments. It sees this:

```php
can:edit,post
#[Authorize('edit', 'post')]
```

and assumes they are the same authorisation.

But in Livewire, `post` in the attribute does not always mean “the route `{post}`”. `BaseAuthorize` intentionally resolves method parameters first, then component properties.

That makes this case risky:

```php
Route::livewire('/posts/{post}', EditPost::class)
    ->can('edit', 'post');

#[Authorize('edit', 'post')]
public function save(Post $post): void
{
    //
}
```

If the page was loaded at `/posts/1`, the route middleware authorised post `1`.

But a Livewire action request can pass `post: 2` to the action. Current Livewire resolves the method parameter to post `2` and runs the attribute authorisation against post `2`.

This PR would skip that second check because the strings still match: `edit,post`. That changes the security semantics from “authorise the actual action argument” to “trust that the route already authorised something with the same parameter name”.

**Ideal Direction**

I think the safe version of this fix is not “skip when the middleware strings match”. It should be “skip only when we can prove the same ability and the same resolved gate arguments have already been authorised during this request”.

I would structure it like this:

1. Let `PersistentMiddleware` keep track of successful route-level `Authorize` decisions for the current Livewire update request.

2. Store them as structured data, not middleware strings:

```php
[
    'ability' => 'edit',
    'arguments' => [$resolvedRoutePost],
]
```

3. Flush that state on `flush-state`, alongside the existing persistent middleware request state.

4. In `BaseAuthorize::call()`, keep the current argument resolution logic. That means:
   - no argument becomes `[]`
   - class names stay class strings
   - method parameters are resolved before component properties
   - component properties are used as fallback

5. After resolving the attribute’s actual arguments, ask `PersistentMiddleware` whether this exact authorisation has already passed.

6. Only return early when:
   - the ability matches after enum normalisation
   - the argument count matches
   - every argument is equivalent
   - Eloquent models compare by model identity, for example same class, connection, table, and key, ideally via `Model::is()`

If anything is uncertain, run `Gate::authorize()` again. A harmless duplicate check is better than skipping an action-level authorisation that may be protecting a different model.

The tests should include the happy path from the report, but also a regression where `/posts/1` is route-authorised and the action receives `post: 2`. That request must still be forbidden.

ghabriel25:
@joshhanley Thanks for the review.

I'll add another test for this case and see if it fails. 

I do agree that it only check `post` as string argument, not model key and this is security risk because action argument can be tampered from client so if I edit my post then I pass another key to the action, it bypass the authorization.

I think its better to leave it as it is. Adding additional check for something that doesnt do too much request payload seems cumbersome so I'll close this for now.

PR #10293, opened 2026-05-17
[4.x] Add support to `wire:navigate` for falling back to browser redirect for external links

MrPunyapal:
## Summary

`wire:navigate` now only intercepts same-origin HTTP(S) page visits. External URLs and browser-native link types (for example, `target="_blank"`, `download`, `mailto:`, `tel:`) fall back to the browser's default behavior.

## Problem

Absolute external URLs were being normalized to same-origin paths and fetched by the SPA loader, which caused links like:

```html
<a href="https://github.com/pinkary-project/pinkary.com" target="_blank" wire:navigate>
  Source Code
</a>

// will open https://pinkary.com/pinkary-project/pinkary.com instead of actual link
```

to open a wrong same-origin path instead of the intended external destination.

## Solution

Before prefetching or intercepting a click, Livewire checks whether the link should be handled by the SPA navigation. Links are bypassed when any of the following apply:

- Destination origin differs from the current origin (cross-origin).
- The anchor has a `target` attribute that is not `_self`.
- The anchor has a `download` attribute.
- The URL uses a non-HTTP(S) protocol (`mailto:`, `tel:`, `javascript:`, etc.).

## Implementation (high level)

- Added `shouldHandleLinkWithNavigate(linkEl, destination)` which returns `true` only for eligible same-origin page visits.
- The `wire:navigate` directive uses this helper for both hover-prefetch and click handling, so browser-native links are never prevented.
- The low-level link handler accepts an optional `shouldIntercept` callback to avoid preventing default behavior for native links.


ghabriel25:
@MrPunyapal Is there a reason using `wire:navigate` on external link?

joshhanley:
@MrPunyapal thanks for the PR! Yeah I have the same question as @ghabriel25, why are you using `wire:navigate` on external links? Just remove `wire:navigate` and then there is no problems.

MrPunyapal:
@joshhanley there was a case in pinkary where we were using drop-down. And while updating new design we moved source code link to drop-down and it was using drop-down item component which had hard-coded wire:navigate (as might not be aware of such things when we created that) and it started behaving incorrectly. Yeah we had to add new extra unwanted prop just to make sure external links do not use navigate and we need to remember this and even find such cases and fix them manually.

And than I shared this issue with community on social media and many community members asked it to PR as should work by default.

calebporzio:
Mmmm kinda makes sense actually. 

Is this a breaking change for any reason right now? 

Also this PR implementation feels a little bit too much (docs might not be necessary or maybe just a quicker mention? idk..) thanks

MrPunyapal:
@calebporzio I'm sure there are no breaking changes.
I removed the docs as well.

docs/wire-navigate.md
```diff
@@ -15,6 +15,8 @@ Below is a simple example of adding `wire:navigate` to links in a nav bar:
 
 When any of these links are clicked, Livewire will intercept the click and, instead of allowing the browser to perform a full page visit, Livewire will fetch the page in the background and swap it with the current page (resulting in much faster and smoother page navigation).
 
+`wire:navigate` only intercepts same-origin page visits. External links, links with a non-`_self` target, download links, and URLs such as `mailto:` or `tel:` automatically fall back to the browser's default behavior.
```
calebporzio, line 18: Good docs note 👌

js/plugins/navigate/index.js
```diff
@@ -41,10 +41,17 @@ export default function (Alpine) {
 
         let preserveScroll = modifiers.includes('preserve-scroll')
 
+        let shouldHandleLink = () => {
```
calebporzio: The current shape of this code makes it such that we have to litter this "shouldHandleLink" concern around. 

It would be much cleaner to have a single early return right here.

Only question is: can we check the "destination" this early in the lifecycle.

It's possible we deliberately wait to check the destination until a click or something.

Possibly that's because `x-bind:` or something might set it.

Anyhow, I think we should try a cleaner, more centralized implementation and add tests for cases we suspect might trip it up.

js/plugins/navigate/links.js
```diff
@@ -75,6 +81,19 @@ export function createUrlObjectFromString(urlString) {
     return urlString !== null && new URL(urlString, document.baseURI)
 }
 
+export function shouldHandleLinkWithNavigate(linkEl, destination = extractDestinationFromLink(linkEl)) {
```
calebporzio: this method might benefit from a few simple comments or a single one up top explaining what cases it's looking for.

js/plugins/navigate/links.spec.js
```diff
@@ -0,0 +1,56 @@
+import { describe, expect, it } from 'vitest'
```
calebporzio: In general we avoid unit tests for little methods like this and prefer to test them outside-in using Dusk (like you also have).

These unit tests are useful for driving out a method, but would make the test suite incredibly bloated and rigid if we did it for everything.

Therefore, i suggest we drop unit tests here, cover what we can in a single dusk test, and if anything else goes untested and causes a bug later, a patch will be made with a test to cover it.

src/Features/SupportNavigate/BrowserTest.php
```diff
@@ -319,6 +319,31 @@ public function test_can_navigate_to_another_page_with_hash_fragment()
         });
     }
 
+    public function test_navigate_is_not_triggered_for_external_links_with_a_target()
```
calebporzio: you can cover more cases than just external if you want. maybe there's a way to get this single test to test all the cases you just added. Like downloads and such...

calebporzio (changes requested):
Thanks for the improvements. Got a few more notes for you.

MrPunyapal:
@calebporzio, thanks for reviewing my PR I have updated the code based on your suggestions.

js/plugins/navigate/index.js
```diff
@@ -44,7 +44,7 @@ export default function (Alpine) {
         shouldPrefetchOnHover && whenThisLinkIsHoveredFor(el, 60, () => {
             let destination = extractDestinationFromLink(el)
 
-            if (! destination) return
+            if (! shouldHandleLinkWithNavigate(el, destination)) return
```
calebporzio: I think this method should be called:

"linkShouldBeHandledNatively(el, destination)" or something like that

This reads much more clearly about what the method determines, and as a bonus, it would be inverted so you would remove the `!` which would make the code even more clear.

src/Features/SupportNavigate/BrowserTest.php
```diff
@@ -319,6 +319,40 @@ public function test_can_navigate_to_another_page_with_hash_fragment()
         });
     }
 
+    public function test_navigate_falls_back_to_browser_for_native_links()
```
calebporzio: This test is fine. It is a LITTLE fake-the-worldy for my taste – for example, it would be nice if it didn't intercept things like mailto and download and let them work and assert real outcomes, but i understand that can be a little labor intensive to pull off and maybe not worth it. I'm fine with this if it's not simple to make it more real-world/real-behavior.

I know most LLMs kinda suck at this. They love mocking and faking things and not testing the actual real result of something for some reason.

calebporzio (changes requested):
Thanks for the changes! I think this PR is ALMOST there. Just two quick notes and then we can merge. Thanks @MrPunyapal

calebporzio:
Thanks for addressing that stuff. Code is looking great! However lots of failing tests related to navigate stuff. Maybe try running that stuff locally and seeing if it fails then looking into fixes thanks

MrPunyapal:
@calebporzio yeah, just finding time to look into it
I'll fix and ping you soon!

calebporzio:
Any update @MrPunyapal ? thanks!

MrPunyapal:
All tests are padding locally now.

calebporzio:
Thanks @MrPunyapal - so happy to get this across the line!

PR #10294, opened 2026-05-18
Fix view method call in attribute-layout.md

KevinWinardi:
There is a typo in the path: it should be 'pages', not 'livewire'. Since the file name contains the ⚡ emoji, we also need to explicitly reference it in the view method: view('pages.posts.⚡index'). 
For single-file components, though, I think using $this->view() is the better practice.

### DIFF
--- a/docs/attribute-layout.md
+++ b/docs/attribute-layout.md
@@ -77,7 +77,7 @@ use Livewire\Component;
 new class extends Component {
     public function render()
     {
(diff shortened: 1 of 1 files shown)


joshhanley:
@KevinWinardi thanks! Yep using `$this->view()` is better

KevinWinardi:
You're welcome  sir.

PR #10295, opened 2026-05-18
[4.x] Defer Alpine init per-component while script module loads

joshhanley:
Alternative to [#9861](https://github.com/livewire/livewire/pull/9861) (Solution A). Same fixes, different approach: keeps `Livewire.start()` synchronous and confines the change to per-component `_x_ignore` juggling, rather than pre-loading all modules ahead of `Alpine.start()`.

# The Scenario

`Alpine.data()` registered inside a single-file component's `<script>` block isn't available when Alpine evaluates `x-data` on that component. The browser console shows:

> `Uncaught ReferenceError: testAlpineData is not defined`

Happens on initial load, after `wire:navigate`, on dynamic child mounts, and when a `#[Lazy]` component hydrates.

```php
<?php

    //
};
?>

<div x-data="testAlpineData">
    <div x-text="message"></div>
</div>

<script>
    Alpine.data('testAlpineData', () => ({
        message: 'alpine-data-loaded',
    }))
</script>
```

# The Problem

`Alpine.start()` walks the DOM synchronously, but component script modules are loaded via async `import()`. Alpine evaluates `x-data="testAlpineData"` before `Alpine.data()` has been registered.

# The Solution

Defer Alpine initialisation on a per-component basis while that component's module is loading.

- `js/lifecycle.js` — in `interceptInit`, if the new component's module is pending, set `_x_ignore` on the root, register a callback to clear it and `Alpine.initTree(el)` once the import resolves, and return early so wire: directive triggers fire on the re-init pass.
- `js/features/supportJsModules.js` `effect` hook — for already-initialised components (lazy hydration / update), set `_x_ignore` on the live root before morph runs, then after the import clear `_x_ignore` and `_x_marker` and call `Alpine.initTree`.
- `js/features/supportJsModules.js` `interceptClone` — Alpine's morph plugin runs `cloneNode(from, to)` which calls `initTree(to)` on the detached `to` element where `_x_ignore` on the live `from` doesn't reach. Mirror the flag onto `to` for any element under a component with a pending module.

Trade-off: a brief flash of uninitialised HTML between morph applying and the module resolving. Solution A avoided this by pre-loading modules upfront; this PR accepts it in exchange for a synchronous boot lifecycle.

Fixes https://github.com/livewire/livewire/discussions/9591
Fixes https://github.com/livewire/livewire/discussions/9830
Fixes https://github.com/livewire/livewire/discussions/9737


calebporzio:
Ok, I like this direction. I have some thoughts on the specifics of the implementation but overall this is the best move considering all the tradeoffs.

joshhanley:
@calebporzio ok sweet! I've push some changes to the PR and now it's ready for review.

PR #10298, opened 2026-05-19
[4.x] Fix duplicate `RequireLivewireHeaders` middleware

### LINKED ISSUE #10287: `RequireLivewireHeaders` middleware applied twice on subsequent request

ghabriel25:
Based on https://github.com/livewire/livewire/pull/10288

## Scenario
Using [Laravel debugbar](https://github.com/fruitcake/laravel-debugbar), each time component updates (action, lazy, or defer) **`RequireLivewireHeaders`** appended twice for single Livewire request. 

I tried to check if its really Livewire issue using logger inside this middleware, however the log only triggered once. I thought maybe this is debugbar issue. But after that, I try to create new test for this case and it actually failed. So I'm convinced that it really appended twice.

## Problem
In `setUpdateRoute()` , there was no guard to check if its exists in route middleware stack before append it. It always append it regardless if its already exists or not therefore its getting duplicated (only for default Livewire route)

```php
{
    ...

    // There is no guard if its already exists in middleware stack so code below always append it.

    // Ensure the header guard middleware is always present, even on custom routes.
    $route->middleware(RequireLivewireHeaders::class);

    ...
}
```

This case is okay if custom routes was registered because developers never register this middleware on route level. However thats not the case if its default route. This middleware already registered along with `web` middleware here:

```php
function boot()
{
    // Register the default route immediately (before routes files load)
    // so it's positioned before any catch-all routes.
            return Route::post(EndpointResolver::updatePath(), $handle)
                ->middleware(['web', RequireLivewireHeaders::class])
                ->name('default-livewire.update');
        });
    }

    $this->skipRequestPayloadTamperingMiddleware();
}
```

## Solution
Check if its exists before append it to middleware stack
```php
// Ensure the header guard middleware is always present, even on custom routes.
// Only append if its not exists on current middleware stack
if (! in_array(RequireLivewireHeaders::class, $route->middleware())) {
    $route->middleware(RequireLivewireHeaders::class);
}
```

This will ensure **RequireLivewireHeaders** middleware not duplicated for both default and custom route

## Files Changes
- `src/Mechanisms/HandleRequests/UnitTest.php` (new test added)

Fix: https://github.com/livewire/livewire/issues/10287

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -101,7 +101,10 @@ class HandleRequests extends Mechanism
         }
 
         // Ensure the header guard middleware is always present, even on custom routes.
-        $route->middleware(RequireLivewireHeaders::class);
+        // Only append if its not exists on current middleware stack
+        if (! in_array(RequireLivewireHeaders::class, $route->middleware())) {
+            $route->middleware(RequireLivewireHeaders::class);
+        }
 
         // Append `livewire.update` to the existing name, if any.
         if (! str($route->getName())->endsWith('livewire.update')) {
--- a/src/Mechanisms/HandleRequests/RequireLivewireHeaders.php
+++ b/src/Mechanisms/HandleRequests/RequireLivewireHeaders.php
@@ -16,14 +16,6 @@ class RequireLivewireHeaders
             abort(404);
         }
 
-        // Remove duplicate middleware from middleware stack
-        // Without this, `RequireLivewireHeaders` will be applied twice
-        if (($route = $request->route()) && isset($route->action['middleware'])) {
-            $route->action['middleware'] = array_unique(array_values(
-                array_filter($route->middleware(), fn ($m) => is_string($m))
-            ));
-        }
-
         return $next($request);
     }
 }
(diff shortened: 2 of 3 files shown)


ghabriel25:
Hi, @calebporzio @joshhanley . Finally find the root cause where its appended twice to middleware stack. Please review, thanks!

calebporzio:
Perfect, thanks for getting to the bottom of this @ghabriel25 - your explanation is crystal clear.

ghabriel25:
@calebporzio Great! 👍

PR #10302, opened 2026-05-21
[4.x] Fix cursor pagination crash on an empty page

### LINKED ISSUE #10297: cursor pagination error
### Livewire version

v4.3

### Laravel version

v13.8

### PHP version

8.5

Chrome on macOS

Error
vendor/livewire/livewire/src/Features/SupportPagination/views/simple-tailwind.blade.php:24
Call to a member function encode() on null

LARAVEL
13.8.0
PHP
8.4.5
UNHANDLED
CODE 0

Error
vendor/livewire/livewire/src/Features/SupportPagination/views/simple-tailwind.blade.php:24
Call to a member function encode() on null

LARAVEL
13.8.0
PHP
8.4.5
UNHANDLED
CODE 0

_No response_

no

joshhanley:
# The Scenario

If a component uses cursor pagination and is on a page other than the first, and the result set then becomes empty, for example records are deleted, a filter is applied, or a stale cursor URL is opened, Livewire throws an error instead of rendering the empty page:

Screenshot 2026-05-21 at 10 31 01AM

```blade
<?php

use App\Models\User;
use Livewire\Component;
use Livewire\WithPagination;

new class extends Component {
    use WithPagination;

    public function deleteLatest()
    {
        User::where('id', '>', 3)->delete();
    }

    public function with()
    {
        return ['users' => User::orderBy('id')->cursorPaginate(3, ['*'], 'page')];
    }
}; ?>

<div>
    <button type="button" wire:click="deleteLatest">Delete latest</button>

    @foreach ($users as $user)
        <p wire:key="user-{{ $user->id }}">{{ $user->name }}</p>
    @endforeach

    {{ $users->links() }}
</div>
```

# The Problem

The simple pagination views call `->encode()` on the cursor directly:

```blade
wire:click="setPage('{{ $paginator->previousCursor()->encode() }}', '{{ $paginator->getCursorName() }}')"
```

`previousCursor()` and `nextCursor()` are derived from the page's items, so they return `null` when the page is empty. But the buttons still render in that state: the previous button shows whenever `onFirstPage()` is `false`, and the next button shows whenever `hasMorePages()` is `true`. So on an empty page the view calls `->encode()` on `null` and throws the error.

# The Solution

Capture the cursor and fall back to the current cursor (`$paginator->cursor()`) when it's `null`.

```blade
@php($previousCursor = $paginator->previousCursor() ?? $paginator->cursor())

wire:click="setPage('{{ $previousCursor?->encode() }}', '{{ $paginator->getCursorName() }}')"
```

This mirrors Laravel's own pagination views, which never dereference the cursor directly. They navigate by URL through the null-safe `previousPageUrl()`/`nextPageUrl()` helpers, which return `null` when there's no cursor:

```php
public function previousPageUrl()
{
    if (is_null($previousCursor = $this->previousCursor())) {
        return null;
    }

    return $this->url($previousCursor);
}
```

So Laravel renders `href=""` on an empty page, reloading the same page instead of throwing. Livewire navigates via `wire:click` rather than URLs, so it needs the encoded cursor; falling back to the current cursor reproduces that same "reload the same page" behaviour.

Applied to both `simple-tailwind.blade.php` and `simple-bootstrap.blade.php`, covering the previous and next links, with regression tests for both empty-page cases across both themes.

Fixes #10297

### DIFF
--- a/src/Features/SupportPagination/UnitTest.php
+++ b/src/Features/SupportPagination/UnitTest.php
@@ -3,10 +3,12 @@
 namespace Livewire\Features\SupportPagination;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Pagination\Cursor;
 use Livewire\Attributes\Computed;
 use Livewire\Component;
 use Livewire\Livewire;
 use Livewire\WithPagination;
+use PHPUnit\Framework\Attributes\DataProvider;
 use Sushi\Sushi;
 use Tests\TestComponent;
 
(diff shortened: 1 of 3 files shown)


ghabriel25:
Hi, @joshhanley There is no guard against this in Flux either. I could submit PR there but this is your idea so I think its better to let you know first.

joshhanley:
@ghabriel25 thanks! I'm aware, but one step at a time. If it get's merged here, then I will sort out Flux 🙂

calebporzio:
thanks!

joshhanley:
I've submitted a Flux PR livewire/flux#2663 with a fix there too.

PR #10303, opened 2026-05-21
[4.x] Fix validation throwing TypeError when rules(), messages() or validationAttributes() return a Collection

### LINKED ISSUE #10301: [v4] HandlesValidation: array_merge() fails when internal AttributeCollection property is present
### Livewire version

v4.3.0

### Laravel version

v13.9.0

### PHP version

v8.4.16

Chrome on macos

Calling `$this->validate()` on a Livewire v4 component throws the following error:

```
TypeError
vendor/livewire/livewire/src/Features/SupportValidation/HandlesValidation.php:138
array_merge(): Argument #1 must be of type array, Illuminate\Support\Collection given
```

## Root cause

Running `get_object_vars($this)` on the component at the time of validation reveals that Livewire v4 injects an internal `attributes` property of type `Livewire\Features\SupportAttributes\AttributeCollection` onto every component instance:

```
"attributes" => Livewire\Features\SupportAttributes\AttributeCollection {#1388}
```

`HandlesValidation::138` calls `array_merge()` over all component properties. When it encounters `AttributeCollection` (which is not an array), PHP throws a `TypeError`.

## Workaround

Bypass `HandlesValidation` by calling Laravel's Validator directly:

```php
public function sendMessage(): void
{
    \Illuminate\Support\Facades\Validator::make(
        ['prompt' => $this->prompt, 'pendingFiles' => $this->pendingFiles],
        $this->rules()
    )->validate();
}
```

This still populates `$errors` in the view correctly.

## Steps to reproduce

1. Create a Livewire component that uses `WithFileUploads`
2. Declare a `public array $pendingFiles = []` property
3. Define a `rules()` method with validation for `pendingFiles`
4. Call `$this->validate()` from any action method

## Minimal reproduction

```php
class Session extends Component
{
    use WithFileUploads;

    public array $pendingFiles = [];
    public string $prompt = '';

    protected function rules(): array
    {
        return [
            'prompt'        => 'required|string|max:10000',
            'pendingFiles'  => 'array|max:5',
            'pendingFiles.*' => 'file|mimes:pdf,jpg|max:10240',
        ];
    }

    public function sendMessage(): void
    {
        $this->validate(); // throws TypeError
    }
}
```

Image

## Expected behavior

`$this->validate()` should skip or safely handle internal Livewire properties (like `attributes`) that are not user-defined and are not arrays.

joshhanley:
# The Scenario

Calling `$this->validate()` throws a `TypeError` and returns a 500 from the Livewire update endpoint.

Screenshot 2026-05-21 at 10 57 03AM@2x

```php
<?php

use Livewire\Component;

new class extends Component {
    public string $prompt = '';

    protected function rules()
    {
        return ['prompt' => 'required'];
    }

    protected function messages()
    {
        // Returns a Collection, not an array.
        return collect(['prompt.required' => 'A prompt is required.'])
            ->mapWithKeys(fn ($message, $key) => [$key => $message]);
    }

    public function save(): void
    {
        $this->validate();
    }
}; ?>

<div>
    <flux:input wire:model="prompt" label="Prompt" />
    <flux:button wire:click="save">Save</flux:button>
</div>
```

# The Problem

This happens whenever a component's `rules()`, `messages()`, or `validationAttributes()` returns a `Collection` instead of a plain array, for example when those values are built with a collection pipeline.

`getRules()`, `getMessages()`, and `getValidationAttributes()` in `HandlesValidation` read the value supplied by the component and pass it straight into `array_merge()`:

```php
if (method_exists($this, 'messages')) $messages = $this->messages();
elseif (property_exists($this, 'messages')) $messages = $this->messages;

// ...

return array_merge($messages, $messagesFromOutside);
```

`array_merge()` requires an `array` for argument 1. If the component returns a `Collection` (or any other `Arrayable`), PHP throws a `TypeError` before validation can run. The same flaw is present in all three methods.

# The Solution

Normalise the component-supplied value to an array before merging. If it implements `Illuminate\Contracts\Support\Arrayable` (which `Collection` does), convert it via `->toArray()`:

```php
if ($messages instanceof Arrayable) $messages = $messages->toArray();
```

This is applied identically in `getRules()`, `getMessages()`, and `getValidationAttributes()`. It mirrors how Laravel's own validation layer normalises `Arrayable` input, for example [`Rule::in()`](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Validation/Rule.php#L120-L127) and [`Rule::notIn()`](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Validation/Rule.php#L135-L141), which both do `if ($values instanceof Arrayable) $values = $values->toArray();`. Array-returning components are completely unaffected.

Fixes #10301

### DIFF
--- a/src/Features/SupportValidation/HandlesValidation.php
+++ b/src/Features/SupportValidation/HandlesValidation.php
@@ -111,6 +111,8 @@ trait HandlesValidation
         if (method_exists($this, 'rules')) $rulesFromComponent = $this->rules();
         else if (property_exists($this, 'rules')) $rulesFromComponent = $this->rules;
 
+        if ($rulesFromComponent instanceof Arrayable) $rulesFromComponent = $rulesFromComponent->toArray();
+
         $rulesFromOutside = array_merge_recursive(
             ...array_map(
                 fn($i) => value($i),
@@ -128,6 +130,8 @@ trait HandlesValidation
         if (method_exists($this, 'messages')) $messages = $this->messages();
         elseif (property_exists($this, 'messages')) $messages = $this->messages;
 
+        if ($messages instanceof Arrayable) $messages = $messages->toArray();
+
         $messagesFromOutside = array_merge(
             ...array_map(
                 fn($i) => value($i),
@@ -144,6 +148,8 @@ trait HandlesValidation
 
         if (method_exists($this, 'validationAttributes')) $validationAttributes = $this->validationAttributes();
         elseif (property_exists($this, 'validationAttributes')) $validationAttributes = $this->validationAttributes;
+
+        if ($validationAttributes instanceof Arrayable) $validationAttributes = $validationAttributes->toArray();
 
         $validationAttributesFromOutside = array_merge(
             ...array_map(
--- a/src/Features/SupportValidation/UnitTest.php
+++ b/src/Features/SupportValidation/UnitTest.php
@@ -941,6 +941,64 @@ class UnitTest extends \Tests\TestCase
             ->post(url('/non-livewire-form'))
             ->assertSee('The bar field is required');
     }
+
+    public function test_validate_handles_rules_returning_a_collection()
+    {
+        Livewire::test(new class extends TestComponent {
+            public string $prompt = '';
+
+            protected function rules()
+            {
+                return collect(['prompt' => 'required']);
+            }
+
+            function save() { $this->validate(); }
+        })
+            ->call('save')
+            ->assertHasErrors(['prompt' => 'required']);
+    }
+
+    public function test_validate_handles_messages_returning_a_collection()
+    {
+        Livewire::test(new class extends TestComponent {
+            public string $prompt = '';
+
+            protected function rules()
+            {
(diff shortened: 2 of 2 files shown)


ghabriel25:
Maybe it could have something like helper method
```php
protected function getValueFromMethodOrProperty($methodOrProperty): array
{
    $value = [];

    if (method_exists($this, $methodOrProperty) {
        $value = $this->{$methodOrProperty}();
    }

    if (property_exists($this, $methodOrProperty) {
        $value = $this->$methodOrProperty;
    }
    
    if ($value instanceof Arrayable) {
        $value = $value->toArray();
    }

    return $value;
}

calebporzio (approved):
Verified locally: the 3 new Collection tests pass on this branch and the full SupportValidation unit file passes (65 tests). Negative control: ran the same 3 tests against main and all 3 fail without the fix. The Arrayable normalization mirrors what getRules() already does for outside rules at line ~513, so this is consistent with existing conventions.

PR #10304, opened 2026-05-21
[3.x] Apply `web` middleware to custom route endpoint if not exists

ghabriel25:
Backport some checking from https://github.com/livewire/livewire/pull/10298 because in [documentation](https://livewire.laravel.com/docs/3.x/installation#configuring-livewires-update-endpoint), there wasn't a mention that `web` middleware should be applied

Also, its actually not appended on subsequent request if its not defined on custom route

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -82,6 +82,14 @@ class HandleRequests extends Mechanism
     {
         $route = $callback([self::class, 'handleUpdate']);
 
+        // Ensure the route includes the `web` middleware group.
+        // Without it, CSRF protection is lost entirely on the update endpoint.
+        // Note: we use middleware() (not gatherMiddleware()) to avoid polluting
+        // the route's computed middleware cache before it's fully configured.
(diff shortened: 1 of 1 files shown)


ghabriel25:
4 failing tests passed on local

Untitled

joshhanley:
@ghabriel25 I’ve just re-run the failed jobs so we can see how that goes.

ghabriel25:
@joshhanley It seems the test need to be adjusted.

I'll look into it later

ghabriel25:
The difference between V3 and V4 browser testing configuration for this failing tests that I'm aware
| | V3 | V4 |
| --- | --- | --- |
| specific driver | ❎  | ☑️ |
| instance cache cleared | ❎  | ☑️ |

I can't find the root cause if testing passed on local but not on github action

joshhanley:
@ghabriel25 seems like an issue with Laravel 11 and 12. Have you run the tests locally with those versions of Laravel installed?

ghabriel25:
I think its because the prefix, v4 has `EndpointResolver` but not v3. I'll fix it soon

ghabriel25:
Im gonna close this for now

PR #10305, opened 2026-05-21
[4.x] Fix nested query parameters with `#[Url]` attribute

hamdyelbatal122:
This PR fixes an issue where nested query parameters (e.g. when binding to Form objects like `#[Url(as: 'form.search')]`) were not being correctly read from the query string on page load or subsequent request referers due to flat array key lookups.

### DIFF
--- a/js/plugins/history/index.js
+++ b/js/plugins/history/index.js
@@ -1,4 +1,4 @@
-import { isObjecty } from "@/utils"
+import { dataDelete, dataGet, dataSet, isObjecty } from "@/utils"
 import historyCoordinator from "./coordinator"
 import { unwrap } from "./utils"
 
@@ -155,7 +155,7 @@ function queryStringUtils() {
 
             let data = fromQueryString(search, key)
 
-            return Object.keys(data).includes(key)
+            return dataGet(data, key) !== undefined
         },
         get(url, key) {
             let search = url.search
@@ -164,12 +164,12 @@ function queryStringUtils() {
 
             let data = fromQueryString(search, key)
 
-            return data[key]
+            return dataGet(data, key)
         },
         set(url, key, value) {
             let data = fromQueryString(url.search, key)
 
-            data[key] = stripNulls(unwrap(value))
+            dataSet(data, key, stripNulls(unwrap(value)))
 
             url.search = toQueryString(data)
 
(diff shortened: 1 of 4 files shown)


joshhanley (approved):
@hamdyelbatal122 thanks for the PR! I added a browser test for this, but it failed. Turns out it also needed a fix on the front end, which I have now added.

# The Scenario

A Livewire component is using a nested URL alias such as `form.search`. The browser URL contains a nested query string value like `?form[search]=hello`, but Livewire does not consistently treat that as the value for `form.search`.

On the initial page load, the property can stay at its default value instead of hydrating from `form[search]`. After the input changes, Livewire can also leave the original nested key in the URL and add a second flat key, producing a URL like `?form[search]=hello&form.search=world`.

After refreshing, the stale nested value can then be read again instead of the latest value.

```php
use Livewire\Attributes\Url;
use Livewire\Component;

new class extends Component
{
    #[Url(as: 'form.search')]
    public string $search = '';

    public function render()
    {
        return <<<'HTML'
        <div>
            <input type="text" wire:model.live="search" />
        </div>
        HTML;
    }
};
```

# The Problem

PHP parses `form[search]=hello` into nested data: `['form' => ['search' => 'hello']]`.

The server-side query string handling was doing flat key lookups in `BaseUrl`, so `form.search` did not resolve to the nested value on initial page loads or when reading the referer URL during Livewire requests.

The browser history code had the same mismatch. It only treated bracketed query keys as nested values when the decoded key started with the tracked key exactly. For a tracked alias of `form.search`, the decoded URL key `form[search]` did not match, so updating the value could create a separate `form.search` query parameter instead of replacing `form[search]`.

# The Solution

The server-side query string lookup now uses nested data access so dot-notated aliases resolve correctly against parsed query data and referer query data.

The browser history query string handling now normalises bracketed keys to dot notation before comparing them with the tracked alias. It also uses the existing `dataGet`, `dataSet`, and `dataDelete` helpers so nested aliases are read, written, and removed consistently.

hamdyelbatal122:
Hey @joshhanley, I've merged the latest `main` branch into this PR branch to resolve any potential conflicts and trigger the latest CI checks (including Laravel 13). It is fully up-to-date and ready for your review/merge. Thanks!

joshhanley:
@hamdyelbatal122 thanks! I've already reviewed it, generally Caleb does the merging 🙂

hamdyelbatal122:
Thanks @joshhanley! Appreciate the extra frontend fix and browser test. Understood — I'll leave it with @calebporzio for merge.

calebporzio:
thanks!

PR #10309, opened 2026-05-23
Fix TemporaryUploadedFile::getSize() after storeAs() (#10238)

### LINKED ISSUE #10238: getSize() on TemporaryUploadedFile throws an error after file has been stored (Livewire 4 only)
### Livewire version

v4.2.4

### Laravel version

v12.56.0

### PHP version

8.4.18

Irrevelant

In Livewire v3 you could do this:

```
$file->storeAs(
  $basePath,
  $fileName
);
```

then access `$file->getSize()` to get the file size.

In Livewire v4, `storeAs()` turns the temporary file into a json file, which makes the subsequent `$file->getSize()` fail:

```
Unable to retrieve the file_size for file at location: livewire-tmp/ZCBYJ28dEA7oUVbBA3osihME3CJtla80c11XPbHD.pdf.
```

The fix is easy (getting the size before storing), but this is not documented in the upgrade guide.

See above

_No response_

Either work as it did in v3, or be documented.

YoussefMansour9:
Fixes #10238

In Livewire v4, `storeAs()` uses a same-disk move optimization. After the move, `getSize()` called `$this->storage->size($this->path)` which fails because the file no longer exists at the original path.

This fix checks the metadata file for the size (which is always stored at upload time) before falling back to the storage size call. This follows the same pattern as `getClientOriginalName()` which already relies on metadata.

### DIFF
--- a/src/Features/SupportFileUploads/TemporaryUploadedFile.php
+++ b/src/Features/SupportFileUploads/TemporaryUploadedFile.php
@@ -46,11 +46,11 @@ class TemporaryUploadedFile extends UploadedFile
 
     public function getSize(): int
     {
-        if (app()->runningUnitTests()) {
-            if (isset($this->metaFileData()['size'])) {
-                return $this->metaFileData()['size'];
-            }
+        if (isset($this->metaFileData()['size'])) {
+            return $this->metaFileData()['size'];
+        }
 
+        if (app()->runningUnitTests()) {
             // This is for backwards compatibility when test file meta data was stored in the filename...
             if (str($this->getFilename())->contains('-size=')) {
                 return (int) str($this->getFilename())->between('-size=', '.')->__toString();


ghabriel25:
I think theres already explanation here https://github.com/livewire/livewire/issues/10238#issuecomment-4339978913 to keep temp object for single use.

So as workaround, its best to assign the size to a variable before store it.

Is there a specific case for you?

YoussefMansour9:
Thanks for the review @ghabriel25! The specific case is when the upload goes through a queue job that calls `storeAs()` and later needs to check the file size (e.g. for processing, thumbnails, or reporting). The workaround of storing size before `storeAs()` works but is easy to miss since `getClientOriginalName()` and other metadata methods still work after storing — the inconsistency can be a subtle bug.

That said, I understand keeping the temp object single-use and it looks like @joshhanley already considered and decided against this approach in the original discussion. Happy to close the PR if that is the maintainers preference. Let me know! :+1:

calebporzio:
I'll let @joshhanley make this call because he has more context 👌

joshhanley:
@YoussefMansour9 thanks for the PR! Yeah I'm not comfortable with this change just yet. I'd appreciate some input and thoughts on what you guys think about my comments https://github.com/livewire/livewire/issues/10238#issuecomment-4339978913 on the original issue as I don't like any of the options particularly haha.

I've marked this PR as a draft for now, while we discuss.

I think my main concern is if we add the changes in this PR is that they won't work for files on S3. The reason is because S3 doesn't have the metadata file.

The metadata file was introduced to fix a file pathname limit for Windows devices, but S3 doesn't have that problem and as such doesn't need the metadata file. We could in theory add the metadata file to S3 anyway, but I opted not to when I implemented that, because it would required an extra signed URL for the metafile to be uploaded to S3 for every file that is to be uploaded (plus other issues if the main file uploads and the meta doesn't, etc). Seemed like an unnecessary overhead when S3 didn't need it as it doesn't have the file pathname issue.

So if we were to use the metadata file anyway if it exists (aka not on S3), my concern would be that developers would have working code locally, that may break when they switch to using S3 in production, which is actually worse than it just not working at all (the current bug).

I'm open to all thoughts and ideas though. I've thought over this one a fair bit and as I mentioned, don't particularly like any of the answer I have come up with so far.

calebporzio:
Mmmm yep tricky. Agree with your thinking on this @joshhanley - I'd rather this be broken for both local and prod than just prod.

Of course we need some kind of answer here though.

Documentation at least would be a good stop-gap, but I wonder what a more durable solution could be.

Is there not way to add a conditional to detect s3 and route around the metadata file? Seems like an opportunity for an abstraction. Something with a unified interface that knows when to access metadata file and when to not you know?

This feels solvable from my perspective but I'm not in the weeds. Thanks everyone.

joshhanley:
@calebporzio so I've done some more digging and exploration on this one and the conclusion that I have come to is that I actually think the `move` PR should be reverted. My reasoning is that Laravel's `storeAs()` method is copy, not move, so we should keep our `storeAs()` inline with that. That then prevents the problem, that this PR addresses, from happening altogether as the original file already exists.

I've submitted a PR #10390 to revert the "Move temporary file if disks are same" PR.

I considered adding a method like `moveAs()` as an alternative, but Laravel doesn't have anything like that. There is a `move()` method that Livewire and Laravel both inherit from Symfony, but it doesn't work without temporary files. So just opted for the revert for now, to solve this current problem.

If you do decide to merge that revert PR, then this PR can be closed.

joshhanley:
Closing this as the revert PR #10390 has been merged. Thanks @YoussefMansour9!

PR #10311, opened 2026-05-27
[3.x] Validate page query parameter in SupportPagination

pkoury:

Yes. Production 500 on PHP 8.5 triggered by a Bingbot crawl of a `WithPagination` page with an overflow-large `?page=` value. No prior discussion. The regression is described in section 5️⃣ below.

2️⃣ Did you create a branch for your fix/feature?

Yes. `fix/pagination-page-overflow-3x` off `3.x`.

3️⃣ Does it contain multiple, unrelated changes?

No. One method changed, one regression test added.

4️⃣ Does it include tests?

Yes. New unit test in `src/Features/SupportPagination/UnitTest.php` covering overflow-large strings, non-numeric strings, zero, negative numbers, and the happy path. Tests fail on the unfixed `3.x` checkout and pass with the fix applied.

## Problem

On PHP 8.4+, `SupportPagination`'s registered `currentPageResolver` closure does an unguarded `(int)` cast on a value sourced from the untrusted `?page=` query string. When the value overflows `PHP_INT_MAX`, the cast throws `ErrorException` and the request 500s. Search engine crawlers reliably reproduce this in production.

```
ErrorException: The float 1.2123123123213123E+22 is not representable as an int, cast occurred
at src/Features/SupportPagination/SupportPagination.php:83
```

Non-numeric strings, zero, and negative numbers are also accepted without validation, which doesn't crash but produces incorrect page numbers downstream.

## Fix

Validate inside the resolver closure with `FILTER_VALIDATE_INT`, falling back to page 1 for invalid, out-of-range, zero, or negative values. Mirrors Laravel's stock `currentPageResolver` in `PaginationState::resolveUsing()` line for line.

```php
Paginator::currentPageResolver(function ($pageName) {
    $this->ensurePaginatorIsInitialized($pageName);

    $page = $this->component->paginators[$pageName];

    if (filter_var($page, FILTER_VALIDATE_INT) !== false && (int) $page >= 1) {
        return (int) $page;
    }

    return 1;
});
```

The validation has to be at the resolver closure (the read point) rather than inside `resolvePage()`. `addUrlHook()` registers a `PaginationUrl` attribute that calls `setPropertyFromQueryString()`, which overwrites `paginators[$pageName]` with the raw query string after `resolvePage()` has returned. Validating at the closure covers both the initial-resolution path and the URL-hook path.

Cursor pagination is unaffected. The `CursorPaginator::currentCursorResolver` closure passes the value to `Cursor::fromEncoded()` and never casts to int.

## Companion

A matching PR against `main` will be opened for the same bug in the 4.x line.

### DIFF
--- a/src/Features/SupportPagination/SupportPagination.php
+++ b/src/Features/SupportPagination/SupportPagination.php
@@ -83,8 +83,12 @@ class SupportPagination extends ComponentHook
             $page = $this->component->paginators[$pageName];
 
             if (filter_var($page, FILTER_VALIDATE_INT) !== false && (int) $page >= 1) {
-                return (int) $page;
+                $this->component->paginators[$pageName] = (int) $page;
+
+                return $this->component->paginators[$pageName];
             }
+
+            $this->component->paginators[$pageName] = 1;
 
             return 1;
         });
--- a/src/Features/SupportPagination/SupportPagination.php
+++ b/src/Features/SupportPagination/SupportPagination.php
@@ -80,7 +80,13 @@ class SupportPagination extends ComponentHook
         Paginator::currentPageResolver(function ($pageName) {
             $this->ensurePaginatorIsInitialized($pageName);
 
-            return (int) $this->component->paginators[$pageName];
+            $page = $this->component->paginators[$pageName];
+
+            if (filter_var($page, FILTER_VALIDATE_INT) !== false && (int) $page >= 1) {
+                return (int) $page;
+            }
+
+            return 1;
         });
     }
 
--- a/src/Features/SupportPagination/SupportPagination.php
+++ b/src/Features/SupportPagination/SupportPagination.php
@@ -83,14 +83,10 @@ class SupportPagination extends ComponentHook
             $page = $this->component->paginators[$pageName];
 
             if (filter_var($page, FILTER_VALIDATE_INT) !== false && (int) $page >= 1) {
-                $this->component->paginators[$pageName] = (int) $page;
-
-                return $this->component->paginators[$pageName];
+                return $this->component->paginators[$pageName] = (int) $page;
             }
 
-            $this->component->paginators[$pageName] = 1;
-
-            return 1;
+            return $this->component->paginators[$pageName] = 1;
         });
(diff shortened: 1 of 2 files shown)


src/Features/SupportPagination/SupportPagination.php
```diff
@@ -80,7 +80,13 @@ protected function setPageResolvers()
         Paginator::currentPageResolver(function ($pageName) {
             $this->ensurePaginatorIsInitialized($pageName);
 
-            return (int) $this->component->paginators[$pageName];
+            $page = $this->component->paginators[$pageName];
+
+            if (filter_var($page, FILTER_VALIDATE_INT) !== false && (int) $page >= 1) {
```
ghabriel25, line 85: You could simplify this without casting two times

```php
if (($page = filter_var($page, FILTER_VALIDATE_INT)) !== false && $page >= 1) {
    return $page;
}
```
Since return value from `filter_var` on success convert it to int
> Validates whether the value is an integer, on success it is converted to type [int](https://www.php.net/manual/en/language.types.integer.php).

reference: https://www.php.net/manual/en/filter.constants.php#constant.filter-validate-int

pkoury: Thanks! I did consider a simplification but ended up going with a mirror of Laravel core's [`PaginationState::resolveUsing()`](https://github.com/laravel/framework/blob/13.x/src/Illuminate/Pagination/PaginationState.php#L22-L24) instead.

Happy to switch to the simpler form if Caleb or Josh prefer it.

ghabriel25: I see. Following Laravel pattern should be fine 👍

joshhanley (approved):
@pkoury thanks for the PR! I made one small tweak to reassign back to the paginators array once the cast/ reset is done to ensure it's kept up to date.

pkoury:
Thanks @joshhanley! That makes sense. Appreciate you tightening that up a bit.

calebporzio:
great ,thanks everyone!

PR #10312, opened 2026-05-27
[4.x] Validate page query parameter in SupportPagination

pkoury:
Companion fix for #10311 (3.x).

Yes. `SupportPagination`'s `currentPageResolver` silently truncates overflow-large `?page=` query values to `PHP_INT_MAX` and accepts non-numeric, zero, and negative values without validation. The same unvalidated `(int)` cast on 3.x throws `ErrorException` on PHP 8.4+ and returned a 500 in production (Bingbot crawl, companion PR #10311). No prior discussion. Regression described in section 5️⃣ below.

2️⃣ Did you create a branch for your fix/feature?

Yes. `fix/pagination-page-overflow` off `main`.

3️⃣ Does it contain multiple, unrelated changes?

No. One method changed, one regression test added.

4️⃣ Does it include tests?

Yes. New unit test in `src/Features/SupportPagination/UnitTest.php` covering overflow-large strings, non-numeric strings, zero, negative numbers, and the happy path. Tests fail on the unfixed `main` checkout and pass with the fix applied.

## Problem

`SupportPagination`'s registered `currentPageResolver` closure does an unguarded `(int)` cast on a value sourced from the untrusted `?page=` query string.

On 3.x, this throws `ErrorException` for overflow-large values on PHP 8.4+ and returns a 500 (production trace in the companion PR).

On `main`, the cast does not throw. Overflow values silently truncate to `PHP_INT_MAX` (9223372036854775807) and resolve as the current page, producing empty result sets without an error log entry. Non-numeric strings, zero, and negative numbers are also accepted without validation and produce incorrect page numbers downstream.

## Fix

Validate inside the resolver closure with `FILTER_VALIDATE_INT`, falling back to page 1 for invalid, out-of-range, zero, or negative values. Mirrors Laravel's stock `currentPageResolver` in `PaginationState::resolveUsing()` line for line.

```php
Paginator::currentPageResolver(function ($pageName) {
    $this->ensurePaginatorIsInitialized($pageName);

    $page = $this->component->paginators[$pageName];

    if (filter_var($page, FILTER_VALIDATE_INT) !== false && (int) $page >= 1) {
        return (int) $page;
    }

    return 1;
});
```

The validation has to be at the resolver closure (the read point) rather than inside `resolvePage()`. `addUrlHook()` registers a `PaginationUrl` attribute that calls `setPropertyFromQueryString()`, which overwrites `paginators[$pageName]` with the raw query string after `resolvePage()` has returned. Validating at the closure covers both the initial-resolution path and the URL-hook path.

Cursor pagination is unaffected. The `CursorPaginator::currentCursorResolver` closure passes the value to `Cursor::fromEncoded()` and never casts to int.

### DIFF
--- a/src/Features/SupportPagination/SupportPagination.php
+++ b/src/Features/SupportPagination/SupportPagination.php
@@ -80,7 +80,13 @@ class SupportPagination extends ComponentHook
         Paginator::currentPageResolver(function ($pageName) {
             $this->ensurePaginatorIsInitialized($pageName);
 
-            return (int) $this->component->paginators[$pageName];
+            $page = $this->component->paginators[$pageName];
+
+            if (filter_var($page, FILTER_VALIDATE_INT) !== false && (int) $page >= 1) {
+                return (int) $page;
+            }
+
+            return 1;
         });
     }
 
--- a/src/Features/SupportPagination/UnitTest.php
+++ b/src/Features/SupportPagination/UnitTest.php
@@ -108,6 +108,29 @@ class UnitTest extends \Tests\TestCase
         })->assertSee('Custom simple pagination theme');
     }
 
+    public function test_invalid_page_query_string_falls_back_to_first_page()
+    {
+        $cases = [
+            '12123123123213123123213', // overflows PHP_INT_MAX, would throw on PHP 8.4 cast
+            'not-a-number',
+            '0',
+            '-3',
+        ];
+
+        foreach ($cases as $value) {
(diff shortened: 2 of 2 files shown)


joshhanley:
@pkoury thanks for the PR! I'm going to close this one, in favour of your V3 PR #10311, as they are identical and all V3 changes get upmerged into V4 anyway. So to save yourself the effort in future, just submitting to V3 will get the fix in both 🙂

PR #10307, opened 2026-05-22
Add existing session errors to support validation

JeremyRed:
This PR resolves #10306. In summary, when progressively migrating to Livewire 4, named error bags generated during traditional HTTP redirects or internal component validation fail to render on Livewire page components. 

While the default error bag works as intended, data contained within custom-named error bags appears to be lost during Livewire's hydration and rendering cycles.

This creates a barrier for incremental adoption, as developers cannot cleanly mix Livewire components with legacy forms that rely heavily on named error bags.

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?

Yes, there have been several discussions and pr attempts to resolve this issue.
- PR attempt: https://github.com/livewire/livewire/pull/4795
- Raised issue: https://github.com/livewire/livewire/pull/7265#issuecomment-1806547989
- PR attempt: https://github.com/livewire/livewire/pull/6715
- Discussion: https://github.com/livewire/livewire/discussions/2585#discussioncomment-1136714

2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed)

Yes.

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.

No.

4️⃣ Does it include tests? (Required)

Yes.

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

Details in #10306

PR #10314, opened 2026-05-28
[4.x] Preserve named error bags during component rendering

### LINKED ISSUE #10306: Livewire rendering cycle does not forward existing named `$errors` from session
### Livewire version

v4.3.0

### Laravel version

v13.11.2

### PHP version

8.5.6

Chrome on MacOS

G'day team. We've recently been progressively migrating an existing project to Livewire 4. Because we are doing this page-by-page, we still have legacy blade forms on these pages that post to traditional Laravel controller routes, heavily relying on named error bags.

The issue we've run into is that named error bags seem to get lost during the Livewire request lifecycle upon redirection.

**Reproduction Summary:**

- A user is on a newly migrated Livewire page.
- They submit an existing legacy form on that page, which posts to a standard Laravel controller route.
- The controller validation fails, and it redirects the user back to the Livewire page using something like `.withErrors($validator, 'bag_name')`.
- Upon landing back on the Livewire page, the named error bag no longer exists in the `$errors` view bag while the default bag works as expected.

Below is a minimal setup to reproduce the issue. It features a standard POST route throwing a validation exception inside a named bag, alongside a Livewire component handling both Livewire validation and the legacy form landing.

**Routes**

```php
<?php

use Illuminate\Validation\ValidationException;

Route::livewire('/', 'pages::test');

Route::post('/error/named', function () {
    throw ValidationException::withMessages([
        'title' => 'ERROR NAMED',
    ])->errorBag('foo');
});

```

**Page Component (pages::test)**

```php
<?php

use Illuminate\Support\Facades\Validator;
use Livewire\Component;

new class extends Component {
    public string $title = '';

    public function save()
    {
        Validator::make(
            ['title' => $this->title],
            ['title' => 'required'],
        )->validate();
    }
};
?>

<div>
    {{-- Livewire form --}}
    <form wire:submit="save">
        <label>
            Livewire Title
            <input type="text" wire:model="title">
            @error('title') <span style="color: red;">{{ $message }}</span> @enderror
        </label>
        <button type="submit">Validate Livewire</button>
    </form>

    {{-- Legacy non-livewire form --}}
    <form action="/error/named" method="post">
        @csrf
        <label>
            Legacy Title
            <input type="text" name="title">
            @error('title', 'foo') <span style="color: red;">{{ $message }}</span> @enderror
        </label>
        <button type="submit">Validate Laravel</button>
    </form>
</div>

```

_No response_

**Expected Behaviour:**

* Clicking **"Validate Laravel"** submits to the traditional POST route, triggers the exception on the foo bag, redirects back to the Livewire page, and renders the error message via @error('title', 'foo').
  
**Actual Behaviour:**

* Any error message relying on the named error bag (foo) fails to appear. The named bag data appears to be dropped during Livewire's hydration/rendering cycle.

hamdyelbatal122:
### Description

This PR resolves an issue where named error bags passed from session redirects (via `.withErrors($validator, 'bag_name')`) were lost during Livewire rendering.

Fixes #10306

### Root Cause

In `SupportValidation.php`, Livewire's component validation hooks (`render` and `renderIsland`) overwrite the globally shared `errors` view variable with a new `ViewErrorBag` containing only the component's default error bag:

```php
$errors = (new ViewErrorBag)->put('default', $this->component->getErrorBag());
```

This completely discards any previously shared `errors` (including named error bags) during the rendering of the component, making them inaccessible in Blade via `@error('field', 'bag_name')` or `$errors->getBag('bag_name')`.

### Solution

Instead of overwriting the shared variable with a new empty bag, we retrieve the previously shared `errors` variable first. If it is a `ViewErrorBag`, we copy all existing bags over to the new `ViewErrorBag` before putting the component's default error bag.

I've also added a unit test to verify this behavior.

### DIFF
--- a/src/Features/SupportValidation/SupportValidation.php
+++ b/src/Features/SupportValidation/SupportValidation.php
@@ -14,6 +14,8 @@ class SupportValidation extends ComponentHook
 
     public static function provide()
     {
+        static::$view = null;
+
         on('flush-state', function () {
             static::$view = null;
         });
--- a/src/Features/SupportValidation/SupportValidation.php
+++ b/src/Features/SupportValidation/SupportValidation.php
@@ -18,17 +18,7 @@ class SupportValidation extends ComponentHook
 
     function render($view, $data)
     {
-        $errors = new ViewErrorBag;
-
-        $previouslySharedErrors = app('view')->getShared()['errors'] ?? null;
-
-        if ($previouslySharedErrors instanceof ViewErrorBag) {
-            foreach ($previouslySharedErrors->getBags() as $key => $bag) {
-                $errors->put($key, $bag);
-            }
-        }
-
-        $errors->put('default', $this->component->getErrorBag());
+        $errors = $this->viewErrorBag();
 
         $revert = Utils::shareWithViews('errors', $errors);
 
@@ -41,17 +31,7 @@ class SupportValidation extends ComponentHook
 
     function renderIsland($name, $view, $data)
     {
-        $errors = new ViewErrorBag;
-
-        $previouslySharedErrors = app('view')->getShared()['errors'] ?? null;
-
-        if ($previouslySharedErrors instanceof ViewErrorBag) {
-            foreach ($previouslySharedErrors->getBags() as $key => $bag) {
-                $errors->put($key, $bag);
-            }
-        }
-
-        $errors->put('default', $this->component->getErrorBag());
+        $errors = $this->viewErrorBag();
 
         $revert = Utils::shareWithViews('errors', $errors);
 
@@ -81,5 +61,20 @@ class SupportValidation extends ComponentHook
         $this->component->setErrorBag($e->validator->errors());
 
         $stopPropagation();
+    }
+
+    protected function viewErrorBag(): ViewErrorBag
+    {
+        $errors = new ViewErrorBag;
+
+        $previouslySharedErrors = app('view')->getShared()['errors'] ?? null;
+
+        if ($previouslySharedErrors instanceof ViewErrorBag) {
+            foreach ($previouslySharedErrors->getBags() as $name => $bag) {
+                $errors->put($name, $bag);
+            }
+        }
+
+        return $errors->put('default', $this->component->getErrorBag());
     }
 }
--- a/src/Features/SupportValidation/SupportValidation.php
+++ b/src/Features/SupportValidation/SupportValidation.php
@@ -18,7 +18,17 @@ class SupportValidation extends ComponentHook
 
     function render($view, $data)
     {
-        $errors = (new ViewErrorBag)->put('default', $this->component->getErrorBag());
+        $errors = new ViewErrorBag;
+
+        $previouslySharedErrors = app('view')->getShared()['errors'] ?? null;
+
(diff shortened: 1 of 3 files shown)


ghabriel25:
@hamdyelbatal122 duplicate PR ? https://github.com/livewire/livewire/pull/10307. Although, the approach taken seems different

joshhanley (approved):
@hamdyelbatal122 thanks for the PR! Yeah this is a duplicate of PR #10307, but I think this implementation is better as it doesn't depend on the session. I've copied the test across from that PR as it's more inline with what was reported in the issue.

hamdyelbatal122:
Hey @joshhanley, I noticed that the Unit tests for Laravel 13 were failing due to the default validation message format (which doesn't include the trailing period in newer Laravel versions). I've updated the test assertion to check for `'The bar field is required'` without the trailing dot to make it compatible across all versions.

The branch has been updated and is ready for the CI runs. Thanks again for the review!

joshhanley:
@hamdyelbatal122 nah they were failing due to an upstream issue in Symfony and Laravel. It got fixed in last weeks release of Laravel.

hamdyelbatal122:
Thanks for the clarification @joshhanley, and for the review. Glad this approach works.

src/Features/SupportValidation/SupportValidation.php
```diff
@@ -62,4 +62,19 @@ function exception($e, $stopPropagation)
...
+
+        $previouslySharedErrors = app('view')->getShared()['errors'] ?? null;
+
+        if ($previouslySharedErrors instanceof ViewErrorBag) {
+            foreach ($previouslySharedErrors->getBags() as $name => $bag) {
+                $errors->put($name, $bag);
+            }
+        }
+
+        return $errors->put('default', $this->component->getErrorBag());
```
calebporzio, line 88: Two thoughts:

- Why not just use the shared error bag in the first place? aren't we effectively creating like a clone of it?
- Accessing `app('view')->getShared()` has a runtime cost. Especially on a page with hundreds of livewire components, that's hundreds of calls to that. (could be memoized - but would have to be cleared for octane-like environments)

hamdyelbatal122: Hi @calebporzio ,

Regarding your thoughts:

1. **Using the shared error bag directly vs cloning:**
Since `ViewErrorBag` is an object, mutating it directly (e.g. by putting the component's own default bag via `put('default', ...)`) would mutate the globally shared instance. This would cause child or sibling components to overwrite the default error bag for their parent and sibling components. By cloning the shared `ViewErrorBag` (which creates a shallow copy of the bags array), we can safely override the `'default'` bag for the current component's render scope without side-effects on other components, and then safely revert back to the previous instance.

2. **Runtime cost of `app('view')->getShared()` & Octane compatibility:**
That's a great point. I have updated the code to memoize the view factory in a static property (`static::$view`) to avoid calling `app('view')` repeatedly, and hooked into Livewire's `flush-state` event to reset it, making it completely safe for persistent server environments like Octane.

I have also simplified the code to use `clone` directly instead of the manual `foreach` loop. I've pushed these changes to the branch!

calebporzio:
thanks!

PR #10315, opened 2026-05-28
[4.x] Replace some of form object method with existing `InteractsWithProperties` trait

ghabriel25:
I notice that some methods inside `Form` class is exactly the same with methods in `InteractsWithProperties` trait. So, instead duplicating the same code, why not just use this trait? afterall, class method override trait method (See [Example #3](https://www.php.net/manual/en/language.oop5.traits.php))

My first thought was maybe its duplicated to avoid conflict with `Component` traits which also use `InteractsWithProperties` however all tests passed so it's all good.

| Similar methods | Removed |
| --- | --- |
| `hasProperty()` | ✅ |
| `getPropertyValue()` | ✅ |
| `only()` | ✅ |
| `except()` | ✅ |
| `fill()` | ✅ |
| `pull()` | ✅ |
| `resetExcept()` | ✅ |
| `reset()` | ❌ |

`reset()` method has different flow so I dont want to touch that part because it will become recursive.

The only thing that should be aware is something like this

```php
class PostForm extends From
{
    use InteractsWithProperties; // ❌

    public $title;

    public $body;
}
```
This will override it all.

### File Changed
- `src/Features/SupportFormObjects/Form.php`

### DIFF
--- a/src/Features/SupportFormObjects/Form.php
+++ b/src/Features/SupportFormObjects/Form.php
@@ -15,14 +15,12 @@ use Livewire\Concerns\InteractsWithProperties;
 
 class Form implements Arrayable
 {
-    use InteractsWithProperties {
-        resetExcept as public;
-    }
-
     use HandlesValidation {
         validate as parentValidate;
         validateOnly as parentValidateOnly;
     }
+
+    use InteractsWithProperties;
 
     function __construct(
         protected Component $component,
--- a/src/Features/SupportFormObjects/Form.php
+++ b/src/Features/SupportFormObjects/Form.php
@@ -11,9 +11,12 @@ use function Livewire\invade;
 use Illuminate\Support\Arr;
 use Livewire\Drawer\Utils;
 use Livewire\Component;
+use Livewire\Concerns\InteractsWithProperties;
 
 class Form implements Arrayable
 {
+    use InteractsWithProperties;
+
     use HandlesValidation {
         validate as parentValidate;
         validateOnly as parentValidateOnly;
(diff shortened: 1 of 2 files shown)


src/Features/SupportFormObjects/Form.php
```diff
@@ -11,9 +11,14 @@
 use Illuminate\Support\Arr;
 use Livewire\Drawer\Utils;
 use Livewire\Component;
+use Livewire\Concerns\InteractsWithProperties;
 
 class Form implements Arrayable
 {
+    use InteractsWithProperties {
+        resetExcept as public;
```
joshhanley: Let's just make `resetExcept` public in `InteractsWithProperties`, instead of changing the visibility here.

joshhanley (changes requested):
@ghabriel25 thanks for the PR! Personally I don't like early abstraction, I tend to follow the rule of 3. If something is copied 3 or more times, then I will work out whether I need an abstraction or not. Often times it's still not needed.

But I think I'm ok with this refactor.

One small change, I think we should just make resetExcept public in `InteractsWithProperties`, then we don't need the visibility override on the trait import.

I've look at all the other methods that you have removed in comparison to the `InteractsWithProperties` implementation and can't see any other differences or concerns. So I'd say it's good to go other than the small change above.

joshhanley (approved):
I've made the changes to make `resetExcept` public.

I did a review to make sure there wasn't any issues with making `resetExcept` public on the component and there are protections in place so it can't be called from the front end. But even if we didn't have that, the implementation is just an array flip, that then gets passed to `reset()`, which is public and has it's own checks.

src/Features/SupportFormObjects/Form.php
ghabriel25: I'll work on it. Thanks!

Just curious why its protected though? Thats why I override it because I dont want any breaking changes

ghabriel25:
Thanks!

ghabriel25:
@joshhanley failing test

```php
function test_wire_loading_targets_exclude_wire_navigate()
{
    Livewire::visit(new class extends Component {
        public function hydrate()
        {
            sleep(1);
        }

        public function render()
        {
            return <<<'HTML'
                <div>
                    <a href="/otherpage" wire:navigate dusk="link" class="text-blue-500" wire:loading.class="text-red-500">Link</a>
                    <button type="button" wire:click="$refresh" dusk="refresh-button">Refresh</button>
                </div>
            HTML;
        }
    })
    ->assertHasClass('@link', 'text-blue-500')
    ->click('@refresh-button')
    ->pause(5)
    ->assertHasClass('@link', 'text-red-500')
    ;
}
```

joshhanley:
@ghabriel25 that failing test is unrelated. We do have flaky tests, so it will happen occassionally.

You also generally don't need to keep merging main in. I would only do it if there is a change on main that will conflict with your changes, otherwise it just creates more noise on the PR, that way it's ready for a clean merge 🙂

ghabriel25:
@joshhanley I see. 

I thought it cant be merge if not updated, thanks for the information.

joshhanley:
@ghabriel25 nah, it's only blocked if there is a conflict with main. Otherwise it would make our lives harder as maintainers 😅

calebporzio:
Thanks!

PR #10316, opened 2026-05-30
[4.x] Use existing component authorization instead Gate facade

ghabriel25:
Maybe this is just personal preference but Livewire component already use `AuthorizesRequests` trait from Laravel so calling `Gate` facade inside **BaseAuthorize** attribute seems redundant.

With this, 
- authorization goes through Component layer rather than bypassing it
- remove slight performance overhead (negligible) from facade.

**Before**
```php
public function call(array $params)
{
    // Action that does not require a model or class...
    if (is_null($this->argument)) {
       Gate::authorize($this->ability);

        return;
    }
    ...
    Gate::authorize($this->ability, $resolved);
}
```
**After**
```php
public function call(array $params)
{
    // Action that does not require a model or class...
    if (is_null($this->argument)) {
       $this->component->authorize($this->ability);

        return;
    }
    ...
    $this->component->authorize($this->ability, $resolved);
}
```

### DIFF
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -4,7 +4,6 @@ namespace Livewire\Features\SupportAuthorization;
 
 use Attribute;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Facades\Gate;
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
 use UnitEnum;
 
@@ -20,7 +19,7 @@ class BaseAuthorize extends LivewireAttribute
     {
         // Action that does not require a model or class...
         if (is_null($this->argument)) {
-            Gate::authorize($this->ability);
+            $this->component->authorize($this->ability);
 
             return;
         }
@@ -33,7 +32,7 @@ class BaseAuthorize extends LivewireAttribute
             $resolved[] = $this->resolveArgument($arg, $parameters);
         }
 
-        Gate::authorize($this->ability, $resolved);
+        $this->component->authorize($this->ability, $resolved);
     }
 
     /**


ghabriel25:
Failing tests from workflow but passed on local
- **Livewire\Features\SupportTesting\UnitTest::test_testing_does_not_create_dynamic_properties_on_test_response**
ErrorException: Since symfony/http-foundation 8.1: Directly setting property "headers" of "Illuminate\Http\Response" is deprecated; pass the header bag as a constructor argument instead. 
Laravel already has fix for this (See https://github.com/laravel/framework/pull/60318)

- **Livewire\Mechanisms\PersistentMiddleware\BrowserTest::test_that_authorization_middleware_is_re_applied**
Did not see expected text [Protected Content] within element [body].
Failed asserting that false is true.

joshhanley:
@ghabriel25 thanks for the PR! Having a look at this, I'm a little concerned because it's not a straight swap. Looking at the `authorize()` method on the trait, it actually runs `parseAbilityAndArguments()` before it calls authorise on the `Gate` class. Where as our implementation just calls `Gate::authorize()` directly without any parsing.

On one hand, it could be seen as an extra benefit that this new implementation has the parsing. But Im also concerned about it being an unintentional breaking change. As such I'm going to close this. But if you were running into issues that this fixes, please let us know and we can look at reopening it.

ghabriel25:
Hi. @joshhanley actually there is a specific case for this that I forgot to mention

- user dont sepecify the ability
- action name is the same as policy method or defined gate name

```php
class PostPolicy
{
    public function create(User $user)
    {
        return true;
    }
}
```
```php
new class extends Component
{
    #[Authorize(Post::class)] // ❌
    public function create()
    {
        $this->authorize(Post::class); // ✅
    }
}
```
However, this needs another test case.

And I think, replacing `Gate` with `$this->component` doesnt fix this issue because resolved method inside this attribute would be `call()` not `create()`. I'll post the issue if needed

PR #10317, opened 2026-06-01
[4.x] Fix autofocus after `wire:navigate`

### LINKED ISSUE #10299: Native autofocus works on full page load but not after wire:navigate
### Livewire version

v4.3.0

### Laravel version

v13.9.0

### PHP version

8.5.4

Chrome on MacOS

The HTML autofocus attribute works on full page loads, but not when navigating via wire:navigate. After SPA-style navigation, elements marked with autofocus are not focused.

```
php artisan make:livewire autofocus-home --sfc --no-interaction
php artisan make:livewire autofocus-target --sfc --no-interaction
```

```php
// routes/web.php
Route::livewire('/autofocus-test', 'pages::autofocus-home');
Route::livewire('/autofocus-target', 'pages::autofocus-target');
```

```blade
{{-- resources/views/pages/autofocus-home.blade.php --}}
<div>
  <p>Start here, then navigate via wire:navigate:</p>
  <a href="/autofocus-target" wire:navigate>
      Go to target page
  </a>
</div>
```

```blade
{{-- resources/views/pages/autofocus-target.blade.php --}}
<div>
  <p>This textarea has the native <code>autofocus</code> attribute:</p>
  <textarea autofocus rows="4" cols="40" placeholder="Should be focused"></textarea>
</div>
```

Small demo of the given code snippets.
_*When the text resizing occurs, those are refreshes that focus the text area_

https://github.com/user-attachments/assets/a9e2316e-40f3-4f11-ba66-d329d9bf7e8e

**Expected**
After wire:navigate, the textarea receives focus (same as a full page load).

**Actual**
After wire:navigate, the textarea is rendered but not focused. A direct visit or hard refresh works as expected.

**Suggestion**
After swapping page content on `wire:navigate`, Livewire could focus the first visible `autofocus` element, similar to how browsers do on initial load.

joshhanley:
# The Scenario

If you have two pages, each with an input that has `autofocus` on it, the browser focuses the input when you load the first page.

But if you then `wire:navigate` to the second page, the second page's `autofocus` input isn't focused.

CleanShot 2026-06-01 at 17 16 55

```php
use Livewire\Component;

class FirstAutofocusPage extends Component
{
    {
        return <<<'HTML'
        <div>
            <input id="first-input" type="text">
            <input id="first-autofocus" autofocus type="text">

            <a href="/second-autofocus" wire:navigate>Go to second autofocus page</a>
        </div>
        HTML;
    }
}

class SecondAutofocusPage extends Component
{
    {
        return <<<'HTML'
        <div>
            <textarea id="second-autofocus" autofocus></textarea>
        </div>
        HTML;
    }
}
```

# The Problem

The browser only applies native `autofocus` automatically during document load. After `wire:navigate` swaps in a new page, the existing document has already processed autofocus, so the newly added element will not be focused automatically.

The navigate plugin already had an `autofocusElementsWithTheAutofocusAttribute()` helper for this case, but it was permanently disabled behind `let autofocus = false`.

# The Solution

This removes the dead `autofocus` flag and runs the existing autofocus helper after the new page has been swapped in and Alpine has initialised.

I did dig back through Livewire's navigate git history and Alpine's navigate git history to see if there was a reason autofocus was disabled and I couldn't find anything that intentionally added it. So I believe it's safe to change it and use it, as I don't think it was disabled to stop a bug.

CleanShot 2026-06-01 at 17 10 36

Fixes #10299

### DIFF
--- a/js/plugins/navigate/index.js
+++ b/js/plugins/navigate/index.js
@@ -12,7 +12,6 @@ import { fetchHtml } from "./fetch"
 let enablePersist = true
 let showProgressBar = true
 let restoreScroll = true
-let autofocus = false
 
 export default function (Alpine) {
 
@@ -123,11 +122,8 @@ export default function (Alpine) {
 
                     afterNewScriptsAreDoneLoading(() => {
                         andAfterAllThis(() => {
-                            setTimeout(() => {
-                                autofocus && autofocusElementsWithTheAutofocusAttribute()
-                            })
-
                             nowInitializeAlpineOnTheNewPage(Alpine)
+                            autofocusElementsWithTheAutofocusAttribute()
 
                             fireEventForOtherLibrariesToHookInto('alpine:navigated')
                             showProgressBar && finishAndHideProgressBar()
@@ -201,9 +197,8 @@ export default function (Alpine) {
                     swapCallbacks.forEach(callback => callback())
 
                     andAfterAllThis(() => {
-                        autofocus && autofocusElementsWithTheAutofocusAttribute()
-
                         nowInitializeAlpineOnTheNewPage(Alpine)
+                        autofocusElementsWithTheAutofocusAttribute()
 
                         fireEventForOtherLibrariesToHookInto('alpine:navigated')
                     })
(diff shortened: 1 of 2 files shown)


calebporzio:
Sounds good, if we can't spot any valid reason it was passed on, we should merge this and just watch out for regressions. thanks!

PR #10319, opened 2026-06-01
[4.x] Fix authorization denied with class name as the only parameter on `#[Authorize]` attribute

### LINKED ISSUE #10318: Authorization failed when only class name provided using `#[Authorize]` attribute
### Livewire version

v4.3

### Laravel version

v13

### PHP version

v8.4

Chrome on Windows

When user only provide class name (not ability name) as attribute parameter for `#[Authorize]` attribute, the authorization failed/denied. But using component authorization however works as expected.

This is because `AuthorizesRequests` guess the ability from action name and use that against resources map ability from Laravel or policy method.

However, `#[Authorize]` attribute call `Gate` directly that cause authorization failed because there was no parsing for ability name if only class name provided

```php
class PostPolicy
{
    public function createPost(User $user)
    {
        return true;
    }
}
```
```php
new class extends Component
{
    public $message;

    #[Authorize(Post::class)] // ❌
    public function createPost()
    {
        $this->authorize(Post::class); // ✅

        $this->message = 'Authorization passed!';
    }
}
```
```blade
<div>
    <button wire:click="createPost">Create Post</button>

    @if($message)
        <div>{{ $message }}</div>
    @endif
</div>
```

N/A

It should work just like component authorization or maybe additional documentation about the difference use case for component authorization and attribute authorization.

ghabriel25:
## Scenario

**Requirement**: policy method name / defined gate name should be same with component action name

When user only provide class name as parameter on `#[Authorize]` attribute, the authorization will be denied. However, using only class name as parameter on `$this->authorize()` works as expected.

Devs would expect it works either way using attribute or component trait method. Just like using transition  works both way

```php
#[Transition(type: 'forward')] // ✅
public function next()
{
    $this->transition(type: 'forward'); // ✅
}
```

## Problem
In `BaseAuthorize::call()`, it directly calls `Gate::authorize` without parsing/guessing the ability name
```php
if (is_null($this->argument)) {
    Gate::authorize($this->ability);

    return;
}
```
`$this->authorize()` flow works inside component action because when no ability provided, **AuthorizesRequests** trait will guess the ability name from action name. So for example

```php
class PostPolicy
{
    public function createPost(User $user)
    {
        return true;
    }
}
```

```php
public function createPost()
{
    $this->authorize(Post::class); // ✅
}
```
Laravel will determine that policy name would be `createPost`.

However, this flow breaks when its called inside attribute because the method name from backtracing would be `call` NOT `createPost`.
```php
#[Authorize(Post::class)] // ❌
public function createPost()
{
    //
}
```

## Solution

Either way, this will make `#[Authorize]` attribute align with `AuthorizesRequests` from component.

**Option 1** (This PR)

Use **AuthorizesRequests** trait inside **BaseAuthorize** and override the `parseAbilityAndArguments` method to return action name instead backtracing method name.

```php
protected function parseAbilityAndArguments($ability, $arguments)
{
    $ability = enum_value($ability);

    if (is_string($ability) && ! str_contains($ability, '\\')) {
        return [$ability, $arguments];
    }

    // Removed
    // $method = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 3)[2]['function'];

    return [$this->normalizeGuessedAbilityName($this->getName()), $ability];
}
```

**Option 2** (Avoid trait conflict)

- Check the ability using `class_exists` or contains `\`
- If exists and no argument provided, parse the ability (just like `AuthorizesRequests` did) before authorization.

Something like

```php
public function call(array $parameters) : void
{
    // Action that does not require a model or class...
    if (is_null($this->argument)) {
        [$ability, $arguments] = $this->parseAbilityAndArguments($this->ability, $this->argument);

        Gate::authorize($ability, $arguments);

        return;
    }

    $arguments = Arr::wrap($this->argument);

    // Resolve each argument (prioritize method parameters first, then component properties)
    $resolved = [];
    foreach ($arguments as $arg) {
        $resolved[] = $this->resolveArgument($arg, $parameters);
    }

    Gate::authorize($this->ability, $resolved);
}

protected function parseAbilityAndArguments($ability, $arguments)
{
    $ability = enum_value($ability);

    if (is_string($ability) && ! str_contains($ability, '\\')) {
        return [$ability, $arguments];
    }

    return [$this->getName(), $ability];
}
```

## File Changed

Fixes: https://github.com/livewire/livewire/issues/10318

### DIFF
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -78,6 +78,6 @@ class BaseAuthorize extends LivewireAttribute
             return [$ability, $arguments];
         }
 
-        return [$this->getName(), $ability];
+        return [$this->normalizeGuessedAbilityName($this->getName()), $ability];
     }
 }
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -3,14 +3,18 @@
 namespace Livewire\Features\SupportAuthorization;
 
 use Attribute;
+use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Facades\Gate;
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
 use UnitEnum;
+
+use function Illuminate\Support\enum_value;
 
 #[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_METHOD)]
 class BaseAuthorize extends LivewireAttribute
 {
+    use AuthorizesRequests;
+
     public function __construct(
         public UnitEnum|string $ability,
         public array|string|null $argument = null,
@@ -20,7 +24,7 @@ class BaseAuthorize extends LivewireAttribute
     {
         // Action that does not require a model or class...
         if (is_null($this->argument)) {
-            Gate::authorize($this->ability);
+            $this->authorize($this->ability);
 
             return;
         }
@@ -33,7 +37,7 @@ class BaseAuthorize extends LivewireAttribute
             $resolved[] = $this->resolveArgument($arg, $parameters);
         }
 
-        Gate::authorize($this->ability, $resolved);
+        $this->authorize($this->ability, $resolved);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


joshhanley (approved):
@ghabriel25 thanks for the PR! Makes sense to me.

I pushed a small change to use `normalizeGuessedAbilityName()` so the attribute path matches Laravel’s `AuthorizesRequests` behaviour.

These two should be equivalent:

```php
public function store()
{
    $this->authorize(Post::class);
}
```

```php
#[Authorize(Post::class)]
public function store()
{
    //
}
```

Laravel maps the guessed `store` ability to `create`, so both forms should authorise against `PostPolicy::create()`. Without that mapping, the attribute form would try `PostPolicy::store()` instead.

ghabriel25:
@joshhanley Thanks!

Issue #10278, opened 2026-05-11
Islands don't carry imports


ganyicz:
### Livewire version

v4.3.0

### Laravel version

v13.0.0

### PHP version

8.4.0

### Browser and operating system

Chrome on MacOS

### Describe the issue you're experiencing

Islands throw when an imported class is used inside of it

### Code snippets to reproduce the issue

```blade
<?php

use Carbon\Carbon;

new class extends \Livewire\Component
{
    public function with()
    {
        return ['timestamp' => Carbon::now()->timestamp];
    }
} ?>

<div class="h-dvh flex flex-col items-center justify-center">
    @island
        {{ Carbon::createFromTimestamp($timestamp)->diffForHumans() }}
    @endisland
</div>
```

### Screenshots/screen recordings

<img width="1702" height="770" alt="Image" src="https://github.com/user-attachments/assets/d84610b8-a7aa-466d-93dd-b51a6fd86648" />

### How do you expect it to work?

Islands should carry over all imports from the SFC

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be removed if I haven't met the criteria above.

joshhanley:
I've had a look into this, and not exactly sure what the best option would be. For SFC, if there are imports at the top of the class, they get extracted into the view when it's compiled, so that way they are available for the view.

But with an island, it gets compiled as a separate step, so we can't make use of that logic. Plus with an MFC or class component, there won't be any imports from the class. So we'd need to do a detection for both `<?php use` and `@use` in the view and also extract them into the island view. Feels like that could be brittle 🤔

VernSG:
Hi @joshhanley, I added a draft PR with a failing regression test that
reproduces the reported SFC import case: #10322 .

The regular SFC view can resolve `Carbon`, but the separately compiled island
view fails with:

```text
Class "Carbon" not found
```

I kept the PR test-only for now because the intended behavior for @use, MFC,
and class components still seems open for discussion. Happy to add the
implementation once the preferred direction is confirmed.

joshhanley:
@VernSG I don't have a preferred direction at this stage, but if you have a possible solution, then we'd welcome a PR! 🙂

VernSG:
[...]

PR #10322, opened 2026-06-02
[4.x] add failing test for SFC island imports

VernSG:
Change-Id: Ifa6a9ac47ee380b8a0e3ad4a8b83c2353c81f1eb

## Description

This reproduction intentionally covers the SFC import case reported in #10278.
It does not attempt to define the expected behavior for `@use`, MFC, or class
components while the implementation direction is still being discussed.

Imports declared in an SFC are available in the regular component view, but
are not carried into the separately compiled island view.

The fixture uses:

```php
use Carbon\Carbon;
and references Carbon from inside an @island block.
```

## Current behavior
The new test fails with:
```test
Class "Carbon" not found
```

## Expected behavior
Imports declared in the SFC should also be available inside its islands.

## Scope
This PR intentionally contains only a failing regression test. It does not
include a framework fix.

## Test

```bash
vendor/bin/phpunit --configuration phpunit.xml.dist \
  --testsuite Unit \
  --filter test_sfc_island_carries_imports_from_the_component
```

This test currently fails as expected with:

```text
Class "Carbon" not found
```

### DIFF
--- a/src/Features/SupportIslands/UnitTest.php
+++ b/src/Features/SupportIslands/UnitTest.php
@@ -9,6 +9,15 @@ use Illuminate\Support\Facades\File;
 
 class UnitTest extends TestCase
 {
+    public function test_sfc_island_carries_imports_from_the_component()
+    {
+        app('livewire.finder')->addLocation(viewPath: __DIR__ . '/fixtures');
+
+        Livewire::test('sfc-island-with-import')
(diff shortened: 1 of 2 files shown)


joshhanley:
@VernSG thanks for the PR! I'm going to close this as the issue already has a detailed reproduction of the issue. But if you have a solution, I would love to see a PR for it!

PR #10323, opened 2026-06-02
[4.x] Replace `EndpointResolver::updatePath()` with path from callback

ghabriel25:
In `setUpdateRoute()`, the update path already passed into the callback
```php
{
    
    ...
}
```
but in `boot()` also use `EndpointResolver::updatePath()` again instead using path from the callback

```php
// app($this::class)->setUpdateRoute(function ($handle) {  // ❌
    // return Route::post(EndpointResolver::updatePath(), $handle) // ❌

app($this::class)->setUpdateRoute(function ($handle, $path) { // ✅
    return Route::post($path, $handle) // ✅
        ->middleware(['web', RequireLivewireHeaders::class])
        ->name('default-livewire.update');
});
```

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -23,8 +23,8 @@ class HandleRequests extends Mechanism
         // Register the default route immediately (before routes files load)
         // so it's positioned before any catch-all routes.
         if (! $this->updateRoute && ! $this->updateRouteExists()) {
-            app($this::class)->setUpdateRoute(function ($handle) {
-                return Route::post(EndpointResolver::updatePath(), $handle)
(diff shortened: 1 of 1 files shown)


joshhanley (approved):
@ghabriel25 thanks for the PR! Yep all looks good to me.

The callback signature of `$handle, $path` is what we recommend people use in the installation docs too, so makes sense to change this inline with that.  https://livewire.laravel.com/docs/4.x/installation#customizing-livewires-update-endpoint

```php
use Livewire\Livewire;
 
{
    public function boot()
    {
        Livewire::setUpdateRoute(function ($handle, $path) {
            return Route::post('/custom' . $path, $handle);
        });
    }
}
```

calebporzio:
thanks!

PR #10327, opened 2026-06-04
[4.x] Prevent file download responses from leaking into action returns

### LINKED ISSUE #10326: File download responses are added to `effects.returns` and break JSON encoding
### Livewire version

v4.3.1

### Laravel version

v13.13.0

### PHP version

8.4.1

Chrome 48.0.7778.181 on Arm-64 / MacOs 26.5.1

First of all: thank you very much for the outstanding work on Livewire! 
Using it happliy since 2023 :-)

# Bug report: 
File download responses are added to `effects.returns` and break JSON encoding

## Environment

- PHP: 8.4.1
- Laravel: 13.13.0
- Livewire: 4.x
- Browser: Chrome
- Livewire assets loaded via `@livewireScripts`
- `config/livewire.php`: `inject_assets => false`

## Minimal reproduction

```php
<?php

namespace App\Http\Livewire;

use Livewire\Component;
use Illuminate\Support\Facades\Storage;

class BugPlayground extends Component
{
    public function downloadFileDownloadWay()
    {
        return Storage::disk('local')->download(
            'livewire-downloads/test.txt',
            'test-download-way.txt',
            [
                'Content-Type' => 'text/plain',
            ]
        );
    }

    public function downloadFileStreamingWay()
    {
        return response()->streamDownload(function () {
            echo Storage::disk('local')->get('livewire-downloads/test.txt');
        }, 'test-streaming-way.txt', [
            'Content-Type' => 'text/plain',
        ]);
    }

    public function render()
    {
        return view('livewire.spielwiese');
    }
}
```

```blade
<div>
    <button wire:click="downloadFileDownloadWay">
        Download via Storage::download
    </button>

    <button wire:click="downloadFileStreamingWay">
        Download via streamDownload
    </button>
</div>
```

The file exists and is valid. The same issue also happens with generated PDF files from DomPDF and Browsershot, but the problem is independent of PDF generation, encoding etc. A plain 8-byte text file is sufficient to reproduce it.

## Actual behaviorPdf

The Livewire request fails with:

```text
InvalidArgumentException
Type is not supported
```

The exception is thrown while Laravel is creating the final `JsonResponse`.

## Important debug findings

`SupportFilePdfloads` correctly detects the download response and correctly creates a JSON-safe download effect.

The debug output from `SupportFileDownloads::dehydrate()` shows:

```json
{
  "download_keys": ["name", "content", "contentType"],
  "name_type": "string",
  "content_type": "string",
  "content_length": 12,
  "contentType_type": "string",
  "download_json_ok": true,
  "download_json_error": "No error"
}
```
So the `download` effect itself is valid.

However, the final Livewire response still contains the original response object inside `effects.returns`:

```json
{
  "bad_paths": [
    {
      "path": "$.components.0.effects.returns.0",
      "type": "Symfony\\Component\\HttpFoundation\\StreamedResponse",
      "json_error": 0,
      "json_error_msg": "No error",
      "public_vars": [
        "version",
        "statusCode",
        "statusText",
        "charset"
      ]
    }
  ],
  "structure": {
    "components": [
      {
        "effects": {
          "returns": [
            "object:Symfony\\Component\\HttpFoundation\\StreamedResponse"
          ],
          "html": "string:1592",
          "download": {
            "name": "string:21",
            "content": "string:12",
            "contentType": "string:10"
          }
        }
      }
    ],
    "assets": []
  }
}
```

This seems to be the root cause: Livewire correctly creates the `download` effect, but the original `StreamedResponse` / `BinaryFileResponse` is still also added to `effects.returns`. Laravel then tries to JSON-encode that response object and fails with `JSON_ERROR_UNSUPPORTED_TYPE`.

_No response_

## Expected behavior

When a component action returns a `StreamedResponse` or `BinaryFileResponse`, Livewire should:

1. convert it into the `download` effect via `SupportFileDownloads`, and
2. not include the original response object in `effects.returns`.

The method return slot should probably remain aligned but contain `null`.

## Relevant code location

In `Livewire\Mechanisms\HandleComponents\HandleComponents::callMethods()` the return value is currently added directly:

```php
$returns[] = $finish($return);
```

and later:

```php
$componentContext->addEffect('returns', $returns);
```

This allows `StreamedResponse` / `BinaryFileResponse` instances to leak into the JSON response.

## Suggested fix

Normalize action return values before adding them to the `returns` effect.

```php
use Symfony\Component\HttpFoundation\StreamedResponse;
use Symfony\Component\HttpFoundation\BinaryFileResponse;
```

```php
protected function normalizeReturnForEffect($return)
{
    if ($return instanceof StreamedResponse || $return instanceof BinaryFileResponse) {
        return null;
    }

    return $return;
}
```

Then in `callMethods()`:

```php
if ($earlyReturnCalled) {
    $returns[] = $this->normalizeReturnForEffect($finish($earlyReturn));

    continue;
}
```

and:

```php
$returns[] = $this->normalizeReturnForEffect($finish($return));
```

After this change, the final response contains the valid `download` effect, while `effects.returns.0` becomes `null` instead of containing the non-serializable response object.

## Why this seems correct

The file response is not useful as a JavaScript action return value. The meaningful frontend instruction is already represented by the `download` effect created by `SupportFileDownloads`.

Keeping the array position in `returns` as `null` preserves call alignment while avoiding an unserializable PHP response object in the JSON payload.

joshhanley:
# The Scenario

A Livewire action can return a Laravel file download response. Livewire correctly converts the response into a `download` effect, but the original response object is also included in the action `returns` effect.

```php
use Illuminate\Support\Facades\Storage;
use Livewire\Component;

class BugPlayground extends Component
{
    public function downloadFileDownloadWay()
    {
        return Storage::disk('local')->download(
            'livewire-downloads/test.txt',
            'test-download-way.txt',
            [
                'Content-Type' => 'text/plain',
            ]
        );
    }
}
```

# The Problem

`SupportFileDownloads` detects `StreamedResponse` and `BinaryFileResponse` instances and converts them into a JSON-safe `download` effect.

However, `HandleComponents::callMethods()` still stores the original action return value in `effects.returns`. For file downloads, that means a Symfony response object can leak into the final Livewire JSON payload.

Returning `null` from the file-download hook is not enough with the current hook pipeline: component hook finish callbacks are used for side effects, and the event bus treats `null` as “keep the original forwarded value”.

# The Solution

After method call hooks finish, Livewire now normalises `StreamedResponse` and `BinaryFileResponse` action returns to `null` before adding them to `effects.returns`.

The `download` effect remains the browser-facing instruction, and the `returns` array keeps a `null` entry so action return indexes stay aligned with the submitted calls.

Fixes #10326

### DIFF
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -13,6 +13,8 @@ use Livewire\Exceptions\TooManyCallsException;
 use Livewire\Drawer\Utils;
 use Livewire\Features\SupportFormObjects\Form;
 use Illuminate\Support\Facades\View;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\StreamedResponse;
 
 class HandleComponents extends Mechanism
 {
@@ -676,7 +678,9 @@ class HandleComponents extends Mechanism
             $finish = trigger('call', $root, $method, $params, $componentContext, $returnEarly, $metadata, $idx);
 
             if ($earlyReturnCalled) {
-                $returns[] = $finish($earlyReturn);
+                $return = $finish($earlyReturn);
+
+                $returns[] = $return instanceof StreamedResponse || $return instanceof BinaryFileResponse ? null : $return;
 
                 continue;
             }
@@ -697,7 +701,9 @@ class HandleComponents extends Mechanism
             $return = wrap($root)->{$method}(...$params);
             if (config('app.debug')) trigger('profile', 'call'.$idx, $root->getId(), [$start, microtime(true)]);
 
-            $returns[] = $finish($return);
+            $return = $finish($return);
+
+            $returns[] = $return instanceof StreamedResponse || $return instanceof BinaryFileResponse ? null : $return;
 
             // Support `Wire:click.renderless`...
             if ($metadata['renderless'] ?? false) {
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -680,6 +680,8 @@ class HandleComponents extends Mechanism
             if ($earlyReturnCalled) {
                 $return = $finish($earlyReturn);
 
+                // File downloads are sent to the browser through the download effect,
+                // so keep the return slot aligned without leaking the response object into JSON.
                 $returns[] = $return instanceof StreamedResponse || $return instanceof BinaryFileResponse ? null : $return;
 
                 continue;
@@ -703,6 +705,8 @@ class HandleComponents extends Mechanism
 
             $return = $finish($return);
 
+            // File downloads are sent to the browser through the download effect,
+            // so keep the return slot aligned without leaking the response object into JSON.
             $returns[] = $return instanceof StreamedResponse || $return instanceof BinaryFileResponse ? null : $return;
 
             // Support `Wire:click.renderless`...
--- a/src/Features/SupportFileDownloads/UnitTest.php
+++ b/src/Features/SupportFileDownloads/UnitTest.php
@@ -6,6 +6,7 @@ use Illuminate\Contracts\Support\Responsable;
 use Illuminate\Support\Facades\Storage;
 use Livewire\Component;
 use Livewire\Livewire;
+use Livewire\Mechanisms\HandleRequests\EndpointResolver;
 use PHPUnit\Framework\ExpectationFailedException;
 
 class UnitTest extends \Tests\TestCase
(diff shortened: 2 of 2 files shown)


src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -697,7 +701,9 @@ protected function callMethods($root, $calls, $componentContext)
             $return = wrap($root)->{$method}(...$params);
             if (config('app.debug')) trigger('profile', 'call'.$idx, $root->getId(), [$start, microtime(true)]);
 
-            $returns[] = $finish($return);
+            $return = $finish($return);
+
+            $returns[] = $return instanceof StreamedResponse || $return instanceof BinaryFileResponse ? null : $return;
```
ghabriel25, line 710: How about `array_filter` outside loop?

joshhanley: > the returns array keeps a null entry so action return indexes stay aligned with the submitted calls.

No, the returns index still needs to exist

ghabriel25: Ah sorry, I mean `array_map` but its okay 👍

calebporzio:
thanks! (don't love the one-liner ternary there though - think it would be nicer as a discrete conditional before the assignment/return)

joshhanley:
@calebporzio yep, I'll fix that and push it.

PR #10328, opened 2026-06-04
[4.x] Fix `wire:sort:item` ids on lazy components

### LINKED ISSUE #10325: wire:sort receives null as the ID when the sortable child component is "lazy"
### Livewire version

v4.3.1

### Laravel version

v13.13.0

### PHP version

8.3.30

Firefox on Arch Linux

When using wire:sort to sort child components, the sort handler receives the correct ID as long as lazy is not set on the child components. As soon as lazy is added, the $id parameter in the handler is always null.

Page Component:
```blade
<?php

use Livewire\Component;

new class extends Component {
    public array $widgets = [
        [
            'id' => 1,
            'title' => "Title 1",
        ],[
            'id' => 2,
            'title' => "Title 2",
        ],[
            'id' => 3,
            'title' => "Title 3",
        ],
    ];

    public function handleSort($id, $position): void
    {
        dump($id, $position);
    }
};
?>

<div>
    <div class="grid grid-cols-3 gap-4" wire:sort="handleSort">
        @foreach ($widgets as $widget)
            <livewire:widget
                wire:key="{{ $widget['id'] }}"
                wire:sort:item="{{ $widget['id'] }}"
                lazy
            >
                {{ $widget['title'] }}
            </livewire:widget>
        @endforeach
    </div>
</div>
```
Widget:
```blade
<?php

use Livewire\Component;


new class extends Component {
};
?>

@placeholder
<flux:card>
    <flux:text>Widget</flux:text>
    <flux:skeleton animate="shimmer" class="h-10 flex items-center justify-center">
        <flux:icon.loading/>
    </flux:skeleton>
</flux:card>
@endplaceholder

<flux:card
    {{ $attributes }}
>
    <flux:text>Widget</flux:text>
    <flux:heading size="xl">{{ $slot }}</flux:heading>
</flux:card>
```

Image
Image

The handleSort $id contains the value passed via wire:sort:item

joshhanley:
# The Scenario

When a sortable list renders lazy child components and forwards `wire:sort:item` to each child root element, sorting those children can call the sort action with a missing item id. The lazy child renders correctly, but the sort action receives `null` instead of the forwarded item id.

```blade
<div>
    <ul wire:sort="sortItem">
        <livewire:child wire:key="item-1" wire:sort:item="item-1" lazy>
            Item 1
        </livewire:child>

        <livewire:child wire:key="item-2" wire:sort:item="item-2" lazy>
            Item 2
        </livewire:child>
    </ul>

    <div>{{ $result }}</div>
</div>
```

# The Problem

Lazy component placeholders are Alpine-initialised before the forwarded `wire:sort:item` attribute exists on the child component root element.

When the lazy component loads, Livewire morphs the placeholder into the rendered child and adds the forwarded `wire:sort:item` attribute. However, the existing `Alpine.interceptInit` path has already run for that element, so Alpine Sort never initialises the equivalent `x-sort:item` directive and never sets `_x_sort_key`.

That means when Sortable fires, Alpine Sort cannot find the item key and Livewire receives `null` for the item id.

# The Solution

This PR adds support for initialising `wire:sort:item` when the directive is added after an element has already gone through Alpine initialisation.

That covers lazy component roots, where the placeholder is initialised first and the forwarded `wire:sort:item` attribute is only applied after the lazy component renders. When that later directive initialisation happens, Livewire binds the matching `x-sort:item` directive so Alpine Sort can register the item key and pass it through to the sort action.

Fixes #10325

### DIFF
--- a/js/features/supportWireSort.js
+++ b/js/features/supportWireSort.js
@@ -2,6 +2,23 @@ import { setNextActionOrigin } from '@/request'
 import { evaluateActionExpression } from '../evaluator'
 import Alpine from 'alpinejs'
 import { extractDirective } from '@/directives'
+import { on } from '@/hooks'
+
+on('directive.init', ({ el, directive }) => {
+    if (! directive.rawName.startsWith('wire:sort:item')) return
+
+    if (el._x_sort_key !== undefined) return
+
+    let modifierString = directive.modifiers.join('.')
+
+    let expression = directive.expression
+
+    Alpine.bind(el, {
+        ['x-sort:item' + modifierString]() {
+            return expression
+        }
+    })
+})
 
 Alpine.interceptInit(el => {
     for (let i = 0; i < el.attributes.length; i++) {
--- a/src/Features/SupportWireSort/BrowserTest.php
+++ b/src/Features/SupportWireSort/BrowserTest.php
@@ -116,6 +116,79 @@ class BrowserTest extends BrowserTestCase
         ->assertSeeIn('@result', 'item:item-1,position:2');
     }
 
+    public function test_wire_sort_item_id_is_passed_for_lazy_child_components()
+    {
+        Livewire::visit([
+            new class extends Component {
+                public $result = '';
+
+                public function sortItem($item, $position)
+                {
+                    $this->result = "item:{$item},position:{$position}";
+                }
+
+                public function render()
+                {
+                    return <<<'HTML'
+                    <div>
+                        <ul dusk="sortable" wire:sort="sortItem">
+                            <livewire:child wire:key="item-1" wire:sort:item="item-1" lazy>
+                                Item 1
+                            </livewire:child>
+
+                            <livewire:child wire:key="item-2" wire:sort:item="item-2" lazy>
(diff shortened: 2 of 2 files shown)


calebporzio:
Hmmm... this is tricky.

A) It's a workaround for a deeper lazy initialization problem
B) The code doesn't express why it exists (as a workaround for that) - could at least be a comment
C) Adding an extra `directive.init` hook is potentially performance degrading for the rest of livewire

All of these types of fixes are hacks around a bad architecture for lazy loading components.

In a perfect world the abstractions are designed such that features like `wire:sort` never have to think about laziness at all.

I understand that's not our current world.

But I just think it's worth examining other solutions or even other workarounds, or potentially exploring the bigger picture a little more before merging this fix. Thanks!

joshhanley:
I've investigated this further and there are two different scenarios here:
- conditionally added `wire:sort` attributes
- attributes forward to placeholders

So I'm actually going to open separate PRs for those and close this one. Either of which will solve the reported issue, but both necessary in their own right I think.

PR #10329, opened 2026-06-04
[3.x] Only report corrupt component payloads in debug mode

calebporzio:
## What changed

Marks `CorruptComponentPayloadException` as `ShouldntReport` so invalid/tampered Livewire snapshots still receive the existing production 419 response without being reported to Laravel exception reporters.

Adds a regression test proving a bad checksum returns 419 and does not hit registered report callbacks.

## Why

Livewire update endpoints are exposed to the public internet and get probed with stale or tampered snapshots. A checksum failure means Livewire rejected the request correctly; it should be treated as expected boundary traffic rather than application error noise in Flare/Sentry/Bugsnag.

## Validation

- `./vendor/bin/phpunit src/Mechanisms/HandleRequests/UnitTest.php`
- `./vendor/bin/phpunit src/Tests/ComponentsAreSecureUnitTest.php`
- `./vendor/bin/phpunit --testsuite Unit --filter='HandleRequests|ComponentsAreSecure'`

### DIFF
--- a/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
+++ b/src/Mechanisms/HandleComponents/CorruptComponentPayloadException.php
@@ -2,10 +2,9 @@
 
 namespace Livewire\Mechanisms\HandleComponents;
 
-use Illuminate\Contracts\Debug\ShouldntReport;
 use Livewire\Exceptions\BypassViewHandler;
 
-class CorruptComponentPayloadException extends \Exception implements ShouldntReport
+class CorruptComponentPayloadException extends \Exception
 {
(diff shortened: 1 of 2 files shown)


calebporzio:
@joshhanley - take a look at this when you get a sec. if you agree and like it, merge it and also put it in 4.x if you can. thanks mate

joshhanley:
@calebporzio had a look and Laravel 10 doesn't have the `ShouldntReport` contract. My other concern is that I think this should report `app.debug` is enabled and silenced if it's not, the same as the 419 response.

I've updated the PR to support these.

PR #10330, opened 2026-06-05
[4.x] Fix flaky test in `PersistentMiddleware` browser test

ghabriel25:
## Problem
The `test_that_authorization_middleware_is_re_applied()` test sometimes failed due to a race condition. The test decorates the browser's fetch function to capture request arguments in `localStorage`, but it was attempting to access those arguments before they were reliably set.

## Root Cause
After triggering a Livewire action with `click('@changeProtected')`, the test immediately navigated to a new page and tried to parse `localStorage.getItem('lastFetchArgs')`. However, the asynchronous fetch interception that stores these arguments may not have completed yet, resulting test failures that only occurred sometimes (when network/timing was slow)

## Solution
Added an explicit wait using `waitUntil()` to ensure `lastFetchArgs` is present in `localStorage` before navigating to the next page.

### DIFF
--- a/src/Mechanisms/PersistentMiddleware/BrowserTest.php
+++ b/src/Mechanisms/PersistentMiddleware/BrowserTest.php
@@ -192,10 +192,12 @@ JS;
                 $script = <<<'JS'
                     let unDecoratedFetch = window.fetch
                     let decoratedFetch = (...args) => {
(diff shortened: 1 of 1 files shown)


joshhanley:
Thanks!

PR #10336, opened 2026-06-08
[4.x] Fix duplicate Alpine x-if elements after back/forward navigation

### LINKED ISSUE #10335: Livewire starter kit bug with passkeys button
### Livewire version

v4.3.1

### Laravel version

v.13.14.0

### PHP version

8.5.4

Chrome on Windows

I posted this on the laravel/maestro github, but I figured it would be useful for you to know as well:

https://github.com/laravel/maestro/issues/151

.

_No response_

.

hamdyelbatal122:
## Summary

- Fixes duplicate DOM nodes when using browser back/forward buttons with `wire:navigate` and Alpine `x-if` / `x-for` templates
- Cleans up Alpine-rendered output before caching the page snapshot on `popstate`, matching the behavior already used when navigating away via a `wire:navigate` link
- Adds a browser regression test that reproduces the passkey button duplication reported in the Laravel starter kit, plus a Vitest test for snapshot caching behavior

Fixes #10335 (reported via [laravel/maestro#151](https://github.com/laravel/maestro/issues/151))

## Root cause

When leaving a page via `wire:navigate`, Livewire calls `cleanupAlpineElementsOnThePageThatArentInsideAPersistedElement()` before storing the page in the back-button snapshot cache. This removes Alpine `x-if` / `x-for` rendered output so only the original `<template>` tags are cached.

When leaving a page via the browser back/forward buttons, the snapshot was updated **without** this cleanup. The cached HTML then contained both the `<template x-if>` tag and Alpine's rendered output. On restore, Alpine initialized again and rendered duplicate elements.

## Test plan

- [x] `npm run test:run` — all Vitest tests pass (including new `history.spec.js`)
- [x] `composer test:unit` — all unit tests pass (1267 tests)
- [x] Browser test `test_alpine_if_does_not_duplicate_elements_after_back_and_forward_navigation` passes
- [x] Verified the new browser test **fails without the fix** (2 buttons instead of 1 after repeated back/forward)
- [x] Existing navigate regression tests pass: `test_alpine_for_loop_still_functions_after_navigation`, `test_back_and_forward_buttons_work_after_page_refresh`

### DIFF
--- a/js/plugins/navigate/history.spec.js
+++ b/js/plugins/navigate/history.spec.js
@@ -1,0 +1,106 @@
+import { describe, it, expect, beforeEach } from 'vitest'
+import {
+    pushUrl,
+    updateCurrentPageHtmlInSnapshotCacheForLaterBackButtonClicks,
+    whenTheBackOrForwardButtonIsClicked,
+} from './history'
+
+function setPageHtml(bodyHtml) {
+    document.documentElement.innerHTML = `<head></head><body>${bodyHtml}</body>`
+}
+
+function countRenderedButtonsInHtml(html) {
+    let doc = (new DOMParser()).parseFromString(html, 'text/html')
+    let container = doc.getElementById('alpine-if-container')
+
+    if (! container) return 0
+
+    let count = 0
+
+    container.querySelectorAll('template[x-if]').forEach(template => {
+        let sibling = template.nextElementSibling
+
+        while (sibling) {
+            if (sibling.matches('button')) count++
+
+            sibling = sibling.nextElementSibling
+        }
+    })
+
+    return count
+}
+
+// Mirrors Alpine.destroyTree() behavior for x-if: rendered nodes live as siblings after the template.
+function cleanupAlpineXIfOutput(root = document.body) {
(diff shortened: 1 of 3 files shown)


joshhanley:
@hamdyelbatal122 thanks for the PR! Unfortunately all of your responses and this PR seem AI generated. Because of that, I had to investigate this issue myself to ensure we can trust the fix and as such, submitted my own PR. So I'm going to close this PR. We've got no problem with you using AI to assist with investigation and fixes (we do it too!), but please be sure to review the code and write the responses yourself, so we can ensure it has been thoroughly checked and fixed. Thanks for being willing to contribute!

hamdyelbatal122:
Thanks @joshhanley — I understand why the PR came across that way, and I appreciate you taking the time to verify the fix properly in #10339.

To be clear: I investigated the issue and wrote the fix myself. I only used AI to polish the wording and format the PR description — not for the diagnosis or the code change. That was my mistake in presentation, and I should have kept the PR text simpler and in my own voice.

I've looked through #10339 and the approach matches what I found. Thanks for closing the loop on #10335, and I'll be more careful with how I present future contributions.

PR #10338, opened 2026-06-09
[4.x] Fix SFC island imports

### LINKED ISSUE #10278: Islands don't carry imports

VernSG:

Yes. This fixes #10278. The issue already has a detailed reproduction, and Josh mentioned that a PR with a possible solution would be welcome.

Yes. The branch is named `fix-sfc-island`.

No. This only changes SFC island compilation so imported PHP aliases are available inside separately compiled island views.

Yes. I added a regression test for an SFC island using an imported `Carbon` alias.

SFC views can include PHP imports like:

```php
use Carbon\Carbon;
```

The compiled parent SFC view can resolve that alias correctly, but `@island` blocks are extracted into separate compiled Blade files. Those extracted island views previously did not receive the parent view's PHP imports, so using the same alias inside an island failed with:

```text
Class "Carbon" not found
```

For example:

```blade
@island(name: 'timestamp', with: ['timestamp' => Carbon::parse('2024-01-01')->timestamp])
    <div>{{ Carbon::createFromTimestamp($timestamp)->year }}</div>
@endisland
```

This PR carries PHP `use` statements from the compiled SFC view into each separately compiled island view, allowing imported aliases to work in both the island `with:` expression and island body.

Tests:

```bash
composer test:unit -- --filter="sfc_island_can_use_imports_from_parent_view"
composer test:unit -- --filter="SupportIslands"
```

Thanks for contributing! 🙌

### DIFF
--- a/src/Compiler/Parser/Parser.php
+++ b/src/Compiler/Parser/Parser.php
@@ -209,15 +209,7 @@ PHP
     {
         // Extract everything between <?php and "new"
         if (preg_match('/\<\?php(.*?)new/s', $classPortion, $matches)) {
-            $preamble = $matches[1];
-
-            // Use (*SKIP)(*FAIL) to ignore "use" statements inside comments...
-            $skipComments = '(?:\/\/[^\n]*|\/\*.*?\*\/)(*SKIP)(*FAIL)';
-
-            // Extract all use statements
-            if (preg_match_all('/' . $skipComments . '|use\s+[^;]+;/s', $preamble, $useMatches)) {
-                $useStatements = implode("\n", $useMatches[0]);
-            }
+            $useStatements = static::extractUseStatements($matches[1]);
         }
 
         // Only add PHP tags and use statements if we found any
@@ -226,5 +218,18 @@ PHP
         }
 
         return $contents;
+    }
+
+    public static function extractUseStatements(string $contents): ?string
+    {
+        // Use (*SKIP)(*FAIL) to ignore "use" statements inside comments...
+        $skipComments = '(?:\/\/[^\n]*|\/\*.*?\*\/)(*SKIP)(*FAIL)';
+
+        // Extract all use statements (excluding closure `use (...)` clauses)...
+        if (! preg_match_all('/' . $skipComments . '|\buse\s+[^;()]+;/s', $contents, $useMatches)) {
+            return null;
+        }
+
+        return $useMatches[0] ? implode("\n", $useMatches[0]) : null;
     }
 }
--- a/src/Features/SupportIslands/Compiler/IslandCompiler.php
+++ b/src/Features/SupportIslands/Compiler/IslandCompiler.php
@@ -98,6 +98,9 @@ class IslandCompiler
         $scopeProviderCode = $this->generateScopeProviderCode($expression);
         $innerContent = $scopeProviderCode . $innerContent;
 
+        // Carry SFC import aliases into separately compiled island views...
+        $innerContent = $this->injectUseStatementsFromCompiledView($innerContent);
+
         // Ensure the cached directory exists...
         File::ensureDirectoryExists(dirname($cachedPath));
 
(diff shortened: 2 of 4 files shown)


joshhanley (changes requested):
@VernSG thanks for the PR! I don't think this covers everything I raised in the issue regarding both php use blocks and blade `@use` blocks. Can you confirm? I think it would be good to see multiple tests around that. Also some comments above the regexs explaining what they do would be good.

VernSG:
> @VernSG thanks for the PR! I don't think this covers everything I raised in the issue regarding both php use blocks and blade `@use` blocks. Can you confirm? I think it would be good to see multiple tests around that. Also some comments above the regexs explaining what they do would be good.

Thanks, you're right. I've updated the PR to cover both PHP `use` blocks and Blade `@use(...)` directives, with separate regression tests for each. I also added comments above the regexes.

Tests:
- `composer test:unit -- --filter="sfc_island_can_use_.*imports_from_parent_view"`
- `composer test:unit -- --filter="SupportIslands"`

calebporzio:
Thanks for this!

Question: Given that we already find and extract import/use statements from the top of an SFC, can we not just re-use that exact same logic or whatever for islands? Seeing the extra regexes and logic in here give me pause.

Lmk, thanks! Ideally we keep this as simple as possible.

VernSG:
> Thanks for this!
> 
> Question: Given that we already find and extract import/use statements from the top of an SFC, can we not just re-use that exact same logic or whatever for islands? Seeing the extra regexes and logic in here give me pause.
> 
> Lmk, thanks! Ideally we keep this as simple as possible.

  Thanks, that makes sense.

  Functionally, the current change does fix the issue and covers both PHP `use` and Blade
  `@use(...)` with separate tests. That said, I agree the implementation may be doing too
  much inside `IslandCompiler`.

  The extra regex logic was added to make both import styles available inside separately
  compiled island views, but I agree it would be better to reuse the existing SFC import
  handling if possible rather than maintain another extraction path here.

  I’ll take another pass and try to simplify it around the existing logic.

VernSG:
> Thanks for this!
> 
> Question: Given that we already find and extract import/use statements from the top of an SFC, can we not just re-use that exact same logic or whatever for islands? Seeing the extra regexes and logic in here give me pause.
> 
> Lmk, thanks! Ideally we keep this as simple as possible.

Im simplified this.

The island compiler now compiles the SFC prefix first and reuses the leading import blocks from that compiled output, instead of keeping separate parsing paths for PHP `use` and Blade `@use(...)`.

So `@use(...)` is handled by Blade first, and the island logic just reuses the result.

I also kept regression coverage for both cases.

calebporzio:
Thanks @VernSG! I took a pass at this and pushed a commit that simplifies the implementation along the lines of my earlier comment:

- The SFC compiler already extracts `use` statements when hoisting class imports into the compiled view (`Parser::injectUseStatementsFromClassPortion`). I pulled that extraction out into a shared `Parser::extractUseStatements()` helper, and the island compiler now reuses it on the leading PHP block(s) of the view — so there's one canonical "find the use statements" implementation instead of two.
- `@use(...)` directives are now just copied into the island file as-is and compiled by Blade there — this avoids re-entrantly invoking `compileString()` on the blade compiler from inside its own `prepareStringsForCompilationUsing` hook, which the previous version did.
- Imports are extracted once per view instead of re-compiled for every island occurrence.
- Added an end-to-end regression test that mounts a real SFC fixture (imports in the class portion, used inside an `@island`) — the exact repro from #10278. Your two compiled-view-level tests are kept as-is.

Verified all three new tests fail with `Class "Carbon" not found` against main and pass with the fix, and the full unit suite is green (1298 tests).

PR #10334, opened 2026-06-06
[4.x] Fix effect overidden by `wire:loading.attr` on renderless action

MaxencePaulin:
Review the contribution guide first at: https://livewire.laravel.com/docs/contribution-guide

1️⃣ Is this something that is wanted/needed? Did you create a discussion about it first?
Not a discussion but an issue about it.

2️⃣ Did you create a branch for your fix/feature? (Main branch PR's will be closed)
Yes

3️⃣ Does it contain multiple, unrelated changes? Please separate the PRs out.
No

4️⃣ Does it include tests? (Required)
Yes

5️⃣ Please include a thorough description (including small code snippets if possible) of the improvement and reasons why it's useful.

I did a detailed issue here : https://github.com/livewire/livewire/issues/10332

I added a failing test and a potential fix. I don't have full knowledge of Livewire core, you can tell me if it's not too easy to fix :)

Fixes https://github.com/livewire/livewire/issues/10332

Thanks for contributing! 🙌

PR #10340, opened 2026-06-09
[4.x] Restore existing attributes after `wire:loading.attr`

### LINKED ISSUE #10332: `x-bind:disabled` binding overridden after renderless action when combined with `wire:loading.attr="disabled"`
### Livewire version

v4.3.1 (v3 too)

### Laravel version

v13.13.0

### PHP version

8.4.14

Brave on Windows

When a button has both `wire:loading.attr="disabled"` and `x-bind:disabled` bound to a condition or property, triggering a #[Renderless] action from the component causes the `x-bind:disabled` binding to be overridden and stop working correctly.

In this example, `x-bind:disabled` should permanently evaluate to true, but after a renderless action is called, the button becomes enabled.

1. Observe that the "Some action" button is disabled
2. Click "Renderless action" to trigger a renderless action
3. Observe that the "Some action" button is no longer disabled

Here is a live Wirebox to reproduce it : https://wirebox.app/b/xg1g6

You can see directly the code snippets used below :

```php
<?php
namespace App\Livewire;

use Livewire\Component;
use Livewire\Attributes\Renderless;

{
    public bool $disabled = true;

    public function render()
    {
        return <<<'BLADE'
            <div
            >
                <form wire:submit="someAction" class="bg-blue-600 p-[10px]">
                    <button
                        type="button"
                        wire:click="renderlessAction"
                        class="px-4 py-2 bg-gray-700 text-white rounded-md cursor-pointer hover:bg-gray-800 disabled:opacity-50 disabled:cursor-not-allowed"
                    >
                        Renderless Action
                    </button>

                    <button
                        type="submit"
                        x-bind:disabled="$wire.$get('disabled')"
                        wire:loading.attr="disabled"
                        class="px-4 py-2 bg-green-700 text-white rounded-md cursor-pointer hover:bg-green-800 disabled:opacity-50 disabled:cursor-not-allowed"
                    >
                        Some action
                    </button>
                </div>
            </div>
        BLADE;
    }

    public function someAction(): void
    {
        //
    }

    #[Renderless]
    public function renderlessAction(): void
    {
        //
    }
}
```

//

The x-bind:disabled binding should be work, even when a renderless action is called.

> [!IMPORTANT]  
>Note that the issue does not occur when either removing the #[Renderless] attribute from the action, or adding wire:click to the button that has `wire:loading.attr="disabled"`.

joshhanley:
# The Scenario

If an element already has an attribute set, such as by Alpine, and that same attribute is changed by `wire:loading.attr`, the attribute can be removed when loading finishes.

In the browser this can leave a button enabled even though the Alpine/Livewire state still says it should be disabled. Renderless actions make this visible because there is no morph afterwards to re-apply the attribute.

CleanShot 2026-06-09 at 21 34 49

```blade

<button
    x-bind:disabled="$wire.$get('disabled')"
    wire:loading.attr="disabled"
>
    Submit
</button>
```

# The Problem

`wire:loading.attr` currently removes the attribute when loading ends. That is correct when Livewire added the attribute temporarily, but incorrect when the attribute already existed before loading started.

A later morph or Alpine effect can sometimes hide the issue by re-applying the attribute. Renderless actions expose it because they do not trigger a morph, so the attribute remains removed.

# Potential solutions

- **Re-run Alpine effects after renderless responses** (#10334) — fixes the reported case but only covers `x-bind`, relies on a private Alpine API, and misses server-rendered attributes.
- **Skip applying the attribute when it already exists** — avoids clobbering, but then loading state can't override the attribute's value during the request (e.g. restore an original `aria-*` value afterwards).
- **Capture the pre-loading state and restore it when loading ends** — general, and matches the save/restore precedent already used for `display` in `toggleBooleanStateDirective`. Chosen.

# The Solution

`toggleBooleanStateDirective` now captures the attribute state before applying `wire:loading.attr` and returns a restore callback that puts the attribute back exactly as it was — re-setting the original value if it existed, removing it if it didn't (which also fixes `.attr.remove` re-adding a never-present attribute with the value `"true"`).

`wire-loading` tracks overlapping loading cycles with a counter so that:

- the pre-loading state is only captured on the first of any overlapping triggers (otherwise a later trigger would capture the state a previous one already set and the attribute would never unset), and
- the state is only restored once every overlapping trigger has finished (previously the first cycle to finish removed the attribute while others were still loading).

Because the counter requires every start to be matched by an end, `applyDelay` now keeps per-cycle `timeout`/`started` state. Previously all cycles shared one closure, so with `.delay` two overlapping cycles could swallow one end callback — harmless before (the next cycle self-healed) but permanent with a counter (it would wedge above zero and the directive would stop applying entirely). A guard against unbalanced end triggers (e.g. a stray `livewire-upload-finish`) protects the counter in the other direction.

CleanShot 2026-06-09 at 21 33 22

Fixes #10332

---

Note: `wire:loading.class` still has the sibling issue (a pre-existing class is removed when loading ends), as do `wire:dirty.attr`/`wire:offline.attr` which share `toggleBooleanStateDirective` but don't consume the restore callback. Left for a follow-up to keep this scoped.

### DIFF
--- a/js/directives/wire-loading.js
+++ b/js/directives/wire-loading.js
@@ -9,14 +9,16 @@ directive('loading', ({ el, directive, component, cleanup }) => {
 
     let [delay, abortDelay] = applyDelay(directive)
 
+    let restoreLoadingState
+
     let cleanupA = whenTargetsArePartOfRequest(component, el, targets, inverted, [
-        () => delay(() => toggleBooleanStateDirective(el, directive, true)),
-        () => abortDelay(() => toggleBooleanStateDirective(el, directive, false)),
+        () => delay(() => restoreLoadingState = toggleBooleanStateDirective(el, directive, true)),
+        () => abortDelay(() => restoreLoadingState ? restoreLoadingState() : toggleBooleanStateDirective(el, directive, false)),
     ])
 
     let cleanupB = whenTargetsArePartOfFileUpload(component, targets, [
-        () => delay(() => toggleBooleanStateDirective(el, directive, true)),
-        () => abortDelay(() => toggleBooleanStateDirective(el, directive, false)),
+        () => delay(() => restoreLoadingState = toggleBooleanStateDirective(el, directive, true)),
+        () => abortDelay(() => restoreLoadingState ? restoreLoadingState() : toggleBooleanStateDirective(el, directive, false)),
     ])
 
     cleanup(() => {
--- a/js/directives/wire-loading.js
+++ b/js/directives/wire-loading.js
@@ -24,6 +24,10 @@ directive('loading', ({ el, directive, component, cleanup }) => {
     }
 
     function endLoading() {
+        // Guard against unbalanced end triggers, otherwise a stray one would
+        // drive the count negative and stop the directive ever applying again...
+        if (activeLoadingCount === 0) return
+
         activeLoadingCount--
 
         // Only restore/remove once every overlapping loading trigger has finished...
@@ -71,23 +75,31 @@ function applyDelay(directive) {
         }
     })
 
-    let timeout
-    let started = false
+    // Each delay cycle gets its own state, otherwise overlapping loading
+    // cycles would share one timeout and swallow each other's end callback...
+    let cycles = []
 
     return [
         (callback) => { // Initiate delay...
-            timeout = setTimeout(() => {
+            let cycle = { started: false }
+
+            cycle.timeout = setTimeout(() => {
                 callback()
 
-                started = true
+                cycle.started = true
             }, duration)
+
+            cycles.push(cycle)
         },
         async (callback) => { // Execute or abort...
-            if (started) {
+            let cycle = cycles.shift()
+
+            if (! cycle) return
+
+            if (cycle.started) {
                 await callback()
-                started = false
             } else {
-                clearTimeout(timeout)
+                clearTimeout(cycle.timeout)
             }
         },
     ]
--- a/js/directives/wire-loading.js
+++ b/js/directives/wire-loading.js
@@ -9,16 +9,37 @@ directive('loading', ({ el, directive, component, cleanup }) => {
 
     let [delay, abortDelay] = applyDelay(directive)
 
+    let activeLoadingCount = 0
     let restoreLoadingState
 
+    function startLoading() {
+        // Only capture the pre-loading attribute state on the first of any
+        // overlapping loading triggers, otherwise we'd capture the state
+        // that a previous overlapping trigger already set...
+        if (activeLoadingCount === 0) {
+            restoreLoadingState = toggleBooleanStateDirective(el, directive, true)
+        }
+
+        activeLoadingCount++
+    }
+
+    function endLoading() {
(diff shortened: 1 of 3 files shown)


ghabriel25:
Just working on this for a week but still testing, my approach exactly the same. 👍 

Still have a long way to understand js section 😆

js/directives/wire-loading.js
```diff
@@ -9,14 +9,16 @@ directive('loading', ({ el, directive, component, cleanup }) => {
 
     let [delay, abortDelay] = applyDelay(directive)
 
+    let restoreLoadingState
+
     let cleanupA = whenTargetsArePartOfRequest(component, el, targets, inverted, [
-        () => delay(() => toggleBooleanStateDirective(el, directive, true)),
-        () => abortDelay(() => toggleBooleanStateDirective(el, directive, false)),
+        () => delay(() => restoreLoadingState = toggleBooleanStateDirective(el, directive, true)),
```
calebporzio: I'm good with this but want to make sure we're not introducing a race condition where the first callback gets called twice before the "undo" callback gets a chance to run.

This would create a scenario where wire:loading.attr never unsets it's value.

joshhanley: @calebporzio yeah I added a test covering this and it does cause issues. I've marked this as draft while I see if I can come up with a solution.

joshhanley:
> I'm good with this but want to make sure we're not introducing a race condition where the first callback gets called twice before the "undo" callback gets a chance to run.
> 
> This would create a scenario where wire:loading.attr never unsets it's value.

@calebporzio have added a fix for this.

calebporzio:
Reviewed this end-to-end (repro'd #10332 locally, ran the tests against main and the branch to confirm they catch the bug). The approach is right — capture/restore matches the `display` precedent in `toggleBooleanStateDirective`, and the counter fixes a real pre-existing bug where the first of two overlapping cycles to finish removed the attribute early.

One real problem found and fixed: **the counter could wedge permanently with `.delay`**. `applyDelay` shared one `timeout`/`started` closure across all cycles, so two overlapping delayed cycles swallowed one `endLoading` — before the counter that self-healed on the next request; with the counter it stuck above zero and `wire:loading.attr` stopped applying on that element **forever**. Verified with a browser test that fails on the branch as-was and passes on main. Pushed:

- `applyDelay` now keeps per-cycle state (FIFO), so every start is matched by exactly one end
- `endLoading` guards against unbalanced end triggers (stray `livewire-upload-finish` would have driven the count negative — same permanent breakage in the other direction)
- `shared.js` restore callback now always returned for the attr branch (`removeAttribute` when it didn't exist) — also fixes `.attr.remove` re-adding a never-present attribute as `"true"` when loading ends
- strengthened the overlapping test (assert the attribute survives the first finish) and added a regression test for the `.delay` wedge

Full `SupportWireLoading` browser suite (25 tests) + JS unit tests pass locally. Follow-up candidates (same root cause, out of scope here): `wire:loading.class` removing pre-existing classes, and `wire:dirty`/`wire:offline` which share the toggle but ignore the restore callback.

PR #10341, opened 2026-06-09
[4.x] Honor scoped route bindings for `mount()` method injection

mateusjunges:
## The Scenario
Consider a full page Livewire component behind scoped route bindings, where a child route key is not globally unique. Here's a common shape:

```php
Route::get('{event}/channels/{channel}', Channel::class)
    ->name('channels.show')
    ->scopeBindings();
```

`Channel` uses slug as its route key, and slugs are only unique within an event (two events can have a channel named "General" with a slug `general`). The `scopeBindings()` call tells laravel to resolve `{channel}` through its parent `{event}` relationship. The component injects the child in `mount()`:

```php
public Event $event;
public Channel $channel;

public function mount(Channel $channel): void
{
    $this->authorize('view', $channel);
}
```

## The Problem
The `Channel` injected into `mount()` was resolved with a global, unscoped lookup, so it could be the channel belonging to a different event (the first that matches that slug). Authorization and any logic afterwards will run against the wrong record, causing 403 or 404s.

The cause lives in `ImplicitRouteBinding::resolveAllParameters()`, which resolved `mount()` parameters before public properties. Resolving mount parameters runs Laravel implicit binding over only the `mount` method signature (`channel`, in this case). At this moment the parent `event` parameter is still a raw slug string rather than a resolved model, so Laravel scoped branch is skipped and it falls back to an unscoped `resolveRouteBinding`. The result is that when a model is both a typed public property and a `mount` parameter, both resolved to the unscoped record.

## The Solution
Resolve public properties before `mount` parameters. Properties resolve in route order, so the parent `event` becomes a resolved model and is seeded into the route via `setParameter` before the child resolves. By the time `mount` parameter resoulution runs, the parent is a `UrlRoutable` and laravel applies the scoped lookup.

### DIFF
--- a/src/Drawer/ImplicitRouteBinding.php
+++ b/src/Drawer/ImplicitRouteBinding.php
@@ -27,10 +27,43 @@ class ImplicitRouteBinding
 
     public function resolveAllParameters(Route $route, Component $component)
     {
-        $props = $this->resolveComponentProps($route, $component);
-        $params = $this->resolveMountParameters($route, $component);
+        // Resolve every bindable route parameter in URI order first (like Laravel's
+        // own implicit binding) so that scoped child bindings always see their
+        // parent as an already-resolved model, regardless of whether the parent
+        // is a public property or a mount() parameter...
+        $this->resolveRouteParametersInOrder($route, $component);
 
-        return $props->merge($params)->all();
+        $params = $this->resolveMountParameters($route, $component);
+        $props = $this->resolveComponentProps($route, $component);
+
+        return $params->merge($props)->all();
+    }
+
+    protected function resolveRouteParametersInOrder(Route $route, Component $component)
+    {
+        $types = $this->getPublicPropertyTypes($component)
+            ->merge($this->getMountParameterTypes($component)->filter())
+            ->filter(function ($className) {
+                return $className && (is_subclass_of($className, UrlRoutable::class) || is_subclass_of($className, BackedEnum::class));
+            });
+
+        foreach ($route->parametersWithoutNulls() as $name => $value) {
+            if (! $types->has($name)) continue;
+
+            $route->setParameter($name, $this->resolveParameter($route, $name, $types->get($name)));
+        }
+    }
+
+    protected function getMountParameterTypes(Component $component)
+    {
+        if (! method_exists($component, 'mount')) {
+            return new Collection();
+        }
+
+        return collect((new ReflectionMethod($component, 'mount'))->getParameters())
+            ->mapWithKeys(function ($parameter) {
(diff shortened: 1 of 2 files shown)


joshhanley (approved):
@mateusjunges thanks for the PR! Looks good to me. Yeah makes sense that we resolve component props before mount params.

calebporzio:
Great find @mateusjunges, and the diagnosis is spot on — the mount pass runs Laravel's implicit binding over just the `mount()` signature, so the parent is still a raw string and the scoped branch gets skipped. The tests are exactly the style we want too (real routes, real requests).

One problem with the reorder though: it swapped which mixed case breaks rather than fixing both. The mirror shape — parent injected in `mount()`, child as a typed public prop — works on `main` today and regressed with the reorder (the child prop resolved before the parent was seeded, so it fell back to an unscoped lookup):

```php
public Book $book;

public function mount(Store $store) { ... }
```

Since the parent can live in either group, no ordering of the two phases fixes both directions. So I pushed a commit to your branch that instead does what Laravel itself does: a pre-pass that resolves every bindable route parameter in URI order (union of prop types + mount signature types) and seeds each back into the route via `setParameter()`. Both existing phases then just pick up the already-resolved models — and both skip already-resolved values, so nothing double-resolves. Your three tests still pass, plus a new test covering the mirror case (verified it fails without the pre-pass). Full binding suite is green locally.

mateusjunges:
Thanks @calebporzio!

eleftrik:
I have a full page Livewire component route which doesn't use scoped bindings:

```php
Route::get('/subscriptions/{customerType:slug}/{offer:slug}', ContractType::class)
    ->withoutScopedBindings()
    ->name('subscription.choose-contract-type');
```

After upgrading Livewire to `4.3.5`, code breaks:

`BadMethodCallException: Call to undefined method Domain\Customers\Models\CustomerType::offers()`

I was able to fix the problem by patching `vendor/livewire/livewire/src/Drawer/ImplicitRouteBinding.php`:

```diff
public function resolveAllParameters(Route $route, Component $component)
{
    // Resolve every bindable route parameter in URI order first (like Laravel's
    // own implicit binding) so that scoped child bindings always see their
    // parent as an already-resolved model, regardless of whether the parent
    // is a public property or a mount() parameter...

+    if ($route->enforcesScopedBindings()) {
        $this->resolveRouteParametersInOrder($route, $component);
+    }

    $params = $this->resolveMountParameters($route, $component);
    $props = $this->resolveComponentProps($route, $component);

    return $params->merge($props)->all();
}
```

@mateusjunges @joshhanley @calebporzio What do you think?

Thank you

PR #10342, opened 2026-06-09
[4.x] Fix `writeToHistory()` mutating the current history state

people-sea:

Yes. Related to filamentphp/filament#20017.

Yes.

No.

Yes.

The history coordinator was reusing `window.history.state` directly before calling `replaceState`.

  That means the current history state could be mutated before the browser history entry was actually replaced. If another library wraps `replaceState` and compares the incoming
  state with `window.history.state`, it can see the already-mutated object and skip the real `replaceState` call.

  This clones the current state before applying Livewire's updates, so `window.history.state` only changes when `replaceState` succeeds.

### DIFF
--- a/js/plugins/history/coordinator.js
+++ b/js/plugins/history/coordinator.js
@@ -53,7 +53,7 @@ class HistoryCoordinator {
     }
 
     writeToHistory(method, url, callback) {
-        let state = window.history.state || {}
+        let state = window.history.state ? { ...window.history.state } : {}
         if (!state.alpine) state.alpine = {}
 
         // Process the state using the callback...
--- a/js/plugins/history/coordinator.js
+++ b/js/plugins/history/coordinator.js
@@ -53,7 +53,8 @@ class HistoryCoordinator {
     }
 
     writeToHistory(method, url, callback) {
-        let state = window.history.state ? { ...window.history.state } : {}
+        let state = window.history.state || {}
+        // let state = window.history.state ? { ...window.history.state } : {}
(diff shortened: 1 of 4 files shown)


joshhanley (changes requested):
@people-sea thanks for the PR! Can you please add a browser test that covers this? Other than that, the rest looks good. Thanks!

people-sea:
Done, Thanks for reviewing!

joshhanley (approved):
@people-sea thanks for that! I ended up changing the browser test to more accurately reflect the reported Filament issue.

I believe the main issue here is that Filament wraps `window.history.replaceState`, which is exposing the problem in Livewire. So wanted to be clear about that in the test.

If I remember correctly, the wrapping was due to our history state bug, where Livewire would write to it for every property in the query string. That has been fixed for a little while now, so I do wonder whether Filament still needs the wrapper?

people-sea:
@joshhanley Thanks! You’re right. I looked into this further. Filament v5 no longer needs the wrapper, while Filament v4 still does because it uses Livewire 3. 

Thank you very much for also investigating this issue so thoroughly.

calebporzio:
Verified this locally — the new browser test fails on main exactly at the scroll-restore assertion and passes with the fix, and both new JS specs fail without the one-line change. Nice catch that `history.state` is a live cached object, and the error-handler spec update is a real improvement too (state is no longer mutated when the write throws). Ready to merge — thanks @people-sea, and thanks @joshhanley for tightening up the browser test.

PR #10345, opened 2026-06-12
Fix: Guess ability from method name when #[Authorize] receives only a class name

sawirricardo:
## Problem

When using `#[Authorize(Post::class)]` with only a class name (no explicit ability), authorization always fails. This is because `BaseAuthorize::call()` passes the FQCN directly as the ability to `Gate::authorize()`:

```
Gate::authorize("App\Models\Post")  // ❌ no such ability exists
```

However, using `$this->authorize(Post::class)` inside the method **works** because Laravel's `AuthorizesRequests::authorize()` guesses the ability from the calling method name via `debug_backtrace()`.

Reported in #10318.

## Solution

Mirror `AuthorizesRequests::parseAbilityAndArguments()` behavior in `BaseAuthorize::call()`:

- When `$ability` contains `\` (class name) and `$argument` is null, guess the ability from the component method name
- Apply Laravel's `resourceAbilityMap()` normalization (`store` → `create`, `edit` → `update`, etc.)
- Pass the class name as the argument to `Gate::authorize()`

```php
// Before: #[Authorize(Post::class)] on method createPost()
Gate::authorize("App\Models\Post")  // ❌ fails

// After:
Gate::authorize("createPost", ["App\Models\Post"])  // ✅ works
```

## Changes

- **`BaseAuthorize.php`** — Added ability-guessing logic when class name used as sole parameter. Added `normalizeGuessedAbilityName()` and `resourceAbilityMap()` mirroring Laravel's `AuthorizesRequests`.
- **`UnitTest.php`** — Added test covering both direct method-name matching (`createPost` → `createPost`) and resource map matching (`store` → `create`).

### DIFF
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -18,6 +18,17 @@ class BaseAuthorize extends LivewireAttribute
 
     public function call(array $parameters) : void
     {
+        // If ability is a class name (contains backslash) and no argument was
+        // provided, guess the ability from the component method name — matching
+        // the behavior of Laravel's AuthorizesRequests::authorize(Model::class).
+        if (is_null($this->argument) && is_string($this->ability) && str_contains($this->ability, '\\')) {
+            $ability = $this->normalizeGuessedAbilityName($this->getName());
+
+            Gate::authorize($ability, [$this->ability]);
+
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


joshhanley:
@sawirricardo thanks for the PR! There is already an open PR #10319 for this, so I'm going to close this one. Thanks!

sawirricardo:
no prob Josh!

PR #10346, opened 2026-06-12
[4.x] Use dedicated feature for `BaseRenderless`

ghabriel25:
`BaseRenderless` was placed inside mechanism like an orphan feature. So I think it would make more sense to use dedicated `SupportRenderless` feature for this just like `Async` attribute

## Before
```text
src/
├── Mechanisms/
│   ├── HandleComponents/
│   │   ├── BaseRenderless.php
```

## After
```text
src/
├── Features/
│   ├── SupportRenderless/
│   │   ├── BaseRenderless.php
```

### DIFF
--- a/src/Attributes/Renderless.php
+++ b/src/Attributes/Renderless.php
@@ -2,7 +2,7 @@
 
 namespace Livewire\Attributes;
 
-use Livewire\Mechanisms\HandleComponents\BaseRenderless;
+use Livewire\Features\SupportRenderless\BaseRenderless;
 
 #[\Attribute]
 class Renderless extends BaseRenderless
--- a/src/Features/SupportEvents/SupportEvents.php
+++ b/src/Features/SupportEvents/SupportEvents.php
@@ -9,7 +9,7 @@ use Livewire\Features\SupportAttributes\AttributeLevel;
 use Livewire\ComponentHook;
 use Livewire\Exceptions\EventHandlerDoesNotExist;
 use Livewire\Features\SupportAuthorization\BaseAuthorize;
-use Livewire\Mechanisms\HandleComponents\BaseRenderless;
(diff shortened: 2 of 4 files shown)


ghabriel25:
@joshhanley I think browser test related to this feature inside `SupportEvents` should be moved too, what do you think?

**Edit:** nevermind, I'll make dedicated browser test for this 👍

joshhanley (approved):
@ghabriel25 thanks for the PR! I'm not sure if we should merge this or not.

From a public API perspective, this should be fine, but it's technically a breaking change if someone is using the attribute directly instead of `Livewire/Attributes/Renderless`.

I've approved this, as if we do decide to go ahead, the code changes are good. Let's see what @calebporzio thinks. But I would lean towards not merging myself.

ghabriel25:
@joshhanley Thanks for the review!

This seems like a good place to consolidate browser tests for renderless actions, which are currently spread across islands, events, etc.

calebporzio:
I agree with the intent and content of this PR, but I think Josh might be right here? People could easily just import this class from the wrong namespace and get bit?

It would be good to have a firm policy on this kind of thing.

What do you think @ghabriel25 - do we freely change internal classes like this? that someone might in theory depend on? Or do we call it an internal class and do what we want. (leaning towards the latter and just merging this but want to make sure I'm not doing the wrong thing.)

ghabriel25:
@calebporzio @joshhanley I think we should officially treat all `Base*` classes (and anything under `Mechanisms\`,  as strictly internal. They are not part of the public API, and we should feel free to refactor/move them as the architecture improves — especially in version 4. 

As for users who accidentally depend on them: that's on them because there isnt single documentation about using it, and we can document it clearly in the upgrade guide / contributing docs.

calebporzio:
Agreed, thanks!

PR #10349, opened 2026-06-15
[4.x] Remove stale `$wire.$watch` cleanups from the component when manually unwatched

joshhanley:
This PR is an alternative to #10347. It preserves the original regression test and commit attribution, while moving the cleanup fix into `$wire.$watch` itself.

# The Scenario

When a file upload input is conditionally rendered inside a long-lived component, toggling the input off should release the input's event listeners and property watcher. Instead, each mount and removal leaves another component cleanup behind for the lifetime of the component, causing a memory leak.

```blade
<button wire:click="$toggle('showUpload')">Toggle</button>

@if ($showUpload)
@endif
```

# The Problem

`$wire.$watch` registers its Alpine watcher cleanup on the Livewire component. Component cleanups are only removed when the whole component is destroyed.

File uploads use `$wire.$watch` from an element-scoped directive:

```js
let unwatchValueReset = component.$wire.$watch(property, (value) => {
    if (! el.isConnected) return

    if (value === null || value === '') {
        el.value = ''
    }

    if (el.multiple && Array.isArray(value) && value.length === 0) {
        el.value = ''
    }
})

cleanup(() => {
    unwatchValueReset()
})
```

When the file input is removed, the directive can call the returned unwatch function, but the original cleanup remains in `component.cleanups`. That lingering cleanup still points at the file input through the watcher callback. If the input came from a larger conditional branch, the browser can keep that detached branch in memory for the rest of the component lifetime.

# The Solution

The solution is to make component cleanups removable so that a feature can dispose of them before the component itself is destroyed.

`addCleanup()` now returns a remover for the cleanup it just registered:

```js
addCleanup(cleanup) {
    this.cleanups.push(cleanup)

    return () => {
        let index = this.cleanups.indexOf(cleanup)

        if (index === -1) return

        this.cleanups.splice(index, 1)
    }
}
```

`$wire.$watch` uses that remover in its returned function, so that calling it removes the component cleanup and then calls Alpine's `unwatch`:

```js
let unwatch = Alpine.watch(getter, callback)

let removeCleanup = component.addCleanup(unwatch)

return () => {
    removeCleanup()
    unwatch()
}
```

The original PR also cleaned up the `livewire-upload-cancel` listener, which had previously been missed, so this PR keeps that change too.


calebporzio (approved):
Verified locally: JS suite passes (vitest, 100 tests), the new Dusk test passes on this branch, and the full SupportFileUploads browser file passes. Negative control: rebuilt the JS bundle from main and ran the new test against it - cleanups grew from 1 to 3 across mount/teardown cycles, confirming the leak this fixes. Also checked that the only internal $wire.$watch caller is supportFileUploads (updated here) and no other addCleanup caller relies on its return value, so the API change is backward compatible.

PR #10352, opened 2026-06-17
[4.x] File preview original name

blazery:
**Background**
I didn't check if this was a requested fix. I ran into this issue and wanted to create a PR. Feel free to shoot it down if the previous behaviour was intended. If the code is not up to standard, I don't mind changing it. 

I was working on a Form with an overview step. The overview step showed recently temporarily uploaded files, and completed uploaded files in one overview. All of these items should be previewable and clickable. Our completed media files could be downloaded or previewed using a custom endpoint, and the others using Livewire's built-in endpoint. 
Files downloaded from the `livewire.preview-file` endpoint, did not have the original filename the user provided, but an encoded version in Livewire v3 and a hash name in v4. I would like the original filename to be returned as to not expose any of the internal file structure of the website.

**Info**
I want the original filename to be returned so as not to expose any of the internal file structure of the website.
I added tests to verify this works with S3 buckets and remote storage. It handles cache hits. Im not sure about performance.

### DIFF
--- a/src/Features/SupportFileUploads/FilePreviewController.php
+++ b/src/Features/SupportFileUploads/FilePreviewController.php
@@ -5,6 +5,7 @@ namespace Livewire\Features\SupportFileUploads;
 use Illuminate\Routing\Controllers\HasMiddleware;
 use Illuminate\Routing\Controllers\Middleware;
 use Livewire\Drawer\Utils;
+use Symfony\Component\HttpFoundation\HeaderUtils;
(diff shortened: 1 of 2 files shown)


joshhanley:
@blazery thanks for the PR! I'm going to close this as you have also submitted a PR #10354 targeting v3. So we will review that one, and if it gets merged, it will then be up-merged into v4. Thanks!

PR #10260, opened 2026-05-02
[4.x] Fix Authorize attribute doesnt support `array` for additional context

ghabriel25:
Support `array` as argument for additional context. Laravel version 10+ already support this.
 https://laravel.com/docs/13.x/authorization#supplying-additional-context

See https://github.com/livewire/livewire/issues/10257 for the explanation.

Related PR: https://github.com/livewire/livewire/pull/10089

Fix: https://github.com/livewire/livewire/issues/10257

PR #10361, opened 2026-06-23
[4.x] Fix authorization using `#[Authorize]`  always fallback to component property

ghabriel25:
First of all, I want to apologize for the oversight on this PR https://github.com/livewire/livewire/pull/10260. I notice this bug from this comment https://github.com/livewire/livewire/pull/10292#pullrequestreview-4494752116

The authorization should prioritize action argument then fall back to component property if there was no argument found but even if the argument on action can be resolved, it always fall back to component property which lead authorization bypassed.

# Scenario

In Livewire component, user can pass a model `id` as argument to be resolved like route model binding in Laravel

This will pass on testing however, it bypassed the authorization if used on Livewire component.

```php
class PostPolicy
{
    public function edit(User $user, Post $post)
    {
        return (int) $post->id === 1
    }
}

new class extends Component
{
    public ?Post $post;

    public function mount()
    {
        $this->post = Post::find(1);
    }

    #[Authorize['edit', 'post'] // this SHOULD be denied, however it BYPASSED
    public function editPost(Post $post)
    {
        // authorization bypassed
    }
}
```
```blade
<button wire:click="editPost('2')">Edit</button>
```

# Problem

In `BaseAuthorize::resolveArgument()`, it only check if provided argument is set in action parameters. This will pass on unit test because all `call()` action using named argument,

```php
Livewire::actingAs(AuthorizationUser::find(1))
    ->test(new class extends TestComponent {
        #[Authorize('edit', 'post')]
        public function editPost(AuthorizationPost $post) : bool
        {
            return true;
        }
    })
    ->call('editPost', post: 1) // here
    ->assertOk();
```

However this will make authorization bypassed if user only provide an `id` to be resolved.

```php
if ($methodArgument instanceof \ReflectionParameter && isset($parameters[$arg)) { // ❌ 
    $model = app($methodArgument->getType()->getName())->resolveRouteBinding($parameters[$arg]); // ❌ 

    if (! $model) {
        throw (new \Illuminate\Database\Eloquent\ModelNotFoundException)->setModel($methodArgument->getType()->getName());
    }

    return $model;
}

// Fall back to component property
return data_get($this->component, $arg);
```
Using the scenario, this will ALWAYS fall back to component property.

# Solution

Check action parameters has either named argument or model `id`.
```php
if ($methodArgument instanceof \ReflectionParameter) {
    $routeBinding = null;

    // parameter keys can be a string if user pass named argument to be resolved
    // e.g wire:click="$dispatch('edit-post', { post: '2' })"
    if (isset($parameters[$arg])) {
        $routeBinding = $parameters[$arg];
    }

    // parameter keys can be a numeric if user only pass an id to be resolved
    // e.g wire:click="editPost('2')"
    elseif (isset($parameters[$methodArgument->getPosition()])) {
        $routeBinding = $parameters[$methodArgument->getPosition()];
    }

    $model = app($methodArgument->getType()->getName())->resolveRouteBinding($routeBinding);

    if (! $model) {
        throw (new \Illuminate\Database\Eloquent\ModelNotFoundException)->setModel($methodArgument->getType()->getName());
    }

    return $model;
}
```

# Test Case
```php
class PostPolicy
{
    public function edit(User $user, Post $post)
    {
        return (int) $post->id === 1
    }
}

new class extends Component
{
    public ?Post $post;

    public function mount()
    {
        $this->post = Post::find(1);
    }

    #[On('edit-post')]
    #[Authorize['edit', 'post']
    public function editPost(Post $post)
    {
        //
    }
}
```
```blade
<button wire:click="editPost('1')">Edit</button> // ✔️ 
<button wire:click="editPost({ post: '1' })">Edit</button> // ✔️ 
<button wire:click="$dispatch('edit-post', { post: '1'})">Edit</button> // ✔️ 

<button wire:click="editPost('2')">Edit</button> // ❌ 
<button wire:click="editPost({ post: '2' })">Edit</button> // ❌ 
<button wire:click="$dispatch('edit-post', { post: '2'})">Edit</button> // ❌ 
```

# Files Changed
- `src/Features/SupportAuthorization/UnitTest.php`

### DIFF
--- a/src/Features/SupportAuthorization/BaseAuthorize.php
+++ b/src/Features/SupportAuthorization/BaseAuthorize.php
@@ -57,9 +57,21 @@ class BaseAuthorize extends LivewireAttribute
         );
 
         if ($methodArgument instanceof \ReflectionParameter) {
-            $routeBindingValue = $this->getRouteBindingValue($arg, $methodArgument->getPosition(), $parameters);
+            $routeBinding = null;
 
-            $model = app($methodArgument->getType()->getName())->resolveRouteBinding($routeBindingValue);
+            // parameter keys can be a string if user pass named argument to be resolved
+            // e.g wire:click="$dispatch('edit-post', { post: '2' })"
+            if (isset($parameters[$arg])) {
+                $routeBinding = $parameters[$arg];
+            }
+
+            // parameter keys can be a numeric if user only pass an id to be resolved
+            // e.g wire:click="editPost('2')"
+            elseif (isset($parameters[$methodArgument->getPosition()])) {
+                $routeBinding = $parameters[$methodArgument->getPosition()];
+            }
+
+            $model = app($methodArgument->getType()->getName())->resolveRouteBinding($routeBinding);
 
             if (! $model) {
                 throw (new \Illuminate\Database\Eloquent\ModelNotFoundException)->setModel($methodArgument->getType()->getName());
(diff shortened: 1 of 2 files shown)


ghabriel25:
@calebporzio @joshhanley Please review this, thanks!

PR #10363, opened 2026-06-23
Add renderless wire model

sohailahmad07:
### The Scenario

Livewire already supports renderless actions with directives like `wire:click.renderless`. This lets an action run on the server without re-rendering the component afterward.

This PR adds the same renderless behavior to live model updates:

```blade
<input type="text" wire:model.renderless.live="title">
```

With this syntax, Livewire sends the updated model value to the server, but skips the component render for that model request.

This is useful for model updates that need to be persisted or processed on the server without morphing the current DOM, such as updating validation state that can be accessed from JavaScript through $errors.

### The Feature

`wire:model.renderless.live` now works as a Livewire-only rendering modifier for live model updates.

Because `wire:model` internally creates an Alpine `x-model` binding, Livewire removes `renderless` before building the Alpine binding. This keeps `renderless` scoped to Livewire and avoids passing it to Alpine as an `x-model` modifier.

```js
modifiers = modifiers.filter(m => m !== 'renderless')
```

The original directive modifiers remain available on `directive.modifiers`, so the request layer can still detect `.renderless` and mark the outgoing model request as renderless.

### Backend Behavior

Live model updates are sent as `$commit` magic actions. This PR adds scoped backend handling for renderless model commits:

```php
($metadata['renderless'] ?? false)
&& ($metadata['type'] ?? null) === 'model.live'
&& $method === '$commit'
```

This ensures only live model commits marked as renderless skip rendering, and prevents other magic actions from accidentally inheriting renderless behavior.

### Test Coverage

A browser regression test was added for:

```blade
wire:model.renderless.live="title"
```

The test verifies that the model value is sent to the server while the component does not re-render until a later normal commit.

### DIFF
--- a/docs/wire-model.md
+++ b/docs/wire-model.md
@@ -304,6 +304,18 @@ Bracket notation is also supported as an alternative:
 
 Both notations are equivalent — `address['city']` resolves the same path as `address.city`. Bracket and dot notation can be mixed freely within a single expression.
 
+## Skipping renders
+
+The `.renderless` modifier sends a live model update to the server without re-rendering the component:
+
+```blade
+<input type="text" wire:model.renderless.live="search">
+```
+
+This is useful for model updates that should be persisted on the server but should not morph the current DOM after the message completes.
+
+Because `.renderless` only affects server messages, use it with `.live`.
+
 ## Going deeper
 
 For a more complete documentation on using `wire:model` in the context of HTML forms, visit the [Livewire forms documentation page](/docs/4.x/forms).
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks @sohailahmad07 — nice find. The root issue is that magic actions (`$commit`, `$set`, etc.) take the early-return path in `callMethods()` and never reach the existing `.renderless` metadata check further down, so the modifier was silently ignored for all of them.

I pushed a couple of tweaks directly to your branch:

- Dropped the `model.live` type restriction so the early-return branch mirrors the generic renderless check — this also fixes things like `wire:click.renderless="$set('title', 'hello')"`, which had the same gap.
- Added a browser test covering the magic-action case alongside yours.

Both browser tests pass locally.

PR #10366, opened 2026-06-24
[4.x] Fix #[Validate()] and rules() rules for shared fields being incorrectly merged

### LINKED ISSUE #10364: #[Validate()] and rules() arrays are incorrectly merged
### Livewire version

v3.8.1 but also v4 is affected

### Laravel version

v12.62.0 but doesn't matter

### PHP version

8.2.30 but doesn't matter

Firefox on Windows; Apache running on linux

The `#[Validate()]` and `rules()` arrays aren't correctly merged. If you have your rules in `#[Validate()]`, the rules in `rules()` are just overwritten.

### Component:
```php
<?php
namespace App\Livewire;

use Livewire\Component;
use Illuminate\Support\Str;

class Test extends Component
{
    #[Validate(['required', 'int', 'min:1'])]
    public int $choices = 1;
    #[Validate(['required', 'min:1'])]
    public array $options = [];

    {
        return [
            'choices' => 'max:' . count($this->options),
        ];
    }

    public function addOption()
    {
        $this->options[] = Str::random(10); # it's pretty irrelevant, what the content is
    }

    public function store()
    {
        dump($this->validate());

        $this->reset();
    }

    public function render()
    {
        return view('livewire.test');
    }
}
```
### Blade:
```blade
<div>
    <form wire:submit="store" class="flex flex-col gap-1">
        <input name="choices" wire:model="choices" type="number" min="1" step="1" />
        @error('choices')
        <p class='text-sm text-red-600 dark:text-red-400'>{{ $message }}</p>
        @enderror
        <button wire:click="addOption" type="button">add Option</button>
        @forelse($options as $option)
        <div class="bg-neutral">{{ $option }}</div>
        @empty
        <div class="bg-neutral">no options there</div>
        @endforelse
        @error('options')
        <p class='text-sm text-red-600 dark:text-red-400'>{{ $message }}</p>
        @enderror
        <button>save</button>
    </form>
</div>
```

_No response_

I want to have my rules right where my variables are declared (and their defaults) and only move the rules that I have to in the rules() array.

It could be easily done in [src/Features/SupportValidation/HandlesValidation.php:121](https://github.com/livewire/livewire/blob/4cdaed0eadd80871f578acf55651b06120f51a59/src/Features/SupportValidation/HandlesValidation.php#L121) with `array_merge_recursive()` instead of `array_merge()`.

lazerg:
# The Scenario

When a component property has rules defined both in a \`rules()\` method and via a \`#[Validate()]\` attribute, only the attribute rules survive.

```php
{
    #[Validate(['required', 'string'])]
    public string $name = '';

    {
        return ['name' => ['min:3']];
    }
}
```

In this example, calling \`getRules()\` returns only \`['name' => ['required', 'string']]\` — the \`min:3\` rule from \`rules()\` is lost entirely.

# The Problem

\`getRules()\` ends with:

```php
return array_merge($rulesFromComponent, $rulesFromOutside);
```

Because \`array_merge\` overwrites string keys, any field appearing in both sources loses its \`rules()\` rules and keeps only the attribute rules.

# The Solution

Change \`array_merge\` to \`array_merge_recursive\` so that per-field rules from both sources are combined into a single array rather than the later source overwriting the earlier one.

Fixes #10364

### DIFF
--- a/src/Features/SupportValidation/HandlesValidation.php
+++ b/src/Features/SupportValidation/HandlesValidation.php
@@ -118,7 +118,7 @@ trait HandlesValidation
             )
         );
 
-        return array_merge($rulesFromComponent, $rulesFromOutside);
+        return array_merge_recursive($rulesFromComponent, $rulesFromOutside);
     }
 
     protected function getMessages()
--- a/src/Features/SupportValidation/UnitTest.php
+++ b/src/Features/SupportValidation/UnitTest.php
@@ -925,6 +925,29 @@ class UnitTest extends \Tests\TestCase
             ->assertSee('second error')
             ->assertHasErrors(['first', 'second']);
     }
+
+    public function test_validate_attribute_and_rules_method_rules_are_merged_for_the_same_property()
+    {
+        Livewire::test(new class extends TestComponent {
+            #[Validate(['required', 'string'])]
+            public $foo = '';
+
+            public function rules()
+            {
+                return [
+                    'foo' => ['min:3'],
+                ];
+            }
+
+            function save() { $this->validate(); }
+        })
+            ->set('foo', 'ok')
+            ->call('save')
+            ->assertHasErrors(['foo' => 'min'])
+            ->set('foo', '')
+            ->call('save')
+            ->assertHasErrors(['foo' => 'required']);
+    }
 }
 
 class ComponentWithRulesProperty extends TestComponent


joshhanley:
@lazerg thanks for the PR! I don't think this is a good idea as some of Laravel's validation methods are order dependent. If we add support for this, we have no way to determine the correct order. As such I'm going to close this. Thanks for being willing to contribute!

PR #10368, opened 2026-06-25
[4.x] Add support for `@continue` and `@break` in morph markers

### LINKED ISSUE #10350: wire:sort with wire:sort:group: TypeError on cross-group drag in empty group
### Livewire version

v4.2.4

### Laravel version

v12.56

### PHP version

8.4.16

Chromium Win

When using wire:sort with wire:sort:group to drag items between groups, a JavaScript error occurs in livewire.js when dropping an item into a different empty group:

TypeError: Cannot read properties of null (reading 'before')
    at Block.appendChild (livewire.js:13659:23)
    at context.patchChildren (livewire.js:13490:20)
    ...

This error breaks further Livewire execution on the page. The server-side callback runs correctly and the database is updated as expected — the issue is purely in the client-side DOM morphing after the update.

Root cause
wire:sort internally calls skipRender() after the sort callback to avoid unnecessary re-renders when reordering within the same group. However, when an item is dragged across groups (wire:sort:group), a full re-render is required because two separate DOM containers change. Livewire does not distinguish between these two cases and suppresses the re-render unconditionally, causing the DOM morphing algorithm to operate on a stale/inconsistent DOM state.

To reproduce
Set up a Livewire component with nested wire:sort and wire:sort:group
Have at least two groups with items
Drag an item from one group into an empty group
Observe the TypeError in the browser console

```<div wire:sort="reorderOfferOptions" wire:key="offer-options-{{ $offer->id }}">
            @forelse ($offerOptions as $option)

                <div wire:sort:item="{{ $option->id }}" wire:key='option_{{ $option->id }}' class="draggable">
                    {{ $option->name }} || {{ $option->id }}

                    <div class="mx-5 draggable-zone" wire:sort="reorderOfferedProducts" wire:sort:group="options"
                        wire:key="sort-zone-{{ $option->id }}" wire:sort:group-id="{{ $option->id }}">
                        @if ($option->offeredProducts->count() > 0)
                            @foreach ($option->offeredProducts as $offeredProduct)
                                <div wire:sort:item="{{ $offeredProduct->id }}"
                                    wire:key="op-{{ $offeredProduct->id }}" class="draggable border-bottom mb-2">
                                    {{ $offeredProduct->offer_option_order }} || {{ $offeredProduct->id }}
                                </div>
                            @endforeach
                        @endif
                    </div>
                </div>
            @empty
            @endforelse
        </div>
```

_No response_

The item is moved to the target group and the DOM updates correctly without any JavaScript errors.

joshhanley:
# The Scenario

When a Blade loop contains `@continue(...)`, a Livewire update can fail during DOM morphing with this browser console error:

```text
Uncaught TypeError: Cannot read properties of null (reading 'before')
```

The server-side action can still run, but the client-side update breaks when the DOM is patched.

```blade
@foreach ($offerOptions as $option)
    @continue($option->offeredProducts->count() == 0)

    <div wire:key="option-{{ $option->id }}">
        {{ $option->name }}
    </div>
@endforeach
```

# The Problem

The browser error happens because Livewire can send incomplete morph marker boundaries to the client.

Livewire did not support Blade `@continue` and `@break` directives when compiling morph markers. When a loop exits through one of those directives, Laravel compiles it into PHP control flow that can skip the closing marker Livewire expected to render. The client morph engine can then lose the matching end comment for the block it is patching and call `before` on a missing node.

# The Solution

Add support for Blade `@continue` and `@break` directives in Livewire morph marker compilation.

Bare directives are supported:

```blade
@continue
@break
```

Conditional directives are supported:

```blade
@continue($shouldSkip)
@break($shouldStop)
```

Numeric-depth directives are supported:

```blade
@continue(2)
@break(2)
```

Fixes #10350

### DIFF
--- a/src/Features/SupportMorphAwareBladeCompilation/BrowserTest.php
+++ b/src/Features/SupportMorphAwareBladeCompilation/BrowserTest.php
@@ -81,5 +81,39 @@ class BrowserTest extends \Tests\BrowserTestCase
         ->assertVisible('@disable')
         ;
     }
-}
 
+    public function test_continue_inside_loop_does_not_break_morphing_when_condition_changes(): void
+    {
+        Livewire::visit(new class extends Component {
+
+            public bool $show = false;
+
+            public function render(): string
+            {
+                return <<<'HTML'
+                <div>
+                    <button type="button" wire:click="showContent" dusk="show">Show</button>
+
+                    @foreach ([1] as $item)
+                        @if (true)
+                            @continue(! $show)
+
+                            <div dusk="content">Content</div>
+                        @endif
+                    @endforeach
+                </div>
+                HTML;
+            }
+
+            public function showContent(): void
+            {
+                $this->show = true;
+            }
+        })
+        ->assertNotPresent('@content')
+        ->waitForLivewire()->click('@show')
+        ->assertVisible('@content')
+        ->assertConsoleLogHasNoErrors()
+        ;
+    }
+}
--- a/src/Features/SupportMorphAwareBladeCompilation/SupportMorphAwareBladeCompilation.php
+++ b/src/Features/SupportMorphAwareBladeCompilation/SupportMorphAwareBladeCompilation.php
@@ -156,6 +156,8 @@ class SupportMorphAwareBladeCompilation extends ComponentHook
                 $template = static::suffixClosingDirective($match[0], $template, $matchPosition);
             } elseif (preg_match($loopEmptyDirectivePattern, $match[0])) {
                 $template = static::suffixLoopEmptyDirective($match[0], $template, $matchPosition);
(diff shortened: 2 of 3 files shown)


calebporzio (approved):
Verified locally: the SupportMorphAwareBladeCompilation unit file passes (136 tests) and the full Unit suite passes (1287 tests) on this branch, since this touches global template compilation. The new Dusk browser test passes here, and as a negative control it reproduces the exact reported error on main (Uncaught TypeError: Cannot read properties of null (reading 'before')). The 10 new provider cases plus the updated @while case all fail against main without the fix. Also sanity-checked the stray-ENDBLOCK cases (bare @break at loop-body depth, @break inside @switch): the Alpine morph algorithm only enters block scanning on a paired BLOCK-open, so an unmatched close marker is treated as an ordinary comment node and is benign.

PR #10369, opened 2026-06-25
[4.x] Fix fetch responses with a http redirect being parsed as Livewire payloads

### LINKED ISSUE #10365: Livewire continues processing redirected HTML response after handling fetch redirect
### What happened?

When a Livewire update request receives an HTTP redirect (for example because an auth/session middleware redirects to `/login`), the browser `fetch()` follows the redirect and Livewire receives the final HTML response.

Livewire detects `response.redirected === true` and invokes the redirect handler, which eventually calls `window.location.href = response.url`. However, the request flow continues processing the already-read response body after that point. Because the body is usually an HTML login page rather than a Livewire JSON payload, Livewire can still try to parse/process the redirected HTML before the browser completes the top-level navigation.

In practice this can cause a brief flash of redirected, unstyled HTML or an error fallback/modal before the navigation to the redirected URL completes.

### Reproduction

1. Have a Livewire component with any action that sends a normal Livewire update request.
2. Make the server respond to that update request with a normal HTTP redirect, such as `302 Location: /login` from an expired session/auth middleware.
3. Trigger the Livewire action after the session expires.
4. The update request is redirected to `/login`, so the final fetch response body is the login page HTML.
5. Livewire calls the redirect handler, but then continues to parse/process that HTML response as if it were the Livewire payload.

### Expected behavior

After Livewire handles a non-prevented `response.redirected` redirect, it should stop processing the response body for that request.

If a userland redirect hook calls `preventDefault()`, then continuing the current flow may still make sense. But when the default redirect behavior actually runs, the request should return early instead of attempting to parse the redirected HTML as JSON.

### Possible fix direction

Have the redirect handler report whether it actually performed the redirect, then return early in the request flow only for non-prevented redirects. Roughly:

```js
if (response.redirected) {
    let handled = handlers.redirect(response.url)

    if (handled) {
        handlers.finish()
        return
    }
}
```

and make the redirect handler return `false` when `preventDefault()` was called, otherwise call `window.location.href = url` and return `true`.

### Why this matters

Applications can work around this by returning JSON/401 for Livewire requests instead of `302`, but a regular Laravel auth/session redirect is a common failure mode. Once Livewire has already decided to perform a top-level redirect, continuing to process the redirected HTML body is surprising and can briefly leak an unstyled intermediate page to users.

joshhanley:
# The Scenario

When a Livewire update request receives a normal HTTP redirect, the browser follows that redirect internally and Livewire receives the final response body. If the redirect target is an HTML page such as a login page, Livewire can briefly show an error modal or log a JSON parse error before the browser finishes navigating to the redirected URL.

```php
use Livewire\Component;

new class extends Component
{
    public function save(): void
    {
        // Any Livewire action that sends an update request...
    }
};
```

```php
public function handle($request, Closure $next)
{
    if ($request->hasHeader('X-Livewire')) {
        return redirect('/login');
    }

    return $next($request);
}
```

# The Problem

Livewire detects `response.redirected` and calls the redirect handler, which sets `window.location.href` to the final fetch URL.

```js
if (response.redirected) {
    handlers.redirect(response.url)
}
```

The request flow then continues into dump detection and JSON parsing when it shouldn't.

```js
if (contentIsFromDump(responseBody)) {
    let dump

    [dump, responseBody] = splitDumpFromContent(responseBody)

    handlers.dump(dump)
}

let responseJson = JSON.parse(responseBody)
```

For normal HTTP redirects followed by `fetch()`, the response body is often the redirected HTML page rather than a Livewire JSON payload. That means Livewire can try to parse HTML as JSON even though it has already started a top-level redirect.

# The Solution

The solution is to treat a redirected fetch response as the end of the Livewire request once the existing redirect handler has run.

Livewire already detects `response.redirected` and calls `handlers.redirect(response.url)`. This change finishes the request and returns immediately after that point, so the redirected HTML body is not passed through dump detection, JSON parsing, or the error modal fallback.

```js
if (response.redirected) {
    handlers.redirect(response.url)
    handlers.finish()

    return
}
```

Fixes #10365

### DIFF
--- a/js/request/index.js
+++ b/js/request/index.js
@@ -527,6 +527,9 @@ async function sendRequest(request, handlers) {
 
     if (response.redirected) {
         handlers.redirect(response.url)
+        handlers.finish()
+
+        return
     }
 
     /**
--- a/src/Features/SupportRedirects/BrowserTest.php
+++ b/src/Features/SupportRedirects/BrowserTest.php
@@ -2,15 +2,38 @@
 
 namespace Livewire\Features\SupportRedirects;
 
-use Livewire\Attributes\On;
-use Tests\BrowserTestCase;
-use Livewire\Livewire;
-use Livewire\Component;
-use Livewire\Attributes\Reactive;
+use Closure;
+use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Route;
+use Livewire\Attributes\On;
+use Livewire\Attributes\Reactive;
+use Livewire\Component;
+use Livewire\Livewire;
+use Symfony\Component\HttpFoundation\Response;
+use Tests\BrowserTestCase;
 
 class BrowserTest extends BrowserTestCase
 {
+    public static function tweakApplicationHook()
+    {
+        return function () {
+            Livewire::addPersistentMiddleware(RedirectLivewireUpdateToHtml::class);
+
+            Route::get('/redirected-livewire-update', RedirectedLivewireUpdatePage::class)
+                ->middleware(['web', RedirectLivewireUpdateToHtml::class]);
+
+            Route::get('/redirected-livewire-update/login', fn () => response(<<<'HTML'
+                <!DOCTYPE html>
+                <html>
+                    <body>
+                        <h1 dusk="redirected-html">Redirected HTML response</h1>
+                    </body>
+                </html>
+                HTML))
+                ->middleware('web');
+        };
+    }
+
     public function test_can_redirect()
     {
         Livewire::visit([new class extends Component {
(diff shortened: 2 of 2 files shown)


calebporzio:
Traced this — since `handlers.redirect()` hard-navigates via `window.location.href`, there was never a reason to keep processing the body after it fires; the old fall-through just tried to `JSON.parse` the redirected HTML (login page etc.) and spewed a console error mid-navigation. Early return + `finish()` matches the other terminal paths in `sendRequest`. Test middleware setup is a nice realistic repro of an auth redirect hitting an update request. Ready to merge.

One pre-existing quirk this made me notice (not this PR's problem): `response.redirected` is also true for infra-level rewrites of the update endpoint itself (trailing-slash 301, http→https), and in that case we navigate the browser to the update URL rather than processing the payload — that was true before this change too. Probably fine, just noting it in case a report ever comes in.

PR #10370, opened 2026-06-25
[3.x] Fix wire:model.blur recreating deleted array entries

people-sea:

Yes. This fixes a regression in `wire:model.blur` where a stale model update can recreate an array item that was removed by the submit action.

 Yes

No
 Yes.

Fixes a regression where `wire:model.blur` could recreate a deleted array item after submit.

  `blur` is already handled by Livewire, so it should not be forwarded to Alpine's `x-model` modifiers.


calebporzio:
thanks!

PR #10374, opened 2026-06-29
[4.x] Add `Middleware` attribute

ghabriel25:
This PR introduces a new `#[Middleware]` attribute like Laravel [middleware attribute](https://laravel.com/docs/13.x/controllers#middleware-attributes) allowing users to apply middleware directly to individual component actions/methods only.

**Important Note**
- Only for component actions/methods
- Only support string as an argument (either full class name or middleware alias)
- `Closure` as argument **not supported** yet (Require php8.5)
- Middleware still need to be registered as `PersistentMiddleware` (See: [Configuring persistent middleware](https://livewire.laravel.com/docs/4.x/security#configuring-persistent-middleware))

To be honest, I really want to make this work as `PersistentMiddleware` but component hook runs after mechanism and I couldnt find a way to implement it as applicable middleware.

Therefore I use the same flow to apply the middleware into pipeline using `Utils::applyMiddleware()`. 

The first obstacle was middleware can return `RedirectResponse` which will throw an error if handled by Livewire. So, inside attribute component hook I restore Laravel redirector back into container to handle it as general redirect (not Livewire navigate).

# Examples

### Basic usage

```php
#[Middleware('auth')]
#[Middleware(EnsureEmailsVerified::class]
public function showContent()
{
    //
}

#[Middleware('guest')]
{
    //
}
```

### Authorization as middleware using component property

```php
public Post $post;

public function editPost()
{
    //
}
```

### Authorization as middleware using method parameter

```php
public function editPost(Post $post)
{
    //
}
```

### Trigger middleware with an event

```php
#[On('mark-as-read')]
#[Middleware('auth')]
public function markAsRead()
{
    //
}
```


joshhanley:
@ghabriel25 can you please mark this as draft until you are finished working on it? Thanks!

ghabriel25:
Hi @joshhanley Sorry for that. I just finish my last check. Its ready for review now, thanks!

src/Features/SupportActionMiddleware/SupportActionMiddleware.php
```diff
@@ -0,0 +1,50 @@
...
+
+    protected static function restoreOriginalRedirector()
+    {
+        $redirectorCacheStack = SupportRedirects::$redirectorCacheStack;
+
+        if (empty($redirectorCacheStack)) return;
+
+        $lastIndex = array_key_last($redirectorCacheStack);
+
+        $cachedRedirector = SupportRedirects::$redirectorCacheStack[$lastIndex];
```
ghabriel25: Should be

```php
$cachedRedirector = $redirectorCacheStack[$lastIndex];
```

calebporzio (commented):
Hey @ghabriel25 — I appreciate the ambition here, and the writeup is honest about the constraints you hit, which I respect. I checked out the branch, ran your new `SupportActionMiddleware` tests (all 8 pass) and the full unit suite (green). But after kicking the tires I don't think we should take this in its current shape — and honestly, the workarounds it needed are the tell.

Two concrete problems I verified locally:

**1. Non-persistent middleware silently no-ops — a security footgun.** `filterMiddlewareByPersistentMiddleware()` drops anything not registered as persistent middleware, without any warning. I added this test to your branch locally and it *passes*:

```php
// BlockEverything aborts(403) unconditionally, but isn't persistent middleware...
    #[Middleware(BlockEverything::class)]
    public function protectedAction() { Session::put('should-have-been-blocked', true); }
})
->call('protectedAction')
->assertOk(); // passes — the "protection" was silently skipped
```

A user who writes `#[Middleware('verified')]` or a custom middleware class will believe the action is protected when it isn't. Yes, it's documented in the PR body, but an attribute named `Middleware` that silently ignores most middleware is a trap we can't ship. (Your own `test_can_apply_multiple_middleware_to_action` hits this: `MiddlewareTest::class` gets filtered out, and the assertions after `expectException` never execute.)

**2. The redirector swap breaks Livewire redirects for any action carrying the attribute.** Because `SupportActionMiddleware` restores Laravel's redirector globally whenever the called method has a `#[Middleware]` attribute, `redirect()` inside that action stops going through Livewire. Verified:

```php
#[Middleware('auth')]
public function goSomewhere() { return redirect('/somewhere'); }
// ->call('goSomewhere')->assertRedirect('/somewhere') FAILS
// remove the attribute and it passes
```

Stepping back on direction: the fact that this needed persistent-middleware filtering, a global redirector swap, and special-casing `can:` into `BaseAuthorize` (duplicating what `#[Authorize]` already does) tells me HTTP middleware just doesn't map cleanly onto component actions — middleware wants a request/response pipeline, and actions aren't that. For the main real use cases we already have `#[Authorize]`, and I'd rather grow that family (dedicated action-level attributes) than expose the middleware stack here.

So I'm going to pass on this one as a feature — genuinely not a knock on the effort, the tests are thorough and the code is easy to follow. If you want to keep pulling on the thread, an issue/discussion proposing the API first would be the right next step before more implementation work.

One small aside that IS worth keeping: your `SupportEvents` addition uses `is_subclass_of($i, BaseMiddleware::class)`, which has the exact base-class-excluded quirk you fixed in #10384 — worth remembering if any of this gets revived.

ghabriel25:
Hi @calebporzio Thanks for the review. Here are some thoughts from me for your inputs

1. I can remove the filtering from persistent middleware but still remove all closure middleware. This way, the middleware still works even without being registered as persistent middleware. I also can filter all authorization based middleware if you prefer only general middleware should be applied for this attribute

2. I can work for this to find another solution so it doesnt break Livewire redirect feature.

Otherwise, feel free to close this. Thanks!

ghabriel25:
@calebporzio I think you're right about this approach. It seems not reliable applying middleware after mechanism already run and I really appreciate your inputs.

Therefore I'll close this one and re-open https://github.com/livewire/livewire/pull/10382 which apply the middleware to the route level based on action.

PR #10375, opened 2026-06-30
[4.x] Fix form `validateOnly()` keeping previous field messages

### LINKED ISSUE #10372: Duplicate validation errors when using Form classes
### Livewire version

v4.3.3

### Laravel version

v13.17.0

### PHP version

8.5.7

Safari on macOS

When validating properties within `Livewire\Form` classes, real-time validation appears to retain validation messages from the previous failed validation attempt.

```php
<?php

namespace App\Livewire;

use Livewire\Component;

{
    public ExampleForm $form;
}
```

```php
<?php

namespace App\Livewire\Forms;

use Livewire\Form;

{
    #[Validate('required|int')]
    public $number = '';
}
```

```blade
<div>
    <input type="text" wire:model.live="form.number" />

    @dump($errors->get('form.number'))
</div>
```

Image

Image

Image

The error bag should only contain validation error messages from the current request.

billyonecan:
Fixes #10372

## Summary

This fixes an issue where validation errors from the previous Livewire request are merged into the current request. As a result, form object `validateOnly()` can display duplicate or stale validation messages for the field being validated.

The fix mirrors what `Livewire\Features\SupportValidation\HandlesValidation` does by excluding the current field when merging messages from the component's error bag.

## Testing

A regression test has been added to cover this issue. The new test fails before this change with the following assertion:

```text
1) Livewire\Features\SupportFormObjects\UnitTest::test_validate_only_does_not_merge_stale_validation_errors
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => 'The title field is required.'
+    1 => 'The title field is required.'
)
```

### DIFF
--- a/src/Features/SupportFormObjects/Form.php
+++ b/src/Features/SupportFormObjects/Form.php
@@ -47,7 +47,7 @@ class Form implements Arrayable
             return $this->parentValidateOnly($field, $rules, $messages, $attributes, $dataOverrides);
         } catch (ValidationException $e) {
             invade($e->validator)->messages = $this->prefixErrorBag(invade($e->validator)->messages)->merge(
-                $this->getComponent()->getErrorBag()
+                $this->getComponent()->errorBagExcept($this->getPropertyName().'.'.$field)
             );
 
             invade($e->validator)->failedRules = $this->prefixArray(invade($e->validator)->failedRules);
--- a/src/Features/SupportFormObjects/Form.php
+++ b/src/Features/SupportFormObjects/Form.php
@@ -47,7 +47,7 @@ class Form implements Arrayable
             return $this->parentValidateOnly($field, $rules, $messages, $attributes, $dataOverrides);
         } catch (ValidationException $e) {
             invade($e->validator)->messages = $this->prefixErrorBag(invade($e->validator)->messages)->merge(
-                $this->getComponent()->errorBagExcept($this->getPropertyName().'.'.$field)
+                $this->getComponent()->errorBagExcept($this->getPropertyName() . '.' . $field)
             );
 
             invade($e->validator)->failedRules = $this->prefixArray(invade($e->validator)->failedRules);
--- a/src/Features/SupportFormObjects/UnitTest.php
+++ b/src/Features/SupportFormObjects/UnitTest.php
@@ -297,17 +297,22 @@ class UnitTest extends \Tests\TestCase
         ;
     }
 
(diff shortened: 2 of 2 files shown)


joshhanley (approved):
@billyonecan thanks for the PR! All looks good to me. I did update the test to more closely reflect the issue you were experiencing, with `required` errors still being shown after it passed and should be showing `integer` errors.

---

# The Scenario

When a form object property is validated as it is updated, the error bag should only contain messages from the current validation pass for that property.

If a field first fails `required` validation and then later fails `integer` validation, the rendered error messages can contain both the current and previous failures:

```php
[
    'The number field must be an integer.',
    'The number field is required.',
]
```

The expected error bag should only include the current validation failure:

```php
[
    'The number field must be an integer.',
]
```

Image

```php
<?php

namespace App\Livewire\Forms;

use Livewire\Form;

{
    #[Validate(['required', 'integer'])]
    public $number = 1;
}
```

```blade
<?php

use Livewire\Component;

new class extends Component {
    public ExampleForm $form;
}; ?>

<div>
    <input type="text" wire:model.live="form.number" />

    @dump($errors->get('form.number'))
</div>
```

# The Problem

`Livewire\Features\SupportFormObjects\Form::validateOnly()` prefixes form object validation errors before they are stored on the parent component.

When validation fails, it was also merging the component's entire existing error bag back into the validator. That preserved unrelated errors, but it also preserved stale errors for the same form field currently being validated.

This meant a previous `form.number` error could be appended to the new `form.number` error, leaving duplicate or outdated validation messages in the rendered error bag.

# The Solution

The solution is to preserve only existing component errors that do not belong to the form field currently being validated.

Instead of merging the component's whole error bag, `Form::validateOnly()` now excludes the prefixed field key from the preserved errors. This keeps sibling field errors intact while allowing the current field's validation messages to be replaced by the latest validation result.

Image

calebporzio:
Traced this through — it mirrors exactly what `HandlesValidation::validateOnly()` already does at the component level with `errorBagExcept()`, and since that helper uses `str()->is()`, wildcard fields exclude correctly too while stale sibling-key errors are preserved. The test proves replacement rather than just presence by asserting the exact bag contents across two different failing rules. Ready to merge — thanks @billyonecan!

Heads-up: this and #10376 both touch the trailing line of `SupportFormObjects/UnitTest.php`, so whichever merges second needs a trivial rebase.

calebporzio:
makes sense, thanks!

PR #10376, opened 2026-06-30
[4.x] Handle exceptions from form lifecycle hooks

### LINKED ISSUE #10373: `BadMethodCallException` when validating within Form classes
### Livewire version

v4.3.3

### Laravel version

v13.17.0

### PHP version

8.5.7

Safari on macOS

When attempting to validate properties within a `Livewire\Form` class, a `BadMethodCallException` is thrown:

```
Method Livewire\Features\SupportRedirects\Redirector::withInput does not exist.
```

```php
<?php

namespace App\Livewire;

use Livewire\Component;

{
    public ExampleForm $form;
}
```

```php
<?php

namespace App\Livewire\Forms;

use Livewire\Form;

{
    public $number = '';

    public function updated(string $property, mixed $value): void
    {
        $this->validateOnly($property);
    }

    {
        return [
            'number' => 'required|int',
        ];
    }
}
```

```blade
<div>
    <input type="text" wire:model.live="form.number" />
</div>
```

_No response_

I'd expect it to work the same way as it would if it wasn't using a `Form` class.

billyonecan:
Fixes #10373

## Summary

This fixes an issue where exceptions thrown from `Livewire\Form` lifecycle hooks are not handled by Livewire's component exception pipeline.

Form lifecycle hooks are invoked via `wrap($form)`, which dispatches exceptions against the `Form` instance. Since component exception hooks are only registered against `Livewire\Component` instances, exceptions thrown from form hooks bypass Livewire's exception handling and instead propagate to Laravel's global exception handler.

For `ValidationException`, this results in Laravel attempting to generate a redirect response (`Redirector::withInput()`), causing a `BadMethodCallException` during a Livewire request.

**Note**: This implementation handles `Livewire\Form` explicitly within `Wrapped` so that exceptions are dispatched against the owning component. If there's a preferred abstraction for mapping wrapped targets to their exception target, I'm happy to update accordingly.

### DIFF
--- a/src/ComponentHookRegistry.php
+++ b/src/ComponentHookRegistry.php
@@ -82,7 +82,7 @@ class ComponentHookRegistry
         });
 
         on('exception', function ($target, $e, $stopPropagation) {
-            if ($target instanceof Form) {
+            if ($target instanceof \Livewire\Form) {
                 $target = $target->getComponent();
             }
 
--- a/src/ComponentHookRegistry.php
+++ b/src/ComponentHookRegistry.php
@@ -82,7 +82,7 @@ class ComponentHookRegistry
         });
 
         on('exception', function ($target, $e, $stopPropagation) {
-            if ($target instanceof \Livewire\Form) {
+            if ($target instanceof \Livewire\Features\SupportFormObjects\Form) {
                 $target = $target->getComponent();
             }
 
--- a/src/ComponentHookRegistry.php
+++ b/src/ComponentHookRegistry.php
@@ -82,6 +82,10 @@ class ComponentHookRegistry
         });
 
         on('exception', function ($target, $e, $stopPropagation) {
+            if ($target instanceof Form) {
+                $target = $target->getComponent();
+            }
+
             if ($target instanceof \Livewire\Component) {
                 static::proxyCallToHooks($target, 'callException')($e, $stopPropagation);
             }
--- a/src/Features/SupportFormObjects/UnitTest.php
+++ b/src/Features/SupportFormObjects/UnitTest.php
@@ -894,6 +894,17 @@ class UnitTest extends \Tests\TestCase
         ;
     }
 
+    function test_can_perform_validate_only_inside_form_updated_hook()
+    {
+        Livewire::test(new class extends TestComponent {
+            public FormWithValidateOnlyInUpdatedHookStub $form;
+        })
+        ->assertHasNoErrors()
+        ->set('form.number', 'abc')
+        ->assertHasErrors('form.number')
+        ;
+    }
+
     function test_can_reset_and_return_property_with_pull_method()
     {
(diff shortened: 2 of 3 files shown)


joshhanley (approved):
@billyonecan thanks for the PR!

I've made a small change to add the check inside the component hook registry, rather than in wrapped itself. Also added a test that reflects the issue you raised. Otherwise all looks good to me!

---

# The Scenario

When a `Livewire\Form` validates from its update lifecycle hook, Livewire should capture the validation failure and add it to the component error bag.

Instead, the validation exception escapes Livewire’s validation handling. In a browser request, Laravel then tries to convert the validation failure into a redirect response, which causes this error:

```text
Method Livewire\Features\SupportRedirects\Redirector::withInput does not exist.
```

Image

```php
<?php // resources/views/components/⚡example.blade.php

use Livewire\Component;

new class extends Component {
    public ExampleForm $form;
};
?>

<div>
    <input type="text" wire:model.live="form.number" />
</div>
```

```php
<?php

namespace App\Livewire\Forms;

use Livewire\Form;

{
    public $number = '';

    public function updated(string $property, mixed $value): void
    {
        $this->validateOnly($property);
    }

    {
        return [
            'number' => ['required', 'integer'],
        ];
    }
}
```

# The Problem

Validation exceptions are normally handled by Livewire’s component exception hooks, which catch the exception, add the validator errors to the component error bag, and stop the exception from propagating.

Form lifecycle hooks are invoked through `wrap($form)`, so exceptions thrown from those hooks are dispatched against the `Livewire\Form` instance. The component hook registry only forwards exception hooks for `Livewire\Component` instances, which means validation exceptions from form hooks never reach `SupportValidation`.

Once the exception escapes Livewire, Laravel handles it like a normal validation failure and attempts to redirect back with input. During a Livewire request, the redirector has been replaced with Livewire’s redirector, so Laravel ends up calling `withInput()` on the wrong redirector.

# The Solution

The solution is to add form detection to the component hook registry so exceptions from form objects are forwarded through their component.

This lets validation exceptions thrown from form lifecycle hooks flow through the existing component exception hooks.

Image

calebporzio:
Great find @billyonecan, and the registry listener is the right place for this — `Wrapped` is the only thing that fires the `exception` event, so mapping the form to its owning component there lets the whole existing pipeline (`SupportValidation`, user `exception()` hooks, testing support) just work.

One fix I pushed to your branch: the check was against `\Livewire\Form`, which is just the empty surface subclass — every other internal check (`FormObjectSynth::match()`, `SupportFormObjects::update()`, etc.) uses the base `Livewire\Features\SupportFormObjects\Form`, so a form extending the base class directly would have been recognized everywhere except here. Swapped it to the base class and added a regression test covering that case (verified it errors with the original check). Full `SupportFormObjects` suite passes locally.

PR #10378, opened 2026-07-01
[4.x] Refactor `SupportAttributes`

ghabriel25:
I notice that most of `SupportAttributes` lifecycle hook use the same pattern

```php
$this->component
    ->getAttributes()
    ->whereInstanceOf(LivewireAttribute::class)
    ->each(...)
```

It can be extracted to dedicate method and pass the callback

### Before
```php
function boot(...$params)
{
    $this->component
        ->getAttributes()
        ->whereInstanceOf(LivewireAttribute::class)
        ->each(function ($attribute) use ($params) {
            if (method_exists($attribute, 'boot')) {
                $attribute->boot(...$params);
            }
        });
}

function render(...$params)
{
    $callbacks = $this->component
        ->getAttributes()
        ->whereInstanceOf(LivewireAttribute::class)
        ->map(function ($attribute) use ($params) {
            if (method_exists($attribute, 'render')) {
                return $attribute->render(...$params);
            }
        });

    return function (...$params) use ($callbacks) {
        foreach ($callbacks as $callback) {
            if (is_callable($callback)) {
                $callback(...$params);
            }
        }
    };
}
```

### After
```php
function boot(...$params)
{
    $this->getLivewireAttributes()->each(function ($attribute) use ($params) {
        if (method_exists($attribute, 'boot')) {
            $attribute->boot(...$params);
        }
    });
}

function render(...$params)
{
    $callbacks = $this->getLivewireAttributes()
        ->map(function ($attribute) use ($params) {
            if (method_exists($attribute, 'render')) {
                return $attribute->render(...$params);
            }
        });

    return function (...$params) use ($callbacks) {
        foreach ($callbacks as $callback) {
            if (is_callable($callback)) {
                $callback(...$params);
            }
        }
    };
}

protected function getLivewireAttributes()
{
    return $this->component
        ->getAttributes()
        ->whereInstanceOf(LivewireAttribute::class);
}
```

This clean a bit the repetitive code.

### DIFF
--- a/src/Features/SupportAttributes/SupportAttributes.php
+++ b/src/Features/SupportAttributes/SupportAttributes.php
@@ -15,7 +15,7 @@ class SupportAttributes extends ComponentHook
             }
         };
 
-        $this->invokeLivewireAttribute($callback);
+        $this->getLivewireAttributes()->each($callback);
     }
 
     function mount(...$params)
@@ -26,7 +26,7 @@ class SupportAttributes extends ComponentHook
             }
         };
 
-        $this->invokeLivewireAttribute($callback);
+        $this->getLivewireAttributes()->each($callback);
     }
 
     function hydrate(...$params)
@@ -37,7 +37,7 @@ class SupportAttributes extends ComponentHook
             }
         };
 
-        $this->invokeLivewireAttribute($callback);
+        $this->getLivewireAttributes()->each($callback);
     }
 
     function update($propertyName, $fullPath, $newValue)
@@ -103,7 +103,7 @@ class SupportAttributes extends ComponentHook
             }
         };
 
-        $this->invokeLivewireAttribute($callback);
+        $this->getLivewireAttributes()->each($callback);
     }
 
     function destroy(...$params)
@@ -114,7 +114,7 @@ class SupportAttributes extends ComponentHook
             }
         };
 
-        $this->invokeLivewireAttribute($callback);
+        $this->getLivewireAttributes()->each($callback);
     }
 
     function exception(...$params)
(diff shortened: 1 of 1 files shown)


src/Features/SupportAttributes/SupportAttributes.php
```diff
@@ -9,45 +9,40 @@ class SupportAttributes extends ComponentHook
...
     {
-        $this->component
-            ->getAttributes()
-            ->whereInstanceOf(LivewireAttribute::class)
-            ->each(function ($attribute) use ($params) {
-                if (method_exists($attribute, 'boot')) {
-                    $attribute->boot(...$params);
-                }
-            });
+        $callback = function ($attribute) use ($params) {
```
joshhanley: I don't think this should be a temporary variable.

joshhanley (changes requested):
@ghabriel25 thanks for the PR! It makes sense, but I also don't think this is a pain point for maintainability, so up to Caleb whether he wants to merge it or not.

But I don't agree with the `$callback` temporary variables. I actually think it makes it harder to read. I would just inline them again.

src/Features/SupportAttributes/SupportAttributes.php
ghabriel25: Done.

calebporzio (commented):
Thanks @ghabriel25! Since Josh punted the call to me: I'm fine taking this. The extraction is legitimate — same filtered collection built eight times over — and the final revision (with Josh's inline feedback addressed) reads clean to me.

Because refactors of attribute handling are easy to get subtly wrong, I verified it's purely behavior-preserving: `getLivewireAttributes()` returns a fresh filtered collection each call just like the inlined version did (no caching or ordering change), and I ran the full unit suite (1295 tests) plus the SupportRenderless browser tests on this branch — all green.

Good with this one.

calebporzio:
thanks @ghabriel25 !

PR #10380, opened 2026-07-02
Fix PHP 8.5 null array offset deprecation in EloquentCollectionSynth::hydrate()

### LINKED ISSUE #10379: PHP 8.5 deprecation: `EloquentCollectionSynth::hydrate()` passes `null` to `Relation::getMorphedModel()`
### Livewire version

v4.3.3

### Laravel version

v13.18.0

### PHP version

8.5.7

Chrome on Linux and Windows

### Overview

After upgrading to PHP 8.5, Livewire triggers a deprecation warning on every subsequent request (update cycle) for components that hold an Eloquent Collection property:

```
ErrorException: Using null as an array offset is deprecated, use an empty string instead
in vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/Relation.php:511
```

The issue seems to come from `EloquentCollectionSynth::hydrate()` (line 66). When an Eloquent Collection is dehydrated with an empty collection (`getQueueableClass()` returns `null`), `$meta['modelClass']` is stored as `null`. During hydration, this `null` value is passed directly to `Relation::getMorphedModel($modelClass)`, which uses it as an array key (`static::$morphMap[$alias]`). In PHP 8.5, using `null` as an array offset is deprecated.

### Steps to reproduce:

1. Set up a new Laravel 13 / Livewire 4 app using PHP 8.5
2. Create the component described below and mount it
3. Navigate in your browser to the route where you mounted the component
4. Click the "Refresh" button to trigger a Livewire roundtrip
5. Observe the deprecation warning in logs

### Suggested fix

In `src/Features/SupportModels/EloquentCollectionSynth.php`, line 66:

```php
// Before:
$modelAlias = Relation::getMorphedModel($modelClass);

// After:
$modelAlias = $modelClass ? Relation::getMorphedModel($modelClass) : null;
```

This matches the guard already present in the `dehydrate()` method (line 34: `if ($modelClass) { ... }`).

```php
<?php

use App\Models\User;
use Livewire\Component;

class UserList extends Component
{
    public Collection $users;

    public function mount(): void
    {
        $this->users = new Collection();
    }

    public function refresh(): void
    {
        $this->users = User::all();
    }

    public function render()
    {
        return <<<'BLADE'
        <div>
            <button wire:click="refresh()">Refresh</button>
            @foreach($users as $user)
                <p>{{ $user->name }}</p>
            @endforeach
        </div>
        BLADE;
    }
}
```

_No response_

Following the steps to reproduce mentioned above should not lead to a deprecation warning in the log.

JulianGlueck:

This fixes a PHP 8.5 deprecation warning. No discussion needed as it's a clear bug fix.

Yes, the branch is named `fix/php85-null-array-offset-eloquent-collection-synth`

No

Yes

**Fixes #10379** 

## Problem

PHP 8.5 deprecated using `null` as an array offset.

In `EloquentCollectionSynth::hydrate()`, when an empty Eloquent Collection is dehydrated, `$meta['modelClass']` is stored as `null` (because `getQueueableClass()` returns `null` for empty collections). During hydration, this `null` is passed to `Relation::getMorphedModel($modelClass)`, which uses it as an array key (`static::$morphMap[$alias]`), triggering the deprecation.

## Fix

Guard against `null` before calling `getMorphedModel()`, matching the guard already present in `dehydrate()` (line 34).

### DIFF
--- a/src/Features/SupportModels/EloquentCollectionSynth.php
+++ b/src/Features/SupportModels/EloquentCollectionSynth.php
@@ -63,7 +63,7 @@ class EloquentCollectionSynth extends Synth {
         $modelClass = $meta['modelClass'];
 
         // If no alias found, this returns `null`
-        $modelAlias = Relation::getMorphedModel($modelClass);
+        $modelAlias = $modelClass ? Relation::getMorphedModel($modelClass) : null;
 
         if (! is_null($modelAlias)) {
             $modelClass = $modelAlias;
--- a/src/Features/SupportModels/UnitTest.php
+++ b/src/Features/SupportModels/UnitTest.php
@@ -325,25 +325,32 @@ class UnitTest extends \Tests\TestCase
         $this->assertCount(1, array_values($articleQueries));
     }
 
-    public function test_eloquent_collection_synth_handles_null_model_class_without_deprecation()
+    public function test_hydrating_an_empty_eloquent_collection_does_not_trigger_deprecations()
     {
-        $component = Livewire::test(ArticleComponent::class);
+        $component = Livewire::test(new class extends \Livewire\Component {
+            public Collection $articles;
 
-        $synth = new EloquentCollectionSynth(
-            new \Livewire\Mechanisms\HandleComponents\ComponentContext($component->instance()),
-            'articles'
-        );
+            public function mount() {
+                $this->articles = new Collection();
+            }
 
-        $meta = [
-            'class' => Collection::class,
-            'modelClass' => null,
-            'keys' => [],
-        ];
+            public function render() { return <<<'HTML'
+                <div>count: {{ count($articles) }}</div>
+            HTML; }
+        });
 
-        // Should not trigger "Using null as an array offset" deprecation
-        $result = $synth->hydrate(null, $meta, fn ($v) => $v);
+        // An empty collection dehydrates with `modelClass` as `null`. Convert
+        // deprecations to exceptions so the test fails if hydrating passes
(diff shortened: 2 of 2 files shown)


joshhanley (approved):
@JulianGlueck thanks for the PR!  Thanks for the fix, all looks good to me.

Looking at the `getMorphedModel()` method, this is a pattern used everywhere in Laravel and Livewire. What a dumb thing to deprecate 🤦‍♂️

```php
public static function getMorphedModel($alias)
{
    return static::$morphMap[$alias] ?? null;
}
```

calebporzio:
Thanks @JulianGlueck — the fix itself is exactly right, and `hydrate()` is the correct place for it since existing snapshots in the wild already carry `modelClass: null` (I also checked `ModelSynth` and the legacy `EloquentModelSynth` — they only ever receive string classes in real flows, so this single guard covers it).

One issue with the test though: it passed even without the fix. Laravel's error handler routes deprecations to the log and our phpunit config doesn't fail on them, so on the PHP 8.5 CI matrix it was green on current main too. I pushed a commit to your branch swapping it for a round-trip test that renders a component with an empty collection (which dehydrates with `modelClass: null`) and converts deprecations to exceptions during `$refresh` — same pattern we use in `SupportMultipleRootElementDetection`. That version will actually fail on PHP 8.5 without your guard.

JulianGlueck:
Thanks & good catch, @calebporzio, I missed that 😅

PR #10381, opened 2026-07-03
[4.x] Add original filename option to temporary upload URLs

joshhanley:
# The Scenario

When rendering a temporary upload preview link, Livewire's generated `temporaryUrl()` points at the `livewire.preview-file` endpoint. If that preview link is downloaded, the response currently suggests Livewire's temporary filename instead of the user's original upload name.

See screencapture in the solution below for the before and after.

```blade
<div>

    @if ($photo)
        <a href="{{ $photo->temporaryUrl() }}" target="_blank">
            Preview
        </a>
    @endif
</div>
```

# The Problem

`TemporaryUploadedFile::temporaryUrl()` currently generates a signed local preview URL with only the temporary filename.

`FilePreviewController` passes that filename to `Utils::pretendPreviewResponseIsPreviewFile()`, and `Utils` uses it as both the storage lookup name and the public download name for `Storage::download()`.

This means the local preview endpoint's `Content-Disposition` filename is tied to Livewire's internal temporary filename.

# The Solution

This is a follow up of PR #10354, which changed it, but that was a breaking change, so this is an alternative solution.

The solution is adding an opt-in `useOriginalFilename` parameter to `TemporaryUploadedFile::temporaryUrl()`.

```blade
<img src="{{ $photo->temporaryUrl(useOriginalFilename: true) }}">
```

When `useOriginalFilename` is false or not present, Livewire keeps the existing response filename behaviour. When true, Livewire includes a signed `useOriginalFilename` flag in the preview URL.

The preview controller verifies the signed URL before deriving the original upload name from `TemporaryUploadedFile`.

It then sets the `Content-Disposition` filename to the original upload name so the downloaded file has the expected name.

CleanShot 2026-07-03 at 10 44 21

Refs #10354

### DIFF
--- a/src/Drawer/Utils.php
+++ b/src/Drawer/Utils.php
@@ -65,7 +65,7 @@ class Utils extends BaseUtils
             fn ($headers) => response($content, 200, $headers));
     }
 
-    static function pretendPreviewResponseIsPreviewFile($filename)
+    static function pretendPreviewResponseIsPreviewFile($filename, $downloadName = null)
     {
         $file = FileUploadConfiguration::path($filename);
         $storage = FileUploadConfiguration::storage();
@@ -73,7 +73,7 @@ class Utils extends BaseUtils
         $lastModified = FileUploadConfiguration::lastModified($file);
 
         return self::cachedFileResponse($filename, $mimeType, $lastModified,
-            fn ($headers) => $storage->download($file, $filename, $headers));
+            fn ($headers) => $storage->download($file, $downloadName ?? $filename, $headers));
     }
 
     static private function cachedFileResponse($filename, $contentType, $lastModified, $downloadCallback)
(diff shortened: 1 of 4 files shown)


calebporzio (approved):
Verified locally: both new tests pass and the full SupportFileUploads unit file passes (59 tests). Negative control against main: the backward-compat test (temporary filename in Content-Disposition) still passes, confirming zero behavior change for existing users, and the opt-in test errors with unknown named parameter as expected. Also confirmed the useOriginalFilename flag is baked into the signed URL, so it can't be toggled onto an existing URL client-side.

calebporzio:
Thanks for working through this and for taking care to preserve backwards compatibility.

I did a broader demand pass before merging this. The two original PRs (#10352 and #10354) represent one affected user targeting two branches, and I could not find any independent reports, reactions, or matching discussions asking for this behavior.

Given that `temporaryUrl()` is a preview API, adding permanent download-filename configuration for this single workflow does not feel justified yet. Applications that need a downloadable temporary upload can provide an explicit download action or route and use `getClientOriginalName()` as the response filename.

I explored the compatible implementation further in #10615, but came away feeling that even the small opt-in adds more framework surface than the current demand warrants. I am going to close this for now. We can revisit a dedicated temporary-download API if more independent demand emerges.

Thanks again for the thoughtful implementation.

PR #10382, opened 2026-07-03

ghabriel25:
Another approach from https://github.com/livewire/livewire/pull/10374

This PR introduces a new `#[Middleware]` attribute like Laravel [middleware attribute](https://laravel.com/docs/13.x/controllers#middleware-attributes) allowing users to apply middleware directly to individual component actions/methods only.

**Important Note**
- Only for component actions/methods
- Only support string as an argument (either full class name or middleware alias)
- `Closure` as argument **not supported** yet

# Feature flow

Applying middleware from attribute to the route and append it as persistent middleware.

Inside `SupportActionMiddleware::gatherActionMiddleware($request)`, all component methods that using `#[Middleware]` attribute resolved using reflection including the one that also use `#[On]` attribute so method name can be resolved if its an event listener.

# Examples

### Basic usage

```php
#[Middleware('auth')]
#[Middleware(EnsureUserIsSubscribed::class)
public function showContent()
{
    //
}

#[Middleware('guest')]
{
    //
}
```

### Trigger middleware with an event

```php
#[On('mark-as-read')]
#[Middleware('auth')]
public function markAsRead()
{
    //
}
```

### DIFF
--- a/src/Attributes/Middleware.php
+++ b/src/Attributes/Middleware.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Attribute;
+use Livewire\Features\SupportActionMiddleware\BaseMiddleware;
+
+#[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_METHOD)]
+class Middleware extends BaseMiddleware
+{
+    //
+}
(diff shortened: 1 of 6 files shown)


ghabriel25:
@joshhanley @calebporzio  What do you think about this approach? is it okay rather than this PR https://github.com/livewire/livewire/pull/10374

I'll start working on the test if both of you approved this approach, thanks!

ghabriel25:
Turns out this one is risky therefore I'm gonna close this one and re-open the original PR (https://github.com/livewire/livewire/pull/10374)

I apologize for opening multiple pull requests. I realize that may have created extra review overhead

PR #10384, opened 2026-07-04
[4.x] Fix `BaseAuthorze` and `BaseRenderless` not triggered using event

### LINKED ISSUE #10383: BaseRenderless and BaseAuthorize attribute doesnt work inside browser test when action trigerred as event listener
### Livewire version

v4.3

### Laravel version

v13

### PHP version

8.4

N/A

When using `#[BaseRenderless]` or `#[BaseAuthorize]` inside browser test, triggering an action as event listener doesnt work.

Its because inside `SupportEvents::call()` filter the attribute as a subclass not instance
```php
$this->component->getAttributes()
    ->filter(fn ($i) => is_subclass_of($i, BaseAuthorize::class))
    ->filter(fn ($i) => $i->getName() === $method)
    ->filter(fn ($i) => $i->getLevel() === AttributeLevel::METHOD)
    ->each(fn ($i) => $i->call($params));
```

### Suggestion

```php
$this->component->getAttributes()
    ->filter(fn ($i) => $i instanceof BaseAuthorize::class)
    ->filter(fn ($i) => $i->getName() === $method)
    ->filter(fn ($i) => $i->getLevel() === AttributeLevel::METHOD)
    ->each(fn ($i) => $i->call($params));
```

```php
public functio test_can_trigger_renderless_attribute_with_event()
{
    Livewire::visit(new class extends Component {
    public $count = 0;

    #[BaseRenderless]
    #[BaseOn('some-event')
    function renderlessAction() { }

    function render()
    {
        $this->count++;

        return Blade::render(<<<'HTML'
        <div>
            <button wire:click="$dispatch('some-event')" dusk="button">{{ $count }}</button>
        </div>
        HTML, ['count' => $this->count]);
    }
    })
    ->assertSeeIn('@button', '1')
    ->waitForLivewire()->click('@button')
    ->assertSeeIn('@button', '1');
}
```

N/A

When user wants to contribute and creating BrowserTest, this could lead to unusual behavior. I also notice some of unit test also using `Base**` attribute.

ghabriel25:
# Scenario

When user wants to contribute to this repo and preparing some browser / unit test related to `#[Authorize]` and `#[Renderless]` attribute, they might be not aware that using `#[BaseAuthorize]` or `#[BaseRenderless]` wont be triggered if action called as event listener.

I'm aware that all `Base*` attribute should be strictly for internal use as stated here https://github.com/livewire/livewire/pull/10346#issuecomment-4809466315.

# Problem

Inside `SupportEvents::call`, both renderless and authorize attribute checked using `is_subclass_of`

```php
$this->component->getAttributes()
    ->filter(fn ($i) => is_subclass_of($i, BaseAuthorize::class) // here
    ->filter(fn ($i) => $i->getName() === $method)
    ->filter(fn ($i) => $i->getLevel() === AttributeLevel::METHOD)
    ->each(fn ($i) => $i->call($params));
```
I think the intent here is a good one that prevent using `Base**` attribute as public API. But, this will cause some unsual issue when running tests. Therefore, I still think this need to be fixed.

# Solution

Change `is_subclass_of` to `instanceof`

```php
$this->component->getAttributes()
    ->filter(fn ($i) => $i instanceof BaseAuthorize)
    ->filter(fn ($i) => $i->getName() === $method)
    ->filter(fn ($i) => $i->getLevel() === AttributeLevel::METHOD)
    ->each(fn ($i) => $i->call($params));
```

I have added regression test for both authorize unit test and renderless browser test.

# Files Changed
- `src/Features/SupportEvents/SupportEvents.php`
- `src/Features/SupportAuthorization/UnitTest.php`
- `src/Features/SupportRenderless/BrowserTest.php`

Fix: https://github.com/livewire/livewire/issues/10383

### DIFF
--- a/src/Features/SupportAuthorization/UnitTest.php
+++ b/src/Features/SupportAuthorization/UnitTest.php
@@ -877,6 +877,23 @@ class UnitTest extends TestCase
             ->dispatch('some-event')
             ->assertOk();
     }
+
+    public function test_base_authorize_attribute_blocks_unauthorized_user_via_event()
+    {
+        Gate::policy(AuthorizationPost::class, AuthorizationPostPolicy::class);
+
+        Livewire::actingAs(AuthorizationUser::find(2))
+            ->test(new class extends TestComponent {
+                #[BaseAuthorize(AuthorizationPost::class)]
+                #[BaseOn('some-event')]
+                public function store() : bool
+                {
+                    return true;
+                }
+            })
+            ->dispatch('some-event')
+            ->assertForbidden();
+    }
 }
 
 class AuthorizationUser extends AuthUser
--- a/src/Features/SupportAuthorization/UnitTest.php
+++ b/src/Features/SupportAuthorization/UnitTest.php
@@ -8,6 +8,7 @@ use Illuminate\Routing\UrlGenerator;
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Support\Facades\Session;
 use Livewire\Attributes\Authorize;
+use Livewire\Features\SupportEvents\BaseOn;
 use Livewire\Livewire;
 use Sushi\Sushi;
 use Tests\TestCase;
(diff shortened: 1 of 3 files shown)


calebporzio (commented):
Hey @ghabriel25 — thanks for this one, and for the clear writeup on the issue. The fix is right: the direct-call path runs these attributes for anything matching at the method level, so the event path filtering with `is_subclass_of` (which excludes the base class itself) was just an inconsistency. `instanceof` is the correct check.

Here's what I ran to verify:

- **Renderless browser test**: fails on `main`'s `SupportEvents.php`, passes with your fix. Solid regression test.
- **Authorize unit test**: I noticed your new test passes on `main` too — it dispatches as an *authorized* user and asserts Ok, which is the same outcome whether the authorize check runs or gets skipped, so it wasn't actually guarding the fix. I pushed a small commit to your branch adding a variant that dispatches as an unauthorized user and asserts Forbidden — verified it fails on `main` and passes here. Hope you don't mind the direct push!
- **Full unit suite**: green on this branch.

Happy with this — good to merge from my side.

ghabriel25:
@calebporzio Thanks for the review.

PR #10388, opened 2026-07-06

ghabriel25:
> [!NOTE]
> This PR includes the changes from PR https://github.com/livewire/livewire/pull/10398 so that tests pass locally.
> Once that PR is merged,  I will rebase/clean this PR and mark as ready

This is the third attempt from
- https://github.com/livewire/livewire/pull/10374
- https://github.com/livewire/livewire/pull/10382

## Summary

This PR adds a new `#[Middleware]` attribute that allows developers to apply Laravel middleware (class names or aliases, with optional parameters) directly to individual Livewire component actions.

Livewire actions are public methods that can be invoked from the frontend via `wire:click`, forms, or events. While route-level middleware (via `PersistentMiddleware`) and `#[Authorize]` already exist, there has been no first-class, declarative way to protect a specific action with arbitrary middleware the same way Laravel controllers support middleware attributes.

This feature closes that gap by enabling early, method-level protection so that unauthorized or restricted calls never reach business logic.

This PR resolve the concern from this comment https://github.com/livewire/livewire/pull/10374#pullrequestreview-4629568219 and some regression test added to ensure that it works as expected. 

Need to mention that this attribute does not behave like `#[Authorize]` which can be called in collection loop inside `SupportEvents` or regular attribute call hook. The reason is this attribute should gather all middleware and apply it all at once (not one by one) just like `PersistentMiddleware` therefore it takes precedence over event and attribute `call()` hook.

## How it works

1. **Collection (boot time)**  
   The `#[Middleware]` attribute stores middleware definitions (class or alias) in the component’s internal store, keyed by method name.

2. **Authorization conversion (boot time)**  
   Any middleware that resolves to `Illuminate\Auth\Middleware\Authorize` (including the `can:` alias) is automatically converted into a real `#[Authorize]` attribute on the method. This keeps the two systems consistent and reuses the existing authorization infrastructure.

3. **Application (call time)**  
   On the `call` lifecycle event (only for requests hitting Livewire update endpoint):
   - The real method name is resolved (including event listeners via `__dispatch`).
   - Middleware already applied by `PersistentMiddleware` is filtered out to avoid unnecessary duplicates.
   - Remaining middleware is applied **all at once** through the existing `Utils::applyMiddleware()` pipeline using a fake request (the same technique used by `PersistentMiddleware`).

Because middleware is gathered and applied as a single batch on the `call` event, it runs **before** normal attribute `call()` hooks and event listeners. This gives middleware precedence over `#[Authorize]`.

## Usage examples

**Basic usage (class or alias)**

```php
#[Middleware('auth')]
#[Middleware(EnsureEmailIsVerified::class)]
public function showContent()
{
    //
}

#[Middleware('guest')]
{
    //
}
```

**With parameters**

```php
#[Middleware('auth')]
#[Middleware('throttle:10')]
public function likePost(Post $post) { ... }

#[Middleware(CustomRoleMiddleware::class . ':admin')]
public function adminOnly() { ... }
```

**With events**

```php
#[On('mark-as-read')]
#[Middleware('auth')]
public function markAsRead()
{
    //
}
```

**Precedence over authorization**

```php
#[Middleware('auth')]
#[Authorize('update', 'post')] // only runs if auth middleware passes
public function save(Post $post) { ... }
```

**Automatic conversion of authorize middleware**

```php
// is converted at boot time into:
// #[Authorize('update', 'post')]
```

## Important behavioral notes

- The attribute only runs on subsequent Livewire update requests (the same constraint as PersistentMiddleware). It does not run on the initial page load.
- Middleware already registered as persistent is automatically excluded to avoid double execution.
- Middleware that is not registered as persistent but is applied both at the route level and via the attribute will run twice. This is intentional.

## What's changed

- New feature: `SupportActionMiddleware`
- New attribute: `Livewire\Attributes\Middleware` (extends internal `BaseMiddleware`)
- Small extension to `HandlesAttributes` (`setMethodAttribute`) so authorize conversion can inject attributes cleanly
- `PersistentMiddleware` now exposes the list of middleware it actually applied so the new feature can filter against it without re-resolving
- Registered in the service provider alongside other features

The design deliberately reuses existing infrastructure (Utils::applyMiddleware, the fake-request pattern, the attribute store, and the event listener resolution logic) rather than introducing a parallel system.


ghabriel25:
@calebporzio @joshhanley This is the best I could think for this feature.

joshhanley:
@ghabriel25 going to close this as it's not easy to rebase, can you fix up and resubmit against `4.x`? Thanks!

PR #10389, opened 2026-07-06
[4.x] Extract renderless-related methods into a `HandlesRenderless` trait

ghabriel25:
I think its time to move renderless-related methods from `Component` to `HandlesRenderless` trait.


src/Component.php
```diff
@@ -21,6 +21,7 @@
 use Illuminate\Support\Traits\Macroable;
 use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 use BadMethodCallException;
+use Livewire\Features\SupportRenderless\HandlesRenderless;
```
joshhanley, line 19:
```suggestion
use Livewire\Features\SupportAttributes\HandlesAttributes;
use Livewire\Features\SupportRenderless\HandlesRenderless;
use Livewire\Exceptions\PropertyNotFoundException;
use Livewire\Concerns\InteractsWithProperties;
use Illuminate\Support\Traits\Macroable;
use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
use BadMethodCallException;
```

joshhanley (approved):
@ghabriel25 thanks for the PR! Looks good to me. Only one small change, which I have committed, is to move the import up with the other features imports.

src/Component.php
joshhanley: Whoops, GitHub suggest didn't work the way I was expecting. Fixing it now.

joshhanley: Fixed.

ghabriel25:
@joshhanley thanks!

PR #10393, opened 2026-07-07
[4.x] Add `assertOnlyHasErrors` method

gdebrauwer:
Sometimes you want to test that a validation error is present, but you want to be sure that no other validation error occurred. Currently, there is no easy way to do that, unless you provide every possible validation error key to the assertHasNoErrors method call.

```php
// Will fail if 'email' error is missing OR if 'name' error is present
// BUT nothing will fail if 'phone' error is present
$component
    ->assertHasErrors(['email'])
    ->assertHasNoErrors(['name']);
```

This PR fixes that by creating a method that checks if the provided errors are present, but also asserts that no other validation errors occurred:

```php
// Will fail if 'email' error is missing OR if 'name' error is present OR if any other error is present
$component->assertOnlyHasErrors(['email']);
```

This works the same way as the `assertOnlyJsonValidationErrors` method that I added in Laravel (https://github.com/laravel/framework/pull/54678).


calebporzio:
i dig it 👌

PR #10394, opened 2026-07-07
[4.x] Chunked, resumable, and multiple-file S3 uploads

calebporzio:
Livewire's upload pipeline learns four new tricks: **multiple file uploads on S3**, **chunked uploads** that aren't bound by PHP's upload limits, **resumable uploads** that survive interruptions and page reloads, and **native S3 multipart** — all riding on one new idea: the *upload plan*.

## The big idea: upload plans

The old flow hard-wired exactly two shapes of upload (`upload:generatedSignedUrl` for local, `upload:generatedSignedUrlForS3` for S3 — single file only). That rigidity was where every limitation lived.

Now `_startUpload` asks a new `UploadPlanner` how the files should travel, and dispatches a single `upload:plan` event with a server-decided strategy:

| Strategy | When | What the JS does |
|---|---|---|
| `form` | small files, local disks | one classic multipart POST (unchanged behavior) |
| `chunked` | files over `chunk_threshold`, local disks | slice in the browser → POST chunks → server stitches, validates, stores |
| `s3` | S3 disks | presigned PUT per file; files over the threshold get a native multipart sub-plan |
| `reject` | declared size already violates your `max:` rule | fail with the real validation message **before any bytes move** |

The server plans, the client executes. The JS went from a 390-line monolith to `js/features/supportFileUploads/{manager,request,strategies/*}.js` — adding a future strategy is one PHP branch + one JS module.

## Chunked uploads

- Chunks POST to a new signed `livewire.upload-chunk` endpoint and are stream-assembled server-side, then run through the **exact same `validateAndStore()` path as a normal upload** — rules, sidecar, signed temp path. Chunking is invisible above that layer.
- Default chunk size is **1MB on local disks — deliberately below a stock `php.ini`'s 2M limit** — so users on untuned servers can upload files of any size your rules allow. `upload_max_filesize` stops mattering.
- Chunk size auto-grows for very large files so no upload ever needs more than 10,000 chunks (also S3's part limit).
- The chunk endpoint defaults to `throttle:600,1` since one file = many rapid requests.

## Resumability

Uploads are fingerprinted with `sha1(session id | name | size | type | lastModified | chunkSize)`. Every plan reports which chunks/parts the server already has, and the JS skips them — so a fresh upload is just a resume from nothing:

- Cancel, interruption, or a full page reload → re-selecting the file resumes where it left off (progress bar starts at the resumed offset, honestly).
- Editing the file bumps `lastModified` → new fingerprint → clean slate. Stale chunks can never be stitched into a new version.
- If every chunk landed but the completion response was lost, the JS re-sends the last chunk to re-trigger assembly.
- No database, no cron: stale chunks ride the existing 24-hour temp-upload cleanup.

## S3: multiple files + multipart

- `S3DoesntSupportMultipleFileUploads` is no longer thrown — the planner emits one presigned URL per file.
- Files over the threshold use real S3 multipart: presigned `UploadPart` URLs from the browser, completion server-side via a new signed `livewire.upload-multipart` endpoint.
- **ETags are collected server-side via `ListParts` at completion time**, so buckets need zero CORS changes beyond what presigned PUTs already require (no `ExposeHeaders: ETag` support tickets).
- Multipart resume works through a tiny fingerprint → `UploadId` mapping in `livewire-tmp/multipart/`; stale mappings self-heal.
- Presigned URLs expire after `max_upload_time` (5 min) — uploads that outlive their signatures (chunked *and* multipart) transparently re-handshake for a fresh plan and continue.

## Security model

One governing rule: **the server trusts nothing it didn't sign.**

- Fingerprints are **session-scoped**, so no user can derive (or be issued a signed reference to) another user's in-flight upload.
- The chunk id is a signed capability — `fingerprint|totalChunks|chunkSize` under HMAC — so the endpoint never trusts client-sent counts or sizes. The worst a hostile client can do is upload a file your `max:` rule already permitted.
- Fingerprints are regex-pinned to 40 hex chars before touching any path; assembled files face identical validation to direct uploads; MIME rules check content, not the client-declared type; both new endpoints are signed + CSRF-protected.

## Config

```php
'temporary_file_upload' => [
    // ...
    'chunking' => true,        // Upload large files in resumable chunks (S3 multipart on S3 disks)...
    'chunk_size' => null,      // Example: 10 * 1024 * 1024 (bytes)  | Default: 1MB (5MB on S3 — the multipart minimum)
    'chunk_threshold' => null, // Files larger than this are chunked | Default: chunk_size
],
```

`livewire-upload-progress` now also carries byte-level `loaded`/`total` (plus `id` and `property` on all upload events), so real "142.5 / 800 MB" status lines are possible.

## Notes for reviewers

- **Internal protocol change:** the `upload:generatedSignedUrl` / `upload:generatedSignedUrlForS3` events no longer exist — anything listening for them (Flux?) needs a grep before this merges.
- **S3 ops note (documented):** abandoned multipart uploads hold invisible storage — buckets want an `AbortIncompleteMultipartUpload` lifecycle rule.
- Users who set a custom upload `middleware` take over both endpoints and may need to raise their own throttle for chunks.

## Verification

- 18 new chunked/multipart unit tests through real HTTP routes (planner strategies, fingerprint determinism *and* session-scoping, capability tampering → 403, bounds → 422, full store→assemble→validate cycles, multipart complete/abort); full suite green.
- New Dusk browser test: 1MB fixture uploads in 64KB chunks through real Chrome, sha256-identical after reassembly.
- Manually verified end-to-end against a real app and a local MinIO: multiple S3 files, multipart, cancel → reload → resume (only missing chunks/parts re-sent, byte-identical results via `cmp`), and forced 401/403 mid-upload to prove the expired-signature re-handshake recovers.


wilianmaique:
up, i need this hehe 😄

kresnasatya:
This is what I'm looking for! 🥹
Rather than make custom chunk, I would like to use the built-in version from the author itself.
Can't wait to try it as soon as possible including integration with Flux UI. We use Pro version for file-upload. 😉

MrMooky:
I see this was merged into `main`, but none of the releases seem to include/mention these new features. Am I missing something?

PerryvanderMeer:
@MrMooky It wasn't merged into 4.x, only into main, so it will be included in Livewire v5

MrMooky:
> @MrMooky It wasn't merged into 4.x, only into main, so it will be included in Livewire v5

Oh. 😐

PR #10397, opened 2026-07-08
[4.x] Add JavaScript synthesizers via `Livewire.synth()`

calebporzio:
## What

PHP synthesizers let component properties hold rich values on the server (Carbon dates, DTOs, enums), but JavaScript only ever sees the raw dehydrated wire value — `$wire.date` is just an ISO string, a `TemporaryUploadedFile` is just a serialized blob string, and so on.

This PR adds the client-side counterpart: **JavaScript synthesizers**. Registering a synth for a PHP synthesizer's key upgrades the raw value into a rich JS object when state is hydrated on the frontend, and converts it back to wire format when state is diffed and sent to the server.

## Usage

```js
    Livewire.synth('cbn', {
        match: (value) => value instanceof Date,

        hydrate: (value, meta) => new Date(value),

        dehydrate: (value) => value.toISOString(),
    })
})
```

With that registered, every Carbon property is a real `Date` on the frontend:

```blade
<span x-text="$wire.publishedAt.toLocaleDateString()"></span>

<button x-on:click="$wire.publishedAt = new Date()">Publish now</button>
```

The API deliberately mirrors the PHP `Synth` class:

- `hydrate(value, meta)` — raw wire value + metadata tuple → rich JS value. Called whenever state arrives from the server (initial load and every response).
- `dehydrate(value)` — rich value → raw wire format. Called when Livewire diffs state to build update payloads. Must return what the PHP synth's `hydrate()` accepts.
- `match(value)` — recognizes rich values in state, so Livewire treats them atomically when diffing and knows how to dehydrate brand-new instances the user assigns (e.g. `$wire.date = new Date()`).

Nothing ships enabled by default — this is purely an opt-in extension point, so no existing behavior changes.

## How it works

- `extractData()` consults the registry: when a `[value, { s: 'key' }]` tuple has a registered JS synth, the value is hydrated (children first) into both `canonical` and `ephemeral`/`reactive` state.
- The diffing utilities (`diff`, `diffAndConsolidate`, `diffAndPatchRecursive`, `deeplyEqual`) treat synth-matched values as **atomic leaves**: they're compared by their dehydrated wire format and emitted into update payloads as raw wire values. Unchanged rich values keep their identity across round trips (no spurious reactive effects).
- Queued updates (`$wire.$set(...)`) and action parameters (`$wire.save(new Date())`) are dehydrated before being serialized into requests.
- `mergeNewSnapshot()` re-extracts the old canonical state from the previous raw snapshot instead of JSON-cloning it, since rich values don't survive a JSON round trip.

Every new code path short-circuits behind a `hasSynths()` check, so apps that never call `Livewire.synth()` run exactly the code they run today.

## Tests

- `js/synths.spec.js` — unit coverage for registration validation, hydration (including nested tuples and metadata pass-through), atomic diffing, wire-format diff emission, patch identity preservation, and non-mutating dehydration.
- `src/Features/SupportJsSynthesizers/BrowserTest.php` — end-to-end coverage: Carbon → `Date` hydration, assigning fresh rich values and round-tripping them through the server, identity preservation across `$refresh`, rich action parameters, and a custom Wireable + JS class pairing with nested `wire:model.live` binding.

## Docs

Extends `docs/synthesizers.md` with a "JavaScript synthesizers" section: the Date quickstart, a breakdown of the three functions, a custom PHP + JS pairing continuing the page's existing `Address` example, caveats (global-per-key registration, atomicity, `wire:model` interplay), and a reference list of built-in synth keys.

### DIFF
--- a/docs/synthesizers.md
+++ b/docs/synthesizers.md
@@ -350,6 +350,31 @@ Now the same `Address` concept exists on both sides of the wire:
 
 You can still mutate individual fields (`$wire.address.street = '...'` or `wire:model="address.street"`) — when a rich value changes, Livewire sends its entire dehydrated form to the server, where the PHP synthesizer hydrates it back into an `Address`.
 
+### Reactivity
+
+Rich values participate in Alpine's reactivity at the property level. Any effect that reads the property — `x-text`, `$wire.$watch()`, `wire:dirty` — re-runs whenever the property is _replaced_:
+
+```blade
+<!-- This re-renders when $wire.publishedAt is assigned a new value... -->
+<div x-text="$wire.publishedAt.toDateString()"></div>
+
+<!-- ...whether from the client: -->
+<button x-on:click="$wire.publishedAt = new Date()">Publish now</button>
+```
+
+The same applies when the property changes server-side: because rich values are atomic, Livewire replaces the whole value when merging a server response, which triggers any effects reading it. And when a rich value _hasn't_ changed, Livewire keeps the existing object (same identity) so effects don't re-run needlessly.
+
+Rich values built from plain classes — like the `Address` example above — are also deeply reactive, just like plain objects: mutating `$wire.address.street` re-runs an effect reading `$wire.address.full`.
+
+The one exception is in-place mutation of native objects like `Date`:
+
+```blade
+<!-- This will NOT re-render the x-text above: -->
+<button x-on:click="$wire.publishedAt.setFullYear(2030)">Won't react</button>
+```
+
+JavaScript proxies can't intercept a `Date`'s internal methods, so Alpine (like Vue) can't observe the mutation. The change is still detected by Livewire's state diffing and sent to the server on the next request — but no frontend effects fire. Treat values like dates as immutable: assign a new instance instead of mutating.
+
 ### Things to know
 
 * **Registration is global per key.** Registering a synth for `cbn` upgrades _every_ Carbon and native date property across your entire application, so make sure your frontend code is prepared for that.
--- a/docs/synthesizers.md
+++ b/docs/synthesizers.md
@@ -244,3 +244,117 @@ class AddressSynth extends Synth
     }
 }
 ```
+
+## JavaScript synthesizers
+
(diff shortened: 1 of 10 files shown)


calebporzio:
## Benchmark: cost of always-registered synths

Since stock synths (e.g. file uploads) will likely ship registered by default, the `hasSynths()` early return will eventually always be true — so the steady-state overhead is what matters. Measured the hot functions on two component shapes: **typical** (~60 state nodes) and **heavy** (1,000-row data table, ~8,000 nodes), with 3 `instanceof`-matching synths registered.

After making `dehydrateTree` copy-on-write (plain subtrees return by reference, zero allocations):

| Operation | 0 synths | 3 synths, plain data | 3 synths, 1k rich values |
|---|---|---|---|
| `deeplyEqual` (typical) | 0.0021 ms | 0.0036 ms | 0.0049 ms |
| `deeplyEqual` (heavy) | 0.33 ms | 0.49 ms | 1.35 ms |
| `diffAndConsolidate` (typical) | 0.0026 ms | 0.0034 ms | 0.0048 ms |
| `diffAndConsolidate` (heavy) | 0.90 ms | 1.00 ms | 1.88 ms |
| `diffAndPatchRecursive` (typical) | 0.0030 ms | 0.0045 ms | — |
| `diffAndPatchRecursive` (heavy) | 0.41 ms | 0.56 ms | — |

Takeaways:

- **Typical components pay ~1–2µs extra per operation** with synths registered — unmeasurable against a network round trip.
- A pathological 8k-node component pays roughly **+0.1–0.3ms per diff/patch** (~10–50% relative on those ops, well under 1ms per full request cycle, <1% of a realistic request).
- The rich-values column is the cost of actually *using* the feature on 1,000 hydrated values at once — opt-in and still ~1ms scale.
- Before the copy-on-write change, `dehydrateTree` deep-copied entire trees on every equality check whenever any synth was registered (heavy `deeplyEqual` was 0.70ms and allocation-heavy). That's fixed; `hasSynths()` is also now a non-allocating length check, so keeping it costs nothing even once it's always true.

PR #10398, opened 2026-07-08
[4.x] Fix `RedirectResponse` being handled by Livewire redirector during request lifecycle

ghabriel25:
# Scenario
### Case 1
Imagine user create custom exception handler that convert the exception to specific redirect response

### Case 2
When testing this PR https://github.com/livewire/livewire/pull/10388, I found that all redirect response during request lifecycle handled by Livewire will throw an exception as shown below

image

```php
<?php

use Livewire\ComponentHook;

class CustomHook extends ComponentHook
{
    function call()
    {
        // this will return RedirectResponse
        throw new AuthenticationException();

        // or
        return to_route('home');
    }
}
```
```php
<?php

use Livewire\Component;

new class extends Component
{
    public function save()
    {
        throw new AuthenticationException();
    }
}
```

This is caused by `RedirectResponse` handled by Livewire custom redirector.

The exception originally comes from `AuthenticationException` with flow

**AuthenticationException**

↓

`redirect()->guest($url)`

↓

**SupportRedirects/Redirector::to()**

Redirector swap happened during component hook boot time and only restored on dehydrate, which means any redirect response during request lifecycle will be handled by Livewire redirector

# Problem

When `RedirectResponse` handled by Livewire redirector, it return the custom redirector instance.

```php
public function to($path, $status = 302, $headers = [], $secure = null)
{
    $this->component->redirect($this->generator->to($path, [], $secure));

    return $this;
}
```

I dont know the reason why this return Livewire redirector instance eventhough its should return `RedirectResponse` instance and not the redirector itself.

# Solution

Always return parent object to handle redirect response, since its general redirect and not using navigate.

```php
public function to($path, $status = 302, $headers = [], $secure = null)
{
    $this->component->redirect($this->generator->to($path, [], $secure));

    return parent::to($path, $status, $headers, $secure);
}
```
This way, it will handles the `RedirectResponse` properly

> [!NOTE]
> since user can return redirect response inside action that either called as normal call or as event listener, I also have checked that return effect from this response can be json-encoded without throwing any error. I just not sure if it should be exlcluded or not from return effects.

### DIFF
--- a/src/Features/SupportRedirects/Redirector.php
+++ b/src/Features/SupportRedirects/Redirector.php
@@ -14,11 +14,7 @@ class Redirector extends BaseRedirector
     {
         $this->component->redirect($this->generator->to($path, [], $secure));
 
-        if (app(HandleRequests::class)->isLivewireRoute()) {
-            return parent::to($path, $status, $headers, $secure);
-        }
-
-        return $this;
+        return parent::to($path, $status, $headers, $secure);
     }
 
     public function away($path, $status = 302, $headers = [])
--- a/src/Features/SupportRedirects/Redirector.php
+++ b/src/Features/SupportRedirects/Redirector.php
@@ -4,6 +4,7 @@ namespace Livewire\Features\SupportRedirects;
 
 use Livewire\Component;
 use Illuminate\Routing\Redirector as BaseRedirector;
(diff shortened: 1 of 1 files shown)


ghabriel25:
Ready to be reviewed, thanks!

PR #10399, opened 2026-07-09
[4.x] Fix redirect from `AuthenticationException` not being handled properly

ghabriel25:
# Scenario
If `AuthenticationException` thrown before component dehydrate, the redirect from this exception will be handled by Livewire custom redirector which will throw an exception as shown below

image

```php
<?php

new class extends Component
{
    public function save()
    {
        throw new AuthenticationException();
    }
}
```
```blade
<div>
    <button type="button" wire:click="save">Save</button>
</div>
```

# Problem
In `HandleRequests::handleUpdate()`, it only catch `TypeError` but swallow `AuthenticationException` that has redirect mechanism to login route. Because Laravel redirector being swapped during component hook boot time, any redirect response during request lifecycle will be handled by Livewire custom redirector which will throw an exception as shown above.

# Solution
Catch `AuthenticationException` and set redirect `effects` 

```php
try {
    [ $snapshot, $effects ] = app('livewire')->update($snapshot, $updates, $calls);
} catch (\TypeError $e) {
    report($e);

    abort(419);
} catch (\Illuminate\Auth\AuthenticationException $e) {
    $effects['redirect'] = $e->redirectTo(request());
    $effects['redirectUsingNavigate'] = true;
}
```
I've look up for Laravel exception that has redirect response and `AuthenticationException` was the only one.

### DIFF
--- a/src/Mechanisms/HandleRequests/HandleRequests.php
+++ b/src/Mechanisms/HandleRequests/HandleRequests.php
@@ -209,6 +209,9 @@ class HandleRequests extends Mechanism
                 if (config('app.debug')) throw $e;
 
                 abort(419);
+            } catch (\Illuminate\Auth\AuthenticationException $e) {
+                $effects['redirect'] = $e->redirectTo(request());
+                $effects['redirectUsingNavigate'] = true;
             }
 
             $componentResponses[] = [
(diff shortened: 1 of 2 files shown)


ghabriel25:
Closing in favor https://github.com/livewire/livewire/pull/10398

PR #10400, opened 2026-07-09
[Spike] Stock `fil` JS synth: rich `TemporaryUpload` objects

calebporzio:
Acid test for #10397 — builds a real stock file-upload synth on top of the JS synthesizer API to validate it before merge. Draft, targeting the `js-synths` branch.

## What it does

File properties are rich objects on the frontend instead of `"livewire-file:..."` strings:

```blade
<span x-text="$wire.photo.name"></span>          <!-- original client filename -->
<span x-text="$wire.photo.extension"></span>
<img x-bind:src="$wire.photo.temporaryUrl()">    <!-- previews without a Blade re-render -->

<button x-on:click="$wire.removeUpload('photos', $wire.photos[0].filename)">Remove</button>
```

Two pieces:

- **PHP**: `FileUploadSynth::dehydrate()` now ships `name` (original client filename) and `previewUrl` (signed preview URL, when previewable) in the tuple metadata.
- **JS**: a `TemporaryUpload` class + `Livewire.synth('fil', ...)` registered by default in `supportFileUploads.js`. `toString()`/`toJSON()` degrade to the raw wire value.

## API verdict

The `Livewire.synth()` API survived unchanged — no signature additions needed:

- **The `meta` parameter proved essential and sufficient.** The original filename isn't derivable client-side (it lives in a server-side `.json` sidecar), and preview URLs must be signed server-side. Both travel naturally in the dehydration tuple's metadata — no component/path context param was needed.
- **Arrays of rich values worked with zero special-casing** — multiple uploads are per-element tuples, so each hydrates independently; splicing one out diffs and round-trips correctly.
- **The upload machinery needed no changes** — uploads set properties server-side, so the rich object is purely a read-side upgrade; `removeUpload`/`cancelUpload`/loading states all work untouched.

## Findings along the way

- `getClientOriginalName()` can decode binary garbage from hashed paths (seen in unit tests) — meta generation guards with `mb_check_encoding` so snapshots can't fail JSON encoding.
- Registering by default means `$wire.photo` changes type from string → object: **a BC decision, not a technical one**. `toString()`/`toJSON()` soften it, but code doing `$wire.photo.startsWith(...)` would break. Shipping it opt-in first is also viable.
- Cost: one tiny storage read (name sidecar) + one URL signing per file property per response, and ~250 bytes of snapshot meta per file.

## Tests

- Entire existing `SupportFileUploads` browser suite passes with the synth **enabled by default** (the real BC signal).
- New browser tests: rich single upload (name/extension/preview image actually loads), multiple uploads as rich arrays, removal via the rich object's temp filename.
- All 55 file-upload PHP unit tests pass with the new meta.

### DIFF
--- a/docs/synthesizers.md
+++ b/docs/synthesizers.md
@@ -350,6 +350,31 @@ Now the same `Address` concept exists on both sides of the wire:
 
 You can still mutate individual fields (`$wire.address.street = '...'` or `wire:model="address.street"`) — when a rich value changes, Livewire sends its entire dehydrated form to the server, where the PHP synthesizer hydrates it back into an `Address`.
 
+### Reactivity
+
+Rich values participate in Alpine's reactivity at the property level. Any effect that reads the property — `x-text`, `$wire.$watch()`, `wire:dirty` — re-runs whenever the property is _replaced_:
+
+```blade
+<!-- This re-renders when $wire.publishedAt is assigned a new value... -->
+<div x-text="$wire.publishedAt.toDateString()"></div>
+
+<!-- ...whether from the client: -->
+<button x-on:click="$wire.publishedAt = new Date()">Publish now</button>
+```
+
+The same applies when the property changes server-side: because rich values are atomic, Livewire replaces the whole value when merging a server response, which triggers any effects reading it. And when a rich value _hasn't_ changed, Livewire keeps the existing object (same identity) so effects don't re-run needlessly.
+
+Rich values built from plain classes — like the `Address` example above — are also deeply reactive, just like plain objects: mutating `$wire.address.street` re-runs an effect reading `$wire.address.full`.
+
+The one exception is in-place mutation of native objects like `Date`:
+
+```blade
+<!-- This will NOT re-render the x-text above: -->
+<button x-on:click="$wire.publishedAt.setFullYear(2030)">Won't react</button>
+```
+
+JavaScript proxies can't intercept a `Date`'s internal methods, so Alpine (like Vue) can't observe the mutation. The change is still detected by Livewire's state diffing and sent to the server on the next request — but no frontend effects fire. Treat values like dates as immutable: assign a new instance instead of mutating.
+
 ### Things to know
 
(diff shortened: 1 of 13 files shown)


calebporzio:
## Round 2: the insane version

Pushed the full rich upload lifecycle. All of it works, verified in real Chrome:

```blade

<!-- Live during upload: blob preview, no server round trip ever -->
<img x-show="$wire.photo" x-bind:src="$wire.photo?.previewUrl">

<!-- Reactive progress bar -->
<div x-show="$wire.photo?.isUploading">
    <div x-bind:style="`width: ${$wire.photo.progress}%`"></div>
</div>

<!-- Self-removing (cancels if still in flight) -->
<button x-on:click="$wire.photo.remove()">×</button>
```

How it works: on upload start, the property optimistically holds a **pending** `TemporaryUpload` built from the native `File` object. `.progress` mutations are natively reactive (Alpine proxies class instances), `.previewUrl` is a `URL.createObjectURL` blob that stays valid after the upload settles (a filename→blob registry bridges the swap to the server-hydrated instance — previews never fetch from the server). When the server confirms, the pending graduates and the response swaps in the hydrated object. A lovely emergent property: pendings **survive intermediate responses for free**, because the patcher only replaces values whose canonical wire form actually changed.

This required exactly two additive API changes to `Livewire.synth()`:

1. **`hydrate(value, meta, context)`** — context is `{ component, path }`, so rich values know where they live (powers `.remove()`, including `photos.1` → `photos` derivation for array elements).
2. **`dehydrate()` may return `undefined`** — "no wire representation yet." Such values are never diffed to the server and are **omitted from arrays** rather than serialized as `null` (that null would have corrupted server state — caught and unit-tested). Diffs that become empty after omission are dropped entirely, so mid-upload requests send no update for the property.

Coverage: 13/13 upload browser tests (4 new: reactive progress lifecycle log, blob-preview-never-hits-server, single self-remove + input reset, array-element self-remove), 126 JS unit tests, and regression sweeps on dirty/query-string/synth suites — all green.

PR #10401, opened 2026-07-09
Stock `fil` JS synth: rich `TemporaryUpload` objects with an optimistic upload lifecycle

calebporzio:
## What

Rebuilds the `fil` synth spike (#10400) on top of the beefier-file-uploads architecture. File properties are rich objects on the frontend instead of `"livewire-file:..."` strings — and the property now holds an optimistic pending object from the instant a file is selected, before a single byte reaches the server:

```blade
<form wire:submit="save">
    <div x-show="$wire.photo">
        <img x-bind:src="$wire.photo?.previewUrl">

        <progress max="100" x-bind:value="$wire.photo?.progress" x-show="$wire.photo?.isUploading"></progress>

        <button type="button" x-on:click="$wire.photo.remove()">Remove</button>
    </div>

</form>
```

Previews come from a local blob URL created from the file already sitting in the browser — no waiting on the tmp upload, no re-downloading a file the client just sent.

## The rich object

Property | Description
--- | ---
`name` | Original client-side filename
`extension` | Derived from `name`
`filename` | Hashed temp filename on the server (`null` while uploading)
`isUploading` / `progress` | Reactive upload lifecycle state
`isPreviewable` / `previewUrl` | Best-available preview: local blob URL when possible, signed server URL otherwise
`temporaryUrl()` | The signed server preview URL (JS equivalent of PHP's `->temporaryUrl()`)
`remove()` | Removes the upload from its property — cancels it if still in flight
`toString()` / `toJSON()` | Degrade to the raw wire value

## How it fits the new architecture

- **Pending lifecycle rides the central progress tracker.** The spike patched progress into per-XHR handlers; here pendings update inside `makeProgressTracker`, so **chunked, resumable, and S3 multipart uploads get reactive rich progress for free**.
- **Re-handshake safe.** Mid-upload plan refreshes call `startUpload()` again for the same upload — pending seeding is guarded to run once per upload object.
- **Meta reconciles with the snapshot-persisted URL cache.** `FileUploadSynth::dehydrate()` now ships `name` plus an eagerly signed preview `url` in tuple meta (so `$wire.photo.temporaryUrl()` works without Blade ever asking). Eager signing goes through `temporaryUrl()`'s existing cache, so S3 presigned URLs still sign at most once per file lifetime and stay morph-stable.
- **Graduation preserves previews.** When an upload finishes, pendings receive their wire value in place (flipping `isUploading` reactively) and their blob URLs are stashed by temp filename, so previews survive the server swapping in freshly hydrated instances — no flash, no refetch. Errored/cancelled uploads revert the property to its pre-upload value and revoke their blob URLs.

## Core synth contract extensions

Two pieces from the spike's final iteration that aren't in the merged #10397:

- `hydrate(value, meta, context)` receives `{ component, path }` so rich values know where they live in state — `remove()` is built on it.
- `dehydrate()` may return `undefined` to mean "no wire representation yet": such values are never emitted into update payloads and are omitted from arrays entirely (rather than serializing as `null`). This is the primitive that lets optimistic pending uploads sit on properties without corrupting server state.

Both are covered in `js/synths.spec.js` and documented in `docs/synthesizers.md`.

## BC note

Registering the synth by default means `$wire.photo` changes type from string → object. `toString()`/`toJSON()` soften the edge, but code doing string ops on file props directly would break. Meta adds ~250 bytes per file property plus one sidecar read + one URL signing per response worst case.

## Tests

- Entire existing `SupportFileUploads` browser suite passes with the synth enabled by default (14 tests), plus `ChunkedUploadsBrowserTest` and the `SupportJsSynthesizers` suite (10 tests).
- New browser tests: rich hydration (name/extension/preview actually loads), reactive progress + blob previews surviving settlement, self-removal (including input clearing), and multiple uploads as rich arrays with per-file removal.
- All 94 file-upload unit tests pass — one assertion updated deliberately: a file that never generated a URL now ships `name` + `url` meta instead of nothing.
- Verified end-to-end in a real app: 100ms after file selection the property holds a pending object with a `blob:` preview; after settling, the blob URL survives the server swap.

## Docs

- `docs/uploads.md`: new "Rich upload objects in JavaScript" section (single + multiple examples, full property reference, CSP note for `blob:` URLs).
- `docs/synthesizers.md`: "Hydration context" and "Values without a wire representation" sections.

## Known follow-ups

- Blob URLs stashed for finished uploads are only revoked via `remove()` — replacing an upload or resetting the property server-side leaves them alive until navigation. Bounded by page lifetime; could be tied into snapshot merging later.
- A network failure on the initial `_startUpload` roundtrip leaves the pending object (and the upload bag — pre-existing) wedged; the re-handshake path already handles this, the initial call could reuse it.


calebporzio:
Fixed a bug found in real-world testing (b4fc09bb): clicking a `remove()` button on a settled upload threw `Cannot read properties of undefined (reading 'shift')` from the `upload:removed` handler.

**Cause:** rich upload objects live inside reactive state, so `$wire.photo.context.component` returns an *Alpine proxy* of the component, not the raw instance. `getUploadManager()`'s WeakMap missed on the proxy identity and created a second manager with duplicate listeners — the original manager's handler crashed on a removal its `removeBag` never saw, killing the finish-callback chain (and silently breaking in-flight cancellation via `remove()` too).

**Fix:** `getUploadManager()` now unwraps with `Alpine.raw()` before the WeakMap lookup. The removal browser test captures `window` errors and asserts the finish callback fires, which reproduces the bug before the fix.

calebporzio:
Removal is now optimistic (b77e0363): `$wire.photo.remove()` nulls the property (or shrinks the array) in the same synchronous frame as the call — no round trip gating the UI.

The interesting bit is how the echo is suppressed without touching PHP: the detachment is applied to **both the reactive and canonical trees**, so the diff never emits it as a property update. `_removeUpload` therefore still runs against the property as the server last knew it, keeping its gated, property-scoped delete (clients still can't delete arbitrary tmp files by name — plain tmp filenames aren't signed, so a "trust the client, delete by name" fallback was deliberately avoided). The response snapshot then reconciles to exactly the state the client already shows.

Browser tests pin the behavior down deterministically: they click and read the property in the same synchronous JS frame (an XHR can't have completed), then confirm the server settles on identical state with no flicker and the finish callback still fires.

calebporzio:
**Footgun fix: array property + file input missing `multiple`** (2a16773c)

Binding `wire:model="photos"` (an array property) to a file input without the `multiple` attribute would swap the array out for a single pending upload object for the whole duration of the upload — breaking any `x-for` iterating the property — before the server response flipped it back to an array.

The server already had an opinion here: `_finishUpload` has always **appended** non-multiple uploads into array properties. The client now honors the same rule — `manager.upload()` detects an array property and routes through the multiple/append path, so the property stays an array through the entire lifecycle (pending object appended, never swapped in bare). This covers both `wire:model` and the manual `$wire.upload()` API (callback still receives a single filename there).

- New browser test asserts the property shape is an array at every reactive change during the upload, and that a second selection through the same single input appends.
- Docs note added under "Handling multiple files": the property's shape wins; `multiple` just tells the browser to allow multi-select.
- Verified in the playground app: pending shape is now `array`, settles at `array(1)`, zero console errors.

calebporzio:
**Type rules now preflight too** (b4524f40)

Size rules were already rejected from declared metadata during the plan handshake — now `image`, `mimes`, `mimetypes`, and `extensions` are as well. Dropping a 200MB video into an `image` field errors instantly instead of after a full upload (or a full S3 multipart dance).

Design notes:
- **Lenient by construction.** Laravel's authoritative check content-sniffs the real file after upload, so the preflight only rejects when *every* declared signal (filename extension, browser MIME type) provably contradicts the rule. Empty or unrecognized signals pass through to the real server-side validation — the one asymmetry is a deliberately misnamed file (real PNG named `.txt`), which preflight-rejects where post-upload sniffing would have accepted.
- **`extensions` is an exact mirror** — Laravel itself validates the client filename's extension, which is exactly what the plan receives.
- **`image` uses the widest extension set across supported framework versions** (svg is included by Laravel ≤11, excluded by 12+ unless `allow_svg`) so the preflight never rejects a file the installed framework would accept.
- **`mimetypes` mirrors Laravel's wildcard families** (`image/*`), and `mimes` gets the same jpg⇔jpeg merge Laravel applies.
- `DeclaredSizeRules` → `DeclaredRules`, returning both size and type constraints; the reject plan flows through the existing `_uploadErrored` path so error messages match post-upload validation exactly.

Coverage: unit tests for each rule's rejection + the lenient pass-through cases, a browser test proving zero bytes reach `livewire-tmp`, and a real-app drive showing **0 requests to the upload endpoint**, instant `The photos.0 field must be an image.` error, and clean revert of the optimistic pending state.

PR #10404, opened 2026-07-11
[4.x] Add `wire:if` and `wire:for` directives

calebporzio:
## Summary

Adds `wire:if` and `wire:for` — the `x-if`/`x-for` counterparts to `wire:show`/`wire:text`. Both must be used on `<template>` tags (like their Alpine equivalents) and evaluate against component properties:

```blade
<template wire:if="confirmingDelete">
    <div>Are you sure?</div>
</template>

<template wire:for="task in tasks" wire:for:key="task.id">
    <li>
        <span wire:text="task.title"></span>
        <button wire:click="remove(task.id)">Remove</button>
    </li>
</template>
```

They react to both server updates and pure client-side mutations (`$wire.show = true`, `$wire.tasks.push(...)`) with no round-trip.

## Implementation notes

**`wire:if`** follows the exact `wire:show` pattern: `Alpine.bind` an `x-if` closure that evaluates through `evaluateActionExpression` (so bare identifiers get `$wire.`-prefixed, Alpine scope keys are left alone).

**`wire:for`** can't use that pattern — `x-for` regex-parses its expression *string* (`item in items`), so a function expression never works. Instead the items half of the expression is rewritten statically: `task in tasks` → `task in $wire.tasks`. Iterator aliases are left untouched, and nested loops work because inner expressions skip Alpine scope keys (`item in list.items` stays unprefixed when `list` comes from the outer loop).

**Keys** live on a dedicated `wire:for:key` attribute (following the `wire:sort:group` naming pattern), because `wire:key` holds a static Blade-rendered string everywhere else while a template loop needs a per-item client-side expression — overloading one attribute with two value languages felt like a trap. Alpine's `:key` also works; a plain `wire:key` on the template logs a console warning pointing at `wire:for:key`. The key expression is passed through `x-bind:key` in the same `Alpine.bind` object — Alpine processes `bind` before `for`, so it lands where `x-for` reads per-item keys from (binding `x-for` alone at `interceptInit` time would run before the template's own `:key` attribute is processed, silently index-keying every list).

## How this interacts with morphing

Alpine morph already has a mechanism for templates: as it patches each element pair it runs `Alpine.cloneNode(from, to)`, so the incoming (`to`) tree's `x-if`/`x-for` templates render their content *into the incoming tree*, and morph then pairs generated content 1:1 against the live content.

That mechanism has a race with `Alpine.transaction()` from #9882 (which correctly blocks reactive effects until after the morph): when a server update changes a list's length, the incoming template renders the **new** count while the live side is frozen at the **old** count — the leftover incoming rows get inserted as ghost/duplicate rows. Reproducible on `main` today with plain `x-for` bound to `$wire` state: growing a 2-item list produces 4 `<li>`s (one empty). This is the same family as #10144 (closed unmerged) and the navigate-cache duplicates fixed in #10336/#10339.

This PR takes the position that morph should not touch template-generated content at all — Alpine owns it on both sides:

1. When morph reaches an `x-if`/`x-for`/`wire:if`/`wire:for` template pair, it `skipUntil()`s past the live generated elements (same mechanism island fragment markers use) and `skip()`s the template pair itself. Skipping the pair prevents the per-element clone-seeding from ever rendering content into the incoming tree, which is what previously ghosted rows. State changes still land after the morph and Alpine re-renders reactively.
2. A guard in the `removing` hook protects generated elements when the incoming HTML ends before reaching them. Alpine still tears them down itself when the template is removed or its expression goes falsy.
3. `wire:if`/`wire:for` bindings are wrapped in `Alpine.skipDuringClone` (same as `wire:navigate`), so they never activate during clone-mode initialization.

Because 1–2 apply to plain Alpine templates too, this also fixes the existing ghost-row bug on `main` — covered by regression tests using plain `x-if`/`x-for` bound to `$wire`.

## Tests

15 browser tests across `SupportWireIf`/`SupportWireFor` covering: toggling in/out of the DOM, initial-true render, negation, client-side-only reactivity (`$wire.show = ...`, `$wire.fruits.push(...)`), server-driven add/remove with exact row counts, `(item, index)` aliases, actions receiving the iterated item, nested loops, keyed reorder preserving DOM state (typed input travels with its row through `array_reverse`), morph survival (typed input inside generated content persists across an unrelated `$refresh` with trailing siblings intact), and plain `x-if`/`x-for` ghost-row regressions.

Green regression sweeps: `SupportWireShow`/`SupportWireText`, `SupportIslands` (27 tests — heaviest consumer of the `skipUntil` machinery), `SupportMorphAwareBladeCompilation`, `AlpineMorphingBrowserTest` (the #9882 suite), and the JS unit suite (126).

Docs included for both directives (`docs/wire-if.md`, `docs/wire-for.md`) and registered in the nav.

### DIFF
--- a/docs/wire-if.md
+++ b/docs/wire-if.md
@@ -19,11 +19,12 @@ The `<template>` tag itself is never displayed — when the expression is truthy
 Here's an example of using `wire:if` to reveal a delete confirmation prompt:
 
 ```php
+<?php
+
 use Livewire\Component;
 use App\Models\Post;
 
-class ShowPost extends Component
-{
+new class extends Component {
     public Post $post;
 
     public $confirmingDelete = false;
@@ -36,12 +37,12 @@ class ShowPost extends Component
 
         return $this->redirect('/posts');
     }
-}
+};
 ```
 
 ```blade
 <div>
-    <button x-on:click="$wire.confirmingDelete = true">Delete Post</button>
+    <button wire:click="confirmingDelete = true">Delete Post</button>
 
     <template wire:if="confirmingDelete">
         <div>
@@ -49,13 +50,13 @@ class ShowPost extends Component
 
             <button wire:click="delete">Yes, delete it</button>
 
-            <button x-on:click="$wire.confirmingDelete = false">Cancel</button>
+            <button wire:click="confirmingDelete = false">Cancel</button>
         </div>
     </template>
 </div>
 ```
 
-When the "Delete Post" button is clicked, the confirmation prompt is added to the page instantly — no server round-trip. When cancelled, it's removed from the DOM entirely.
+When the "Delete Post" button is clicked, the confirmation prompt is added to the page instantly — property assignments like `wire:click="confirmingDelete = true"` update client-side state directly, without a server round-trip. When cancelled, it's removed from the DOM entirely.
 
 > [!info] Single root element
 > Like `x-if`, the `<template>` tag must contain a single root element.
--- a/docs/wire-if.md
+++ b/docs/wire-if.md
@@ -1,0 +1,84 @@
+Livewire's `wire:if` directive conditionally renders an element based on the result of an expression — entirely on the client, without a server round-trip.
+
+It is the Livewire equivalent of Alpine's [`x-if`](https://alpinejs.dev/directives/if) directive, evaluated against your component's properties instead of Alpine data.
+
+Unlike `wire:show`, which toggles an element's visibility using CSS (`display: none`), `wire:if` adds and removes the element from the DOM entirely. And unlike Blade's `@if`, the condition is re-evaluated instantly on the client whenever the property changes — no network request required.
+
+Because the content must be able to be added and removed from the page, `wire:if` must be used on a `<template>` tag:
+
+```blade
+<template wire:if="expression">
+    <div>...</div>
+</template>
+```
+
+The `<template>` tag itself is never displayed — when the expression is truthy, its contents are rendered directly after it in the DOM.
+
+## Basic usage
+
+Here's an example of using `wire:if` to reveal a delete confirmation prompt:
+
+```php
+use Livewire\Component;
+use App\Models\Post;
+
+class ShowPost extends Component
+{
+    public Post $post;
+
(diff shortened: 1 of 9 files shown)


ghabriel25:
Awesome!

docs/wire-if.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+The `<template>` tag itself is never displayed — when the expression is truthy, its contents are rendered directly after it in the DOM.
+
+## Basic usage
+
+Here's an example of using `wire:if` to reveal a delete confirmation prompt:
+
+```php
+use Livewire\Component;
```
calebporzio, line 24: we should make this example in the single file component format right? reference the rest of the docs and see what we're doing for full component examples.

docs/wire-if.md
```diff
@@ -0,0 +1,84 @@
...
+        $this->post->delete();
+
+        return $this->redirect('/posts');
+    }
+}
+```
+
+```blade
+<div>
+    <button x-on:click="$wire.confirmingDelete = true">Delete Post</button>
```
calebporzio: can we make this wire:click instead of x-on:click - will that even work with an assignment? curious

docs/wire-if.md
calebporzio: Done — both wire-if and wire-for examples now use the single-file anonymous class format (`new class extends Component {...}`) matching the rest of the docs (wire-sort, wire-intersect, components.md). Pushed in d667b54d.

docs/wire-if.md
calebporzio: Yep — it works. `wire:click` maps to `x-on:` and evaluates through the same `$wire.`-prefixing contextualizer as everything else, so `wire:click="confirmingDelete = true"` becomes `$wire.confirmingDelete = true`: a pure client-side ephemeral assignment, no request fired. Verified in a real browser — the client-side reactivity test now uses exactly this form (`wire:click="show = ! show"` toggling a `wire:if` with no `waitForLivewire()`), and the docs example was switched to `wire:click` assignments for both the open and cancel buttons. Also called this out in the docs prose since it's a nice non-obvious pattern.

PR #10405, opened 2026-07-11
Add CSS View Transitions support to wire:navigate

calebporzio:
## Summary

Adds native [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API) support to `wire:navigate`, closing the long-standing #6329 (the most-upvoted open feature idea after partial rendering).

Three ways to opt in:

```blade
<a href="/posts" wire:navigate.transition>Posts</a>
```

```php
// config/livewire.php — enables transitions for every navigation,
// including browser back/forward visits...
'navigate' => [
    'transitions' => true,
],
```

```js
Livewire.navigate('/posts', { transition: true })
```

Because the swap is wrapped in `document.startViewTransition()`, shared-element morphs require no new API — elements with matching `view-transition-name` styles across pages animate between positions automatically, and all animation timing/easing is customizable with standard `::view-transition-*` CSS.

## Implementation notes

- New `js/plugins/navigate/transition.js` exposes `transitionPageSwap(transition, update)`, which wraps both page-swap sites in the navigate code (link navigations and cached back/forward restores). The global flag is enabled via a direct module import from `supportNavigate.js` — no new public API surface.
- Guards are shared with `wire:transition`: falls back to an instant swap when the API is unsupported, when the user prefers reduced motion, or when a modal dialog is open — and both features now use the same extracted `skipTransitionWhenDialogOpens()` observer to skip mid-animation if a dialog opens.
- The global config plumbs through the script tag as `data-navigate-transitions` (and `livewireScriptConfig.navigateTransitions`), following the existing `show_progress_bar` pattern.
- `wire-navigate.js`'s hand-written attribute map (5 entries for 2 modifiers) is now generated, so all 16 ordered combinations of `.hover` / `.preserve-scroll` / `.transition` register correctly — including for `wire:current`'s selector list.
- Per-link transitions apply only to that forward navigation; back/forward transitions are tied to the global config so popstate behavior stays predictable.

## Tests

- 3 new Dusk tests in `SupportNavigate/BrowserTest.php`: no transition by default, `.transition` modifier fires one, global config transitions plain links and the back button.
- Full `SupportNavigate` browser suite passes (58 tests, 383 assertions).

## Docs

- `docs/navigate.md`: new "Animating page visits with view transitions" section with CSS customization examples and a duplicate-`view-transition-name` warning.
- `docs/wire-navigate.md`: `.transition` modifier section + reference table row.

Closes #6329

### DIFF
--- a/docs/navigate.md
+++ b/docs/navigate.md
@@ -277,6 +277,63 @@ In addition to `wire:navigate`, you can manually call the `Livewire.navigate()` 
 </script>
 ```
 
+## Animating page visits with view transitions
+
+By default, `wire:navigate` swaps in the new page instantly. If you'd like page visits to animate instead, Livewire supports the browser's [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API).
+
+To animate a single link's navigation, add the `.transition` modifier:
+
+```blade
+<a href="/posts" wire:navigate.transition>Posts</a>
+```
+
+Now, when this link is clicked, the browser will smoothly crossfade between the old and new page.
+
+If you'd like to enable transitions for every navigation in your application—including browser back and forward button visits—you can enable them globally in your application's `config/livewire.php` file:
+
+```php
+'navigate' => [
+    'transitions' => true,
+],
+```
+
+You can also trigger a transition when navigating programmatically:
+
+```js
+Livewire.navigate('/posts', { transition: true })
+```
+
+### Customizing transitions
+
+Because Livewire uses the native View Transitions API, you can customize animations entirely in CSS. For example, here's how you would swap the default crossfade for a quick slide-up effect:
+
+```css
+@keyframes slide-up {
+    from { transform: translateY(1rem); opacity: 0; }
+}
+
+::view-transition-new(root) {
+    animation: slide-up 0.2s ease-out;
+}
+```
+
+You can also "morph" a specific element from one page into an element on the next page by giving them both the same `view-transition-name`. For example, to animate a post's title from an index page onto its detail page:
+
+```blade
+<!-- /posts -->
+<h2 style="view-transition-name: post-title-{{ $post->id }}">{{ $post->title }}</h2>
+
+<!-- /posts/1 -->
+<h1 style="view-transition-name: post-title-{{ $post->id }}">{{ $post->title }}</h1>
+```
+
+Now, when a user navigates between these pages, the title will visually glide from its old position into its new one.
+
+> [!warning] View transition names must be unique
+> Like an `id` attribute, a `view-transition-name` can only appear once per page. If two elements on the same page share a name, the browser will skip the transition entirely. When rendering names inside loops, always include a unique identifier like `$post->id`.
+
+In browsers that don't support the View Transitions API, and for users who have enabled their operating system's reduced motion setting, Livewire will skip the animation and swap the page instantly.
+
 ## Using with analytics software
 
 When navigating pages using `wire:navigate` in your app, any `<script>` tags in the `<head>` only evaluate when the page is initially loaded.
--- a/docs/navigate.md
+++ b/docs/navigate.md
@@ -293,6 +293,8 @@ If you'd like to enable transitions for every navigation in your application—i
 
 ```php
 'navigate' => [
+    // ...
+
     'transitions' => true,
 ],
 ```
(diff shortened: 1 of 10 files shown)


docs/navigate.md
```diff
@@ -277,6 +277,63 @@ In addition to `wire:navigate`, you can manually call the `Livewire.navigate()`
...
+<a href="/posts" wire:navigate.transition>Posts</a>
+```
+
+Now, when this link is clicked, the browser will smoothly crossfade between the old and new page.
+
+If you'd like to enable transitions for every navigation in your application—including browser back and forward button visits—you can enable them globally in your application's `config/livewire.php` file:
+
+```php
+'navigate' => [
+    'transitions' => true,
```
calebporzio, line 298: should we have a `// ...` here to signify there are other configs in navigate?

docs/navigate.md
```diff
@@ -277,6 +277,63 @@ In addition to `wire:navigate`, you can manually call the `Livewire.navigate()`
...
+```js
+Livewire.navigate('/posts', { transition: true })
+```
+
+### Customizing transitions
+
+Because Livewire uses the native View Transitions API, you can customize animations entirely in CSS. For example, here's how you would swap the default crossfade for a quick slide-up effect:
+
+```css
+@keyframes slide-up {
```
calebporzio, line 313: Would this apply to every view transition on the page? it doesn't seem scoped to full page transitions. (see our `wire:transition` api)

docs/navigate.md
```diff
@@ -277,6 +277,63 @@ In addition to `wire:navigate`, you can manually call the `Livewire.navigate()`
...
+    from { transform: translateY(1rem); opacity: 0; }
+}
+
+::view-transition-new(root) {
+    animation: slide-up 0.2s ease-out;
+}
+```
+
+You can also "morph" a specific element from one page into an element on the next page by giving them both the same `view-transition-name`. For example, to animate a post's title from an index page onto its detail page:
```
calebporzio: We use `wire:transition="some-name"` in the `wire:transition` API to assign transition names so the API feels nice and clean.

This feels like a spritual divergence from that.

This feature and `wire:transition` are more intertwined in my head than in this PR.

js/plugins/navigate/index.js
```diff
@@ -6,6 +6,7 @@ import { restoreScrollPositionOrScrollToTop, storeScrollInformationInHtmlBeforeN
 import { isPersistedElement, putPersistantElementsBack, storePersistantElementsForLater } from "./persist"
```
calebporzio, line 6: I like the changes in this file. they all look good.

src/Features/SupportNavigate/BrowserTest.php
```diff
@@ -301,6 +313,74 @@ public function test_can_navigate_to_page_without_reloading()
         });
     }
 
+    public function test_wire_navigate_does_not_use_view_transitions_by_default()
+    {
+        $this->browse(function (Browser $browser) {
```
calebporzio, line 320: monkey patching startViewTransition ensures that the view transition api is getting hit.

This assures that code path gets hit but makes no actual assertions that it would work in the browser.

I'm sure its a tricky thing to test but if you could get the timing right and reliable you could assert css state mid-transition? idk... I'm fine if you deem it too flaky of a test to write, but i at least want you to be made aware i try to avoid tests like these. they don't test the real world. far less useful to the project.

docs/navigate.md
calebporzio: Added in 5559de4c.

docs/navigate.md
calebporzio: You're right — as written it targeted every view transition's root (in practice morphs pin the root with `!important` so it mostly only showed during navigations, but that's incidental, not intentional). Fixed properly in 5559de4c: page transitions now carry a `navigate` transition type, so the docs example scopes with `html:active-view-transition-type(navigate)::view-transition-new(root)`. Firefox only supports the bare-callback form of `startViewTransition()`, so it falls back to an untyped transition there (same try/catch pattern as `transitionDomMutation`) — scoped CSS just won't match in Firefox and it keeps the default crossfade.

docs/navigate.md
calebporzio: Agreed — unified in 5559de4c. Cross-page morphs now use `wire:transition="post-title-{{ $post->id }}"`, same API as morphs: `transitionPageSwap()` reuses `setTransitionNames()`/`clearTransitionNames()` from wire-transition.js to assign the underlying `view-transition-name` just-in-time on both sides of the swap and clear it when the animation finishes (no lingering stacking contexts). One deliberate constraint: only *named* `wire:transition` elements participate in page transitions — unnamed ones share the `match-element` default, which is safe inside one component's morph but would collide across two full pages and make the browser abort the whole transition. Docs updated to teach the wire:transition form.

src/Features/SupportNavigate/BrowserTest.php
calebporzio: Fair — added a real-behavior test in 5559de4c alongside the monkey-patch ones. It stretches the animation to 1.5s via CSS on the test pages, then asserts the browser is *genuinely mid-animation* using the `:active-view-transition` pseudo-class (which only matches while a real transition is running), that the incoming page's named `wire:transition` element received its `view-transition-name`, that unnamed ones didn't, and that names are cleared once `:active-view-transition` stops matching. The pseudo-class flips are what the browser itself reports, so it's not timing-guessy — `waitUntil` polls across a 1.5s window. Kept the monkey-patch tests for the opt-in semantics (proving a plain link *doesn't* transition is hard to assert positively any other way).

PR #10408, opened 2026-07-13
[4.x] Add the `$upload` action and `wire:drop` — paste, drop, and file-picker uploads

calebporzio:
## The problem

Building a ChatGPT-style prompt box — paste a screenshot while typing, drag a file anywhere on the page, click "Attach" to open the OS picker — currently requires hand-rolling all three gestures on top of a hidden `<input type="file">`: manual `dragover` cancellation, dragenter/dragleave counters to avoid overlay flicker, clipboard inspection, and `$refs.fileInput.click()` hacks.

`<input type="file" wire:model="...">` has been Livewire's only template-level entrance for files. This PR adds the rest.

## The API

One new magic action, usable inside any event listener:

```blade
<textarea wire:paste="$upload('photos')"></textarea>

<div wire:drop="$upload('photos')">Drop files here</div>

<button type="button" wire:click="$upload('photos')">Attach files</button>
```

`$upload` follows one rule: **it takes files from the argument, else from the event, else from the user.**

- A paste or drop that carries files → those files upload into the property
- A file-capable event *without* files (a plain text paste) → no-op; the browser's default proceeds untouched
- An event that *can't* carry files (a click) → the OS file picker opens

And one new directive, `wire:drop`, because a bare `drop` listener never fires unless `dragover` is cancelled — a plain wildcard listener physically can't power a dropzone. `wire:drop` owns the whole drag lifecycle:

- Only engages with drags carrying files (`DataTransfer.types` includes `Files`) — dragging selected text does nothing
- Applies `data-dragging` to the element while files are dragged over it, so overlays are pure CSS (same move as `data-loading`)
- Handles the dragenter/dragleave depth-counting so the attribute doesn't flicker over child elements
- `.window` accepts drops anywhere on the page (Alpine's `.window` semantics), with `data-dragging` still landing on the directive's element as a styling anchor

A complete ChatGPT-style prompt box needs zero custom JavaScript and no hidden inputs:

```blade
<div class="group" wire:drop.window="$upload('attachments')">
    <div class="hidden group-data-dragging:grid fixed inset-0 place-items-center bg-black/50">
        Drop files to attach
    </div>

    <form wire:submit="send">
        <template x-for="file in $wire.attachments" :key="file.name">
            <div>
                <img x-show="file.isPreviewable" :src="file.previewUrl">
                <progress x-show="file.isUploading" :value="file.progress" max="100"></progress>
                <button type="button" x-on:click="file.remove()">&times;</button>
            </div>
        </template>

        <textarea wire:model="message" wire:paste="$upload('attachments')"></textarea>

        <button type="button" wire:click="$upload('attachments', { accept: 'image/*,.pdf' })">
            Add photos & files
        </button>

        <button type="submit" x-bind:disabled="$wire.attachments.some(f => f.isUploading)">Send</button>
    </form>
</div>
```

### Options

`$upload(property, filesOrEvent?, { accept, multiple, append })`

- `accept` — filters pasted/dropped files and constrains the picker, matching native `accept` semantics (client-side convenience only; server validation remains the gate)
- `multiple` — defaults to whether the property currently holds an array
- `append` — multiple uploads append (default) or replace

### JS door

`$wire.$upload()` is modernized to return a promise (resolves on finish, `null` on cancel/dismiss, rejects on failure) and accepts a File/FileList/array, an Event to pull files from, or an options object. **The v3 callback signature still works unchanged** — detected by a function in the third argument.

## Design notes

Walked through `api-design.md`:

- **No new vocabulary was minted for the gestures.** `wire:paste` and `wire:click` are the existing wildcard/event directives; `$upload` joins the existing magic-action family (`$set`, `$toggle`, `$dispatch`) and mirrors the existing `$wire.$upload()` — the word "upload" already belongs to this capability through every door (`WithFileUploads`, `livewire-upload-*` events, the docs page).
- **`wire:drop` is claimed from the wildcard** because the wildcard version was inert (no `dragover` cancellation → the browser never fires `drop`). Anyone with an existing `wire:drop="handler"` gets a strict upgrade: their handler now actually fires for file drops, with `$event` in scope, plus `data-dragging` for free.
- **Event inheritance** uses `window.event`, which the DOM spec sets synchronously during any dispatch — so `$upload` works identically inside `wire:paste`, `x-on:paste`, and any addEventListener callback, with no evaluator or wildcard changes. (Consequence: an async wrapper like `.debounce` loses the event context and falls through to the picker — documented trade-off, and not a meaningful combination for paste/drop anyway.)
- **Zero new server surface.** Every gesture funnels into the existing upload manager, so signed URLs, throttling, chunked/resumable transport, S3 multipart, validation, `wire:loading` targeting, and the rich upload synth objects all compose with no new threat model.

## Verification

- **9 new browser tests** (`UploadActionBrowserTest`) covering: paste upload, text-paste non-hijack (asserts `defaultPrevented === false`), drop upload + `data-dragging` lifecycle, text-drag ignore, `.window` drops landing state on the directive element, picker flow (CDP file-chooser interception; asserts the input is removed after selection), `accept` filtering, single-property first-file semantics, and the legacy callback signature.
- Full `SupportFileUploads` browser suite (26 tests), upload unit suite (99 tests), and JS specs (126) all green.
- Exercised end-to-end in a real Laravel app: a ChatGPT-style chat page driven by Playwright through real Chrome — synthetic clipboard/drag events, a **real trusted click opening a real file chooser**, chip previews from blob URLs, optimistic remove, and send round-trip with server-side original filenames. 12/12 scenarios passed.

### DIFF
--- a/docs/uploads.md
+++ b/docs/uploads.md
@@ -294,23 +294,23 @@ new class extends Component {
 ```
 
 ```blade
-<div class="group relative" wire:drop.window="$upload('attachments')">
+<div class="relative" wire:drop.window="$upload('attachments')">
     {{-- A full-screen overlay, shown while files are dragged over the page... --}}
-    <div class="hidden group-data-dragging:grid fixed inset-0 place-items-center bg-black/50 text-white">
+    <div class="hidden in-data-dragging:grid fixed inset-0 place-items-center bg-black/50 text-white">
         Drop files to attach
     </div>
 
     <form wire:submit="send">
         {{-- Pending and finished attachments, powered by rich upload objects... --}}
-        <template x-for="file in $wire.attachments" :key="file.name">
+        <template wire:for="file in attachments" wire:for:key="file.name">
             <div>
-                <img x-show="file.isPreviewable" x-bind:src="file.previewUrl">
+                <img wire:show="file.isPreviewable" wire:bind:src="file.previewUrl">
 
-                <span x-text="file.name"></span>
+                <span wire:text="file.name"></span>
 
-                <progress x-show="file.isUploading" x-bind:value="file.progress" max="100"></progress>
+                <progress wire:show="file.isUploading" wire:bind:value="file.progress" max="100"></progress>
 
-                <button type="button" x-on:click="file.remove()">&times;</button>
+                <button type="button" wire:click="file.remove()">&times;</button>
             </div>
         </template>
 
@@ -320,7 +320,7 @@ new class extends Component {
             Add photos & files
         </button>
 
-        <button type="submit" x-bind:disabled="$wire.attachments.some(file => file.isUploading)">
+        <button type="submit" wire:bind:disabled="attachments.some(file => file.isUploading)">
             Send
         </button>
     </form>
@@ -693,7 +693,15 @@ These functions exist on a JavaScript component object, which can be accessed us
 </script>
 ```
 
-`$wire.$upload()` returns a promise that resolves when the upload finishes (or with `null` if the picker is dismissed or the upload is cancelled) and rejects if it fails. For upload state while in flight — progress, previews — read the property's [rich upload objects](#rich-upload-objects-in-javascript) reactively.
+`$wire.$upload()` returns a promise that resolves with the [rich upload object](#rich-upload-objects-in-javascript) now living on the property — or an array of them when uploading multiple files (just the files from this upload, not the property's previously uploaded ones). If the picker is dismissed or the upload is cancelled it resolves with `null`, and if the upload fails it rejects:
+
+```blade
+<script>
+    let photo = await $wire.$upload('photo')
+
+    photo.previewUrl // Instant local preview...
+</script>
+```
 
 The legacy callback signature from earlier versions of Livewire continues to work:
 
--- a/docs/uploads.md
+++ b/docs/uploads.md
@@ -247,6 +247,119 @@ Properties holding multiple uploads hydrate into arrays of rich objects, so each
 
 Rich upload objects are powered by [JavaScript synthesizers](/docs/synthesizers#javascript-synthesizers). When sent back to the server or stringified via `JSON.stringify()`, they degrade to their raw wire value automatically.
 
+## Uploading beyond file inputs
+
+Modern interfaces accept files from more than a file input: users paste screenshots into a message box, drag files onto the page, and click "Attach" buttons that open the system's file picker.
+
+Livewire supports all of these through a single action: `$upload`. Use it inside any event listener — most commonly `wire:paste`, [`wire:drop`](/docs/wire-drop), and `wire:click`:
+
+```blade
+<textarea wire:paste="$upload('photos')"></textarea>
+
+<div wire:drop="$upload('photos')">Drop files here</div>
+
+<button type="button" wire:click="$upload('photos')">Attach files</button>
+```
+
+`$upload` follows one rule: **it takes files from the argument, else from the event, else from the user.**
+
+* When the triggering event carries files — pasted screenshots, dropped files — those files upload into the property
+* When the event *can* carry files but doesn't — a plain text paste — nothing happens, and the browser's default behavior proceeds untouched
+* When the event *can't* carry files — a click — the browser's file picker opens and the user's selection uploads instead
+
+Uploads started this way are ordinary Livewire uploads: they flow through the same temporary-upload pipeline as `wire:model` file inputs, respect your validation rules, and hydrate into [rich upload objects](#rich-upload-objects-in-javascript) for instant previews and progress.
+
+Here's a complete chat-style message box that accepts attachments from all three gestures:
+
+```php
+<?php // resources/views/components/⚡message-box.blade.php
+
+use Livewire\Attributes\Validate;
+use Livewire\WithFileUploads;
+use Livewire\Component;
+
+new class extends Component {
+    use WithFileUploads;
+
+    public $message = '';
+
+    #[Validate(['attachments.*' => 'file|max:10240'])]
+    public $attachments = [];
+
+    public function send()
+    {
+        // ...
+    }
+};
+```
+
+```blade
+<div class="group relative" wire:drop.window="$upload('attachments')">
+    {{-- A full-screen overlay, shown while files are dragged over the page... --}}
+    <div class="hidden group-data-dragging:grid fixed inset-0 place-items-center bg-black/50 text-white">
+        Drop files to attach
+    </div>
+
+    <form wire:submit="send">
+        {{-- Pending and finished attachments, powered by rich upload objects... --}}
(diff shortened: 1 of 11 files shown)


docs/uploads.md
```diff
@@ -247,6 +247,119 @@ Properties holding multiple uploads hydrate into arrays of rich objects, so each
...
+```blade
+<div class="group relative" wire:drop.window="$upload('attachments')">
+    {{-- A full-screen overlay, shown while files are dragged over the page... --}}
+    <div class="hidden group-data-dragging:grid fixed inset-0 place-items-center bg-black/50 text-white">
+        Drop files to attach
+    </div>
+
+    <form wire:submit="send">
+        {{-- Pending and finished attachments, powered by rich upload objects... --}}
+        <template x-for="file in $wire.attachments" :key="file.name">
```
calebporzio: we can use the new wire:for stuff

docs/uploads.md
```diff
@@ -247,6 +247,119 @@ Properties holding multiple uploads hydrate into arrays of rich objects, so each
...
+    {
+        // ...
+    }
+};
+```
+
+```blade
+<div class="group relative" wire:drop.window="$upload('attachments')">
+    {{-- A full-screen overlay, shown while files are dragged over the page... --}}
+    <div class="hidden group-data-dragging:grid fixed inset-0 place-items-center bg-black/50 text-white">
```
calebporzio: can use TW `in-data-...` no need for group

docs/uploads.md
```diff
@@ -247,6 +247,119 @@ Properties holding multiple uploads hydrate into arrays of rich objects, so each
...
+            </div>
+        </template>
+
+        <textarea wire:model="message" wire:paste="$upload('attachments')"></textarea>
+
+        <button type="button" wire:click="$upload('attachments', { accept: 'image/*,.pdf' })">
+            Add photos & files
+        </button>
+
+        <button type="submit" x-bind:disabled="$wire.attachments.some(file => file.isUploading)">
```
calebporzio: wire:bind

docs/uploads.md
```diff
@@ -560,10 +673,33 @@ These functions exist on a JavaScript component object, which can be accessed us
...
+
+    // Remove single file from multiple uploaded files...
+    $wire.$removeUpload('photos', uploadedFilename)
+
+    // Cancel an upload...
+    $wire.$cancelUpload('photos')
+</script>
+```
+
+`$wire.$upload()` returns a promise that resolves when the upload finishes (or with `null` if the picker is dismissed or the upload is cancelled) and rejects if it fails. For upload state while in flight — progress, previews — read the property's [rich upload objects](#rich-upload-objects-in-javascript) reactively.
```
calebporzio: think we could return the reactive object representing the upload? would feel nice but have questions around multiple and if array should return

src/Features/SupportFileUploads/UploadActionBrowserTest.php
```diff
@@ -0,0 +1,329 @@
...
+
+            public $photo;
+
+            function render() { return <<<'HTML'
+            <div>
+                <div dusk="zone" wire:drop="$upload('photo')">Drop a file here</div>
+
+                <span dusk="name" x-text="$wire.photo?.name"></span>
+            </div>
+            HTML; }
```
calebporzio, line 241: all these tests feel super not real life and dont give me confidence. they do too much manual JS to simulate real things

docs/uploads.md
calebporzio: Done in f7d54a5c — the chat example now uses `wire:for="file in attachments"` + `wire:for:key="file.name"`, with `wire:show`/`wire:text`/`wire:bind` on the chips instead of the `x-*` equivalents. Also dogfooded the exact same markup in the Upload Lab chat page and re-verified end-to-end (which is how the arrow-function evaluator bug below surfaced).

docs/uploads.md
calebporzio: Done — `hidden in-data-dragging:grid` with no `group` class needed, in both the uploads.md example and the two wire-drop.md examples. Verified the variant works against the real overlay in the Upload Lab.

docs/uploads.md
calebporzio: Done — `wire:bind:disabled="attachments.some(file => file.isUploading)"`. This uncovered a real pre-existing bug: `contextualizeExpression` was prefixing arrow-function params (`file` → `$wire.file`), producing a syntax error for any `wire:bind`/`wire:text`/`wire:show` expression containing an arrow. Fixed in 4d269ea9 with vitest coverage (incl. multi-param and destructured params), and verified live that the send button actually disables mid-upload and re-enables after.

docs/uploads.md
calebporzio: Done in f7d54a5c. On the multiple question, I went with **mirroring the shape of the upload**: a single upload resolves the property's rich object, a `multiple` upload resolves an array holding **just this batch** — not the property's previously uploaded files. Rationale: you awaited *this* call, so you get *these* uploads; the full list is always available reactively on `$wire.photos` anyway.

```js
let photo = await $wire.$upload('photo')      // TemporaryUpload
let batch = await $wire.$upload('photos', files) // TemporaryUpload[] (this batch only)
```

Picker dismissed / upload cancelled still resolves `null`. One nuance worth knowing: by resolve time the server round-trip has completed, so the object you get is the hydrated instance — `previewUrl` is the signed server URL unless a local blob URL was already created during flight. Covered in `test_upload_promise_resolves_with_rich_upload_objects`.

src/Features/SupportFileUploads/UploadActionBrowserTest.php
calebporzio: Totally fair — rewrote the whole file in f7d54a5c around real browser input. What changed:

- **Paste**: a real PNG is written to the **actual clipboard** (CDP `Browser.grantPermissions` + `navigator.clipboard.write`), then delivered by a **real ⌘V/Ctrl+V keystroke** — the browser generates the trusted paste event itself. The text-paste test now asserts the text *actually lands in the textarea* (the browser default really ran), not a `defaultPrevented` flag.
- **Drop**: CDP `Input.dispatchDragEvent` with `files: [realPathOnDisk]` — the browser synthesizes the trusted dragenter/dragover/dragleave stream from pointer coordinates exactly like an OS file drag. The drag-away test moves the pointer to another element and the *browser* fires the real dragleave.
- **Picker**: real click → real chooser (CDP-intercepted only because headless Chrome auto-dismisses dialogs) → selection delivered via `DOM.setFileInputFiles` with a real file from disk, which fires the trusted change event — same mechanism Puppeteer/Playwright use for uploads.

The only remaining `new File(...)` construction is in the two JS-door tests (`$wire.$upload(name, file)` / legacy callbacks), where constructing a File **is** the real-life usage. All 10 green, plus the full uploads suite.

PR #10414, opened 2026-07-16
[4.x] Fix `wire:loading.delay` remaining visible after concurrent requests

people-sea:

Yes. I did not open a discussion because this is a small bug fix.

Yes.

No.

Yes.

When concurrent requests share the same delayed loading state, each request creates a timeout, but only the latest timeout can be cleared. An earlier timeout may still run after both requests finish, leaving the loading element visible.

This keeps track of the active requests so the timeout is started once and cleared after the last request finishes.

https://github.com/user-attachments/assets/aa305c79-71c9-4fa6-9ac4-47d7451034f1

https://github.com/user-attachments/assets/dffa956d-6e41-455a-8a7c-60fd35eda922

### DIFF
--- a/js/directives/wire-loading.js
+++ b/js/directives/wire-loading.js
@@ -48,11 +48,16 @@ function applyDelay(directive) {
         }
     })
 
+    let active = 0
     let timeout
     let started = false
 
     return [
         (callback) => { // Initiate delay...
+            active++
+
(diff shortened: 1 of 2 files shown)


people-sea:
Superseded by #10340, the applyDelay rewrite there covers this. My failing test passes on that branch, so closing.

PR #10416, opened 2026-07-16
[4.x] Add an opt-in delta update engine for large components

DanielSpravtsev:

Opening as a draft for feedback — take a look and let's see what makes sense to do with it @calebporzio @joshhanley 

`feature/stateful-delta-engine`

No. It's one feature, plus its tests and docs.

Yes — PHP unit tests, a browser test, and JS tests covering the diff/patch round-trip, integrity checks, the fallback paths, and the size thresholds.

## The idea

Morphing already does the smart thing on the client: it only touches the DOM nodes that actually changed. But to get there, the browser still downloads the component's *entire* rendered HTML on every update. On a 200-row table where a single status cell flips, that's the whole table crossing the wire to deliver a few bytes of real change.

This PR adds a `delta` update engine that closes that gap. The server hangs on to the previous render, diffs the new one against it, and sends only the bytes that changed as a short list of byte-offset patches. The browser reconstructs the exact HTML and feeds it to the same morph pipeline it already uses. PHP still renders the component every time — the only thing that changes is how much of the result crosses the wire.

It's opt-in and off by default:

```php
// config/livewire.php

'update_engine' => 'delta', // default: 'morph'
```

Existing components and Blade templates don't change, and with the default `morph` setting nothing about Livewire's behavior moves at all.

## How it works

The first eligible update sends full HTML and establishes an exact, byte-for-byte baseline:

```json
{ "html": "<div>...</div>", "htmlHash": "8aef..." }
```

A later update can come back as patches instead:

```json
{
    "htmlDelta": {
        "base": "8aef...",
        "patches": [
            { "start": 13842, "delete": 7, "insert": "cnVubmluZw==" }
        ]
    },
    "htmlHash": "b391..."
}
```

On the browser side it checks that `base` matches the render it's currently showing, applies the patches by byte offset, hashes the result with SHA-256, and only morphs once that hash matches `htmlHash`. Because the reconstructed HTML flows through the normal morph, `wire:key`, focus handling, nested components, and morph hooks all keep working untouched. If anything doesn't line up, it throws the baseline away and quietly asks for a full render.

## It only kicks in when it's worth it

The engine doesn't try to delta everything — that would just add overhead to small components. Anything under 8 KiB stays on plain morph and never touches the render store or adds delta metadata to the response.

Above that, a delta actually has to earn its place: it needs to save at least 10% of the payload, and — because responses are usually gzipped on the wire — it has to still win after a fast gzip estimate and clear a 1 KiB absolute floor. Otherwise Livewire just sends full HTML. The defaults:

```php
'delta' => [
    'store' => null,   // your default cache store
    'ttl' => 300,
    'minimum_html_bytes' => 8192,
    'minimum_savings' => 0.1,
    'minimum_compressed_savings_bytes' => 1024,
    'compression_aware' => true,
],
```

Components can cross that line either way. Grow past the threshold and one full response re-seeds a baseline; shrink back under it and things drop back to plain morph.

## More than one patch at a time

A single prefix/suffix diff falls apart the moment two far-apart regions change — think of moving a Kanban card, where you delete it from one column and insert it into another. So `HtmlDelta` can emit several non-overlapping patches that all reference the same original render and apply atomically:

```json
"patches": [
    { "start": 120,  "delete": 86, "insert": "" },
    { "start": 2980, "delete": 0,  "insert": "PGFydGljbGU+..." }
]
```

Anchor discovery has fixed limits on scan steps and patch count. If a diff stops being productive, the encoder falls back to a single replacement patch and lets the normal size checks decide whether full HTML is the better deal anyway — so worst-case work stays bounded.

## Where the previous render lives

To build a delta the server needs the render the browser is currently showing, so after the first update it keeps one render per active component in Laravel's cache:

```php
'delta' => [
    'store' => 'redis',
    'ttl' => 300,
],
```

A file cache is fine on a single server. Once updates for the same component can land on different servers you'll want a shared store like Redis. Only the latest eligible render is kept, and expiry, a missing entry, a raced revision, or an unavailable store all just fall back to full HTML.

## Integrity and security

Delta updates change how HTML is shipped, not how requests are authorized. The browser never sends rendered HTML or patches for the server to apply — only the hash of the render it currently holds, and that's only trusted when it matches the hash inside Livewire's checksum-protected snapshot. The render gets SHA-256-verified on both ends: the server re-hashes the cached copy before diffing, and the browser re-hashes the reconstructed HTML before morphing. A tampered or corrupted patch never reaches the DOM — it triggers a full resync instead.

> **Worth calling out:** the cache holds raw rendered HTML, which can contain user-specific data. The docs cover this — keep the store on a private network with an app-specific prefix, use Redis ACLs and TLS across trust boundaries, and keep the TTL short.

## Benchmarks

These cover the render effect only — the snapshot is identical on both transports, so it's excluded.

### Payload size

| Scenario | Rendered | Morph raw | Delta raw | Morph gzip | Delta gzip | Gzip saved |
|---|---:|---:|---:|---:|---:|---:|
| Todo toggle | 415 B | 460 B | 273 B | 232 B | 240 B | −3.4% |
| Kanban move | 3,616 B | 3,866 B | 440 B | 599 B | 341 B | 43.1% |
| 200-row table status | 35,828 B | 38,444 B | 235 B | 3,410 B | 205 B | 94.0% |

With the default thresholds the Todo and the 3.6 KiB Kanban components both stay on morph — the win is too small to bother with — and only the 35.8 KiB table actually uses a delta. That's the whole point of the size gate: don't spend cache and hashing time chasing savings that don't matter.

### Server cost (file cache)

| Scenario | Morph p50 | Delta + file p50 | Extra |
|---|---:|---:|---:|
| Todo, forced delta | 0.004 ms | 0.611 ms | +0.607 ms |
| Kanban, forced delta | 0.018 ms | 0.826 ms | +0.808 ms |
| 200-row table | 0.165 ms | 1.901 ms | +1.736 ms |

For the table, the delta cuts about 3.2 KB off the gzipped response — roughly 5 ms of transfer on a 5 Mbps link, so after the server and client overhead it comes out about 2.9 ms ahead in this local run. At high throughput the CPU can cost more than the transfer it saves, which is exactly why this is opt-in, hybrid, and compression-aware rather than the default.

> **On the numbers:** the byte measurements are deterministic and I'd stand behind them. The CPU timings were taken with PHP 8.4 WASM under Node against the host filesystem (warm cache, `LOCK_EX` writes, no `fsync`), so treat them as directional — they should really be rerun on native PHP-FPM/Octane with production storage and real components.

## Compatibility and rollback

No component changes are needed, and everything you'd expect keeps working: actions, Blade rendering, validation and authorization, redirects and events, renderless actions, nested components, `wire:key`, the existing morph, and the normal snapshot/hydration model. The reconstruction path also accepts the original single-patch shape, so cached test state and rolling deployments stay safe.

To turn it off, flip the setting back:

```php
'update_engine' => 'morph',
```

There's no migration and no persistent application state to unwind. If an interaction doesn't need a server render at all, Alpine state or a renderless action is still the right tool — the delta engine is for when you *do* need to render but most of the output stays put.

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -216,6 +216,33 @@ return [
 
     /*
     |---------------------------------------------------------------------------
+    | Update Engine
+    |---------------------------------------------------------------------------
+    |
+    | The default "morph" engine returns the component's full rendered HTML
+    | after each update. The "delta" engine caches eligible renders and sends
+    | only the bytes that changed, as long as the render is big enough and the
+    | saving is worth it. It verifies a SHA-256 hash on both ends and falls back
+    | to full HTML whenever a baseline can't be trusted.
+    |
+    */
+
+    'update_engine' => env('LIVEWIRE_UPDATE_ENGINE', 'morph'),
+
+    'delta' => [
+        'store' => env('LIVEWIRE_DELTA_STORE'),
+        'ttl' => env('LIVEWIRE_DELTA_TTL', 300),
+        'minimum_html_bytes' => env('LIVEWIRE_DELTA_MINIMUM_HTML_BYTES', 8192),
+        'minimum_savings' => env('LIVEWIRE_DELTA_MINIMUM_SAVINGS', 0.1),
+        'minimum_compressed_savings_bytes' => env(
+            'LIVEWIRE_DELTA_MINIMUM_COMPRESSED_SAVINGS_BYTES',
+            1024,
+        ),
+        'compression_aware' => env('LIVEWIRE_DELTA_COMPRESSION_AWARE', true),
+    ],
+
+    /*
+    |---------------------------------------------------------------------------
     | Smart Wire Keys
     |---------------------------------------------------------------------------
     |
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -220,10 +220,10 @@ return [
     |---------------------------------------------------------------------------
     |
     | The default "morph" engine returns the component's full rendered HTML
-    | after each update. The "delta" engine caches eligible renders and sends
-    | only the bytes that changed, as long as the render is big enough and the
-    | saving is worth it. It verifies a SHA-256 hash on both ends and falls back
-    | to full HTML whenever a baseline can't be trusted.
+    | after each update. The experimental "delta" transport negotiates exact,
+    | stateless chunk and explicit fragment updates, with an optional cache
+    | accelerator. Every reconstructed snapshot and render is verified with
+    | SHA-256 and falls back safely whenever optional state is unavailable.
     |
     */
 
@@ -232,6 +232,7 @@ return [
     'delta' => [
         'store' => env('LIVEWIRE_DELTA_STORE'),
         'ttl' => env('LIVEWIRE_DELTA_TTL', 300),
+        'cache_accelerator' => env('LIVEWIRE_DELTA_CACHE_ACCELERATOR', true),
         'minimum_html_bytes' => env('LIVEWIRE_DELTA_MINIMUM_HTML_BYTES', 8192),
         'minimum_savings' => env('LIVEWIRE_DELTA_MINIMUM_SAVINGS', 0.1),
         'minimum_compressed_savings_bytes' => env(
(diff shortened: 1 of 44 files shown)


ghabriel25:
@DanielSpravtsev I think this is really a good improvement, however it seems too broad to implement this at component level? Maybe scale it down to specific things like for page that contains table. The API could be `wire:table` or something like that

DanielSpravtsev:
> @DanielSpravtsev I think this is really a good improvement, however it seems too broad to implement this at component level? Maybe scale it down to specific things like for page that contains table. The API could be `wire:table` or something like that

Thanks — I understand the concern about the scope.
At this point, this PR is closer to an experiment and a discussion starter than a finalized API. I’m exploring a broader problem in Livewire’s current transport: the full snapshot travels between the browser and server, while the server also returns the full rendered HTML even when only a few bytes have changed.

I’m now thinking about whether this should become a more global transport-level solution — potentially treating both the snapshot and rendered HTML as revisions that can be transferred as deltas, while always retaining a safe full-payload fallback. If I arrive at a solid design, I may substantially rework this PR around that broader approach.
I don’t think wire:table would be the right abstraction, though. The problem isn’t specific to tables — the same issue exists with large lists, Kanban boards, dashboards, trees, feeds, and other mostly-stable components. Tying the optimization to a particular HTML/UI concept would make the API narrow and require developers to manually identify cases that the transport layer could potentially handle automatically.

If explicit scoping is needed, I think a generic component- or fragment-level opt-in would make more sense. Otherwise, an adaptive global transport could decide automatically based on the rendered size and the actual savings. Tables would then benefit naturally without Livewire needing a table-specific primitive.
So for now I’m treating this PR as an exploration of the idea, and I’m open to reshaping it if a cleaner global approach emerges. 

Thanks for the feedback!

joshhanley:
@DanielSpravtsev thanks for the PR! We recently created a `4.x` branch for ongoing Livewire v4 development. Because this PR was based on `main`, we’re going to close it. Could you please rebase or cherry-pick your changes onto `4.x` and open a new PR targeting `4.x`? Thanks!

PR #10418, opened 2026-07-17
[4.x] Fix authorize attribute bypass exception hook

### LINKED ISSUE #10410: [4.x] `#[Authorize]` attribute doesnt trigger component `exception` hook
### Livewire version

v4.3

### Laravel version

v13.2

### PHP version

8.4

Chrome on Windows

When using `#[Authorize]` attribute, the component `exception()` hook not triggered.

```php
Gate::define('interact', fn () => false);
```
```php
<?php

use Livewire\Component;

new class extends Component
{
    public function saveWithExceptionHook()
    {
        $this->authorize('interact');
    }

    #[Authorize('interact')]
    public function saveWithoutExceptionHook()
    {
        //
    }

    {
        $stopPropagation();

        dd($e); // only triggered when saveWithExceptionHook() called
    }
}
```

N/A

It should work for both cases.

ghabriel25:
Based on https://github.com/livewire/livewire/pull/10422. This PR provide more simple fix rather than complex method override and trait aliasing.

> [!NOTE]
> Please give a little time to compare these 2 PR as both provide different approach. This PR doesnt cover `ModelNotFoundException` like the other PR does.
> If one the these PR is more suitable for the fix, feel free to open/close the PR, thanks!
> 
> I apologize for the unconvenience, best regards.

# Scenario
When using `#[Authorize]` attribute to perform authorization as pre-method execution. It doesnt trigger the component `exception()` hook.

```php
<?php

use App\Models\Post;
use Livewire\Component;

new class extends Component
{
    #[Authorize('edit', 'post')]
    public function save(Post $post)
    {
        //
    }

    {
        $stopPropagation();

        dd($e->getMessage()); // not triggered
    }
}
```

# Problem
Attribute `call()` handled by `SupportAttributes::call()` which is running before the actual method get called so any exception thrown will not get handled by component exception hook.

# Solution
Back to use `$this->component->authorize()` with a wrapper. Now `HandlesAuthorization` trait only override `parseAbilityAndArguments()` to provide the right method name either from attribute or backtrace. This way, any exception thrown can be catched from component `exception()` hook.

# Files Changed
- New `HandlesAuthorization` trait
- Move `AuthorizesRequests` trait from `Component` to `HandlesAuthorization` trait
- Override `parseAbilityAndArguments()` but add fallback method name from backtrace to ensure it resolve the correct method name
- Added regression tests to make sure both works as expected either using `#[Authorize]` attribute or `$this->authorize()`

Fix: https://github.com/livewire/livewire/issues/10410

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -20,14 +20,13 @@ use Livewire\Features\SupportRenderless\HandlesRenderless;
 use Livewire\Exceptions\PropertyNotFoundException;
 use Livewire\Concerns\InteractsWithProperties;
 use Illuminate\Support\Traits\Macroable;
-use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 use BadMethodCallException;
+use Livewire\Features\SupportAuthorization\HandlesAuthorization;
 
 abstract class Component
 {
     use Macroable { __call as macroCall; }
 
-    use AuthorizesRequests;
     use InteractsWithProperties;
     use HandlesEvents;
     use HandlesIslands;
@@ -44,6 +43,7 @@ abstract class Component
     use HandlesSlots;
     use HandlesHtmlAttributeForwarding;
     use HandlesRenderless;
+    use HandlesAuthorization;
 
     protected $__id;
     protected $__name;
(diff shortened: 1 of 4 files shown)


ghabriel25:
Closing in favor https://github.com/livewire/livewire/pull/10422

PR #10422, opened 2026-07-18
[4.x] Fix `#[Authorize]` attribute doesnt trigger component exception hook

### LINKED ISSUE #10410: [4.x] `#[Authorize]` attribute doesnt trigger component `exception` hook

ghabriel25:
# Scenario
When using `#[Authorize]` attribute to perform authorization as pre-method execution. It doesnt trigger the component `exception()` hook.
```php
class PostPolicy
{
    public function edit(User $user, Post $post)
    {
        return false;
    }
}
```

```php
<?php

use App\Models\Post;
use Livewire\Component;

new class extends Component
{
    #[Authorize('edit', 'post')]
    public function save(Post $post)
    {
        //
    }

    {
        $stopPropagation();

        dd($e->getMessage()); // not triggered
    }
}
```

# Problem
Attribute `call()` handled by `SupportAttributes::call()` which is running before the actual method get called so any exception thrown will not get handled by component exception hook.

# Solution
Delegate authorization to `HandlesAuthorization` trait and inside `BaseAuthorize::call()`, wrap the component before performing the authorization from that trait. This way, any exception thrown can be catched from component `exception()` hook. All method parameters resolution handled just like before.

# Whats Changed
- New `HandlesAuthorization` trait
- Move `AuthorizesRequests` trait from `Component` to `HandlesAuthorization` trait
- Added regression tests to make sure both works as expected either using `#[Authorize]` attribute or `$this->authorize()`

Fix: https://github.com/livewire/livewire/issues/10410


ghabriel25:
@calebporzio @joshhanley Hi! Could you please review this PR and let me know if the unit tests need any cleanup?

ghabriel25:
Going to close this and recreate new pr instead chery picking the commits

joshhanley:
@ghabriel25 yep I couldn't find a clean way to do this one either.

PR #10423, opened 2026-07-18
Add Selection — a first-class primitive for selecting rows

calebporzio:
## Selection — a first-class primitive for selecting rows

> Work-in-progress: shaping out a selection utility for tables/lists. This PR is the minimal core so the diff can be reviewed as it grows.

### What it gives you

```blade
<?php

use Livewire\Selection;

new class extends Component
{
    public Selection $selection; // auto-initialized, no mount() needed

    public function deleteSelected()
    {
        Transaction::whereIn('id', $this->selection->all())->delete();

        $this->selection->clear();
    }
};
?>

<template wire:if="selection.any()">
    <button wire:click="selection.clear()">Clear selection</button>
    <button wire:click="deleteSelected">Delete selected</button>
</template>

@foreach ($this->transactions as $transaction)
    <input type="checkbox" wire:model="selection" value="{{ $transaction->id }}" />
@endforeach
```

- `wire:model="selection"` on checkboxes toggles key membership natively
- `selection.clear()` / `selection.any()` / `selection.count()` etc. work client-side in any directive expression (instant, no request — syncs like a deferred model update)
- The same API exists server-side: `all()`, `any()`, `isEmpty()`, `count()`, `contains()`, `select()`, `deselect()`, `clear()`

### How it works

**Backend** (`src/Features/SupportSelection/`): a `Selection` class (surfaced as `Livewire\Selection`, mirroring `Livewire\Form`) and a `SelectionSynth` (key: `sel`) that dehydrates to a plain keys array + class meta. Typed `public Selection $selection` properties are auto-initialized by the feature hook, like form objects.

**Frontend** (`js/features/supportSelection.js`): a built-in JS synth registered through the existing `Livewire.synth` plumbing. It hydrates the wire value into a `Selection extends Array` — because it's a real array:

- Alpine's checkbox `x-model` treats it as a list and toggles membership natively (`Symbol.species` keeps `concat`/`filter` results as `Selection`)
- `wire:if="selection.any()"` is reactive for free
- the existing atomic synth diffing handles client→server sync unchanged

Key comparisons are loose (`==`) on both sides, since checkbox values arrive as strings while server keys are usually ints.

### Deliberately not in this PR

- **Dual-mode selection** — "select all results" across a paginated set with an *except* list. The keys array is kept behind `Selection`'s API (never accessed as a raw array) so a mode/except pair can be added without touching call sites.
- Public docs, `$wire.$selection`-style magics, Flux integration.

### Tests

- `UnitTest.php` — auto-init, client set → hydrated `Selection`, action-side API, synth key/class meta, loose comparison, hydration class validation
- `BrowserTest.php` — checkbox `wire:model` + server sync, `selection.clear()`/`select()` from directive expressions with `wire:if` reactivity, server-initialized + server-mutated selections reaching checkboxes

---

## Update: synth `bind` contract

The client-side implementation now rests on a new **general synth capability** rather than selection-specific machinery: a JS synth may declare `bind()` — its contract for how elements `wire:model` to its rich values.

```js
Livewire.synth('sel', {
    match, hydrate, dehydrate,

    bind({ el, component, path, modifiers, get, set, notify, cleanup }) {
        if (! isCheckbox(el)) return false // decline → default x-model handling

        Alpine.bind(el, {
            ['x-effect']() { el.checked = get().contains(el.value) },
            ['@change']() { get().toggle(el.value); notify() },
        })
    },
})
```

`wire:model` delegates element wiring to the synth while keeping network timing (`.live`, triggers, debounce) for itself — mirroring its existing file-input special case, generalized into a protocol. This will matter for other rich synth types beyond Selection.

What it bought Selection immediately:

- **Alpine's array `x-model` semantics are out of the loop** — toggling mutates the selection in place, instances are never replaced, and the `concat`/`filter` overrides are gone.
- **`selectPage()` reads a binding registry, not the DOM** — elements register per component/path as they bind (with directive cleanup), so teleports, islands, and morphs are all covered structurally.
- **Dual-mode stays open** — checked-state is derived by the synth (`contains`), so an all/except mode later just changes that derivation.

Covered by a new public-contract browser test in `SupportJsSynthesizers` (a userland synth binding a `Date` to a text input) plus the existing selection suite; `SupportDataBinding` (42 tests) passes unchanged.
---

## Update 2: `bindTo` on values, `merge` on synths

The binding contract moved from the synth to the **value** (`b99a3c59`), and the system is now three named layers:

> **Synths serialize. Values bind. Bindings register. Durability via merge.**

- **`bindTo(binding)` on any rich value** — `wire:model` duck-checks the bound value instead of looking up its synth. This makes binding a primitive at *any* state depth: a future `wire:model="selection.selectAll"` header checkbox is just a getter returning a bindable facet (checked/indeterminate/toggle-page), which the synth-keyed lookup could never reach. Contract implementations read through `binding.get()` — never a captured `this` — so bindings self-heal if the value is replaced.
- **`merge(existing, incoming)` on synths** — server-driven changes now update the existing rich instance in place instead of swapping it, so instance identity survives round-trips (`window.ref === $wire.selection` after a server mutation — there's a browser test). Mirrors `FormObjectSynth` reusing `$existing` on the PHP side. Explicit userland replacement (`$wire.foo = new Thing`) still replaces — deliberate.
- **`WeakMap` value→synth memo** in `hydrateValue()` — diffing/dehydration hot paths skip the `match()` scan.

Selection uses all three; its binding registry now stores value-accessor facets instead of raw elements.
---

## Update 3: the `wireModel` accessor contract (final shape)

`bindTo` is gone. The contract is now one method on the value, returning the **same plain `get`/`set` accessor shape Alpine's x-model already honors** — x-model never stops running, we just hand it a smarter accessor:

```js
class Selection extends Array {
    wireModel(el, { cleanup }) {
        if (el.type !== 'checkbox' || ! el.hasAttribute('value')) return   // ← escape hatch

        let select = () => el.isConnected && this.select(el.value)
        this.__page().add(select)
        cleanup(() => this.__page().delete(select))

        return {
            get: () => this.contains(el.value),      // a checkbox models a boolean
            set: checked => checked ? this.select(el.value) : this.deselect(el.value),
        }
    }
}
```

Three possible returns — the whole contract:

| return | meaning |
|---|---|
| *nothing* | not mine — the element binds like normal data |
| `{ get, set }` | x-model binds the element to this accessor instead of the raw property |
| `false` | I wired the element myself — wire:model stands down (the file-upload takeover, generalized) |

Because a checkbox now models a *boolean*, Alpine's native machinery does all element wiring (checked rendering, change handling, form reset, `.fill`, modifiers) — Selection has **zero listeners and zero effects**. Every interaction is an in-place mutation, `merge()` keeps the instance alive across round-trips, and the page registry collapsed to a non-enumerable `Set` of closures *on the instance itself* — deleting the module registry, `__adopt`, and all component/path context. `selectPage()` naively calls the stored closures.

The `{ cleanup, effect }` helper bag supports richer bindings later — e.g. a `selection.selectAll` header-checkbox facet using `effect` to drive `el.indeterminate` while its accessor models "is the whole page selected."

New coverage: accessor remap + full-takeover browser tests in `SupportJsSynthesizers`; all selection, data-binding (42), and JS spec (138) suites green.
---

## Update 4: the `page` facet — header checkbox

```blade
<input type="checkbox" wire:model="selection.page" />   {{-- the column-header select-all --}}
```

- **checked** ⇢ the whole rendered page is selected
- **indeterminate** ⇢ the page is partially selected (driven by the contract's `effect` helper)
- **toggle** ⇢ `selectPage()` / `deselectPage()` — off-page selections survive a header uncheck

This is the first sub-property facet riding `interceptWireModel` — a getter on `Selection` returning a contract-bearing object — and it required **zero changes to wire:model**, which was the point of moving the contract onto values. The instance registry now stores value thunks (so facets can *query* page state, not just act on it), and `deselectPage()` / `pageSelected()` joined the public API.

Verified live: header toggles all 10 rows, one row unchecked flips it to indeterminate, clicking the indeterminate header re-selects the page (native browser behavior), and pagination correctly resets header state per page while selections persist. 3 new browser tests; all suites green (9 selection, 54 synth+data-binding, 138 JS specs).
---

## Update 5: dual-mode — `selectAll()` with an except-list

```blade
<flux:button wire:click="selection.selectAll()">Select all {{ $this->transactions->total() }}</flux:button>
```

**Wire format**: `{ mode: 'include' | 'except', keys: [...] }` — mode names what the keys *mean*: the selection itself, or the exceptions to "everything".

**The payoff of routing everything through `contains()`/`select()`/`deselect()`**: those three flip semantics in except mode, and *nothing else changed* — bound checkboxes, the `page` facet, header indeterminate, `selectPage()` all work identically in both modes. On the demo dashboard, selecting all and paginating shows page 2's rows checked without their keys ever being stored.

**API** (parity both sides): `selectAll()`, `isAll()`, `isAllSelected()`, `except()`. `clear()` always returns to an empty include-mode selection.

**Hostile-to-footguns**: enumerating an all-mode selection fails loudly — PHP `all()`/`toArray()`/`count()` throw with a pointed message rather than silently handing the *exception list* to a `whereIn()` (which would target exactly the wrong rows). JS `count()` returns `null` in all-mode; display pairs `isAll()` with a Blade-echoed total (kept deliberately client-side-light — server-side totals/`query()` scoping arrive with query-awareness later).

Also in this update: the checkbox gate recognizes Flux's `ui-checkbox` (same `.value`/`.checked`/`.indeterminate` contracts).

Verified live: select all → "2,479 selected" via Blade-echoed total, excepting a row drops the count, pagination shows every page checked, exceptions and mode survive round-trips, clear resets. 4 new unit tests + 2 new browser tests; suites: 8 unit, 12 selection browser, 54 synth+data-binding, 138 JS specs.
---

## Update 6: `keys()`, `outOf()`, `whereSelected()` — and the security model

**Renames**: `all()` → `keys()` (ambiguous once dual-mode landed; `all` is reserved for the select-all concept). `count()` contract: computable when the selection knows its total, loud failure (PHP) / `null` (JS) when it can't.

**`outOf()`** — feed the selection its result-set total:

```php
#[Computed]
public function transactions()
{
    return $this->selection->outOf(Transaction::latest()->paginate(10)); // pass-through
}
```

```blade
<flux:text wire:text="selection.count() + ' selected'"></flux:text>
```

No Blade echo inside expressions — the total rides the **snapshot meta** (server-owned, checksummed, restored across requests even when the computed doesn't run), so `count()` just works in both modes on both sides.

**`whereSelected()`** — the mode branch, absorbed:

```php
Transaction::whereSelected($this->selection)->delete();
// include → whereIn(key, keys) · except → whereNotIn(key, except) · column defaults to qualified model key
```

## Security: selection keys are client input

A selection is a **client-writable property** — that's its job — so a hostile client can forge *any* keys **and the mode flag** (`{ mode: 'except', keys: [] }` claims "everything is selected"). Treat `keys()`/`except()` with exactly the trust you'd give `request()->input('ids')`.

What the framework guarantees: the synth validates the hydrated class against `Selection`, the payload is size-capped, keys only ever reach queries as bindings (no injection), and the **total can't be forged** (meta rides the checksummed snapshot; only `outOf()` writes it). What the framework *cannot* know is ownership — so the app-level rule is:

> **Never point a selection at an unscoped query.** `auth()->user()->transactions()->whereSelected($selection)` — forged keys intersect to nothing, and except-mode's blast radius is bounded to rows the user owns.

With scoping in place, the invariant is: **a forged selection can never do more than the user could do by clicking every checkbox themselves.** `whereSelected` is designed as a *constraint you add to an already-scoped builder* — it composes with scoping, never replaces it — and the loud `keys()`/`count()` throws in all-mode exist so the classic unscoped-`whereIn` shortcut can't be written by accident. The docs page will carry this section prominently; future query-awareness (`$selection->query()`) can make the scoping itself server-defined.

Suites: 13 unit / 12 browser / 54 synth+data-binding / 138 JS specs, live dashboard verified ("2,479 selected" via meta-fed total, no expression echo).
---

## Update 7: `setTotal()`, the unscoped guard, and auto-initialization

**`outOf()` → `setTotal()`** — the naming interrogation surfaced a design call: totals are now an *explicit opt-in*, not a core concept. "All selected (3 excluded)" needs zero total plumbing and is the documented default; feeding a total for a computed all-mode `count()` is a rarely-written escape hatch (`tap($paginator, fn ($p) => $this->selection->setTotal($p))`), and escape-hatch setters get honest setter names.

**`whereSelected` unscoped guard** — a select-all selection against a completely unconstrained query (no wheres, no global scopes) now **throws** unless acknowledged with `unscoped: true`, a visible marker for code review. Guarded in except-mode only, deliberately: that's the asymmetric catastrophe (four forged bytes → `whereNotIn` targets the whole table), while include-mode forged keys are ordinary request-input territory.

**Auto-initialization** — `public Selection $selection` now springs to life with no `mount()` assignment, via a *general* mechanism: any synth may define `initialize($type, $assign)`, and `SupportAutoInitializedProperties` fills typed uninitialized public properties at boot. The speed guarantee, benchmarked (10-property component):

| strategy | per boot |
|---|---|
| old full-reflection scan (what form objects did on every boot) | 649–1,043 ns |
| new cached scan, nothing to initialize (common case) | **74 ns** |
| new cached scan, 2 selections initialized | 330 ns |
| cold discovery (once per class per process) | 3,750 ns |

Discovery is cached **per component class**, and form objects migrated onto the shared cache (their per-boot reflection scan is deleted) — so a 1,000-component page boots **~3× faster than main** (~0.33 ms vs ~1.04 ms). The `$assign` callback lets a synth control ordering around the assignment (forms boot after it, preserving historical behavior).

Verified: full unit suite (1,387 tests) green, 24 selection+synth browser tests, live dashboard with `mount()` deleted.
---

## Update 8: verified against real `flux:checkbox`

The dashboard demo now runs entirely on `<flux:checkbox>` (rows + page-facet header) — no native inputs. Verified live via Playwright:

- row toggling in **both directions** (the direction that would break if Alpine didn't know `ui-checkbox` — and it does: Alpine's `isCheckbox()` is `el.type === 'checkbox' || localName === 'ui-checkbox' || 'ui-switch'`, so the whole accessor path treats Flux elements natively)
- the `selection.page` header: checked / **indeterminate** (Flux's `data-indeterminate` styling attribute engages, rendering the native-looking dash), toggle both ways
- select-all across pagination, exceptions, clear, morph round-trips — zero JS errors

Also added `ui-switch` to the selection gate to match Alpine's own checkbox grouping. Note for grouped usage: `flux:checkbox.group` stops event propagation from its children, but `wire:model` listens on the element itself, so per-checkbox bindings survive inside groups too.


js/directives/wire-model.js
```diff
@@ -119,6 +120,36 @@ directive('model', ({ el, directive, component, cleanup }) => {
         bindings['@keydown.enter'] = () => update()
     }
 
+    // Rich synth values can declare their own element binding contract.
+    // When the bound value's synth defines bind(), it owns the element
+    // wiring (instead of Alpine's x-model) while network timing —
+    // .live, triggers, debounce — stays wire:model's job...
+    let synth = findSynthByValue(dataGet(component.ephemeral, expression))
```
calebporzio: is this an expensive thing to add to the wire:model initialization path? It's not deep which is good, but still just curious.

js/directives/wire-model.js
```diff
@@ -120,14 +119,17 @@ directive('model', ({ el, directive, component, cleanup }) => {
...
-    if (synth?.bind) {
-        let handled = synth.bind({
+    // Bindable values: a rich value anywhere in state may define
+    // bindTo(binding) — its contract for owning the element wiring
+    // (instead of Alpine's x-model) while network timing — .live,
+    // triggers, debounce — stays wire:model's job. Implementations
+    // should read state through binding.get(), never a captured `this`,
+    // so bindings survive the value being replaced. Return false to
+    // decline an element and fall back to default handling...
+    let bound = dataGet(component.ephemeral, expression)
```
calebporzio: before this addition we weren't accessing the value on wire:model init right? like we are doing a direct get without a microtask or anything. will this cause any timing problems? like accessing a value before the full initialization sweep or something? seems fine but there's always weird timing things like this that get funky.

js/directives/wire-model.js
```diff
@@ -142,8 +144,8 @@ directive('model', ({ el, directive, component, cleanup }) => {
 
         if (handled !== false) {
             // Network trigger listeners (@blur/@change/@enter) still apply.
-            // They bind after the synth's own listeners, so a synth mutation
-            // on the same event lands before the network request fires...
+            // They bind after the value's own listeners, so a mutation on
+            // the same event lands before the network request fires...
             Alpine.bind(el, bindings)
 
             return
```
calebporzio: I wonder if now that this is so generalized we should put this bind contract in Alpine and then we wouldn't need this early return (i suppose we would need it in alpine though lol, but might be cleaner and now taht i think about it it might even EXIST there already lol) look at alpine's x-model source and wsee what you see

js/features/supportSelection.js
```diff
@@ -2,31 +2,33 @@ import { registerSynth } from '@/synths'
...
-// contract: component -> path -> Set(el). The registry IS the current page —
-// teleported elements stay covered because registration doesn't care where
-// an element lives...
-let bindings = new WeakMap()
+// Every binding attached to a selection, registered through the bindTo
+// contract: component -> path -> Set(binding). Each binding knows how to
+// read its element's value — the registry IS the current page, and it's
+// queryable (unlike events), which is what page-level affordances like
+// selectPage() and a future select-all facet need...
+let registry = new WeakMap()
```
calebporzio: I still wanna see if there's some way to avoid a registry or at least a registry that stores granular things like el and value directly.

i think i would prefer a registry of closures that contain their own el/value state and we can just naively call them

js/features/supportSelection.js
```diff
@@ -2,31 +2,33 @@ import { registerSynth } from '@/synths'
...
 
-function register(component, path, el, cleanup) {
-    let byPath = bindings.get(component)
+function register(component, path, binding, cleanup) {
+    let byPath = registry.get(component)
```
calebporzio: this is fine but look in utils for custom types like WeakSet or something. hybrids i created to avoid this repetitive boilerplate. its fine if they aren't suitable.

js/features/supportSelection.js
```diff
@@ -57,13 +59,11 @@ export class Selection extends Array {
...
+    // to this selection, each asked for its element's value...
     selectPage() {
-        for (let el of this.__boundEls()) {
-            if (! el.isConnected) continue
-
-            let value = elValue(el)
+        for (let binding of this.__bindings()) {
+            let value = binding.value()
 
             if (value !== undefined && value !== null && value !== '') this.select(value)
```
calebporzio: relevant to my previous comment. i feel like select page (if it is to use a registry) should just be naively looping through stored items and calling a callback that does the desired action.

js/features/supportSelection.js
```diff
@@ -73,16 +73,42 @@ export class Selection extends Array {
         this.splice(0, this.length)
     }
 
-    __boundEls() {
+    // Selections own their element semantics: checked means "this value is
+    // selected", toggling mutates the selection in place. Reads go through
+    // binding.get() — never `this` — so the binding keeps working even if
+    // this instance is ever replaced...
+    bindTo({ el, component, path, get, notify, cleanup }) {
```
calebporzio: not sure about this name. doesn't seem like an important contract that gets picked up by a core thing like wire:model.

feels like it should be __bindTo at LEST or wireModelBind()? idk that sucks too.

js/features/supportSelection.js
```diff
@@ -73,16 +73,42 @@ export class Selection extends Array {
...
+        let value = () => el.isConnected ? (el.getAttribute('value') ?? el.value) : undefined
+
+        register(component, path, { el, value }, cleanup)
+
+        Alpine.bind(el, {
+            ['x-effect']() {
+                el.checked = get().contains(value())
+            },
+
+            ['@change']() {
```
calebporzio: isn't change already registered from wire-model still? is it fine to have both?

I wonder if this whole architecture should be more like a hook/passthrough where this callback gets an opportunity to extract a get/set callback to call when it wants but all the normal alpine bindings happe nincluding x-model even i think? idk...

js/features/supportSelection.js
```diff
@@ -109,25 +131,9 @@ registerSynth('sel', {
...
-            ['@change']() {
-                get().toggle(elValue(el))
-
-                notify()
-            },
-        })
+    // Server-driven changes update the existing instance in place so its
+    // identity (and adopted context) survive the round-trip...
+    merge: (existing, incoming) => {
+        existing.splice(0, existing.length, ...incoming)
```
calebporzio, line 217: looks good.

js/directives/wire-model.js
```diff
@@ -28,6 +28,28 @@ directive('model', ({ el, directive, component, cleanup }) => {
...
+    // Bindable values: a rich value may define wireModel(el, helpers) to
+    // decide how an element binds to it. Three possible returns:
+    //   nothing      → not mine; the element binds like normal data
+    //   { get, set } → x-model binds the element to this accessor instead
+    //                  of the raw property (get feeds renders, set receives
+    //                  input) — network timing stays wire:model's job
+    //   false        → the value wired the element itself; do nothing more
+    let bound = dataGet(component.$wire, expression)
+
+    let accessor = typeof bound?.wireModel === 'function'
```
calebporzio: i think the name is still kinda meh. i think we should brainstorm a precedent.

if there is something we can follow like an existing hook or pattern.

or alternatively, we maybe go with an excrutiatingly explicit name like: interceptWireModel or something. should think about __ prefixes or a livewire specific one which i think we have used before? idk... or just intercept? idk... that's too bare probably.

calebporzio: Renamed to `interceptWireModel` in 61b76b58. Precedent survey: `toLivewire`/`fromLivewire` (plain explicit protocol names, no sigils), `interceptRequest`/`interceptMessage`/`$wire.$intercept` (existing public vocabulary for "handed an operation before default behavior — proceed, remap, or take over", which matches this contract's three returns exactly), `__` prefix ruled out (house rule: internal-only, this is public extender surface). The interceptor family won.

js/features/supportSelection.js
```diff
@@ -38,39 +38,81 @@ export class Selection extends Array {
...
     selectPage() {
-        this.__page().forEach(select => select())
+        this.__pageValues().forEach(value => this.select(value))
+    }
+
+    deselectPage() {
+        this.__pageValues().forEach(value => this.deselect(value))
+    }
+
+    pageSelected() {
```
calebporzio: maybe we call this "pageIsSelected"? i don't love the tripple word. if this is an established precedent for libraries like this than im good with it. but isEmpty reads great. pageSelected might read just as good just want to be sure...

js/features/supportSelection.js
```diff
@@ -38,39 +38,81 @@ export class Selection extends Array {
...
+            interceptWireModel(el, { effect }) {
+                if (el.type !== 'checkbox') return
+
+                effect(() => {
+                    let values = selection.__pageValues()
+                    let selected = values.filter(value => selection.contains(value))
+
+                    el.indeterminate = selected.length > 0 && selected.length < values.length
+                })
+
```
calebporzio, line 121: i sort of expected this to return false and handle all the stuff itself but am pleasantly surprised that is not the case. just want to make sure that there's no scenario where the checkbox would be put into an intermitant state and then took out of it for one reason or another when it hsould  stay indeterminant but for some reason the effect didn't run or something idk.. just something to tripple check.

js/features/supportSelection.js
calebporzio: Went with `isPageSelected()` (cdca3b01). Precedent: is-LEADING is the established boolean convention — this class's own `isEmpty()`, `Array.isArray`, AG Grid `isSelected()`, TanStack Table `getIsAllPageRowsSelected()` — while subject-first (`pageIsSelected`) has almost no library precedent. Bonus: it kills a real ambiguity — `pageSelected()` reads plausibly as "the selected items of the page" (a list).

js/features/supportSelection.js
calebporzio: Triple-checked. The design is declarative — indeterminate is re-derived from state, never toggled imperatively — so staleness requires an untracked dependency. Full audit: (1) only our effect ever writes `el.indeterminate` (x-model/morph don't know it exists); (2) tracked deps: the registry Set (row checkboxes registering/cleanup during morphs trigger it) and selection contents (every mutation path incl. server `merge` triggers it); (3) the browser natively clears indeterminate on header click, but a header click always produces a non-partial state (selectPage/deselectPage), so that clear is always correct — row clicks don't touch the property, they mutate selection → effect re-derives; (4) the one real gap was a server round-trip morph where deps DON'T change and the effect never re-runs — the property has to survive the morph on its own. It does, and there's now a browser test pinning exactly that (partial → $refresh → still indeterminate, cdca3b01). Remaining known caveat, shared feature-wide: wholesale userland replacement (`$wire.selection = [...]`) orphans the captured instance until the next hydration.

src/Features/SupportSelection/SupportSelection.php
```diff
@@ -11,5 +12,18 @@ public static function provide()
         app('livewire')->propertySynthesizer(
             SelectionSynth::class
         );
+
+        // Constrain a query to a selection — whereIn in include mode,
```
calebporzio, line 16: this looks great i just want to make sure there isn't SOME way to protect people? like checking if the query is scoped already and if not requring a forced param ro something? idk... just feels super dangerous without SOME kind of protection for users.

src/Features/SupportSelection/Selection.php
```diff
@@ -10,11 +10,35 @@ class Selection implements Arrayable, \Countable, \IteratorAggregate, \JsonSeria
...
     public function __construct(
         protected array $keys = [],
         protected string $mode = 'include',
     ) {}
 
+    // Tell the selection how many results exist so count() can be computed
+    // in select-all mode. Accepts a paginator (and passes it through, so a
+    // computed can return `$this->selection->outOf(Model::paginate())`) or
+    // a plain total...
+    public function outOf($total)
```
calebporzio: I'd like to interrogate "outOf" as a name.

"deriveTotalFrom()" is the most explicity i can think of.

would also be good to brainstorm other orthogonal options...

src/Features/SupportSelection/SupportSelection.php
calebporzio: Built (d4a5d9b0): `whereSelected` now **refuses a select-all selection against an unconstrained query** — throws with a pointed message unless the builder already carries wheres/global scopes, or the caller acknowledges with `unscoped: true` (a visible marker for code review). Guard is scoped to except-mode deliberately: that's the catastrophic asymmetric failure (four forged bytes → whereNotIn targets the whole table), while include-mode forged keys are ordinary request-input territory bounded by whatever scoping exists — guarding it would tax every single-user app's happy path. Global scopes count as constraints (soft-tenancy passes). Covered by unit tests incl. the global-scope case.

src/Features/SupportSelection/Selection.php
calebporzio: Landed on `setTotal()` (0d5222ef) after a wider brainstorm. The interrogation was right: every candidate name (`from`, `within`, `deriveTotalFrom`) hurt because the *concept* was awkward — a value being narrated facts about a neighbor. The design answer: totals are now an explicit opt-in, not a core concept — "All selected (3 excluded)" needs zero total plumbing and becomes the documented default; feeding a total for a computed all-mode `count()` is a rarely-written escape hatch via `tap($paginator, fn ($p) => $this->selection->setTotal($p))`. Escape-hatch setters get honest setter names. (Considered and rejected: render-time auto-discovery of the paginator — slightly too magical.) Pass-through behavior deleted along with the pretty-sentence ambitions.

src/Features/SupportAutoInitializedProperties/SupportAutoInitializedProperties.php
```diff
@@ -0,0 +1,67 @@
...
+        $component = $this->component;
+
+        $entries = static::$discovered[$component::class] ??= static::discover($component::class);
+
+        foreach ($entries as [$property, $typeName, $synthClass]) {
+            // Only fill properties nothing else has initialized (a default,
+            // a hydration, an earlier hook)...
+            if ($property->isInitialized($component)) continue;
+
+            $synth = new $synthClass(new ComponentContext($component), $property->getName());
```
calebporzio: is this kind of surgical into the synth system? I'm getting vibes that maybe this shouldn't belong here?

src/Features/SupportAutoInitializedProperties/SupportAutoInitializedProperties.php
```diff
@@ -0,0 +1,67 @@
...
+
+            // The synth assigns through the callback so it controls ordering
+            // around the assignment (e.g. form objects boot after)...
+            $synth->initialize($typeName, fn ($value) => $property->setValue($component, $value));
```
calebporzio: i'm wondering if this should live directly inside the synth system? like this entire feature? I am totally fine if there is a deliberate reason for it to be a dedicated feature. like it's nice to give it a name and a file and a unit test as a nice neat package, but the synth system isn't something that should be smeared around the codebase.

src/Features/SupportFormObjects/FormObjectSynth.php
```diff
@@ -16,6 +16,34 @@ static function match($target)
...
+    // Typed updates without snapshot meta pass through untouched — the
+    // same behavior as before forms were matchable by type...
+    function hydrateFromType($type, $value)
+    {
+        return $value;
+    }
+
+    // Uninitialized `public PostForm $form` properties spring to life.
+    // The form's boot runs AFTER the property assignment, preserving the
+    // ordering of the old per-boot reflection scan...
```
calebporzio: i don't like comments that reference a timeline? like "old per-bootreflection scan.." these feel like comments meant forme to read right now in this moment. not comments to live in the codebase for eternity. please edit them all with this in mind. this will likely mean deleteing many or at least editing them to be more evergreen.

src/Features/SupportAutoInitializedProperties/SupportAutoInitializedProperties.php
calebporzio: Agreed and moved: the scan, the class-keyed cache, and the synth construction now live in `HandleSynths::initializeProperties()` — synth machinery homed in the synth mechanism. The feature folder shrank to exactly the neat-package part you described: a boot hook that calls the mechanism, a name, a docblock, and the tests. `findSynthClassByType` left the public surface (inlined into discovery).

src/Features/SupportFormObjects/FormObjectSynth.php
calebporzio: Swept the whole diff for timeline phrasing. Fixed: the hydrateFromType comment now explains WHY forms pass through (a form can't be rebuilt from a raw value alone — needs component + booted attribute state), the initialize comment explains why boot runs after assignment (boot-time logic reaches the form through its component), and a stale browser-test comment/name still referencing instance replacement (which no longer even happens) became test_select_page_composes_with_manual_toggles.

src/Mechanisms/HandleSynths/HandleSynths.php
```diff
@@ -183,13 +183,54 @@ protected function findByTarget($target, $context, $path)
         return new ($this->typeCache[$type])($context, $path);
     }
 
-    public function findSynthClassByType(string $type): ?string
+    // Typed public properties whose synthesizer defines initialize() are
+    // filled automatically at boot. Discovery is cached per component
+    // class: one reflection pass per class per process, then each boot
+    // pays an array lookup plus an isInitialized() check per entry...
+    protected static array $initializable = [];
```
calebporzio, line 192: static array's can become memory leaks if an octane environment. need to make sure to have a flush mechanism if this is the case. look for prior art about this across the codebase.

src/Features/SupportAutoInitializedProperties/SupportAutoInitializedProperties.php
```diff
@@ -12,56 +11,13 @@
...
-
-            $synthClass = app(HandleSynths::class)->findSynthClassByType($type->getName());
-
-            if ($synthClass === null || ! method_exists($synthClass, 'initialize')) continue;
-
-            $entries[] = [$property, $type->getName(), $synthClass];
-        }
-
-        return $entries;
+        app(HandleSynths::class)->initializeProperties($this->component);
```
calebporzio: i suppose this is better, however shouldn't it live somewhere else? feels a little weird to just hook int oboot from a feature class. where do the rest of the synth methods get called? find that location and see if this fits there right alongside it. thanks

src/Mechanisms/HandleSynths/HandleSynths.php
calebporzio: Prior art found and followed (1ba93473): the codebase convention is `on('flush-state', ...)` registered in a mechanism's boot() — HandleComponents, ExtendBlade, FrontendAssets, PersistentMiddleware all do exactly this, and Octane fires `Livewire::flushState()` between requests. HandleSynths now registers the same listener and resets the discovery cache. Worth noting the leak profile: the cache is keyed by component class holding ReflectionProperty entries, so it's bounded by the number of component classes in the app — flush-state makes it strictly correct rather than fixing unbounded growth, but correct is correct.

src/Features/SupportAutoInitializedProperties/SupportAutoInitializedProperties.php
calebporzio: Found the location and it fit perfectly (1ba93473): the mechanism's synth calls all live in HandleComponents (`$this->synths->hydrate/dehydrate` in hydrateProperties/dehydrateProperties), so initialization now happens there too — `$this->synths->initializeProperties($component)` immediately after instantiation in both mount() and fromSnapshot(). Components are born initialized; no lifecycle hook involved. The feature class and its bootFeatures registration are deleted — the folder keeps the name + UnitTest only, which has precedent (SupportOctane is exactly this shape). Full 1,387-test unit suite green after the move.

src/Features/SupportAutoInitializedProperties/UnitTest.php
```diff
@@ -9,6 +9,15 @@
...
 use Tests\TestComponent;
 
+/**
+ * Typed public properties whose synthesizer knows how to initialize them
+ * (an initialize() method on the synth) spring to life automatically:
+ *
+ *     public Selection $selection;   // no mount() assignment needed
+ *
+ * The machinery lives in the synth system — see
+ * HandleSynths::initializeProperties().
```
calebporzio: dont you think this unit test should live somewhere else and this feature folder should be deleted? thanks.

calebporzio: Yep — folder deleted. The tests merged into `src/Mechanisms/HandleSynths/UnitTest.php`, which already existed and is where the mechanism's other behavior is tested. Auto-initialization now has zero footprint outside the synth system: the machinery, the lifecycle calls, and the tests all live with the mechanism.

docs/selection.md
```diff
@@ -0,0 +1,294 @@
+Checking rows in a table and acting on them — archive these, delete those, export the lot — is one of the most common patterns in data-driven interfaces. It's also surprisingly fiddly to build by hand: tracked keys scattered across an array property, "select all" state that lies after paginating, header checkboxes with three states, and bulk queries that better not trust the client.
```
calebporzio: This reads a little too cute. "acting on them - archive these, delete those, export the lot -" "surprisingly fiddly"

here's a quick course correction:.

Livewire provides a dedicated `Selection` object for easily building selection interfaces like checkboxes in table rows, item selection in a media library etc... All the tricky bits including true "select-all" (select all rows in the database), exclusion, select page, etc... are all taken care of for you so that building interfaces like this becomes an absolute breeze.

(i know the next paragraph is kinda like that. so don't take my writing as gospel, just a starting point for you to kinda see the things i like adnd don't like.) i personalyl think docs should have less voice than you injected initially.

docs/selection.md
```diff
@@ -0,0 +1,294 @@
...
+
+Livewire provides a dedicated `Selection` object for this. It tracks checked keys across renders and pagination, powers "select page" and "select all" controls, and constrains bulk-action queries to exactly the selected rows.
```
calebporzio: this is good

docs/selection.md
```diff
@@ -0,0 +1,294 @@
...
+
+use Livewire\Attributes\Computed;
+use Livewire\WithPagination;
+use Livewire\Component;
+use Livewire\Selection;
+
+new class extends Component {
+    use WithPagination;
+
+    public Selection $selection;
```
calebporzio: thinking we should call this "selected"? what do you think? like: public Selection $selected

docs/selection.md
```diff
@@ -0,0 +1,294 @@
...
+
+    public function archiveSelected()
+    {
+        auth()->user()->invoices()
+            ->whereSelected($this->selection)
+            ->update(['archived_at' => now()]);
+
+        $this->selection->clear();
+
+        unset($this->invoices); // Bust the memoized list so this render reflects the update...
```
calebporzio: do we have to actually bust it? computeds only memoize during a single round trip. if a user presses "archiveSelected", this code will run before $this->invoices is ever accessed I believe. so no need for this.

docs/selection.md
```diff
@@ -0,0 +1,294 @@
...
+<input type="checkbox" wire:model="selection.page">
+```
+
+This checkbox is fully wired automatically:
+
+- It's checked when every rendered row checkbox is selected
+- It shows the native indeterminate state when only some are
+- Toggling it selects or deselects every rendered row — keys selected on other pages are left untouched
+
+"The page" means whatever row checkboxes are currently rendered in the browser. Because that's a client-side notion, the page methods — `selectPage()`, `deselectPage()`, and `isPageSelected()` — are available in directive expressions and JavaScript rather than on the PHP object:
```
calebporzio: change "The page" means whatever row checkboxes are

to just: "The page" means whatever checkboxes are

docs/selection.md
```diff
@@ -0,0 +1,294 @@
...
+
+```blade
+<button wire:click="selection.selectPage()">Select page</button>
+```
+
+## Selecting all results
+
+Selecting every row on the current page is often just the first step — the user really wants all 2,500 matching records. Enumerating every key client-side would be impractical, so `selectAll()` flips the selection into *select-all* mode: it now represents every result, and tracks only the *exceptions* — rows unchecked since.
```
calebporzio: i think this could read a little better, especially around the emdashes which are maybe a little much.

docs/selection.md
```diff
@@ -0,0 +1,294 @@
...
+
+```blade
+<button wire:click="selection.selectAll()">
+    Select all {{ $this->invoices->total() }}
+</button>
+```
+
+Everything keeps working in select-all mode. Bound checkboxes render checked, unchecking a row records an exception, the page facet and `contains()` answer correctly, and `whereSelected()` constrains queries with `whereNotIn` instead of `whereIn` — automatically.
```
calebporzio: "the page facet" - no one knows what that means. way too jargony.

"answer correctly" feels a little weird too.

docs/selection.md
```diff
@@ -0,0 +1,294 @@
...
+<template wire:if="! selection.isAllSelected()">
+    <button wire:click="selection.selectAll()">
+        Select all {{ $this->invoices->total() }}
+    </button>
+</template>
+```
+
+### Counting a select-all selection
+
+In select-all mode, the selection can't know its own count — that requires knowing how many results exist. Feed it a total from your paginator with `setTotal()`:
```
calebporzio: "Feed it a total from your paginator" is abrupt.

"You can feed it a total from your paginator" is nicer

take this note for the rest of this docs page. look at how i write documentation on other pages and try to mimic that tone. friendly, voiceless, clear, concise but not terse.

docs/selection.md
```diff
@@ -0,0 +1,294 @@
...
+| `clear()` | Deselect everything and leave select-all mode |
+
+Keys are compared loosely — checkbox values arrive as strings while database keys are often integers, so `'1'` and `1` refer to the same row.
+
+### PHP-only methods
+
+| Method | Description |
+|--------|-------------|
+| `setTotal($total)` | Feed a result total for select-all counts — accepts a paginator or an integer |
+| `total()` | The stored total, or `null` |
```
calebporzio: why not give total() to client as well? same return style. i say you add it to client as well and move it up from this category to the reference category before it.

docs/selection.md
calebporzio: Rewrote the intro (00ad6cde) with your draft as the base: one plain sentence naming the object and the interfaces it's for, then the capability list, with "all the tricky parts are taken care of for you" carrying the pitch. The cute framing paragraph is gone entirely.

docs/selection.md
calebporzio: Agreed, and adopted throughout (00ad6cde). $selected reads better at every call site: wire:model="selected" on a checkbox reads as a plain fact, selected.count() + ' selected' echoes the UI label, whereSelected($this->selected) stops double-saying "selection". The type name already carries the noun, so the property is free to read like state. Docs + the playground demo now use it everywhere.

docs/selection.md
calebporzio: You're right — removed. And verified the claim in the playground rather than trusting it: deleted the demo's unset($this->transactions) too, clicked through delete-selected live, and the list refreshes correctly because the computed is first accessed at render, after the action ran.

docs/selection.md
calebporzio: Changed.

docs/selection.md
calebporzio: Rewritten (00ad6cde): three plain sentences, no em-dashes — page-is-just-the-first-step, why enumeration is impractical, then what selectAll() flips the selection into, with exceptions defined in-line.

docs/selection.md
calebporzio: De-jargoned: now "the header checkbox and contains() behave exactly as you'd expect". Also swept "facet" from the rest of the page — the page section now says "bind it to selected.page" instead of naming a concept nobody's met.

docs/selection.md
calebporzio: Took this page-wide (00ad6cde): reread pagination.md for register and swept the whole page — "You can feed it a total", "you can branch on the mode instead", "you can acknowledge it explicitly", fewer em-dashes throughout, and the reference-table fragments turned into sentences.

docs/selection.md
calebporzio: Built (00ad6cde): total() on the client Selection with the same return style (the stored total, or null if one was never set), moved from PHP-only into the shared reference table — setTotal() is now the only PHP-only method. Mentioned in the counting section too, and covered by a browser test: server-side setTotal(2500), then a wire:text bound to selected.total() renders it and count() computes from it after selectAll().

PR #10427, opened 2026-07-21
[4.x] Support minute durations for `wire:poll`

### LINKED ISSUE #10419: `wire:poll` silently ignores unrecognized duration modifiers (e.g. `.10m`) and polls every 2 seconds
### Livewire version
v4.3.3

### Laravel version
v13.x

### Description

`wire:poll`'s duration parser only recognizes `ms` and `s` units. Any other
modifier — most dangerously a minutes-style one like `wire:poll.10m` — does not
match either regex and **silently falls back to the 2000 ms default**, with no
warning of any kind.

The failure mode is nasty because it is invisible in the UI: the component
polls *more* often than intended, so everything looks fine while every open
page hammers the server at 300x the intended rate. We shipped
`wire:poll.10m` on 5 dashboard widgets across three production environments
and only discovered it while chasing a different console-noise symptom
(companion issue: #10420, cancelled poll actions surfacing as unhandled rejections):
each open dashboard was issuing a bundled Livewire update every ~3–5 seconds
instead of one every 10 minutes.

`.10m` reads as perfectly plausible syntax (`10m` is a common duration
shorthand in e.g. Go, Prometheus, systemd), and the docs only show `s`/`ms`
examples without stating that anything else is ignored.

### Root cause

`js/directives/wire-poll.js`:

```js
function extractDurationFrom(modifiers, defaultDuration) {
  let durationInMilliSeconds;
  let durationInMilliSecondsString = modifiers.find((mod) => mod.match(/([0-9]+)ms/));
  let durationInSecondsString = modifiers.find((mod) => mod.match(/([0-9]+)s/));
  if (durationInMilliSecondsString) {
    durationInMilliSeconds = Number(durationInMilliSecondsString.replace("ms", ""));
  } else if (durationInSecondsString) {
    durationInMilliSeconds = Number(durationInSecondsString.replace("s", "")) * 1e3;
  }
  return durationInMilliSeconds || defaultDuration;   // ← silent fallback
}
```

`"10m"` matches neither `/([0-9]+)ms/` nor `/([0-9]+)s/`, so the function
returns `defaultDuration` (2000).

### Steps to reproduce

1. Add `wire:poll.10m` to any component's root element.
2. Open the page with DevTools → Network.
3. Observe a Livewire update request every ~2 seconds instead of every 10 minutes.

### Expected behavior

Any of these would prevent the footgun (in order of preference):

1. Support `m` as a unit (`.10m` → 600 000 ms). A duration-looking modifier
   that parses like a duration should behave like one. (I'm aware minute
   support was declined back in #674 for v1 — but that predates how common
   long-interval dashboard polling is now.)
2. If minutes stay unsupported: **warn** when a modifier looks like a
   duration but isn't recognized, e.g. anything matching `/^\d+[a-z]+$/`
   that isn't `ms`/`s` → `console.warn('[wire:poll] unrecognized duration
   ".10m", using default 2s')`.
3. At minimum, document that unrecognized modifiers silently fall back to
   the 2 s default.

### Actual behavior

Silent fallback to 2000 ms; no console output; no docs mention.

joshhanley:
This PR is a companion to PR https://github.com/livewire/livewire/pull/10426 which adds a warning for unsupported durations.

# The Scenario

When using `wire:poll` for dashboard widgets that only need occasional updates, a duration such as `wire:poll.10m` appears to represent ten minutes but instead polls every two seconds.

```blade
<div wire:poll.10m>
    ...
</div>
```

# The Problem

The `wire:poll` duration parser only recognises milliseconds and seconds. A duration using `m` is not parsed, so Livewire falls back to the default two-second interval.

# The Solution

The solution is to support minutes as a polling duration.

```blade
<div wire:poll.10m>
    ...
</div>
```

Fixes #10419

### DIFF
--- a/docs/polling.md
+++ b/docs/polling.md
@@ -49,6 +49,8 @@ The best way to reduce requests in this scenario is simply to make the polling i
 You can manually control how often the component will poll by appending the desired duration to `wire:poll` like so:
 
 ```blade
+<div wire:poll.1m> <!-- In minutes... -->
+
 <div wire:poll.15s> <!-- In seconds... -->
 
 <div wire:poll.15000ms> <!-- In milliseconds... -->
--- a/docs/wire-poll.md
+++ b/docs/wire-poll.md
@@ -52,6 +52,8 @@ The best way to reduce requests in this scenario is simply to make the polling i
 You can manually control how often the component will poll by appending the desired duration to `wire:poll` like so:
 
 ```blade
+<div wire:poll.1m> <!-- In minutes... -->
+
 <div wire:poll.15s> <!-- In seconds... -->
 
 <div wire:poll.15000ms> <!-- In milliseconds... -->
@@ -88,6 +90,7 @@ wire:poll="action"
 
 | Modifier | Description |
 |----------|-------------|
+| `.[number]m` | Poll interval in minutes (e.g., `.1m`) |
 | `.[number]s` | Poll interval in seconds (e.g., `.15s`) |
 | `.[number]ms` | Poll interval in milliseconds (e.g., `.15000ms`) |
 | `.keep-alive` | Continue polling even when the tab is in the background |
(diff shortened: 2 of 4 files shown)


ghabriel25:
@joshhanley Do you think its time to have helper method for duration modifier? That would be a great benefit from many places in js section

joshhanley:
@ghabriel25 yeah maybe. We can have a look at that if this PR gets merged 🙂

calebporzio:
@copilot resolve the merge conflicts in this pull request

PR #10441, opened 2026-07-23
Detect temporary upload mime type from actual file contents

tanthammar:
Same fix as #10440, ported to v4


joshhanley:
@tanthammar thanks for the PR! Going to close this in favour of the v3 one, as we up-merge all fixes from v3 to v4 anyway 🙂

PR #10442, opened 2026-07-24
[4.x] Refactor component hooks to use existing helper method from `ComponentHook` class

ghabriel25:
# Motivation
Inside `ComponentHook` class, there are already data store helper methods that can be used from component hooks. However, currently there are many component hooks that still using `store` function instead those helper methods

- `SupportEvents`
- `SupportJsEvaluation`
- `SupportJsModules`
- `SupportLazyLoading`
- `SupportNestedComponentListeners`
- `SupportScriptsAndAssets`
- `SupportWireModelingNestedComponents`

# What's Changed
- Change all `store()` to existing helper methods from `ComponentHook` class
- In addition, adds `$this->storeFind()` to `ComponentHook` class

### DIFF
--- a/src/ComponentHook.php
+++ b/src/ComponentHook.php
@@ -108,6 +108,11 @@ abstract class ComponentHook
         return store($this->component)->get($key, $default);
     }
 
+    function storeFind($key, $iKey = null, $default = null)
+    {
+        return store($this->component)->find($key, $iKey, $default);
+    }
+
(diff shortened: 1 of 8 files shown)


calebporzio:
great! thanks!

ghabriel25:
@calebporzio I missed `SupportLifecycleHooks`, I'll push it right away

PR #10443, opened 2026-07-24
Virtual properties: #[Virtual] methods that construct synth-backed properties

calebporzio:
## What

**Virtual properties**: a public method marked `#[Virtual]` acts as a property. The method is the property's *constructor* — it runs fresh every request, so it's the home for configuration, closures, and anything else that doesn't serialize:

```php
use Livewire\Attributes\Virtual;

new class extends Component
{
    // Before...
    // public Selection $selected;

    // After...
    #[Virtual]
    public function selected(): Selection
    {
        return new Selection(keys: ['bar'], mode: 'except');
    }
};
```

```blade
<flux:checkbox wire:model="selected.page" label="Select all" />
<flux:checkbox wire:model="selected" value="foo" label="Foo" />
```

On the wire it's indistinguishable from a normal property — it dehydrates through its normal synth into snapshot `data`, the JS layer needs zero changes, and `wire:model` just works.

## Where things live

Everything sits where its siblings already sit:

- **`$component->all()` is the one abstraction for "every property"** — declared + virtual, virtuals always last. Dehydration is literally `$data = $component->all()`, and view data, update validation, validation data, and island scope all use it. No more double-reaching into Utils for declared and somewhere else for virtual.
- **`HandlesVirtualProperties` trait on `Component`** (and `Form`) owns discovery (static per-class reflection cache, validated for return types + name collisions) and the per-instance `$__virtualProperties` lookup, plus access: `has/get/set/unsetVirtualProperty()`, `initializeVirtualProperties()`. No synth logic in the trait.
- **Initialization is one step**: `HandleSynths::initializeProperties()` initializes declared typed properties via their synths, then virtual properties via their methods — on both the mount and hydrate paths, symmetrically.
- **Synthy stuff stays in synth land**: hydration lives inline in `hydrateProperties`, mirroring the declared flow line-for-line — declared keys go `$this->synths->hydrate(...)` → assignment; virtual keys go `$this->synths->hydrateInto(...)` → `setVirtualProperty(...)`.
- **`Component::__get`/`__unset`** consult the trait directly. `#[Virtual]`/`BaseVirtual` is a pure marker whose only runtime job is throwing `CannotCallVirtualPropertyDirectlyException` when the method is invoked as an action.

## The lifecycle stories

**Mount** — boot initializes declared + virtual properties in one pass (before `mount()`). `$this->selected` works in PHP, `$selected` in Blade. At dehydrate, `all()` feeds every property through normal synth dehydration.

**Hydrate** — boot initializes the same way: every `#[Virtual]` method runs and supplies a fresh configured instance. Then `hydrateProperties` walks the snapshot and hydrates raw state INTO those instances via the synth contract `hydrateInto($target, $value, $meta)` — a pure in-place mutation. Synths without `hydrateInto` fall back to plain `hydrate()` and the instance is replaced. (Note: the method runs before declared properties hydrate, so it should construct from config — client-synced state arrives via `hydrateInto` a moment later.)

**Update** — `hydrateForUpdate` passes root updates through `hydrateInto` on the live instance (the one from initial hydration) with the raw update value — identity and configured state survive client writes. Applies to declared properties too when their synth implements `hydrateInto` (`public Selection $selection` updates in place, matching the JS `sel` synth's in-place `merge`). Wrong-type writes hit the same `TypeError` → 419 path as typed declared props; `''`/`null` and `unset($this->selected)` re-initialize — the method runs again.

## Embedded, not tacked on

Because `all()`, `hasProperty()`, `fill()`, and `reset()` are virtual-aware, the surrounding framework picks virtuals up as if they'd always existed: `$this->validate()` sees them, island views get them, `$this->reset('selected')` re-runs the method, `only()`/`pull()`/`except()` just work.

## Verified

- **Browser tests**: two real-Chrome scenarios — method-built state hydrating into actual checkboxes, toggles syncing back via `wire:model`, `$refresh` round trips, server-side action mutations reaching the DOM, and except-mode + server-owned totals surviving round trips
- **Unit tests** (13): initialization, mount/action/view access, snapshot shape, guard rails, per-request re-construction (static counter), `reset()`, root-update instance identity, unset semantics, synth fallback, hostile-payload type safety
- Full unit suite green; `SupportSelection` browser suite green; real HTTP round trips against the playground app
- Perf: benchmarked at **+0.22–0.45%, within noise** for components without virtuals (details in review thread); discovery is a static per-class cache, per-request cost is an array lookup

## Reviewed

Two independent second-opinion passes (Codex for design/correctness, a security agent against the hostile-client threat model) — findings and fixes in [this comment](#issuecomment-5073646139). Five items fixed (direct-action bypass on non-camelCase virtual methods, declared-property in-place-update regression, total type/signature enforcement, validation-error persistence, `#[Locked]`/`#[Url]` targeting); the rest traced and confirmed safe.

## Lifecycle model (resolved after review)

Virtual properties **construct on first access, like `#[Computed]`** — never eagerly at boot. Since first-access is always after mount/hydration, a method can read sibling property state (`return new Report($this->filters)`), and a lazy placeholder doesn't run an expensive body it never touches. Dehydration counts as an access, so they're constructed and serialized into **every** snapshot (including placeholders) exactly like declared properties — no mid-lifecycle "materializes later" gap. Root updates apply to a **clone** of the method-built instance, so `updating`/`updated` hooks see the correct before/after values while the method's configuration survives the write. Contract: keep method bodies cheap constructors and defer real work into the returned object's callbacks (same as `#[Computed]`).

## Open questions

- **Mount params**: `<livewire:x :selected="...">` doesn't target virtual properties yet — params fall through to HTML attributes.
- **`Selection` mode strings**: `new Selection(mode: 'exclude')` silently falls back to include mode — might deserve a loud failure (separate PR).


src/Features/SupportPropertyFactories/UnitTest.php
```diff
@@ -0,0 +1,332 @@
...
+                $e = $e->getPrevious();
+            }
+
+            Assert::fail('Exception thrown, but none in the chain was ['.$class.'].');
+        }
+
+        Assert::fail('Expected exception ['.$class.'] was not thrown.');
+    }
+
+    function test_a_factory_property_survives_a_round_trip()
```
calebporzio: Im fine with most of the unit tests up until this point.

at this point (things that assert things between requests and such)

I'd be much more comfortable with a browser test for all of this.

I want to know that the real situation is happening: a real component dehydratesa nd mounts into a real JS runtime and a real refresh sends it all back and rehydrates and such.

src/Features/SupportVirtualProperties/UnitTest.php
```diff
@@ -0,0 +1,332 @@
...
+
+        $selected = $component->get('selected');
+
+        Assert::assertSame(['a', 'b'], $selected->keys());
+        Assert::assertSame(10, $selected->total());
+    }
+
+    function test_root_updates_are_hydrated_into_the_live_instance_rather_than_replacing_it()
+    {
+        Livewire::test(new class extends TestComponent {
```
calebporzio, line 302: this is an example of a test that i think can stay a unit test.

src/Features/SupportVirtualProperties/UnitTest.php
```diff
@@ -0,0 +1,332 @@
...
+                $this->selected->select($key);
+            }
+        })
+            ->call('pick', 'foo')
+            ->call('pick', 'baz')
+            ->assertSet('selected', fn ($selected) => $selected->keys() === ['foo', 'baz']);
+    }
+
+    function test_unsetting_a_factory_property_resets_it_to_a_fresh_factory_instance()
+    {
```
calebporzio, line 364: interesting. after you unset does it lazilly reset? like it waits until another access to refresh it? or somehow hook into unset?

src/Features/SupportPropertyFactories/SupportPropertyFactories.php
```diff
@@ -0,0 +1,71 @@
...
+use function Livewire\store;
+
+class SupportPropertyFactories extends ComponentHook
+{
+    static function provide()
+    {
+        // Rebuild every factory property at the start of each subsequent
+        // request: the factory method supplies the configured instance,
+        // then its synth hydrates the client's raw state into it...
+        on('hydrate', function ($component, $memo, $context) {
```
calebporzio: this all makes sense to me.

just want to take a minute to inquire about performance overhead any of this might add?

might be worth spawning a subagent to compare a large page with many livewire components that all have many properties page load before this PR and after kinda thing.

and think about low hanging perf fruit we might be missing.

fine if negligable just want to make sure.

src/Features/SupportPropertyFactories/SupportPropertyFactories.php
```diff
@@ -0,0 +1,71 @@
...
+        // Unsetting a factory property resets it back to a freshly
+        // constructed factory instance on next access...
+        on('__unset', function ($target, $property) {
+            $attribute = static::findFactoryAttribute($target, $property);
+
+            if ($attribute) store($target)->unset('propertyFactories', $property);
+        });
+    }
+
+    static function factoryAttributes($component)
```
calebporzio: i think i like the separatio nof having all these methods here, but an alternative implementation of course would be having a HandlesFactories trait to go on the compoennt instance and track everything directly there. but this is arguably better.

src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -259,6 +260,11 @@ public function fromSnapshot($snapshot)
 
         $context = new ComponentContext($component);
 
+        // Keep the raw dehydrated data around so features that manage
+        // properties without a backing declaration (property factories)
+        // can hydrate themselves from it later in the request...
+        $context->snapshotData = $data;
+
         $this->hydrateProperties($component, $data, $context);
```
calebporzio, line 262: just a thought, but you know the "if property doesn't exist, continue" in the loop inside hydrateProperties. What id instead of "continue" we did another check for the factory directly? Would it make any of this implementation more streamlined? more performant? more first-class principle instead of tacked-on-thing?

I don't want to do that if it's not a huge advantage though

src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -286,7 +292,10 @@ public function snapshot($component, $context = null)
 
     protected function dehydrateProperties($component, $context)
     {
-        $data = Utils::getPublicPropertiesDefinedOnSubclass($component);
+        $data = [
+            ...Utils::getPublicPropertiesDefinedOnSubclass($component),
+            ...SupportPropertyFactories::getFactoryProperties($component),
```
calebporzio: ahhh see, here we're leaking concerns from a standalone feature class into here. that's a smell.

src/Mechanisms/HandleComponents/HandleComponents.php
calebporzio: yeah and if we move factory hydration into hydrateproperties we don't need to store it on the context because it won't happen in a later step.

src/Features/SupportPropertyFactories/UnitTest.php
calebporzio: Done — moved the between-request coverage into a real `BrowserTest.php`: two scenarios with actual checkboxes wire:modeled to a factory Selection in Chrome. They cover: factory state hydrating into the browser (checkbox starts checked from factory config), a checkbox toggle syncing back, $refresh round trips (with a rendered `$runs` counter proving the factory re-runs each request), a server-side action mutation reaching the checkboxes, and except-mode + server-owned totals surviving round trips. The corresponding unit tests were removed.

src/Features/SupportVirtualProperties/UnitTest.php
calebporzio: Kept as a unit test 👍 (instance identity via spl_object_id is only observable server-side anyway).

src/Features/SupportVirtualProperties/UnitTest.php
calebporzio: It hooks into unset via the `__unset` magic + event (same channel computeds use for cache busting), and the reset is lazy: the handler just forgets the stored value, and the next access re-runs the factory. In practice "lazy" is short-lived — dehydration at the end of that same request reads the property to build the snapshot, so it springs back as a fresh factory instance by then at the latest. Setting it to `''`/`null` from the client behaves the same way (mirroring how typed declared props unset on empty-string).

src/Features/SupportPropertyFactories/SupportPropertyFactories.php
calebporzio: Benchmarking this now (large page, many components × many properties, before/after) — will post numbers here. Structurally the marginal cost for components *without* factories is small: `getAttributes()` was already computed every request by SupportAttributes (it's memoized per component instance), so the additions are just an iteration over that collection at the touch points. The `on('hydrate')` listener is gone entirely in the latest push — hydration moved into `hydrateProperties` per your other comment.

src/Features/SupportPropertyFactories/SupportPropertyFactories.php
calebporzio: This class is actually gone now — reworked per your hydrateProperties/smell comments: there's a core `VirtualProperty` interface in the mechanism namespace, `BaseFactory` implements it, and `HandleComponents` only knows the contract. No feature class, no trait, no bootFeatures registration — the attribute IS the integration.

src/Mechanisms/HandleComponents/HandleComponents.php
calebporzio: Did exactly this — and it turned out to be the key that un-smelled everything. `hydrateProperties` now offers unclaimed keys to the component's virtual properties directly in the loop (ordering is safe: virtual props always serialize after declared ones, and the snapshot is checksummed so the client can't reorder). `$context->snapshotData` is gone, and so is the separate `on('hydrate')` listener — factory hydration happens in the same pass as everything else.

src/Mechanisms/HandleComponents/HandleComponents.php
calebporzio: Agreed — fixed by making the concept first-class instead of the feature: new `VirtualProperty` interface lives in `Mechanisms/HandleComponents/` (a property with no backing declaration, managed by an attribute). `HandleComponents` consults only that contract at every touch point (dehydrate, view data, hydrate, update validation, writes, __get/__unset). `#[Factory]` is just its first implementation — the `SupportPropertyFactories` import is gone.

src/Features/SupportPropertyFactories/SupportPropertyFactories.php
calebporzio: Benchmark results are in (subagent ran this in isolated worktrees — `main` @ 9f2f53bb vs this branch, same PHP 8.4.6 binary, vendor identical):

**Workload**: a component with 22 public properties (strings, ints, arrays, 2 Collections), **no** factory attributes — 300 full mount+render+dehydrate cycles, plus 600 update requests (`set` + `$refresh`). One timed pass per fresh PHP process, 10 processes per side, alternating branches to cancel thermal drift.

| Workload | main (median) | branch (median) | delta |
|---|---|---|---|
| 300 mounts | 242.9 ms | 244.0 ms | +1.1 ms (**+0.45%**, ~4µs/mount) |
| 600 update requests | 545.1 ms | 546.3 ms | +1.2 ms (**+0.22%**, ~2µs/request) |

The run distributions overlap almost completely (spreads ±1.5%), so the deltas are within noise. On a page with 100 components this is on the order of 0.5–1.5ms total.

**Why it's structurally cheap**: `getAttributes()` was already computed every request by `SupportAttributes` (it's memoized per component, and every lifecycle hook already does a `whereInstanceOf()` over it that allocates more than `virtualProperties()` does) — so this adds zero new reflection. Instrumented call counts per no-factory component: mount = 2 calls, update = 3–4 calls, each a foreach over the (usually tiny) memoized collection.

**One piece of low-hanging fruit found and shipped**: the `on('__get')` listener runs on every magic property access — so a Blade view hammering `$this->someComputed` in a loop would rebuild the map N times. Fixed by memoizing `virtualProperties()` per attribute collection (WeakMap keyed on the collection instance — invalidation-safe because `mergeOutsideAttributes()` always replaces the collection). That path is now O(1) after first access.

src/Features/SupportVirtualProperties/HandlesVirtualProperties.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        $name = (string) str($name)->camel();
+
+        $this->__virtualProperties[$name] = $this->{static::virtualPropertyMethods()[$name]['name']}();
+    }
+
+    // Apply raw wire state to the freshly constructed instance. The synth
+    // mutates it in place via hydrateInto so everything the method
+    // configured on it (closures included) survives the trip...
+    function hydrateVirtualProperty($name, $valueOrTuple, $context)
```
calebporzio: not sure i love this here. i feel like the synth should get what it needs from thne virtual property and then call hydrateInto itself? idk... but synth stuff is so tender that i don't like spreading it around the codebase

src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -39,6 +39,8 @@ public function mount($name, $params = [], $key = null, $slots = [])
 
         $this->synths->initializeProperties($component);
 
+        $component->initializeVirtualProperties();
```
calebporzio: this is an asymetry. 

i suppose we're trading twi things off here:

we could either:

A) put virtual property initialzation (or some of it) into the $this->synths->initailizeProperties() method. (which feels right when looking at this code). but it would move virtual property concerns outside the trait where they currently live. but it feels like a first-party enough feature that maybe it should?

src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -286,7 +288,10 @@ public function snapshot($component, $context = null)
 
     protected function dehydrateProperties($component, $context)
     {
-        $data = Utils::getPublicPropertiesDefinedOnSubclass($component);
+        $data = [
+            ...Utils::getPublicPropertiesDefinedOnSubclass($component),
+            ...$component->getVirtualProperties(),
```
calebporzio: there's an aymmetry. might be fine. but it's worth investigating.

src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -298,7 +303,16 @@ protected function dehydrateProperties($component, $context)
...
     {
         foreach ($data as $key => $value) {
-            if (! property_exists($component, $key)) continue;
+            if (! property_exists($component, $key)) {
+                // A key with no backing declaration may be a virtual
+                // property (a #[Virtual] method) — the method constructs
+                // the instance, then the raw wire value hydrates INTO it.
+                // Virtual keys always serialize after declared ones, so
+                // the method runs against fully-hydrated state...
+                if ($component->hasVirtualProperty($key)) $component->hydrateVirtualProperty($key, $value, $context);
```
calebporzio: like my earlier note. i don't think hydrateVirtualProperty should be called here you know? its contents should be exploded out a lil bit

src/Features/SupportVirtualProperties/HandlesVirtualProperties.php
calebporzio: Agreed and gone — `hydrateVirtualProperty` is out of the trait entirely, so the trait no longer touches anything synthy. The hydration now lives inline in `hydrateProperties`, flowing through the synth coordinator exactly like declared properties do (`$this->synths->hydrateInto(...)` + `setVirtualProperty(...)`, mirroring the `$this->synths->hydrate(...)` + assignment right below it). All hydrateInto logic is contained in `HandleSynths`.

src/Mechanisms/HandleComponents/HandleComponents.php
calebporzio: Went with A — `HandleSynths::initializeProperties()` now ends with `$component->initializeVirtualProperties()`, so declared-typed synth init and virtual init are one initialization step, and the extra line in `mount()` is gone. Bonus symmetry: since `fromSnapshot` calls the same method, the hydrate path now boots virtuals eagerly exactly like the mount path (your original spec — every request: construct fresh, then hydrate raw state INTO it). One consequence worth knowing: the method runs before declared properties hydrate on update requests, so it should construct from config, not from client-synced state — anything client-synced arrives via hydrateInto a moment later.

src/Mechanisms/HandleComponents/HandleComponents.php
calebporzio: This asymmetry (and the constant declared-vs-virtual double-reach) is now solved with one abstraction: `$component->all()` returns declared + virtual, and it's THE way everything enumerates properties. `dehydrateProperties` is literally `$data = $component->all();`, `getView` and `updateProperty` validation use it too — and switching `getDataForValidation()` and island scope to `all()` made virtuals validatable and available in island views for free. `fill()`/`reset()`/`only()`/`pull()` are virtual-aware now as well (`reset('selected')` re-runs the method — a nice semantic fit).

src/Mechanisms/HandleComponents/HandleComponents.php
calebporzio: Exploded out — the loop now reads symmetrically with the declared path: virtual branch does `$child = $this->synths->hydrateInto($component->getVirtualProperty($key), $value, $context, $key)` then `$component->setVirtualProperty($key, $child)`; declared branch does `$child = $this->synths->hydrate(...)` then assignment. Trait = lookup + access only.

calebporzio:
## Second-opinion review pass (Codex + a security-focused audit)

Ran two independent reviews over the whole branch — Codex for design/correctness, and a security agent against Livewire's hostile-client threat model. Both traced the actual code (not just the diff). Here's what came back and what I did.

### Fixed

**1. Direct-action bypass for non-camelCase virtual methods** (both reviewers, MEDIUM). The `#[Virtual]` call-block matched on the *camelCased property name*, but the action dispatcher matches on the *raw method name* — so a `#[Virtual] public function select_all()` slipped past the guard and was invokable as an action with attacker `params`, returning its value in the `returns` effect. Fixed by filtering the raw virtual method names out of the callable-action allowlist in `callMethods` (alongside `render`), which is casing-agnostic. Regression test added with a snake_case method.

**2. In-place root updates mutating DECLARED properties before hooks/locks** (Codex, merge blocker). My generalized in-place `hydrateInto` branch fired for declared `public Selection` too, mutating the live instance before `update`/`Locked` hooks ran — a behavior change for existing properties. Scoped the branch to virtual properties only (`hasVirtualProperty($path)`); declared properties revert to their exact main-branch replace-then-assign path.

**3. Incomplete type/signature enforcement** (both, LOW). Builtin/union return types skipped the write-time guard; methods with required params were discovered then died with `ArgumentCountError`; non-public methods were accepted. Since virtual properties exist to construct *objects* that don't serialize well, discovery now requires a concrete class return type, public visibility, and zero required params — which makes the write-time guard a total `instanceof`, no builtin/union/void gaps. Duplicate canonical names (`selected_items` + `selectedItems`) now throw instead of silently last-winning.

**4. Validation errors on virtual properties discarded** (Codex). Error dehydration filtered keys through `Utils::hasProperty` (`property_exists`-only), dropping virtual-property errors from the snapshot memo. Switched to the component's virtual-aware `hasProperty()`. Errors now survive round trips (test added).

**5. `#[Locked]`/`#[Url]` silently no-op on a method** (Codex). Both were declared with bare `#[\Attribute]`, so putting `#[Locked]` on a `#[Virtual]` method gave a false sense of protection. Restricted both to `TARGET_PROPERTY` so it fails loudly (test added). Small related hardening — happy to split it out if you'd prefer.

### Confirmed safe (traced, not assumed)

Deep/dotted-path updates against a virtual root (base `Synth::__call` throws on the missing `set`), `SecurityPolicy` (the in-place branch never instantiates from client meta — the instance is server-built; meta rides the checksummed snapshot), locked/reactive/query-string cross-shadowing (exact-name collisions are impossible by the discovery guard, and property attributes can't target methods), and the discovery cache under Octane (bounded by component count, one class name per literal).

### Two documented limitations (want your call before I build machinery)

- **`updating` hook ordering for virtual props**: a virtual-property update mutates the live instance in `hydrateForUpdate`, so `updatingSelected()` sees post-update state, unlike a declared object property (which sees the old instance until assignment). No security impact (virtual props can't be locked). A correct fix means deferring the in-place mutation past `trigger('update')` — a two-phase hydrate. Worth it, or document as intended?
- **Lazy placeholder eager construction**: virtual methods run in `initializeProperties`, before the lazy `skipMount` decision — so an expensive/side-effecting constructor runs during the placeholder request (and again on the real load). This mirrors how synth-initialized typed properties already behave, but unlike `mount()` (which lazy *does* skip). Proper fix defers virtual init past the lazy decision and omits virtuals from the placeholder snapshot — that touches lazy internals, so I'd rather you weigh in than have me reshape it unilaterally.

Full unit suite green (1412), plus the SupportSelection and SupportLazyLoading browser suites, plus real HTTP round trips against the playground app.

calebporzio:
## Both open questions resolved

**1. `updating` hook ordering — fixed (clone-on-update).** A root update now applies to a `clone` of the method-built instance rather than mutating it in place. The original stays put during `trigger('update')`, so `updatingSelected()` sees the old value and `updatedSelected()` sees the new — matching declared and nested-object update semantics. The clone carries the method's configuration/closures, so config still survives the write (that was always the real goal — instance *identity* was never the requirement). Tested.

**2. Lazy placeholder eager construction — fixed (lazy-on-access + always-dehydrate).** Virtual properties now construct on first access (like `#[Computed]`) instead of eagerly at boot. Two payoffs beyond the lazy fix:
- Because first-access is always after mount/hydration, a virtual method can now **read sibling property state** (e.g. `return new Report($this->filters)`) — impossible under eager construction, which ran before mount. Tested.
- A lazy component's placeholder no longer runs an expensive method body it never touches.

We deliberately **did not** skip virtuals from the placeholder snapshot — dehydration still constructs and serializes them every request, so they're present in every snapshot (including placeholders) exactly like declared properties. That avoids introducing a "data that materializes mid-lifecycle" gap with no precedent, and keeps the client-side view of them consistent. The one assumption is that method bodies are cheap constructors with expensive work deferred into the returned object's callbacks — same contract as `#[Computed]`, and documented as such.

Verified: 21 unit tests (incl. hook-ordering and sibling-state), 3 browser tests (incl. a lazy + virtual round trip in real Chrome), full unit suite (1414) and the SupportLazyLoading/SupportSelection browser suites all green.

src/Concerns/InteractsWithProperties.php
```diff
@@ -35,7 +37,13 @@ public function fill($values)
 
         foreach ($values as $key => $value) {
             if (in_array(Utils::beforeFirstDot($key), $publicProperties)) {
-                data_set($this, $key, $value);
+                // Root writes to virtual properties land in their lookup —
```
calebporzio: remove comment

src/Concerns/InteractsWithProperties.php
```diff
@@ -56,6 +64,14 @@ public function reset(...$properties)
         foreach ($properties as $property) {
             $property = str($property);
 
+            // Virtual properties reset by re-initializing — the method
```
calebporzio: remove comment

src/Features/SupportSelection/SelectionSynth.php
```diff
@@ -46,6 +46,21 @@ function dehydrate($target)
         ], $meta];
     }
 
+    // Apply the client's raw state to an existing instance in place
+    // rather than building a fresh one — anything construction configured
+    // on it survives (this is how virtual properties round-trip)...
```
calebporzio: i think you can scratch this comment.

src/Features/SupportValidation/SupportValidation.php
```diff
@@ -58,9 +58,11 @@ function dehydrate($context)
 
         // Only persist errors that were born from properties on the component
         // and not from custom validators (Validator::make) that were run.
+        // The component's own hasProperty() is virtual-property aware, so
+        // errors on virtual properties survive to the next request too...
```
calebporzio: scratch this comment addition (keep what comment was previously there.)

src/Features/SupportVirtualProperties/HandlesVirtualProperties.php
```diff
@@ -0,0 +1,164 @@
...
+    function setVirtualProperty($name, $value)
+    {
+        $name = (string) str($name)->camel();
+
+        $method = static::virtualPropertyMethods()[$name];
+
+        // Writes must land as the class the method promised — the same
+        // contract a typed public property enforces on assignment. The
+        // return type is always a concrete class (enforced at discovery),
+        // so this instanceof check is total...
```
calebporzio: seems like an incomplete comment. please rewrite/evaluate/ or remove

src/Features/SupportVirtualProperties/HandlesVirtualProperties.php
```diff
@@ -0,0 +1,164 @@
...
+            throw new \TypeError(
+                'Virtual property ['.$name.'] on component ['.static::class.'] must be an instance of ['.$method['type'].'].'
+            );
+        }
+
+        $this->__virtualProperties[$name] = $value;
+    }
+
+    // Unsetting re-initializes: the method runs again and the fresh
+    // instance takes the old one's place in the lookup...
```
calebporzio: useless comment?

src/Features/SupportVirtualProperties/HandlesVirtualProperties.php
calebporzio: Done.

src/Concerns/InteractsWithProperties.php
calebporzio: Removed.

src/Concerns/InteractsWithProperties.php
calebporzio: Removed.

src/Features/SupportSelection/SelectionSynth.php
calebporzio: Removed — the hydrateInto contract is documented once on `HandleSynths::hydrateInto`; the per-synth method is self-evident.

src/Features/SupportValidation/SupportValidation.php
calebporzio: Reverted to the original comment; my addition is gone.

src/Features/SupportVirtualProperties/HandlesVirtualProperties.php
calebporzio: Rewritten: "Discovery guarantees a concrete class return type, so a plain instanceof covers every case — no builtin/union gaps to handle." (The point was: no builtin/union branch is needed because the return type is always a concrete class.)

src/Features/SupportVirtualProperties/HandlesVirtualProperties.php
calebporzio: Reduced to one line stating the non-obvious bit — unsetting a virtual property reconstructs it rather than removing it (it has no empty state). Happy to drop entirely if you'd still rather.

PR #10446, opened 2026-07-24
[4.x] Refactor `SupportLifecycleHooks` to use existing helper methods

ghabriel25:
Leftover from this PR https://github.com/livewire/livewire/pull/10442


joshhanley:
Thanks!

PR #10447, opened 2026-07-25
[4.x] Rename internal `WithFileUploads` trait to `HandlesFileUploads`

ghabriel25:
# Why
`WithFileUploads` trait exists in two different directory
- `src\Features\SupportFileUploads\WithFileUploads.php`
- `src\WithFileUploads.php`

This can be confusing from IDE perpective when importing the trait (See image below)

image

Yes, its well documented but I think its a good to be explicitly use different name so I rename it to `HandlesFileUploads`. This is internal change so there is no bc at all.

However, it still has some concern that some people already use this internal trait but we already agreed that internal class, trait etc should be strictly not used as public API.

image

# Whats Changed
- Rename `src\Features\SupportFileUploads\WithFileUploads.php` to `src\Features\SupportFileUploads\HandlesFileUploads.php`
- Change the imported trait in `src\WithFileUploads.php` to use this new namespace.

### DIFF
--- a/src/Features/SupportFileUploads/HandlesFileUploads.php
+++ b/src/Features/SupportFileUploads/HandlesFileUploads.php
@@ -7,7 +7,7 @@ use Illuminate\Http\UploadedFile;
 use Livewire\Attributes\Renderless;
 use Livewire\Facades\GenerateSignedUploadUrlFacade;
 
-trait WithFileUploads
+trait HandlesFileUploads
 {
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for the PR! I’m going to close this because renaming the internal upload trait for IDE clarity doesn’t provide enough value to justify the churn and potential impact on people who may already reference it.

PR #10448, opened 2026-07-25
[4.x] Refactor `SupportIsolating` to use `instanceof` instead `is_subclass_of`

ghabriel25:
Inside `SupportIsolating::shouldIsolate()` still using `is_subclass_of` instead `instanceof`. All `Base*` attribute are strictly for internal uses, however there is a chance future contributors not aware of this and creating browser test using base attribute.

This is to ensure that both from public API use case and tests align without any error that some people not aware of

# Whats Changed
- Change `is_subclass_of` to `instanceof`
- Added regression browser test


calebporzio:
Thanks for the PR! I’m going to close this because the behavior difference only applies when directly using the internal `BaseIsolate` attribute, rather than the supported public attribute. That doesn’t justify changing the existing implementation.

PR #10450, opened 2026-07-26
Boot form objects returned from #[Virtual] property methods

calebporzio:
## The bug

A `Form` object returned from a `#[Virtual]` method never gets booted. Its `#[Validate]`, `#[Url]`, and `#[Locked]` attributes are never registered on the component, and its `boot()` never runs.

```php
public DeclaredForm $form;              // works

#[Virtual]
public function draft(): VirtualForm    // silently broken
{
    return new VirtualForm($this, 'draft');
}
```

Verified over real HTTP against a playground app — the declared form behaves, the virtual one doesn't:

| Behavior | Declared | Virtual (before) |
|---|---|---|
| `boot()` | runs | never runs |
| `#[Validate]` on `validate()` | errors | `MissingRulesException` |
| Validate-on-update | fires | silent |
| `#[Url]` read + effect | both | both dead |
| `#[Locked]` vs. hostile write | throws | **overwritten, HTTP 200** |
| Consolidated whole-form update | expands per-field | not expanded |

The `#[Locked]` bypass is the sharp edge: moving a form from a declared property to a virtual one silently dropped a security contract.

## Why

**A birth path the synth never sees.** `FormObjectSynth::bootFormObject()` — which merges the form's attributes onto the component and runs its `boot()` — is only reachable from `initialize()` and `hydrate()`'s fresh-construction branch. A `#[Virtual]` method is a third way a form comes into existence, and on later requests `hydrate()` finds the just-materialized instance and takes the *reuse* branch, which skips boot.

**Attributes live in one-shot windows.** `SupportAttributes` iterates the collection at `trigger('mount')`/`trigger('hydrate')` and calls each attribute's lifecycle hook. Anything merged after that window silently gets none. Virtual properties materialize lazily — by design, so a method can read sibling state set in `mount()` — which on the initial request is always after the window has closed.

## The fix

- **`HandleSynths::adopt()`** — a synth may finish constructing a value born in user code, opting in via `method_exists` exactly like `hydrateInto()`. `FormObjectSynth::adopt()` runs the existing boot flow. All virtual construction funnels through one `constructVirtualProperty()` that stores the value *before* adopting, so re-entrant reads don't reconstruct.
- **Lifecycle-window catch-up** — `SupportAttributes` records each one-shot window (before iterating, so a mid-window merge doesn't double-run); `mergeOutsideAttributes()` replays missed windows on late arrivals only. Arrival time stops changing behavior — here and for every future late merge (`setPropertyAttribute()` mid-action, SupportPagination's hand-patched query-string call).
- **`Attribute::setValue` writes through its sub-target** — routing through the component would re-enter `__get` mid-write, where PHP's magic-method recursion guard bypasses `__get` and `data_set` shadows the materializing virtual with a plain array. Reachable today via `#[Url]` + a query param, so client-influenceable.
- **Consolidated updates** resolve virtual forms (exact key match — `hasVirtualProperty()` camelCases, which would let `Form` alias onto `form`) and skip fields echoed back identical to their *snapshot* value, compared as the client sees them (one number type, synth tuples unwrapped) so filler never trips `#[Locked]` while every real change still expands and gets gated.
- **Boot is per-instance** — a method that memoizes its form plus a `reset()` would otherwise merge its attributes twice and double its rules; a reconstructed form at the same path replaces its predecessor's attributes rather than accumulating them.

Deliberately *not* added: `FormObjectSynth::hydrateInto()`. Forms are identity-stable by design ("form objects should never be fully replaced") — a clone-and-replace root update would orphan every merged attribute's sub-target binding.

No new user-facing API. A virtual form simply behaves like a declared one.

## Verified

- 13 new tests in `SupportVirtualProperties/UnitTest.php`, each written red first (boot-once, `#[Validate]` at rest and on update, `#[Locked]` on granular and consolidated paths, `#[Url]` read + effect, first-access-inside-`mount()` recursion-guard corruption, memoized-instance reset, JSON-lossy locked float echo, aliased-path resolution)
- Full unit suite green (1427), and the touched feature suites are order-stable across several random seeds
- Real-Chrome browser suites for virtual properties, form objects, query string, validation, locked properties, selection, pagination, lazy loading
- Real HTTP wire requests against a playground app: `boot()` runs exactly once per request, `?vq=` lands in the form and registers its URL effect, validation errors surface, and the locked-property tamper now throws instead of returning 200
- ~70 adversarial probes from independent reviewers (hostile client, lifecycle, integration), including a pre-fix baseline build confirming the `#[Locked]` bypass was live before and closed after

## Follow-ups (not in this PR)

- Pre-existing `FormObjectSynth` 500s, declared forms included: a consolidated update with an unknown key set to `null` hits a `ReflectionException` in `propertyIsTyped()`; root `set('form', null)`/`''` hits a `foreach` TypeError in `hydrateFormProperties()`. Two small guards would make these clean rejections.
- `SupportPagination`'s manual `setPropertyFromQueryString()` call exists to paper over the missed mount window that catch-up now heals generally — deletable, best with its own browser-test pass.
- `Form::toArray()` uses reflection-only `getPublicProperties()`, so a `#[Virtual]` property *on a form* dehydrates nowhere. Half-wired today; worth a deliberate decision.

### DIFF
--- a/src/Features/SupportAttributes/Attribute.php
+++ b/src/Features/SupportAttributes/Attribute.php
@@ -77,7 +77,16 @@ abstract class Attribute
             }
         }
 
-        data_set($this->component, $this->levelName, $value);
+        // Write straight onto the sub-target when there is one. Routing
+        // through the component would re-enter Component::__get mid-write —
+        // and if the property is a #[Virtual] being materialized right now,
+        // PHP's magic-method recursion guard bypasses __get entirely and
+        // data_set would shadow it with a plain array property...
+        if ($this->subTarget) {
+            data_set($this->subTarget, $this->subName, $value);
+        } else {
+            data_set($this->component, $this->levelName, $value);
+        }
     }
 
     protected function tryingToSetStringOrIntegerToEnum($subject)
--- a/src/Features/SupportAttributes/HandlesAttributes.php
+++ b/src/Features/SupportAttributes/HandlesAttributes.php
@@ -2,6 +2,8 @@
 
 namespace Livewire\Features\SupportAttributes;
 
+use function Livewire\store;
+
 trait HandlesAttributes
 {
     protected AttributeCollection $attributes;
@@ -21,5 +23,24 @@ trait HandlesAttributes
     function mergeOutsideAttributes(AttributeCollection $attributes)
     {
         $this->attributes = $this->getAttributes()->concat($attributes);
+
+        // Attributes can arrive after a one-shot lifecycle window has
+        // already run (a #[Virtual] form object materializing mid-request).
+        // Replay the missed windows on just the late arrivals so arrival
+        // time never changes behavior...
+        $arrivals = $attributes->whereInstanceOf(Attribute::class);
+
+        foreach (store($this)->get('attributeWindows', []) as [$window, $params]) {
+            $arrivals->each(function ($attribute) use ($window, $params) {
+                if (method_exists($attribute, $window)) {
+                    $attribute->$window(...$params);
+                }
+            });
+        }
+    }
+
+    function forgetAttributesWhere($callback)
+    {
+        $this->attributes = $this->getAttributes()->reject($callback)->values();
     }
 }
--- a/src/Features/SupportAttributes/SupportAttributes.php
+++ b/src/Features/SupportAttributes/SupportAttributes.php
@@ -9,6 +9,8 @@ class SupportAttributes extends ComponentHook
 {
     function boot(...$params)
     {
+        $this->recordWindow('boot', $params);
+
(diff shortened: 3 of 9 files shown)


calebporzio:
Superseded-in-spirit by #10451, which rebuilds this from scratch on review feedback: no synth `adopt()`, no lifecycle-window record/replay, no `setValue` reroute, no snapshot-echo comparison — 126 source lines against 441 here.

It buys that by accepting one constraint: `#[Virtual]` methods construct alongside declared typed properties, so a virtual method can't read a mount parameter or anything `mount()` assigns. Same limitation declared form objects already have.

Take one or the other, not both.

PR #10453, opened 2026-07-27
[4.x] Extract payload & structure validation into dedicated `PayloadGuards` class

ghabriel25:
## Motivation

The validation blocks were becoming noisy and mixed concerns with the core request/component handling logic. Moving pure guard logic into a dedicated class keeps the happy path clean while preserving identical behavior (same exceptions, same abort codes, same config keys).

## Summary

Extracts all request-payload and component-update validation/guard logic into a dedicated `PayloadGuards` class.

Previously this logic was scattered across `HandleRequests::handleUpdate()` and `HandleComponents::update()` / `callMethods()`. Centralizing it improves readability of the main request & component handling flow.

### Added
- `src/Mechanisms/HandleRequests/PayloadGuards.php`
  - `verify()` – payload size limit, components array structure, max components limit
  - `verifyPayloadMaxCalls()` – max calls limit
  - `verifySnapshotStructure()` – snapshot shape validation
  - `verifyCallsStructure()` – calls shape validation

### Updated
- `HandleRequests.php` – replaced the large inline validation block with `PayloadGuards::verify()`
- `HandleComponents.php` – delegated snapshot structure, calls structure, and max-calls checks to `PayloadGuards`

## Notes

- Pure refactor – no intentional behavior changes.
- All existing config keys (`livewire.payload.max_size`, `max_components`, `max_calls`) and exception classes remain unchanged.

### DIFF
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -8,10 +8,10 @@ use Livewire\Mechanisms\HandleSynths\HandleSynths;
 use Livewire\Exceptions\PublicPropertyNotFoundException;
 use Livewire\Exceptions\MethodNotFoundException;
 use Livewire\Exceptions\MaxNestingDepthExceededException;
-use Livewire\Exceptions\TooManyCallsException;
 use Livewire\Drawer\Utils;
 use Livewire\Features\SupportFormObjects\Form;
 use Illuminate\Support\Facades\View;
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks for the PR! I’m going to close this because it moves existing payload validation into a new abstraction without changing behavior. Since this is security-sensitive request handling, I’d rather leave the tested code in place unless a concrete bug requires changing it.

PR #10456, opened 2026-07-27
[4.x] Handle empty successful responses

### LINKED ISSUE #10424: Empty error modal + no retry when iOS Safari 18+ resolves a Livewire POST with a zero-byte 200 (dead keep-alive reuse)
### Livewire version
v4.3.3

### What happened

Safari 18+ (iOS and macOS) has a network regression: when it reuses a pooled keep-alive connection that the server or edge closed while the tab was suspended (screen off, tab backgrounded, Safari closed and reopened), it does not retry the request the way Chrome, Firefox, and Safari 17 do. For GETs Safari retries safe methods, so page loads recover. For POSTs it silently resolves `fetch()` with a **200 whose body is zero bytes**. Since every Livewire action is a POST to the update endpoint, Livewire apps are disproportionately affected: the first action after a phone comes back from idle fails this way.

References for the WebKit regression:
- https://discussions.apple.com/thread/256112607 (includes Wireshark analysis)

1. `response.text()` resolves to `""`, `JSON.parse("")` throws `SyntaxError: JSON Parse error: Unexpected EOF`.
2. The catch routes to `handlers.error` -> `showHtmlModal(responseBody)`, which renders the zero-byte body: a **completely blank full-screen dialog** with no diagnostic value for the user or the developer.
3. `rejectActionPromises` rejects action promises nothing awaits, so the console also fills with `Unhandled Promise Rejection` noise.

### Reproduction (no Safari needed)

The blank-modal handling reproduces in any browser by making the update endpoint return what Safari manufactures. Minimal Laravel middleware:

```php
class SimulateDroppedLivewireResponses
{
  public function handle(Request $request, Closure $next): Response
  {
      if ($request->isMethod('POST')
          && str_ends_with((string) $request->route()?->getName(), 'livewire.update')
          && Cache::add('dropped:'.$request->ip(), 1, 10)) {
          return response('', 200, ['Content-Type' => 'application/json']);
      }

      return $next($request);
  }
}
```

Append it to the `web` group, click any `wire:click` button: blank modal, parse error, unhandled rejections. The next request inside 10s passes, mimicking the real pattern (only the first request after an idle gap rides the dead socket).

### Suggested improvements

1. **Treat an ok-but-empty body as a transport failure, not an error page.** A 2xx with zero bytes can never be a real Livewire payload; showing it in `showHtmlModal` renders a blank dialog. It should take the failure path (or its own path) instead.
2. **Retry once on transport-level failure.** Browsers deliberately never auto-retry non-idempotent requests, so a framework whose whole protocol is POST needs its own answer to dead-socket reuse. A single retry when `fetch()` rejects or resolves 2xx-with-empty-body is safe (the request never produced a response) and fixes this class of failure for every Livewire app on iOS. We shipped exactly this as a `window.fetch` wrapper in production and verified it A/B against the middleware above: unpatched -> blank modal; patched -> transparent retry, action completes.
3. **Route action-promise rejections somewhere catchable** so transport failures do not surface as unhandled rejections.

joshhanley:
# The Scenario

Sometimes when Safari resumes after being suspended, a Livewire update request can resolve with a successful status but an empty response body. The action then appears to fail, and Livewire displays a completely blank error modal.

This can affect any Livewire action, including ordinary `wire:click` interactions.

CleanShot 2026-07-27 at 21 18 03

```blade
<?php

use Livewire\Component;

new class extends Component {
    public int $count = 0;

    public function increment(): void
    {
        $this->count++;
    }
};
?>

<div>
    <p>{{ $count }}</p>

    <button wire:click="increment">Increment</button>
</div>
```

# The Problem

Livewire treats the response as successful because its HTTP status is `200`, then attempts to parse the empty body as JSON.

When parsing fails, the response is passed to the normal HTML error handler. Because the body is empty, `showHtmlModal()` renders a blank full-screen dialog with no useful information or recovery path.

Automatically retrying the request is unsafe because Livewire cannot determine whether the action reached the server before its response was lost.

# The Solution

The solution is to detect an empty successful response before attempting to parse it as JSON and route it through Livewire's existing error-response handling.

The default handling displays a concise confirmation (similar to 419) asking the user whether they want to refresh the page. Refreshing restores the latest server state without repeating an action that may already have completed.

CleanShot 2026-07-27 at 21 15 55

Request interceptors receive the response first and can call `preventDefault()` to suppress or replace the default confirmation.

Fixes #10424

### DIFF
--- a/js/request/index.js
+++ b/js/request/index.js
@@ -388,6 +388,12 @@ function sendMessages() {
                     return
                 }
 
+                if (response.ok && responseBody.trim() === '') {
+                    confirm('The server’s response was lost, refresh?') && window.location.reload()
+
+                    return
+                }
+
                 if (response.aborted) return
 
                 showHtmlModal(responseBody)
@@ -527,6 +533,13 @@ async function sendRequest(request, handlers) {
 
     if (response.redirected) {
         handlers.redirect(response.url)
+        handlers.finish()
+
+        return
+    }
+
+    if (responseBody.trim() === '') {
+        handlers.error({ response, responseBody })
         handlers.finish()
 
         return
--- a/js/request/index.js
+++ b/js/request/index.js
@@ -389,7 +389,9 @@ function sendMessages() {
                 }
 
                 if (response.ok && responseBody.trim() === '') {
-                    confirm('The server’s response was lost, refresh?') && window.location.reload()
+                    confirm(
+                        'The server returned an empty response.\nWould you like to refresh the page?'
+                    ) && window.location.reload()
 
                     return
                 }
--- a/src/Features/SupportErrorResponses/BrowserTest.php
+++ b/src/Features/SupportErrorResponses/BrowserTest.php
@@ -74,7 +74,7 @@ class BrowserTest extends \Tests\BrowserTestCase
         Livewire::visit(Component::class)
             ->click('@refresh')
             ->waitForDialog()
-            ->assertDialogOpened('The server’s response was lost, refresh?')
+            ->assertDialogOpened("The server returned an empty response.\nWould you like to refresh the page?")
             ->dismissDialog()
             ->pause(100)
             ->assertMissing('#livewire-error')
--- a/src/Features/SupportErrorResponses/BrowserTest.php
+++ b/src/Features/SupportErrorResponses/BrowserTest.php
@@ -2,6 +2,7 @@
 
 namespace Livewire\Features\SupportErrorResponses;
 
+use Illuminate\Support\Facades\Route;
 use Livewire\Component as BaseComponent;
 use Livewire\Livewire;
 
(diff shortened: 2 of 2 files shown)


jonathan-bird:
@joshhanley I agree this is the best solution. Users can also hook into any aborted and set their own retry too like I have:  https://github.com/livewire/livewire/issues/10424#issuecomment-5020989027

Up to you if you want to handle it like this too with `.text()` and header checks, or whether your `.trim()` is sufficient:

```
const emptyBody = response.ok
            && ! response.headers.has('X-Livewire-Stream')
            && (await response.clone().text()).trim() === '';
```

pkoury:
This is also happening in Safari when using wire:navigate and the user leaves the page before a request finishes. So after this PR the user would see the "response was lost" dialog.

calebporzio:
only nit here is the confirm message. Is there a precedent for it? maybe ask an llm or two what it should be and compare responses? thanks

PR #10460, opened 2026-07-28
Fix _removeUpload for Collection properties

### LINKED ISSUE #10459: _removeUpload silently fails when file upload property is a Collection
### Description

When using `WithFileUploads` with a property set to an `Illuminate\Support\Collection` (e.g. `public Collection $photos`), calling `_removeUpload` does nothing. The temporary file is not deleted from storage, the `upload:removed` event is not dispatched, and the file remains in the collection.

`_finishUpload` and `validateIncomingUpload` both handle `Collection` instances, but `_removeUpload` currently only checks `is_array($uploads)`.

### Reproduction

```php
use Livewire\Component;
use Livewire\WithFileUploads;
use Illuminate\Support\Collection;

class UploadPhotos extends Component
{
    use WithFileUploads;

    public Collection $photos;

    public function mount()
    {
        $this->photos = collect();
    }
}
```

1. Upload multiple files into `$photos`. `_finishUpload` successfully merges them into the `Collection`.
2. Click to remove one of the uploaded files from the UI (which calls `$wire._removeUpload('photos', tmpFilename)`).

### Expected Behavior

The specified temporary file should be deleted from storage, removed from the `$photos` collection, and the `upload:removed` event should be dispatched.

### Actual Behavior

`_removeUpload` checks `is_array($uploads)` which evaluates to `false` for a `Collection`. The method exits without doing anything.

### Context / Cause

In `Livewire\Features\SupportFileUploads\WithFileUploads.php`:

```php
function _removeUpload($name, $tmpFilename)
{
    $uploads = $this->getPropertyValue($name);

    if (is_array($uploads) && isset($uploads[0]) && $uploads[0] instanceof TemporaryUploadedFile) {
        // ...
    }
}
```

`$uploads instanceof Collection` is not handled here.

hamdyelbatal122:
Fixes #10459

### Summary

When storing multiple file uploads in a `Collection` property (e.g. `public Collection $photos`), calling `_removeUpload` would fail silently because it only checked `is_array($uploads)`.

`_finishUpload` and `validateIncomingUpload` both support `Collection` instances, but `_removeUpload` was missing it.

### Changes

- Added `Collection` support to `_removeUpload` in `WithFileUploads.php` while preserving the property's original `Collection` type when updated.
- Added a unit test to verify removing temporary uploads from a `Collection` property.

### DIFF
--- a/src/Features/SupportFileUploads/WithFileUploads.php
+++ b/src/Features/SupportFileUploads/WithFileUploads.php
@@ -134,17 +134,25 @@ trait WithFileUploads
     {
         $uploads = $this->getPropertyValue($name);
 
-        if (is_array($uploads) && isset($uploads[0]) && $uploads[0] instanceof TemporaryUploadedFile) {
-            $this->dispatch('upload:removed', name: $name, tmpFilename: $tmpFilename)->self();
+        $isCollection = $uploads instanceof \Illuminate\Support\Collection;
 
-            app('livewire')->updateProperty($this, $name, array_values(array_filter($uploads, function ($upload) use ($tmpFilename) {
-                if ($upload->getFilename() === $tmpFilename) {
-                    $upload->delete();
-                    return false;
-                }
+        if (is_array($uploads) || $isCollection) {
+            $items = $isCollection ? $uploads->all() : $uploads;
 
-                return true;
-            })));
+            if (isset($items[0]) && $items[0] instanceof TemporaryUploadedFile) {
+                $this->dispatch('upload:removed', name: $name, tmpFilename: $tmpFilename)->self();
+
+                $filtered = array_values(array_filter($items, function ($upload) use ($tmpFilename) {
+                    if ($upload->getFilename() === $tmpFilename) {
+                        $upload->delete();
+                        return false;
+                    }
+
+                    return true;
+                }));
+
+                app('livewire')->updateProperty($this, $name, $isCollection ? collect($filtered) : $filtered);
+            }
         } elseif ($uploads instanceof TemporaryUploadedFile && $uploads->getFilename() === $tmpFilename) {
             $uploads->delete();
 
--- a/src/Features/SupportFileUploads/WithFileUploads.php
+++ b/src/Features/SupportFileUploads/WithFileUploads.php
@@ -136,23 +136,21 @@ trait WithFileUploads
 
         $isCollection = $uploads instanceof \Illuminate\Support\Collection;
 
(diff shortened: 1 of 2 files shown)


src/Features/SupportFileUploads/WithFileUploads.php
```diff
@@ -134,17 +134,25 @@ function _removeUpload($name, $tmpFilename)
     {
         $uploads = $this->getPropertyValue($name);
 
-        if (is_array($uploads) && isset($uploads[0]) && $uploads[0] instanceof TemporaryUploadedFile) {
-            $this->dispatch('upload:removed', name: $name, tmpFilename: $tmpFilename)->self();
+        $isCollection = $uploads instanceof \Illuminate\Support\Collection;
```
ghabriel25, line 137: You could simplify it with
```php
$uploads = $this->getPropertyValue($name);

$isArrayable = $uploads instance of \Illuminate\Contracts\Support\Arrayable;

$uploads = $isArrayable ? $uploads->toArray() : $uploads;

// original code

app('livewire')->updateProperty($this, $name, $isArrayable ? collect($filtered) : $filtered);

hamdyelbatal122: Thanks for the review @ghabriel25!

Converting to a plain array works for reading, but when calling `app('livewire')->updateProperty(...)` at the end, setting a strictly typed property (e.g. `public Collection $photos;`) to a plain array causes PHP to throw a `TypeError`.

That's why we preserve the `Collection` instance when updating the property.

ghabriel25: Apologize for incomplete snippet (updated). What I mean was as a flag so later can be used to check before setting the property back.

hamdyelbatal122: Thanks for clarifying @ghabriel25 

I've updated the implementation to extract `$items` upfront and flatten the conditional checks, while using `$isCollection` to safely preserve the property's type when updating it back.

Appreciate the review

joshhanley:
@hamdyelbatal122 thanks for the PR! We recently created a `4.x` branch for ongoing Livewire v4 development. Because this PR was based on `main`, we’re going to close it. Could you please rebase or cherry-pick your changes onto `4.x` and open a new PR targeting `4.x`? Thanks!

hamdyelbatal122:
Re-created targeting `4.x` branch here: #10485

PR #10461, opened 2026-07-29
[4.x] Adds `#[Debounce]` attribute for property-level `wire:model` debounce timing

ghabriel25:
# Summary

Adds a `#[Debounce]` PHP attribute so a property can declare its network debounce duration without requiring `.debounce.Xms` on every `wire:model.live` binding in Blade.

### Before

```blade
<input wire:model.live.debounce.250ms="search">
```

### After

```php
#[Debounce(250)]
public string $search = '';
```
```blade
<input wire:model.live="search">
```

Omitting the duration uses Livewire’s existing default of `150ms`:

```php
#[Debounce]
public string $search = '';
```

# Behaviour

| Blade | Attribute on property | Network timing |
| - | - | - |
| `wire:model.live="search"` | `#[Debounce]` | 150ms (default) |
| `wire:model.live="search"` | none | 150ms (default) |
| `wire:model.live="search"` | `#[Debounce(250)]` | 250ms (from attribute) |
| `wire:model.live.debounce.500ms="search"` | `#[Debounce(250)]` | 500ms (Blade modifier wins) |
| `wire:model="search"` | `#[Debounce(250)]` | still deferred — attribute does not force `.live` |

The attribute only supplies the debounce duration. It does not change whether a network request is sent. Users still need `.live` (or another network trigger) when they want live updates.
When both the attribute and a Blade `.debounce.Xms` modifier are present, the Blade modifier takes precedence so a template can override the property default without removing the attribute.

# What's Changed
- Feature class: `src/Features/SupportDebounce/BaseDebounce.php`
- Public alias: `src/Attributes/Debounce.php`
- Modify debounce duration if provided: `js/directives/wire-model.js`
- Inscribe debounce effects if provided when navigating: `js/component.js`

### DIFF
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -66,7 +66,7 @@ directive('model', ({ el, directive, component, cleanup }) => {
     let networkOnChange = networkModifiers.includes('change') || networkModifiers.includes('lazy')
     let networkOnEnter = networkModifiers.includes('enter')
     let hasNetworkTriggers = networkOnBlur || networkOnChange || networkOnEnter
-    let isDebounced = networkModifiers.includes('debounce')
+    let isDebounced = networkModifiers.includes('debounce') || componentHasDebounceEffects(component, expression)
     let isThrottled = networkModifiers.includes('throttle')
 
     // Trigger a network request
@@ -86,7 +86,11 @@ directive('model', ({ el, directive, component, cleanup }) => {
 
     // Apply debounce/throttle from network modifiers
     if ((shouldSendNetwork && ! hasNetworkTriggers && isRealtimeInput(el) && ! isDebounced && ! isThrottled) || isDebounced) {
-        debouncedUpdate = debounce(debouncedUpdate, parseModifierDuration(networkModifiers, 'debounce') || 150)
+        let debounceDuration = parseModifierDuration(networkModifiers, 'debounce') 
+            ?? componentEffectsDuration(component, expression, 'debounce')
+            ?? 150
+
+        debouncedUpdate = debounce(debouncedUpdate, debounceDuration)
     }
 
     if (isThrottled) {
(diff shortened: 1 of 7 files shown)


js/directives/wire-model.js
```diff
@@ -108,7 +108,9 @@ directive('model', ({ el, directive, component, cleanup }) => {
 
     // Apply debounce/throttle from network modifiers
     if ((shouldSendNetwork && ! hasNetworkTriggers && isRealtimeInput(el)) || isDebounced) {
-        debouncedUpdate = debounce(debouncedUpdate, parseModifierDuration(networkModifiers, 'debounce') || 150)
+        let debounceDuration = parseModifierDuration(networkModifiers, 'debounce')
```
hamdyelbatal122, line 89: Since `0` is falsy in JavaScript, using `||` causes `0 || 150` to evaluate to `150` instead of `0` when using `#[Debounce(0)]` or `.debounce.0ms`.

We can fix this by using the nullish coalescing operator (`??`) instead of `||`:

```js
let debounceDuration = parseModifierDuration(networkModifiers, debounce) 
    ?? componentEffectsDuration(component, expression, debounce)

debouncedUpdate = debounce(debouncedUpdate, debounceDuration ?? 150)
```

hamdyelbatal122 (commented):
Overall this is a fantastic PR @ghabriel25! I left an inline code review comment below regarding a minor edge-case with 0ms duration.

js/directives/wire-model.js
ghabriel25: @hamdyelbatal122 Thanks for the review! I've submitted a fix https://github.com/livewire/livewire/pull/10463 instead resolving the issue here. I'll use that fix if it's merged, till then I think I'll leave it as it is since it's a bug from the start.

However, I still add the regression test for this bug.

ghabriel25:
Going to close this as its not a proper way to handle both `wire:model.live` and `wire:click` which has debounce mechanism

PR #10462, opened 2026-07-29
[4.x] Discard stale responses when a newer request is in flight for model.live

iBotPeaches:
## Problem

When requests go out for a live component they can resolve out of order creating a pretty broken experience. The root cause being the older (late) response is parsed and displayed. This can be shown pretty simply with influenced delay on a DB query and the video below.

https://github.com/user-attachments/assets/099dbe32-e6ee-4cea-af8e-500ba5d9de3c

(I added some console logs to prove with simple incremental values that they can be parsed out of order)

This has been reported many times throughout the years: #8341, #2480 and #9030.

## Solution

You'd think a solution would be to cancel outbound requests when a newer one is made, but Livewire shares outbound calls with other components. You cannot safely abort requests due to this "batch" method. Instead, we decide to mark each message with a monotonic value that is incremented at payload build time assigned to each component (and scope) in a WeakMap.

When response comes back, any message who version is behind the component's current version is dropped. The promises are resolved as empty, but nothing is applied snapshot/effect wise.

https://github.com/user-attachments/assets/0b795557-1fe2-44ee-9635-9a2083d859d5

With more console logs. You can see the late response is ignored and thus the intended (current) version is the only one applied.

In testing I learned you can't apply this to every request, so we harden it down to only `model.live` types as those are exactly where the `.debounce` flaw occurs.

I did not understand why `requests.forEach` was done twice, the 2nd being to apply the `message.scope`, so I inlined that so I could apply my `.liveVersion` at same time.

## Testing
I cannot figure out a way to test this, because it seems the Dusk test server is a single process. So it doesn't matter the order of requests I send (even if I control the timing) because they run in order. So the entire parallel request aspect of the test I cannot test.

### DIFF
--- a/js/request/index.js
+++ b/js/request/index.js
@@ -4,7 +4,7 @@ import { MessageRequest, PageRequest } from './request.js'
 import { InterceptorRegistry } from './interceptor.js'
 import { trigger, triggerAsync } from '@/hooks.js'
 import { showHtmlModal } from '@/utils/modal.js'
-import { MessageBus, scopeSymbolFromMessage } from './messageBus.js'
+import { getLiveVersionFor, incrementLiveVersionFor, MessageBus, scopeSymbolFromMessage } from './messageBus.js'
 import Message from './message.js'
 import Action from './action.js'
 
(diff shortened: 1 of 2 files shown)


joshhanley:
@iBotPeaches thanks for the PR! We recently created a `4.x` branch for ongoing Livewire v4 development. Because this PR was based on `main`, we’re going to close it. Could you please rebase or cherry-pick your changes onto `4.x` and open a new PR targeting `4.x`? Thanks!

iBotPeaches:
Done - v4.x - https://github.com/livewire/livewire/pull/10488

PR #10463, opened 2026-07-30
[4.x] Use null coalescing instead logical `OR` in `wire-model.js` debounce and throttle duration

ghabriel25:
## Summary
Improve correctness and precision of debounce/throttle duration parsing in `wire-model.js` align with how alpine parse the duration modifer

## Problem
`parseModifierDuration()` currently returns a string (e.g. `"150"`) or `undefined`.
That value is then used like this:

```js
debounce(..., parseModifierDuration(...) || 150)
throttle(..., parseModifierDuration(...) || 150)
```

Two issues arise:

1. **String** → **number** coercion – `setTimeout` receives a string and must coerce it. While browsers handle this, it is unnecessary and slightly imprecise.
2. **Zero is treated as falsy** – A legitimate duration that forces to integer `0` is currently ignored and falls back to the default `150`. Zero is a valid and useful value (“no artificial delay”).

## Solution
- Parse the duration to an integer with `Number(duration)` so the function returns a proper number (or `undefined`).
- Replace the logical OR (`||`) with nullish coalescing (`??`) so only `null`/`undefined` trigger the 150 ms default.
Integer `0` is now correctly honored.

```js
// before
parseModifierDuration(...) || 150

// after
parseModifierDuration(...) ?? 150
```

### DIFF
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -242,7 +242,7 @@ function parseModifierDuration(modifiers, key) {
     if (index === -1) return undefined
 
     let nextModifier = modifiers[modifiers.indexOf(key)+1] || 'invalid-wait'
-    let duration = nextModifier.split('ms')[0]
+    let duration = parseInt(nextModifier.split('ms')[0], 10)
 
     return ! isNaN(duration) ? duration : undefined
 }
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -90,7 +90,7 @@ directive('model', ({ el, directive, component, cleanup }) => {
     }
 
     if (isThrottled) {
-        debouncedUpdate = throttle(debouncedUpdate, parseModifierDuration(networkModifiers, 'throttle') || 150)
+        debouncedUpdate = throttle(debouncedUpdate, parseModifierDuration(networkModifiers, 'throttle') ?? 150)
     }
 
     // Build the bindings object
@@ -244,5 +244,5 @@ function parseModifierDuration(modifiers, key) {
     let nextModifier = modifiers[modifiers.indexOf(key)+1] || 'invalid-wait'
     let duration = nextModifier.split('ms')[0]
 
-    return ! isNaN(duration) ? duration : undefined
+    return ! isNaN(duration) ? parseInt(duration) : undefined
 }
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -242,7 +242,7 @@ function parseModifierDuration(modifiers, key) {
     if (index === -1) return undefined
 
     let nextModifier = modifiers[modifiers.indexOf(key)+1] || 'invalid-wait'
-    let duration = parseInt(nextModifier.split('ms')[0], 10)
+    let duration = nextModifier.split('ms')[0]
 
-    return ! isNaN(duration) ? duration : undefined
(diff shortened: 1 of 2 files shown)


joshhanley:
@ghabriel25 thanks for the PR! Can you please split the fixes into two PRs and add tests for them. Thanks!

ghabriel25:
@joshhanley I'll do it right away, thanks!

ghabriel25:
Hi @joshhanley, do you think the PR should be split into two different PR like this? the issue was different but they are related each other. I think it should be as one fix so it can be tested with the regression test.

I feel like doing something wrong here.

joshhanley:
@ghabriel25 nah all good, I miss-read "This PR address 2 issue found in wire-model.js". If you think it should all be in the one PR, then that's fine! But yeah a test or two covering them would be good.

ghabriel25:
@joshhanley Done. The tests should be sufficient enough for this update. It's really tricky to get the right timing, because it passes on local but not on github action

ghabriel25:
Hi @calebporzio @joshhanley , just want to make sure if this is intended behavior that `.throttle` modifier will always debounced as long as it has `.live` modifier?

```blade
<input wire:model.live.throttle.50ms="search" />
```
```js
// Apply debounce/throttle from network modifiers
}

if (isThrottled) {
}
```
will end up
```js
throttle(debounce(update, 150), 50)
```

This is current behavior
| Binding | What actually runs |
| - | - |
| `wire:model.live` | `debounce(update, 150)` |
| `wire:model.live.debounce.300ms` | `debounce(update, 300)` |
| `wire:model.live.throttle` | `throttle(debounce(update, 150), 150)` |
| `wire:model.live.throttle.500ms` | `throttle(debounce(update, 150), 500)` |

joshhanley:
@ghabriel25 hmm seems strange to me that throttle is also debounced 🤔 but it's been a while since I looked at that code, so not sure exactly. Best thing to do would be to look back through older commits to see what it did previously (incase it's a recent unintentional change). 

@ghabriel25 also we've recently created a 4.x branch. I've updated the PR bases to be 4.x on your PRs. Are you able to rebase your changes on the 4.x branch for each of them? Thanks

ghabriel25:
@joshhanley I'll check the histroy because it doesnt make sense pure throttle requests get auto debounced just for having `.live` modifier

Also, I'll rebase all my PR into new branch, thanks for the information

joshhanley:
@ghabriel25 thanks! 😁

joshhanley:
@ghabriel25 I've marked your PRs as draft for now. Let me know if you want a hand rebasing any of them. Feel free to mark them as ready again once they are rebased.

ghabriel25:
@joshhanley Thanks! honestly I need a help for rebasing all my PR but its fine I guess. I'll learn the hard way sooner or later

joshhanley:
@ghabriel25 I'll have a quick look at them. Don't push any changes just yet. I'll let you know which I will need you to have a look at 🙂

ghabriel25:
@joshhanley Thanks a lot. I wont push anything atm.

joshhanley:
@ghabriel25 done! I've fixed up most of them. Just 3 big ones that were not possible to rebase. I've closed them for now.  Once you have reworked them, please open new PRs for them against `4.x`. Thanks!

PR #10464, opened 2026-07-30
[4.x] Fix `parseModifierDuration` in `wire-model.js` returns string number insead integer

ghabriel25:
## Summary
In `wire-model.js`, function `parseModifierDuration` returns string number or `undefined` eventhough `setTimeout` expect non-negative integer as delay.

From [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout)
> If the value isn't a number, implicit [type coercion](https://developer.mozilla.org/en-US/docs/Glossary/Type_coercion) is silently done on the value to convert it to a number. This can lead to unexpected and surprising results — see [Non-number delay values are silently coerced into numbers](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#non-number_delay_values_are_silently_coerced_into_numbers) for an example.

## Solution
Since `setTimeout` expect non-negative integer as argument, it would make more sense to provide an integer instead string number. In `parseModifierDuration`, parse the duration before checking

```js
function parseModifierDuration(modifiers, key) {
    let index = modifiers.indexOf(key)
    if (index === -1) return undefined

    let nextModifier = modifiers[index + 1] || 'invalid-wait'
    let duration = parseInt(nextModifier.split('ms')[0], 10)

    return ! isNaN(duration) ? duration : undefined
}
```

In addition use existing `index` instead repeating `modifiers.indexOf(key)`

### DIFF
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -263,8 +263,8 @@ function parseModifierDuration(modifiers, key) {
     let index = modifiers.indexOf(key)
     if (index === -1) return undefined
 
-    let nextModifier = modifiers[modifiers.indexOf(key)+1] || 'invalid-wait'
-    let duration = nextModifier.split('ms')[0]
+    let nextModifier = modifiers[index + 1] || 'invalid-wait'
+    let duration = parseInt(nextModifier.split('ms')[0], 10)
 
     return ! isNaN(duration) ? duration : undefined
 }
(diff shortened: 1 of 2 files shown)


ghabriel25:
Closing in favor https://github.com/livewire/livewire/pull/10463 but the test will be moved there

PR #10465, opened 2026-07-30
Improve query string parsing and effect handling

joshhanley:
WIP


joshhanley:
Closing to target v3 instead.

PR #10471, opened 2026-07-31
[4.x] Recursively hydrate nested synthesizers during property updates

lam0819:

No discussion yet — this is a bug fix that comes with a failing test, so it seemed more useful to bring the reproduction. Happy to open a discussion if you'd prefer to talk it through first.

Yes — `fix-nested-synth-hydration-on-property-updates`.

No. One change in `HandleSynths`, plus tests.

Yes — 7 tests, all of which fail on `main` without the change:

- `src/Mechanisms/HandleSynths/UnitTest.php` — the mechanism itself (recursion, missing meta, `__rm__`, already-hydrated values, security policy over nested meta)
- `src/Features/SupportFileUploads/UnitTest.php` — the real-world symptom (nested uploads surviving an ancestor update)

**5️⃣ Please include a thorough description of the improvement and reasons why it's useful.**

### The bug

Non-primitive values ride the wire as a `[value, meta]` tuple, where the meta says which synthesizer rebuilds the PHP value. Full hydration walks that tree recursively:

```php
// HandleSynths::hydrate()
return $synth->hydrate($value, $meta, function ($name, $child) use ($context, $path) {
    return $this->hydrate($child, $context, "{$path}.{$name}");
});
```

Property updates did not:

```php
// HandleSynths::hydratePropertyUpdate()
return $synth->hydrate($value, $meta, function ($name, $child) {
    return $child;
});
```

So when an update targets a path *above* a synthesized value, the synth at the updated path is applied but everything nested under it comes back as raw wire data.

### How you hit it

Nested repeaters. With state shaped like `data.sections.{section}.rows.{row}.image`, editing any field in the section sends an update for the ancestor path (`data.sections.first`), not for the image leaf. The payload contains the already-uploaded file as its serialized string, and — because the browser strips meta out of update payloads — there is no meta on it. The property is then repopulated with:

```php
'image' => 'livewire-file:BToZJYNz....jpg'   // a string, not a TemporaryUploadedFile
```

The upload is still on disk, but nothing downstream recognises it any more, so the file "disappears" from the form. In Filament this shows up as a nested repeater losing images from rows the user never touched. Related report: https://github.com/livewire/livewire/discussions/6203

It isn't specific to file uploads — any nested synthesized value (collections, enums, Carbon instances, custom synths) is affected; uploads just make the loss visible.

### The fix

Hydrate nested children during an update too, using the meta from the **previous snapshot** rather than the update payload:

```php
$childPath = "{$path}.{$name}";

if (! $childMeta = $this->getMetaForPath($raw, $childPath)) return $child;

return $this->hydratePropertyUpdate([$child, $childMeta], $context, $childPath, $raw);
```

`$raw` is the snapshot data that `update()` already verified with `Checksum::verify()`, and it is the same source the top-level `hydratePropertyUpdate()` call already trusts for meta — the update payload never gets a say in which class is instantiated. Because the recursion goes back through `hydratePropertyUpdate()`, every nested tuple still passes `SecurityPolicy::validateClass()` and still honours the `__rm__` removal sentinel.

Two deliberate bail-outs:

- **No meta in the snapshot → the value is left alone.** New children (a repeater row added this request) are plain data, exactly as before.
- **A child that is already a PHP object is left alone.** Real update payloads are JSON, so an object can only have been set server-side — `Livewire::test()->set('list', [SomeEnum::TEST])`, for instance. Running those back through a synth would be wrong (and would break `EnumUnitTest::test_an_array_of_enums`).

`hydratePropertyUpdate()`'s new `$raw` argument is optional and defaults to `null` (old, non-recursive behaviour), so external callers of the method are unaffected.

### Notes

- `vendor/bin/phpunit --testsuite Unit`: 1421 tests, only the 5 `FrontendAssets` failures that already fail on a clean `main` in my environment (Laravel 13-dev / PHP 8.5).
- The same bug exists on `3.x`, where these methods live in `HandleComponents` instead — happy to send a backport PR if you want it there.

### DIFF
--- a/src/Features/SupportFileUploads/UnitTest.php
+++ b/src/Features/SupportFileUploads/UnitTest.php
@@ -1216,6 +1216,79 @@ class UnitTest extends \Tests\TestCase
 
         $this->assertInstanceOf(TemporaryUploadedFile::class, $test->viewData('photo'));
     }
+
+    public function test_nested_uploads_survive_an_update_to_one_of_their_ancestor_paths()
+    {
+        // Nested repeaters (a Filament repeater inside a repeater, for
+        // example) update an ancestor path rather than the file's own leaf
+        // path. The browser strips synthesizer meta out of update payloads,
+        // so already-uploaded files arrive as their raw serialized strings
+        // and have to be re-hydrated from the previous snapshot's meta...
+        $component = Livewire::test(NestedFileUploadComponent::class)
+            ->set('data.sections.first.rows.one.image', UploadedFile::fake()->image('one.jpg'))
+            ->set('data.sections.first.rows.two.image', UploadedFile::fake()->image('two.jpg'));
+
+        $one = $component->viewData('data')['sections']['first']['rows']['one']['image'];
+        $two = $component->viewData('data')['sections']['first']['rows']['two']['image'];
+
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $one);
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $two);
+
+        // Typing into a caption of one row sends the whole section back...
+        $component->set('data.sections.first', [
+            'rows' => [
+                'one' => ['image' => $one->serializeForLivewireResponse(), 'caption' => 'Updated'],
+                'two' => ['image' => $two->serializeForLivewireResponse(), 'caption' => ''],
+            ],
+        ]);
+
+        $rows = $component->viewData('data')['sections']['first']['rows'];
+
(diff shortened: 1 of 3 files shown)


joshhanley:
@lam0819 thanks for the PR! We recently created a `4.x` branch for ongoing Livewire v4 development. Because this PR was based on `main`, we’re going to close it. Could you please rebase or cherry-pick your changes onto `4.x` and open a new PR targeting `4.x`? Thanks!

PR #10474, opened 2026-07-31
[4.x] Fix single-file component detection when a PHP attribute has an array argument

### LINKED ISSUE #10472: [4.3.4] SFC no longer detected when a PHP attribute has an array argument (regression from #10367)
### Livewire version

v4.3.4

### Laravel version

v13.23.0

### PHP version

8.4.23

N/A — this fails during server-side component resolution, before anything reaches the browser (macOS, Darwin 25.5.0)

After upgrading v4.3.3 → v4.3.4, single-file components whose PHP attribute receives an **array argument** are no longer detected as SFCs. Resolving them throws:

```
Livewire\Exceptions\ComponentNotFoundException: Unable to find component: [pages::bug-repro]
```

The cause is the narrowed SFC detection regex from #10367 in `src/Finder/Finder.php`:

```diff
- return preg_match('/\<\?php.*new\s+.*class/s', $contents) === 1;
+ return preg_match('/\<\?php.*\bnew\s+(?:#\[[^\]]*\]\s*)*class\b/s', $contents) === 1;
```

The attribute branch `#\[[^\]]*\]` stops at the **first** `]`. An attribute like `#[Layout('layouts.app', ['title' => 'Dashboard'])]` contains a `]` from its own array argument, so the match fails and the file stops being treated as an SFC.

Layout data as a second argument is part of the attribute's public API — `BaseLayout::__construct(public $name, public $params = [])` — and #10367 states the change "keeps valid SFC forms working, including inline and multiline attributes". The fixtures added there (`single-file-component-with-inline-php-attribute.blade.php`, `single-file-component-with-multiline-php-attribute.blade.php`) only use attributes without array arguments, so this form was not covered.

In our application this silently broke 11 of 139 SFCs — every page component that passes a title to its layout.

1. Create `resources/views/pages/bug-repro.blade.php` (with `pages` mapped to `resources/views/pages` in `component_namespaces`). The layout name is irrelevant — what matters is the array argument:

```blade
<?php

use Livewire\Attributes\Layout;
use Livewire\Component;

new
#[Layout('layouts.app', ['title' => 'Dashboard'])]
class extends Component {
    public string $message = 'Hello';
};

?>

<div>
    {{ $message }}
</div>
```

2. Resolve it — via a route, `<livewire:pages::bug-repro />`, or tinker:

```php
Livewire::mount('pages::bug-repro');
```

Result on v4.3.4:

```
Livewire\Exceptions\ComponentNotFoundException: Unable to find component: [pages::bug-repro]
```

3. Control case — change **only** the attribute in that same file and it resolves again:

```blade
#[Layout('layouts.app')]
```

| Livewire | attribute | result |
| --- | --- | --- |
| v4.3.3 | `#[Layout('layouts.app', ['title' => 'Dashboard'])]` | renders |
| v4.3.4 | `#[Layout('layouts.app', ['title' => 'Dashboard'])]` | `ComponentNotFoundException` |
| v4.3.4 | `#[Layout('layouts.app')]` | renders |

_No response_

An SFC with an attribute that takes an array argument should keep resolving, exactly as it did in v4.3.3 and as #10367 intended.

Allowing one level of bracket nesting in the attribute branch is enough:

```diff
- '/\<\?php.*\bnew\s+(?:#\[[^\]]*\]\s*)*class\b/s'
+ '/\<\?php.*\bnew\s+(?:#\[(?:[^\[\]]|\[[^\[\]]*\])*\]\s*)*class\b/s'
```

Verified against our codebase (139 SFCs): the current regex matches 128, the patched one matches 139, and the Volt fixture from #10367 (`volt-functional-component.blade.php`) still correctly resolves to `null`, so the original fix does not regress.

A `token_get_all()`-based check would of course be more robust than a regex, but the above is the minimal change.

lazerg:
`#10367` narrowed single-file component detection to `new` followed by optional `#[...]` attributes and then `class`. The attribute part, `#\[[^\]]*\]`, stops at the first `]`, so an attribute whose argument contains an array closes early. The file is no longer seen as an SFC, and resolving it throws `ComponentNotFoundException`.

This stopped resolving after 4.3.4:

```blade
new
#[Layout('layouts.app', ['title' => 'Dashboard'])]
class extends Component { /* ... */ };
```

Allowing one level of bracket nesting in the attribute keeps array arguments working, and Volt functional components are still skipped as `#10367` intended.

Fixes #10472

### DIFF
--- a/src/Finder/Finder.php
+++ b/src/Finder/Finder.php
@@ -395,7 +395,7 @@ class Finder
         // Light touch check: Look for the pattern that indicates an SFC
         // Pattern: <?php followed by 'new class' (with potential attributes/newlines between)
         // This distinguishes SFCs from regular Blade views
-        return preg_match('/\<\?php.*\bnew\s+(?:#\[[^\]]*\]\s*)*class\b/s', $contents) === 1;
+        return preg_match('/\<\?php.*\bnew\s+(?:#\[(?:[^\[\]]|\[[^\[\]]*\])*\]\s*)*class\b/s', $contents) === 1;
     }
 
     protected function hasValidMultiFileComponentSource(string $dir, string $fileBaseName): bool
--- a/src/Finder/Fixtures/single-file-component-with-array-argument-php-attribute.blade.php
+++ b/src/Finder/Fixtures/single-file-component-with-array-argument-php-attribute.blade.php
@@ -1,0 +1,14 @@
+<?php
+
+use Livewire\Attributes\Layout;
+use Livewire\Component;
+
+new
+#[Layout('layouts.app', ['title' => 'Dashboard'])]
+class extends Component
+{
+    //
+};
+?>
+
+<div>Finder Test Single File Component With Array Argument Attribute</div>
--- a/src/Finder/UnitTest.php
+++ b/src/Finder/UnitTest.php
@@ -239,6 +239,17 @@ class UnitTest extends \Tests\TestCase
         $this->assertEquals(__DIR__ . '/Fixtures/single-file-component-with-multiline-php-attribute.blade.php', $path);
     }
 
+    public function test_can_resolve_single_file_component_with_array_argument_php_attribute()
+    {
+        $finder = new Finder();
+
+        $finder->addLocation(viewPath: __DIR__ . '/Fixtures');
+
+        $path = $finder->resolveSingleFileComponentPath('single-file-component-with-array-argument-php-attribute');
+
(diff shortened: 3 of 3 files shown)


joshhanley:
@lazerg thanks for the PR! You're PR targets the main branch instead of the 4.x branch and there are a few extra scenarios it needed to cover. So I've fixed this in PR #10480 which targets the 4.x and tagged a new release v4.3.5. Because of that I'm going to close this. Thanks for being willing to contribute!

lazerg:
No worries, thanks for fixing it up and getting it released.

PR #10476, opened 2026-08-01
[4.x] Only match a component's class namespace on whole segments

calebporzio:
## Problem

`Finder::generateNameFromClass()` strips a configured class namespace using a plain prefix match. That treats a sibling namespace that merely begins with the same characters as though it were inside the configured namespace.

With the default `App\Livewire` class namespace:

```text
App\Livewire\Forms\ContactForm  → forms.contact-form
App\LivewireForms\ContactForm   → forms.contact-form
```

The second class should normalize to `app.livewire-forms.contact-form`. Instead, both classes collapse to the same component name. Rendering or routing to `App\LivewireForms\ContactForm` can therefore mount `App\Livewire\Forms\ContactForm` without an exception. If no colliding component exists, resolution fails with a misleading `ComponentNotFoundException`.

## Fix

Require the normalized namespace prefix to end at a segment boundary:

```php
$fullName->startsWith($namespace . '.')
```

Classes genuinely inside the configured namespace still have that namespace stripped. Prefix-sharing sibling namespaces keep their complete normalized name and resolve back to their original class.

## Tests

Added Finder-level coverage for both registration paths:

- class locations registered with `addLocation()`
- named class namespaces registered with `addNamespace()`

Each test normalizes a class from a prefix-sharing sibling namespace and verifies that the resulting name resolves back to the same class.

Validation:

- `vendor/bin/phpunit src/Finder/UnitTest.php`
- 63 tests, 100 assertions
- Manual full-page route verification confirms the base branch mounts the wrong component for a prefix-sharing class, while this branch routes both classes correctly


joshhanley:
I've changed the base for this PR to `4.x` and rebased it on `4.x`.

PR #10479, opened 2026-08-02
Prevent infinite request loops when referencing component methods in reactive directives

### LINKED ISSUE #10477: Component method is called repeatedly when using `wire:text`
### Livewire version

v4.3

### Laravel version

v13.0

### PHP version

8.4

Chrome on Windows

Using `wire:text` to a component method causing the method get called repeatedly. While `wire:text` used for component property, it doesnt do any guards against calling component method.

```php
<?php

use Livewire\Component;

new class extends Component
{
    public $bar = 'bar';

    public function foo()
    {
        return $this->bar;
    }
}
```
```blade
<div>
    <span wire:text="foo"></span>
</div>
```

https://github.com/user-attachments/assets/fcb320a3-7338-4f04-b265-d4e26b42c36d

It should prevent to calling the method, or at least retrieve the method once but still checking the return type can be shown as a text or not.

hamdyelbatal122:
Fixes #10477

### Problem
When client-side reactive binding directives (`wire:text`, `wire:show`, `wire:bind`, `wire:if`) evaluate an expression that points to a PHP component method instead of a public property (e.g., `<span wire:text="foo"></span>`), accessing `$wire.foo` returns a fallback action wrapper function.

Because Alpine's expression evaluator automatically executes functions returned inside reactive getters (`x-text`, `x-show`, etc.), Alpine invokes the action function on every render cycle. This fires `fireAction()`, sending a network request to the server. When the server response updates component state, Alpine's reactive effect re-evaluates the directive, triggering another request in an infinite HTTP request loop.

### Solution
1. Added `evaluateReactiveExpression` to mark execution context while evaluating reactive DOM binding directives.
2. Combined `@vue/reactivity` active watcher detection (`getCurrentWatcher()`) to catch both `wire:*` directives and direct Alpine `x-text="$wire.foo()"` usage.
3. Added a guard in `fireAction()` that prevents firing server network requests during reactive evaluations, logging a clear console warning instead (`Livewire: Cannot call server method "foo" inside reactive bindings...`).
4. Added Dusk browser tests in `SupportWireText/BrowserTest.php` to prevent regressions.

### DIFF
--- a/js/directives/wire-bind.js
+++ b/js/directives/wire-bind.js
@@ -1,5 +1,5 @@
 import Alpine from 'alpinejs'
-import { evaluateActionExpression } from '../evaluator'
+import { evaluateReactiveExpression } from '../evaluator'
 
 Alpine.interceptInit(el => {
     for (let i = 0; i < el.attributes.length; i++) {
@@ -12,7 +12,7 @@ Alpine.interceptInit(el => {
 
             Alpine.bind(el, {
                 ['x-bind' + remainder]() {
-                    return evaluateActionExpression(el, expression)
+                    return evaluateReactiveExpression(el, expression)
                 }
             })
         }
--- a/js/directives/wire-if.js
+++ b/js/directives/wire-if.js
@@ -1,4 +1,4 @@
-import { evaluateActionExpression } from '../evaluator'
+import { evaluateReactiveExpression } from '../evaluator'
 import Alpine from 'alpinejs'
 
 // Skipped during clone-mode init (morphing seeds incoming trees that way) —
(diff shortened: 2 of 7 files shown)


joshhanley:
@hamdyelbatal122 thanks for the PR! We recently created a `4.x` branch for ongoing Livewire v4 development. Because this PR was based on `main`, we’re going to close it. Could you please rebase or cherry-pick your changes onto `4.x` and open a new PR targeting `4.x`? Thanks!

hamdyelbatal122:
Re-created targeting `4.x` branch here: #10484

PR #10486, opened 2026-08-03

ghabriel25:
> [!NOTE]
> This PR includes the changes from PR https://github.com/livewire/livewire/pull/10398 so that tests pass locally.
> Once that PR is merged,  I will rebase/clean this PR and mark as ready

## Summary

This PR adds a new `#[Middleware]` attribute that allows developers to apply Laravel middleware (class names or aliases, with optional parameters) directly to individual Livewire component actions.

Livewire actions are public methods that can be invoked from the frontend via `wire:click`, forms, or events. While route-level middleware (via `PersistentMiddleware`) and `#[Authorize]` already exist, there has been no first-class, declarative way to protect a specific action with arbitrary middleware the same way Laravel controllers support middleware attributes.

This feature closes that gap by enabling early, method-level protection so that unauthorized or restricted calls never reach business logic.

This PR resolve the concern from this comment https://github.com/livewire/livewire/pull/10374#pullrequestreview-4629568219 and some regression test added to ensure that it works as expected. 

Need to mention that this attribute does not behave like `#[Authorize]` which can be called in collection loop inside `SupportEvents` or regular attribute call hook. The reason is this attribute should gather all middleware and apply it all at once (not one by one) just like `PersistentMiddleware` therefore it takes precedence over event and attribute `call()` hook.

## How it works

1. **Collection (boot time)**  
   The `#[Middleware]` attribute stores middleware definitions (class or alias) in the component’s internal store, keyed by method name.

2. **Authorization conversion (boot time)**  
   Any middleware that resolves to `Illuminate\Auth\Middleware\Authorize` (including the `can:` alias) is automatically converted into a real `#[Authorize]` attribute on the method. This keeps the two systems consistent and reuses the existing authorization infrastructure.

3. **Application (call time)**  
   On the `call` lifecycle event (only for requests hitting Livewire update endpoint):
   - The real method name is resolved (including event listeners via `__dispatch`).
   - Middleware already applied by `PersistentMiddleware` is filtered out to avoid unnecessary duplicates.
   - Remaining middleware is applied **all at once** through the existing `Utils::applyMiddleware()` pipeline using a fake request (the same technique used by `PersistentMiddleware`).

Because middleware is gathered and applied as a single batch on the `call` event, it runs **before** normal attribute `call()` hooks and event listeners. This gives middleware precedence over `#[Authorize]`.

## Usage examples

**Basic usage (class or alias)**

```php
#[Middleware('auth')]
#[Middleware(EnsureEmailIsVerified::class)]
public function showContent()
{
    //
}

#[Middleware('guest')]
{
    //
}
```

**With parameters**

```php
#[Middleware('auth')]
#[Middleware('throttle:10')]
public function likePost(Post $post) { ... }

#[Middleware(CustomRoleMiddleware::class . ':admin')]
public function adminOnly() { ... }
```

**With events**

```php
#[On('mark-as-read')]
#[Middleware('auth')]
public function markAsRead()
{
    //
}
```

**Precedence over authorization**

```php
#[Middleware('auth')]
#[Authorize('update', 'post')] // only runs if auth middleware passes
public function save(Post $post) { ... }
```

**Automatic conversion of authorize middleware**

```php
// is converted at boot time into:
// #[Authorize('update', 'post')]
```

## Important behavioral notes

- The attribute only runs on subsequent Livewire update requests (the same constraint as PersistentMiddleware). It does not run on the initial page load.
- Middleware already registered as persistent is automatically excluded to avoid double execution.
- Middleware that is not registered as persistent but is applied both at the route level and via the attribute will run twice. This is intentional.

## What's changed

- New feature: `SupportActionMiddleware`
- New attribute: `Livewire\Attributes\Middleware` (extends internal `BaseMiddleware`)
- Small extension to `HandlesAttributes` (`setMethodAttribute`) so authorize conversion can inject attributes cleanly
- `PersistentMiddleware` now exposes the list of middleware it actually applied so the new feature can filter against it without re-resolving
- Registered in the service provider alongside other features

The design deliberately reuses existing infrastructure (Utils::applyMiddleware, the fake-request pattern, the attribute store, and the event listener resolution logic) rather than introducing a parallel system.

### DIFF
--- a/src/Attributes/Middleware.php
+++ b/src/Attributes/Middleware.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Livewire\Attributes;
+
+use Livewire\Features\SupportActionMiddleware\BaseMiddleware;
+
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_METHOD)]
+class Middleware extends BaseMiddleware
+{
+    //
+}
--- a/src/Features/SupportActionMiddleware/BaseMiddleware.php
+++ b/src/Features/SupportActionMiddleware/BaseMiddleware.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Livewire\Features\SupportActionMiddleware;
+
+use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
+
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_METHOD)]
+class BaseMiddleware extends LivewireAttribute
+{
+    public function __construct(public string $middleware)
+    {
+        //
+    }
+
+    function boot()
+    {
+        $middleware = $this->storeFind('middlewareAttributes', $this->getName(), []);
+
+        $middleware[] = $this->middleware;
+
+        $this->storePush('middlewareAttributes', $middleware, $this->getName());
+    }
+}
--- a/src/Features/SupportActionMiddleware/BrowserTest.php
+++ b/src/Features/SupportActionMiddleware/BrowserTest.php
@@ -1,0 +1,359 @@
+<?php
+
+namespace Livewire\Features\SupportActionMiddleware;
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Blade;
+use Illuminate\Support\Facades\Route;
+use Livewire\Attributes\Middleware;
+use Livewire\Attributes\On;
+use Livewire\Component;
+use Livewire\Livewire;
+use Tests\BrowserTestCase;
+use Tests\TestComponent;
+
+class BrowserTest extends BrowserTestCase
+{
+    public static function tweakApplicationHook()
+    {
+        return function () {
+            Route::livewire('/somewhere', new class extends TestComponent {})
+                ->middleware('web')
+                ->name('somewhere');
+
+            Route::livewire('/somewhere-else', new class extends TestComponent {})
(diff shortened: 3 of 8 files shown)


ghabriel25:
Hi, @joshhanley. I think I need help for this. I dont know why it keeps rebasing main branch. I apologize for the inconvenience I made.

joshhanley:
@ghabriel25 no worries! I've rebased it on 4.x. Does it look correct to you?

ghabriel25:
@joshhanley Yup, this is how it should be! Thanks! 

Need to learn more about git 😅 

It seems some updates from main branch not included. Its `Attribute` refactor that uses `$this->storeFind()` helper which lead to failing tests

joshhanley:
@ghabriel25 no worries! Yeah I could work it out, but I just get Codex to fix it 😁

If there was stuff on main that was merged after Caleb started some of his v5 stuff, then yeah they won't be on the 4.x branch. Feel free to PR them to `4.x` and reference the original PR or commit on `main` and I will get them merged.

calebporzio:
Thanks for the substantial work here. I’m going to close this because action middleware is a large, security-sensitive API direction that we haven’t agreed to. As mentioned on the earlier iterations, please start with an issue or discussion covering the concrete use cases and proposed API before implementing it. For now, I’d prefer dedicated action-level attributes such as `#[Authorize]` over exposing the full middleware stack on component methods.

PR #10489, opened 2026-08-03
[4.x] Recursively hydrate nested synthesizers during property updates

lam0819:
Re-targeted from #10471 at @joshhanley's request — same change, branched off `4.x` instead of `main`.

No discussion — this is a bug fix that comes with a failing test, so it seemed more useful to bring the reproduction. Happy to open one if you'd prefer to talk it through.

Yes — `fix-nested-synth-hydration-on-property-updates-4x`, targeting `4.x`.

No. One change in `HandleSynths`, plus tests.

Yes — 9 tests. Six reproduce the bug and fail on `4.x` without the change, one guards the already-hydrated testing path, and two lock down the snapshot/update trust boundary:

- `src/Mechanisms/HandleSynths/UnitTest.php` — the mechanism itself (recursion, missing meta, `__rm__`, already-hydrated values, security policy over nested meta, forged update tuples, and checksum-protected nested meta)
- `src/Features/SupportFileUploads/UnitTest.php` — the real-world symptom (nested uploads surviving an ancestor update)

**5️⃣ Please include a thorough description of the improvement and reasons why it's useful.**

### The bug

Non-primitive values ride the wire as a `[value, meta]` tuple, where the meta says which synthesizer rebuilds the PHP value. Full hydration walks that tree recursively:

```php
// HandleSynths::hydrate()
return $synth->hydrate($value, $meta, function ($name, $child) use ($context, $path) {
    return $this->hydrate($child, $context, "{$path}.{$name}");
});
```

Property updates did not:

```php
// HandleSynths::hydratePropertyUpdate()
return $synth->hydrate($value, $meta, function ($name, $child) {
    return $child;
});
```

So when an update targets a path *above* a synthesized value, the synth at the updated path is applied but everything nested under it comes back as raw wire data.

### How you hit it

Nested repeaters. With state shaped like `data.sections.{section}.rows.{row}.image`, editing any field in the section sends an update for the ancestor path (`data.sections.first`), not for the image leaf. The payload contains the already-uploaded file as its serialized string, and — because the browser strips meta out of update payloads — there is no meta on it. The property is then repopulated with:

```php
'image' => 'livewire-file:BToZJYNz....jpg'   // a string, not a TemporaryUploadedFile
```

The upload is still on disk, but nothing downstream recognises it any more, so the file "disappears" from the form. In Filament this shows up as a nested repeater losing images from rows the user never touched. Related report: https://github.com/livewire/livewire/discussions/6203

It isn't specific to file uploads — any nested synthesized value (collections, enums, Carbon instances, custom synths) is affected; uploads just make the loss visible.

### The fix

Hydrate nested children during an update too, using the meta from the **previous snapshot** rather than the update payload:

```php
$childPath = "{$path}.{$name}";

if (! $childMeta = $this->getMetaForPath($raw, $childPath)) return $child;

return $this->hydratePropertyUpdate([$child, $childMeta], $context, $childPath, $raw);
```

`$raw` is the snapshot data that `update()` already verified with `Checksum::verify()`, and it is the same source the top-level `hydratePropertyUpdate()` call already trusts for meta — the update payload never gets a say in which class is instantiated. Because the recursion goes back through `hydratePropertyUpdate()`, every nested tuple still passes `SecurityPolicy::validateClass()` and still honours the `__rm__` removal sentinel.

### Security contract

The update payload controls values, never type identity. Synthesizer keys, classes, and other hydration metadata are read only from the previous snapshot after `Checksum::verify()` authenticates it with the application key. A tuple-shaped update value is therefore ordinary input data; it cannot select a synthesizer or class.

The security tests exercise both sides of that boundary:

- An update value shaped exactly like a synth tuple and claiming a denylisted `Symfony\Component\Process\Process` class remains ordinary data inside the Collection selected by the authenticated snapshot meta.
- Replacing that nested Collection meta inside the previous snapshot is rejected with `CorruptComponentPayloadException` before update hydration runs.
- Authenticated nested classes still pass through `SecurityPolicy::validateClass()` and each synth's own class constraints during recursive hydration.

This grants no new write authority: a client that can submit a consolidated ancestor update could already submit the equivalent granular leaf update. The change makes both routes apply the same snapshot-derived hydration contract.

Two deliberate bail-outs:

- **No meta in the snapshot → the value is left alone.** New children (a repeater row added this request) are plain data, exactly as before.
- **A child that is already a PHP object is left alone.** Real update payloads are JSON, so an object can only have been set server-side — `Livewire::test()->set('list', [SomeEnum::TEST])`, for instance. Running those back through a synth would be wrong (and would break `EnumUnitTest::test_an_array_of_enums`).

`hydratePropertyUpdate()`'s new `$raw` argument is optional and defaults to `null` (old, non-recursive behaviour), so external callers of the method are unaffected.

### Relationship to #10431

#10431 ("Keep rich synth values out of consolidated diffs") attacks the same symptom from the client: it stops `diffAndConsolidate()` from replacing per-leaf diffs with a whole-parent update when the subtree holds rich synth values. That's a good mitigation, but it's on `main` only, and it's complementary rather than overlapping:

- It can only see values JS knows are rich — types registered via `Livewire.synth()`. Server-only synthesizers (`Collection`, `Carbon`, enums, `stdClass`, userland `Synth`/`Wireable`) have no JS counterpart, so they're still consolidated and still lose their type.
- On `4.x` uploads aren't a JS synth at all (that arrived with `js/features/supportFileUploads/synth.js` on `main`), so a backport of #10431 alone would not fix the upload case here.
- Ancestor updates also arrive by routes the differ doesn't control — `$wire.$set('data.sections', {...})`, a package writing parent state, or a hand-crafted request. The server should reconstruct correctly regardless of how the payload was produced.

### Notes

- `vendor/bin/phpunit --testsuite Unit` on this branch: 1344 tests, with exactly the same 5 pre-existing `FrontendAssets` failures that a clean `4.x` produces in my environment (Laravel 13-dev / PHP 8.5) — 1335 tests there, same 4 errors + 1 failure.
- The same bug exists on `3.x`, where these methods live in `HandleComponents` — happy to send that backport too if you want it.

### DIFF
--- a/src/Features/SupportFileUploads/UnitTest.php
+++ b/src/Features/SupportFileUploads/UnitTest.php
@@ -941,6 +941,79 @@ class UnitTest extends \Tests\TestCase
         Storage::disk('default-disk')->assertExists('images/avatar.jpg');
         Storage::disk('tmp-for-tests')->assertMissing('images/avatar.jpg');
     }
+
+    public function test_nested_uploads_survive_an_update_to_one_of_their_ancestor_paths()
+    {
+        // Nested repeaters (a Filament repeater inside a repeater, for
+        // example) update an ancestor path rather than the file's own leaf
+        // path. The browser strips synthesizer meta out of update payloads,
+        // so already-uploaded files arrive as their raw serialized strings
+        // and have to be re-hydrated from the previous snapshot's meta...
+        $component = Livewire::test(NestedFileUploadComponent::class)
+            ->set('data.sections.first.rows.one.image', UploadedFile::fake()->image('one.jpg'))
+            ->set('data.sections.first.rows.two.image', UploadedFile::fake()->image('two.jpg'));
+
+        $one = $component->viewData('data')['sections']['first']['rows']['one']['image'];
+        $two = $component->viewData('data')['sections']['first']['rows']['two']['image'];
+
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $one);
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $two);
+
+        // Typing into a caption of one row sends the whole section back...
+        $component->set('data.sections.first', [
+            'rows' => [
+                'one' => ['image' => $one->serializeForLivewireResponse(), 'caption' => 'Updated'],
+                'two' => ['image' => $two->serializeForLivewireResponse(), 'caption' => ''],
+            ],
+        ]);
+
+        $rows = $component->viewData('data')['sections']['first']['rows'];
+
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $rows['one']['image']);
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $rows['two']['image']);
+        $this->assertEquals('one.jpg', $rows['one']['image']->getClientOriginalName());
+        $this->assertEquals('two.jpg', $rows['two']['image']->getClientOriginalName());
+        $this->assertEquals('Updated', $rows['one']['caption']);
+    }
+
+    public function test_a_row_added_alongside_a_nested_upload_doesnt_disturb_it()
+    {
+        $component = Livewire::test(NestedFileUploadComponent::class)
+            ->set('data.sections.first.rows.one.image', UploadedFile::fake()->image('one.jpg'));
+
+        $one = $component->viewData('data')['sections']['first']['rows']['one']['image'];
+
+        // A brand new row has no counterpart in the previous snapshot, so
+        // there's no meta to hydrate it with — it stays plain data while its
+        // existing sibling is still reconstructed...
+        $component->set('data.sections.first', [
+            'rows' => [
+                'one' => ['image' => $one->serializeForLivewireResponse(), 'caption' => ''],
+                'three' => ['image' => null, 'caption' => 'New row'],
+            ],
+        ]);
+
+        $rows = $component->viewData('data')['sections']['first']['rows'];
+
+        $this->assertInstanceOf(TemporaryUploadedFile::class, $rows['one']['image']);
+        $this->assertEquals(['image' => null, 'caption' => 'New row'], $rows['three']);
+    }
+}
+
+class NestedFileUploadComponent extends TestComponent
+{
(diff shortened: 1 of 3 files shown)


lukas-staab:
Independent confirmation of this from production, with a userland synth rather than file uploads (but also reprocible with livewire supported Carbon as well, see below)

  Laravel finance app on Livewire v4.3.1. A public `array $posts` is an array of table rows with money values,
  each row carrying `Cknow\Money\Money` values hydrated by a small userland `Synth`. Editing a
  deferred-bound field (`wire:model`, no `.live`) in *every* row and then clicking save produced a non-synth error.

  The money leaves arrived as their raw display strings (`"1.234,50 €"`), exactly as described here, not as `1234.5` as they should have. 

  Two details that may be worth adding to the case:

  1. **The trigger wasn't a shape change.** No rows were added or removed — this is the
     `changedCount === totalChildren` branch of `diffRecursive` (`totalChildren > 1`, one changed
     field per row). So the "array grew or shrank" framing in #10431 undersells the reach: an
     array of N > 1 rich values consolidates as soon as the user touches something in each row.
     With a single row it never fires, which is why this sat unnoticed in a long-running app.

  2. **It isn't only a crash — it can corrupt data silently.** Our type happens to have a
     locale-dependent parser downstream. Had the string reached the model cast instead of a method
     call, `"1.234,50 €"` would have been read as `1.23` under a non-German locale and written to
     the database as a plausible wrong amount, with no error anywhere. A crash was the lucky outcome.

  I can confirm #10431 does not help here even in principle: Our `MoneySynth` is server-only with no JS
  counterpart (yet), so `containsSynthValue()` cannot see it. The recursive server-side hydration in this
  PR is what would fix it. Reproducing it needs nothing custom, though — a plain
  `public array $rows` holding `Carbon` instances degrades identically on 4.x: a `rows.0.when`
  update hydrates to `Carbon`, a consolidated `rows` update leaves a `string`.

calebporzio:
This looks good to me. Just need to run it by a few agents for security hardening, then I think we're good to go.

PR #10492, opened 2026-08-03
[4.x] Fix throttle requests get debounced by `.live` modifier

ghabriel25:
## Problem
When using pure `.throttle` modifier on input element
```blade
<input wire:model.live.throttle.30ms="search" />
```
It will get auto debounced by `150ms` from `.live` modifier eventhough it doesnt include `.debounce` modifier. This is hardly get noticed without browser test.

Looking back at the commits history, this was introduced in https://github.com/livewire/livewire/pull/9597 and follow up by https://github.com/livewire/livewire/pull/9832 but never get fixed. The main problem was logical `||` from this

```js
}

if (isThrottled) {
}
```
From example above, this will produce
```js
debouncedUpdate = throttle(debounce(update, 150), 30)
```

Why? because even if it doesnt have `.debounce` modifier, it still have `.live` modifier which will apply default `150ms` debounce duration.

## Solution
Add guard for `.live` that it's not throttled / debounced
```js
if ((shouldSendNetwork && ! hasNetworkTriggers && isRealtimeInput(el) && ! isDebounced && ! isThrottled) || isDebounced) {
}

if (isThrottled) {
}
```
This way, throttle requests will only get debounced if there is `.debounce` modifier

### DIFF
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -85,7 +85,7 @@ directive('model', ({ el, directive, component, cleanup }) => {
     let debouncedUpdate = update
 
     // Apply debounce/throttle from network modifiers
-    if ((shouldSendNetwork && ! hasNetworkTriggers && isRealtimeInput(el)) || isDebounced) {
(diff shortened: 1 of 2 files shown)


calebporzio:
great, thank you @ghabriel25 !

PR #10494, opened 2026-08-03
[4.x] Fix server-side JS evaluation on mount running before the component is registered

### LINKED ISSUE #10493: Server-side JS evaluation fails to call Livewire actions if not explicitly deferred
### Livewire version

4.3.4

### Laravel version

12

### PHP version

8.5

Chrome on MacOs

After migrating our Livewire 3 codebase to Livewire 4, I noticed a couple hurdles with regards to javascript execution timing. Server-side js evaluation on mount is executed before the Livewire component is registered 
and called actions are "failing" silently because $wire proxy catch error to return a no-op (see screenshot)
Image

It seems like an undocumented breaking change, but also just a bug overall as it does not behave like advertized in the docs

https://wirebox.app/b/x5z3w

Image

Accessing $wire object is spelled out in the documentation:
https://livewire.laravel.com/docs/4.x/javascript#server-side-javascript-evaluation
So one should expect it to work without deferring the javacript calls

lazerg:
When a component calls `$this->js()` inside `mount()`, that JavaScript ships in the initial page payload as a mount-time effect. Livewire ran those effects from the `Component` constructor, before `initComponent` had registered the component in the JS store. So any `$wire` action in that JavaScript couldn't resolve its component, the `$wire` proxy caught the error and returned a silent no-op, and the action never fired.

Processing the initial effects after the component is registered lets those mount-time `$wire` calls resolve normally.

Fixes #10493

### DIFF
--- a/js/component.js
+++ b/js/component.js
@@ -49,9 +49,6 @@ export class Component {
         el.$wire = this.$wire
 
         this.cleanups = []
-
-        // Effects will be processed after every request, but we'll also handle them on initialization.
-        this.processEffects(this.effects)
     }
 
     addActionContext(context) {
--- a/js/store.js
+++ b/js/store.js
@@ -16,6 +16,9 @@ export function initComponent(el) {
 
     components[component.id] = component
 
+    // Process initial effects after registering so mount-time effects like `$this->js()` can resolve `$wire`...
+    component.processEffects(component.effects)
+
     return component
 }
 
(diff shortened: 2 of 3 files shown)


lazerg:
The one red check is a flaky island poll timing test in SupportRequestInteractions, not something this change touches. It passes locally on this branch, and it failed on only one CI shard.

calebporzio:
great, thank you!

Issue #10495, opened 2026-08-03
`#[Renderless]` action suppresses the re-render of other actions batched into the same request


TWithers:
### Livewire version

v4.3.3

### Laravel version

v13.23.0

### PHP version

8.5.8

### Browser and operating system

Dia on MacOS

### Describe the issue you're experiencing

When two component actions are invoked in the same JavaScript tick (e.g. `$wire.setNameToBar(); $wire.toggleComplete()`), Livewire bundles them into a single network request that performs one render at the end of the request. If any action in that batch is marked `#[Renderless]` (or calls `$this->skipRender()`), the skip-render flag is stored on the component for the entire request, so the DOM is never patched, even though a second (non-renderless) action in the same batch mutated state that should update the view. The state change is persisted server-side (a page refresh shows the new values), but the UI silently fails to update. Chaining the calls so they run as two separate requests (`$wire.setNameToBar().then(() => $wire.toggleComplete())`) works around it, because the second request renders normally.

### Code snippets to reproduce the issue

Component (single-file component):

```php
<?php

use Livewire\Attributes\Renderless;
use Livewire\Component;

new class extends Component {
    public bool $isComplete = false;
    public string $name = '';

    #[Renderless]
    public function setNameToBar(): void
    {
        $this->name = 'Bar';
    }

    public function toggleComplete(): void
    {
        $this->isComplete = ! $this->isComplete;
    }
};
```

```html
<div>
    <p>Name: {{ $name ?: '(empty)' }}</p>
    <p>Complete: {{ $isComplete ? 'Yes' : 'No' }}</p>
</div>
```

The JavaScript that triggers it:
```javascript
// ❌ Batched into one request — renderless setNameToBar() suppresses the render,
//    so "Name" and "Complete" do NOT update in the DOM (but do change server-side).
$wire.setNameToBar(); $wire.toggleComplete()

// ✅ Chained into two requests — the second call renders normally, UI updates.
$wire.setNameToBar().then(() => $wire.toggleComplete())
```

### How do you expect it to work?

`#[Renderless]` should only skip rendering for its own action. When a non-renderless action is part of the same batched request and mutates state, Livewire should still re-render the component at the end of that request. In other words, "renderless" should be scoped per-action, not applied to the whole component for the entire request, so batching a renderless call alongside a normal one shouldn't swallow the render the normal call needs. (At minimum, if this is intended, it would help to document that a renderless action suppresses rendering for everything batched with it.)

[...]

PR #10496, opened 2026-08-03
[4.x] Renderless batched render skipped

TWithers:
Resolves issue #10495 

Includes unit and browser test.

**The change:** Instead of skipping immediately, each renderless action now just *records* that it opted out of rendering (a per-request count). After all the actions in the request have run, `callMethods` skips the render only when the count covers **every** action in the request. So a lone renderless action still skips (count == total), but a renderless action batched with a normal one now renders (count < total). Imperative `$this->skipRender()`, event listeners, islands, redirects, slots, and nesting are untouched.


joshhanley:
@TWithers thanks for the PR! I'm closing this as it targets `main`. Can you open a new PR targeting `4.x` branch please? Thanks!

PR #10497, opened 2026-08-03
[4.x] Renderless batched render skipped

TWithers:
Resolves issue https://github.com/livewire/livewire/issues/10495

Includes unit and browser test.

The change: Instead of skipping immediately, each renderless action now just records that it opted out of rendering (a per-request count). After all the actions in the request have run, callMethods skips the render only when the count covers every action in the request. So a lone renderless action still skips (count == total), but a renderless action batched with a normal one now renders (count < total). Imperative $this->skipRender(), event listeners, islands, redirects, slots, and nesting are untouched.

### DIFF
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -557,7 +557,7 @@ class HandleComponents extends Mechanism
 
                 // Support `.renderless` on magic actions like `wire:model.renderless.live`...
                 if ($metadata['renderless'] ?? false) {
-                    $root->skipRender();
+                    $this->markCallRenderless($root);
                 }
 
                 $returns[] = $return;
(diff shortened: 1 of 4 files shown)


src/Mechanisms/HandleComponents/HandleComponents.php
```diff
@@ -594,13 +594,26 @@ protected function callMethods($root, $calls, $componentContext)
...
+        // (e.g. `$wire.renderlessAction(); $wire.normalAction()` collapsed into a
+        // single request) must not suppress the render the normal action needs.
+        if (count($calls) > 0 && store($root)->get('renderlessCallCount', 0) >= count($calls)) {
+            $root->skipRender();
+        }
+
         $componentContext->addEffect('returns', $returns);
     }
 
+    protected function markCallRenderless($root)
```
ghabriel25, line 612: I think this method should be in `HandlesRenderless`

ghabriel25 (changes requested):

calebporzio:
This has been replaced by #10502, which has now been merged. Thanks for the original work here—it directly informed the final fix.

PR #10501, opened 2026-08-04
[4.x] Harden GitHub Actions and add a continuous check

Totara-thib:
Answering the contribution checklist up front: this is a CI hardening fix rather than a feature, so no prior discussion thread; it is on a dedicated branch; it contains three related commits that belong together; it includes no tests because it changes no runtime code, the CI runs themselves are the verification.

**Commit 1 pins every action to its commit sha**, versions kept as comments. A tag like `@v2` is a movable pointer: whoever controls the action, or anyone who compromises it, can re-point it and your next run executes their code with the job's token, including in the release workflow that builds and publishes with `contents: write`. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066). All shas were resolved from the upstream repos and cross checked against their release tags. The pins stay maintainable: a dependabot config with the `github-actions` ecosystem bumps them automatically, one small block.

**Commit 2 adds `permissions: contents: read` to the two test workflows.** All ten test jobs never use the GitHub token, so the scope is exact. The release workflow already declares `contents: write` and is untouched.

**Commit 3 adds [Plumber](https://github.com/getplumber/plumber) to CI**, the tool I used to find the issues in the first place, so none of this quietly drifts back:

- Scans the workflows on each push to 4.x and on each PR, fails when something regresses: an unpinned action, a job without permissions, a known CVE. The gate passes at 85 points, so one small finding does not block your PRs.
- The config is a small overlay inheriting the CLI's built-in baseline: it trusts setup-php and turns off two controls with a comment saying why (branch protection, because 4.x currently has no protection rules, that is a settings choice worth a look; and the release npm cache, scoping that cache key is a good follow-up but its own change). Findings land in the security tab as SARIF.
- Adds a score badge to the README next to your packagist ones. It works like OpenSSF Scorecard's published results: runs publish the score to score.getplumber.io, the badge shows the state of 4.x, a failed publish never fails your CI, and it reads UNKNOWN in gray until the first run.

One heads up: if the org uses an Actions allowlist in settings, patterns written against tags (like `owner/action@v2`) stop matching once refs are shas and workflows refuse to start. Entries need to be `owner/action@*` in that case.

To be fully transparent: I work on Plumber. If you do not want the tool in your CI, no hard feelings, say so and I remove that commit from the PR, the two hardening commits are the part that matters and they stand entirely on their own.

### DIFF
--- a/.github/workflows/plumber.yml
+++ b/.github/workflows/plumber.yml
@@ -1,0 +1,34 @@
+name: Plumber
+
+on:
+  push:
+    branches: [4.x]
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  plumber:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      security-events: write
+      # Needed by score-push to publish the score for the README badge.
+      id-token: write
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+
+      - uses: getplumber/plumber@40bd6b5bb8ff4f0944feacaeb41dea6bce08d62d # v0.4.28
+        with:
+          # Code scanning upload needs security-events write, which PRs
+          # from forks do not get. The report stays available as a
+          # workflow artifact there.
+          upload-sarif: ${{ github.event.pull_request.head.repo.fork != true }}
+          # Publishes the score to score.getplumber.io, which feeds the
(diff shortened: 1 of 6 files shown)


calebporzio:
this feels like an advertisement disguising as a PR. closing thx

Totara-thib:
> this feels like an advertisement disguising as a PR. closing thx

As I said I could remove the commit including Plumber which by the way is a free open source project. 
I understand, cheers!

PR #10502, opened 2026-08-04
[4.x] Fix renderless calls incorrectly skipping render when batched with normal calls

### LINKED ISSUE #10495: `#[Renderless]` action suppresses the re-render of other actions batched into the same request

ghabriel25:
Based on https://github.com/livewire/livewire/pull/10497 so the credit goes to the PR's owner @TWithers .

## Summary
When a renderless action (`#[Renderless]` or `renderless` metadata) is batched in the same request as a normal action, Livewire was calling `skipRender()` as soon as the renderless call ran. That flag is request-scoped, so the later normal call could mutate view state without a re-render, leaving the DOM stale.

## What's changed
- Introduce `markAsRenderless()` and `isRenderless()` on `HandlesRenderless` as a soft per-call flag (does not set `skipRender`).
- `BaseRenderless`and `SupportEvents` now marks the call as renderless instead of calling `skipRender()` immediately.
- In `HandleComponents::callMethods`, count renderless calls across the batch and call `skipRender()` only if every call in the batch is renderless.
- Reset the mark at the start of each call so one renderless call cannot poison the next.
- Add unit tests for mixed batches (attribute + metadata paths).

## Simulated batches
- `[#[Renderless] a, normal b]`
Soft count 1 of 2 → no end `skipRender()` → **renders**.
- `[normal b, #[Renderless] a]`
Reset soft flag each call; count 1 of 2 → **renders**. Order-independent for soft marks.
- `[#[Renderless] a]`
Count 1 of 1 → `skipRender()` → **no render**. Same as main.
- `[metadata renderless, metadata renderless]`
Count 2 of 2 → skip. Same as main’s “all `.renderless`”.
- `[#[Renderless] a, normal b]` **where** `b` **calls** `$this->skipRender()`
Hard flag set in `b` → **no render**, even though soft count is 1. Correct for imperative API.
- `[#[Renderless] a, normal b]` **where** `b` throws
Exception aborts; no successful render response. Soft end-of-loop skip never runs.
- `[__dispatch renderless listener, normal b]`
Events path marks soft; count 1 of 2 → **renders**. Covered by PR test.
- Empty `$calls`
`count($calls) > 0` guard → PR does not call `skipRender()` from the counter. Render still depends on other flags (redirect, island, etc.).
- `forceRender()` **then all-renderless batch**
End `skipRender()` no-ops because of `forceRender` → **renders**. Same guard as main.
- **Double-count risk (attribute + metadata on same call)**
Condition is `isRenderless() || metadata['renderless']` — one increment per call, not two. Safe.

Fix: https://github.com/livewire/livewire/issues/10495

### DIFF
--- a/src/Features/SupportEvents/SupportEvents.php
+++ b/src/Features/SupportEvents/SupportEvents.php
@@ -9,7 +9,6 @@ use Livewire\Features\SupportAttributes\AttributeLevel;
 use Livewire\ComponentHook;
 use Livewire\Exceptions\EventHandlerDoesNotExist;
 use Livewire\Features\SupportAuthorization\BaseAuthorize;
-use Livewire\Features\SupportRenderless\BaseRenderless;
 
 class SupportEvents extends ComponentHook
 {
@@ -38,17 +37,6 @@ class SupportEvents extends ComponentHook
             $returnEarly(
                 wrap($this->component)->$method(...$params)
             );
-
-            // Here we have to manually check to see if the event listener method
-            // is "renderless" as it's normal "call" hook doesn't get run when
-            // the method is called as an event listener...
-            $isRenderless = $this->component->getAttributes()
-                ->filter(fn ($i) => $i instanceof BaseRenderless)
-                ->filter(fn ($i) => $i->getName() === $method)
-                ->filter(fn ($i) => $i->getLevel() === AttributeLevel::METHOD)
-                ->count() > 0;
-
-            if ($isRenderless) $this->component->markAsRenderless();
         }
     }
 
--- a/src/Features/SupportEvents/SupportEvents.php
+++ b/src/Features/SupportEvents/SupportEvents.php
@@ -48,7 +48,7 @@ class SupportEvents extends ComponentHook
                 ->filter(fn ($i) => $i->getLevel() === AttributeLevel::METHOD)
                 ->count() > 0;
 
-            if ($isRenderless) $this->component->skipRender();
+            if ($isRenderless) $this->component->markAsRenderless();
         }
     }
 
--- a/src/Features/SupportRenderless/BaseRenderless.php
+++ b/src/Features/SupportRenderless/BaseRenderless.php
@@ -11,6 +11,6 @@ class BaseRenderless extends LivewireAttribute
     {
         $this->component->skipIslandsRender();
 
-        $this->component->skipRender();
+        $this->component->markAsRenderless();
     }
 }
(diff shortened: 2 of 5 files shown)


TWithers:
I agree with the placement of those methods. Going to go ahead and close my PR in favor of this one.

ghabriel25:
@TWithers I hope you don't mind.

ghabriel25:
Hi @calebporzio @joshhanley, from my perspective, I'm not entirely convinced renderless should be action-scoped or request-scoped, since I don't think we can realistically support both behaviors at the same time.

PR #10505, opened 2026-08-05
[4.x] Prefer `new Collection()` over `collect()` helper

ghabriel25:
## Summary

This PR is a pure refactor that replaces calls to Laravel’s global `collect()` helper with direct instantiation of `new Collection(...)` throughout the Livewire 4.x codebase.

It aligns Livewire with a convention adopted in the Laravel framework itself, which prefers `new Collection()` for:

- Slightly better performance (one fewer function call)
- A shorter call stack
- Makes DX slightly better when navigating the framework
- Consistency across the framework internals

The change has no intended behavioral impact. All existing Collection method chaining, filtering, mapping, etc. continue to work exactly as before.

## Motivation
The `collect()` helper is a thin convenience wrapper without any logic in it. Calling the helper adds an unnecessary extra stack frame and a global function lookup.

Laravel core has begun preferring the explicit constructor in its own source for the reasons listed above. This PR brings Livewire in line with that style.

### DIFF
--- a/src/Drawer/BaseUtils.php
+++ b/src/Drawer/BaseUtils.php
@@ -2,6 +2,8 @@
 
 namespace Livewire\Drawer;
 
+use Illuminate\Support\Collection;
+
 class BaseUtils
 {
     protected static $reflectionCache = [];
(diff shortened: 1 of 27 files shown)


calebporzio:
Thanks for the PR! I’m going to close this because replacing `collect()` across the codebase is a broad style refactor without meaningful behavioral value. We’re avoiding this kind of churn right now.

PR #10507, opened 2026-08-05
[4.x] Consolidate lazy loading setup to `HandlesLazyLoading` trait

ghabriel25:
## Summary

Move the transient `isLazyLoadMounting` / `isLazyLoadHydrating` flags from ad-hoc component-store keys (set/read via `ComponentHook::storeSet/Get`) into an explicit `HandlesLazyLoading` trait on the base `Component`. Other hooks that need to know about the lazy-loading lifecycle now call typed methods on the component instead of reading magic strings from the store.

## Motivation

Several hooks must behave differently while a lazy component is still in its placeholder phase versus when it is actually mounting/hydrating:

- **`SupportEvents`** — must *not* register listeners during the placeholder mount; *must* register them when `__lazyLoad` triggers the real mount.
- **`SupportJsModules`** — must *not* emit the `scriptModule` effect during the placeholder mount (the view hasn’t run, so `@assets` hasn’t either); *must* emit it on the real mount/hydrate.

Previously those hooks reached into the component store with the same string keys that `SupportLazyLoading` wrote. That worked only because `ComponentHook::storeGet/Set` already target `store($this->component)`, but the coupling was implicit and fragile.

This change:

1. Makes the lifecycle state part of the component’s public internal API (same pattern as `skipMount()`, `skipHydrate()`, `HandlesEvents`, `HandlesIslands`, …).
2. Removes the need for other hooks to know the exact store key names.
3. Keeps storage location and runtime behavior identical.

## Changes

**New file**

- `src/Features/SupportLazyLoading/HandlesLazyLoading.php`  
  Trait with:
  - `lazyLoadMounting()` / `lazyLoadHydrating()` — set the flags
  - `isLazyLoadMounting()` / `isLazyLoadHydrating()` — read the flags (default `false`)

**Updated**

- `src/Component.php` — `use HandlesLazyLoading;`
- `src/Features/SupportLazyLoading/SupportLazyLoading.php`
  - `mount()` → `$this->component->lazyLoadMounting()`
  - `hydrate()` → `$this->component->lazyLoadHydrating()`
  - `dehydrate()` → `$this->component->isLazyLoadMounting()` / `isLazyLoadHydrating()`
- `src/Features/SupportEvents/SupportEvents.php` — same boolean checks via the component
- `src/Features/SupportJsModules/SupportJsModules.php` — same boolean checks via the component

**Intentionally unchanged**

- `isLazyIsolated` stays on the hook store — it is only consumed inside `SupportLazyLoading` itself.

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -22,6 +22,7 @@ use Livewire\Concerns\InteractsWithProperties;
 use Illuminate\Support\Traits\Macroable;
 use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 use BadMethodCallException;
+use Livewire\Features\SupportLazyLoading\HandlesLazyLoading;
 
 abstract class Component
 {
@@ -44,6 +45,7 @@ abstract class Component
     use HandlesSlots;
     use HandlesHtmlAttributeForwarding;
     use HandlesRenderless;
+    use HandlesLazyLoading;
 
     protected $__id;
     protected $__name;
--- a/src/Features/SupportEvents/SupportEvents.php
+++ b/src/Features/SupportEvents/SupportEvents.php
@@ -55,9 +55,9 @@ class SupportEvents extends ComponentHook
     function dehydrate($context)
     {
         // Don't register listeners until a lazy component has fully mounted...
-        if ($this->storeGet('isLazyLoadMounting') === true) return;
+        if ($this->component->isLazyLoadMounting()) return;
 
-        if ($context->isMounting() || $this->storeGet('isLazyLoadHydrating') === true) {
+        if ($context->isMounting() || $this->component->isLazyLoadHydrating()) {
             $listeners = static::getListenerEventNames($this->component);
 
             $listeners && $context->addEffect('listeners', $listeners);
--- a/src/Features/SupportJsModules/SupportJsModules.php
+++ b/src/Features/SupportJsModules/SupportJsModules.php
@@ -45,13 +45,13 @@ class SupportJsModules extends ComponentHook
         // Don't add scriptModule effect during lazy-loading placeholder mount.
         // The component's view isn't rendered yet, so @assets won't have run.
         // The scriptModule will be added when __lazyLoad triggers the real mount.
-        if ($this->storeGet('isLazyLoadMounting') === true) return;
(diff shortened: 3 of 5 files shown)


joshhanley:
@ghabriel25 thanks for the PR! My main concern with this is it's adding public methods to the component, where as before it was contained in the feature files.

Now I'm aware we already have some traits like this, but I personally think we should try to eliminate as many methods on the base component class as possible, not add to them. I wouldn't close this, as @calebporzio might have a different opinion on that though!

ghabriel25:
Hi, @joshhanley Your concern is valid and honestly that what I thought before push this PR, however I think this feature should have internal helper as people can create their custom component hook which might need some check if component `isLazyLoadMounting`, `isLazyLoadHydrating` or not.

Hardcoding the key in too many places seems not right at least to me. I tend to make it one source so it easier to understand the flow.

Also I kind of showing my intent in this fix https://github.com/livewire/livewire/pull/10502

joshhanley:
@ghabriel25 yeah I agree the key being scattered all over the place isn't great. I do wonder whether there is a deeper concept here, for like controlling component related stuff like this without necessarily stacking it onto the component itself 🤔

ghabriel25:
@joshhanley I think some kind of forward call mechanism to specific hook would be great or a trait for `ComponentHook` which can be used for all component hooks, something like

```php
trait InteractsWithLazyLoading
{
    protected function markLazyLoadMounting(): void
    {
        $this->storeSet('isLazyLoadMounting', true);
    }

    protected function markLazyLoadHydrating(): void
    {
        $this->storeSet('isLazyLoadHydrating', true);
    }

    protected function isLazyLoadMounting(): bool
    {
        return $this->storeGet('isLazyLoadMounting', false);
    }

    protected function isLazyLoadHydrating(): bool
    {
        return $this->storeGet('isLazyLoadHydrating', false);
    }
}
```

Other component hooks can follow this pattern.

calebporzio:
Thanks for exploring this. I’m going to close it because it restructures internal lazy-loading state without fixing user-facing behavior. We’re keeping the existing implementation to avoid unnecessary API and maintenance surface.

PR #10509, opened 2026-08-06
[4.x] Move lazy-loading lifecycle flags into `InteractsWithLazyLoading` trait

ghabriel25:
This is an alternative for https://github.com/livewire/livewire/pull/10507

## Summary

Extract the transient lazy-loading flags (`isLazyLoadMounting`, `isLazyLoadHydrating`, and `isLazyIsolated`) from magic store keys into a dedicated `InteractsWithLazyLoading` trait that is mixed into the base `ComponentHook` class.

Other hooks that need to know about the lazy-loading lifecycle can now call typed, protected methods instead of reading magic strings from the store.

## Motivation

Several hooks must behave differently while a lazy component is still in its placeholder phase versus when it is actually mounting/hydrating:

- **SupportEvents** — must *not* register listeners during the placeholder mount; *must* register them when `__lazyLoad` triggers the real mount.
- **SupportJsModules** — must *not* emit the `scriptModule` effect during the placeholder mount (the view hasn’t run, so `@assets` hasn’t either); *must* emit it on the real mount/hydrate.

Previously those hooks reached into the component store with the same string keys that `SupportLazyLoading` wrote. That worked only because `ComponentHook::storeGet/Set` already target `store($this->component)`, but the coupling was implicit and fragile.

This change:

1. Makes the lifecycle state an explicit part of the **hook system** (not the public Component API).
2. Removes the need for other hooks to know the exact store key names.
3. Keeps storage location and runtime behavior identical.
4. Avoids adding any new methods to the base `Component` class (addresses the concern raised on the previous approach).

## Changes

**New file**

- `src/Concerns/InteractsWithLazyLoading.php`  
  Trait with protected helpers:
  - `markLazyLoadMounting()` / `markLazyLoadHydrating()` / `markLazyIsolated()`
  - `isLazyLoadMounting()` / `isLazyLoadHydrating()` / `isLazyIsolated()`

**Updated**

- `src/ComponentHook.php` — `use InteractsWithLazyLoading;`
- `src/Features/SupportLazyLoading/SupportLazyLoading.php`
- `src/Features/SupportEvents/SupportEvents.php`
- `src/Features/SupportJsModules/SupportJsModules.php`

**Intentionally unchanged**

- `src/Component.php` — no new public methods

## Pattern for future similar cases

When multiple component hooks need to share transient state that lives on the component store:

1. Create `src/Concerns/InteractsWith{Feature}.php`
2. Mix the trait into `ComponentHook`
3. Use protected `mark*` / `is*` methods inside the hooks
4. Never put these helpers on the public `Component` class unless they are intentionally part of the user-facing API

This keeps the Component surface area small while still giving us a single source of truth and typed access.

### DIFF
--- a/src/ComponentHook.php
+++ b/src/ComponentHook.php
@@ -2,8 +2,12 @@
 
 namespace Livewire;
 
+use Livewire\Concerns\InteractsWithLazyLoading;
+
 abstract class ComponentHook
 {
+    use InteractsWithLazyLoading;
+
     protected $component;
 
     function setComponent($component)
--- a/src/Concerns/InteractsWithLazyLoading.php
+++ b/src/Concerns/InteractsWithLazyLoading.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Livewire\Concerns;
+
+trait InteractsWithLazyLoading
+{
+    protected function markLazyLoadMounting()
+    {
+        $this->storeSet('isLazyLoadMounting', true);
+    }
+
+    protected function markLazyLoadHydrating()
+    {
+        $this->storeSet('isLazyLoadHydrating', true);
+    }
+
+    protected function markLazyIsolated($isolate)
+    {
+        $this->storeSet('isLazyIsolated', $isolate);
+    }
+
+    protected function isLazyLoadMounting()
+    {
+        return $this->storeGet('isLazyLoadMounting', false);
+    }
+
+    protected function isLazyLoadHydrating()
+    {
+        return $this->storeGet('isLazyLoadHydrating', false);
+    }
+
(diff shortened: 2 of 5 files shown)


ghabriel25:
@joshhanley What do you think about this approach? I think this solve the concern about exposing too many method in base component.

If this is going to be preferred flow, this PR https://github.com/livewire/livewire/pull/10499 should be reverted and follow the same pattern as well

calebporzio:
Thanks for exploring this. I’m going to close it because it restructures internal lazy-loading state without fixing user-facing behavior, and we’re avoiding new abstractions in working areas right now.

PR #10511, opened 2026-08-08
[4.x] Cleanup `wire-model.js` from redundant and dead code

ghabriel25:
Inside `js/directives/wire-model.js`, there are 2 redundant codes and 1 dead code
```js
// redundant
function debounce(func, wait) {}
function throttle(func, limit) {}

// dead code
function isDirty(subject, dirty) {}
```

Alpine already has `debounce` and `throttle` that uses **exactly** mechanism.
- https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/utils/debounce.js
- https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/utils/throttle.js

As for dead code, its declared (not exported) but never used inside `wire-model.js` because checking dirty state now delegated to `wire-dirty.js`

### Before
```js
if ((shouldSendNetwork && ! hasNetworkTriggers && isRealtimeInput(el) && ! isDebounced && ! isThrottled) || isDebounced) {
}

if (isThrottled) {
}
```

### After
```js
if ((shouldSendNetwork && ! hasNetworkTriggers && isRealtimeInput(el) && ! isDebounced && ! isThrottled) || isDebounced) {
    debouncedUpdate = Alpine.debounce(debouncedUpdate, parseModifierDuration(networkModifiers, 'debounce') || 150)
}

if (isThrottled) {
    debouncedUpdate = Alpine.throttle(debouncedUpdate, parseModifierDuration(networkModifiers, 'throttle') || 150)
}
```

### DIFF
--- a/js/directives/wire-model.js
+++ b/js/directives/wire-model.js
@@ -86,11 +86,11 @@ directive('model', ({ el, directive, component, cleanup }) => {
 
     // Apply debounce/throttle from network modifiers
     if ((shouldSendNetwork && ! hasNetworkTriggers && isRealtimeInput(el) && ! isDebounced && ! isThrottled) || isDebounced) {
-        debouncedUpdate = debounce(debouncedUpdate, parseModifierDuration(networkModifiers, 'debounce') || 150)
+        debouncedUpdate = Alpine.debounce(debouncedUpdate, parseModifierDuration(networkModifiers, 'debounce') || 150)
     }
 
     if (isThrottled) {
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the PR! I’m going to close this because it’s cleanup of existing working code rather than a bug fix. We’re keeping churn low right now and focusing on behavior-backed fixes or agreed-upon features.

PR #10512, opened 2026-08-08
[4.x] Add `hasModifier()` helper to `Directive` and migrate existing usages

ghabriel25:
## Summary

Adds a `hasModifier(key)` helper to the JavaScript `Directive` class and migrates existing call sites from the low-level `directive.modifiers.includes(...)` pattern.

This is a pure internal refactor with no behavioural changes. It improves consistency and readability, and brings the JS side in line with the existing PHP `WireDirective::hasModifier()` method.

### Motivation

- Many directives repeatedly reach into the raw `modifiers` array.
- A small helper makes the intent clearer and gives us a single place to evolve modifier checking in the future (aliases, normalization, etc.).
- Creates nice symmetry with the PHP API.

### Changes

- Added `Directive#hasModifier(key)` in `js/directives.js`.
- Updated all (or the majority of) call sites in:
  - `shared.js`
  - `wire-confirm.js`
  - `wire-current.js`
  - `wire-ignore.js`
  - `wire-loading.js`
  - `wire-poll.js`
  - `wire-replace.js`
  - `wire-wildcard.js`
  - `supportPreserveScroll.js`
  - `supportWireSort.js`

### DIFF
--- a/js/directives.js
+++ b/js/directives.js
@@ -107,6 +107,10 @@ export class Directive {
         return methods[0].params
     }
 
+    hasModifier(key) {
+        return this.modifiers.includes(key)
+    }
+
     parseOutMethodsAndParams(rawMethod) {
         let methods = []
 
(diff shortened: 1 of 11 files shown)


calebporzio:
Thanks for the PR! I’m going to close this because it’s a refactor of existing working code without a behavior change. We’re keeping churn low right now and focusing on bug fixes or agreed-upon features.

PR #10513, opened 2026-08-09
[4.x] Discard stale live model responses after newer responses land

calebporzio:
## Problem

`wire:model.live` requests are intentionally allowed to overlap so typing does not wait for each previous request to finish. That means they can also finish out of order:

1. Typing `j` sends the first request.
2. Typing `k` sends a newer request containing `jk`.
3. The `jk` response arrives and renders first.
4. The older `j` response arrives afterward and merges its stale snapshot and effects.

The input and server-rendered state can therefore roll backward from `jk` to `j`.

Aborting the older HTTP request is not generally safe because a Livewire request may contain messages for multiple components.

This is an alternative implementation for #10488. The underlying behavior has also been reported in #8341, #2480, and #9030.

## Approach

The existing `model.live` fast path in `interactions.js` is the one place where a new action is intentionally allowed to overlap active messages in the same request scope.

At that point, the incoming action captures the active all-`model.live` messages that it will supersede. If the incoming action's response later lands, those captured messages are stamped as superseded. A superseded response still settles its action and finishes its lifecycle, but it does not merge its snapshot or run its effects.

This keeps the rule local and avoids a separate version registry:

- No global counters or `WeakMap` ordering state.
- Component and island scopes remain independent through the existing message bus scope matching.
- Every active older live-model message is captured, so three or more overlapping requests work the same way as two.
- Mixed and non-live-model messages are not stamped.
- Existing bound-child behavior remains serialized.

The important distinction is that sending a newer request does not immediately invalidate an older one. The older message is stamped only when the newer response is accepted. If the newer request is cancelled or fails at the network or HTTP layer, the older successful response can still land.

Skipped component payloads count as accepted responses and also stamp the older messages, since allowing an older snapshot afterward would still roll the component backward.

## Testing

One PHP browser test recreates the user-visible `j` / `jk` failure deterministically. It lets the first HTTP response complete but holds that first browser `fetch` promise for 500ms, allowing the second response to render first. After releasing the older response, both the input and server-rendered value must remain `jk`.

The browser test fails on unchanged `4.x` with the rendered value rolling back, and passes with this change.

```sh
php vendor/bin/phpunit --configuration phpunit.xml.dist src/Tests/LiveModelResponseOrderingBrowserTest.php
npm run test:run
npm run build
```


iBotPeaches:
thanks! Learned a bit reviewing this to see where my PR failed

PR #10514, opened 2026-08-09
[4.x] Introduce `Effects` class and replace `Object.prototype.hasOwnProperty.call` checks

ghabriel25:
## Summary
Introduce a small `Effects` class for Livewire’s server-driven side-effect payloads, and replace every `Object.prototype.hasOwnProperty.call(effects, …)` check with `effects.has()` / `effects.hasValue()`. Construction is centralized at the few places effects enter the system so the rest of the codebase can use a clear API without changing behavior.

## Why this is needed
Effects are a plain object parsed from `wire:effects` or from the JSON response (`redirect`, `html`, `listeners`, `scripts`, `download`, `dispatches`, `styleModule`, `globalStyleModule`, etc.). Feature code repeatedly asks **“does this payload include key X?”** in the most defensive way possible:

```js
Object.prototype.hasOwnProperty.call(effects, 'globalStyleModule') && effects.globalStyleModule
```

That pattern appears **20 times** across core and feature files. It is:
1. **Hard to read** — the intent is “does this effect exist?”, not “borrow `hasOwnProperty` from `Object.prototype`.”
2. **Easy to get wrong** — some sites check ownership only, others ownership and truthiness; the difference is easy to miss in review.
3. **Inconsistent** — mix of `effects.key`, `effects['key']`, and the long `hasOwnProperty` form for the same idea.
4. **Awkward to extend** — any new helper (safe get, “has non-empty value”, iteration helpers) would be duplicated or require another shared util with no clear owner.

Livewire already treats effects as a distinct concept (`constructor`, `mergeNewSnapshot`, `processEffects`, `inscribeSnapshotAndEffectsOnElement`, request success path). Giving that concept a type makes the boundary explicit and removes the need for the low-level ownership ceremony at every call site.

## What's Changed
1. **New `js/effects.js`**
    - Own keys stay on the instance so existing `effects.redirect`, `effects.html`, etc. keep working.
    - `has()` / `hasValue()` encode 2 patterns already used in the codebase (ownership vs ownership + truthy).
    - `toJSON` keeps `JSON.stringify` / `wire:effects` attributes as plain data.
2. **Construct `Effects` only at entry points**

| Files | Where |
| - | - |
| `js/component.js` | Constructor (`wire:effects`), `mergeNewSnapshot`, `replayUpdate`, `processEffects` |
| `js/request/index.js` | After unpacking `effects` from the component payload on success |

After these sites, `component.effects`, `component.originalEffects`, `message.responsePayload.effects`, and the `effects` argument to `on('effect', …)` are always `Effects` instances (or normalized inside `processEffects`).

3. **Replace every effects ownership check**

| File | Keys |
| - | - |
| `js/component.js` | `listeners`, `url`, `scripts` (on `originalEffects` when inscribing) |
| `js/request/message.js` | `returns`, `returnsMeta` |
| `js/features/supportCssModules.js` | `styleModule`, `globalStyleModule` |
| `js/features/supportListeners.js` | `listeners` |
| `js/features/supportScriptsAndAssets.js` | `scripts` |
| `js/features/supportRedirects.js` | `redirect` |
| `js/features/supportNavigate.js` | `redirectUsingNavigate` |
| `js/features/supportFileDownloads.js` | `download` |
| `js/features/supportDispatches.js` | `dispatches` |
| `js/features/supportQueryString.js` | `url` |
| `js/features/supportJsEvaluation.js` | `js`, `xjs` |
| `js/features/supportJsModules.js` | `scriptModule` |
| `js/features/supportMorphDom.js` | `html` |
| `js/features/supportIslands.js` | `islandFragments` |
| `js/features/supportLaravelEcho.js` | `listeners` |

Example:
```js
// before
if (Object.prototype.hasOwnProperty.call(effects, 'redirect')) { ... }

if (Object.prototype.hasOwnProperty.call(effects, 'globalStyleModule') && effects.globalStyleModule) { ... }

// after
if (effects.has('redirect')) { ... }

if (effects.hasValue('globalStyleModule')) { ... }
```

### DIFF
--- a/js/component.js
+++ b/js/component.js
@@ -29,8 +29,8 @@ export class Component {
 
         this.name = this.snapshot.memo.name
 
-        this.effects = new Effects(JSON.parse(el.getAttribute('wire:effects') ?? '{}'))
-        this.originalEffects = new Effects(deepClone(this.effects))
+        this.effects = Effects.from(JSON.parse(el.getAttribute('wire:effects') ?? '{}'))
+        this.originalEffects = Effects.from(deepClone(this.effects))
 
         // "canonical" data represents the last known server state.
         this.canonical = extractData(deepClone(this.snapshot.data))
@@ -83,7 +83,7 @@ export class Component {
 
         this.snapshot = snapshot
 
-        this.effects = effects instanceof Effects ? effects : new Effects(effects)
+        this.effects = Effects.from(effects)
 
         this.canonical = extractData(deepClone(snapshot.data))
 
@@ -151,7 +151,7 @@ export class Component {
      * users interact with, triggering reactive effects.
      */
     processEffects(effects, request) {
-        effects = effects instanceof Effects ? effects : new Effects(effects)
+        effects = Effects.from(effects)
 
         // This is for BC.
         trigger('effects', this, effects)
--- a/js/component.js
+++ b/js/component.js
@@ -29,7 +29,7 @@ export class Component {
 
         this.name = this.snapshot.memo.name
 
-        this.effects = new Effects(JSON.parse(el.getAttribute('wire:effects') || {}))
+        this.effects = new Effects(JSON.parse(el.getAttribute('wire:effects') ?? '{}'))
         this.originalEffects = new Effects(deepClone(this.effects))
 
         // "canonical" data represents the last known server state.
--- a/js/component.js
+++ b/js/component.js
@@ -3,6 +3,7 @@ import { generateWireObject } from '@/$wire'
 import { findComponentByEl, findComponent, hasComponent } from '@/store'
 import { trigger } from '@/hooks'
 import { setNextActionOrigin } from '@/request'
+import { Effects } from '@/effects'
 
 export class Component {
     constructor(el) {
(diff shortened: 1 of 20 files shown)


calebporzio:
Hey @ghabriel25 - thank you for all the PRs lately, however, in the interest of keep our codebase stable right now, I'm going to ask that we don't refactor existing/functioning/tested areas of the codebase and only focus PRs on bugfixes or occasionally new features.

Also, if there is a recurring AI task or something finding and generating these I'd ask that you disable it in the interest of keeping the maintanance burden under control for me as I have to manually review each one.

Thanks for all your effort!

ghabriel25:
@calebporzio I understand. Just for information, I always craft the code by myself and never use AI for such thing. However, I do use AI to generate PR description. I apologize for the inconvenience.

calebporzio:
Oh no way, that's actually great to hear. And thank you more broadly for all your quality contributions lately.

Don't hesitate to PR something if you're compelled, but yeah I'm more interested in features and bugfixes than refactors ATM.

PR #10529, opened 2026-08-11
Batch implicit island morphs and preserve ordered output

calebporzio:
## What changed

- Direct `renderIsland()` calls are retained as repeatable explicit commands.
- Implicit `morph` renders are queued once per island target and flushed after every call has finished, using final component state.
- Implicit `append` and `prepend` renders execute at each call's finisher and retain every intermediate state.
- Buffered island fragments are applied sequentially in response order.

## Why

Island replacement is state synchronization. Rendering after each call captures an intermediate state, so implicit morphs are deferred and coalesced into one final-state render.

Append, prepend, and direct `renderIsland()` calls are ordered output commands. Their repetition and call-time state are observable, so each command must be retained and applied in order.

Sequential browser application prevents an earlier async morph from finishing late and clobbering later output.

## Test coverage

- A batched implicit `+1` then `+10` morph renders once and paints final state `11`.
- Batched implicit appends produce `0, 1, 11`.
- Batched implicit prepends produce `11, 1, 0`.
- An explicit morph at state `1` followed by an explicit append at state `2` leaves both nodes in order.

Validation performed locally:

- Focused island browser tests: 3 passed, 6 assertions
- SupportIslands unit tests: 22 passed, 38 assertions
- Livewire JavaScript bundle rebuilt for the linked playground

## Follow-up boundary

Mixed implicit morph and append/prepend batches still need a whole-target ordering decision. This PR establishes the two independent behaviors without attempting that larger planning step.

### DIFF
--- a/js/features/index.js
+++ b/js/features/index.js
@@ -4,6 +4,7 @@
 import './supportListeners';
 import './supportScriptsAndAssets';
 import './supportJsEvaluation';
+import './supportIslands';
 import './supportMorphDom';
 import './supportDispatches';
 
@@ -17,7 +18,6 @@ import './supportQueryString';
 import './supportLaravelEcho';
 import './supportStreaming';
 import './supportRedirects';
-import './supportIslands';
 import './supportNavigate';
 import './supportEntangle';
 import './supportSlots';
--- a/js/features/supportIslands.js
+++ b/js/features/supportIslands.js
@@ -61,9 +61,9 @@ interceptMessage(({ message, onSuccess, onStream }) => {
                 fragments = payload.effects.islandFragments
             }
 
-            fragments.forEach(async fragmentHtml => {
+            for (let fragmentHtml of fragments) {
                 await renderIsland(message.component, fragmentHtml)
-            })
+            }
         })
     })
 })
--- a/js/request/message.js
+++ b/js/request/message.js
@@ -224,9 +224,9 @@ export default class Message {
     }
 
     async invokeOnMorph() {
-        await Promise.all(
-            this.interceptors.map(interceptor => interceptor.onMorph())
-        )
+        for (let interceptor of this.interceptors) {
+            await interceptor.onMorph()
+        }
     }
 
     invokeOnRender() {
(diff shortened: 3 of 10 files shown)


calebporzio:
## Island rendering behavior we settled on

This is the behavioral contract we worked toward while testing the combinations in the playground.

### Included in this PR

- **Implicit `morph` renders are batched per island target.** If several normal calls target the same island in one request, it renders once after all calls using the final component state.
- **Implicit `append` / `prepend` renders happen in place.** Every eligible call emits a fragment, in call order, with the state from that point in the request.
- **Explicit `renderIsland()` calls are never deduplicated.** Every direct call emits a fragment and the browser applies them in the same order the server sent them.
- **`.renderless` and `#[Renderless]` are call-local.** They suppress that call's implicit island render, but do not suppress an explicit `renderIsland()` and do not prevent a later normal call from producing the final implicit morph.
- **Buffered island fragments are applied sequentially in the browser.** A later fragment cannot finish before and be clobbered by an earlier one.
- **Island fragments finish before the root morph.** The root render therefore observes the DOM produced by the island fragments instead of racing them.
- **Linked island instances keep their normal fan-out behavior.** One logical island render still produces the corresponding fragment for each mounted token.
- **Lazy islands keep lazy mounting when implicit morphs are coalesced.**

### Intentionally deferred

- **Mixed implicit modes for the same island in one request** (for example `morph`, then `append`) do not yet preserve the complete call timeline. The ideal behavior is exact server call order, but the implementation we tried required the general component-call system and islands to know too much about each other, so we reverted it.
- **`skipRender()` is not yet island-target-aware.** The ideal behavior is a hard veto scoped to the render target instead of a component-wide ambiguous switch.
- **Overlapping streamed island morphs** are still allowed to overlap. Fully serial streamed application can be revisited separately.
- **`replace` is not an island mode.** The temporary server-side reference was removed; unrelated stream replacement behavior is outside this PR.

The guiding rule is: **the response should describe the intended DOM operations in server order, and the browser should apply those operations sequentially in that order.**

PR #10533, opened 2026-08-12
[4.x] Fix `Route::lazy()` and `Route::defer()` macros not available for package service providers

marcheffels:

This fixes the same class of bug as #9903, just for the `lazy()`/`defer()` route macros instead of `Route::livewire()`.

Yes.

No, single focused change.

Not yet, that's why this is a draft. I'd appreciate some help on how to best write a test for this, if one is needed.

5️⃣ Please include a thorough description...

# The Scenario

When a third-party package uses `Route::lazy()` or `Route::defer()` in its service provider to register a lazy-loaded Livewire route, the application throws a `BadMethodCallException` because the macros aren't registered yet at the time the package's routes are loaded.

# The Problem

The `lazy()` and `defer()` macros were registered in `SupportLazyLoading::provide()`, which runs during `LivewireServiceProvider::boot()`. Laravel's package discovery loads and boots service providers in an unpredictable order. If a third-party package's service provider boots *before* Livewire's service provider, the `lazy()`/`defer()` macros don't exist yet.

This is the same underlying issue that was fixed for `Route::livewire()` in #9903.

# The Solution

Register the `lazy()`/`defer()` macros inside `HandleRouting::register()` (a mechanism). Mechanisms run their `register()` phase before any service provider's `boot()` phase runs, so this guarantees the macros are available before any package's `boot()` method runs, regardless of provider boot order.

### DIFF
--- a/src/Features/SupportLazyLoading/SupportLazyLoading.php
+++ b/src/Features/SupportLazyLoading/SupportLazyLoading.php
@@ -25,8 +25,6 @@ class SupportLazyLoading extends ComponentHook
 
     static function provide()
     {
-        static::registerRouteMacro();
-
         on('flush-state', function () {
             static::$disableWhileTesting = false;
         });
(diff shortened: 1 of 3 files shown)


calebporzio:
thanks!

PR #10534, opened 2026-08-12
[4.x] Add `#[Debounce]` attribute for property and method

ghabriel25:
> [!NOTE]
> I intentionally filter the attribute for `wire:model`, `wire:click`, and `wire:submit` because thats all I find on Livewire documentation ([wire:model modifiers](https://livewire.laravel.com/docs/4.x/wire-model#modifiers), [wire:click modifiers](https://livewire.laravel.com/docs/4.x/wire-click#modifiers), [wire:submit modifiers](https://livewire.laravel.com/docs/4.x/wire-submit#modifiers)
>
> I prefer not to include all wildcard that might caused issue later on

# Summary

Adds a `#[Debounce]` PHP attribute so properties **and** methods can declare their debounce duration once in PHP, without repeating `.debounce.Xms` on every `wire:model.live`, `wire:click`, or `wire:submit` binding.

### Before

```blade
<input wire:model.live.debounce.250ms="search">
<button wire:click.debounce.500ms="save">Save</button>
<form wire:submit.debounce.500ms="save">...</form>
```

### After

```php
#[Debounce(250)]
public string $search = '';

#[Debounce(500)]
public function save() { /* ... */ }
```

```blade
<input wire:model.live="search">
<form wire:submit="save">...</form>
```

Omitting the duration uses Livewire’s existing default of `150ms` for properties and `250ms` for methods:

```php
#[Debounce]
public string $search = '';

#[Debounce]
public function save() { ... }
```

### Properties (`wire:model`)

| Blade | Attribute | Network timing |
| --- | --- | --- |
| `wire:model.live="search"` | `#[Debounce]` | `150ms` (default) |
| `wire:model.live="search"` | none | `150ms` (default for realtime inputs) |
| `wire:model.live="search"` | `#[Debounce(250)]` | `250ms` (attribute) |
| `wire:model.live.debounce.500ms="search"` | `#[Debounce(250)]` | `500ms` (**Blade wins**) |
| `wire:model="search"` | `#[Debounce(250)]` | still deferred — attribute does **not** force `.live` |
| `wire:model.live.blur="search"` | `#[Debounce(2000)]` | no attribute debounce (explicit network trigger) |
| `wire:model.lazy="search"` | `#[Debounce(2000)]` | no attribute debounce (`.lazy` → change trigger) |

The attribute only supplies the **duration**. It does not decide whether a network request is sent. Users still need `.live` (or another network trigger) for live updates.

### Methods (`wire:click`, `wire:submit`, …)

| Blade | Attribute | Result |
| --- | --- | --- |
| `wire:click="save"` | `#[Debounce(500)]` on `save` | Alpine `.debounce.500ms` applied |
| `wire:click="save()"` / `wire:click="save(1)"` | same | method name stripped before `(` for effect lookup |
| `wire:click.debounce.100ms="save"` | `#[Debounce(500)]` | **Blade wins** (attribute not appended) |
| `wire:submit="save"` | `#[Debounce(500)]` on `save` | same path as `wire:click` via `wire-wildcard` |

### Duration normalization (PHP)

| Input | Normalized |
| --- | --- |
| `#[Debounce]` / omitted | properties: `150ms`, methods: `250ms` |
| `#[Debounce(250)]` | `250ms` |
| `#[Debounce(0)]` | `0` |
| `#[Debounce(-100)]` | `0` (clamped) |
| `#[Debounce('300ms')]` / `'300MS'` | `300ms` |
| `#[Debounce('0.5s')]` / `'2s'` | `500ms` / `2000ms` |
| `#[Debounce('250')]` | `250ms` (bare number → ms) |
| `#[Debounce('fast')]` / `''` / `'ms'` | properties: `150ms`, methods: `250ms` (fallback) |

### Other supported targets

- **Form object properties** — effect key is the full path (`form.q`), matching `wire:model.live="form.q"`
- **Nested array properties** — e.g. `filters.q` when `#[Debounce]` is on `$filters`
- **`$parent` expressions** — `wire:model.live="$parent.search"` resolves debounce from the parent component’s effects
- **History navigation** — `debounce` is re-inscribed from `originalEffects` (same durability pattern as `url` / `scripts` / `listeners`)

## What's changed

| Area | File |
| --- | --- |
| Feature class | `src/Features/SupportDebounce/BaseDebounce.php` |
| Public alias | `src/Attributes/Debounce.php` |
| Property network timing | `js/directives/wire-model.js` |
| Action directives (`click`, `submit`, …) | `js/directives/wire-wildcard.js` |
| Persist effects across history | `js/component.js` |
| Browser tests | `src/Features/SupportDebounce/BrowserTest.php` |
| Unit tests | `src/Features/SupportDebounce/UnitTest.php` |

### Design notes (aligned with existing Livewire patterns)

- **Mount-only dehydrate** — same contract as `BaseUrl`: effect is pushed only while `$context->isMounting()`, so subsequent requests do not re-send the map.
- **Effect key** — `$this->getName()` (property path or method name), including dotted form-object paths.
- **JS lookup** — prefers `originalEffects.debounce` over `effects.debounce` so the map survives `mergeNewSnapshot`.
- **Precedence** — Blade `.debounce.Xms` always wins over the attribute for both models and actions.
- **Method expressions** — name is taken as the substring before `(` so `save`, `save()`, and `save(1)` all resolve to the `save` effect key.
- **Alpine format** — actions append `.debounce.${duration}ms` (two modifiers), matching Alpine’s expected shape.

## Testing

**Unit**

- Property / method effect push, defaults, string normalization, zero/negative clamp
- Annotated-only isolation; property + method independent durations
- Mount-only (effect absent on subsequent request)
- Form object dotted key (`form.q`)

**Browser**

- Attribute delays `wire:model.live` network updates
- Blade modifier overrides attribute (property + form object)
- Attribute does not force `.live` on deferred models
- Only annotated property/method is delayed
- Form object, nested array, `$parent` expressions
- Explicit triggers (`.blur`, `.enter`, `.lazy`) ignore attribute duration
- `.live.throttle` still respects attribute debounce window
- Zero-duration attribute
- History back/forward keeps attribute debounce
- `wire:click` / parameterized method / `wire:submit` rapid fire collapses to one request
- Property + method independence (different durations, sequential asserts)


calebporzio:
Hmmm this is a tricky one. I like the concept but have the following hangups:

- Merging this will open the door for many more features like it: Throttle, Once, etc... i imagine. In fact if we merged this, people would expect Throttle to exist as well. Also maybe even #[Live] to control weather a property should be wire:model.live by default you know?
- It's not applied to wildcard listeners which could trip some people up
- I don't think I've personally needed this. It's not that often that I have a large amount of repitition of a listener AND it's debounced you know?

So all in all, I think I want to punt on this feature right now. I don't think there's a strong enough case for it just yet. thanks for the effort!

PR #10536, opened 2026-08-14
[4.x] Fix reactive Eloquent models throws `CannotMutateReactivePropException` when relations are lazy-loaded

ghabriel25:
## Problem

`BaseReactive` detected changes with:

```php
crc32(json_encode($this->getValue()))
```

For Eloquent models, `json_encode()` includes loaded relations. A normal app pattern:

```php
#[Reactive]
public Article $article;

public function render()
{
    return view('...', [
        'author' => $this->article->author->name, // lazy load
    ]);
}
```

Lazy load model mutates the in-memory parent model (relation loaded), changes the hash, and throws on dehydrate:
`Cannot mutate reactive prop [article] in component: [...]`

That conflicts with how Livewire treats models elsewhere:
- `ModelSynth` only dehydrates **class + key** (no attributes, no relations)
- `SupportReactiveProps::modelMatchesSnapshot()` compares identity + dirty/changed flags for skip, and also ignores relations

So mutation detection was stricter (and noisier) than the rest of the model pipeline.

`BaseReactive` also lacked the `json_encode === false` guard that `valuesMatch()` already has (NAN/INF/etc. can collide on `crc32(false)`).

## Solution
Add a shared hasher on `SupportReactiveProps` and use it from `BaseReactive` for mount / hydrate / dehydrate:
- **Models:** hash `class` + `key` + `getAttributes()` — not relations
  - Relation reads (`$this->article->author` — lazy load) → no throw
  - Attribute writes (`$this->article->title = 'New title'`) → still throw
  - Different model identity → still throw
- **Other values:** `crc32(json_encode($value))`, or `null` if encoding fails (same idea as `valuesMatch`)

```php
static function hashValue(mixed $value): ?string
{
    if ($value instanceof Model) {
        $class = $value::class;

        $morphMap = Relation::morphMap();

        $alias = in_array($class, $morphMap)
            ? array_search($class, $morphMap, true)
            : $class;

        $payload = json_encode([
            'class' => $alias,
            'key' => $value->getKey(),
            'attributes' => $value->getAttributes(),
        ]);

        return $payload === false ? null : (string) crc32($payload);
    }

    $json = json_encode($value);

    return $json === false ? null : (string) crc32($json);
}
```

`shouldSkipUpdate()` / `modelMatchesSnapshot()` are **unchanged**. Skip remains identity-based, consistent with `ModelSynth`.

> [!IMPORTANT]
> There are still some issues thats not fixed with this PR as I'm not sure if we should support these cases:
> 1. Reactive collections / arrays of models – still use full json_encode, so the original bug can still appear.
> 2. Custom objects that implement JsonSerializable and embed models

## Tests
Five tests:
- One regression test (accessing lazy load relation)
- Four complementary tests to make sure this PR doesnt break anything

### Why attributes are in the model hash
Hashing attributes preserves the existing reactive contract (“child must not mutate the prop”) while excluding relations so display-only lazy loads are safe.
`isDirty()` / `wasChanged()` alone are not used on dehydrate: the parent may pass an already-dirty model; those flags would false-positive after `setValue`. Comparing attribute snapshots from hydrate → dehydrate avoids that.

### Why cast `crc32` value to string
On `32-bit` builds the value from `crc32` can be negative because PHP integers are signed. On `64-bit` builds it is always non-negative. See [PHP Manual](https://www.php.net/manual/en/function.crc32.php)

Discussions: https://github.com/livewire/livewire/discussions/6090, https://github.com/livewire/livewire/discussions/6033

### DIFF
--- a/src/Features/SupportReactiveProps/BaseReactive.php
+++ b/src/Features/SupportReactiveProps/BaseReactive.php
@@ -17,7 +17,7 @@ class BaseReactive extends LivewireAttribute
 
         $this->storePush('reactiveProps', $property);
 
-        $this->originalValueHash = crc32(json_encode($this->getValue()));
+        $this->originalValueHash = SupportReactiveProps::hashValue($this->getValue());
     }
 
     public function hydrate()
@@ -32,13 +32,10 @@ class BaseReactive extends LivewireAttribute
 
             // Only trigger lifecycle hooks if value actually changed
             // Use the same comparison method as dehydrate() for consistency (handles arrays/objects)
-            $currentHash = crc32(json_encode($currentValue));
-            $updatedHash = crc32(json_encode($updatedValue));
-
-            if ($currentHash !== $updatedHash) {
+            if (! SupportReactiveProps::hashesMatch($currentValue, $updatedValue)) {
                 // Set value immediately so lifecycle hooks (boot/hydrate/booted) see updated data
                 $this->setValue($updatedValue);
-                $this->originalValueHash = crc32(json_encode($updatedValue));
+                $this->originalValueHash = SupportReactiveProps::hashValue($updatedValue);
 
                 // Queue the update trigger for after hydrate so updating*/updated* hooks fire
                 // Pass both old and new values plus a setValue callback so that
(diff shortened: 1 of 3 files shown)


ghabriel25:
Already mark as ready to be reviewed but forgot to change the text in the tests. I shouldnt include my nickname whatsoever thats unrelated to Livewire

PR #10539, opened 2026-08-14
Keep loading active through navigate redirects

calebporzio:
## Goal

Treat a same-origin `navigate: true` redirect as part of the action's loading lifecycle without saving transient loading UI into navigation history.

## Requirements

- [x] Keep existing developer-declared `wire:loading` state active while destination HTML is fetching.
- [x] Keep automatic `data-loading` indicators and form protection active through the destination fetch.
- [x] Clear transient loading state before serializing the departing page into history.
- [x] Restore a clean, non-loading page through Back/Forward navigation.
- [x] Clear loading when navigation fails, is cancelled, or is prevented.
- [x] Affect only elements already participating in loading behavior; do not introduce a global page lock.
- [x] Leave ordinary actions and non-navigate redirects unchanged.
- [x] Avoid leaked listeners or stale state across repeated navigations.

## Incremental plan

1. Define the forward-navigation contract with a failing browser test.
2. Keep loading active while destination HTML is pending.
3. Give Navigate ownership of a shared navigation lifecycle.
4. Move loading cleanup onto that lifecycle before the history snapshot.
5. Add Back/Forward restoration coverage.
6. Cover loading variants and listener cleanup.

## Completed behavior

Navigate owns an active transaction with four meaningful phases:

- `start` occurs only after the cancelable navigate event is accepted.
- `ready` occurs after destination HTML is available and before the departing page is serialized.
- `cancelled` occurs when the destination fetch fails or is aborted.
- `finished` occurs after the new page has initialized and `alpine:navigated` fires.

`wire:loading` watches for navigation only while successful response effects are being processed: after `onSuccess` and before `onEffect`. If an effect starts navigation, loading is handed to that transaction until its destination is ready or cancelled. Otherwise loading finishes normally before morphing. Request cancellation, errors, and failures continue through the existing `onFinish` fallback.

Automatic request-owned busy states use the same narrow detection window through `releaseAfterNavigation()`. This keeps the action-origin `data-loading` marker, submit-button disabling, and form input read-only state active until the destination is ready or cancelled. Without navigation they retain their existing request-finish cleanup.

The navigation transaction owns the paired `ready` and `cancelled` listener cleanup behind `onDestinationSettled()`. The callback carries no duplicated status value; a consumer that needs the outcome reads `navigation.state` or uses a phase-specific hook. This removes all knowledge of `redirect` and `redirectUsingNavigate` effects from loading features; only Navigate interprets those effects.

Browser coverage runs the complete lifecycle: loading during a delayed destination fetch, synchronized visual and disabled states, protected form controls, clean Back/Forward restoration, a second successful navigation from restored history, recovery after cancellation, and prevention.

## Verification

- Ten focused browser tests pass with 67 assertions across redirect loading, automatic `data-loading`, and form protection.
- The Flux playground keeps the spinner and disabled state synchronized for the full destination fetch.
- Back/Forward restores the original button without transient busy attributes.
- A successful retry after an aborted navigation confirms the lifecycle does not retain stale cancellation state.
- The playground reports `disabled`, `data-loading`, and visible spinner state together and has no console errors.
- Production JavaScript build passes.
- `git diff --check` passes.


people-sea:
Hi @calebporzio, any chance this could be backported to 3.x? I had a quick look, but it seems like a fairly involved change.

PR #10543, opened 2026-08-15
[4.x] Support exception hook for computed properties

ghabriel25:
## Summary
Allow exceptions thrown from `#[Computed]` attribute to trigger the component's `exception` hook.

## Problem
Exception thrown inside computed properties was ignored by component exception hook while there is non-zero chance it can throw exception such `ModelNotFoundException`. Inside `BaseComputed::evaluateComputed()`, the component wrapped with `invade()` not `wrap()` which is make sense since computed properties doesnt have any parameters to resolved.

## Solution
Using `wrap()` + `Component::tap()`  following how `BaseAuthorize::call()` forwards the exception to component exception hook.

> [!NOTE]
> Replacing `invade()` with `wrap()` will make it more simple **but** might cause breaking change because computed properties could be a protected method

```php
protected function evaluateComputed()
{
    $value = null;

    wrap($this->component)->tap(function ($component) use (&$value) {
        $value = invade($component)->{parent::getName()}();
    });

    return $value;
}
```

## Test
- Add unit test for the exception hook on computed properties

Related discussion: https://github.com/livewire/livewire/discussions/8373

### DIFF
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -3,6 +3,7 @@
 namespace Livewire\Features\SupportComputed;
 
 use function Livewire\invade;
+use function Livewire\trigger;
 
 use Livewire\Features\SupportAttributes\Attribute;
 use Illuminate\Support\Facades\Cache;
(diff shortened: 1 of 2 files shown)


calebporzio:
thanks!

PR #10548, opened 2026-08-17
[4.x] Fall back to Eloquent casts when filling typed properties

ghabriel25:
## What this fixes

Passing an Eloquent model to `Form::fill()` serializes it with `toArray()`. Backed enum casts therefore become strings before Livewire assigns them to the form, causing a `TypeError` when the destination property is typed as that enum.

```php
class UserForm extends Form
{
    public ?UserStatus $status;
}

$this->form->fill($user);
```

## Approach

Keep the existing serialized `toArray()` value as the default. If PHP rejects that value because it does not satisfy the destination property type, retry the assignment with Eloquent's native `getAttribute()` value.

This lets Eloquent supply the enum, date, or custom-cast object required by a typed property without introducing separate reflection and union-type matching logic.

## Compatibility

- Untyped and already-compatible properties keep their existing serialized values.
- Loaded relations remain serialized arrays when the destination accepts them.
- Typed properties can fall back to Eloquent's native cast value.
- Invalid native values still raise the original kind of `TypeError` instead of being silently coerced.

## Validation

- Added a regression test for filling a typed backed-enum Form property from an enum-casted model attribute.
- Added a compatibility test proving an untyped property still receives the serialized string.
- Focused Form and fill suites: 53 tests, 295 assertions.
- Full unit suite: 1,378 tests, 3,921 assertions, with the repository's existing skips and incomplete tests.

Fixes the behavior reported in https://github.com/livewire/livewire/discussions/9689.


ghabriel25:
@calebporzio @joshhanley I think this could lead breaking change for something like

```php
new class extends Component
{
    public Post $post;
    public $created_at;

    public function mount()
    {
        $this->fill($this->post);
    }
}
```

**Before**
`$created_at` expected to be a `string`

**After**
`$created_at` now instance of `Carbon`

calebporzio:
i went with a simpler strategy. thanks!

ghabriel25:
Thanks, really insightfull!

PR #10559, opened 2026-08-20
[4.x] Fix casted Eloquent attributes bypassing validation under legacy model binding

ghabriel25:
## Affected version
- 3.x (Backport)
- 4.x (Current PR)

## Summary

When validating nested Eloquent model properties (especially with **legacy model binding**), Livewire prepared validation data via `Model::toArray()`, which applies attribute **casts**. That meant user input could be mutated before the validator ran—for example `32.5` on an `integer`-cast field became `32`, so `integer` rules passed incorrectly.

This delegate unwraps models to `EloquentModelSynth` (legacy models) instead of `toArray()` (casted values), so validation sees the values the user actually submitted.

## The problem

Minimal snippet to reprocude the issue
```blade
<div>
    <input wire:model="user.age" />
</div>
```

```php
'legacy_model_binding' => true,
```

```php
<?php

use App\Models\User;
use Livewire\Component;

new class extends Component
{
    public User $user;

    protected function rules(): array
    {
        return [
            'user.age' => 'required|integer'
        ];
    }

    public function save()
    {
        $this->validate(); // this is actually passed the validation

        $this->user->save();
    }
};
```
1. User enters `32.5` into `wire:model="user.age"`.
2. On `$this->validate()`, data was unwrapped with `$model->toArray()`.
3. Cast turned `32.5` → `32` **before** the validator ran.
4. The `integer` rule passed even though the input was not an integer.

## Solution

In `HandlesValidation::unwrapDataForValidation()`:
- Prefer **validate before cast**
- Prefer synthesizer `unwrapForValidation()` when available.
- For `Model` instances, delegate to `EloquentModelSynth` legacy models instead of falling through to `Arrayable::toArray()`
- Keep existing `Arrayable` behavior for other values.

## What is not fully covered
1. **Boolean soft values** — `'yes'`, `'true'`, `''` often won’t restore because of PHP’s loose `!=` with `true/false`.
2. **Float type-only changes** — `'3.14'` vs `3.14` not restored (`!=` is false).
3. **Dates** — only if raw string ≠ serialized casted string.
4. **Custom `CastsAttributes`** — if casted form is an object, `is_scalar(casted)` is false → keep casted (same idea as the synth’s dehydrate path for class casts).
5. **Any rule that expects the post-cast representation** (e.g. already-rounded decimal) — may now see pre-cast input instead.

Fixes: https://github.com/livewire/livewire/discussions/3400

### DIFF
--- a/src/Features/SupportLegacyModels/EloquentModelSynth.php
+++ b/src/Features/SupportLegacyModels/EloquentModelSynth.php
@@ -7,6 +7,7 @@ use Illuminate\Database\ClassMorphViolationException;
 use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Livewire\Drawer\Utils;
 
 class EloquentModelSynth extends Synth
 {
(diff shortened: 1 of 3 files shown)


calebporzio:
thanks!

PR #10561, opened 2026-08-20
Fix wire action collisions with Alpine scope variables

### LINKED ISSUE #10558: wire:click / wire:submit actions incorrectly skipped when name collides with nested Alpine x-data property
### Livewire version

v4.4.1

### Laravel version

v13.24.0

### PHP version

8.4.14

Brave on Windows

## Context

I originally reported this in #10121 ("Wire:submit not executed when x-data contain same name property / function"). In that discussion, a fix was suggested (#10137  "Fix Alpine scope leaking across Livewire component boundary"), but unfortunately that didn't solve the issue on my end.

I'm on 4.4.1, which already includes that fix, and the problem is still present. I believe that fix addressed a different, narrower case (Alpine scope leaking *above* the `wire:id` component boundary), not the general name collision case described below.

I did some digging and confirmed this worked correctly on v4.1.4 (before #9974) and started breaking with v4.2.0.

## Root cause

Looking at the source ([js/evaluator.js](https://github.com/livewire/livewire/blob/v4.4.1/js/evaluator.js)), `contextualizeExpression` collects every key from every nested `_x_dataStack` between the element and the `wire:id` root (via `getAlpineScopeKeys`), then skips prefixing with `$wire.` for any expression matching one of those keys, regardless of how deeply nested the Alpine scope is or how far it sits from the element.

In v3, the behavior was to always prefix with `$wire.`.
In v4, some JavaScript expressions can be accepted directly. Since #9974, it also first checks whether a property with the same name exists in any Alpine scope between the element and the Livewire component, and skips the `$wire.` prefix if so.

## Reproduction

I have anonymous Blade components that are reusable and have their own Alpine scope, sometimes with slots (modals, inputs, or reusable pieces, etc.). Sometimes an Alpine property in one of these nested scopes shares its name with an action or property of the Livewire component. Given that there can be multiple nested Alpine scopes, this can easily happen with generic names used to keep the architecture easy to understand.

Here's a minimal live reproduction : https://wirebox.app/b/4n26o

#### Steps to reproduce the issue
1. Open the live reproduction
2. Click "Toggle menu" to reveal the task list
3. Click "Open" next to any task
4. Observe: nothing happens, the server-side open($id) action is never called (no dd() output)

```php
<?php
namespace App\Livewire;

use Livewire\Component;
use Livewire\Attributes\Title;

#[Title('Task menu')]
class Menu extends Component
{
    /**
     * @var array{id: int, title: string}
     */
    public array $tasks = [
        ['id' => 1, 'title' => 'task 1'],
        ['id' => 2, 'title' => 'task 2'],
    ];

    public function open(int $id)
    {
        dd("Open {$id}");
    }

    public function render()
    {
        return view('livewire.menu');
    }
}
```

```blade
<div class="flex w-full bg-zinc-900 h-full items-center justify-center text-gray-300 px-16">
    <flux:card class="space-y-3">

        <div>
            <flux:heading size="lg">Menu</flux:heading>
        </div>

        <div>
            <div x-data="{ open: false }">
                <button
                    class="rounded-md bg-slate-800 py-2 px-4 border border-transparent text-center text-sm text-white transition-all shadow-md hover:shadow-lg focus:bg-slate-700 focus:shadow-none active:bg-slate-700 hover:bg-slate-700 active:shadow-none disabled:pointer-events-none disabled:opacity-50 disabled:shadow-none ml-2"
                    x-on:click="open = !open"
                >
                    Toggle menu
                </button>
                <div x-show="open" class="mt-2">
                    <ul class="flex flex-col gap-y-2">
                        @foreach ($tasks as $task)
                            <li class="flex">
                                {{ $task['title'] }}
                                <button
                                    class="ms-auto rounded-md bg-slate-800 py-2 px-4 border border-transparent text-center text-sm text-white transition-all shadow-md hover:shadow-lg focus:bg-slate-700 focus:shadow-none active:bg-slate-700 hover:bg-slate-700 active:shadow-none disabled:pointer-events-none disabled:opacity-50 disabled:shadow-none"
                                    wire:click="open({{ $task['id'] }})"
                                >
                                    Open
                                </button>
                            </li>
                        @endforeach
                    </ul>
                </div>
            </div>
        </div>
    </flux:card>
</div>
```

//

`wire:click="action"` should deterministically call the server-side action named "action", regardless of any Alpine scope in between. This was the behavior in v3 and up to v4.1.4.

For maintenance and code review purposes, this determinism matters: anyone reading `wire:click="action"` should be able to tell at a glance that it calls a publicly exposed server-side method, without needing to trace every Alpine scope between the element and the component root.

In my opinion, it's worth keeping a clear boundary between server actions (via `wire:click`) and JavaScript actions: `x-on:click` can already do both, JavaScript and explicit calls to the Livewire component via `$wire`

Of course, I'm using the click event listener as an example here, but this applies to anything relying on the evaluator.

calebporzio:
Fixes #10558.

## Problem

When a `wire:*` action name collided with a nested Alpine scope variable, the contextual evaluator left the action Alpine-scoped. For example, `wire:click="open(1)"` inside `x-data="{ open: false }"` attempted to call the boolean instead of the Livewire component method.

## Fix

Prefer the leading target of non-reactive `wire:*` action expressions over colliding Alpine scope keys, while preserving Alpine scope resolution for action parameters and reactive directives such as `wire:text` and `wire:show`. The target detection also handles bare statements, semicolons, and comments.

## Verification

- `npm run test:run` (110 passed, 1 skipped)
- `vendor/bin/phpunit --configuration phpunit.xml.dist src/Features/SupportJsEvaluation/BrowserTest.php` (20 passed, 29 assertions)

### DIFF
--- a/js/evaluator.js
+++ b/js/evaluator.js
@@ -65,7 +65,7 @@ export function evaluateReactiveExpression(el, expression, options = {}) {
 export function evaluateActionExpression(el, expression, options = {}) {
     if (! expression || expression.trim() === '') return
 
-    let contextualExpression = contextualizeExpression(expression, el)
+    let contextualExpression = contextualizeExpression(expression, el, ! isEvaluatingReactiveExpression())
 
     try {
         let result = Alpine.evaluateRaw(el, contextualExpression, options)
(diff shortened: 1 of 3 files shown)


MaxencePaulin:
Thanks, you were fast!

PR #10563, opened 2026-08-20
[4.x] Fix browser events dispatched by an update firing before the morph inserts their target

### LINKED ISSUE #10560: v4.4.1 regression: browser events dispatched by an update fire before the morph inserts their target (Flux modals never open)
### Livewire version

v4.4.1

### Laravel version

v13.26.1

### PHP version

8.4.23

Chrome/Chromium (Playwright) on macOS 15

Upgrading from v4.4.0 to v4.4.1 broke every Flux modal that is **rendered and shown by the same Livewire update**.

When one action both causes the `<flux:modal>` to be rendered for the first time (a plain `@if` in the view flipping from false to true) and calls `Flux::modal(...)->show()` in the same round-trip, the `modal-show` browser event is dispatched **before** the morph has inserted the `<dialog>` into the DOM. The event fires into the void:

- After the update, the `<dialog>` **is present** in the DOM (`document.querySelector('dialog') !== null`),
- but it never receives the `open` attribute (`document.querySelector('dialog[open]') === null`) — the modal never appears.
- Manually re-dispatching the same `modal-show` event from the console after the update opens the modal fine, which shows the event/listener wiring itself is intact and only the ordering is wrong.

Bisected against our browser test suite: **v4.4.0 passes, v4.4.1 fails**, with no other dependency changed. Modals that are already present in the DOM before the action (not newly rendered by the same update) still open fine.

Looking at the v4.4.0...v4.4.1 diff, plausible candidates are the island-morph batching/ordering change (785465b7, #10529) and the serialization of `invokeOnMorph` in `js/request/message.js`, but I haven't bisected to a single commit.

**Minimal reproduction repo (one component, one Pest browser test that passes on 4.4.0 and fails on 4.4.1):** https://github.com/jtrod/livewire-4.4.1-modal-show-repro

Current workaround for anyone hitting this:

```json
"conflict": {
    "livewire/livewire": "4.4.1"
}
```

Component:

```php
use Flux\Flux;
use Livewire\Component;

class ModalDemo extends Component
{
    public ?string $editing = null;

    public function edit(): void
    {
        $this->editing = 'demo-record';

        Flux::modal('demo')->show();
    }

    public function render()
    {
        return view('livewire.modal-demo');
    }
}
```

View:

```blade
<div>
    <flux:button wire:click="edit" data-test="open">Edit</flux:button>

    @if ($editing !== null)
        <flux:modal name="demo" data-test="modal">
            <flux:heading size="lg">The modal opened</flux:heading>
        </flux:modal>
    @endif
</div>
```

Steps:

1. Visit the page and click **Edit**.
2. On v4.4.1: nothing visible happens. In the console, `document.querySelector('dialog')` returns the dialog, but `document.querySelector('dialog[open]')` returns `null`.
3. `composer require livewire/livewire:4.4.0` and repeat: the modal opens.

Pest browser test from the repro repo:

```php
it('opens a flux modal rendered and shown in the same livewire update', function () {
    $page = visit('/');

    $page->assertScript('document.querySelector("dialog") === null', true)
        ->click('@open')
        ->wait(2)
        ->assertScript('document.querySelector("dialog") !== null', true)
        ->assertScript('document.querySelector("dialog[open]") !== null', true);
});
```

_No response_

Browser events queued by an update (such as Flux's `modal-show`) should be dispatched **after** the morph has applied the update's DOM changes, as in v4.4.0 — so an element rendered by the same update can receive them. The dialog should open.

jtrod:
Fixes #10560

## Root cause

#10529 (commit `785465b7`, shipped in v4.4.1) changed `Message.invokeOnMorph()` to a serial `for … await interceptor.onMorph()` loop. Every `interceptMessage()` registration contributes a default no-op `async () => {}` `onMorph`, and each serial await burns a microtask generation — so `supportDispatches`' triple-`queueMicrotask` deferral now flushes before `supportMorphDom`'s interceptor morphs the DOM. Events targeting elements rendered by the same update are silently lost: `->el()`/`->ref()` dispatches resolve their target via `querySelectorAll` at dispatch time, and even bubbling dispatches only reach listeners already bound. Under 4.4.0's `Promise.all`, the morph mutated the DOM synchronously before any microtask flushed.

Concrete symptom: a Flux modal rendered and shown in one round-trip never opens.

## The fix

In `js/features/supportDispatches.js` only — message-driven effects now dispatch from the message's own `onMorph` hook (post-morph, pre-paint; `supportDispatches` is imported after `supportIslands` and `supportMorphDom`, and `invokeOnMorph()` awaits in registration order, so it runs after the DOM changes apply). The triple-queueMicrotask path is kept for request-less effects (initial page load via `store.js`), guarded by `if (request) return`. This restores the 4.4.0 contract — existing tests from #9933 already depend on post-morph dispatch.

## Tests

Three browser tests in `src/Features/SupportEvents/BrowserTest.php`:

1. Window listener on an element rendered by the same update that dispatches.
2. `->el('#target')` targeting an element rendered by the same update (the exact Flux mechanism — non-bubbling).
3. Mount-time dispatch on initial page load (covers the request-less fallback path).

Tests 1–2 fail on current `4.x` HEAD via wait timeout, pass with the fix. Full `SupportEvents` and `SupportIslands` browser suites and the vitest suite pass (`SupportIslands` protects #10529's island behavior). Also verified end-to-end against the repro app from #10560 (passes patched, fails on v4.4.1).

## Behavior notes

Deliberate edge-behavior changes, worth reviewer eyes:

- A message cancelled between effect-processing and the morph no longer fires its dispatches.
- If an earlier interceptor's `onMorph` throws, dispatches are dropped along with the rest of the morph pipeline — and conversely a throw during dispatch now propagates out of `invokeOnMorph()` (skipping `invokeOnRender`), though DOM listener exceptions don't propagate out of `dispatchEvent()` so userland handlers can't trigger that.
- Global `interceptMessage` `onMorph` callbacks registered by userland now run after dispatches — component-scoped interceptors (`$wire.$intercept`) still run before, since the registry orders component-scoped interceptors first.

## Separate pre-existing observation (not addressed here)

`supportSlots.js` uses a fire-and-forget `fragments.forEach(async …)` in its `onMorph`, so it resolves before its slot morphs complete; additionally `supportSlots` is imported after `supportDispatches`, so events targeting elements inside a freshly rendered slot fragment remain subject to the pre-existing race. Happy to file a separate issue.

## Note

`dist/` intentionally not included, per contributor convention.

### DIFF
--- a/js/features/supportDispatches.js
+++ b/js/features/supportDispatches.js
@@ -1,26 +1,57 @@
 import { dispatch, dispatchEl, dispatchRef, dispatchSelf, dispatchTo } from '@/events'
 import { on } from '@/hooks'
+import { interceptMessage } from '@/request'
 
-on('effect', ({ component, effects }) => {
-    // Wrapping this in a triple queueMicrotask...
-    // The first one puts it after all other "effect" hooks...
-    // The second one puts it after all reactive Alpine effects
-    // (that are processed via flushJobs in scheduler)...
-    // The third one puts it after morph changes have been applied...
+// Dispatches coming back from a server round-trip are anchored to the message's own
+// post-morph hook instead of a detached queueMicrotask chain. `Message.invokeOnMorph()`
+// awaits each interceptor's `onMorph()` serially, in registration order, and this file
+// is imported AFTER `supportIslands` and `supportMorphDom` in `features/index.js`, so by
+// the time this callback runs the new HTML has already been morphed into the DOM —
+// post-morph, but still pre-paint.
+//
+// The ordering is load-bearing for two reasons: element-targeted dispatches (`->el()` /
+// `->ref()`) resolve their target by querying the component's element at dispatch time,
+// and even a bubbling `dispatch()` only reaches listeners that are already bound — both
+// fail silently for an element this same update rendered if the event fires before the
+// morph.
+//
+// Note the converse: `supportSlots` is imported later still, so its slot-fragment morphs
+// have not yet run when this fires — events targeting elements inside a freshly rendered
+// slot are still subject to the original race.
+interceptMessage(({ message, onSuccess }) => {
+    onSuccess(({ payload, onMorph }) => {
+        onMorph(() => {
+            dispatchEvents(message.component, getDispatches(payload.effects))
+        })
+    })
+})
+
+// Effects that don't originate from a message never reach the interceptor above — the
+// only such case is the initial page load, where `store.js` calls `processEffects()`
+// without a request. Those mount-time dispatches still go out from here.
+//
+// Wrapping this in a triple queueMicrotask...
+// The first one puts it after all other "effect" hooks...
+// The second one puts it after all reactive Alpine effects
+// (that are processed via flushJobs in scheduler)...
+// The third one puts it after morph changes have been applied...
+on('effect', ({ component, effects, request }) => {
+    if (request) return
+
     queueMicrotask(() => {
         queueMicrotask(() => {
             queueMicrotask(() => {
-                let dispatches = []
-
-                if (Object.prototype.hasOwnProperty.call(effects, 'dispatches') && effects.dispatches) {
-                    dispatches = effects.dispatches
-                }
-
-                dispatchEvents(component, dispatches)
+                dispatchEvents(component, getDispatches(effects))
             })
         })
     })
 })
+
+function getDispatches(effects) {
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for the contribution! I uncovered other missing pieces and went with a different approach, so closing this PR in favor of this other one: #10568  THANKs!!

PR #10566, opened 2026-08-21
[4.x] Improve Eloquent models restoration on subsequent request

ghabriel25:
> [!NOTE]
> I notice this behavior when creating comment feature for my app where it has load comment button to refresh the component which lead to duplicate queries

```
// $post->comments
User A comment
  - User B comment
  - User A comment
  - User B comment
  - User A comment
```

This will create duplicate queries for the same user model even with eager loading enabled. Because each model hydration will create another query. 

Imagine how many queries it will created for **Like a comment** feature where it only has a list of user who likes the comment and each of that user appear on the same page of comment section page

## The Scenario

Livewire restores Eloquent models from snapshot metadata (`class` + `key`) on every request via `ModelSynth` and `EloquentCollectionSynth`. Each property is hydrated on its own. The only reuse today is for models already resolved by route binding (`PersistentMiddleware::$resolvedRouteModels`). https://github.com/livewire/livewire/pull/9935

Minimal snippet to reproduce the issue
```php
<?php

namespace App\Livewire;

use App\Models\User;
use Livewire\Component;

class UserManager extends Component
{
    public User $selected;
    
    #[Computed]
    public function users()
    {
        return User::query()->limit(3)->get();
    }

    public function mount()
    {
        $this->selected = $this->users->first();
    }

    public function select(int $id)
    {
        $this->selected = $this->users->firstWhere('id', $id)
            ?? User::findOrFail($id);
    }

    public function render()
    {
        return view('livewire.user-manager');
    }
}
```

```blade
<div>
    <flux:main>
        <flux:heading size="xl">Users</flux:heading>

        <ul class="space-y-6">
            @foreach ($this->users as $user)
                <li class="space-x-4" wire:key="{{ $user->id }}">
                    <span>{{ $user->name }}</span>
                    <flux:button wire:click="select({{ $user->id }})">
                        Select
                    </flux:button>
                </li>
            @endforeach
        </ul>

        <hr class="my-4">

        <div class="space-y-4">
            <p>Selected: {{ $selected->name }} (ID: {{ $selected->id }})</p>
            <flux:button wire:click="$refresh">Refresh</flux:button>
        </div>
    </flux:main>
</div>
```
> [!NOTE]
> The output below still same regardless using `#[Computed]` attribute or using `Collection` as property
> Tested with php version 8.3, 8.4, and 8.5

1. Click `Refresh`, notice duplicate queries
    - `select * from "users" where "users"."id" = 1 limit 1`
    - `select * from "users" where "users"."id" in (1, 2, 3)`
2. Select a user
    - `select * from "users" where "users"."id" in (1, 2, 3)`
3. Click `Refresh` again, same result with (1)

## The Problem

The same database row is often represented more than once in a single request graph, for example:

* A parent holds `Collection $posts` and a child (or the parent itself) also holds `Post $selected` that is one of those posts
* Nested / reactive children each receive the same `User`, `Order`, or `Team` the parent already loaded
* Several components on one page hydrate the same model after a shared action or island update
* A full-page component uses a route-bound model while child components also declare that model as a public property

Each of those paths can call `newQueryForRestoration()` again for an id that was already loaded earlier in the same request. Under PHP 8.4+ the query is deferred until the proxy is touched, but it still runs once per property, not once per unique `class:key`.

## The Solution

> [!IMPORTANT]
> There are two different cases here:
> **Selected rendered first → collection second:**
> 1. Selected materializes
>     - cache miss
>     - `select * from "users" where "users"."id" = 1 limit 1`
>     - remembers the model
> 2. Collection materializes
>     - cache hit only for rendered selected
>     - `select * from "users" where "users"."id" in (2, 3)` (selected model **excluded**)
>     - remembers remaining models
>
>  **Collection rendered first → selected second**
> 1. Collection materializes
>     - cache miss
>     - `select * from "users" where "users"."id" in (1, 2, 3)`
>     - remembers all models
> 2. Selected materializes
>     - hit cache
>     - no additional query (selected model already **included** in collection query)

Extend the existing request-scoped map on `PersistentMiddleware` so **any** materialized model can be reused for the rest of the request—not only route bindings:

1. **`rememberResolvedModel(Model $model)`** — store original model under `class:key` when a model is loaded. The first materialization of a given identity wins for the rest of the request.
5. **`ModelSynth`** — consult the cache before querying, and again inside the lazy-proxy callback (so a collection that materializes later can still satisfy a sibling model property).
6. **`EloquentCollectionSynth`** — after a bulk restore, remember each model so later `ModelSynth` hydrates can skip their own `SELECT`.

No change to snapshot shape or checksum behaviour. The map is still cleared on `flush-state`.

## Tests

To be honest, the tests was written by AI as I'm not sure how to test this. I only test this on fresh Laravel + Livewire apps.

* Two hydrates of the same `class:key` issue one query
* Models remembered from an `EloquentCollectionSynth` bulk load are reused by a following `ModelSynth` hydrate with no extra query
* `Collection` + selected model of the same class share one query on `$refresh` (end-to-end)

### DIFF
--- a/src/Features/SupportModels/EloquentCollectionSynth.php
+++ b/src/Features/SupportModels/EloquentCollectionSynth.php
@@ -77,21 +77,37 @@ class EloquentCollectionSynth extends Synth {
         }
 
         return $this->makeLazyProxy($class, $meta, function () use ($modelClass, $keys, $meta) {
-            // We are using Laravel's method here for restoring the collection, which ensures
-            // that all models in the collection are restored in one query, preventing n+1
-            // issues and also only restores models that exist.
-            $collection = (new $modelClass)->newQueryForRestoration($keys)->useWritePdo()->get();
+            $mechanism = app(PersistentMiddleware::class);
+
+            // Reuse models already loaded in this request (route binding or ModelSynth).
+            $missing = [];
+            foreach ($keys as $key) {
+                if (! $mechanism->getResolvedRouteModel($modelClass, $key)) {
+                    $missing[] = $key;
+                }
+            }
+
+            $collection = collect();
+
+            if (count($missing) > 0) {
+                // We are using Laravel's method here for restoring the collection, which ensures
+                // that all models in the collection are restored in one query, preventing n+1
+                // issues and also only restores models that exist.
+                $collection = (new $modelClass)->newQueryForRestoration($missing)->useWritePdo()->get();
+
+                // Cache every model so individual ModelSynth hydrates can reuse them.
+                foreach ($collection as $model) {
+                    $mechanism->rememberResolvedModel($model);
+                }
+            }
 
             $collection = $collection->keyBy->getKey();
 
-            // Cache every model so individual ModelSynth hydrates can reuse them.
-            foreach ($collection as $model) {
-                app(PersistentMiddleware::class)->rememberResolvedModel($model);
-            }
-
             return new $meta['class'](
-                collect($meta['keys'])->map(function ($id) use ($collection) {
-                    return $collection[$id] ?? null;
+                collect($keys)->map(function ($id) use ($mechanism, $modelClass, $collection) {
+                    return $mechanism->getResolvedRouteModel($modelClass, $id)
+                        ?? $collection[$id]
+                        ?? null;
                 })->filter()
             );
         });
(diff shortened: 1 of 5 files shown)


calebporzio:
going with #10633 instead that came out of this. thanks!

PR #10568, opened 2026-08-21
[4.x] Close asynchronous morph timing gaps

### LINKED ISSUE #10560: v4.4.1 regression: browser events dispatched by an update fire before the morph inserts their target (Flux modals never open)

calebporzio:
Fixes #10560

## What

- Wait for every slot fragment morph before the message morph phase continues.
- Add an explicit message-level `onMorphed` phase after all awaited morph work and before action promises resolve.
- Run message-driven browser events and server-dispatched JavaScript from `onMorphed` instead of detached timing guesses.
- Preserve the established pre-morph `@script` phase so scripts can register Alpine data, directives, and other behavior before newly rendered elements initialize.
- Add a `component.initialized` boundary after initial effects are processed so mount-time JavaScript and dispatches preserve their startup ordering.
- Await asynchronous `onStream` handlers so streamed island morphs complete before action promises resolve and before the next chunk is processed.
- Document the simple `onMorphed` hook, advanced `onMorph` hook, startup boundary, server JavaScript timing, and stream awaiting behavior.

## Timing boundaries

- `@assets` remain awaited during payload interception before component response handling.
- `@script` remains pre-morph because it is commonly used to register Alpine behavior consumed by the incoming DOM.
- `$this->js()` / `xjs` run post-morph so server-dispatched JavaScript can safely access DOM rendered by the same response.
- Script-defined `$js` actions are still registered before `xjs` invokes them.

## Why

The nested `queueMicrotask()` chain only guessed when morphing would be complete. Once message morph interceptors became serially awaited—and especially when slot fragments or streamed islands contributed their own asynchronous morphs—DOM-dependent effects could run before newly rendered targets and listeners existed, or an action promise could resolve before the UI finished updating.

The lifecycle boundaries now make that ordering explicit without changing the historical initialization contract of `@script`.

## Tests

The following behavior-observing browser tests were added and verified failing before their corresponding production patches, then passing afterward:

- Server-dispatched JavaScript can access DOM rendered by the same response.
- A conditional `@script` can register Alpine data before newly transitioned DOM initializes.
- An awaited action does not resolve until a streamed, transitioned island has finished morphing.
- A direct browser event reaches a target rendered by the same response.
- An awaited action does not resolve until a transitioned slot displays its server-rendered content.

Additional verification after the timing-history correction:

- Combined Scripts and Assets, JS Evaluation, and Events browser suites: 55 tests, 93 assertions.
- JavaScript suite: 110 passed, 1 pre-existing skip.
- `npm run build`.


js/features/supportDispatches.js
```diff
@@ -1,7 +1,18 @@
...
+
+interceptMessage(({ message, onSuccess }) => {
+    onSuccess(({ payload, onMorphed }) => {
+        onMorphed(() => {
+            dispatchEvents(message.component, getDispatches(payload.effects))
+        })
+    })
+})
+
+on('effect', ({ component, effects, request }) => {
```
calebporzio: interesting...this is i suppose because effects can sometimes run without being part of a request lifecycle? like in the case of a component intializing and distpaching from mount and embedding effects directly that run during initailization?

hmmm... this feels weird to have to check both. it also feels weird to have an if (request) return early return basedon the existance of request or not...

js/features/supportDispatches.js
```diff
@@ -1,7 +1,18 @@
 import { dispatch, dispatchEl, dispatchRef, dispatchSelf, dispatchTo } from '@/events'
 import { on } from '@/hooks'
+import { interceptMessage } from '@/request'
+
+interceptMessage(({ message, onSuccess }) => {
+    onSuccess(({ payload, onMorphed }) => {
+        onMorphed(() => {
+            dispatchEvents(message.component, getDispatches(payload.effects))
```
calebporzio, line 8: but this is so clean and nice. so this is a good direction we're heading i think.

js/features/supportDispatches.js
```diff
@@ -10,18 +21,18 @@ on('effect', ({ component, effects }) => {
...
-
-                dispatchEvents(component, dispatches)
+                dispatchEvents(component, getDispatches(effects))
             })
         })
     })
 })
 
+function getDispatches(effects) {
+    if (! Object.prototype.hasOwnProperty.call(effects, 'dispatches')) return []
```
calebporzio: why hasOwnProperty.call? would effects.dispatches || [] or whatever not suffice alone? curious why this is here.

js/features/supportDispatches.js
calebporzio: Yep — the request-less case is initial component effects such as a dispatch from `mount()`. I moved that path to `component.init`, so response dispatches are explicitly message-scoped through `onMorphed` and initial dispatches are explicitly initialization-scoped. The `if (request)` provenance check is gone, and the mount-time browser test covers the initial path.

js/features/supportDispatches.js
calebporzio: Agreed — the ownership check was inherited from the previous implementation and adds nothing here. `getDispatches()` now just returns `effects.dispatches || []`.

PR #10571, opened 2026-08-21
[3.x] Ignore a repeat `__lazyLoad` call instead of throwing

### LINKED ISSUE #10570: Repeat `__lazyLoad` call throws MethodNotFoundException since v3.8.5
### Livewire version
v3.8.5 (regression introduced in this release; v3.8.4 and earlier are unaffected)

### Summary

Since v3.8.5, a repeat `__lazyLoad` call throws `MethodNotFoundException`, returning a 500 from
`/livewire/update` and leaving the lazy component stuck on its placeholder. The duplicate call is
generated by Livewire's own placeholder and commit queue, so no application code is required to
trigger it — a user scrolling a lazy component in and out of view is enough.

```
Livewire\Exceptions\MethodNotFoundException:
Unable to call component method. Public method [__lazyLoad] not found on component
at src/Mechanisms/HandleComponents/HandleComponents.php:517
```

### Reproduction

Two `__lazyLoad` calls on the same component, which is exactly what the browser sends (see below).
Added to `src/Features/SupportLazyLoading/UnitTest.php`, alongside the existing
`test_a_lazy_component_loads_with_its_own_mount_params`:

```php
public function test_a_repeat_lazy_load_call_is_ignored_rather_than_throwing()
{
    SupportLazyLoading::$disableWhileTesting = false;

    Livewire::component('lazy-alpha', LazyAlpha::class);

    $html = html_entity_decode(Livewire::mount('lazy-alpha', ['level' => 5]));
    preg_match("/__lazyLoad\('([^']+)'\)/", $html, $matches);

    $this->assertNotEmpty($matches[1] ?? null);

    Livewire::test('lazy-alpha')
        ->call('__lazyLoad', $matches[1])
        ->call('__lazyLoad', $matches[1])   // throws on v3.8.5, passes on v3.8.4
        ->assertOk()
        ->assertSee('level:5');
}
```

### Why the browser sends a second `__lazyLoad`

1. `SupportLazyLoading::generatePlaceholderHtml()` stamps `x-intersect="$wire.__lazyLoad('<encoded>')"`
   on the placeholder root. There is no `.once` modifier, so the observer fires again every time the
   element re-enters the viewport — trivially reproducible with momentum scrolling on mobile.
2. If it fires while the first resume is still in flight, `CommitBus.corraleCommitsIntoPools()` skips
   the second commit (`findPoolWithComponent()` hits) and leaves it queued; `sendAnyQueuedCommits()`
   flushes it once the first response resolves.
3. `Commit.toRequestPayload()` reads `this.component.snapshotEncoded` **at send time**, which by then
   is the post-load snapshot. So the duplicate arrives with `memo.lazyLoaded: true`.

A production payload showing exactly this — a valid `__mountParamsContainer` param paired with an
already-resumed snapshot:

```json
{
  "calls": [{ "method": "__lazyLoad", "params": ["eyJkYXRhIjp7ImZvck1vdW50Ijp..."] }],
  "snapshot": { "memo": { "name": "widgets.company.net-income", "lazyLoaded": true, ... } }
}
```

### Why it throws on 3.8.5 but not 3.8.4

`19a4a028` ("Tidy up lazy loading internals", backported to 3.x in #10545) added a guard to
`SupportLazyLoading::call()`:

```php
 function call($method, $params, $returnEarly)
 {
     if ($method !== '__lazyLoad') return;

+    // Only applies while a lazy load is being resumed.
+    if ($this->storeGet('isLazyLoadHydrating') !== true) return;
```

`isLazyLoadHydrating` is only set by `hydrate()`, which returns early when `$memo['lazyLoaded'] === true`.
So for an already-resumed component the hook declines to intercept, and `__lazyLoad` falls through to
`HandleComponents::callMethods()`, which throws.

Before that commit, `call()` intercepted unconditionally: the duplicate re-ran `resurrectMountParams()`
(the `forComponent` check passes, it is the same component) and `callMountLifecycleMethod()`, then
returned early. A redundant re-mount, but no exception.

I checked every v3 tag: the guard is absent in v3.0.0 through v3.8.4 and present in v3.8.5. It is also
present on 4.x, so this affects that branch too.

### Suggested fix

Remember that the component arrived already-resumed, and swallow the repeat call rather than letting it
reach method resolution:

```php
 public function hydrate($memo)
 {
     if (! isset($memo['lazyLoaded'])) return;
-    if ($memo['lazyLoaded'] === true) return;
+
+    if ($memo['lazyLoaded'] === true) {
+        $this->storeSet('hasAlreadyLazyLoaded', true);
+
+        return;
+    }
```

```php
     // Only applies while a lazy load is being resumed.
-    if ($this->storeGet('isLazyLoadHydrating') !== true) return;
+    if ($this->storeGet('isLazyLoadHydrating') !== true) {
+        if ($this->storeGet('hasAlreadyLazyLoaded') === true) $returnEarly();
+
+        return;
+    }
```

The component still re-renders and returns HTML, so the duplicate request simply re-morphs the
already-correct output.

Adding `.once` to the placeholder's `x-intersect` would remove the duplicate request altogether, but it
would also remove the retry when a lazy load genuinely fails on a flaky connection, so it seems like the
wrong trade on its own.

Opening a PR with this fix and the test shortly.

robert-stanciu:
Fixes #10570.

## The problem

Since v3.8.5, a repeat `__lazyLoad` call throws `MethodNotFoundException` — a 500 from `/livewire/update`, leaving the lazy component stuck on its placeholder. No application code is needed to trigger it: the duplicate is produced by Livewire's own placeholder and commit queue when a user scrolls a lazy component in and out of view.

1. `generatePlaceholderHtml()` stamps `x-intersect="$wire.__lazyLoad('<encoded>')"` on the placeholder root, with no `.once` modifier, so it re-fires on every re-entry into the viewport.
2. Fire it while the first resume is in flight and `CommitBus.corraleCommitsIntoPools()` skips the second commit (`findPoolWithComponent()` hits), leaving it queued for `sendAnyQueuedCommits()`.
3. `Commit.toRequestPayload()` reads `snapshotEncoded` at send time, so the queued duplicate goes out carrying the post-load snapshot (`memo.lazyLoaded: true`).
4. `hydrate()` returns early on that memo, so `isLazyLoadHydrating` is never set, so the guard added in 19a4a028 makes `call()` decline to intercept — and `__lazyLoad` reaches `HandleComponents::callMethods()`, which throws.

Before 19a4a028 the same duplicate was absorbed: `call()` intercepted unconditionally, re-ran `resurrectMountParams()` (the `forComponent` check passes, it is the same component) and `callMountLifecycleMethod()`, then returned early. I checked each v3 tag — the guard is absent in v3.0.0 through v3.8.4 and present from v3.8.5.

## The fix

Record that a component arrived already-resumed, and swallow the repeat call instead of letting it reach method resolution. A genuine first-time resume is unaffected, and since only method dispatch is skipped the component still re-renders, so the duplicate request just re-morphs the already-correct output.

I kept it to the narrow case rather than making `call()` swallow every `__lazyLoad` it does not handle, so a `__lazyLoad` on a component that was never lazy still surfaces as an error.

## Tests

`test_a_repeat_lazy_load_call_is_ignored_rather_than_throwing` in the existing `SupportLazyLoading\UnitTest`, following the `Livewire::mount()` + `preg_match` pattern the neighbouring mount-params tests already use. It reproduces the production error without the fix and passes with it.

`--testsuite Unit --filter SupportLazyLoading` is green (5 tests, 12 assertions).

## Note on `.once`

Adding `.once` to the placeholder's `x-intersect` would stop the duplicate request being sent at all, but it would also remove the retry when a lazy load genuinely fails on a flaky connection — so it looked like the wrong trade to make on its own. Happy to add it as well if you would rather have both.

### DIFF
--- a/src/Features/SupportLazyLoading/SupportLazyLoading.php
+++ b/src/Features/SupportLazyLoading/SupportLazyLoading.php
@@ -76,7 +76,12 @@ class SupportLazyLoading extends ComponentHook
     public function hydrate($memo)
     {
         if (! isset($memo['lazyLoaded'])) return;
-        if ($memo['lazyLoaded'] === true) return;
+
+        if ($memo['lazyLoaded'] === true) {
+            $this->storeSet('hasAlreadyLazyLoaded', true);
+
+            return;
+        }
 
         $this->component->skipHydrate();
 
@@ -99,7 +104,16 @@ class SupportLazyLoading extends ComponentHook
         if ($method !== '__lazyLoad') return;
 
         // Only applies while a lazy load is being resumed.
-        if ($this->storeGet('isLazyLoadHydrating') !== true) return;
+        if ($this->storeGet('isLazyLoadHydrating') !== true) {
+            // The placeholder's `x-intersect` has no `.once` modifier, so it can fire again
+            // while the first resume is still in flight. That duplicate is queued client-side
+            // and sent afterwards carrying the post-load snapshot, so it lands here on an
+            // already-resumed component. Swallow it rather than letting it fall through to
+            // method resolution, which would throw MethodNotFoundException.
+            if ($this->storeGet('hasAlreadyLazyLoaded') === true) $returnEarly();
+
+            return;
+        }
 
         [ $encoded ] = $params;
 
--- a/src/Features/SupportLazyLoading/UnitTest.php
+++ b/src/Features/SupportLazyLoading/UnitTest.php
@@ -61,6 +61,26 @@ class UnitTest extends \Tests\TestCase
             ->assertSee('level:5');
     }
 
+    public function test_a_repeat_lazy_load_call_is_ignored_rather_than_throwing()
+    {
+        SupportLazyLoading::$disableWhileTesting = false;
+
+        Livewire::component('lazy-alpha', LazyAlpha::class);
+
+        $html = html_entity_decode(Livewire::mount('lazy-alpha', ['level' => 5]));
+        preg_match("/__lazyLoad\('([^']+)'\)/", $html, $matches);
+
+        $this->assertNotEmpty($matches[1] ?? null);
+
+        // The placeholder's `x-intersect` has no `.once` modifier, so the browser can queue a
+        // second `__lazyLoad` behind the first and send it with the post-load snapshot.
+        Livewire::test('lazy-alpha')
+            ->call('__lazyLoad', $matches[1])
+            ->call('__lazyLoad', $matches[1])
+            ->assertOk()
+            ->assertSee('level:5');
+    }
+
     public function test_a_mount_params_container_is_scoped_to_its_own_component()
     {
         SupportLazyLoading::$disableWhileTesting = false;


calebporzio:
Thanks for this — the writeup in #10570 is one of the best bug reports we've gotten. You traced it exactly right, down to `Commit.toRequestPayload()` reading `snapshotEncoded` at send time.

One thing to flag: we'd already landed a fix for this on 4.x in #10562 a day before you opened this, which is why 4.x users are seeing it too — it just hasn't been tagged yet. That version is slightly wider than yours. It also changes `dehydrate()` to carry `lazyLoaded: true` forward in the memo:

```php
} elseif (
    $this->storeGet('isLazyLoadHydrating') === true
    || $this->storeGet('isLazyLoadAlreadyLoaded') === true
) {
    $context->addMemo('lazyLoaded', true);
}
```

Without that half, the repeat call is swallowed but the response drops `lazyLoaded` from the memo — so a *third* `__lazyLoad` hydrates with no memo key, the flag is never set, and it throws again. I confirmed it against your branch's shape: it survives call two and throws on call three. Momentum scrolling can queue more than one duplicate behind the first, so it's reachable in the same scenario you described.

Since it's release day I've opened #10586 to backport the 4.x version verbatim so the branches stay in sync, rather than have you round-trip on it. Closing this in favor of that one — but the diagnosis was entirely yours and you'll be credited in the release notes.

Agreed on `.once`, by the way. Same conclusion here: it'd remove the retry on a flaky connection, so it's the wrong trade on its own.

PR #10572, opened 2026-08-21
fix route model binding issue that causes 404 on update route

jasperdevries:

I encountered this issue during work where we saw this happen on our website, i did not create a discussion first. 

yes i did create a branch.

It just this change, with test.

yes

What happened is that when having a livewire component on a 404 page and that component calling the livewire update route, we got another 404 page in a popup. After some digging we found that when livewire calls that route it also tries to build/fake the route that 404'd. This had been caught for routes that do not exist but not for routes that do exist but where the route model binding throws a 404.
This change fixes that by catching the `ModelNotFoundException` when trying to apply the middleware to the fake request in order to not recieve a 404 on the update route.

Thanks for contributing! 🙌

### DIFF
--- a/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
+++ b/src/Mechanisms/PersistentMiddleware/PersistentMiddleware.php
@@ -3,6 +3,7 @@
 namespace Livewire\Mechanisms\PersistentMiddleware;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Routing\Router;
 use Livewire\Mechanisms\Mechanism;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
@@ -118,7 +119,11 @@ class PersistentMiddleware extends Mechanism
         // Only send through pipeline if there are middleware found
         if (is_null($middleware) || $middleware === []) return;
 
-        Utils::applyMiddleware($request, $middleware);
+        try {
+            Utils::applyMiddleware($request, $middleware);
+        } catch (ModelNotFoundException $e) {
+            return;
+        }
 
         $this->middlewareAppliedFor[$routeKey] = true;
 
--- a/src/Mechanisms/PersistentMiddleware/UnitTest.php
+++ b/src/Mechanisms/PersistentMiddleware/UnitTest.php
@@ -6,6 +6,7 @@ use Illuminate\Http\Request;
 use Illuminate\Routing\RouteCollection;
 use Illuminate\Support\Facades\Facade;
 use Illuminate\Support\Facades\Route;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Livewire\Component as BaseComponent;
 use Livewire\Livewire;
 use Livewire\Mechanisms\HandleRequests\EndpointResolver;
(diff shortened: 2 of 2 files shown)


ghabriel25:
If application decide it should be 404 then swallowing the exception means it is treating failure of `SubstituteBindings` as a non-fatal condition and then treating the middleware application as effectively completed., am I right?

```
Livewire update request
        │
        ▼
PersistentMiddleware
        │
        ▼
makeFakeRequest()
        │
        ▼
find original route
        │
        ▼
gather persistent middleware
        │
        ▼
Utils::applyMiddleware($request, $middleware)
        │
        ▼
SubstituteBindings
        │
        ▼
Implicit route binding
        │
        ├── model exists
        │      ↓
        │   continue
        │
        └── model missing
               ↓
       ModelNotFoundException
               ↓
          PR catches it
               ↓
          continue Livewire

**Comment by jasperdevries** [comment]
yes, but at this point we're already on a 404 page and this fake request will only trigger another error on the update route when called on a 404 page. This is essentially the same fix that has been done for when the route cannot be found in the `getRouteFromRequest` method.

**Comment by ghabriel25** [comment]
Because its route level issue (route exists but model not found) I think its better to fix it on route level too. Something like

```php
{

    if (! $route) return [];

    $router = app('router');

    try {
        $router->substituteBindings($route);
        $router->substituteImplicitBindings($route);
    } catch (ModelNotFoundException $e) {
        return [];
    }

    $middleware = $router->gatherRouteMiddleware($route);

    return $this->filterMiddlewareByPersistentMiddleware($middleware);
}

**Comment by ghabriel25** [comment]
I have tried both your fix and my fix. All bypass authorization middleware which cause at least 2 to 3 browser tests fail.
- `test_explicit_route_model_binding_does_not_break_with_lazy_child`
- `test_that_authorization_middleware_is_re_applied`
- `test_that_authorization_middleware_is_re_applied_on_page_components`

I think we should wait @calebporzio or @joshhanley to solve this as it includes security related things.

I'll leave my scenario here as reference

1. Publish Laravel error pages
```bash
php artisan vendor:publish --tag=laravel-errors
```
2. Add Livewire component to `404.blade.php` page
3. Calling component action will trigger another 404 page but this time inside a modal

https://github.com/user-attachments/assets/4ff1f09a-6a2f-4d12-9d7b-10a7ea6f92c2

calebporzio:
Thanks for tracking this down and providing the reproduction.

We are going to leave interactive Livewire components on route-binding exception pages unsupported at the framework level. Swallowing ModelNotFoundException during persistent middleware replay can allow an action to proceed without middleware that comes afterward, including authorization. Precisely distinguishing an initial binding failure from a model deleted after a successful render requires carrying exception provenance through the snapshot lifecycle, which is too much machinery for this edge case.

Applications that accept the tradeoff can remove SubstituteBindings from Livewire persistent middleware replay in AppServiceProvider:

```php
use Illuminate\Routing\Middleware\SubstituteBindings;
use Livewire\Livewire;

public function boot(): void
{
    Livewire::setPersistentMiddleware(
        array_values(array_diff(
            Livewire::getPersistentMiddleware(),
            [SubstituteBindings::class],
        ))
    );
}
```

Initial page requests will still perform route binding, but Livewire updates will not re-run it. Authentication and other persistent middleware remain enabled.

The tradeoff is that Livewire updates will no longer revalidate scoped or custom route bindings. Do not use this workaround if route binding is part of tenant isolation or authorization, or if updates must detect that a bound model has been deleted.

PR #10578, opened 2026-08-23
[4.x] Fix form reset throws `TypeError` on non-nullable properties without defaults

ghabriel25:
## Summary
`Form::reset()` does not handle non-nullable typed properties that have no default. A fresh form instance leaves those properties uninitialized, `data_get()` returns `null`, and `data_set()` throws a `TypeError`.

`FormObjectSynth::set()` and `InteractsWithProperties::reset()` already handle this case by unsetting the property. `Form::reset()` did not.

## Scenario

```php
class PostForm extends Form
{
    public ?Post $post;

    public string $title;

    public int $views_count;

    public function setPost(Post $post)
    {
        $this->post = $post;

        $this->title = $post->title;

        $this->views_count = $post->views_count;
    }
}
```
```php
new class extends Component
{
    public PostForm $form;

    public function mount(Post $post)
    {
        $this->form->setPost($post);
    }

    public function resetForm()
    {
        $this->form->reset(); // TypeError: Cannot assign null to property ...::$title of type string
    }
}
```

## Problem

`Form::reset()` always did:
```php
data_set($this, $property, data_get($freshInstance, $property));
```

On a fresh instance, uninitialized typed properties yield `null` from `data_get()`, which cannot be assigned to `string` / `int`.

## Solution

- Check if property exists on Form object
- Reflect whether the property is initialized on the fresh instance
- If not, `data_forget()` instead of assigning `null`
- Otherwise, `data_set()` from the fresh instance as before

However, this still will throws exception if non-nullable typed properties accessed before assigning its value
```
Error: Typed property ...::$title must not be accessed before initialization
```

There is no way to mitigate this error so I add additional information to [Resetting form fields](https://livewire.laravel.com/docs/4.x/forms#resetting-form-fields)

```php
new class extends Component
{
    public Post $post;

    public PostForm $form;

    public function mount(Post $post)
    {
        $this->post = $post;

        $this->form->setPost($post);
    }

    public function resetForm()
    {
        $this->form->reset();

        $this->form->setPost($this->post);  // Assign after reset to prevent exception
    }
}
```

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -350,36 +350,57 @@ $this->reset(['title', 'content']);
 ```
 
 > [!important]
-> Form properties declared as non-nullable typed properties without defaults (e.g. `public string $title`) are left uninitialized after `reset()` (and when Livewire clears them via the form synthesizer). Accessing them in a view before assignment will throw. Use a default (`= ''`), a nullable type (`?string $title = null`), or null-safe access in the view.
+> Form properties declared as non-nullable typed properties without defaults (e.g. `public string $title`) are left uninitialized after `reset()` (and when Livewire clears them via the form synthesizer). Accessing them in a view before assignment will throw. Assign after reset, use a default value (`= ''`), a nullable type (`?string $title = null`), or null-safe access in the view (`$form?->title`).
 
 ```php
 <?php
 
 namespace App\Livewire\Forms;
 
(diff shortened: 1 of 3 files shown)


ghabriel25:
Closing in favor https://github.com/livewire/livewire/pull/10593

Issue #10588, opened 2026-08-24
v4.4.2 Exception Handling Changes Ignore $stopPropagation() Closure Invocation


rmdavi:
### Livewire version

v4.4.2

### Laravel version

v13.26.1

### PHP version

8.4.23

### Browser and operating system

Unit tests, Linux and MacOS

### Describe the issue you're experiencing

Since v4.4.2's changes to [exception handling in computed property](https://github.com/livewire/livewire/pull/10543), `$stopPropgation()` is not effective and return from the computed property continues unimpeded.

### Code snippets to reproduce the issue

Example code:

```php
#[Computed]
public function computedPropExample()
{
    $obj = '1234';

    throw_if($obj === '1234', new Exception('Test ex'));

    return $obj;
}

public function exception(Throwable $e, Closure $stopPropagation)
{
    $stopPropagation();
}
```

We will still see '1234' as computedPropExample, even with an exception fired.

### How do you expect it to work?

Catching the exception and firing the `$stopPropagation()` closure should prevent the computed value from returning.

### Please confirm (incomplete submissions will not be addressed)

- [x] I have provided easy and step-by-step instructions to reproduce the bug.
- [x] I have provided code samples as text and NOT images.
- [x] I understand my bug report will be removed if I haven't met the criteria above.

ghabriel25:
@rmdavi Thanks for reporting! I tried your code snippets and it works as expected, there's no return value from the computed property. Removing `$stopPropagation()` from exception hook still throws the exception. Could you clear the cache then try again?

rmdavi:
Ah, you know what, @ghabriel25 I think you are right that the return is not coming through as I had noted. What I am getting my tests is a `null` return, where before there was no return at all.

ghabriel25:
@rmdavi Could you provide minimal example for that so I can test it?

rmdavi:
Of course, @ghabriel25 

```php
	public function mount()
	{
		$value = $this->exampleComputedProp;

		throw_if($value === null, new Exception('Computed prop unexpectedly null'));
	}

	#[Computed]
	public function exampleComputedProp()
	{
		$obj = '1234';

		throw_if(true, new Exception('The value cannot be loaded'));

		return $obj;
	}

	public function exception(Throwable $e, Closure $stopPropagation)
	{
		$stopPropagation();
	}

```

ghabriel25:
Hi @rmdavi Could you explain more about your test? Because I tried again with your new snippet it still works as expected.

[...]

PR #10591, opened 2026-08-25
[4.x] Fix exception thrown from computed properties does not stop propagating when called from lifecycle hooks

ghabriel25:
## Scenario

Accessing computed properties from view which throws an exception now handled by this PR https://github.com/livewire/livewire/pull/10543. However this introduce new regression where computed properties accessed inside component method (action / lifecycle hooks) which create nested exception handler.

## Problem

Component action / lifecycle hooks already use `Wrapped::__call`, so the regression create nested exception handler. If inner handler catch the exception and stop propagating, outer handler will treat it as no exception thrown therefore it will continue the exceution eventhough it shouldnt.

## Solution

Before returning the value from `evaluateComputed()`, we need to determine whether the computed property is being accessed from a component method or while rendering the view.

`HandleComponents::$renderStack` provides the necessary context: when the computed property is accessed while rendering the view, the stack is populated, so the computed method needs to be wrapped to ensure exceptions are handled by Livewire's `exception()` hook. When it is accessed from a component method that is already wrapped by Livewire, the stack is empty, so the computed method can be invoked directly and let the existing outer exception handler handle the exception.

However, an additional case occurs when a computed property is accessed from the component's `render()` method itself.

Currently, `HandleComponents::render()` resolves the view and properties before entering `trackInRenderStack()`:
```php
[ $view, $properties ] = $this->getView($component);

return $this->trackInRenderStack(...);
```
`getView()` invokes the component's `render()` method through `wrap()`:
```php
if (method_exists($component, 'render')) {
    $viewOrString = wrap($component)->render();
}
```
Therefore, when `render()` accesses a computed property, `$renderStack` is still empty. The computed property would be evaluated directly, causing its exception to propagate to the outer `wrap($component)->render()` handler. If the exception is handled with `$stopPropagation()`, that wrapper returns null, which subsequently causes an unrelated `ViewException` when `getView()` attempts to process the render result.

To avoid this, `getView()` needs to execute inside `trackInRenderStack()`:
```php
return $this->trackInRenderStack($component, function () use ($component) {
    [ $view, $properties ] = $this->getView($component);

    // ...
});
```
This ensures `$renderStack` is populated for the entire render lifecycle, including the component's `render()` method and the Blade view.

As a result:

- Component action/method → computed property → **no additional wrapper**
- Component `render()` → computed property → **computed is wrapped**
- Blade view → computed property → **computed is wrapped**

This prevents nested exception handlers for component actions while ensuring exceptions from computed properties accessed during rendering are correctly handled by Livewire's `exception()` hook.

## Fix
This PR only fix regression behavior that surface through this issue https://github.com/livewire/livewire/issues/10588 but does not solve the author's issue

### DIFF
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -2,7 +2,7 @@
 
 namespace Livewire\Features\SupportComputed;
 
-use function Livewire\{ invade, trigger };
+use function Livewire\{ invade, trigger, wrap};
 
 use Livewire\Features\SupportAttributes\Attribute;
 use Illuminate\Support\Facades\Cache;
@@ -158,19 +158,17 @@ class BaseComputed extends Attribute
 
     protected function evaluateComputed()
     {
-        try {
+        if ($this->storeGet('exceptionHandlingDepth', 0) > 0) {
             return invade($this->component)->{parent::getName()}();
-        } catch (\Throwable $e) {
-            $shouldPropagate = true;
-
-            $stopPropagation = function () use (&$shouldPropagate) {
-                $shouldPropagate = false;
-            };
-
-            trigger('exception', $this->component, $e, $stopPropagation);
-
-            $shouldPropagate && throw $e;
         }
+
+        $value = null;
+
+        wrap($this->component)->tap(function ($component) use (&$value) {
+            $value = invade($component)->{parent::getName()}();
+        });
+
+        return $value;
     }
 
     public function getName()
--- a/src/Features/SupportComputed/BaseComputed.php
+++ b/src/Features/SupportComputed/BaseComputed.php
@@ -161,7 +161,7 @@ class BaseComputed extends Attribute
     {
         $evaluated = fn () => invade($this->component)->{parent::getName()}();
 
-        if (empty(HandleComponents::$renderStack)) {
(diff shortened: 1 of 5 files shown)


calebporzio:
gonna need a lot more context here lol

ghabriel25:
@calebporzio Yeah, I still working on it at the moment. It has valid regression test here https://github.com/livewire/livewire/issues/10588 that fails at current behavior. I'll close this if I can't find any solution. Thanks!

ghabriel25:
@calebporzio Done

rmdavi:
@ghabriel25 I applied this patch locally and am still running into the same failures in our test suite as well as in the failing mock test I shared.

ghabriel25:
Hi @rmdavi I have updated the PR description. Yes, this PR doesnt solve your issue only surfacing regression that I've found through your report. Sorry for the misunderstanding.

ghabriel25:
@calebporzio I'm not sure if `render()` should be treated different than other action / lifecycle hooks. Let me know and I'll change it as before along with its test.

calebporzio:
Reverting the original PR that caused this. it's causing too many problems. thanks anyways!

ghabriel25:
I've found the last piece to solve it but its fine I guess.

PR #10592, opened 2026-08-25
Fix Livewire updates from route binding error pages

calebporzio:
## What changed

When Laravel renders an `errors::*` view because route model binding failed, mark Livewire components rendered on that error page in their checksummed snapshot. Subsequent updates for those components skip persistent middleware replay because the original route middleware pipeline never completed.

This is an alternative to #10572. Catching `ModelNotFoundException` around the full persistent middleware pipeline can convert a partial pipeline into a successful update and skip middleware that follows `SubstituteBindings`, including authorization. This approach limits the exception to snapshots the server can prove came from Laravel's route-model-binding error response.

A once-valid route whose model disappears later does not receive the marker, so persistent middleware continues to fail closed.

## Test

- Added a browser regression that renders a Livewire counter inside Laravel's model-binding 404 view and verifies it can update from 0 to 1.
- Ran `./vendor/bin/phpunit --configuration phpunit.xml.dist src/Mechanisms/PersistentMiddleware/UnitTest.php src/Mechanisms/PersistentMiddleware/BrowserTest.php`
- Result: 13 tests, 38 assertions.

### DIFF
--- a/src/Mechanisms/PersistentMiddleware/BrowserTest.php
+++ b/src/Mechanisms/PersistentMiddleware/BrowserTest.php
@@ -35,6 +35,11 @@ class BrowserTest extends BrowserTestCase
 
     public static function tweakApplicationHook() {
         return function() {
+            config()->set('view.paths', [
+                __DIR__.'/fixtures/views',
+                ...config('view.paths'),
+            ]);
+
             Livewire::addPersistentMiddleware([AllowListedMiddleware::class, IsBanned::class]);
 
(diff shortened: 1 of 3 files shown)


ghabriel25:
Love this approach! another new insight for me to handle it at request level 👍

calebporzio:
Closing this approach. Correctly distinguishing an initial route-binding 404 from a model that disappears after a successful render requires carrying exception provenance through the component snapshot lifecycle. That is too much framework machinery for this edge case. A userland opt-out is documented on #10572.

PR #10594, opened 2026-08-25
Add return types on public FrontendAssets methods

AJenbo:
This allows `@livewireScriptConfig` and a few other things to pass type checking in applications that have that enabled.


AJenbo:
Thanks

PR #10596, opened 2026-08-26
[4.x] Don't include redirect responses in the returns effect

### LINKED ISSUE #10595: [4.4.2] Returning a redirect from an action can break the JSON response
### Livewire version

v4.4.2

### Laravel version

v13.26.1

### PHP version

8.4.23

Safari 26.5.2 on macOS 26 (also reproduced on Chrome; it is a server-side error, so the browser is not relevant)

Since v4.4.2, returning a redirect from an action can make the whole Livewire response fail to serialise:

```
InvalidArgumentException: Type is not supported
vendor/laravel/framework/src/Illuminate/Http/JsonResponse.php:91
```

The request returns a 500 and the UI hangs, even though the action itself completed (in our case the record was deleted).

Cause looks like #10555, which changed `Livewire\Features\SupportRedirects\Redirector::to()`:

```diff
-        return $this;
+        return parent::to($path, $status, $headers, $secure);
```

Before, `return redirect()->route(...)` from an action returned the **Redirector**, whose properties are all protected, so it serialised to `{}` and passed through harmlessly. Now it returns a real `Illuminate\Http\RedirectResponse`, which lands in `effects.returns` and gets JSON encoded by `Router::toResponse()`.

We confirmed the location by logging the payload in a `Livewire\on('response')` hook when `json_encode` fails:

```
Livewire payload is not JSON encodable at payload.components.0.effects.returns.0 [Illuminate\Http\RedirectResponse]
```

`HandleComponents::callMethods()` already nulls `StreamedResponse` and `BinaryFileResponse` in `returns` for exactly this reason.

One caveat worth flagging: this is environment dependent. On our production server the `RedirectResponse` fails to encode, but locally the same object encodes fine, so you may not hit the 500 immediately. Either way, the response object should probably not be in `returns` at all.

Single-file component:

```php
<?php

use Livewire\Component;

new class extends Component
{
    public function go()
    {
        return redirect()->route('dashboard');
    }
};

?>

<div>
    <button type="button" wire:click="go">Go</button>
</div>
```

Steps:

1. Install Livewire v4.4.2.
2. Add the component above to any page, pointing at any existing named route.
3. Click the button.
4. Inspect the `POST /livewire/update` response. `effects.returns[0]` now contains the serialised `RedirectResponse` instead of the `{}` it contained on v4.4.1. On our production environment this throws `InvalidArgumentException: Type is not supported` and returns a 500.

To see the difference without needing the 500, `composer require livewire/livewire:4.4.1` and compare `effects.returns[0]` in the response.

_No response_

Returning a redirect from an action is idiomatic Laravel and is shown in the Livewire docs, so it should not be able to break the response payload.

The redirect is already delivered through the separate `redirect` effect, so the response object in `returns` is redundant. I would expect `RedirectResponse` to be nulled there, the same way `StreamedResponse` and `BinaryFileResponse` already are:

```php
if ($return instanceof StreamedResponse || $return instanceof BinaryFileResponse) {
    $return = null;
}
```

lazerg:
Since #10555, `Redirector::to()` returns a real `RedirectResponse` instead of `$this`. An action that returns `redirect()->route(...)` now drops that response object into `effects.returns`, where it gets JSON encoded along with the rest of the payload. Depending on what the response is carrying, encoding can fail with `InvalidArgumentException: Type is not supported`, and the update request 500s even though the action itself ran fine.

The redirect already reaches the browser through the `redirect` effect, so the response object has no reason to sit in `returns`. It's now nulled out there, the same way `StreamedResponse` and `BinaryFileResponse` already are.

Fixes #10595

### DIFF
--- a/src/Features/SupportRedirects/UnitTest.php
+++ b/src/Features/SupportRedirects/UnitTest.php
@@ -288,6 +288,15 @@ class UnitTest extends \Tests\TestCase
         $this->assertEquals('livewire-is-awesome', Session::get('success'));
     }
 
+    public function test_redirect_response_is_not_included_in_returns()
+    {
+        $component = Livewire::test(TriggersRedirectStub::class);
+
+        $component->runAction('triggerRedirectHelper');
+
+        $this->assertNull($component->effects['returns'][0]);
+    }
+
     protected function registerNamedRoute()
     {
         Route::get('foo', function () {
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -14,6 +14,7 @@ use Livewire\Features\SupportEvents\SupportEvents;
 use Livewire\Features\SupportFormObjects\Form;
 use Illuminate\Support\Facades\View;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
 class HandleComponents extends Mechanism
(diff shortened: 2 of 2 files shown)


calebporzio:
thanks!

Issue #10597, opened 2026-08-26
Intermittent "No hint path defined for [layouts]" on Filament panel pages (page-layout capture race in SupportPageComponents)


ahmadabuzaiid:
### Livewire version

v4.4.2

### Laravel version

v13.29.0

### PHP version

8.5.9

### Browser and operating system

Chrome on Ubuntu 26.04 lts

### Describe the issue you're experiencing

**Package**: livewire/livewire
**Version**: v4.4.1 and v4.4.2 (same behavior, diffed SupportPageComponents.php between them — no relevant change)
**Laravel Version**: v13.26.1 / v13.29.0 (both affected)
**PHP Version**: 8.5.9
**Also using**: filament/filament v5.7.6 (filament/support requires livewire/livewire: ^4.1)

### Problem description

Filament panel pages intermittently throw:

```
InvalidArgumentException
No hint path defined for [layouts].
vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php:111
```

This is the same underlying symptom as filamentphp/filament#19020 (fixed on the Filament side in filamentphp/filament#19049 by scaffolding `resources/views/layouts/app.blade.php`), but that fix only applies to the "individual components" installer flow. Filament *panels* never scaffold that file, because `Filament\Pages\BasePage::render()` always calls `->layout(...)` explicitly and should never need the `config('livewire.component_layout')` default. In our case it does anyway, intermittently.

### What we found tracing SupportPageComponents.php

`interceptTheRenderOfTheComponentAndRetreiveTheLayoutConfiguration()` opens a listener window around `app('livewire')->mount(...)`:

```php
$handler = once(function ($target, $view, $data) use (&$layoutConfig, &$slots) {
    ...
    $layoutConfig = $view->layoutConfig ?? new PageComponentConfig;
    ...
});
on('render', $handler);
on('render.placeholder', $handler);
$callback(); // mounts + renders the page component tree
off('render', $handler);
off('render.placeholder', $handler);
```

`once()` here (Livewire's own helper in `src/helpers.php`, not Laravel's global one) only lets the *first* `render`/`render.placeholder` event within this single call actually run the capture logic — intentionally, per the comment, so nested child components rendered during the same page mount don't overwrite the parent page's layout.

[...]

PR #10598, opened 2026-08-26
[4.x] Fix `SupportPageComponents` render listener cleanup on skip-render or exceptions

ghabriel25:
> [!NOTE]
> I discovered this while reading Livewire codebase but I think this is somehow related with this issue  https://github.com/livewire/livewire/issues/10597

## Problem

When `SupportPageComponents::interceptTheRenderOfTheComponentAndRetreiveTheLayoutConfiguration()` registers its temporary `render` listeners, they are only removed after the callback completes successfully.

```php
on('render', $handler);
on('render.placeholder', $handler);

$callback();

off('render', $handler);
off('render.placeholder', $handler);
```

If the `$callback()` throws an exception (or the render path is skipped entirely via `shouldSkipRender` / short-circuit mount), the cleanup code is skipped and the listeners remain registered on the event bus. On a later request that **does** reach the page-component path, the already-spent `once()` handler is a no-op → `$layoutConfig` stays `null` then it will fallback to Livewire default layout.

```php
$layoutConfig = intercept(...) ?: new PageComponentConfig;
```

`PageComponentConfig` defaults to `config('livewire.component_layout') = 'layouts::app'`.

## Solution

This change wraps the callback in a try/finally block so the temporary listeners are always removed, while keeping the existing render interception flow unchanged.

A regression test is added to verify that both `render` and `render.placeholder` listeners are cleaned up when the callback throws an exception.

### DIFF
--- a/src/Features/SupportPageComponents/SupportPageComponents.php
+++ b/src/Features/SupportPageComponents/SupportPageComponents.php
@@ -2,7 +2,7 @@
 
 namespace Livewire\Features\SupportPageComponents;
 
-use function Livewire\{on, off, once};
+use function Livewire\{on, once};
 use Livewire\Drawer\ImplicitRouteBinding;
 use Livewire\ComponentHook;
 use Livewire\Mechanisms\HandleRouting\LivewirePageController;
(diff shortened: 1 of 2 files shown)


calebporzio:
good catch!

PR #10599, opened 2026-08-26
Add continuous visibility timing to wire:intersect

cyppe:
## Why

`wire:intersect` can currently tell when an element crosses a visibility threshold, but not whether it remains there. A fast scroll, layout shift, or backgrounded tab can therefore trigger work that was intended only for genuinely viewed content.

This adds a continuous-visibility contract that reads naturally in Blade:

```blade
<div wire:intersect.once.half.dwell.500ms="trackImpression">
    ...
</div>
```

The action runs only after the element stays at least half visible for 500 milliseconds. Dropping below the threshold or moving the page to the background resets the dwell interval.

Leave actions use the same vocabulary:

```blade
<video wire:intersect:leave.half.dwell.750ms="pausePreview">...</video>
```

This makes qualified impressions, deferred media work, reading-progress signals, and similar visibility-sensitive actions declarative without application-owned observers and timers.

## What changed

- Documents `.dwell` with its 250 ms default, explicit durations, thresholds, leave behavior, page-visibility reset, and `.once` composition.
- Adds a narrow browser contract proving the complete modifier suffix reaches Alpine, plus Livewire-owned morph initialization and renderless action integration.
- Keeps the Livewire bridge itself unchanged: `wire:intersect` already forwards every modifier to Alpine `x-intersect`, so duplicating the dwell state machine in Livewire would create two implementations.

## Dependency

This draft depends on [alpinejs/alpine#4884](https://github.com/alpinejs/alpine/pull/4884), which implements the continuous dwell primitive in `@alpinejs/intersect`.

Livewire 4.x currently installs `@alpinejs/intersect` 3.16.3, which does not include `.dwell`. Livewire CI also links Alpine `main`, so this browser contract becomes available after the Alpine PR merges and the normal published Alpine dependency bump is applied. This PR should not merge before that sequence; it deliberately contains no fork reference, unpublished dependency, or duplicate JavaScript implementation.

## Validation

- Focused browser contract: 3 tests, 6 assertions (using the locally linked Alpine feature branch)
- JavaScript suite: 110 passed, 1 skipped
- PHP unit suite: 1,391 tests, 3,964 assertions, 13 skipped, 3 incomplete
- Production asset build passed
- Diff checks passed

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -733,18 +733,25 @@ Run actions when elements enter or leave the viewport, similar to Alpine's [`x-i
 <div wire:intersect:leave="pauseVideo">...</div>
 <div wire:intersect.half="loadMore">...</div>
 <div wire:intersect.full="startAnimation">...</div>
+<div wire:intersect.half.dwell.500ms="trackImpression">...</div>
 
 <!-- With options -->
 <div wire:intersect.margin.200px="loadMore">...</div>
 <div wire:intersect.threshold.50="trackScroll">...</div>
+<div wire:intersect.parent="loadWithinScroller">...</div>
 ```
 
 Available modifiers:
 - `.once` - Fire only once
+- `.dwell.Xms` - Require continuous intersection for a duration (250ms by default)
 - `.half` - Wait until half is visible
 - `.full` - Wait until fully visible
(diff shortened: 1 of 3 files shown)


calebporzio:
this is great! thanks!

PR #10600, opened 2026-08-26
[4.x] Add `wire:dirty.persist` for dirty state that survives round-trips

stebrl:
## The problem

`$dirty` compares the reactive state against `component.canonical`, and `canonical` is replaced on every snapshot merge (`js/component.js`, `mergeNewSnapshot()`).

So `wire:dirty` answers *"are there changes the server hasn't seen yet"*, not *"are there unsaved changes"*. Any request resets it — a `wire:poll`, a lazy load, a `wire:model.live` update, an unrelated action elsewhere in the component. For the most common thing people reach for it for, an "unsaved changes" indicator or navigation guard, it reports clean while the form is still unsaved.

## The change

Components keep a second reference point next to `canonical`:

```js
// "baseline" represents the last state the component considers "saved". Unlike
// "canonical", it survives round-trips: it is only replaced by an explicit rebaseline.
this.baseline = extractData(deepClone(this.snapshot.data))
```

| API | |
|---|---|
| `wire:dirty.persist` | measures against the last **saved** state instead of the last server response |
| `$wire.$dirty('title', { persist: true })` | the same from Alpine |
| `$this->rebaseline()` | server-side: declares the state it is returning as saved |
| `$wire.$rebaseline()` | client-side equivalent |

```blade
<form wire:submit="save">
    <input type="text" wire:model="title">

    <div wire:dirty.persist>Unsaved changes...</div>
</form>
```

```php
public function save()
{
    Post::create($this->only('title'));

    $this->rebaseline();
}
```

`.persist` composes with `wire:target`, `.remove` and `.class`.

Existing behaviour is untouched — `persist` is opt-in and the default path still compares against `canonical`.

## Two supporting changes

**The post-commit refresh now re-derives the state** instead of hard-coding `false`:

```diff
-refreshDirtyStatesByComponent.each(component, i => i(false))
+refreshDirtyStatesByComponent.each(component, recompute => recompute(true))
```

For a plain `wire:dirty` this is the same answer the hard-coded `false` gave — once the snapshot is merged, `canonical` and `reactive` match — but a `.persist` directive has to be asked rather than assumed. The `force` flag preserves the existing "re-apply after the morph" behaviour.

**The injected stylesheet hides `[wire:dirty.persist]`** the way it already hides `[wire:dirty]`. The selector matches the exact attribute name, so a modifier variant would otherwise flash visible on first paint.

## Tests

Four browser tests in `src/Features/SupportDirty/BrowserTest.php`. The first one pins the *current* behaviour (an unrelated round-trip clears `wire:dirty`) so the difference is visible in the suite:

- `wire:dirty` is reset by an unrelated roundtrip
- `wire:dirty.persist` survives an unrelated roundtrip
- `$dirty` accepts a persist option and `$rebaseline` clears it
- persist composes with the class modifier

Docs updated in `docs/wire-dirty.md`.

### DIFF
--- a/docs/wire-dirty.md
+++ b/docs/wire-dirty.md
@@ -39,55 +39,6 @@ By adding the `.remove` modifier to `wire:dirty`, you can instead show an elemen
 <div wire:dirty.remove>The data is in-sync...</div>
 ```
 
-## Persisting dirty state across requests
-
-By default, "dirty" means "the client has changes the server hasn't seen yet". Any request resets it — including requests that save nothing, like `wire:poll`, a lazy load, or an unrelated action elsewhere in the component.
-
-That is the wrong measure for an "unsaved changes" warning, where you want to know whether anything has changed since the last *save*. The `.persist` modifier measures against the last saved state instead, and only an explicit rebaseline clears it:
-
-```blade
-<form wire:submit="save">
-    <input type="text" wire:model="title">
-
-    <button type="submit">Save</button>
-
-    <div wire:dirty.persist>Unsaved changes...</div> <!-- [tl! highlight] -->
-</form>
-```
-
-Tell Livewire when the state was saved by calling `rebaseline()` from the action that saved it:
-
-```php
-public function save()
-{
-    Post::create($this->only('title'));
-
-    $this->rebaseline(); // [tl! highlight]
-}
-```
-
-Until `save()` runs, the indicator stays visible — a poll ticking in the background no longer makes the form look saved.
-
-`.persist` composes with the other modifiers and with `wire:target`:
-
-```blade
-<div wire:dirty.persist wire:target="title">Unsaved title...</div>
-
-<div wire:dirty.persist.remove>All changes saved</div>
-
-<input wire:model="title" wire:dirty.persist.class="border-yellow-500">
-```
-
-### From Alpine
-
-`$dirty` takes the same option, and `$rebaseline()` clears it from the client:
-
-```blade
-<div x-show="$wire.$dirty('title', { persist: true })">Unsaved title...</div>
-
-<button x-on:click="$wire.$rebaseline()">Mark as saved</button>
-```
-
 ## Targeting property updates
 
 Imagine you are using `wire:model.live.blur` to update a property on the server immediately after a user leaves an input field. In this scenario, you can provide a "dirty" indication for only that property by adding `wire:target` to the element that contains the `wire:dirty` directive.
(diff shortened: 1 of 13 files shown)


stebrl:
**Optional follow-up: `#[Rebaseline]`**

One thing I left out deliberately, in case you'd rather keep this PR to a single concept.

Calling `$this->rebaseline()` from inside the action works, but it puts a correctness detail in the method body, where it can end up on the wrong branch — after an early return, or before a `validate()` that throws.

An attribute would move it to a declaration, in the same shape as `#[Renderless]`:

```php
#[Rebaseline]
public function save()
{
    $this->validate();

    Post::create($this->only(['title']));
}
```

`SupportAttributes::call()` (`SupportAttributes.php:56`) runs whatever an attribute's `call()` returns as a finish callback, and `HandleComponents.php:584` invokes it after the method. So the exception case is handled for free: a failed `validate()` never reaches the rebaseline. `$this->rebaseline()` would stay for conditional saves, the way `$this->skipRender()` stays alongside `#[Renderless]`.

It's about ten lines plus a unit test. Happy to push it here, or leave it for a separate PR — whichever you prefer. Naming is open too; `#[Rebaseline]` matches the `.persist` internals but reads a bit technical.

calebporzio:
Thanks for putting this together — the distinction you identified is real, and the investigation was useful.

After working through the framework-level versions in #10618, I think we should keep this in userland. Livewire can know whether state has reached the server, but it can't know which fields an application considers durably saved, whether a partial save should advance the baseline, or whether a failed save counts. Keeping `wire:dirty` scoped to client/server synchronization leaves that meaning precise.

A user can track an application-specific saved baseline with Alpine. This version advances only when the server dispatches `post-saved`, and it captures the submitted snapshot so an edit made while the request is in flight remains dirty:

```php
<?php // resources/views/components/post/⚡edit.blade.php

use Livewire\Component;
use App\Models\Post;

new class extends Component {
    public Post $post;

    public string $title = '';
    public string $body = '';

    public function save()
    {
        $this->validate([
            'title' => ['required'],
            'body' => ['required'],
        ]);

        $this->post->update($this->only(['title', 'body']));

        $this->dispatch('post-saved');
    }
};
?>

<form
    wire:submit="save"
    x-data="{
        saved: null,
        pending: null,

        snapshot() {
            return JSON.stringify({
                title: $wire.title,
                body: $wire.body,
            })
        },

        init() {
            this.saved = this.snapshot()
        },

        beginSave() {
            this.pending = this.snapshot()
        },
    }"
    x-on:submit="beginSave()"
    x-on:post-saved.window="saved = pending"
>
    <input wire:model="title">
    <textarea wire:model="body"></textarea>

    <div x-show="saved !== snapshot()">Unsaved changes...</div>

</form>
```

Because `post-saved` is dispatched after validation and persistence, failed saves leave the old baseline intact. Thanks again for exploring this.

PR #10603, opened 2026-08-26
[v4.x] fix: wire:dirty losing reactivity for targets listed after a dirty one

peaklabs-dev:
## Issue

When `wire:dirty` watches multiple targets (e.g. `wire:dirty.class="unsaved" wire:target="a, b, c"`), the directive permanently stops reacting to any target listed *after* one that becomes dirty and is then committed. Issue reported on coollabsio/coolify#11265.

## Root cause

The dirty check loops over targets inside an `Alpine.effect` and `breaks` as soon as one target is dirty. Alpine's reactivity re-tracks effect dependencies on every run, only properties actually read during the latest run stay subscribed. Breaking early means later targets aren't read, so the effect drops its subscription to them.

The dirty state is then cleared out-of-band after a commit (the `on('commit')` hook calls `refreshDirtyState(false)` directly instead of re-running the effect), so the effect never performs a full read again.

### DIFF
--- a/js/directives/wire-dirty.js
+++ b/js/directives/wire-dirty.js
@@ -47,12 +47,10 @@ export function checkDirty(component, targets) {
     if (targets === undefined) {
         isDirty = JSON.stringify(component.canonical) !== JSON.stringify(component.reactive)
     } else if (Array.isArray(targets)) {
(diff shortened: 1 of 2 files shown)


calebporzio:
fixed in #10616

PR #10607, opened 2026-08-27
[4.x] Don't reuse snapshot synthesizer meta when an updated value changes type

### LINKED ISSUE #10606: [v4.4.1] [v4.4.2] Regression in hydration process, test fails after package update
### Livewire version

v4.4.2

### Laravel version

v13.29.0

### PHP version

8.5.9

Chrome on Arch Linux

There seems to be some kind of regression in the hydrate logic of Wireable objects. In my big production project a test started to fail after updating to version v4.4.1 or v4.4.2. The test passes in previous versions including v.4.4.0. 

I have created a repository that replicates the error: [livewire-4-testing-array-hydration-error](https://github.com/reafeichtinger/livewire-4-testing-array-hydration-error/blob/main/tests/Feature/ExampleTest.php#L7-L25)

Setting the array property to a primitive value, e.g. [int => int|float] throws an exception.
This only happens if it was set ot an object (implementing the Wireable interface) first.
```bash
foreach() argument must be of type array|object, int given

 at vendor/livewire/livewire/src/Features/SupportWireables/WireableSynth.php:42
    38▕         if (! isset($meta['class']) || ! is_a($meta['class'], Wireable::class, true)) {
    39▕             throw new \Exception('Livewire: Invalid wireable class.');
    40▕         }
    41▕ 
 ➜  42▕         foreach ($value as $key => $child) {
    43▕             $value[$key] = $hydrateChild($key, $child);
    44▕         }
    45▕ 
    46▕         return $meta['class']::fromLivewire($value);
 ```
For now my workaround is to set the object last but it's not ideal. Maybe this is intentional but maybe this points to a regression in the v4.4.1 update.

In a test:
```php
$this->component->set('array', [0 => new EmptyObject]);
$this->component->set('array', [0 => 1]);
```

The Object:
```php
class EmptyObject implements Wireable
{
    public function toLivewire(): array
    {
        return [];
    }

    {
        return new static;
    }
}
```

_No response_

Same as before, the test should pass and not throw an exception.

lazerg:
## The problem

Since 4.4.1, updating a property to a value of a different shape than the one in the previous snapshot throws. Setting an array item to a `Wireable` object and then to an integer gives `foreach() argument must be of type array|object, int given` from `WireableSynth::hydrate()`.

Recursive update hydration reconstructs nested values from the meta in the previous snapshot, since the browser strips meta out of update payloads. It was applying that meta to whatever came back for the path, even when the new value can't be what the meta describes, so a synthesizer expecting an array got handed a scalar.

## The fix

Reuse the previous meta only when the incoming value still has the same wire shape as the value it was recorded for, comparing array against scalar against null rather than the exact PHP type. Anything else passes through as-is, which is what happened before 4.4.1.

Comparing the exact type would have been too strict: the browser sends numbers back as strings, so an int-backed enum arrives as `"2"` where the snapshot holds `2`, and that has to keep hydrating.

## Verification

- `phpunit --testsuite=Unit src/Mechanisms/HandleSynths/UnitTest.php` — 20 passed. Both new tests fail on 4.x without the change.
- Full unit suite: 1393 tests, no new failures.
- The nested denylist and forged-meta tests from #10489 still reach the security path rather than returning early at the new check.

Fixes #10606

### DIFF
--- a/src/Mechanisms/HandleSynths/HandleSynths.php
+++ b/src/Mechanisms/HandleSynths/HandleSynths.php
@@ -116,7 +116,7 @@ class HandleSynths extends Mechanism
             // The update can also swap a child between an array and a scalar —
             // a Wireable replaced by an integer, say — and the previous meta no
             // longer describes a value of that shape...
-            if (is_array($child) !== is_array($childData) || is_null($child) !== is_null($childData)) return $child;
+            if (! $this->metaStillApplies($child, $childData)) return $child;
 
             // The child value is untrusted update data and may itself look like a
             // synthetic tuple. Always pair it with the authenticated snapshot meta;
@@ -130,10 +130,10 @@ class HandleSynths extends Mechanism
         // This is a trust boundary: $value came from the update payload, while
         // $raw came from the snapshot that Checksum::verify() authenticated. Synth
         // identity and class must only ever be selected from the latter...
-        $meta = $this->getMetaForPath($raw, $path);
+        [$data, $meta] = $this->getTupleForPath($raw, $path);
 
         // If we have meta data already for this property, let's use that to get a synth...
-        if ($meta) {
+        if ($meta && $this->metaStillApplies($value, $data)) {
             return $this->hydratePropertyUpdate([$value, $meta], $context, $path, $raw);
         }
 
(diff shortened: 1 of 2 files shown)


ghabriel25:
@lazerg Top-level (or any) update can replace a synthesized value with a different wire shape — e.g. a `Wireable` property set to an `int`. Previous meta no longer describes the incoming value.

lazerg:
Confirmed, I only guarded the recursion. A top-level `Wireable` set to an int still blew up in `WireableSynth::hydrate()` on this branch.

Pulled the shape check into a helper and applied it in `hydrateForUpdate()` before the snapshot meta gets used, so it now covers the updated path itself for any synth. Test added for the `Wireable` case: a8bc776c1

PR #10611, opened 2026-08-28
[4.x] Guard array-shaped synthetizers against non-array update values

### LINKED ISSUE #10606: [v4.4.1] [v4.4.2] Regression in hydration process, test fails after package update

ghabriel25:
Alternative approaches from #10607 but more likely as simple guard against recursive hydrate nested synthesizers which targeting array-shaped synths.

## Problem

Since v4.4.1 (#10489), recursive update hydration re-applies synthesizer meta from the previous snapshot to nested values. That correctly keeps Collections / uploads / etc. intact when an ancestor path is updated.

It breaks when a nested (or top-level) value **changes shape** — e.g. an array item that was a `Wireable` is later set to a scalar:

```php
$component->set('array', [0 => new EmptyObject]); // Wireable
$component->set('array', [0 => 1]);               // int
// foreach() argument must be of type array|object, int given
// at WireableSynth.php
```

`WireableSynth`, `CollectionSynth`, `StdClassSynth`, and `FormObjectSynth` always loop the value. After #10489 they can receive a scalar (or `__rm__`) while still carrying the old meta.

`ArraySynth` already handled this case:

```php
// If we are "hydrating" a value about to be used in an update,
// Let's make sure it's actually an array before try to set it.
// This is most common in the case of "__rm__" values, but also
// applies to any non-array value...
if (! is_array($value)) {
    return $value;
}
```

The other array-shaped synthesizers did not. #10489 only exposed the gap by reusing snapshot meta on updates.

## Solution

Introduce a marker interface and enforce the rule once in `HandleSynths`:

```php
interface ArrayShapedSynth
{
    //
}
```

Synths that implement it: `ArraySynth`, `CollectionSynth` (via inheritance), `StdClassSynth`, `WireableSynth`, `FormObjectSynth`.

In `hydratePropertyUpdate`, after resolving the synth from snapshot meta:

```php
if ($synth instanceof ArrayShapedSynth && ! is_array($value)) {
    return $value;
}
```

- Snapshot meta is still attached exactly as today (security contract unchanged: meta only from the checksum-verified snapshot).
- Array-shaped synths are skipped when the incoming value is not an array; scalar-friendly synths (Carbon, Enum, Int, Float, Stringable) are unaffected.
- Third-party / userland synths that assume array wire data implement the same interface — no central key list and no special `registerSynth()` path.

`ArraySynth`’s existing local guard is left in place for non-update paths. No per-synth one-liners are added on the other classes.

No changes to the #10489 recursion path. Nested Collections and uploads still re-hydrate when the wire value remains array-shaped; type swaps and `__rm__` fall through cleanly.

This is an alternative to per-synth guards and to approaches that compare the incoming value’s shape against the previous value before attaching meta. Here, meta is always taken from the snapshot; we only refuse to apply it when the value cannot be what that meta describes.

## Tests

- Nested Collection replaced with a scalar is left alone; siblings still re-hydrate
- Nested Wireable / stdClass replaced with a scalar is left alone
- Top-level Collection / Wireable / stdClass replaced with a scalar is left alone
- Replacing a Wireable array item with a scalar via `Livewire::test()->set()` does not throw
- Form object shape change is left alone

## Docs

Add the information about this in [Synthetizers](https://livewire.laravel.com/docs/4.x/synthesizers)

Fixes: https://github.com/livewire/livewire/issues/10606

### DIFF
--- a/docs/synthesizers.md
+++ b/docs/synthesizers.md
@@ -244,3 +244,30 @@ class AddressSynth extends Synth
     }
 }
 ```
+
+## Array-shaped synthesizers
+
+If your synthesizer's `hydrate()` method assumes the wire value is an array
+(e.g. it runs `foreach`, builds a collection, or calls `fromLivewire()` on
+array data), implement the `ArrayShapedSynth` marker interface:
+
+```php
+use Livewire\Mechanisms\HandleComponents\Synthesizers\ArrayShapedSynth;
+use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
+
+class MoneySynth extends Synth implements ArrayShapedSynth
+{
+    public static $key = 'money';
+
+    public function hydrate($value, $meta, $hydrateChild)
+    {
+        // $value is guaranteed to be an array on the update path when
+        // this interface is implemented; non-arrays are returned as-is
+        // by HandleSynths before hydrate() is called.
+        // ...
+    }
+}
+```
+> [!important]
+> Do not implement this interface on synths that legitimately hydrate scalars
+(dates, enums, ints, floats, strings).
--- a/src/Features/SupportFormObjects/FormObjectSynth.php
+++ b/src/Features/SupportFormObjects/FormObjectSynth.php
@@ -5,10 +5,11 @@ namespace Livewire\Features\SupportFormObjects;
 use Livewire\Drawer\Utils;
 use Livewire\Mechanisms\HandleComponents\Synthesizers\Synth;
 use Livewire\Features\SupportAttributes\AttributeCollection;
+use Livewire\Mechanisms\HandleComponents\Synthesizers\ArrayShapedSynth;
 
 use function Livewire\wrap;
 
-class FormObjectSynth extends Synth {
+class FormObjectSynth extends Synth implements ArrayShapedSynth {
     public static $key = 'form';
 
     static function match($target)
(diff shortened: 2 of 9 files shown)


ghabriel25:
There is still a gap for `FormObjectSynth` but I dont want to touch that if there is no issue reported

PR #10613, opened 2026-08-28
Fix Livewire updates from route binding error pages

calebporzio:
## Problem

When route model binding renders a custom 404 page containing a Livewire component, interacting with that component opens a second 404 in Livewire's error modal instead of running the action.

The component's snapshot records the failed route's path and method even though that route's middleware pipeline never completed. On update, Livewire treats those values as a successful middleware context, rematches the route, and hits the same missing binding before the component action can run.

## Approaches considered

- **Minimum:** catch `ModelNotFoundException` during persistent middleware replay. This is tiny, but it fails open: middleware after `SubstituteBindings`, including authorization, can be skipped.
- **Maximum:** record the exact middleware context that completed during every initial request and replay only that context. This is conceptually pure, but it redesigns the persistent middleware lifecycle for a narrow error-page bug.
- **Evaporate:** render exception pages through a separate internal/fallback route, or make them non-Livewire, so the failed route never becomes the component's update context. That moves the problem into Laravel/application error-page architecture or removes the desired interactivity.

## Chosen path

Capture the provenance on the original request: when Laravel renders `errors::*` because route model binding failed, mark each component in its checksummed snapshot. Only those verified snapshots skip persistent middleware replay.

The marker is restored per component on hydration and inherited by children mounted later, so it survives repeated interactions without leaking across bundled components. A component rendered from a valid route receives no marker; if its model disappears later, the update still fails closed before its action runs.

## Levels

1. [Level 1: Ignore missing models during middleware replay](https://github.com/livewire/livewire/commit/9e651f5179715d6cc407b2657e7a4e1984f30c80) — the smallest readable patch and the heart of the original proposal.
2. [Level 2: Preserve verified error-page context](https://github.com/livewire/livewire/commit/260ee4c45b552672b42a6ee618cb01a61a9c8676) — replaces the broad catch with the balanced, fail-closed solution and adds end-to-end and security regressions.

## Verification

- Reproduced the original behavior in a real Laravel app: `Count: 0` opened a second 404 in Livewire's modal.
- Verified the final playground behavior through `Count: 0 → 1 → 2` with no error modal.
- `./vendor/bin/phpunit --configuration phpunit.xml.dist src/Mechanisms/PersistentMiddleware/UnitTest.php` — 5 tests, 9 assertions.
- `composer test:browser -- --filter=PersistentMiddleware` — 9 tests, 35 assertions.

This supersedes #10592 and addresses the report and minimal proposal in #10572.

## Contribution checklist

1. This behavior is wanted and already demonstrated by #10572 and #10592; no separate discussion was opened.
2. The change is on a dedicated branch.
3. The PR contains one focused fix.
4. It includes unit and browser coverage.

### DIFF
--- a/src/Mechanisms/PersistentMiddleware/BrowserTest.php
+++ b/src/Mechanisms/PersistentMiddleware/BrowserTest.php
@@ -35,6 +35,11 @@ class BrowserTest extends BrowserTestCase
 
     public static function tweakApplicationHook() {
         return function() {
+            config()->set('view.paths', [
+                __DIR__.'/fixtures/views',
+                ...config('view.paths'),
+            ]);
+
             Livewire::addPersistentMiddleware([AllowListedMiddleware::class, IsBanned::class]);
 
             // Overwrite the default route for these tests, so the middleware is included
@@ -85,6 +90,8 @@ class BrowserTest extends BrowserTestCase
             Livewire::component('child-with-modelable', ChildComponentWithModelable::class);
             Livewire::component('page-with-explicit-binding', PageComponentWithExplicitBinding::class);
             Livewire::component('child-for-explicit-binding', ChildForExplicitBinding::class);
+            Livewire::component('error-page-counter', ErrorPageCounter::class);
+            Livewire::component('error-page-child', ErrorPageChild::class);
 
             Route::get('/page-with-reactive-child/{post}', PageComponentWithReactiveChild::class)
                 ->middleware('web');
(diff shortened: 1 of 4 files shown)


calebporzio:
Closing this after working through the minimal, provenance-aware, and architectural approaches. The complete version is too sprawling for this edge case, while the small versions have ambiguous middleware and authorization semantics. A userland workaround is documented on #10572.

PR #10615, opened 2026-08-28
[4.x] Allow original filenames for temporary preview downloads

calebporzio:
This follows up on #10381 and applies the opt-in approach to the current `4.x` signed-preview URL flow.

## Problem

- **What the user experiences:** Downloading a locally served temporary preview uses Livewire's opaque temporary hash as the suggested filename.
- **What they expected:** The download should be suggested with the name of the file they uploaded, such as `avatar.jpg`.
- **What caused the difference:** The local preview response uses the same value as both the private storage lookup key and the public `Content-Disposition` filename. S3/GCS already distinguish these concerns and return the original upload name.

## Solution

The final API preserves existing behavior by default and opts into the original name with `temporaryUrl(useOriginalFilename: true)`. The option travels inside the signed URL and is only honored after signature validation.

- [Level 1 — separate the storage key from the download name](https://github.com/livewire/livewire/commit/b82645a5cb8ccf51e66d035e06249acb70145fae): the smallest readable fix, using the original name for every local preview download.
- [Level 2 — make the behavior opt-in](https://github.com/livewire/livewire/commit/7df879ea27e28268a7c23911189ae18a00a108e0): preserve the current default, add the named argument, and carry the choice through the signed preview URL.
- [Level 3 — verify and document the boundary](https://github.com/livewire/livewire/commit/6ba30d9dce6bfb1fbc4e22591ba4042ffa08bb79): prove the option cannot be appended after signing and document the user-facing API.

```blade
<a href="{{ $photo->temporaryUrl(useOriginalFilename: true) }}">Download preview</a>
```

## Verification

- Reproduced the hashed filename against unmodified `4.x`.
- Passed all 1,397 unit tests (3 incomplete, 13 skipped).
- Passed all 65 `SupportFileUploads` unit tests.
- Verified in a real Laravel app that the default URL returns the temporary name, the opt-in URL returns the original name, and a tampered URL returns 401.

### DIFF
--- a/docs/uploads.md
+++ b/docs/uploads.md
@@ -179,6 +179,12 @@ However, Livewire solves this issue by providing a temporary, signed URL that pr
 
 This URL is protected against showing files in directories above the temporary directory. And, because it's signed, users can't abuse this URL to preview other files on your system.
 
+If users can download the preview, you may use the original upload name as the suggested filename by passing `useOriginalFilename: true`:
+
+```blade
+<a href="{{ $photo->temporaryUrl(useOriginalFilename: true) }}">Download preview</a>
+```
+
 > [!tip] S3 temporary signed URLs
 > If you've configured Livewire to use S3 for temporary file storage, calling `->temporaryUrl()` will generate a temporary, signed URL to S3 directly so that image previews aren't loaded from your Laravel application server.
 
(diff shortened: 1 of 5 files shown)


calebporzio:
Closing this after the demand and API-surface review in #10381. The implementation is sound, but the exact behavior currently has one direct requester and is better handled by an explicit userland download action until broader demand emerges.

PR #10617, opened 2026-08-28
[4.x] Preserve structured strings in URL properties

calebporzio:
## Problem

- **What does the user experience?** Visiting a page with a scalar query value such as `?search=[123]` changes a string-backed `#[Url]` property into an array. Consumers such as Filament then hit an `Array to string conversion` error and the page fails to load.
- **What did they expect?** The scalar query value should remain the literal string `[123]`, so searching and rendering continue normally. Actual query arrays written with bracket notation, such as `?search[0]=123`, should still become arrays.
- **What caused the difference?** `BaseUrl` runs every scalar query value through `json_decode()`. Structured-looking text decodes into an array before the property is assigned, even though PHP's query parser had already distinguished a scalar value from a bracket-notation array.

This is an alternative implementation of the problem reported in [#10602](https://github.com/livewire/livewire/pull/10602) and [filamentphp/filament#20372](https://github.com/filamentphp/filament/issues/20372).

## Solution

`Utils::propertyTypeMatchesValue()` now owns the general rule for strictly comparing a value's native PHP type with named, nullable, union, intersection, class, and interface property types. It deliberately reports an unknown result for missing or untyped properties instead of guessing.

`BaseUrl` uses that utility to compare both the decoded structured value and the original scalar. It preserves the original only when the property type prefers it; legacy untyped properties fall back to the type of their current value. Existing scalar coercion for numbers and booleans remains unchanged, while real bracket-notation arrays continue through the existing array path.

The solution is intentionally built in independently workable levels:

- [**Level 1 — initialized values**](https://github.com/livewire/livewire/commit/10ebfb0b4ac9b0ad2e0c990970611192908e001c): the smallest readable guard; preserves structured-looking scalar text when the current property value is a string.
- [**Level 2 — uninitialized typed values**](https://github.com/livewire/livewire/commit/0745efaaef95fa5925efa58db026f544a8d177a0): extends the guard to uninitialized nullable string properties by consulting their declared type.
- [**Level 3 — complete string-capable coverage**](https://github.com/livewire/livewire/commit/ca0eb1879bd2181f3f56f31cb95d4b65a1872ac0): covers the public `#[Url]` attribute, legacy query-string declarations, union types, JSON object strings, and verifies in a browser that bracket-notation arrays remain arrays.
- [**Level 4 — global property type matching**](https://github.com/livewire/livewire/commit/3f3251d4d12d5b58f08de8b3f54787804e101cce): extracts the policy into a documented global utility with a dedicated type matrix, then makes `BaseUrl` select between original and decoded values through that utility.

## Verification

- `vendor/bin/phpunit --testsuite Unit` — 1,404 tests, 3,985 assertions
- `vendor/bin/phpunit src/Features/SupportQueryString/BrowserTest.php` — 36 tests, 288 assertions

### DIFF
--- a/src/Drawer/BaseUtils.php
+++ b/src/Drawer/BaseUtils.php
@@ -147,4 +147,60 @@ class BaseUtils
 
         return $property->hasType() && (! $property->isInitialized($target));
     }
+
+    /**
+     * Compare a value's native PHP type with a property's declared type.
+     *
+     * This deliberately does not apply PHP's weak scalar coercion. It returns
+     * null when the property is missing or untyped because there is no declared
+     * type contract to compare the value against.
+     */
+    static function propertyTypeMatchesValue($target, $property, $value): ?bool {
+        if (! $property || ! property_exists($target, $property)) return null;
+
+        $reflectionProperty = static::getProperty($target, $property);
+
+        if (! $type = $reflectionProperty->getType()) return null;
+
+        return static::typeMatchesValue($type, $value, $reflectionProperty->getDeclaringClass());
+    }
+
+    protected static function typeMatchesValue($type, $value, $declaringClass): bool {
+        if ($value === null) return $type->allowsNull();
+
+        if ($type instanceof \ReflectionUnionType) {
+            foreach ($type->getTypes() as $unionType) {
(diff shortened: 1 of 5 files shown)


people-sea:
@calebporzio thanks！

PR #10618, opened 2026-08-28
[4.x] Persist dirty state until explicitly marked clean

calebporzio:
Builds on the underlying need identified in #10600 with a baseline that distinguishes local edits from clean server-originated changes.

## Problem

- **What does the user experience?** An unsaved-changes indicator powered by `wire:dirty` appears while typing, then disappears after any unrelated message such as `wire:poll`, even though the save action never ran.
- **What did they expect?** The indicator should remain dirty until the edited state is explicitly accepted as saved.
- **What caused the difference?** `wire:dirty` compares the browser's reactive state to `component.canonical`, and every successful server response replaces `canonical`. It therefore measures “has the server received this state?” rather than “has this state been saved?”

## Solution

This adds an opt-in persistent baseline through `wire:dirty.persist` and `$dirty(..., { persist: true })`. Persistent baselines survive unrelated messages, absorb clean server-originated changes, and advance only when the application explicitly calls `markAsClean()`.

The implementation is deliberately staged so each commit is a workable level of the solution:

1. **[Level 1 — persistent baseline](https://github.com/livewire/livewire/commit/d6d6f9edbaf13bee117392dd6e4b5ececfcdb6b3):** Adds `.persist`, the `$dirty` option, and `$wire.$markAsClean()` as the smallest client-side expression of the idea.
2. **[Level 2 — reconcile clean server changes](https://github.com/livewire/livewire/commit/573b981fc6df6b37f019e8434864cc9ebb051ce2):** Advances clean portions of the baseline during snapshot merges, so polling and server-owned sibling updates do not become false unsaved edits.
3. **[Level 3 — authoritative save checkpoint](https://github.com/livewire/livewire/commit/19903f776cd02573359ea6493ca5ef828ee951a0):** Adds `$this->markAsClean()`, applies the returned canonical snapshot as the saved checkpoint, preserves newer in-flight browser edits, and documents the API.

## Other paths considered

- **Minimum:** Keep an initial client snapshot and reset it manually. Small, but server changes become false positives and successful saves have no authoritative checkpoint.
- **Maximum:** Introduce a first-class edit-session tracker with path-level save acknowledgements, partial saves, and concurrency metadata. Complete, but much larger than this opt-in feature needs.
- **Evaporative:** Move persistence semantics into a dedicated form transaction API that automatically owns save boundaries. Harmonious, but requires reconstructing the form/action model.
- **Userland:** Store selected initial values in Alpine, compare them manually, and reset after a successful save event. No framework change, but every application must rebuild lifecycle and concurrency handling.

## Verification

- 114 JavaScript tests pass (1 existing skipped), including four new baseline reconciliation cases.
- Focused server-effect unit test passes (3 assertions).
- Four focused browser tests pass (16 assertions), covering unrelated messages, clean server changes, server-side checkpoints, and a newer edit made while save is in flight.
- Both Livewire browser bundles build successfully.
- A real Laravel playground confirms clean polls stay clean, edits remain dirty across polls, save clears the indicator, and the following poll remains clean.

### DIFF
--- a/docs/wire-dirty.md
+++ b/docs/wire-dirty.md
@@ -3,9 +3,9 @@ In a traditional HTML page containing a form, the form is only ever submitted wh
 
 However, Livewire is capable of much more than traditional form submissions. You can validate form inputs in real-time or even save the form as a user types.
 
-In these "real-time" update scenarios, it can be helpful to signal to your users when a form or subset of a form has been changed, but hasn't been saved to the database.
+In these "real-time" update scenarios, it can be helpful to signal when a form or subset of a form has changed but hasn't reached the server.
 
-When a form contains un-saved input, that form is considered "dirty". It only becomes "clean" when a network request has been triggered to synchronize the server state with the client-side state.
+When the browser contains input that differs from the server state, that input is considered "dirty". It becomes "clean" when a message synchronizes the browser and server state.
 
 ## Basic usage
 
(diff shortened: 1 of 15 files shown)


calebporzio:
Closing this exploration after deciding to keep saved/unsaved semantics in userland. The framework can precisely report client/server synchronization, while durable persistence boundaries, partial saves, and application-owned field groups belong to the application.

The userland recommendation and concurrency-safe snippet are posted on #10600: https://github.com/livewire/livewire/pull/10600#issuecomment-5462407218

PR #10620, opened 2026-08-28
[4.x] Avoid unnecessary validation data unwrapping

calebporzio:
## Problem

`validateOnly()` gets slower as a form grows, even though it validates a single field: every top-level public property — plain scalars and arrays included — is pushed through `findSynth()` during validation prep. For string values this routes through `findByKey()`, costing ~15µs per property *per call*, so large scalar forms pay a real tax on every keystroke validation (#9416, and the dominant cost in #10565's report).

## Findings

- Scalars and arrays are already valid Laravel validation data — the synth lookup on them is pure waste. No built-in synth matches a non-object value, so nothing is skipped that could have mattered.
- Skipping it also fixes a latent fatal: a string property whose *value* equals a synth key (e.g. `'wrbl'`) was routed through `findByKey()`, resolved the Wireable synth, and fataled calling `->toLivewire()` on a string.

## Potential solutions

- **Bail early for non-objects** — two lines, cannot change validation results, captures the reported cost.
- **Also keep `Arrayable`/`ArrayAccess` siblings lazy during `validateOnly()`** — explored and rejected: dependent rules that reference a wrapped sibling break silently (wildcard parameters never match objects through `Arr::dot()`, whole-object `same:`/`gt:` comparisons fail on type mismatch, model siblings leak live cast/hidden/lazy-load semantics), and guarding all of that requires a rule-parameter–scanning heuristic that isn't worth its weight.

## Proposed solution

The non-object bail alone:

```php
// Scalars and arrays are already valid Laravel validation data...
if (! is_object($value)) return $value;
```

Objects take exactly the same path as before — synth `unwrapForValidation()`, then `Arrayable::toArray()` — so validation behavior is byte-for-byte identical; the only observable change is the `'wrbl'` fatal becoming a passing validation.

## Verification

- 10,000 preparation passes over 38 scalar fields: **5,385ms → 17ms**. Re-measured through a real browser round trip (`wire:model.live.blur` → `updated()` → `validateOnly()`) on a 16-property component: 1,000 unwrap passes dropped **624ms → 1.1ms**.
- New tests pin the fast path (a synth spy proving scalars never hit `findSynth()`), the eager-sibling semantics any future laziness attempt must preserve (wildcard `in_array:options.*.id`, whole-Collection `same:`, enum-cast model `required_if`, `withValidator()` + `sometimes()`, and a negative-path dependent rule), and the synth-key string fatal.
- Full unit suite: 1,432 tests, 4,068 assertions; 13 skipped, 3 incomplete. Branch is merged with the current `4.x` head.

No documentation change is needed: the public validation API is unchanged.

### DIFF
--- a/src/Features/SupportValidation/HandlesValidation.php
+++ b/src/Features/SupportValidation/HandlesValidation.php
@@ -512,6 +512,8 @@ trait HandlesValidation
     protected function unwrapDataForValidation($data)
     {
         return collect($data)->map(function ($value) {
+            // Scalars and arrays are already valid Laravel validation data...
+            if (! is_object($value)) return $value;
 
             $synth = app('livewire')->findSynth($value, $this);
 
--- a/src/Features/SupportValidation/HandlesValidation.php
+++ b/src/Features/SupportValidation/HandlesValidation.php
@@ -2,7 +2,6 @@
 
 namespace Livewire\Features\SupportValidation;
 
-use ArrayAccess;
 use function Livewire\invade;
 use function Livewire\store;
 use Illuminate\Contracts\Support\Arrayable;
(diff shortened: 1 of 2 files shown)


ghabriel25:
@calebporzio just throwing a reference here for dependant rules https://github.com/laravel/framework/blob/13.x/src%2FIlluminate%2FValidation%2FValidator.php#L690

PR #10628, opened 2026-08-29
[4.x] Fix #[Session] non-primitives under JSON sessions

calebporzio:
## Decision brief

Fresh Laravel applications now default to JSON session serialization ([laravel/laravel@75cef50](https://github.com/laravel/laravel/commit/75cef503c6edc3447dd79053a648ea981857e15b)). That makes `#[Session]` fail for typed non-primitive properties: a `Collection` survives Livewire updates through the snapshot synth pipeline, but after a full page refresh Laravel reads the JSON session value as a plain array and PHP rejects assigning it to `?Collection`.

```text
Cannot assign array to property ...::$list of type ?Illuminate\Support\Collection
```

Fixes #10250. Supersedes #10386.

### Demand and product call

- Confirmed demand is one reporter and one explicit beneficiary, with no duplicate reports or independent “me too” comments found.
- The affected surface is broader than that count: Laravel's new default exposes every new app that combines `#[Session]` with Collection, Carbon, enum, model, nested, or custom synthesized values.
- This should be fixed in core rather than documented away. The complete userland escape hatch is switching session serialization back to `php`, but that undoes Laravel's security-motivated default. Array-backed state is a safe narrow workaround, not a satisfactory `#[Session]` contract.

### Prior decisions

- #10252 was the original fix from @joshhanley. Review identified three constraints: don't create a careless hydration pathway, don't leak `HandleComponents` internals, and don't add cost or behavior changes to PHP-serialized sessions.
- #10276 extracted and secured `HandleSynths` specifically so features like `#[Session]` can reuse the registered synth pipeline without reaching into `HandleComponents`.
- #10630 added the missing storage-facing abstraction over `HandleSynths`: an authenticated, versioned `PersistedValueCodec`. It is now merged, and this PR builds directly on that core abstraction.
- #10386 applied that mechanism, but stored bare synth tuples. Any pre-existing session array shaped like `[data, ['s' => ...]]` could therefore be mistaken for Livewire metadata.
- Laravel solved its analogous JSON-session `ViewErrorBag` type loss with a narrowly owned codec in [framework #42090](https://github.com/laravel/framework/pull/42090). This PR follows the same boundary while reusing Livewire's general synth registry.

## Solution

Under `session.serialization=json`, `BaseSession` delegates its value to the `PersistedValueCodec` introduced in #10630. The codec decides whether synthesis is necessary and owns recursive traversal, component context, the strict versioned envelope, authentication, and synth security checks.

`BaseSession` now knows only that JSON sessions require storage encoding. It no longer understands synth tuples or calls `hydrate()` directly. Reads are decoded regardless of the current serializer so configuration changes do not strand encoded state; corrupt or stale values are forgotten and the property's default is preserved.

Consequences:

- Collections, Carbon, nested synthesized values, and registered custom synths round-trip through JSON sessions.
- JSON primitives remain raw.
- PHP-serialized sessions remain raw and pay no synth cost.
- Legacy tuple-shaped arrays remain ordinary arrays.
- Every encoded envelope is signed with the application key and scoped to its resolved session key.
- Client-authored, tampered, unsupported-version, and wrong-scope envelopes are rejected before synth hydration.
- Components intentionally using the same explicit session key can continue to share values.
- Authentic values for removed synthesizers are discarded safely instead of permanently failing every request.
- Class metadata still passes through `SecurityPolicy::validateClass()` before any synth can instantiate it.
- No `unserialize()` path is introduced.
- Forging a decodable persisted value requires `APP_KEY`, the same trust failure that permits forging Livewire snapshot checksums.

## Progressive implementation

Each commit is a usable stage rather than a disconnected piece:

The prerequisite first makes the change easy in [#10630](https://github.com/livewire/livewire/pull/10630). This PR then builds the user-facing fix in usable stages:

1. 🛹 [Fix session-backed Collections under JSON serialization](https://github.com/livewire/livewire/commit/74cdc617) — `BaseSession` delegates to the authenticated codec and solves the reporter's exact typed-Collection refresh failure.
2. 🛴 [Persist every synthesized value in JSON sessions](https://github.com/livewire/livewire/commit/b9151770) — proves the same path for Carbon and nested synthesized values.
3. 🚲 [Lock in session compatibility and security boundaries](https://github.com/livewire/livewire/commit/f823d846) — proves raw PHP/primitives, explicit-key sharing, legacy tuple safety, stale-state recovery, and rejection of client-authored model envelopes under both serializers.
4. 🚗 [Complete JSON session compatibility across the browser lifecycle](https://github.com/livewire/livewire/commit/762dd09c) — preserves JSON-native arrays, proves registered custom synths and Eloquent models, asserts JSON mode, and exercises add → full refresh → add.

## Reproduction and verification

Confirmed before the fix on the reporter's exact Livewire `v4.2.4` (`7d0bfa46`):

1. Load the component successfully.
2. Click **Add**; `1` renders.
3. Refresh the page; Laravel returns HTTP 500 with the typed-property error above.

Confirmed after the fix in the same Laravel 13 playground:

1. Click **Add**; `1` renders.
2. Refresh; `1` is restored with no error.
3. Click **Add** again; `1, 2` renders.

Automated verification:

- Persisted-value codec: 12 tests, 19 assertions.
- SupportSession feature suite: 16 tests, 20 assertions.
- Full unit suite: 1,504 tests, 4,251 assertions; 13 existing skips and 3 existing incomplete tests.
- Focused browser regression: 1 test, 5 assertions.
- Real Laravel 13 playground: `1` after update → `1` after full refresh → `1, 2` after the next update.
- Independent review reproduced an unsigned-envelope `ModelSynth` exploit against the first design; the final signed design directly tests and rejects that path under JSON and PHP sessions.

Session-backed Eloquent values retain normal `ModelSynth` lazy-loading semantics; this PR does not force model queries during `BaseSession::mount()`.

## Immediate userland workaround

Apps unable to upgrade can keep JSON-native data in the session and derive a Collection:

```php
#[Session]
public array $listData = [];

#[Computed]
public function list(): Collection
{
    return collect($this->listData);
}
```

This works for JSON-safe items but does not preserve arbitrary nested object types.

Implementation credit: this is rebuilt from @joshhanley's original work in #10252; Josh is credited as a co-author in the first commit.

### DIFF
--- a/src/Features/SupportSession/BaseSession.php
+++ b/src/Features/SupportSession/BaseSession.php
@@ -29,9 +29,13 @@ class BaseSession extends LivewireAttribute
             return;
         }
 
+        // A stored value can become unassignable after it was persisted: its
+        // type no longer matches the property (\TypeError), or it's a scalar
+        // backing a since-removed enum case (\ValueError from Enum::from()).
+        // Either way the stale value is forgotten and the default preserved.
         try {
             $this->setValue($fromSession);
-        } catch (\TypeError) {
+        } catch (\TypeError|\ValueError) {
             Session::forget($this->key());
         }
     }
--- a/src/Features/SupportSession/BaseSession.php
+++ b/src/Features/SupportSession/BaseSession.php
@@ -3,12 +3,16 @@
 namespace Livewire\Features\SupportSession;
 
 use Livewire\Features\SupportAttributes\Attribute as LivewireAttribute;
+use Livewire\Mechanisms\HandleSynths\CorruptPersistedValueException;
+use Livewire\Mechanisms\HandleSynths\PersistedValueCodec;
 use Illuminate\Support\Facades\Session;
 use Attribute;
 
 #[Attribute(Attribute::TARGET_PROPERTY)]
 class BaseSession extends LivewireAttribute
 {
+    protected $codec;
+
     function __construct(
         protected $key = null,
     ) {}
@@ -17,9 +21,19 @@ class BaseSession extends LivewireAttribute
     {
         if (! $this->exists()) return;
 
-        $fromSession = $this->read();
+        try {
+            $fromSession = $this->read();
+        } catch (CorruptPersistedValueException) {
+            Session::forget($this->key());
 
-        $this->setValue($fromSession);
+            return;
+        }
+
+        try {
+            $this->setValue($fromSession);
+        } catch (\TypeError) {
+            Session::forget($this->key());
+        }
     }
 
     public function dehydrate($context)
@@ -34,12 +48,43 @@ class BaseSession extends LivewireAttribute
 
     protected function read()
     {
-        return Session::get($this->key());
+        // Always decode so changing session serializers doesn't strand values
+        // that Livewire encoded before the configuration changed.
+        $key = $this->key();
+
+        return $this->codec()->decodeFromStorage(
+            Session::get($key),
+            $this->component,
+            $this->getName(),
+            $key,
+        );
     }
 
     protected function write()
     {
-        Session::put($this->key(), $this->getValue());
+        $value = $this->getValue();
+        $key = $this->key();
+
+        if ($this->sessionIsJsonSerialized()) {
+            $value = $this->codec()->encodeForStorage(
+                $value,
+                $this->component,
+                $this->getName(),
+                $key,
+            );
+        }
+
+        Session::put($key, $value);
+    }
+
+    protected function sessionIsJsonSerialized()
+    {
+        return config('session.serialization', 'php') === 'json';
+    }
+
+    protected function codec()
+    {
+        return $this->codec ??= app(PersistedValueCodec::class);
     }
 
     protected function key()
--- a/src/Features/SupportSession/JsonSerializationBrowserTest.php
+++ b/src/Features/SupportSession/JsonSerializationBrowserTest.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Livewire\Features\SupportSession;
+
+use Illuminate\Support\Collection;
+use Livewire\Attributes\Session;
+use Livewire\Component;
+use Livewire\Livewire;
+use Tests\BrowserTestCase;
+
+class JsonSerializationBrowserTest extends BrowserTestCase
+{
+    public static function tweakApplicationHook()
+    {
+        return function () {
+            config(['session.serialization' => 'json']);
+        };
+    }
+
+    public function test_collection_properties_survive_a_full_page_refresh()
+    {
+        Livewire::visit(new class extends Component {
+            #[Session]
+            public ?Collection $list = null;
+
+            public function mount(): void
+            {
+                $this->list ??= collect();
+            }
+
+            public function add(): void
+            {
+                $this->list->push($this->list->count() + 1);
+            }
+
+            public function render()
+            {
+                return <<<'HTML'
+                <div>
+                    <span dusk="serialization">{{ config('session.serialization') }}</span>
+                    <button dusk="button" wire:click="add">Add</button>
+                    <span dusk="list">{{ $list->implode(', ') }}</span>
+                </div>
+                HTML;
+            }
+        })
+            ->assertSeeIn('@serialization', 'json')
+            ->assertSeeIn('@list', '')
+            ->waitForLivewire()->click('@button')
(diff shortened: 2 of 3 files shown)


calebporzio:
## Review

Reviewed this together with the codec it builds on (#10630). Set up a fresh Laravel 13 app (which now defaults to `session.serialization=json`) with this branch and verified the whole story end‑to‑end.

**Verdict: solid, ship-able.** The bug is real, the fix works, and the security design holds up under adversarial probing.

### Verified
- **Bug reproduces** exactly as described — a `#[Session] ?Collection` throws `TypeError: Cannot assign array to property …$list of type ?Collection` (HTTP 500) on a full page refresh with the pre-fix code.
- **Fix works in a real browser** — Collection *and* a session-backed `Carbon` both survive a full refresh and keep accumulating, no 500.
- **Security core is sound.** I specifically tried to forge a decodable envelope without `APP_KEY` via the client‑writable `#[Session]` property path (the exact threat model this design targets) and could not: the HMAC signature gate rejects unsigned/tampered/wrong‑version/wrong‑scope envelopes before `HandleSynths::hydrate()` is ever reachable, and `SecurityPolicy::validateClass()` sits behind it as defense‑in‑depth. Sign/verify is also byte‑stable across the JSON round‑trip (floats, exponents, non‑sequential keys, slashes, unicode, big ints, negative zero) — no legitimate value fails its own signature, so no silent data loss from that path.
- Codec + session unit suites (29 tests) and the browser test pass locally; CI is green.

### Fix I pushed to this branch (`b108b173`)
The PR's own contract is *"corrupt or stale values are forgotten and the property's default is preserved."* `mount()` honored that for `CorruptPersistedValueException` and `\TypeError`, but **not for a stale enum backing value**: restoring a `#[Session] ?SomeEnum` whose stored scalar no longer maps to a case coerces through `Enum::from()`, which throws `\ValueError` (sibling of `\TypeError`, not a subclass) at `Attribute.php:76` — inside `mount()`'s `try`, but past its `catch`. Result: a persistent 500 on every page load, session key never forgotten. Widened the catch to `\TypeError|\ValueError` and added a regression test (`test_stale_enum_backing_values_are_discarded_from_typed_properties`). Confirmed it fails without the fix and passes with it.

### One design call worth making (not fixed — it's your call)
A `#[Session] ?Model $post` whose row is **deleted out of band** has the same "sticky 500" shape, but for a different reason and it can't be fixed the same way. On PHP 8.4+, `ModelSynth::hydrate()` returns an *uninitialized lazy proxy*, so `decodeFromStorage()` succeeds and `mount()`'s guards see nothing wrong — the `firstOrFail()` fires later at first property access (usually Blade render), **outside** any mount guard → `ModelNotFoundException` on every load, key never forgotten. (On PHP <8.4 the proxy resolves eagerly inside the codec's `catch`, so it degrades gracefully — a version‑dependent divergence.)

I deliberately did **not** fix this, because:
- It only affects `#[Session]` on a whole model, which the docs already call a *"Bad use"* (they steer you to store the ID).
- It's not a regression against any shipping behavior — non‑primitive JSON sessions never worked before this PR.
- The only real fix is to force‑initialize lazy proxies inside the codec's guarded decode, which **reverses this PR's deliberate choice** to keep normal lazy‑loading semantics and *"not force model queries during `BaseSession::mount()`."*

So it's a genuine trade‑off between graceful recovery and lazy semantics for a discouraged usage. If you want the recovery, initializing lazy proxies in `PersistedValueCodec::decodeFromStorage()` (inside the existing `try`) does it and also erases the 8.3/8.4 divergence. Otherwise a one‑line note in the docs that whole‑model session properties don't self‑heal from deleted rows would close the gap.

### Minor, optional (didn't push — matters of taste)
- The `signatureFor()` domain string + storage‑key binding are the load‑bearing security choices in the file but carry no comment on *why*.
- `PersistedValueCodec`'s `$key` param (the storage key) is easy to conflict with the HMAC key in a file whose whole job is HMAC — `$storageKey` would read clearer.
- `CorruptPersistedValueException::__construct($previous)` could type the param as `?\Throwable`.
- `write()` could use the twin of `read()`'s comment explaining why encoding is json‑only while decoding is always‑on.

Everything else — the envelope strictness, the "always decode so switching serializers doesn't strand values" choice, and the altitude of extracting the codec as its own class — I probed and came away agreeing with. Nice work.

PR #10629, opened 2026-08-29
[4.x] Fix wire transition skipped after flux toast auto-dismissed

### LINKED ISSUE #2783: `wire:transition` stops working after Livewire actions using Flux::toast()
### Flux version

v2.17.1

### Livewire version

v4.4.2

### Tailwind version

v4.3.3

Firefox 154.0 / Windows 11

### What is the problem?

After updating Livewire to v4.4.2, `wire:transition` no longer works correctly when a Livewire action calls `Flux::toast()`.

The issue appears to be related to the changes introduced to fix Flux #2740 / Livewire #10552.

When an action calls `Flux::toast()`, the expected `wire:transition` animation is skipped. More importantly, in my application, subsequent Livewire actions that do NOT call `Flux::toast()` can also have their `wire:transition` skipped.

The original #2740 issue was about the toast being visually covered by a View Transition. However, the current behavior appears to affect the transition itself, rather than only the toast's visual stacking.

This worked correctly before updating to Livewire v4.4.2.

.... 

    ...

`wire:transition` should continue to work normally for Livewire actions that update the DOM.

When `Flux::toast()` is triggered, the toast should be displayed correctly without disabling or skipping unrelated `wire:transition` animations.

In particular, an action that opens a Flux toast should not cause transitions from subsequent Livewire actions to be skipped.

The fix for Flux #2740 should prevent the toast from being covered by the View Transition, while normal `wire:transition` behavior should remain unchanged for actions that do not involve top-layer UI.

ghabriel25:
### Problem

`wire:transition` currently skips a View Transition whenever there is an open modal dialog or popover:

```js
if (document.querySelector('dialog:modal, :popover-open')) {
    return callback()
}
```

This is necessary for open popovers because View Transition pseudo-elements can paint above the top layer. Starting a transition while a popover is open can cause the transition to render over the popover.

However, `:popover-open` only tells us that the popover itself is open. It does not tell us whether the popover currently has any **visible UI**.

Some components keep a popover open as a persistent container while its contents are hidden. In that case, `:popover-open` still matches even though there is nothing visible that needs to be protected from the View Transition.

This causes subsequent Livewire transitions to be skipped unnecessarily.

### Solution

Keep the existing behavior for modal dialogs, but for popovers only skip the transition when the popover contains a visible element:

```js
if (document.querySelector('dialog:modal')) return callback()

for (let popover of document.querySelectorAll(':popover-open')) { 
    for (let el of popover.querySelectorAll('*')) { 
        if (el.checkVisibility()) return callback() 
    } 
}
```

`checkVisibility()` lets us determine whether the open popover actually contains visible UI without relying on any framework- or component-specific selectors.

This keeps the behavior generic and avoids coupling Livewire's transition implementation to a particular component library or DOM structure.

### Why not just check `:popover-open`?

An open popover can remain mounted and open while its contents are hidden. Treating every `:popover-open` element as blocking would therefore incorrectly skip transitions when there is no visible popover.

At the same time, removing the `:popover-open` check entirely would reintroduce the issue where an actually visible popover can be covered by View Transition pseudo-elements.

The new check handles both cases:

* **Visible popover** → skip the transition.
* **Open but visually inactive popover** → allow the transition.
* **Modal dialog** → skip the transition as before.

### Why `checkVisibility()`?

Checking text content, child count, or specific element names would be unreliable:

* A popover can contain non-textual UI.
* A popover can contain DOM nodes that are hidden.
* Empty/non-empty DOM structure does not necessarily correspond to visible UI.
* Component-specific selectors would couple Livewire to another package's implementation details.

`checkVisibility()` directly answers the relevant question: whether there is currently a visible element inside the open popover.

Fixes: https://github.com/livewire/flux/issues/2783

### DIFF
--- a/js/directives/wire-transition.js
+++ b/js/directives/wire-transition.js
@@ -29,23 +29,6 @@ function clearTransitionNames(root) {
     })
 }
 
-function hasBlockingTopLayer() {
-    if (document.querySelector('dialog:modal')) return true
-
-    for (let el of document.querySelectorAll(':popover-open')) {
-        // Flux toast hosts: only block while a real toast panel is present
-        if (el.localName === 'ui-toast' || el.localName === 'ui-toast-group') {
-            if (el.querySelector('[data-flux-toast-dialog]')) return true
-            continue // open host, no panel → do not block transitions
-        }
-
-        // Other popovers (menus, dropdowns, etc.): still block
-        return true
-    }
-
-    return false
-}
-
 export async function transitionDomMutation(fromEl, toEl, callback, options = {}) {
     // Skip transitions entirely if requested...
     if (options.skip) return callback()
@@ -61,7 +44,13 @@ export async function transitionDomMutation(fromEl, toEl, callback, options = {}
     // Skip entirely if a top-layer element is already open (transitions behind
     // it are invisible to the user and the ::view-transition pseudo-elements
     // would paint above it during animation)...
-    if (hasBlockingTopLayer()) return callback()
+    if (document.querySelector('dialog:modal')) return callback()
+
+    for (let el of document.querySelectorAll(':popover-open')) {
+        if ([...el.querySelectorAll('*')].some(el => el.checkVisibility())) {
+            return callback()
+        }
+    }
 
     // Set transition names right before the transition starts (not permanently)...
     setTransitionNames(fromEl, options)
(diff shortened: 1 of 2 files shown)


calebporzio:
thanks!

PR #10635, opened 2026-08-31
[4.x] Fix skip render always fallback to default layout

ghabriel25:
### Summary

This PR fixes an issue where page components that call `skipRender()` always fall back to the default layout, even when the component defines its own layout using a view `->layout()` macro or the `#[Layout]` attribute.

The problem occurs because `skipRender()` bypasses the normal `render` event, which is currently responsible for allowing the page-component layout handler to inspect the component's rendered view and capture its layout configuration.

This change ensures the component's view is resolved before handling `skipRender()` and introduces a dedicated `skip.render` event so the page-component layout configuration can still be captured when rendering is skipped.

### Problem

When a page component calls `skipRender()` during its lifecycle, Livewire skips the normal rendering flow.

For example:

```php
class MyPage extends Component
{
    public function mount()
    {
        $this->skipRender();
    }

    public function render()
    {
        return view('my-page')
            ->layout('layouts.app');
    }
}
```

Previously, because the `render` event was never triggered for the skipped render, the page-component layout interceptor never had an opportunity to capture the layout configuration.

As a result, the page component could incorrectly fall back to the default layout instead of using the layout explicitly configured by the component.

The same issue affected layout and title attributes.

#### 1. Introduce a `skip.render` event

When `skipRender()` is active, Livewire now triggers a `skip.render` event with the component and resolved view.

This allows consumers that need to inspect render metadata to behave consistently whether the component is actually rendered or skipped.

The existing `render` event remains unchanged for the normal rendering path.

#### 2. Capture layout configuration during skipped renders

The page-component layout interceptor now listens to:

* `render`
* `render.placeholder`
* `skip.render`

This ensures layouts and titles are captured regardless of whether the component produces normal HTML or skips its render.

### Expected behavior

With this change, a page component such as:

```php
class MyPage extends Component
{
    public function mount()
    {
        $this->skipRender();
    }

    public function render()
    {
        return view('my-page')
            ->layout('layouts.custom');
    }
}
```

will continue to use `layouts.custom` rather than falling back to the application's default layout.

### Tests

Added regression coverage for the different ways a page component can configure its layout while skipping rendering:

* Layout configured through the view `->layout()` macro.
* Layout configured through the `#[Layout]` attribute.
* Title configured through the `#[Title]` attribute.

The tests verify that the configured layout/title information is still respected when `skipRender()` is called.

All existing behavior for normal rendering remains unchanged.

### DIFF
--- a/src/Features/SupportPageComponents/SupportPageComponents.php
+++ b/src/Features/SupportPageComponents/SupportPageComponents.php
@@ -91,7 +91,7 @@ class SupportPageComponents extends ComponentHook
 
         // Only run this handler once for the parent-most component. Otherwise child components
         // will run this handler too and override the configured layout...
-        $handler = once(function ($target, $view, $data) use (&$layoutConfig, &$slots) {
+        $handler = once(function ($target, $view, $data = []) use (&$layoutConfig, &$slots) {
             $layoutAttr = $target->getAttributes()->whereInstanceOf(BaseLayout::class)->first();
             $titleAttr = $target->getAttributes()->whereInstanceOf(BaseTitle::class)->first();
 
@@ -114,12 +114,14 @@ class SupportPageComponents extends ComponentHook
 
         $render = on('render', $handler);
         $renderPlaceholder = on('render.placeholder', $handler);
+        $skipRender = on('skip.render', $handler);
 
         try {
             $callback();
         } finally {
             $render();
             $renderPlaceholder();
+            $skipRender();
         }
 
         return $layoutConfig;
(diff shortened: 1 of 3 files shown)


calebporzio:
i don't know about this. what's confusing so me is:

A) why would someone skipRender in mount? what would get rendered?
B) if render was skipped i would expect the render method to not run...

and this adds new semantics like a skip.render event and i try to avoid adding mechanisms like that especially for such a small edge case.

going to close this as i have hangups and don't see demand for it.

re-open with explanation and better case if I'm wrong. THANKS

calebporzio:
Ahhh i see that there is an issue #10597 - next time link to it in the PR description thanks!

I still feel the same about the PR though lol and have the same questions.

ghabriel25:
Yes, I think that issue because `render` event not fired. Honestly, I do agree that it shouldnt have new mechanism like that.

Layout configuration required render method called, but that seems awkward if skip render calls render.

PR #10645, opened 2026-08-31
[4.x] Only mark a lazy component loaded once it has actually resumed

### LINKED ISSUE #10643: dehydrate() marks a lazy component as loaded without it having mounted, so a later __lazyLoad renders it unmounted
### Livewire version

v3.8.7 (logic is identical on 4.x — see below)

### Laravel version

v13.29.0

### PHP version

8.5.8

Not browser-specific — reproduced in a PHPUnit test. Originally observed in Chrome on macOS.

`SupportLazyLoading::dehydrate()` writes `lazyLoaded: true` for a component that has never mounted, and since #10586 a later `__lazyLoad` against that snapshot is swallowed and the component renders with every property at its class default.

The mismarking comes from `hydrate()`, which sets `isLazyLoadHydrating` whenever the memo says `lazyLoaded === false` — that is, on *any* request that hydrates a not-yet-loaded lazy component, whatever the request is for. `dehydrate()` then treats that flag as "finished loading". Nothing in the path checks whether `callMountLifecycleMethod()` actually ran.

**This is partly already known and partly already fixed.** #10589 (v3.8.7, backporting #10221 on 4.x) closed the trigger I originally hit: a lazy placeholder no longer registers its event listeners, so a dispatched event can't reach it before it resumes. That PR's own description names the consequence —

> Worse, that request's `dehydrate()` stamps `lazyLoaded: true`, so the component is marked resumed without ever having mounted and stays wrong until a page reload.

— but the fix is in `SupportEvents`, and the stamping in `SupportLazyLoading::dehydrate()` is unchanged. I've confirmed on v3.8.7 that the `listeners` effect is no longer emitted for a placeholder, and that a request which does reach the component still flips the memo and still renders it unmounted.

So this report is about the residual: whether the mismarking itself is worth hardening now that the known route to it is shut, or whether #10589 is considered sufficient. I don't have a browser-reachable trigger on v3.8.7 to offer — if you'd rather wait for one, closing this is fair.

Filing it because the failure mode is silent and the guard looks cheap: a component whose `render()` tolerates default state just renders wrong, and one whose `mount()` initialises something `render()` dereferences throws. A typed property assigned in `mount()` gives "must not be accessed before initialization"; an array built in `mount()` throws on its first key access. The docs' own `UpdateProfile` example (`$name` / `$email` set in `mount()`) has that shape.

One detail that makes it hard to spot: mount *parameters* survive, because they're applied to matching public properties outside the skipped lifecycle hook. Only mount-*derived* state is missing, so the component looks half-initialised rather than blank.

In the style of `src/Features/SupportLazyLoading/UnitTest.php`:

```php
public function test_a_lazy_component_reached_before_it_loads_still_mounts()
{
    SupportLazyLoading::$disableWhileTesting = false;

    Livewire::component('lazy-gamma', LazyGamma::class);

    $component = Livewire::test('lazy-gamma', ['level' => 5]);

    preg_match("/__lazyLoad\('([^']+)'\)/", html_entity_decode($component->html()), $matches);

    $this->assertFalse($component->snapshot['memo']['lazyLoaded']);

    // Any request that reaches the component before __lazyLoad does.
    $component->dispatch('refresh-panels');

    $component
        ->call('__lazyLoad', $matches[1])
        ->assertSee('level:5')
        ->assertSee('mounts:1');
}
```

```php
#[Lazy]
class LazyGamma extends Component {
    public $level = 0;
    public $mounts = 0;

    public function mount($level = 0) {
        $this->level = $level;
        $this->mounts++;
    }

    #[On('refresh-panels')]
    public function refreshPanels() {}

    public function placeholder() {
        return '<div>Loading...</div>';
    }

    public function render() {
        return '<div>level:'.$this->level.' mounts:'.$this->mounts.'</div>';
    }
}
```

Result on v3.8.6 and v3.8.7: `level:5 mounts:0` — it renders, but `mount()` never ran. On v3.8.4 the same test passes, because `call()` used to resurrect the params and mount unconditionally.

Being straight about what this snippet does and doesn't show: `Livewire::test()->dispatch()` invokes the listener server-side, so it bypasses the client-side listener registration that #10589 now suppresses. It demonstrates the `dehydrate()` mismarking and its consequence; it is not evidence of a browser-reachable path on v3.8.7.

How the consequence changed across versions, from the same mismarked snapshot:

| Version | `call('__lazyLoad')` on a mismarked snapshot | Result |
|---|---|---|
| ≤ 3.8.4 | resurrects the params and mounts unconditionally | correct output; the bad memo is invisible |
| 3.8.5 | declines to intercept (`isLazyLoadHydrating !== true`), no early return | `MethodNotFoundException` — #10570 |
| 3.8.6, 3.8.7 | `isLazyLoadAlreadyLoaded` short-circuits it (#10586) | renders unmounted, silently |

#10586 is right that a genuine repeat call should be swallowed and re-render — its note that "the component still re-renders and the duplicate request simply re-morphs the already-correct output" holds when the snapshot really is post-mount. The gap is only that a snapshot can reach that branch without ever having been mounted.

`hydrate()`, `dehydrate()` and `call()` in `SupportLazyLoading` are identical on `4.x` as of v4.4.3, so if this is worth changing it applies to both branches. I have not run the repro against 4.x — that comparison is by inspection only.

_No response_

`lazyLoaded: true` should mean the component has mounted. I'd expect `dehydrate()` to key off whether the lazy load was actually resumed rather than off `isLazyLoadHydrating` — for example a flag set immediately after `callMountLifecycleMethod($mountParams)` in `call()`, required (along with `isLazyLoadAlreadyLoaded`) before writing the memo. A component that was merely hydrated while unloaded then keeps `lazyLoaded: false`, and a later `__lazyLoad` mounts it normally.

Happy to open a PR for that if you'd take one.

Separately, and a bigger behavioural call than the memo flag: during such a request the component renders at all, with default state, and that render is what the browser morphs in. Keeping it in placeholder mode for any request that isn't `__lazyLoad` may be more correct, and would line up with the "lying dormant until it enters the viewport and is fully loaded" intent quoted in #10589.

lazerg:
## Problem

`SupportLazyLoading::hydrate()` sets `isLazyLoadHydrating` whenever the memo says `lazyLoaded === false`, which is every request that hydrates a lazy component that has not loaded yet, not just the `__lazyLoad` resume. `dehydrate()` then reads that same flag as "finished loading" and writes `lazyLoaded: true`.

Nothing checks whether `callMountLifecycleMethod()` ran. So any request that reaches an unloaded lazy component before `__lazyLoad` does, a dispatched event for instance, stamps the snapshot as loaded. The later `__lazyLoad` is then ignored by the guard added in #10562, and the component renders with `mount()` never having run.

It is easy to miss in the wild because mount params are still applied straight to matching public properties, so the component looks half initialized rather than blank. Anything assigned inside `mount()` is not there: a typed property throws "must not be accessed before initialization", an array property throws on key access.

#10589 removed the original trigger by no longer registering client-side listeners for a placeholder, but the mismarking itself is untouched, and any request that reaches the component still hits it.

### Solution

Key `lazyLoaded: true` off whether the lazy load was actually resumed rather than off `isLazyLoadHydrating`. `call()` sets `isLazyLoadResumed` right after `callMountLifecycleMethod()` runs, and `dehydrate()` requires that flag or the existing `isLazyLoadAlreadyLoaded` check before marking the snapshot loaded.

A component merely hydrated while still unloaded now keeps `lazyLoaded: false` and carries its `lazyIsolated` flag forward, so a later `__lazyLoad` mounts it normally.

Scoped to the memo only. Such a request still renders the component rather than falling back to its placeholder, which is the broader behavioral question raised in the issue and left for a separate call.

Verification:

- `test_a_lazy_component_reached_before_it_loads_still_mounts` dispatches an event at the component before its resume request and asserts the resume still mounts it. Without the change it renders `level:5 mounts:0`
- Lazy-loading unit suite: 11 tests, 26 assertions
- Full unit suite: 1450 tests, no new failures

Fixes #10643

### DIFF
--- a/src/Features/SupportLazyLoading/SupportLazyLoading.php
+++ b/src/Features/SupportLazyLoading/SupportLazyLoading.php
@@ -118,18 +118,22 @@ class SupportLazyLoading extends ComponentHook
         $this->component->skipHydrate();
 
         $this->storeSet('isLazyLoadHydrating', true);
+        $this->storeSet('isLazyIsolated', $memo['lazyIsolated'] ?? false);
     }
 
     function dehydrate($context)
     {
-        if ($this->storeGet('isLazyLoadMounting') === true) {
-            $context->addMemo('lazyLoaded', false);
-            $context->addMemo('lazyIsolated', $this->storeGet('isLazyIsolated'));
-        } elseif (
-            $this->storeGet('isLazyLoadHydrating') === true
+        if (
+            $this->storeGet('isLazyLoadResumed') === true
             || $this->storeGet('isLazyLoadAlreadyLoaded') === true
         ) {
             $context->addMemo('lazyLoaded', true);
+        } elseif (
+            $this->storeGet('isLazyLoadMounting') === true
+            || $this->storeGet('isLazyLoadHydrating') === true
+        ) {
+            $context->addMemo('lazyLoaded', false);
+            $context->addMemo('lazyIsolated', $this->storeGet('isLazyIsolated'));
         }
     }
 
@@ -153,6 +157,8 @@ class SupportLazyLoading extends ComponentHook
         $mountParams = $this->resurrectMountParams($encoded);
 
         $this->callMountLifecycleMethod($mountParams);
+
+        $this->storeSet('isLazyLoadResumed', true);
 
         $returnEarly();
     }
--- a/src/Features/SupportLazyLoading/UnitTest.php
+++ b/src/Features/SupportLazyLoading/UnitTest.php
@@ -182,6 +182,17 @@ class LazyAlpha extends Component {
 }
 
 #[Lazy]
+class LazyBeta extends Component {
+    public function placeholder() {
+        return '<div>Loading...</div>';
+    }
+
+    public function render() {
+        return '<div>beta</div>';
+    }
+}
+
+#[Lazy]
 class LazyGamma extends Component {
     public $level = 0;
     public $mounts = 0;
@@ -200,17 +211,6 @@ class LazyGamma extends Component {
 
     public function render() {
         return '<div>level:'.$this->level.' mounts:'.$this->mounts.'</div>';
-    }
-}
-
-#[Lazy]
-class LazyBeta extends Component {
-    public function placeholder() {
-        return '<div>Loading...</div>';
-    }
-
-    public function render() {
-        return '<div>beta</div>';
     }
 }
 
--- a/src/Features/SupportLazyLoading/UnitTest.php
+++ b/src/Features/SupportLazyLoading/UnitTest.php
@@ -6,6 +6,7 @@ use Illuminate\Support\Facades\Route;
 use Livewire\Attributes\Defer;
 use Livewire\Attributes\Layout;
 use Livewire\Attributes\Lazy;
+use Livewire\Attributes\On;
 use Livewire\Component;
 use Livewire\Exceptions\MethodNotFoundException;
 use Livewire\Livewire;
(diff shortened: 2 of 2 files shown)


lazerg:
Closing since the linked issue #10643 was already resolved.

PR #10647, opened 2026-08-31
[4.x] Don't forward array attributes to a lazy placeholder

### LINKED ISSUE #10646: [4.x] Lazy placeholder crashes on array-valued forwarded attribute since v4.4.3 (trim(): array given)
**Livewire version:** v4.4.3
**Laravel version:** v13.29.0
**PHP version:** 8.5.9
**Browser and operating system:** Chrome on Linux (also reproduces server-side, browser-independent — see repro below)

Upgrading from v4.4.2 to v4.4.3 breaks any lazy component that is mounted with an **array-valued** parameter that isn't a declared public property (i.e. it falls through to Livewire's "extra params become forwarded HTML attributes" bucket). Rendering the lazy placeholder throws:

```
TypeError: trim(): Argument #1 ($string) must be of type string, array given
```

at `Illuminate\View\ComponentAttributeBag::__toString()` (`vendor/laravel/framework/src/Illuminate/View/ComponentAttributeBag.php:502`), the instant the placeholder view renders — before the component itself ever mounts.

### Root cause

Diffing `v4.4.2...v4.4.3` on `src/Features/SupportHtmlAttributeForwarding/SupportHtmlAttributeForwarding.php` shows a new method added in this release:

```php
public function renderPlaceholder($view, $properties)
{
    $attributes = $this->component->getHtmlAttributes();

    $view->with(['attributes' => new ComponentAttributeBag($attributes)]);
}
```

`getHtmlAttributes()` returns whatever mount params didn't match a declared property (`HandleComponents::mount()` → `separateParamsAndAttributes()` → `withHtmlAttributes()`). This is brand new in v4.4.3 — in v4.4.2, `$attributes` was simply never set for a placeholder view at all.

If one of those non-property params is legitimately an **array** (a completely normal thing to pass down to a component — e.g. a filters array, a config array, anything not meant as a literal HTML attribute), it now flows straight into `ComponentAttributeBag`, and `ComponentAttributeBag::__toString()` calls `trim($value)` on every attribute value with no array handling — unlike `class`/`style`, which the bag already knows how to accept as arrays.

Any placeholder view using the common defensive pattern `$attributes ?? new ComponentAttributeBag` (very natural, since `$attributes` used to simply not exist for a placeholder) now receives a populated bag it never asked for, and blows up rendering it — even though the view itself never touches the offending key.

I hit this in a real app via **Filament v5**: every Filament v5 widget is lazy by default, and Filament's dashboard forwards its `pageFilters` array (date range + selected clinic, used by `InteractsWithPageFilters`) as an extra mount param to each widget. That param isn't a declared property on the widget, so it becomes a forwarded HTML attribute, and Filament's `filament::components.loading-section` placeholder view crashes on it under v4.4.3. This isn't Filament-specific though — the minimal repro below doesn't use Filament at all.

```php
// app/Livewire/Counter.php
namespace App\Livewire;

use Illuminate\Contracts\View\View;
use Livewire\Attributes\Lazy;
use Livewire\Component;

#[Lazy]
class Counter extends Component
{
    // Deliberately no declared property matching "pageFilters" below —
    // that's what makes it fall through to "forwarded html attributes"
    // instead of being consumed as a real mount parameter.

    public function placeholder(): View
    {
        return view('counter-placeholder');
    }

    public function render(): View
    {
        return view('counter');
    }
}
```

```blade
{{-- resources/views/livewire/counter.blade.php --}}
<div>Loaded!</div>
```

```blade
{{-- resources/views/counter-placeholder.blade.php --}}
<div {{ ($attributes ?? new \Illuminate\View\ComponentAttributeBag)->class(['loading']) }}>
    Loading...
</div>
```

```blade
{{-- Any page that mounts it with an array-valued extra param: --}}
<div>
    @livewire('counter', ['pageFilters' => ['startDate' => '2026-01-01', 'clinicId' => null]])
</div>
```

Rendering this page throws the `trim(): Argument #1 ($string) must be of type string, array given` error the instant the lazy placeholder renders. On v4.4.2, the exact same code renders the placeholder fine.

I verified this precisely with a Pest test toggling only the `livewire/livewire` version (everything else identical):
- `composer require livewire/livewire:4.4.2` → `view('page')->render()` passes.
- `composer require livewire/livewire:4.4.3` → same code, same test → fails every time with the exact error above.

Passing an array-valued extra param to a lazy component (perfectly valid before v4.4.3) shouldn't crash the placeholder render. `renderPlaceholder()`'s new attribute-forwarding should either skip/ignore non-string-coercible attribute values, or `ComponentAttributeBag::__toString()`/this forwarding path should handle an array value the same way `->class()`/`->style()` already do, rather than newly breaking any existing lazy component + custom placeholder that never opted into this new automatic behavior.

## Please confirm

lazerg:
A lazy component's placeholder crashes with `trim(): Argument #1 ($string) must be of type string, array given` when a non-property mount param holds an array. #10392 started passing forwarded HTML attributes into the placeholder view, and `ComponentAttributeBag::__toString()` has no array handling, so an array-valued param breaks the render even when the placeholder never touches that key. Filament v5 hits this on every lazy widget, since it forwards a `pageFilters` array.

Array values are now skipped when building the placeholder's bag. `->class()` and `->style()` already flatten arrays to strings before they reach the bag, so nothing renderable is lost.

Fixes #10646

### DIFF
--- a/src/Features/SupportHtmlAttributeForwarding/SupportHtmlAttributeForwarding.php
+++ b/src/Features/SupportHtmlAttributeForwarding/SupportHtmlAttributeForwarding.php
@@ -23,7 +23,7 @@ class SupportHtmlAttributeForwarding extends ComponentHook
 
     public function renderPlaceholder($view, $properties)
     {
-        $attributes = $this->component->getHtmlAttributes();
+        $attributes = array_filter($this->component->getHtmlAttributes(), fn ($value) => ! is_array($value));
 
         $view->with(['attributes' => new ComponentAttributeBag($attributes)]);
     }
--- a/src/Features/SupportHtmlAttributeForwarding/UnitTest.php
+++ b/src/Features/SupportHtmlAttributeForwarding/UnitTest.php
@@ -4,7 +4,9 @@ namespace Livewire\Features\SupportHtmlAttributeForwarding;
 
 use Tests\TestCase;
 use Livewire\Livewire;
+use Livewire\Features\SupportLazyLoading\SupportLazyLoading;
 use Livewire\Component;
+use Livewire\Attributes\Lazy;
 
 class UnitTest extends TestCase
 {
@@ -46,5 +48,28 @@ class UnitTest extends TestCase
         ->assertSeeHtml('id="error-alert"')
         ->assertSeeHtml('data-testid="my-alert"')
         ;
+    }
+
+    public function test_array_html_attributes_are_not_forwarded_to_a_placeholder()
+    {
+        SupportLazyLoading::$disableWhileTesting = false;
+
+        Livewire::component('alert', LazyAlert::class);
+
+        $html = Livewire::mount('alert', ['lazy' => true, 'pageFilters' => ['status' => 'active'], 'id' => 'error-alert']);
+
+        $this->assertStringContainsString('id="error-alert"', $html);
+        $this->assertStringNotContainsString('pageFilters=', $html);
+    }
+}
+
+#[Lazy]
+class LazyAlert extends Component {
+    public function placeholder() {
+        return '<div {{ $attributes }}>Loading...</div>';
+    }
+
+    public function render() {
+        return '<div {{ $attributes }}>Alert</div>';
     }
 }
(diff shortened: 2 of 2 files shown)


calebporzio:
closing in favor of #10649  thanks!

PR #10653, opened 2026-09-01
[4.x] Fix wire:navigate breaking when a teleport targets an element inside @persist

### LINKED ISSUE #10652: wire:navigate: teleports inside x-persist are put back while the element is detached
### Livewire version

v4.3.3

### What happens

`putPersistantElementsBack()` in `js/plugins/navigate/persist.js` runs the put-back callback before it re-attaches the persisted element:

```js
old._x_wasPersisted = true;
callback(old, i);                              // unPackPersistedTeleports(old)
Alpine.mutateDom(() => { i.replaceWith(old) });
```

`unPackPersistedTeleports` calls `_x_teleportPutBack()`, which resolves its target through `Alpine`'s `getTarget()`:

```js
document.querySelector(expression)
```

At that point `old` is still detached, so a teleport whose target lives inside the persisted subtree cannot be found. `getTarget()` warns and returns `null`, then `placeInDom` does `null.appendChild(...)`:

```
Alpine Warning: Cannot find x-teleport element for selector: "#some-id"
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'appendChild')
    at placeInDom
    at el._x_teleportPutBack
    at unPackPersistedTeleports
    at putPersistantElementsBack
    at swapCurrentPageWithNewHtml
```

Because the exception escapes `swapCurrentPageWithNewHtml`, `nowInitializeAlpineOnTheNewPage` never runs. The new page is rendered but completely inert: no Livewire, no Alpine, nothing until a manual refresh.

It looks intermittent because it needs the incoming HTML to have no element matching the selector either. In our case the persisted component is `#[Lazy]`, so the destination page only ships its placeholder.

### Reproduction

1. In the layout, persist a lazy component:

```blade
@persist('sidebar')
    <livewire:sidebar />
@endpersist
```

2. Have that component render a teleport pointing at a target inside itself, which is what Mary UI's tabs do:

```blade
<div id="labels"></div>

<template x-teleport="#labels">
    <span>Label</span>
</template>
```

3. Load a page, let the lazy component resolve, then click any `wire:navigate` link.

### Expected

The teleport is put back into the persisted element, or a missing target is skipped rather than thrown through the navigation.

### Note

`unPackPersistedPopovers` is called from the same place, so it resolves against the document under the same conditions.

Swapping the two statements so `i.replaceWith(old)` runs before `callback(old, i)` looks like it would be enough, though I have not tested whether anything else depends on the current order.

lazerg:
Clicking a `wire:navigate` link can leave the destination page completely inert. No Livewire, no Alpine, nothing until a manual refresh. It happens when a persisted element holds an `x-teleport` pointing at a target inside itself and the destination page ships no matching target, which is what you get when the persisted component is `#[Lazy]` and only its placeholder comes down.

```blade
@persist('sidebar')
    <div id="labels"></div>

    <template x-teleport="#labels">
        <span>Label</span>
    </template>
@endpersist
```

`putPersistantElementsBack()` ran its put-back callback before re-attaching the persisted element, so `unPackPersistedTeleports()` resolved teleport targets through `document.querySelector()` while that element was still detached. The target could not be found, `placeInDom` threw on `null.appendChild(...)`, and since the exception escaped `swapCurrentPageWithNewHtml`, `nowInitializeAlpineOnTheNewPage` never ran.

Every persisted element now goes back on the page first and the callbacks run afterwards, so a teleport can also resolve a target that lives inside a different persist. That also mirrors `storePersistantElementsForLater()`, which already packs up while the element is still attached.

Fixes #10652

### DIFF
--- a/src/Features/SupportNavigate/BrowserTest.php
+++ b/src/Features/SupportNavigate/BrowserTest.php
@@ -254,6 +254,61 @@ class BrowserTest extends \Tests\BrowserTestCase
         ;
     }
 
+    public function test_teleports_targeting_an_element_inside_a_persist_are_put_back()
+    {
+        $this->registerComponentTestRoutes([
+            '/second' => new class extends Component {
+                public function render(){ return <<<'HTML'
+                    <div>
+                        <div>
+                            On second page
+                        </div>
+
+                        @persist('header')
+                            <div>Loading...</div>
+                        @endpersist
+                    </div>
+                HTML; }
+            },
+        ]);
+
+        Livewire::visit(new class extends Component {
+            public function render(){
+                return <<<'HTML'
+                    <div>
+                        <div>
+                            On first page
+                        </div>
+
+                        @persist('header')
+                            <div x-data="{ outerScopeCount: 0 }">
+                                <div id="labels"></div>
+
+                                <template x-teleport="#labels">
+                                    <div>
+                                        <span x-text="outerScopeCount" dusk="target"></span>
+                                        <button x-on:click="outerScopeCount++" dusk="button">inc</button>
+                                    </div>
+                                </template>
+                            </div>
+                        @endpersist
+
+                        <a href="/second" wire:navigate dusk="link">Go to second page</a>
+                    </div>
+                HTML;
+            }
+        })
+        ->assertSeeIn('@target', '0')
+        ->click('@button')
+        ->assertSeeIn('@target', '1')
+        ->click('@link')
+        ->waitForText('On second page')
+        ->assertSeeIn('@target', '1')
+        ->click('@button')
+        ->assertSeeIn('@target', '2')
+        ;
+    }
+
     public function test_can_configure_progress_bar()
     {
         $this->browse(function ($browser) {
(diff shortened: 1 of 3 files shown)


lazerg:
The red Legacy L11 / PHP 8.5 job is `test_different_delay_durations`, a timing-sensitive `wire:loading.delay` test. It is unrelated to this change, and it flakes locally as well: it failed on the first run and passed on the next two. The same job passed on the previous commit here.

src/Features/SupportNavigate/BrowserTest.php
```diff
@@ -254,6 +254,122 @@ public function render(){
...
+                            </div>
+                        @endpersist
+
+                        <a href="/second" wire:navigate dusk="link">Go to second page</a>
+                    </div>
+                HTML;
+            }
+        })
+        ->assertSeeIn('@target', '0')
+        ->click('@button')
```
ghabriel25: I'd suggest to use 
```php
->waitForLivewire()->click()
```
just a guard to prevent flaky browser test when directly calling `assertSeeIn()` after click

Except if you're using `waitFor()` after click

Otherwise, this is good fix.

calebporzio:
looks good to me! thanks!

PR #10656, opened 2026-09-02
[4.x] Fix `$wire.$watch` crash from lazy placeholder node conflict

### LINKED ISSUE #10655: `$wire.$watch()` throws on destroy when a lazy component's placeholder shares node positions with its real markup
### Livewire version

v4.4.3

### Laravel version

v13.30.1

### PHP version

8.4.12

Brave on Windows

When a lazily-loaded component's placeholder and its real markup share the same node positions, Livewire's morph reuses that DOM node instead of replacing it. If the real markup's node has an `x-data` calling `$wire.$watch(...)` in its `init()`, the watcher gets created during Alpine's internal "clone" pass, a broken, non-functional watcher reference ends up registered on the component instead of the element.

When the component is later destroyed, calling this broken watcher's cleanup throws:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'stop')
```

This aborts the rest of the component's cleanup, so it never actually gets removed from Livewire's internal registry, `window.Livewire.all()` still lists it after destruction. It causes a memory leak each time the component is recreated and destroyed.

You can directly test it on https://4lr6d.wirebox.app:8443

If you want the code of the wirebox : https://wirebox.app/b/4lr6d

Steps to reproduce : 
1. Toggle the child component
2. Toggle the child component again
3. Open the console and see the error
4. Check how many components stay in window.Livewire.all()

Parent component :
```php
<?php
namespace App\Livewire;

use Livewire\Component;

class Main extends Component
{
    public bool $childShown = false;

    public function render()
    {
        return view('livewire.main');
    }

    public function toggleChild(): void
    {
        $this->childShown = !$this->childShown;
    }
}
```
Parent view :
```blade
<div class="flex flex-col w-full h-full items-center justify-center text-gray-300 px-16">
    <div>
        Parent
    </div>
    <div>
        <button
            class="bg-blue-600 p-2 rounded"
            wire:click="toggleChild"
        >
            Toggle child
        </button>
    </div>
    
    @if ($childShown)
        <div>
            <livewire:child lazy />
        </div>
    @endif
</div>
```

Child component : 
```php
<?php
namespace App\Livewire;

use Livewire\Component;

class Child extends Component
{
    public function placeholder()
    {
        return view('placeholder-child');
    }

    public function render()
    {
        return view('livewire.child');
    }
}
```
Child placeholder : 
```blade
<div class="flex w-full h-full items-center justify-center text-gray-300 px-16">
    Loading...
</div>
```
Child view :
```blade
<div
    x-data="{
        init() {
            $wire.$watch('search', () => {});
        }
    }"
    class="flex w-full h-full items-center justify-center text-gray-300 px-16"
>
    Child
</div>
```

Image

Destroying the component should not throw, and the component should be fully removed from Livewire's internal JS registry (`window.Livewire.all()`).

MaxencePaulin:
## The problem

`$wire.$watch` ties its cleanup to the component, not the element calling it. Alpine's morph "clone" pass runs `x-data` `init()` on a throwaway node before diffing, with its reactivity neutered, this happens whenever a lazy placeholder shares node positions with its real markup. The resulting broken watcher gets stuck in `component.cleanups` and throws when the component is destroyed, aborting the rest of the cleanup and leaving the component registered in `window.Livewire.all()` forever.

More details and reproduction here : #10655 

## The fix

Tie `$wire.$watch`'s cleanup to the calling element when possible, like `$wire.$entangle` already does. Direct usage (`@script`, file uploads) falls back to the component's lifetime, unchanged. Extracted into `js/features/supportWatch.js`.

## Tests

- [x] Added a regression test: lazy component, placeholder/real markup share node positions, `$wire.$watch` in `x-data`
- [x] Fails before the fix (same error), passes after
- [x] `SupportLazyLoading` (24 tests) pass
- [x] `SupportFileUploads` (10 tests) pass, no collision with file upload's own watcher cleanup
- [x] `JavascriptUnwatchBrowserTest` passes, manual unwatch from `x-data` still works
- [x] `SupportJsModules`/`SupportWireIntersect` failures are pre-existing on `4.x` (confirmed unrelated to this change)

Fixes #10655

### DIFF
--- a/js/$wire.js
+++ b/js/$wire.js
@@ -1,6 +1,7 @@
 import { cancelUpload, removeUpload, upload, uploadMultiple } from './features/supportFileUploads'
 import { dispatch, dispatchEl, dispatchRef, dispatchSelf, dispatchTo, listen } from '@/events'
 import { generateEntangleFunction } from '@/features/supportEntangle'
+import { generateWatchFunction } from '@/features/supportWatch'
 import { findComponentByEl } from '@/store'
 import { dataGet, dataSet } from '@/utils'
 import Alpine from 'alpinejs'
@@ -119,6 +120,8 @@ Alpine.magic('wire', (el, { cleanup }) => {
 
             if (['$entangle', 'entangle'].includes(property)) {
                 return generateEntangleFunction(component, cleanup)
+            } else if (['$watch', 'watch'].includes(property)) {
+                return generateWatchFunction(component, cleanup)
             }
 
             return component.$wire[property]
@@ -277,18 +280,7 @@ wireProperty('$toggle', (component) => (name, live = true) => {
 })
 
 wireProperty('$watch', (component) => (path, callback) => {
-    let getter = () => {
-        return dataGet(component.reactive, path)
-    }
-
-    let unwatch = Alpine.watch(getter, callback)
-
-    let removeCleanup = component.addCleanup(unwatch)
-
-    return () => {
-        removeCleanup()
-        unwatch()
-    }
+    return generateWatchFunction(component)(path, callback)
 })
 
 wireProperty('$effect', (component) => (callback) => {
--- a/js/features/supportWatch.js
+++ b/js/features/supportWatch.js
@@ -1,0 +1,23 @@
+import { dataGet } from '@/utils'
+import Alpine from 'alpinejs'
+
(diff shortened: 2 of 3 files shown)


MaxencePaulin:
Thanks for reviewing and merging! 🙌

PR #10658, opened 2026-09-03
[4.x] Make `EndpointResolver` container-resolvable

ghabriel25:
## Summary

Make Livewire's endpoint resolver container-resolvable, allowing applications to customize Livewire's endpoint paths without modifying Livewire internals.

The existing `EndpointResolver` API is retained as a facade, while the actual implementation is defined by a new `EndpointResolverInterface`. Why facade over direct `app(EndpointResolverInterface::class)` in its method ? Because originally its used as internal class not public API.

Livewire provides a `DefaultEndpointResolver` implementation, and applications can replace it by binding their own implementation of the interface in the service container.

## Motivation

Livewire generates a number of HTTP endpoints for its functionality, including component updates, frontend assets, file uploads, previews, and component assets.

By default, these endpoints use a prefix derived from the application's `APP_KEY`:

```text id="w7tr8s"
/livewire-xxxxxxxx/update
/livewire-xxxxxxxx/livewire.js
/livewire-xxxxxxxx/upload
```

The endpoint prefix is installation-specific by design, but relying directly on `APP_KEY` can be problematic in some deployment environments.

In particular, applications may build Docker images or cache routes in an environment where the runtime exposed`APP_KEY`.

There are also applications that need to control their Livewire endpoint paths because of infrastructure requirements such as:

* WAF rules and allowlists
* Reverse-proxy configuration
* Custom routing requirements
* Docker/image build environments
* Hosting environments with restrictions on generated paths
* Applications that need Livewire endpoints under a specific URL prefix

Currently, the endpoint paths are implemented directly by `EndpointResolver`, which makes it difficult for an application to replace the endpoint resolution behavior.

This change introduces a container-resolvable extension point while preserving the existing `EndpointResolver` API.

Related discussion:
- https://github.com/livewire/livewire/discussions/10310
- https://github.com/livewire/livewire/discussions/7480

## Design

The endpoint resolver is split into three pieces:

```text id="7k4m2p"
EndpointResolver
    │
    │ Facade
    ▼
EndpointResolverInterface
    │
    ├── DefaultEndpointResolver
    │
    └── CustomEndpointResolver
```

### `EndpointResolverInterface`

The interface defines the contract for resolving all Livewire endpoint paths:

```php id="q8l1se"
interface EndpointResolverInterface
{
    public function prefix(): string;

    public function updatePath(): string;

    public function scriptPath(bool $minified = false): string;

    public function mapPath(bool $csp = false): string;

    public function uploadPath(): string;

    public function previewPath(): string;

    public function componentJsPath(): string;

    public function componentCssPath(): string;

    public function componentGlobalCssPath(): string;
}
```

Applications that need custom endpoint behavior can implement this interface.

### `DefaultEndpointResolver`

The existing endpoint-generation behavior is moved into `DefaultEndpointResolver`.

This class implements `EndpointResolverInterface` and provides the same default paths Livewire currently uses.

For applications that do not customize the resolver, this implementation is used automatically.

### `EndpointResolver` facade

The existing `EndpointResolver` class becomes a Laravel facade targeting `EndpointResolverInterface`.

This means existing Livewire code can continue to use:

```php
EndpointResolver::updatePath();
EndpointResolver::scriptPath();
EndpointResolver::uploadPath();
```

without needing to know which concrete resolver is configured.

The calls now resolve through the container:

```text id="2omc1r"
EndpointResolver::updatePath()
        ↓
EndpointResolver facade
        ↓
EndpointResolverInterface
        ↓
DefaultEndpointResolver
```

or, when an application provides a custom implementation:

```text id="8a7d2m"
EndpointResolver::updatePath()
        ↓
EndpointResolver facade
        ↓
EndpointResolverInterface
        ↓
CustomEndpointResolver
```

This allows the internal Livewire API to remain concise while making the implementation replaceable.

## Container registration

Livewire registers the default resolver using:

```php id="s5v8cw"
$this->app->singletonIf(
    EndpointResolverInterface::class,
    DefaultEndpointResolver::class,
);
```

`singletonIf()` is intentional.

Livewire should provide a default implementation, but it should not overwrite an implementation that the application has already registered.

This allows applications to customize endpoint resolution using Laravel's normal container mechanisms.

## Customizing endpoint paths

An application can implement `EndpointResolverInterface` and register its implementation in a service provider.

For example:

```php id="k3j4va"
class CustomEndpointResolver implements EndpointResolverInterface
{
    public function prefix(): string
    {
        return '/custom-livewire';
    }

    public function updatePath(): string
    {
        return $this->prefix() . '/custom-update';
    }

    // ...
}
```

If user only needs to override `prefix()` method then they can extends `DefaultEndpointResolver` class without interface

```php
use Livewire\Mechanisms\HandleRequests\DefaultEndpointResolver;

class CustomEndpointResolver extends DefaultEndpointResolver
{
    public function prefix(): string
    {
        return '/custom-livewire';
    }
}
```

The implementation can then be registered:

```php id="x1n8qb"
public function register(): void
{
    $this->app->singleton(
        EndpointResolverInterface::class,
        CustomEndpointResolver::class,
    );
}
```

After registration, Livewire automatically uses the custom implementation through the `EndpointResolver` facade.

For example:

```php
EndpointResolver::updatePath();
```

will resolve to the custom implementation and return the application's configured path.

The custom binding should be registered during the service-provider registration phase so that it is available before Livewire registers its routes.

## Why `singletonIf()`?

The default binding uses `singletonIf()` rather than `singleton()` so that the application remains in control of the binding.

Conceptually:

```text id="6i1v4z"
Application registers custom resolver
            ↓
EndpointResolverInterface → CustomEndpointResolver

Livewire registers default
            ↓
singletonIf() sees existing binding
            ↓
DefaultEndpointResolver is NOT registered
```

Without `singletonIf()`, Livewire could replace an application-provided implementation depending on provider registration order.

## Backwards compatibility

The existing `EndpointResolver` class remains the API used by Livewire internally:

```php
EndpointResolver::prefix();
EndpointResolver::updatePath();
EndpointResolver::scriptPath();
EndpointResolver::uploadPath();
```

These calls remain static and therefore do not require applications or packages to migrate from a static API to instance methods.

The difference is that `EndpointResolver` now acts as a facade rather than containing the endpoint-generation implementation directly.

This allows the implementation to become replaceable without changing the API used by existing code.

## Endpoint coverage

The resolver covers all endpoint types currently generated by Livewire:

* Endpoint prefix
* Component update endpoint
* JavaScript asset endpoint
* Minified JavaScript asset endpoint
* JavaScript source map endpoint
* CSP source map endpoint
* File upload endpoint
* File preview endpoint
* Component JavaScript endpoint
* Component CSS endpoint
* Global component CSS endpoint

All of these paths are resolved through the same `EndpointResolverInterface` implementation.

This ensures that applications overriding the resolver do not have to customize individual endpoint types separately.

## Route registration and route caching

The resolver is used when Livewire registers and generates URLs for its endpoints.

As a result, a custom resolver changes the actual endpoint paths used by Livewire rather than only changing URLs rendered into the frontend.

This is important for applications using route caching.

The custom implementation needs to be registered before Livewire registers its routes so that the custom paths are used when the routes are created and cached.

This also addresses deployment environments where the default `APP_KEY`-derived endpoint prefix cannot be determined during the build or route-cache generation phase.

## Testing

The changes include tests covering the default resolver and custom container bindings.

The custom resolver test registers:

```php
$this->app->singleton(
    EndpointResolverInterface::class,
    CustomEndpointResolver::class,
);
```

and verifies that `EndpointResolver` resolves the configured implementation and returns the custom paths.

The test covers every method in `EndpointResolverInterface`, including:

* Prefix
* Update
* JavaScript
* Minified JavaScript
* Source map
* CSP source map
* Upload
* Preview
* Component JavaScript
* Component CSS
* Global component CSS

Integration tests should additionally verify that the custom resolver is used by Livewire's registered routes, ensuring that the container override affects the actual runtime behavior rather than only the resolver itself.

## Result

Applications that do not customize the resolver see no behavior change:

```text id="5h3z9p"
EndpointResolver
      ↓
EndpointResolverInterface
      ↓
DefaultEndpointResolver
      ↓
Existing Livewire endpoint behavior
```

Applications that need custom endpoint paths can provide their own implementation:

```text id="0z8d4n"
EndpointResolver
      ↓
EndpointResolverInterface
      ↓
CustomEndpointResolver
      ↓
Application-defined endpoint paths
```

This provides a Laravel-native extension point for Livewire endpoint resolution while keeping the existing `EndpointResolver` API intact.


sajjadhossainshohag (commented):
Looks good. Also, add a doc; it will be good.

ghabriel25:
@sajjadhossainshohag I'm not native speaker so I'll leave the documentation to the mainteners 😄

ghabriel25:
What I have in mind is `DefaultEndpointResolver` should be in `src/DefaultEndpointResolver.php` (empty class) that extends `BaseEndpointResolver` which stays inside `src/Mechanisms/HandleRequests/BaseEndpointResolver.php` so user have 2 options here

1. Implements the interface and override all methods
2. Only extends `DefaultEndpointResolver` so they can manage which method should be overrided

calebporzio:
closing for same reason as alt PR

PR #10659, opened 2026-09-03
[4.x] Keep children mounted during an island render

### LINKED ISSUE #10657: Children mounted during an island render are dropped from the parent's children memo (island append rebuilds them on the next render)
### Livewire version

v4.4.0 (the relevant code is unchanged in v4.4.3)

### Laravel version

v13.25.0

### PHP version

8.5.8

Chrome 152 on macOS 26.6 (not browser-specific; reproduces in `Livewire::test`)

Nested components mounted during an island render are dropped from the parent's `children` memo. On the parent's next full render they are treated as never rendered, mounted again with new ids and full HTML, and the morph replaces the live components (their state is lost, their DOM is rebuilt).

Mechanism: when an action is scoped to an island (`wire:island`, `wire:island.append`, `$wire.$island(...)`), `SupportIslands::call` calls `$this->component->skipRender()` and renders only the island. Any `<livewire:child>` mounted inside that island render pushes itself into the parent's `children` store as usual. Then on dehydrate:

```php
// src/Features/SupportNestingComponents/SupportNestingComponents.php
function dehydrate($context)
{
    if ($this->component->shouldSkipRender()) {
        $this->keepRenderedChildren();   // children = previousChildren — the just-mounted ones are lost
    }
    ...
    $context->addMemo('children', $this->getChildren());
}

function keepRenderedChildren()
{
    $this->storeSet('children', $this->storeGet('previousChildren'));
}
```

`keepRenderedChildren()` was written for `#[Renderless]`, where nothing can have been mounted. With islands it overwrites the children the island just mounted.

Practical impact: `wire:island.append` cannot be used for infinite scroll when the appended rows are Livewire components. Every append works, then the next ordinary render of the parent rebuilds every appended row.

I checked whether this is intended: the islands docs' "Considerations" list template-variable scope, loops/conditionals and state synchronization, nothing about nested components, and `SupportIslands/UnitTest.php` / `BrowserTest.php` have no test that renders a `<livewire:…>` child inside an active island (only the reverse, an island inside a lazy child). If nested components inside islands are meant to be unsupported, a line in Considerations would save people the hunt.

Two components (any registered names work; shown as classes so the repro runs in a plain Pest test):

```php
use Livewire\Component;
use Livewire\Livewire;

class ProbeIslandChild extends Component
{
    public int $n = 0;

    {
        return '<div>child {{ $n }}</div>';
    }
}

class ProbeIslandParent extends Component
{
    public int $appendFrom = 0;

    public int $limit = 1;

    // Append renders only the new row; a full render renders the whole window.
    public function loadMore(): void
    {
        $this->appendFrom = $this->limit + 1;
        $this->limit++;
    }

    public function fullRender(): void
    {
        $this->appendFrom = 0;
    }

    {
        return <<<'BLADE'
        <div>
            @island(name: 'cards', always: true)
                @foreach (range($appendFrom ?: 1, $limit) as $i)
                    <livewire:probe-island-child :n="$i" :wire:key="'c'.$i" />
                @endforeach
            @endisland
        </div>
        BLADE;
    }
}
```

Test:

```php
Livewire::component('probe-island-child', ProbeIslandChild::class);

$c = Livewire::test(ProbeIslandParent::class);
dump(array_keys($c->snapshot['memo']['children']));          // ['c1']

$append = [
    'method' => 'loadMore', 'params' => [], 'path' => '',
    'metadata' => ['island' => ['name' => 'cards', 'mode' => 'append']],
];
$c->update(calls: [$append]);                                 // island fragment contains "child 2"
$c->update(calls: [$append]);                                 // island fragment contains "child 3"
dump(array_keys($c->snapshot['memo']['children']));          // still ['c1']  <-- c2, c3 dropped

$c->call('fullRender');
dump($c->html());
// c1 is rendered as a stub  <div wire:id="…" wire:name="probe-island-child" wire:key="c1"></div>
// c2 and c3 are rendered as FRESH mounts: full HTML + wire:snapshot, with NEW wire:ids
```

Observed on v4.4.0:

```
after mount:            children = [c1 => Mq2fiIDkOJ2BXfBoOLTB]
after 2 island appends: children = [c1]                          (c2 id in the fragment: 4Z3fmssXz3I0dvyPCQf1)
after fullRender:       c1 stub (same id), c2 + c3 fresh mounts   (c2 now vXaL9l8RoIUbNLAplkOR)
```

In the browser the same sequence replaces the appended children's DOM on the next parent render.

Not needed; the test output above shows it.

Children mounted during an island render should be remembered like children mounted during a full render, so a later render emits stubs for them and keeps the live components. Merging instead of overwriting fixes it:

```php
function keepRenderedChildren()
{
    $this->storeSet('children', array_merge(
        $this->storeGet('previousChildren', []),
        $this->storeGet('children', []),
    ));
}
```

lazerg:
A `<livewire:child>` mounted during an island render never makes it into the parent's `children` memo. The next ordinary render of the parent treats those children as if they had never rendered, mounts them again with new ids and full HTML, and the morph throws away the live components along with their state.

An island-scoped action (`wire:island`, `wire:island.append`, `$wire.$island(...)`) calls `skipRender()` and renders only the island. Children mounted inside that island push themselves into the parent's `children` store as usual, and then dehydrate discards them:

```php
function dehydrate($context)
{
    if ($this->component->shouldSkipRender()) {
        $this->keepRenderedChildren(); // children = previousChildren
    }
    ...
    $context->addMemo('children', $this->getChildren());
}
```

`keepRenderedChildren()` was written for `#[Renderless]`, where nothing can have been mounted, so overwriting was safe. With islands it drops whatever the island just mounted.

The practical cost is that `wire:island.append` can't be used for infinite scroll when the appended rows are components: every append works, and then the next render of the parent rebuilds all of them.

The fix keeps the previously rendered children and layers this request's children over the top, so an entry mounted during the island render survives and a key rendered in both requests still resolves to the newer child.

Two notes on the details:

`array_replace` rather than the `array_merge` suggested in the issue. `array_merge` renumbers integer-like keys, and a numeric `wire:key` reaches this array as a real integer key, so merging turns keys `[1, 2]` into `[0, 1]` and breaks the lookup for every numeric-keyed child, including ones that work today. `array_replace` preserves the keys.

This covers append mode, which is what the issue reports. Morph-mode islands lose children too, but for a different reason: `SupportNestingComponents` is registered ahead of `SupportIslands`, so the children memo is written before `SupportIslands::dehydrate()` runs `renderImplicitIslandMorphs()`, which is where morph children mount. That looks fixable by re-adding the memo after the morph render, and it passes the suite locally, but it's a separate cause and touches hook ordering, so I left it out rather than widen this PR. Happy to add it here or open a follow-up, whichever you prefer.

Reported and diagnosed in #10657, along with the proposed fix.

Fixes #10657

### DIFF
--- a/src/Features/SupportIslands/UnitTest.php
+++ b/src/Features/SupportIslands/UnitTest.php
@@ -677,4 +677,64 @@ class UnitTest extends TestCase
 
         File::deleteDirectory($compiledPath);
     }
+
+    public function test_children_mounted_during_an_island_render_are_kept_in_the_children_memo()
+    {
+        Livewire::component('island-child', new class extends \Livewire\Component {
+            public $number = 0;
+
+            public function render() {
+                return '<div>child {{ $number }}</div>';
+            }
+        });
+
+        $component = Livewire::test(new class extends \Livewire\Component {
+            public $from = 0;
+
+            public $to = 1;
+
+            public function loadMore()
+            {
+                $this->from = $this->to + 1;
+                $this->to++;
+            }
+
+            public function render() {
+                return <<<'HTML'
+                <div>
+                    @island(name: 'rows', always: true)
+                        @foreach (range($from ?: 1, $to) as $number)
+                            <livewire:island-child :$number :wire:key="'row-'.$number" />
+                        @endforeach
+                    @endisland
+                </div>
+                HTML;
+            }
+        });
+
+        $this->assertSame(['row-1'], array_keys($component->snapshot['memo']['children']));
+
+        $component->update(calls: [
+            [
+                'method' => 'loadMore',
+                'params' => [],
+                'path' => '',
+                'metadata' => [
+                    'island' => [
+                        'name' => 'rows',
+                        'mode' => 'append',
+                    ],
+                ],
+            ],
+        ]);
+
+        $this->assertSame(['row-1', 'row-2'], array_keys($component->snapshot['memo']['children']));
+
+        $appendedChildId = $component->snapshot['memo']['children']['row-2'][1];
+
+        $component->call('$refresh');
+
+        $this->assertStringContainsString('wire:id="'.$appendedChildId.'" wire:name="island-child" wire:key="row-2"', $component->html());
+        $this->assertStringNotContainsString('child 2', $component->html());
+    }
 }
--- a/src/Features/SupportIslands/UnitTest.php
+++ b/src/Features/SupportIslands/UnitTest.php
@@ -689,8 +689,7 @@ class UnitTest extends TestCase
         });
 
         $component = Livewire::test(new class extends \Livewire\Component {
-            public $from = 0;
-
+            public $from = 1;
             public $to = 1;
 
             public function loadMore()
@@ -703,7 +702,7 @@ class UnitTest extends TestCase
                 return <<<'HTML'
                 <div>
                     @island(name: 'rows', always: true)
-                        @foreach (range($from ?: 1, $to) as $number)
+                        @foreach (range($from, $to) as $number)
                             <livewire:island-child :$number :wire:key="'row-'.$number" />
                         @endforeach
                     @endisland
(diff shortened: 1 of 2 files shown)


calebporzio:
thanks!

PR #10660, opened 2026-09-03
[4.x] Complete truncated component-hooks docs and add exception() example

striker561:
## What

`docs/component-hooks.md` on 4.x is truncated, it stops mid-sentence at the literal text `You can` and never finishes its only practical example.

This PR keeps the existing content unchanged (intro, generic `MyComponentHook`, `AppServiceProvider` registration) and replaces the unfinished CSV example with two complete examples that use only public Livewire APIs:

1. **Enforcing a rule before an action runs** — the `call()` hook firing before every action, using its `$returnEarly` argument to redirect unsubscribed users away from `create` actions across the whole app:

```php
public function call($method, $params, $returnEarly)
{
    if ($method !== 'create') {
        return;
    }

    if (Auth::user()?->subscribed()) {
        return;
    }

    $returnEarly(
        $this->component->redirect(route('subscription'))
    );
}
```

2. **Handling exceptions** — the `exception()` hook intercepting exceptions that escape an action, using $stopPropagation()` so handled exceptions aren't re-thrown to Laravel's exception handler:

```php
{
    if (! $e instanceof SubscriptionExpiredException) {
        return;
    }

    $this->component->dispatch('notify', 'Your subscription has expired.');

    $stopPropagation();
}
```

## Why

The page introduces the `call()` and `exception()` hooks in the generic example but never shows how to use either for real, so readers learn the hooks exist, not what they're for. The truncation leaves the page feeling unfinished.

## Tests

N/A, documentation-only change; no PHP or JavaScript behavior modified.

## Docs

Replaces the truncated content of `docs/component-hooks.md` with two working, source-verified examples.

### DIFF
--- a/docs/component-hooks.md
+++ b/docs/component-hooks.md
@@ -99,49 +99,72 @@ class AppServiceProvider extends ServiceProvider
 }
 ```
 
-Now that you've seen the broad overview of Component Hooks, here's a more practical example of using them to provide useful functionality for your application.
+## Enforcing a rule before an action runs
 
-Let's say you wanted to support the ability to return a CSV from any Livewire action, and it would automatically trigger a file download. For example, you could return a Csv from a method called `save` inside a `CreatePost` component:
+The `call()` hook runs before every action on every component. Calling the `$returnEarly` argument prevents the action from running. This makes it possible to enforce rules across the whole application from a single place.
 
-```php
-use Livewire\Component;
-
-class CreateUser extends Component
-{
-    public $username = '';
-
-    public $email = '';
-
-    public function something()
-    {
-        return new Csv();
-    }
-
-    // ...
-}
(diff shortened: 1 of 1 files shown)


calebporzio:
this PR is too opinionated (fixes the truncation but adds more things) would be better to sperate out. i just created a PR and merged it to fix the truncation and i don't want to merge the rest of this PR as-is. thanks for putting in the effort!

striker561:
Alright, thanks! I was digging into the `exception()` hook because I needed a way to handle certain application exceptions globally across Livewire components, without having to add the hook or a trait to each component.

I found the existing example, but couldn't find anything documenting the global registration approach, which is why I included it here.

Would you be open to a separate PR specifically for documenting that use case?

PR #10661, opened 2026-09-04
[4.x] Support container-resolvable for `EndpointResolver`

ghabriel25:
This is an alternative for https://github.com/livewire/livewire/pull/10658 where the concern is `EndpointResolver` class already used by consumers in their internal codebase.

This change from facade to direct container resolver

**#10658**
```php
use Illuminate\Support\Facade;

class EndpointResolver extends Facade
{
    public static function getFacadeAccessor()
    {
        return \Livewire\Mechanisms\HandleRequests\EndpointResolverInterface::class;
    }
}
```

**This PR**
```php
class EndpointResolver
{
    protected static function resolver(): EndpointResolverInterface
    {
        return app(EndpointResolverInterface::class);
    }

    public static function prefix(): string
    {
        return static::resolver()->prefix();
    }
    
    // ...
}
```

### DIFF
--- a/src/LivewireServiceProvider.php
+++ b/src/LivewireServiceProvider.php
@@ -7,6 +7,8 @@ use Livewire\Compiler\Compiler;
 use Illuminate\Foundation\Console\AboutCommand;
 use Composer\InstalledVersions;
 use Livewire\Compiler\CacheManager;
+use Livewire\Mechanisms\HandleRequests\DefaultEndpointResolver;
+use Livewire\Mechanisms\HandleRequests\EndpointResolverInterface;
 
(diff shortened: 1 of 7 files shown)


calebporzio:
Sorry @ghabriel25 - this is too much reworking of plumbing for my comfort.

maybe there's another, simpler, smaller diff way to accomplish what you're looking for.

another thing: if you route every one of those methods through static::resolver() which calls app(), that's a lot of container lookups and those are not free (performance wise).

this is the kind of thing I fear would bite me if i merged it. Sorry, but "no for now". thanks!

ghabriel25:
@calebporzio Its fine. "Maybe" it was the reason why these two have similar failing test, not sure though.

PR #10664, opened 2026-09-04
[4.x] Allow custom x-cloak variants by targeting empty x-cloak attributes

QuentinGab:
## What does this PR do?

This PR changes Livewire's default `x-cloak` CSS selector from:

```css
[x-cloak] {
    display: none !important;
}
```

to:

```css
[x-cloak=""] {
    display: none !important;
}
```

## Why is this needed?

The current `[x-cloak]` selector matches any element containing the `x-cloak` attribute, regardless of its value. This makes it impossible to introduce value-based `x-cloak` variants without Livewire's default style also applying to them.

For example, an application may want to cloak an element only below a specific responsive breakpoint:

```html
<div x-cloak="lg">
    ...
</div>
```

```css
@media (width < theme(--breakpoint-lg)) {
    [x-cloak="lg"] {
        display: none !important;
    }
}
```

With Livewire's current `[x-cloak]` rule, `x-cloak="lg"` is always hidden until Alpine initializes, including on screens at or above the `lg` breakpoint. This defeats the purpose of the responsive variant.

Using `[x-cloak=""]` restricts Livewire's default style to the standard empty `x-cloak` attribute:

```html
<div x-cloak>
    ...
</div>
```

It therefore preserves the existing behavior for regular `x-cloak` usage while allowing applications to define their own value-based variants such as `x-cloak="lg"`.

Alpine still removes the `x-cloak` attribute normally when it initializes, regardless of its value, so this change only makes Livewire's CSS selector more specific and extensible.

### DIFF
--- a/src/Mechanisms/FrontendAssets/FrontendAssets.php
+++ b/src/Mechanisms/FrontendAssets/FrontendAssets.php
@@ -141,11 +141,11 @@ class FrontendAssets extends Mechanism
                 --livewire-progress-bar-color: {$progressBarColor};
             }
 
-            [x-cloak] {
+            [x-cloak=""] {
                 display: none !important;
             }
 
-            [wire\:cloak] {
+            [wire\:cloak=""] {
                 display: none !important;
             }
 


calebporzio:
hmmmm this makes sense and is somewhat tempting, however, it feels kind of like a breaking change - like there MUST be some applications that get caught in the crossfire here.

however, the "breakage" is just an early paint and not like a 500 error or something.

ok, why not. talked myself into merging thanks.

QuentinGab:
Hi @calebporzio, Thanks for merging it.

I just noticed an edge case with this. I have opened a new PR: https://github.com/livewire/livewire/pull/10675

PR #10666, opened 2026-09-04
[4.x] Fix @assets inside implicitly rendered islands not shipping

### LINKED ISSUE #10665: [4.x] @assets inside an implicitly rendered island never ships since v4.4.1
Since v4.4.1 (#10529, batched implicit island morphs), an `@assets` block rendered inside an island that is rendered **implicitly** is dropped from the response. That covers `@island(defer: true)`, `@island(lazy: true)`, a `skip: true` island loaded via `wire:intersect="$refresh"`, and any `wire:*` action originating inside an island. An explicit `$this->renderIsland()` from an action still ships its assets. v4.4.0 works. Reproduced on v4.4.1, v4.4.2 and v4.4.3.

### Cause

`SupportIslands::call` now queues the render through `triggerImplicitIslandRender()` and flushes it in `SupportIslands::dehydrate()` via `renderImplicitIslandMorphs()`.

`SupportScriptsAndAssets` is registered earlier in `LivewireServiceProvider` (around line 174, islands around line 203), so its `dehydrate()` has already copied the component's asset store into `static::$renderedAssets` before the island renders and pushes its `@assets` (`@endassets` compiles to `store($this)->push('assets', …)`). `HandleRequests` then serialises `assets` without them.

### Repro

```blade
<div>
    @island(defer: true)
        @placeholder <div>loading</div> @endplaceholder
        <div data-loaded>ready</div>
        @assets <script src="/x.js" data-x></script> @endassets
    @endisland
</div>
```

Expected: the `__lazyLoadIsland` response carries `assets` with `/x.js`, as on 4.4.0.
Actual on 4.4.1 to 4.4.3: `assets` is empty and the script never reaches `<head>`.

Real-world hit: `<flux:editor>` declares its script inside its own `@assets`. Inside a deferred island it never loads `flux/editor.js`, so the editor renders as a blank box with no toolbar.

### Fix shape

Drain implicit morphs before any hook dehydrates. Either call `$component->renderImplicitIslandMorphs()` right before `trigger('dehydrate', …)` in `HandleComponents`, or register `before('dehydrate', …)` from `SupportIslands::provide()`. Test: after a `__lazyLoadIsland` call, assert `SupportScriptsAndAssets::getAssets()` contains the island's asset.

### Workaround in app code, until then

```php
\Livewire\before('dehydrate', function ($component) {
    if (method_exists($component, 'renderImplicitIslandMorphs')) {
        $component->renderImplicitIslandMorphs();
    }
});
```

lazerg:
`@assets` inside an implicitly rendered island (deferred, lazy, `skip` + `wire:intersect`, or any `wire:*` action fired from inside an island) stopped shipping after #10529 batched implicit island morphs into `SupportIslands::dehydrate()`. `SupportScriptsAndAssets` is registered earlier in the service provider, so its `dehydrate()` already copies the component's asset store into the response before the island actually renders and pushes its `@assets` block.

This adds a `before('dehydrate', ...)` hook from `SupportIslands::provide()` that flushes queued implicit island renders ahead of every other feature's `dehydrate()`, regardless of registration order, so the pushed assets are already in the store by the time they're collected. An explicit `$this->renderIsland()` call from an action runs synchronously during the action itself, so it was never affected and stays unaffected.

Fixes #10665

Verification:
- Focused SupportIslands unit suite: 27 tests, 54 assertions
- Focused SupportIslands + SupportScriptsAndAssets browser suites: 53 tests
- Full unit suite: 1462 tests

### DIFF
--- a/src/Features/SupportIslands/SupportIslands.php
+++ b/src/Features/SupportIslands/SupportIslands.php
@@ -6,12 +6,19 @@ use Livewire\Features\SupportIslands\Compiler\IslandCompiler;
 use Illuminate\Support\Facades\Blade;
 use Livewire\ComponentHook;
 
+use function Livewire\before;
+
 class SupportIslands extends ComponentHook
 {
     public static function provide()
     {
         static::registerInlineIslandPrecompiler();
         static::registerIslandDirective();
+
+        // Flush implicit island renders before any dehydrate hook, so other hooks (e.g. assets) see their output...
+        before('dehydrate', function ($component) {
+            $component->renderImplicitIslandMorphs();
+        });
     }
 
     public static function registerIslandDirective()
--- a/src/Features/SupportIslands/UnitTest.php
+++ b/src/Features/SupportIslands/UnitTest.php
@@ -5,6 +5,7 @@ namespace Livewire\Features\SupportIslands;
 use Tests\TestCase;
 use Livewire\Livewire;
 use Livewire\Features\SupportIslands\Compiler\IslandCompiler;
+use Livewire\Features\SupportScriptsAndAssets\SupportScriptsAndAssets;
 use Illuminate\Support\Facades\File;
 
 class UnitTest extends TestCase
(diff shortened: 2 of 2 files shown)


calebporzio:
interesting, i wouldn't think this would be supported in the first place?

lazerg:
It worked fine in 4.4.0. #10529's batching of implicit island morphs is what broke it: `SupportScriptsAndAssets::dehydrate()` now runs before the island's queued render actually flushes, so the `@assets` push happens after assets are already collected. Real-world hit from the issue: `<flux:editor>` declares its script via `@assets`, so it silently loses its toolbar inside any deferred or lazy island. Repro and full details in #10665.

calebporzio:
ahhhh there it is. thanks for the explanation.

PR #10667, opened 2026-09-05
[4.x] Fix `@assets` and `@scripts` blocks inside implicit rendered islands not available

### LINKED ISSUE #10665: [4.x] @assets inside an implicitly rendered island never ships since v4.4.1

ghabriel25:
> [!NOTE]
> This is an alternative for https://github.com/livewire/livewire/pull/10666

## Problem

Assets and scripts registered inside an implicitly rendered deferred island are not included in the response.

Implicit island morphs are rendered during `SupportIslands::dehydrate()`, which runs after `SupportScriptsAndAssets::dehydrate()`. This means `@assets` and `@script` blocks registered while rendering the island are added too late to be picked up by the normal scripts/assets dehydration step.

As a result, the island itself is rendered correctly, but assets and scripts declared inside it are missing from the response.

## Solution

Collect any scripts and assets registered during implicit island rendering after the island morphs have been rendered.

The existing scripts/assets state and memo are used to avoid re-registering anything that was already processed during the request. Newly registered scripts are pushed onto the response effects, while newly registered assets are added back to the rendered assets collection.

This keeps the existing scripts/assets handling intact and only accounts for registrations that happen during implicit island rendering.

## Tests

Added a regression test covering a deferred implicit island containing both:

* `@assets` with an external script
* `@script` with an inline script

The test verifies that:

* the deferred island is rendered when it is loaded;
* the asset is included in the response assets payload;
* the script is included in the component effects.

Fixes: https://github.com/livewire/livewire/issues/10665

### DIFF
--- a/src/Features/SupportIslands/SupportIslands.php
+++ b/src/Features/SupportIslands/SupportIslands.php
@@ -5,6 +5,7 @@ namespace Livewire\Features\SupportIslands;
 use Livewire\Features\SupportIslands\Compiler\IslandCompiler;
 use Illuminate\Support\Facades\Blade;
 use Livewire\ComponentHook;
+use Livewire\Features\SupportScriptsAndAssets\SupportScriptsAndAssets;
 
 class SupportIslands extends ComponentHook
 {
@@ -74,6 +75,10 @@ class SupportIslands extends ComponentHook
     public function dehydrate($context)
     {
         $this->component->renderImplicitIslandMorphs();
+
+        // Collect assets/scripts immediately after island rendering,
+        // as SupportScriptsAndAssets::dehydrate runs before this hook.
+        SupportScriptsAndAssets::collectComponentPendingAssetsAndScripts($this->component, $context);
 
         $context->addMemo('islands', $this->component->getIslands());
 
--- a/src/Features/SupportIslands/UnitTest.php
+++ b/src/Features/SupportIslands/UnitTest.php
@@ -7,6 +7,8 @@ use Livewire\Livewire;
 use Livewire\Features\SupportIslands\Compiler\IslandCompiler;
 use Illuminate\Support\Facades\File;
 
+use function Livewire\invade;
+
 class UnitTest extends TestCase
 {
     public function test_class_component_island_recovers_when_cached_file_is_deleted_between_requests()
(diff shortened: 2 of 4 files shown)


calebporzio:
went with the other pr thanks

PR #10668, opened 2026-09-06
[4.x] Fix stale `wire:navigate.hover` prefetch HTML after a component action

ghabriel25:
## Summary

This PR fixes an issue where HTML prefetched by `wire:navigate.hover` could become stale after a Livewire component action modified the requirement to visit destination page.

### Problem

`wire:navigate.hover` can prefetch a destination page before the user clicks a navigation link. That prefetched HTML is then cached and reused when the navigation actually occurs.

However, if a Livewire component action runs after the prefetch completes, the current page state may change before the user navigates. In this situation, the previously prefetched navigation response can no longer reliably represent the correct state of the application.

Because the stale prefetched HTML remained cached, a subsequent `wire:navigate` navigation or `$this->redirect(..., navigate: true)` could incorrectly reuse that outdated response instead of fetching fresh HTML.

```blade
<a href="{{ route('profile.show') }}" wire:navigate.hover>Profile</a>
```
```php
Route::get('/profile', ShowProfile::class)
    ->middleware(HavePassword::class)
    ->name('profile.show');

Route::get('/create-password', CreatePassword::class)
    ->middleware(HaveNoPassword::class)
    ->name('password.create');
```

In this situation, user hover the profile link (prefetch) then click then link to create a password less than `30s` which is prefetched html time limit stored. Because the stored html is `/create-password`, it will navigate there however user already create a password so it will redirect back again to original page.

Related discussion: https://github.com/livewire/livewire/discussions/9386

### Solution

This PR ensures that prefetched navigation HTML is invalidated when a component action is called.

As a result, after a Livewire action changes the application state, a later `wire:navigate` navigation will no longer use a prefetch response created before that action. Instead, navigation can proceed using up-to-date HTML.

This prevents stale prefetched content from being rendered after the state of the current page has changed.

### Tests

* `test_livewire_action_clears_navigate_prefetch_cache`
* `test_livewire_action_clears_prefetch_so_protected_route_is_refetched`
* `test_navigate_fallback_when_in_flight_prefetch_cleared_by_livewire_request`

### DIFF
--- a/js/plugins/navigate/prefetch.js
+++ b/js/plugins/navigate/prefetch.js
@@ -18,24 +18,24 @@ export function prefetchHtml(destination, callback, errorCallback) {
 
     if (prefetches[uri]) return
 
-    prefetches[uri] = { finished: false, html: null, whenFinished: () => {}, whenFailed: () => {}, expiry: null }
+    let state = { finished: false, html: null, whenFinished: () => {}, whenFailed: () => {}, expiry: null }
+
+    prefetches[uri] = state
 
     // Bound the lifetime of an in-flight prefetch. If it takes too long,
     // invalidate it and allow any waiting navigation to fall back.
-    prefetches[uri].expiry = setTimeout(() => {
-        invalidatePrefetch(uri)
+    state.expiry = setTimeout(() => {
+        invalidatePrefetch(uri, state)
     }, cacheDuration)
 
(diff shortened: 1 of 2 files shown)


ghabriel25:
Closing as this introduce too many gap on current mechanism

PR #10669, opened 2026-09-07
[4.x] Fix stale wire:navigate.hover prefetches after Livewire requests

ghabriel25:
> [!NOTE]
> Based on https://github.com/livewire/livewire/pull/10668 with a minimal fix.

## Affected version
- 3.x (backport)
- 4.x (current PR)

## Problem

`wire:navigate.hover` prefetches and caches the destination HTML for up to 30 seconds.

If a Livewire request changes application state during that period, the cached HTML can become stale. Subsequent navigation may then use the stale prefetched response instead of making a fresh request.

This is particularly problematic for protected routes. For example, a route may initially redirect to a password-creation page when prefetched, but become accessible after a Livewire action updates the session. Without invalidating the prefetch, navigating to the route can continue using the stale prefetched response.

Prefetches that redirect also must not be cached under the original URL. The response belongs to the redirected destination, while the cache entry is keyed by the original URL.

## Changes

* Invalidate all navigation prefetches when a Livewire request is sent.
* Fall back to a normal navigation request when an in-flight prefetch is invalidated while navigation is waiting for it.
* Ignore responses from prefetch requests that were invalidated while they were in flight.
* Do not cache prefetched responses when the request was redirected to a different URL.
* Ensure an invalidated prefetch's expiry timer cannot expire a newer prefetch for the same URL.
* Ignore responses from invalidated prefetch requests, including when a newer prefetch for the same URL has already replaced them.

## Why

A prefetched page represents application state at the time the prefetch was made. A subsequent Livewire request can change that state, including session state, authorization state, or other data used to render the destination.

Discarding the prefetch when a Livewire request is sent ensures that the next navigation gets a fresh response.

Because an in-flight prefetch cannot necessarily be cancelled, its response may arrive after the prefetch has been invalidated. If a new prefetch for the same URL has started in the meantime, the old response must not complete or modify the newer prefetch. Each prefetch therefore verifies that it is still the current prefetch for the URL before handling its response.

When an invalidated prefetch is replaced by a new prefetch for the same URL, the old prefetch's 30-second expiry timer must not remove the new cache entry. Each prefetch should retain its own cache lifetime.

The existing `whenFailed` fallback is used for in-flight prefetches, so no additional navigation-tracking mechanism is required.

This intentionally keeps the existing 30s cache timer mechanism (no broader prefetch lifecycle rewrite).

## Trade-off

Every sent Livewire request clears the hover prefetch cache. That is intentional: correctness for auth/session/state takes priority over keeping speculative prefetches.

Clearing runs on `onSend` so cancelled-before-send requests do not discard the cache.

## Tests

Added coverage for:

* Clearing a cached navigation prefetch after a Livewire action.
* Refetching a previously protected route after its authorization state changes.
* Falling back to a fresh navigation request when an in-flight prefetch is invalidated by a Livewire request.
* Ensuring an invalidated prefetch response cannot complete or interfere with a newer prefetch for the same URL.
* Ensuring an invalidated prefetch cannot expire a newer prefetch for the same URL before the newer prefetch's 30-second cache lifetime has elapsed.
* Redirected prefetch responses are not cached for the original URL.

Fixes: https://github.com/livewire/livewire/discussions/9386

### DIFF
--- a/js/plugins/navigate/prefetch.js
+++ b/js/plugins/navigate/prefetch.js
@@ -28,7 +28,13 @@ export function prefetchHtml(destination, callback, errorCallback) {
 
     if (prefetches[uri]) return
 
-    prefetches[uri] = { finished: false, html: null, whenFinished: () => setTimeout(() => delete prefetches[uri], cacheDuration), whenFailed: () => {} }
+    let state = { finished: false, html: null, whenFinished: () => {}, whenFailed: () => {} }
+
+    state.whenFinished = () => setTimeout(() => {
+        if (prefetches[uri] === state) delete prefetches[uri]
+    }, cacheDuration)
+
+    prefetches[uri] = state
 
     performFetch(uri, (html, routedUri, status) => {
         let state = prefetches[uri]
--- a/js/plugins/navigate/prefetch.js
+++ b/js/plugins/navigate/prefetch.js
@@ -37,10 +37,8 @@ export function prefetchHtml(destination, callback, errorCallback) {
     prefetches[uri] = state
 
     performFetch(uri, (html, routedUri, status) => {
-        let state = prefetches[uri]
-
         // The prefetch may have been invalidated while its request was in flight.
-        if (! state) return
+        if (prefetches[uri] !== state) return
 
         storeCurrentPageStatus(status)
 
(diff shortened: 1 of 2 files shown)


ghabriel25:
I cannot reproduce the lifetime-timer regression when old entry `whenFinished()` callback runs will remove new entry with the same uri

ghabriel25:
Hi, @calebporzio ready for review. Throwing some agents for this would be great!

PR #10681, opened 2026-09-08
[4.x] Support `wire:navigate` with hash fragment

ghabriel25:
## Summary

Add support for URL fragments when using `wire:navigate`, so navigating to a URL with a hash scrolls to the corresponding element like a normal browser navigation.

## Problem

When `wire:navigate` navigates to a URL containing a fragment, Livewire replaces the page content but does not perform the browser's default fragment scrolling.

For example, navigating to:

```html
<a href="/posts#comments" wire:navigate>
    Comments
</a>
```

will update the URL to `/posts#comments`, but the page will remain at its current scroll position instead of scrolling to `#comments`.

Related discussion: https://github.com/livewire/livewire/discussions/8073

## Solution

Handle the fragment after the new page has been rendered:

1. Resolve the fragment target

   * Match elements by `id`
   * Support percent-encoded fragments
   * Support legacy `<a name="...">` anchors
   * Handle the special `#top` fragment
2. Scroll to the target when it exists
3. Fall back to the existing scroll-to-top behavior when there is no fragment
4. Let fragment navigation take precedence over `preserve-scroll`, matching normal browser navigation behavior

Back/forward navigation continues to use the existing scroll restoration behavior.

## Tests

* Navigate to an element using a URL fragment
* Navigate to a missing fragment
* Navigate using a percent-encoded fragment
* Navigate to `#top`
* Navigate to a legacy named anchor
* Navigate to a fragment with `preserve-scroll`
* Preserve fragment scroll position when navigating back and forward


ghabriel25:
Closing in favor https://github.com/livewire/livewire/pull/10692

PR #10684, opened 2026-09-08
[4.x] Prevent malformed synthesized values from throwing before property type handling

### LINKED ISSUE #10683: enum property in component with malformed POST causes wrongful error 500
### Livewire version

4.3.3

### Laravel version

13.21.1

### PHP version

8.5.8

Firefox on MacOS

A component with a backed-enum typed property throws an unhandled `TypeError` (HTTP 500, reported to Sentry as an application bug) whenever the `updates` map in an update request carries a non-scalar value for that property.

```
TypeError: App\Enums\AuthIndexForm::from(): Argument #1 ($value) must be of type int, array given
  at vendor/livewire/livewire/src/Mechanisms/HandleComponents/Synthesizers/EnumSynth.php:34
```

```php
enum Status: int
{
    case Draft = 0;
    case Published = 1;
}

class Post extends Component
{
    public Status $status = Status::Draft;

    public function render()
    {
        return '<div></div>';
    }
}
```

```php
Livewire::test(Post::class)->set('status', []); // TypeError, 500
```

_No response_

Livewire already recognises and handles this exact class of traffic. `HandleComponents::setComponentPropertyAwareOfTypes()` (line 647) reads:

```php
// This is almost certainly a bot/scanner probing typed properties
// with wrong-type values. Abort with 419 directly so it never
// reaches the top-level catch (which would report it as a real bug).

abort(419);
```

ghabriel25:
## Summary

Fix malformed property updates causing certain synthesizers to throw while hydrating values before Livewire reaches its existing typed-property error handling.

This affects:

- `EnumSynth`
- `CarbonSynth`
- `StringableSynth`

These synthesizers know the expected shape of their dehydrated values. When an incoming value does not match that shape, it is passed through unchanged instead of being given to an object constructor or conversion method that would throw.

This allows the invalid value to reach Livewire's existing property assignment path, where invalid typed-property assignments are handled consistently.

Array-shaped synthesizers remains protected by `ArrayShapeSynthesizerInterface`

## Problem

A malformed property update can currently throw inside a synthesizer.

For example:

```text
update property
    ↓
EnumSynth::hydrate()
    ↓
Enum::from([])
    ↓
TypeError
```
Similarly, malformed values can be passed to constructors or conversion helpers in `CarbonSynth` and `StringableSynth`.

Because these exceptions occur during synthesizer hydration, they happen before Livewire's existing typed-property error handling can process the invalid value.

## Solution

**EnumSynth**

`BackedEnum::from()` only accepts `string` and `int` values.

Both hydration paths are guarded:
- `hydrate()`
- `hydrateFromType()`

Malformed values are returned unchanged instead of calling `::from()` with an unsupported value.

**CarbonSynth**

Carbon's dehydrated representation is a date/time string.

Both hydration paths are guarded:
- `hydrate()`
- `hydrateFromType()`

Non-string values are returned unchanged instead of being passed to the `Carbon`/`DateTime` constructor.

**StringableSynth**

`StringableSynth` dehydrates values to strings.

Non-string values are returned unchanged instead of being passed to `str()` during hydration.

This allows malformed values to continue to Livewire's existing property assignment and typed-property error handling.

Fixes: https://github.com/livewire/livewire/issues/10683

### DIFF
--- a/src/Mechanisms/HandleComponents/Synthesizers/CarbonSynth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/CarbonSynth.php
@@ -41,11 +41,15 @@ class CarbonSynth extends Synth {
     static function hydrateFromType($type, $value) {
         if ($value === '' || $value === null) return null;
 
+        if (! is_string($value)) return $value;
+
         return new $type($value);
     }
 
     function hydrate($value, $meta) {
         if ($value === '' || $value === null) return null;
+
+        if (! is_string($value)) return $value;
 
         $type = $meta['type'] ?? null;
 
--- a/src/Mechanisms/HandleComponents/Synthesizers/EnumSynth.php
+++ b/src/Mechanisms/HandleComponents/Synthesizers/EnumSynth.php
@@ -16,6 +16,8 @@ class EnumSynth extends Synth {
     static function hydrateFromType($type, $value) {
         if ($value === null || $value === '') return null;
 
+        if (! is_int($value) && ! is_string($value)) return $value;
+
         return $type::from($value);
     }
 
@@ -28,6 +30,8 @@ class EnumSynth extends Synth {
 
     function hydrate($value, $meta) {
         if ($value === null || $value === '') return null;
+
+        if (! is_int($value) && ! is_string($value)) return $value;
 
         $class = $meta['class'] ?? null;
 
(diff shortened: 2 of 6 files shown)


edalzell:
Related probe/noise failure mode on the `calls` side (invalid method names / bad `__lazyLoad` params → reportable 500 instead of quiet 419): https://github.com/livewire/livewire/issues/10688

afraca:
Great straight forward solution, thanks. There's no urgency, but would be nice to get it in soon @calebporzio  :) 

edit: my apologies , I see multiple PR's are dealing with this and it is a broader issue. Good luck on that 👍

PR #10687, opened 2026-09-08
[4.x] Don't read uninitialized typed properties when expanding consolidated form object updates

### LINKED ISSUE #10685: Uninitialized typed property is read in expandConsolidatedFormObjectUpdates() before #[Locked] runs, turning tampered updates into 500s
## Livewire version

v4.4.4 (also v4.4.2; the line is unchanged on main as of 2026-09-08)

## Laravel version

v13.29.0

## PHP version

8.4.22

Not browser-specific (server-side). Observed from Chrome on macOS user agents; reproduced with `Livewire::test()`.

**Not a security issue.** Nothing is exposed or bypassed: `#[Locked]` still protects the property, and the request fails either way. The problem is the failure mode. Automated probes that replay captured Livewire payloads against public pages turn this into a stream of spurious 500s (and, for us, error alerts landing in Slack) where the intended result is a quiet 419 from the lock. This report is about squashing that noise.

A component with a typed public property that is never assigned (declared, but not set in `mount()`) dehydrates that property as `null` in the snapshot. On the next request the property is therefore still uninitialized after hydration. If an incoming update for that property carries an **array** value, `HandleComponents::expandConsolidatedFormObjectUpdates()` reads the property to test `instanceof Form` and PHP throws:

```
Error: Typed property App\Livewire\ContentPage::$group must not be accessed before initialization
#0 vendor/livewire/livewire/src/Mechanisms/HandleComponents/HandleComponents.php(395): HandleComponents->expandConsolidatedFormObjectUpdates()
#1 vendor/livewire/livewire/src/Mechanisms/HandleComponents/HandleComponents.php(220): HandleComponents->updateProperties()
#2 vendor/livewire/livewire/src/LivewireManager.php(132): HandleComponents->update()
#3 vendor/livewire/livewire/src/Mechanisms/HandleRequests/HandleRequests.php(205): LivewireManager->update()
```

This happens **before** `updateProperty()` runs, so none of the `update` hooks fire. In particular `#[Locked]` never gets to throw `CannotUpdateLockedPropertyException`. The practical effect: a request that tampers with a locked property on a public page renders as a 500 instead of the 419 the lock is designed to produce. We are seeing exactly this from automated probes against a guest-facing component in production.

The offending check is at `src/Mechanisms/HandleComponents/HandleComponents.php:419`:

```php
if (is_array($value) && property_exists($component, $path) && $component->$path instanceof Form) {
```

`property_exists()` is true for a declared-but-uninitialized typed property, and the read that follows is what throws.

```php
use Livewire\Component;
use Livewire\Livewire;

{
    #[Locked]
    public stdClass $group; // declared, never assigned

    {
        return '<div></div>';
    }
};

// Snapshot data at this point: {"group": null}

$testable->set('group', ['a' => 1]);

// Actual: Error: Typed property Livewire\Component@anonymous::$group must not be accessed before initialization
// Expected: Livewire\Features\SupportLockedProperties\CannotUpdateLockedPropertyException
```

Steps:

1. Declare a typed public property on a component and do not assign it in `mount()`.
2. Render the component once (the snapshot carries the property as `null`).
3. Send an update for that property whose value is an array.
4. Observe the uninitialized-property `Error` from `expandConsolidatedFormObjectUpdates()`.

The same property without `#[Locked]` throws the same `Error`; with `#[Locked]` the point is that the lock never runs.

Note the property type does not matter as long as it is not `array`: an uninitialized `public array $items;` dehydrates as `[]` rather than `null`, so it is initialized again on the next request and the bug does not show.

The pre-pass should skip properties that hold no value rather than reading them. For example:

```php
if (is_array($value) && isset($component->$path) && $component->$path instanceof Form) {
```

`isset()` is false for an uninitialized typed property (and for a non-public property name, which `property_exists()` also matches), and true for an initialized `Form`. With that change the update falls through to `updateProperty()`, the `update` hooks run, and a `#[Locked]` property throws `CannotUpdateLockedPropertyException` as intended.

I have not opened a PR because I am not close enough to the hydration internals to be confident about side effects of the change; happy to test a fix against our production payloads.

lazerg:
A component with a typed public property that is never assigned dehydrates as `null`, so the property is still uninitialized on the next request. `expandConsolidatedFormObjectUpdates()` guards its `instanceof Form` read with `property_exists()`, which is true for a declared-but-uninitialized typed property, and the read itself throws.

That happens before `updateProperty()`, so no `update` hooks run and `#[Locked]` never gets a chance to throw. A tampered update on a locked property comes back as a 500 instead of the 419 the lock is meant to produce.

The pre-pass now skips properties that hold no value, using the same `Utils::propertyIsTypedAndUninitialized()` check `FormObjectSynth` already relies on. The update falls through to `updateProperty()` and the lock behaves as expected.

Fixes #10685

### DIFF
--- a/src/Features/SupportLockedProperties/UnitTest.php
+++ b/src/Features/SupportLockedProperties/UnitTest.php
@@ -43,6 +43,20 @@ class UnitTest extends \Tests\TestCase
         ->set('foo.count', 2);
     }
 
+    function test_cant_update_uninitialized_typed_locked_property_with_an_array()
+    {
+        $this->expectException(CannotUpdateLockedPropertyException::class);
+        $this->expectExceptionMessage(
+            'Cannot update locked property: [group]'
+        );
+
+        Livewire::test(new class extends TestComponent {
+            #[BaseLocked]
+            public \stdClass $group;
+        })
+        ->set('group', ['count' => 1]);
+    }
+
     function test_can_update_locked_property_with_similar_name()
     {
         Livewire::test(new class extends TestComponent {
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -416,7 +416,7 @@ class HandleComponents extends Mechanism
         $expanded = [];
 
         foreach ($updates as $path => $value) {
-            if (is_array($value) && property_exists($component, $path) && $component->$path instanceof Form) {
+            if (is_array($value) && property_exists($component, $path) && ! Utils::propertyIsTypedAndUninitialized($component, $path) && $component->$path instanceof Form) {
                 foreach ($value as $key => $child) {
                     $expanded["{$path}.{$key}"] = $child;
                 }
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -416,7 +416,11 @@ class HandleComponents extends Mechanism
         $expanded = [];
 
         foreach ($updates as $path => $value) {
-            if (is_array($value) && property_exists($component, $path) && ! Utils::propertyIsTypedAndUninitialized($component, $path) && $component->$path instanceof Form) {
+            if (is_array($value)
+                && property_exists($component, $path)
+                && ! Utils::propertyIsTypedAndUninitialized($component, $path)
+                && $component->$path instanceof Form
+            ) {
                 foreach ($value as $key => $child) {
                     $expanded["{$path}.{$key}"] = $child;
                 }


lazerg:
The three red jobs all failed on the "Download built assets" step, so this is an artifact hiccup in CI and not something my change causes.

lazerg:
The Browser 2 shard failure is the same two PersistentMiddleware tests that already fail on 4.x itself (run 34221382220), so it predates this change as well.

lazerg:
Hey @calebporzio, please check when you have time.

edalzell:
Related: same class of scanner noise on malformed `calls` rather than `updates` — https://github.com/livewire/livewire/issues/10688

PR #10689, opened 2026-09-09
[4.x] Quiet-abort malformed call probes with 419

### LINKED ISSUE #10688: Malformed `calls` (invalid method names / bad `__lazyLoad` params) throw reportable 500s instead of a quiet 419
## Livewire version

v4.4.2 (same failure mode on current `4.x` as of 2026-09-08)

## Laravel version

v13.x

## PHP version

8.5.8

Not browser-specific (server-side). Observed from automated probes against public pages; reproduced by posting a crafted Livewire update body.

**Not a security issue.** Nothing is exposed or bypassed. The problem is the failure mode. Automated scanners fuzz Livewire `update` payloads by putting garbage in `components[].calls` — e.g. `method: "|"`, non-string methods, or `__lazyLoad` with params like `["|"]` instead of a base64 JSON snapshot. Those requests currently throw reportable exceptions (HTTP 500 / error tracker noise) where the intended posture — matching typed-property scanner handling — is a quiet 419.

Livewire already distinguishes scanner-injected **property** `TypeError`s and `abort(419)`s them outside debug (`HandleRequests` / `setComponentPropertyAwareOfTypes`, see #10211). There is no equivalent gate for **calls**:

```php
// HandleComponents::callMethods()
if (! in_array($method, $methods)) {
    throw new MethodNotFoundException($method);
}
```

So `method: "|"` becomes `MethodNotFoundException` and lands in application error reporting. Invalid `__lazyLoad` params similarly blow up later (checksum / array-offset `ErrorException`) instead of being rejected as a malformed payload.

This is the same class of production noise as #10685 / #10683 — probes against public Livewire endpoints — just aimed at `calls` instead of `updates`.

Related app-side workaround we use today: ignore reporting when a call method is not a valid PHP identifier, or when `__lazyLoad`’s first param is not a base64-decoded JSON array snapshot. Prefer the framework to quiet-reject these the way it already does for wrong-type property updates.

```php
use Livewire\Component;
use Livewire\Livewire;
use Livewire\Exceptions\MethodNotFoundException;

{
    {
        return '<div></div>';
    }
};

// Simulate a probe call: method name is not a PHP identifier.
// In a real update POST this is components[0].calls = [{"method":"|","params":[],"metadata":[]}].
try {
    $testable->call('|');
} catch (MethodNotFoundException $e) {
    // Actual (production, app.debug=false): reportable 500 / error tracker alert
    // Expected: quiet abort(419), same as scanner TypeErrors on property assignment
}

// Second shape: __lazyLoad with a non-snapshot param (e.g. "|").
// components[0].calls = [{"method":"__lazyLoad","params":["|"],"metadata":[]}]
// Actual: ErrorException from checksum / offset access while decoding the mount snapshot
// Expected: quiet abort(419) for a malformed lazy-load payload
```

Steps:

1. Render any public Livewire component once and capture a valid update URL + CSRF token.
2. POST an update body whose `calls` entry uses `method: "|"` (or another non-identifier), with an otherwise valid snapshot.
3. Observe `MethodNotFoundException` / 500 instead of a quiet 419.
4. Repeat with `method: "__lazyLoad"` and `params: ["|"]`; observe a reportable `ErrorException` instead of a quiet 419.

Valid PHP method names that simply do not exist on the component (e.g. `missingAction`) can keep today’s `MethodNotFoundException` behaviour — this report is about payloads that are not plausible method calls at all.

## Screenshots/screen recordings

N/A — server-side exception noise from probes.

Before resolving / invoking a call:

1. If `method` is not a string matching a PHP identifier (`^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$`), `abort(419)` outside debug (same posture as #10211).
2. If `method === '__lazyLoad'` and `params[0]` is not a string that base64-decodes to a JSON array snapshot, `abort(419)` outside debug instead of letting checksum / offset access throw.

That keeps real missing methods reportable while matching the existing scanner soft-path for typed-property probes.

Happy to test a fix against production probe payloads.

## Please confirm (incomplete submissions will not be addressed)

edalzell:
## Summary

Malformed `components[].calls` from scanners currently produce reportable 500s:

- `method: "|"` (or other non-PHP-identifiers) → `MethodNotFoundException`
- `__lazyLoad` with params like `["|"]` instead of a base64 JSON snapshot → checksum / offset `ErrorException`

Livewire already quiet-`abort(419)`s scanner-injected **property** `TypeError`s outside debug. This applies the same posture to **calls**.

Valid PHP method names that simply do not exist on the component (e.g. `missingAction`) still throw `MethodNotFoundException`.

Fixes #10688

## Changes

- `HandleComponents::callMethods()`: reject non-string / non-identifier method names with `abort(419)` (rethrow as `MethodNotFoundException` when `app.debug`)
- `SupportLazyLoading`: validate `__lazyLoad` params[0] is a base64 JSON array snapshot before `resurrectMountParams()`; otherwise `abort(419)` (or `CorruptComponentPayloadException` in debug)
- Use strict `base64_decode` when resurrecting mount params

## Test plan

- [x] `vendor/bin/phpunit --testsuite Unit src/Mechanisms/HandleRequests/UnitTest.php`
- [x] `vendor/bin/phpunit --testsuite Unit src/Features/SupportLazyLoading/UnitTest.php`
- [ ] Confirm `|` method / bad `__lazyLoad` probes return 419 outside debug and are not reported

### DIFF
--- a/src/Features/SupportLazyLoading/SupportLazyLoading.php
+++ b/src/Features/SupportLazyLoading/SupportLazyLoading.php
@@ -148,7 +148,16 @@ class SupportLazyLoading extends ComponentHook
         // Only applies while a lazy load is being resumed.
         if ($this->storeGet('isLazyLoadHydrating') !== true) return;
 
-        [ $encoded ] = $params;
+        $encoded = $params[0] ?? null;
+
+        // Probes send "|" / non-strings instead of a base64 snapshot. Quiet 419 outside debug.
+        if (! $this->lazyLoadParamLooksValid($encoded)) {
+            if (config('app.debug')) {
+                throw new CorruptComponentPayloadException;
+            }
+
+            abort(419);
+        }
 
         $mountParams = $this->resurrectMountParams($encoded);
 
@@ -229,7 +238,7 @@ class SupportLazyLoading extends ComponentHook
     // Rebuild the captured mount params from the container snapshot.
     function resurrectMountParams($encoded)
     {
-        $snapshot = json_decode(base64_decode($encoded), associative: true);
+        $snapshot = json_decode(base64_decode($encoded, strict: true), associative: true);
 
         $this->registerContainerComponent();
 
@@ -249,6 +258,21 @@ class SupportLazyLoading extends ComponentHook
         $hook->setComponent($this->component);
 
         $hook->mount($params);
+    }
+
+    protected function lazyLoadParamLooksValid(mixed $encoded): bool
+    {
+        if (! is_string($encoded) || $encoded === '') {
+            return false;
+        }
+
+        $decoded = base64_decode($encoded, strict: true);
+
+        if ($decoded === false) {
+            return false;
+        }
+
+        return is_array(json_decode($decoded, associative: true));
     }
 
     // A throwaway component used to carry mount params across the round-trip.
--- a/src/Mechanisms/HandleComponents/HandleComponents.php
+++ b/src/Mechanisms/HandleComponents/HandleComponents.php
@@ -532,11 +532,25 @@ class HandleComponents extends Mechanism
         }
 
         $returns = [];
+
+        // Validate method names before any call bookkeeping — probes send "|", null, arrays, etc.
+        foreach ($calls as $call) {
+            $method = $call['method'] ?? null;
+
+            if (! $this->isValidComponentMethodName($method)) {
+                if (config('app.debug')) {
+                    throw new MethodNotFoundException(is_string($method) ? $method : get_debug_type($method));
+                }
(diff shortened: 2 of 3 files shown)


calebporzio:
Yooo 👋

i wonder if there is a broader story here. Like if we should apply this 419-not-loud-500 treatment further upstream so anything of this nature is treated that way. Lemme throw an agent at it.

edalzell:
Fable it, my dude

calebporzio:
thanks Erin, merged the other one

PR #10692, opened 2026-09-09
Fix fragment scrolling after wire:navigate

calebporzio:
Navigating with `<a href="/posts#comments" wire:navigate>` updates the URL and replaces the page, but leaves the fragment target below the viewport. This change scrolls the target into view after the existing scroll restoration completes.

Fragment lookup supports element IDs, percent-encoded IDs, literal percent signs, and legacy named anchors. Missing targets retain the existing scroll-to-top behavior. Cached Back/Forward positions and explicit `preserve-scroll` continue to take precedence.

Validation:
- Confirmed the new browser regression fails against unmodified `4.x` and passes with the fix.
- Nine focused Dusk browser tests pass with 45 assertions, covering fragment scrolling, missing targets, history restoration, and existing link/programmatic scroll preservation. PHPUnit reports two deprecations.
- Rebuilt JavaScript assets for browser verification; generated bundles are excluded from the commit.

This branch contains only the fragment-scrolling fix and browser coverage. No separate discussion was created.

### DIFF
--- a/js/plugins/navigate/scroll.js
+++ b/js/plugins/navigate/scroll.js
@@ -26,9 +26,34 @@ export function restoreScrollPositionOrScrollToTop() {
 
     queueMicrotask(() => {
         queueMicrotask(() => { // Double microtask here to make sure scrolling restoration is the LAST thing to happen. (Even after Alpine's x-init functions)...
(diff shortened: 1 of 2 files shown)


ghabriel25:
Run all related PR tests against this and all green, closing mine 👍

PR #10704, opened 2026-09-15
[4.x] Support direct assignment to virtual properties

prateekbhujel:
Right now if you assign to a virtual property directly inside an action like `$this->selected = new Selection(...)` or `$this->items = collect(['a'])`, it creates a dynamic property on the component instead of updating the virtual property.

Because `Component` has `__get()` and `__unset()` for virtual properties but no `__set()`, a few things break:
1. The assignment bypasses `setVirtualProperty()` and its `TypeError` check.
2. It assigns to a dynamic property which triggers a deprecation notice on PHP 8.2+.
3. During dehydration `InteractsWithProperties::all()` reads `$this->getVirtualProperties()`, which still has the old value from `$this->__virtualProperties` and overwrites the dynamic property, so the assignment is silently lost on the next request.

This adds `__set()` to `Component.php` to route direct assignments through `setVirtualProperty()`, matching `__get()` and `__unset()`. Added tests covering roundtrip persistence, type errors on invalid assignment, and `data_set()`.


calebporzio:
nice

PR #10715, opened 2026-09-15
[4.x] Check all matching events in callback assertions

calebporzio:
When a component dispatches the same event twice, `assertDispatched()` with a callback only checks the first occurrence. For example, after dispatching `foo(id: 1)` and `foo(id: 2)`, these equivalent assertions disagree:

```php
$component->assertDispatched('foo', id: 2); // Passes
$component->assertDispatched('foo', fn ($event, $params) => $params['id'] === 2); // Fails
```

The inverse callback assertion, `assertNotDispatched()`, also incorrectly passes for that second event. This changes callback matching to search all events with the requested name and stop at the first match, consistent with named-parameter assertions.

Related report: https://github.com/livewire/livewire/discussions/8614. Its general concern is inconsistent parameter matching; the reproducible case addressed here is multiple occurrences of the same event with a callback. Credit to @utsavsomaiya, who previously proposed this fix as part of the closed #10046. This PR isolates that bug fix with regression coverage.

Both new regression tests failed before the change. Afterward, the event and testing suites pass: **77 tests, 269 assertions**, PHP 8.4.6 / Laravel 13.31.0.

```sh
vendor/bin/phpunit src/Features/SupportEvents/UnitTest.php src/Features/SupportTesting/UnitTest.php
```

Coverage includes matching first and later occurrences, ignoring other event names, missing events, and rejecting `assertNotDispatched()` when a later occurrence matches.


utsavsomaiya:
Wohooo 🥳!

PR #10718, opened 2026-09-15
[4.x] Prevent DOMDocument ValueError in PHP 8.4+ and null body warning on empty markup

prateekbhujel:
Hey @calebporzio — while looking closely at `ext/dom` behavior across PHP 8.4+:

In PHP 8.4+, `DOMDocument::loadHTML()` validates string arguments strictly and throws:
`ValueError: DOMDocument::loadHTML(): Argument #1 ($source) must not be empty` when given an empty or whitespace-only string (whereas PHP 8.1–8.3 silently warned or returned false).

Additionally:
1. When `$html` is `null` or non-string, `preg_replace()` emits `Deprecated: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated` on PHP 8.1+.
2. When a rendered component view produces only HTML comments (e.g. `<!-- comment -->`) or stripped `<script>`/`<style>` tags, `loadHTML()` produces a document without a `<body>` element. Accessing `$body->childNodes` then emits `Attempt to read property "childNodes" on null`.

This guards `SupportMultipleRootElementDetection::getRootElementCount()`:
1. Returns `0` early if `$html` is not a string or empty after trimming (before regex parsing).
2. Returns `0` if `$body` is null before iterating over child nodes.

Added unit tests covering `null`, empty string, whitespace, and comment/script-only templates.


prateekbhujel:
Thanks @calebporzio. The way you narrowed this to the cases that can actually reach `getRootElementCount()`, then covered both the real component behavior and the direct count contract, made the fix much tighter. I learned a lot from how you scoped it before adding the guards. Appreciate you taking it through.

Also interesting to see Fable used for the refinement in those follow-up commits.

PR #10721, opened 2026-09-16
[4.x] Memoize computed properties that return null within a request

fgilio:
Hi!

I got bit by this in one of our apps, and then saw that it was actually affecting several other apps too.

I think the fix is simple and safe, it was talked about [here](https://github.com/livewire/livewire/discussions/3164) and a similar fix was implemented [here](https://github.com/livewire/livewire/pull/10183).

---

A `#[Computed]` property that returns `null` is never memoized. Its body re-runs on every read within the same request. Non-null values memoize fine.

The per-request memo in `BaseComputed::handleMagicGet()` is stored with `??=`:

```php
$returnValue(
    $this->requestCachedValue ??= $this->evaluateComputed()
);
```

`??=` only assigns when the left side is null, so an honest `null` result (an `->first()` that finds nothing, an optional relation, a "current item" lookup with no match) looks the same as "not evaluated yet". Reading `$this->foo` four times in one request evaluates `foo()` four times. With this change it's once.


fgilio:
Wohoooo!! Thanks, @calebporzio!

